PR #202, opened 2022-03-22
sip: Application Configuration

lann:
[Rendered](https://github.com/fermyon/spin/blob/config-sip/docs/content/sips/000-app-config.md)

Issues to resolve:
- [x] Document that circular references are invalid
- [x] Document rationale for config "tree" (to support third-party components w/ external schema)
- [x] Clarify split between "schema" and "value" sections; maybe don't use "simple" form for schema (?)

### DIFF
--- a/docs/content/sips/000-app-config.md
+++ b/docs/content/sips/000-app-config.md
@@ -1,0 +1,161 @@
+title = "SIP xxx - Application Configuration"
+template = "main"
+date = "2022-03-22T14:53:30Z"
+
+---
+
+Summary: A configuration system for Spin applications.
+
+Owner: lann.martin@fermyon.com
+
+Created: March 22, 2022
+
+## Background
+
+It is common for applications to require configuration at runtime that isn't known at build time or is too sensitive to be stored with build artifacts.
+
+A few examples:
+
+- Logging configuration
+- Per-channel (production, staging, etc) service dependency URLs
+- Database secrets
+
+## Proposal
+
+### Configuration is a tree, with the app at its root
+
+e.g.:
+- app_key
+- component1
+  - key1
+  - key2
+- component2
+  - key3
+  - nested-component
+    - key4
+
+### Each key has a "path" corresponding to its position in the tree
+
+- e.g. `component2.nested-component.key4`
+
+- Config keys are restricted to allow unambiguous encoding as environment variables
+  / file paths:
+  - Start with a letter (required for env vars)
+  - Only lowercase ascii alphanum and `-` (could be `_` instead but only one of those)
+    - Only one `-` at a time (to allow joining paths as env vars with `__`)
+
+### Configuration "schema" is defined by the app and components
+```toml
+[config] # [application.config]?
+key1 = "default_value"
+# ...is equivalent to:
+key1 = { default = "default_value" }
+
+# "required" fields must be given a value
+key2 = { required = true }
+
+# "secret" field values should be handled with care (e.g. not logged)
+key3 = { required = true, secret = true }
+
+[[component]]
+id = "component1"
+...
+# Reusable components would define their config in their own component manifest
+[[component.config]]
+key1 = "value"  # Path: component1.key1
+...
+# Component dependency config can be overridden
+[[component.dependencies.nested-component]]
+key4 = "override"
+```
+
+### Configuration defaults can reference other keys
+```toml
+[config]
+app_key = "app_value"
+app_key2 = "prefix-{{ app_key }}-suffix"  # "prefix-app_value-suffix"
+
+[[component.config]]
+key1 = "value1"
+# Component config can reference root (app) keys:
+key2 = "root_ref={{ app_key }}"   # "root_ref=app_value"
+# and its own keys:
+key3 = "self_ref={{ .key1 }}"     # "self_ref=value1"
+
+[[component.dependencies.nested-component.config]]
+own_key = ""
+# Dependency config can reference root, parent (dependant), and own keys:
+root_ref = "{{ app_key }}"
+parent_ref = "{{ ..key1 }}"
+self_ref = "{{ .own_key }}
+```
+
+### Configuration "providers" populate configuration
+
+_I haven't thought this through too much. Would this live in `spin.toml`? How is config resolved across multiple providers?_
+```toml
+[[config_provider]]
+type = "toml_file"
+path = "config.toml"
+
+[[config_provider]]
+type = "environment"
+# Some config might not be appropriate for some providers
+exclude = "component2.*"
+```
+
+#### Example providers
+
+- Environment provider
+  - `component2.nested-component.key4` -> `SPIN_APP__COMPONENT2__NESTED_COMPONENT__KEY4`
+  - 😬 _Boy howdy thats ugly! Better ideas?_
+- File provider
+  ```toml
+  [component2.nested-component]
+  key4 = "value"
+  ```
+- Vault provider
+
+### Configuration is exposed via component interface
+
+`spin-config.wit`
+```
+// Missing key is a runtime error
+get_config: function(key: string) -> string
+```
+- Since each component gets its own instance of the `spin-config` import, the executor can resolve paths automatically and only expose a component's own config to it.
+
+### Other design options
+
+#### Typed configuration
+
+The above assumes only string values, but we could include some typing:
+```toml
+# Simple form is typed implicitly from its default value
+string_key = "value"
+string_key = { type = "string", default = "value" }
+
+number_key = 123
+number_key = { type = "number", default = 123 }
+
+required_string = { type = "string", required = true }
+# Would require e.g. base64 encoding in some places
+encryption_key = { type = "bytes", required = true, secret = true}
+```
+
+- This would complicate everything for _us_ but is nicer for users.
+
+#### WASI "configfs"
+
+For languages without component support, we could expose config as synthetic mounted files:
+
+```ruby
+key1_value = File.read("/config/key1")
+
+# Typed config
+key1_value = JSON.parse(File.read("/config/key1.json"))
+
+# "bytes" type
+encryption_key = File.read("/config/encryption_key.raw")
+```
+  
--- a/docs/content/sips/000-app-config.md
+++ b/docs/content/sips/000-app-config.md
@@ -27,9 +27,8 @@ It is common for applications to require configuration at runtime that isn't kno
 Configuration within a "parent" (component or application) consists of a number of configuration "slots":
 - slots are uniquely identified within their parent by a string "key"; in order to allow unambiguous conversion to environment variables or file paths, keys are constrained:
   - must start with a letter (required for env vars)
-  - consisting of only lowercase ascii alphanum and `-`
-    - only one `-` at a time and not at the end (to allow separating env vars with `__`)
-    - `-` matches WIT syntax but `_` might be more familiar to users
+  - consisting of only lowercase ascii alphanum and `_` (`[a-z0-9_]`)
+    - only one `_` at a time and not at the end (to allow delimiting in env vars with `__`)
 - a slot must _either_ be marked as "required" _or_ must be given a default value
 - a slot may be marked as "secret", in which case any associated value should be handled with care (e.g. not logged)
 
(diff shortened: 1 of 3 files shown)


fibonacci1729 (dismissed):

michelleN (dismissed):

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,137 @@
...
+`spin.toml`:
+```toml
+[config]
+app_root = "/app"
+...
+[[component.config]]
+# Note: no '.'s needed when referencing top-level app config
+work_root = "{{ app_root }}/work"  # -> "/app/work"
+[[component.dependencies.dep1.config]]
+dep_root = "{{ ..work_root }}/dep" # -> "/app/work/dep"
```
vdice, line 64: Do we need an explicit mention around how to reference 'ancestor' configs?

I see `..work_root` in this example, so it appears that a dep directly under an ancestor (one level down) can access an ancestor's config via `..<key>`.  Does it follow that a dep two levels down would use `...<key>`?  Or perhaps "two levels down" (or more) is not a valid use case and not supported...

lann: We don't even support dependencies yet, so I don't know exactly how this will end up working, but the code I'm working on allows any depth (up to a hard-coded limit of 1000 at the moment :slightly_smiling_face:).

These relative refs are a convenience to reduce duplication; ultimately all references will have to resolve to literal strings and top-level keys (i.e. those in `[config]`) and I expect that to be the most common way to use references.

lann: Ah but I see I don't mention it at all in the prose up here. Yeah I'll add something.

vdice (approved):

docs/content/sips/000-app-config.md
lann: OK, there are more words now.

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,143 @@
...
+
+`spin.toml`:
+```toml
+[config]
+app_root = "/app"
+log_file = "{{ app_root }}/log.txt"    # -> "/app/log.txt"
+...
+[[component.config]]
+work_root = "{{ app_root }}/work"      # -> "/app/work"
+work_out = "{{ .work_root }}/output"   # -> "/app/work/output"
```
itowlson, line 68: I wonder about the value of this.  What if we had a rule that app settings propagated to components (and there were rules against shadowing)?  I suppose that would cause issues for code the app author didn't control.  But this feels a bit fiddly, like having to go edit stuff if a config setting moved from component level to app level because we started using it in a new place.

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,143 @@
...
+key1 = { default = "default_value" }
+# required & secret slot 
+key2 = { required = true, secret = true }
+```
+
+Default values can use template strings to reference other slots.
+```toml
+[config]
+key1 = { required = true }
+key2 = "prefix-{{ key1 }}-suffix"
```
itowlson, line 49: What happens if there is a circular reference?  Do we assume that something will have a value injected to break the cycle, and then derive the remainder from there?  (And error if everything is left at default.)  This feels like it could get complicated, identifying which direction to propagate depending on which value is set.  Would it be safer to mandate that the graph of default value references must be acyclic?

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,143 @@
...
+string_key = { type = "string", default = "value" }
+
+number_key = 123
+number_key = { type = "number", default = 123 }
+
+required_string = { type = "string", required = true }
+# "bytes" would require e.g. base64 encoding in some places
+encryption_key = { type = "bytes", required = true, secret = true}
+```
+- This would complicate the implementation but might be nice for users.
```
itowlson, line 128: Expected types and conversion semantics might vary from language to language too, not sure.  Like "number" - what does that translate to in JS vs AssemblyScript vs C?

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,143 @@
...
+
+For languages without component support, we could expose config as synthetic mounted files:
+
+```ruby
+key1_value = File.read("/config/key1")
+
+# Typed config; `.json` encodes values to JSON
+key1_value = JSON.parse(File.read("/config/key1.json"))
+
+# "bytes" type; `.raw` decodes from base64
```
itowlson, line 140: I would have extremely mixed feelings about using the file extension to communicate desired type coercion/conversion... I see the value in helping with conversions but this feels like it might be too cunning for its own good.

A simpler approach initially would seem to be to use environment variables.  Is the value of files that the contents can be injected lazily (e.g. not fetching a secret unless it's needed)?

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,143 @@
...
+### Configuration "providers" resolve application configuration
+
+When resolving the value of an application configuration slot, providers will be queried in-order for a value. If no value is returned by any provider, the resolution will either use the default value or fail (if the slot is "required").
+
+Provider configuration is handled by spin at instantiation time (`spin up`).
+
+_Note: Provider configuration is TBD; as TOML it could look like:_
+
+```toml
+[[config-provider]]
```
itowlson, line 82: You show this using multiple-table syntax, implying there can be more than one provider.  Maybe worth illustrating why this has value e.g. wanting to keep credentials in a secure vault, but directory names as EVs set by Hippo.

What happens if two providers give values for the same config item?  Error, or first one wins, or last one wins?  Does this mean we need to check every provider for every config item, _just in case_ it has a value?  First or last wins is potentially brittle to someone "tidying" the provider list and not testing well enough 'because it's only tidying up.'

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Configuration "providers" resolve application configuration
+
+When resolving the value of an application configuration slot, providers will be queried in-order for a value. If no value is returned by any provider, the resolution will either use the default value or fail (if the slot is "required").
+
+Provider configuration is handled by spin at instantiation time (`spin up`).
+
+_Note: Provider configuration is TBD; as TOML it could look like:_
+
+```toml
```
itowlson, line 81: Where does this provider configuration live?  It is presumably separate from the spin.toml - I assume that as TOML it would be a distinct file, but it could also be passed as a list of providers on the command line, etc?  Maybe make the intent more explicit?

docs/content/sips/000-app-config.md
itowlson: I guess the whole hierarchy thing feels a bit unnecessary but I assume it's for components with config items with the same name but you want them to have different values?  Might be good to put a few words in about the motivation.

itowlson (changes requested):
Sorry it took me so long to get to this!  Looks good but a few suggestions/questions.

lann:
> Sorry it took me so long to get to this! Looks good but a few suggestions/questions.

Thanks for the careful review! I'll take a closer look tomorrow, but I'm realizing that part of the reason this is confusing is that we don't currently have a "component manifest", so I'm shoehorning component schema into `[[components.config]]`.

I'd like to have two different kinds of config sections which could appear in both app and component manifests:
- one "config schema" (`[config]`) section per app/component manifest that defines its "public" configuration; this is where `default`/`required`/`secret` would be used. I still like the "simple form" here but I can see the argument to avoid it
- a "config values" (`[[components.dependencies.config]]`?) section per dependency, which must confirm to that component's schema; here the entries really just have a single value/template so the simple form would probably be the only form.

I'm conflating the two in this design currently which is indeed confusing. I'll think about whether this can be improved without spec'ing component manifests as well.

itowlson:
> a "config values" (`[[components.dependencies.config]]`?) section per dependency, which must confirm to that component's schema

Okay, I am going to be _super_ pedantic here, to make sure I am understanding this right: the per-dependency section must conform to the _dependency's_ schema, right?

If I've understood that right, I'm a bit unclear on why a component config expresses schema and a dependency config expresses values.  Why is the person configuring the app not allowed to configure the settings required by dependencies?  Is it because the 'main' component wants to control how the dependency is configured, e.g. making sure it uses the same object store?  Is this true of all config items though, e.g. does the main component need to care where the dependency puts its logs, or where its gets the credentials for its internal database?  Is the intent that in "don't care" cases the main component would expose config items that the dependency directly copied?  This makes sense in terms of inulating the application configurer from changes to dependencies - it would be on the application author to re-map, or declare a breaking change, or whatever.

Okay, I _think_ I've talked myself round to what motivates your statement.  But we could definitely do with capturing that line of thought (if indeed that is the line of thought).  And having the same syntax for an overridable default and a set value does feel like a possible source of confusion.

lann:
> having the same syntax for an overridable default and a set value does feel like a possible source of confusion.

Agreed; I'm adding a list of "issues to resolve" with this to the top comment to track.

docs/content/sips/000-app-config.md
lann: Yes, I should specify that circular references are invalid. The prototype code enforces this with a simple recursion depth limit.

docs/content/sips/000-app-config.md
lann: Yeah, this is to support "third-party" components. I'll add words.

docs/content/sips/000-app-config.md
```diff
@@ -0,0 +1,143 @@
...
+[config]
+app_root = "/app"
+log_file = "{{ app_root }}/log.txt"    # -> "/app/log.txt"
+...
+[[component.config]]
+work_root = "{{ app_root }}/work"      # -> "/app/work"
+work_out = "{{ .work_root }}/output"   # -> "/app/work/output"
+[[component.dependencies.dep1.config]]
+dep_root = "{{ ..work_root }}/dep"     # -> "/app/work/dep"
```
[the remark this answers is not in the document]
lann: This would be defined in the hypothetical "component manifest"'s `[config]`.

docs/content/sips/000-app-config.md
lann: Currently:
> providers will be queried in-order for a value

This is an area where I expect us to make changes in the future. Just shooting for a reasonable starting point here.

docs/content/sips/000-app-config.md
lann: Everything down here is under the "Future design options" section, so I'm not trying to fully specify it. My vague idea on types is that they would be :wave:mostly JSON-compatible:wave: (perhaps excluding `null` and including some non-unicode byte string type), thus punting the problem to JSON libraries.

docs/content/sips/000-app-config.md
lann: > A simpler approach initially would seem to be to use environment variables. Is the value of files that the contents can be injected lazily (e.g. not fetching a secret unless it's needed)?

Yes, environment variables have to be loaded eagerly, which is fine for "local" config providers (env, files) but is probably unworkable for remote (etcd, vault) providers. I actually almost scuttled the whole idea of unifying config and secrets when I realized this problem. :grimacing:

radu-matei (dismissed):
I think this is a great starting point.
What do you think an MVP for this would look like?

I think this is in a state where we can merge and iterate as we discuss specific implementation details.

lann:
> I think this is a great starting point. What do you think an MVP for this would look like?

I've been working on implementation in parallel: https://github.com/fermyon/spin/compare/lann/config-wip

> I think this is in a state where we can merge and iterate as we discuss specific implementation details.

Agreed; I'll make one more update to address some of @itowlson's comments and then merge.

radu-matei:
@itowlson would you mind re-reviewing this, please?

docs/content/sips/002-app-config.md
```diff
@@ -0,0 +1,141 @@
...
+- Top-level references use just the key name: `{{ top_level_key }}`
+- "Sibling" references use a single `.` prefix: `{{ .sibling_key }}`
+- "Ancestor" references use multiple `.`s: `{{ ..parent_dep_key }}`, `{{ ...grandparent_key }}`
+- Circular / infinitely recursive references are not permitted
+
+`spin.toml`:
+```toml
+[config]
+app_root = { default = "/app" }
+log_file = "{{ app_root }}/log.txt"    # -> "/app/log.txt"
```
itowlson, line 63: Is this a fixed value, or a sneaky instance of the old syntax for defaults?

itowlson (approved):
Nice!  One nit that could do with clarifying, but otherwise good to go!

docs/content/sips/002-app-config.md
lann: dagnabbit

lann:
@radu-matei Now my small update cleared the _approvals_, too...

radu-matei (approved):

radu-matei:
Yeah, that seems to be a bit aggressive.
Going back to not dismissing reviews at all..

PR #247, opened 2022-03-24
Statically link `openssl` using vendored feature

### LINKED ISSUE #240: Statically link OpenSSL in the Spin binary
On macOS, we keep hitting linking issues that prevent us from running Spin without installing OpenSSL 1.1.1 and manually linking it. For example, we have the following in the docs:

> On an M1 macOS machine you might need to install / configure OpenSSL@1.1 by running `brew install openssl@1.1 && sudo ln -s /opt/homebrew/Cellar/openssl@1.1/1.1.1m /usr/local/openssl-aarch64`

We should figure out a way to statically link OpenSSL.

fibonacci1729:
close #240


lann:
@dicej I think this is the most straightforward way to statically link openssl

PR #273, opened 2022-03-28
SIP: Spin templates

itowlson:
WIP SIP for post-0.1 templates

### DIFF
--- a/docs/content/sips/draft-000-templates.md
+++ b/docs/content/sips/draft-000-templates.md
@@ -1,0 +1,287 @@
+title = "SIP DRAFT-000 - Templates"
+template = "main"
+date = "2022-03-22T14:53:30Z"
+
+---
+
+Summary: Proposes changes to the `spin new` syntax and to the template
+authoring format
+
+Created: March 28, 2022
+
+Status: DRAFT
+
+## Background
+
+We have an initial cut of a template system planned for 0.1, but design and UX
+work on this was somewhat constrained by the 0.1 deadline.  The 0.1 design is
+closely based on `cargo generate` because it was a readily available, er,
+template in the language we were using.  This proposal discusses goals and
+design options for taking the template system forward.
+
+The existing work gives us a solid basis with a well-documented template syntax
+using the Liquid engine, safe failure modes via staging, and other good
+stuff.  No changes are proposed to that core.  This SIP relates to user
+experience (for both consumers and authors) and broader architecture.
+
+## Goals
+
+* Convenient, forgiving user experience for consumers
+* Foundation for delivery via IDE
+* Minimise what authors need to know
+
+## Should we have a template system at all?
+
+It's worth considering this because there are a ton of existing templating
+systems, some language specific (e.g. `cargo generate`, `dotnet new`), some
+generic (e.g. Yeoman).  These are likely to be idiomatic to the languages
+people use.  Is it worth having our own?
+
+Reasons why we should:
+
+* Some languages _don't_ have template support, or require the user to go
+  install yet more tools.
+* It allows documentation to assume Spin templates instead of having to
+  write different docs for every language.
+* Having our own system doesn't preclude offering Spin templates via
+  language-idiomatic systems too.
+
+Reasons why we shouldn't:
+
+* It's a decent chunk of work to do it well.
+* It's extra code for every copy of Spin to carry around.
+
+I believe the general consensus is that the case for having templates built
+in is compelling, but it's always worth checking that assumption.
+
+## Identifying a template
+
+Templates are currently identified by a repository ID and directory name.
+Because Spin is a multi-language runtime, it would be good to also identify
+templates by language, so that a user could easily see and choose templates
+in their preferred language.
+
+The question of the repository ID is what we want "adding" a template to mean.
+Does adding a template:
+
+* Record a mapping of a repository ID to a Git location
+* Copy a template to the local system
+
+At the moment it seems to do both, which I found a bit confusing. If we take
+the view that a template is copied, then we no longer need the repository
+ID and can address templates by their names along.
+
+This, by the way, is the approach taken by the `dotnet` templating system,
+where you run `dotnet new --install <package-name>` to install a template
+package, and then use template by running `dotnet new <template-name>`,
+with an optional `-lang` option to specify the language.
+
+Because this does not qualify every template with a unique Git URL, it does
+require some consideration of name clashes.  There may also be value in
+allowing users to run templates without installing them.
+
+## User experience: starting a template
+
+The `spin new` command create a new Spin application.  (We may want to consider
+also allowing it to create a new component in an existing application.)
+
+Proposed syntax:
+
+| Command        | Behaviour |
+|----------------|-----------|
+| `spin new`     | Prompt from installed templates, then run chosen template |
+| `spin new http` | Runs the `http` template. If this is available in more than one language, prompts for language or uses default language |
+| `spin new http -lang haskell` | Runs the Haskell `http` template |
+| `spin new -lang haskell` | Prompt from installed Haskell-language templates, then run chosen template (maybe not useful?) |
+
+### What if there are no templates?
+
+We don't have an installer so we can't guarantee to have templates present the
+first time the user runs `spin new`.  The `cargo generate` experience for this
+is to print an error about a config file path; this is not good.
+
+If the user runs `spin new` (with any options) and we do not have any templates
+installed, we should offer to install the default ones from Spin itself.  We should
+also point them at the `spin templates install` command to add more.
+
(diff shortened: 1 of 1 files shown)


docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+Templates are currently identified by a repository ID and directory name.
+Because Spin is a multi-language runtime, it would be good to also identify
+templates by language, so that a user could easily see and choose templates
+in their preferred language.
+
+The question of the repository ID is what we want "adding" a template to mean.
+Does adding a template:
+
+* Record a mapping of a repository ID to a Git location
+* Copy a template to the local system
```
lann, line 68: It seems to me that this should be handled by caching (for remote git repos).

docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+| `spin new -lang haskell` | Prompt from installed Haskell-language templates, then run chosen template (maybe not useful?) |
+
+### What if there are no templates?
+
+We don't have an installer so we can't guarantee to have templates present the
+first time the user runs `spin new`.  The `cargo generate` experience for this
+is to print an error about a config file path; this is not good.
+
+If the user runs `spin new` (with any options) and we do not have any templates
+installed, we should offer to install the default ones from Spin itself.  We should
```
lann, line 104: If "added" templates were just repo refs (first option up yonder :point_up:), these defaults could be prepopulated.

docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+also point them at the `spin templates install` command to add more.
+
+### Template search strategy
+
+If the user runs `spin http foo`:
+
+* If there is NO default language set:
```
lann, line 111: Is a default language required? Seems like you could deploy e.g. bartholomew without caring what language it is written in.

lann: Oh look theres a whole section for it...

docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+
+```
+spin new https://github.com/fermyon/spin/templates/http/rust
+```
+
+There would be no language option.
+
+We'd need to enforce that IDs couldn't look like URLs.
+
+### What's this 'default language' thing?
```
lann, line 172: Is a default language required? Seems like you could deploy e.g. bartholomew without caring what language it is written in.

itowlson: Oh that's interesting - you're thinking of templates with no source code, just references to existing Wasm modules?  (With the static files being the reason you need a separate application.)

I mean... no.  A default language isn't required.  It is a vague thought to save people having to type `-lang rust` EVERY. DAMN. TIME.  But yeah, if we did it, we would need to figure out how to make it interact with "no language" templates.  Thanks for the catch.

docs/content/sips/draft-000-templates.md
itowlson: You mean cloning the entire repo into the local environment?  That's what we do at the moment; it feels heavy to me but it's absolutely a viable way.

lann: I just mean _if_ there is some compelling (performance?) reason to have a local copy, treat it as a cache rather than the primary result of "adding" a template.

itowlson: I'm not too concerned about performance, though the offline aspect bears thinking about (ye olde "working on my conference demos on the plane to the conference" scenario).  What I'm concerned about is convenience - being able to type `spin new http` rather than bashing out GitHub URLs - and discoverability - being able to browse the templates one has "installed" or "favourited" or however we choose to think of it, rather than having the whole of URL-space to choose from _grin_.  I should probably be clearer in tying the usage experience back to the "goals" section.

docs/content/sips/draft-000-templates.md
itowlson: I think the source of the confusion might be that I'm envisaging this as a filter/selector on templates that declare what language their code is in, and it's coming across as "we will convert the template content into Haskell for you using _checks notes_ Brian"

Is there a clearer term I could use for this?  Nothing concise is leaping to mind...

lann: Could be filtering on "tags", with special UI for tags that look like e.g. `lang:haskell`.

itowlson: Could be but I'm not sure what other use cases there would be for this _grin_.  Would you be able to expand on what you're thinking of? (or can jump on a call if you want to think out loud at me!)

lann: Wildly speculating: trigger type tags (`http`, `redis`), or a `cms` tag for Bartholomew et al.

itowlson: We probably need trigger type as a first class citizen anyway, so that when adding a component to an app it can offer compatible options.  Could be worth implementing it using a general mechanism for sure!

itowlson: One thing about encouraging authors to provide search keywords today, even if they're not terribly useful with the small number of templates we have today, is that if we set up a templates registry, it will be usefully searchable from day one - we won't need to bribe authors with cake to go back and add keywords to their templates.

lann: Especially if the UI can "nudge" template authors towards appropriate tags.

docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+
+Created: March 28, 2022
+
+Status: DRAFT
+
+## Background
+
+We have an initial cut of a template system planned for 0.1, but design and UX
+work on this was somewhat constrained by the 0.1 deadline.  The 0.1 design is
+closely based on `cargo generate` because it was a readily available, er,
```
radu-matei, line 18: "er"?

docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+  install yet more tools.
+* It allows documentation to assume Spin templates instead of having to
+  write different docs for every language.
+* Having our own system doesn't preclude offering Spin templates via
+  language-idiomatic systems too.
+
+Reasons why we shouldn't:
+
+* It's a decent chunk of work to do it well.
+* It's extra code for every copy of Spin to carry around.
```
radu-matei, line 52: Considering our early discussions around splitting the runtime part of Spin, and distributing the templates implementation only for developers, would this still be a concern?

docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+first time the user runs `spin new`.  The `cargo generate` experience for this
+is to print an error about a config file path; this is not good.
+
+If the user runs `spin new` (with any options) and we do not have any templates
+installed, we should offer to install the default ones from Spin itself.  We should
+also point them at the `spin templates install` command to add more.
+
+### Template search strategy
+
+If the user runs `spin http foo`:
```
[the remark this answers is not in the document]
radu-matei: I don't think the doc has been updated with `spin new foo`?

docs/content/sips/draft-000-templates.md
radu-matei, line 111: Grouping by language seems to add a lot of complexity to our implementation, and overhead for a user creating an application.
It seems like grouping by language might be the obvious default, but what if instead I want to group templates by their trigger type, for example? 

Not saying that is what we should do, only that we should make sure grouping by language is the obvious default.

docs/content/sips/draft-000-templates.md
radu-matei: > A default language isn't required. It is a vague thought to save people having to type -lang rust EVERY. DAMN. TIME.

Could we _just_ rely on template IDs and/or paths in a repo to identify templates, and leave grouping (whether to group them by tags, languages, or something else) to the template authors?

docs/content/sips/draft-000-templates.md
```diff
@@ -0,0 +1,365 @@
...
+
+```
+spin templates install --dir <PATH>
+spin templates install --git <URL> [--branch foo]
+```
+
+The current convention for Git sources is that the templates should be located
+at `/templates`.  I'm not sure whether directory sources expect a `templates`
+subdirectory or if you specify the full path - I thought the former but it
+seemed to work when I used the latter. But maybe I was wrong.
```
radu-matei, line 199: The current convention (which has not been thought out carefully) is to expect a `templates` directory _in the root of the repository_.

radu-matei:
We have had templates support for a while now — do we want to merge this PR?

PR #307, opened 2022-04-01
Application Configuration

lann:


Cargo.toml
```diff
@@ -56,3 +57,6 @@ members = [
 [[bin]]
 name = "spin"
 path = "src/bin/spin.rs"
+
+[patch.crates-io]
+bindle = { path = "../bindle" }
```
lann, line 62: https://github.com/deislabs/bindle/pull/329

lann:
https://github.com/fermyon/spin/blob/950e8c64a8155c4599d9c32150af6566096fed0b/examples/wip-appconfig/spin.toml#L8-L17

https://github.com/fermyon/spin/blob/950e8c64a8155c4599d9c32150af6566096fed0b/examples/wip-appconfig/src/lib.rs#L14-L19

```sh
$ curl 127.0.0.1:3000/hello 
Hello, Fermyon
```

lann:
```sh
$ SPIN_APP_NAME=Local spin-release up
```
```sh
$ curl 127.0.0.1:3000/hello
Hello, Local
```

tests/http/simple-spin-rust/src/lib.rs
```diff
@@ -12,24 +13,24 @@ impl spin_http::SpinHttp for SpinHttp {
...
+                Ok(text) => Response {
+                    status: 200,
+                    headers: None,
+                    body: Some(text.as_bytes().to_vec()),
+                },
+                Err(e) => Response {
+                    status: 500,
+                    headers: None,
+                    body: Some(format!("ERROR! {:?}", e).as_bytes().to_vec()),
+                },
```
lann, line 25: This whole diff chunk is just `rustfmt`.

lann:
I think this is enough for an MVP. There is one hard-coded config provider (`EnvProvider`, which understands `SPIN_APP_<KEY>`).

wit/ephemeral/spin-config.wit
```diff
@@ -0,0 +1,10 @@
+// Get a configuration value for the current component.
+// The config key must match one defined in in the component manifest.
+get-config: function(key: string) -> expected<string, error>
```
radu-matei, line 3: As we want to add values other than strings, do we want to have a `get-config-string` function instead?

crates/config/src/host_component.rs
```diff
@@ -0,0 +1,46 @@
+use std::sync::Arc;
+
+use crate::{Error, Key, Path, Resolver, Result};
+
+mod wit {
```
radu-matei, line 5: I like this a lot, we might want to namespace the other bindings as well.

crates/config/src/host_component.rs
```diff
@@ -0,0 +1,46 @@
...
+            resolver,
+        })
+    }
+}
+
+impl<'a> wit::spin_config::SpinConfig for ComponentConfig {
+    fn get_config(
+        &mut self,
+        key: &str,
+    ) -> std::result::Result<std::string::String, wit::spin_config::Error> {
```
radu-matei, line 30: nit: Would you mind adding `use` statements for standard library objects here? It makes reading things a bit easier.

crates/config/src/host_component.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub fn new(component_id: impl Into<String>, resolver: Arc<Resolver>) -> Result<Self> {
+        let component_root = Path::new(component_id)?;
+        Ok(Self {
+            component_root,
+            resolver,
+        })
+    }
+}
+
+impl<'a> wit::spin_config::SpinConfig for ComponentConfig {
```
radu-matei, line 26: nit: is the lifetime parameter here actually needed?

crates/engine/src/lib.rs
```diff
@@ -7,6 +7,8 @@ pub mod io;
 
 use anyhow::{bail, Context, Result};
 use io::IoStreamRedirects;
+use spin_config::host_component::ComponentConfig;
+use spin_config::Resolver;
```
radu-matei, line 11: nit: merge the two statements.

crates/engine/src/lib.rs
```diff
@@ -7,6 +7,8 @@ pub mod io;
...
 use spin_manifest::{Application, CoreComponent, DirectoryMount, ModuleSource};
 use std::{collections::HashMap, io::Write, path::PathBuf, sync::Arc};
 use tokio::task::JoinHandle;
```
radu-matei, line 14: nit-nit: could you also merge these two statements?

radu-matei (approved):
Just a few nits, but overall LGTM.

Thanks!

crates/config/src/host_component.rs
lann: I actually _had_ to (or at least this was the first solution that came to mind) because the generated bindings were named the same as the crate...

That said, yeah I think its a good pattern to make it obvious where the generated types are coming from.

crates/config/src/host_component.rs
lann: Whoops, this was vscode's generated trait impl for some reason... :thinking:

crates/config/src/host_component.rs
lann: Sure doesn't look like it, does it? I think I had a revision where `ComponentConfig` borrowed `Resolver` or something.

crates/engine/src/lib.rs
lann: Is there a tool we can use to enforce one `use` style?

lann: I'd be fine with https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#crate

lann: Oh its unstable-only :angry:

crates/config/src/host_component.rs
radu-matei: Alternatively, I think you could do `use <> as <>`.

PR #319, opened 2022-04-05
Rename config to manifest

### LINKED ISSUE #318: Rename `config` crate
After discussion about a better name for application config, we're going to rename the current `config` crate to `manifest` and use `config` for #307 unless we can come up with better names for those two things.

lann:
Fixes #318


lann:
> [Remove unused examples/spin-timer/Cargo.lock](https://github.com/fermyon/spin/pull/319/commits/f34641da4d6d2b54fa11b84766022de26d865d91)

@radu-matei I noticed the crate rename didn't affect this lock file, and I think it isn't being used at all because there is no empty `[workspace]` in the `Cargo.toml` making the example part of the workspace. Was that intended?

radu-matei (approved):

PR #327, opened 2022-04-06
ci: Only run cargo fmt / clippy on Linux

lann:
Linux runners consistently finish faster than Mac/Windows, so we'll omit
platform-agnostic checks from those two to make the CI process faster.


lann:
Funny how things start working when you use the _correct_ syntax...

Issue #181, opened 2022-03-18
Implement host support for connecting to a Redis database


radu-matei:
With #155, we have support for triggering  components for new messages on a queue of a given Redis instance.

It would be great if we had explicit support for connecting to a Redis instance (for both accessing the key/value store and writing on the queue).

Because [the Redis protocol](https://redis.io/topics/protocol) isn't implemented over HTTP, this would have to be a host implementation rather than another component.

Note that this would be separate from the generic host implementations (for a key/value store) we are evaluating for Spin.

PR #328, opened 2022-04-06
Implement Redis publish interface

fibonacci1729:
ref #181


technosophos:
Manually tested, and it appears to be working.

fibonacci1729:
@technosophos `get` & `set` are now supported. I'm working on tests now.

fibonacci1729:
The `outbound-redis` example contains an `HTTP` component that uses all three `redis` APIs.

crates/outbound-redis/src/lib.rs
```diff
@@ -0,0 +1,35 @@
+use outbound_redis::*;
+use redis::Commands;
+
+pub use outbound_redis::add_to_linker;
+
+wit_bindgen_wasmtime::export!("../../wit/ephemeral/outbound-redis.wit");
+
+/// A simple implementation to support outbound Redis commands.
+#[derive(Default, Clone)]
+pub struct OutboundRedis {}
```
radu-matei, line 10: nit: remove the `{}` if there are no fields associated with the struct.

radu-matei (approved):
This looks great!
Manually tested, and everything seems to work properly:

```
❯ redis-cli monitor
OK
"GET" "mykey"
"SET" "spin-example" "Eureka!"
"PUBLISH" "messages" "Hello Redis"
```

LGTM

wit/ephemeral/outbound-redis.wit
```diff
@@ -0,0 +1,13 @@
+use * from redis-types
+
+// Publish a Redis message to the specificed channel and return an error, if any.
+publish: function(address: string, channel: string, payload: payload) -> expected<_, error>
+
+// Get the value of a key. An error is returned if the value stored at key is not
+// a string.
+get: function(address: string, key: string) -> expected<option<string>, error>
```
radu-matei, line 8: I just realized that we should probably get and set byte arrays as opposed to strings.

fibonacci1729: return as well?

radu-matei: Yeah, that's what I was thinking.

wit/ephemeral/outbound-redis.wit
```diff
@@ -0,0 +1,12 @@
+use * from redis-types
+
+// Publish a Redis message to the specificed channel and return an error, if any.
+publish: function(address: string, channel: string, payload: payload) -> expected<_, error>
+
+// Get the value of a key. An error is returned if the value stored at key is not
```
radu-matei, line 6: The comment here (re: an error if the value is not a string) is no longer accurate.

PR #329, opened 2022-04-06
fix(docs): remove the content_encoding config for docs

flynnduism:
Removes the `gzip` config from the `site.toml`, which was breaking the site locally. 

Spin was running, but hitting localhost:3000 the browser couldn't render the site. It was showing a "Content Encoding Error"  that I've never seen before.

Signed-off-by: flynnduism <ronan@fermyon.com>

### DIFF
--- a/docs/config/site.toml
+++ b/docs/config/site.toml
@@ -1,7 +1,6 @@
 title = "Spin Documentation"
 base_url = "http://localhost:3000"
 about = "Spin — a framework for building fast, secure, and composable microservices with WebAssembly"
-content_encoding = "gzip"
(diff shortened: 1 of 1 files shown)


radu-matei:
I'm not sure why this is happening.
This is running fine for me locally, and I thought #285 fixed this.

Investigating.

technosophos:
I didn't even realize that line did anything anymore. You might just need to update your local copy of Spin

radu-matei (approved):
Given this is causing issues, I think we should remove it and track it in an issue.

LGTM

PR #335, opened 2022-04-07
ci(.github): add dispatch workflow

vdice:
- Adds a dispatch.yml GH workflow for detecting file changes and dispatching an event as appropriate
- First use case is to detect doc updates and then dispatch an event to a destination repo (which handles (re-)deploying the docs site)

Tested on fork: https://github.com/vdice/spin/actions/runs/2111223700

### DIFF
--- a/.github/workflows/dispatch.yml
+++ b/.github/workflows/dispatch.yml
@@ -1,0 +1,35 @@
+name: Dispatch Event
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


vdice:
(Note: build failing due to unrelated [issue](https://github.com/fermyon/spin/pull/336))

radu-matei (approved):

PR #336, opened 2022-04-07
chore(templates): appease clippy

vdice:
Noticed a clippy err on a recent PR.   Indeed, `main` CI failed with the same: https://github.com/fermyon/spin/runs/5871351506?check_suite_focus=true

Just updated per clippy's suggestion.  Open to different approach...


radu-matei:
Why did it fail on `main` in the first place is the real question..

radu-matei (approved):

lann:
Because this lint was just enabled by default with 1.60

radu-matei:
Ah, that _does_ make sense, interesting that GitHub Actions have the new version of Rust this fast.

Issue #298, opened 2022-03-30
Lightweight SDK for building AssemblyScript HTTP components


radu-matei:
ref https://github.com/fermyon/spin-kitchensink/blob/main/assemblyscript-outbound-http/index.ts

radu-matei:
We haven't had traction on this request, closing for now. We can reconsider this in the future if people are interested in an SDK for AssemblyScript — note that you can still build a Wagi component in AssemblyScript.

PR #338, opened 2022-04-08
feat: add lightweight AssemblyScript SDK for Wagi HTTP components

radu-matei:
ref #298 

This commit starts the work for a lightweight Spin SDK
for AssemblyScript.
Particularly, it starts the work on defining the HTTP
objects needed, and adds a work-in-progress for reading
request data from the environment (for Wagi components).
This is just a work-in-progress, as reading the request
body and writing the response body are currently not implented.

With this SDK, writing an HTTP component in AssemblyScript:

```typescript
export function _start(): void {
    handleRequest((request: Request): Response => {
        for (var i = 0; i < request.headers.size; i++) {
            Console.error("Key: " + request.headers.keys()[i]);
            Console.error("Value: " + request.headers.values()[i] + "\n");
        }
        return new ResponseBuilder(StatusCode.FORBIDDEN)
            .header("foo", "bar")
            .body(String.UTF8.encode("Hello, Fermyon!\n"));
    });
}
```

Note that this still follows the existing convention for
calling the `_start` function in the module for a Wagi
component.

(We are currently experimenting with going through the C bindings
for writing components in other languages, currently for CGO.
If that experiment is successful, it would allow us, for the list
of languages supported with an SDK, to stop relying on the Wagi
compatibility layer and fully implement all types of components.
That work is still early. Depending on its outcome, we will decide
how to approach the AssemblyScript SDK.)


sdk/as/http/cgi.ts
```diff
@@ -0,0 +1,85 @@
...
+    let method = requestMethodFromEnv();
+    let headers = requestHeadersFromEnv();
+    // let body = requestBodyFromEnv();
+
+    return new Request(uri, method, headers);
+}
+
+export function sendResponse(response: Response): void {
+    printHeaders(response);
+    // wasi.Descriptor.Stdout.write(changetype<u8[]>(response.body));
```
radu-matei, line 15: Reading the request body is currently not working.

sdk/as/http/cgi.ts
```diff
@@ -0,0 +1,85 @@
+import * as wasi from 'as-wasi';
+import { Method, Request, Response } from "./http";
+
+export function requestFromCgi(): Request {
+    let uri = requestUriFromEnv();
+    let method = requestMethodFromEnv();
+    let headers = requestHeadersFromEnv();
+    // let body = requestBodyFromEnv();
```
radu-matei, line 8: Reading the request body is currently not working.

sdk/as/http/http.ts
```diff
@@ -0,0 +1,238 @@
...
+            return Method.TRACE;
+        } else if (m == "PATCH" || m == "patch") {
+            return Method.PATCH;
+        } else {
+            return Method.GET;
+        }
+    }
+
+    /** Convert an HTTP method into a string. */
+    export function from(m: Method): string {
```
radu-matei, line 143: Terrible function name.

sdk/as/http/http.ts
```diff
@@ -0,0 +1,238 @@
...
+    DELETE,
+    CONNECT,
+    OPTIONS,
+    TRACE,
+    PATCH,
+}
+
+export namespace Method {
+    /** Parse a string into an HTTP method. */
+    export function parse(m: string): Method {
```
radu-matei, line 118: Terrible function name.

radu-matei:
TODO:

- [ ] decide whether this SDK will still follow the Wagi model for HTTP components
- [ ] fix request and response bodies
- [ ] add outbound HTTP
- [ ] add tests

lann:
I know `as` is a typical abbreviation in that ecosystem, but maybe give it a longer name here?

itowlson:
As a heads up, one thing to test for in this is handling of the special `_routes` entry point.  They believe they have fixed the issue with non-default entry points, but from their description (https://github.com/AssemblyScript/assemblyscript/pull/2157) I'm not sure it will work for WAGI.

(And this may be on us.  WAGI doesn't really treat the module as a reactor, but instead kinda treats it as two commands in a trenchcoat, which works okay with Rust but might be a bit ambiguous in terms of the WASI spec, I'm not sure.)

radu-matei:
Oh, this brings up an excellent point that we (read I) should have documented and/or implemented — see #342

radu-matei:
AssemblyScript recently removed WASI support from the core project, and we no longer support the old WASI experimental HTTP library, so defering this SDK for now.

Happy to revisit if people are interested.

ref #699

Issue #337, opened 2022-04-08
Component IDs can no longer contain the dash `-` character


radu-matei:
After #307, component IDs can no longer contain the dash `-` character:

For example, change the component ID to the following in any example:

```toml
[[component]]
id = "dash-test"
```

And you get the following _runtime_ error when trying to process a request:

```
ERROR spin_http_engine: Error processing request: invalid config key: invalid character '-'
```

cc @lann

lann:
Ahhh yeah it _should_ only be happening if you have a `[component.config]` on that component (which is itself perhaps a bug, but at least semi-intentional).

radu-matei:
We should definitely consider imposing that restriction on component IDs.

My initial confusion came from the fact that this was a "runtime" error, rather than a "startup" error, and that part is still a bit unclear from how the configuration currently works.

itowlson:
What's the status of this?  I have some components in a sample that have hyphens, and was about to update them to use config...

itowlson:
Answer: adding a config section gave me an error

lann:
I don't think the config system really needs this restriction any more. If we want to remove the restriction we can; it will just take some refactoring to the config `Tree` struct.

PR #339, opened 2022-04-08
Mitigate issue #337

lann:
A previous iteration of config design intentionally restricted component
IDs to be valid config keys. That restriction isn't necessary in the
current design but still exists.

This removes the restriction for components that don't use config.

ref #337


radu-matei (approved):
Manually tested with a component with dashes, this is working again, thanks!

LGTM.

PR #346, opened 2022-04-13
Implement a key/value store interface

### LINKED ISSUE #47: Implement a key/value store interface
Spin applications should have a default key/value store implementation available at runtime, ideally following https://github.com/fermyon/wasi-experimental-toolkit.

ecumene:
# A generic key/value Store interface for Spin (#47)

close #47

```
get: function(key: string) -> expected<option<list<u8>>, error>
set: function(key: string, value: list<u8>) -> expected<_, error> 
```

This is an interface for a key/value store for Spin components. The purpose of this store is to hold onto some data between triggers or store expensively computed values or share data between components. For local development, the keys will be persisted between runs via a JSON file, but in production, a [Redis](https://redis.io/) instance will be used as a backend.

## Persisting to disk

Persisting to disk is preferable for local development, but don't rely on this as a data store. Think of it as ephemeral cache for your components.

### Configuring Key/Value Store Path

This will store the `db.json` in your current directory
`spin up --kv-dir .`
`spin up --K .`

## Persisting to Redis

Persisting to Redis is preferable for deployments due to its performance and resiliency. 

### Configuring the Redis address

In your App Manifest, add this line to a component:
```diff
[[component]]
id = "key-value"
source = "target/wasm32-wasi/release/rust_key_value.wasm"
+redis_address = "redis://localhost/"
```

 - [x] Tests
 - [x] Redis/Json example


wit/ephemeral/key-value.wit
```diff
@@ -0,0 +1,12 @@
+// General purpose error.
+enum error {
+    success,
+    error,
+}
```
ecumene, line 5: I copied this from the `redis-types` types. But why would you need `Error::Success?`

radu-matei: Oh yeah, this was left over from older versions of WITX without `expected`.
We should probably update all interfaces to remove the `success` from error enums.

crates/key-value/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Default, Clone)]
+pub struct KeyValue {
+    pub(crate) db: HashMap<String, Vec<u8>>,
+}
+
+impl KeyValue {
+    pub fn load_json() -> Result<KeyValue, anyhow::Error> {
+        // TODO: Config this path
+        let mut file = match std::fs::File::open("db.json") {
```
radu-matei, line 18: Spin stores logs in `~/.spin/<appid>/logs/<componentid_<stderr>/<stdout>.txt`.

Would it make sense to store the JSON file by default in something similar to `~/.spin/appid/key-value/<>.json`, potentially with the possibility to override that when starting Spin?

radu-matei:
To answer the rest of your questions:

- we currently don't have explicit tests for the WIT interfaces. It would be great if we could add some, but I don't think it's a requirement to have it as part of this PR, as I think we will bikeshed over that for a bit.
- re: configuration, this might be a really good tie-in with #202 and #307 
- for the Redis backend, same as ^^, this would be set in the application configuration, and the implementation for it would be *really* similar to #328, which we could probably extract in a super simple crate.

This is starting to look great, thanks a lot for your contribution!

Mossaka:
Should we have a `delete: function(key: string) -> expected<_, error>` function in the interface?

ecumene:
> Should we have a `delete: function(key: string) -> expected<_, error>` function in the interface?

Good point!

Maybe we could also make the value in set optional alternatively. I'll wait for @radu-matei 's input.

radu-matei:
Having a `delete` function definitely makes sense to me.

@ecumene which value do you think should be optional?

ecumene:
```diff
get: function(key: string) -> expected<option<list<u8>>, error>
-set: function(key: string, value: list<u8>) -> expected<_, error> 
+set: function(key: string, value: option<list<u8>>) -> expected<_, error> 
```

radu-matei:
Ok, I see what you mean.
I think I prefer an explicit `delete` function instead here. What do you think?

ecumene:
`delete` is more clear on first glance 👍🏻

PR #347, opened 2022-04-13
feat(sdk/go/redis): Tinygo sdk for outbound redis

adamreese:
ref: #328

### DIFF
--- a/sdk/go/outbound_redis/Makefile
+++ b/sdk/go/outbound_redis/Makefile
@@ -1,0 +1,7 @@
+.PHONY: generate
+generate:
+	wit-bindgen c --import ../../../wit/ephemeral/outbound-redis.wit
+
+.PHONY: clean
+clean:
+	rm outbound-redis.h outbound-redis.c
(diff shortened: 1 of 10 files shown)


sdk/go/outbound_redis/Makefile
```diff
@@ -0,0 +1,7 @@
+.PHONY: generate
```
radu-matei, line 1: Could there be a single Makefile for the Go SDK that handles both the bindings generation and building / testing the SDK?

adamreese: Yes but I want to give testing and code management more though. I don't want it to hold up this PR.

sdk/go/outbound_redis/internals.go
```diff
@@ -0,0 +1,66 @@
...
+import "C"
+import (
+	"errors"
+	"unsafe"
+)
+
+func publish(addr, channel string, payload []byte) error {
+	a := redisStr(addr)
+	c := redisStr(channel)
+	p := C.outbound_redis_payload_t{ptr: &payload[0], len: C.size_t(len(payload))}
```
radu-matei, line 13: nit: could the variables here be a bit more explicit, please?
I can see it being very difficult to debug anyway, particularly given that it's almost entirely CGO.

fibonacci1729 (approved):

PR #351, opened 2022-04-13
Diagnostic for running Wagi component under Spin executor

itowlson:
If you accidentally forget the `executor` line on a Wagi component, you get the enigmatic error "failed to find function export `canonical_abi_free`".  This has happened to several of us _cough cough_ - for some of us _cough_ more than once _cough cough cough_.

This PR provides an additional diagnostic around the error message - a user viewing the logs at the default level will now see:

```
09:55 $ spin up 
Serving HTTP on address http://127.0.0.1:3000
2022-04-13T21:55:09.687838Z ERROR spin_http_engine: Error processing request: component is not compatible with Spin executor - should this use the Wagi executor?

Caused by:
    failed to find function export `canonical_abi_free`
```

Very open to improving the message, or to detecting the condition in a less hideous way!

### DIFF
--- a/crates/http/src/spin.rs
+++ b/crates/http/src/spin.rs
@@ -42,7 +42,9 @@ impl HttpExecutor for SpinHttpExecutor {
         let (store, instance) =
             engine.prepare_component(component, None, Some(io_redirects.clone()), None, None)?;
 
-        let resp_result = Self::execute_impl(store, instance, base, raw_route, req).await;
+        let resp_result = Self::execute_impl(store, instance, base, raw_route, req)
+            .await
+            .map_err(contextualise_err);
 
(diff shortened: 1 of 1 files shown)


radu-matei (approved):
Unless we start returning specific errors (as opposed to `anyhow::Errors`, which we should explore at some point), I don't see a prettier way to check for this error, so this LGTM.

crates/http/src/spin.rs
```diff
@@ -190,6 +192,18 @@ impl SpinHttpExecutor {
     }
 }
 
+fn contextualise_err(e: anyhow::Error) -> anyhow::Error {
+    if e.to_string()
+        .contains("failed to find function export `canonical_abi_free`")
```
lann, line 197: Gross! :+1:

PR #352, opened 2022-04-14
feat: add spin build command to run local build command for components

radu-matei:
ref #384

This commit adds a new top level Spin command that simplifies building more
complex applications. Specifically, this command will execute the build commands
set in each local component configuration:

```toml
[component.build]
command = "cargo build --target wasm32-wasi --release --manifest-path http-rust/Cargo.toml"
```

The build commands for all components in the local application manifest are
currently run sequentially (running multiple resource-intensive compilation processes
concurrently does not seem to be the most efficient choice currently).

Optionally, after all compilation processes are finished, if the `--up` flag is
passed, the application can be started without calling `spin up`.

This means that the entire developer "inner loop" for working on a Spin application
can be `spin build --up`.


crates/build/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+async fn build_component(raw: RawComponentManifest, src: impl AsRef<Path>) -> Result<()> {
+    match raw.build {
+        Some(b) => {
+            log::info!("Executing the build command for component {}.", raw.id);
+            let first = b
+                .command
+                .first()
+                .context("Component build command must have at least one item.")?;
+
+            Exec::cmd(first)
```
lann, line 32: This does PATH lookup?

crates/build/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+                .cwd(
```
lann, line 33: Doesn't need to be this PR, but it would be nice to be able to override this with e.g. `component.build.workdir`

crates/build/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+                    src.as_ref()
+                        .parent()
+                        .context("The application file did not have a parent directory.")?,
+                )
+                .args(&b.command[1..])
+                .stream_stdout()?;
```
lann, line 39: I'm guessing this will eat stdout? If a build process _has_ stdout you might want to see it... :shrug:

src/commands/bindle.rs
```diff
@@ -1,21 +1,11 @@
-use crate::parse_buildinfo;
+use crate::{opts::*, parse_buildinfo};
```
lann, line 1: Is this supposed to be in this PR?

lann: nm I get it; not sure the `*` is worth it here.

src/commands/build.rs
```diff
@@ -0,0 +1,53 @@
...
+            .opts
+            .app
+            .as_deref()
+            .unwrap_or_else(|| DEFAULT_MANIFEST_FILE.as_ref());
+        let RawAppManifestAnyVersion::V1(app) = raw_manifest_from_file(&manifest_file).await?;
+
+        spin_build::build(app, manifest_file).await?;
+
+        match self.up {
+            true => {
```
lann, line 33: `if self.up`/`else`

src/commands/build.rs
```diff
@@ -0,0 +1,53 @@
...
+                    opts: UpOpts {
+                        address: self.opts.address,
+                        app: Some(manifest_file.into()),
+                        bindle: None,
+                        server: None,
+                        tmp: self.opts.tmp,
+                        env: self.opts.env,
+                        tls_cert: self.opts.tls_cert,
+                        tls_key: self.opts.tls_key,
+                        log: self.opts.log,
```
lann, line 44: Would `...self.opts` work here?

src/opts.rs
```diff
@@ -0,0 +1,83 @@
...
+pub const BUILDINFO_OPT: &str = "BUILDINFO";
+pub const INSECURE_OPT: &str = "INSECURE";
+pub const STAGING_DIR_OPT: &str = "STAGING_DIR";
+pub const HIPPO_SERVER_URL_OPT: &str = "HIPPO_SERVER_URL";
+pub const HIPPO_URL_ENV: &str = "HIPPO_URL";
+pub const JSON_MIME_TYPE: &str = "application/json";
+pub const BUILD_UP_OPT: &str = "UP";
+
+#[derive(StructOpt, Debug)]
+pub struct UpOpts {
```
lann, line 24: I would lean towards leaving this in `up.rs`. You're importing `UpCommand` in `build` anyway, and splitting these into another file means more jumping between files when making changes to `up`.

crates/build/src/lib.rs
lann: Also, should probably turn non-zero exit status into an error here.

src/commands/bindle.rs
radu-matei: There are a lot of opts in this file, I'd rather not import every single one..

crates/build/src/lib.rs
radu-matei: Good catch!

radu-matei:
I _think_ I addressed most of the comments, @lann.

src/commands/build.rs
```diff
@@ -0,0 +1,55 @@
...
+            .unwrap_or_else(|| DEFAULT_MANIFEST_FILE.as_ref());
+        let RawAppManifestAnyVersion::V1(app) = raw_manifest_from_file(&manifest_file).await?;
+
+        spin_build::build(app, manifest_file).await?;
+
+        if self.up {
+            let cmd = UpCommand {
+                opts: UpOpts {
+                    address: self.opts.address,
+                    app: Some(manifest_file.into()),
```
lann, line 39: Does `app` need to be part of `UpOpts`? If it wasn't you could just `opts: self.opts` here.

lann (approved):

radu-matei:
Ok, did another round of updates to this PR, most importantly, `command` is now a single string, which will be executed by a shell.

crates/build/src/lib.rs
radu-matei: Opened #406 to track this.

Issue #349, opened 2022-04-13
default feature flag values for `spin bindle push`


bacongobbler:
If I perform the following

```
$ spin bindle push
error: The following required arguments were not provided:
    --file <APP_CONFIG_FILE>
    --bindle-server <BINDLE_SERVER_URL>
```

A few questions popped up into my mind:

1. should the default value of `--file` be spin.toml similar to `spin up`?
2. Should the default value of `--bindle-server` be `http://localhost:8080/v1`?

Same for `spin bindle prepare`: should `--staging-dir` default to a staging directory similar to `spin bindle push`?

itowlson:
Agree `--file` should default.

I'm not enthusiastic about defaulting `--staging-dir` because it's not obvious what it should be.  `push` uses a temporary directory which it cleans up afterwards, but that doesn't work for `prepare` because the whole purpose of doing a prepare instead of a push is that you have access to the standalone bits - so you need to know where they are, and likely want to control where they go.

I'm also dubious about defaulting `--bindle-server`.  The trouble is that if you forget to specify it - and _aren't_ pushing locally - you'll now end up with a cryptic "connection refused" message rather than a "you forgot to tell us which server" message.  Maybe that's obvious enough - after all, it's the same as you'd get if you entered the wrong URL - but if you're explicitly specifying the URL it at least forced you to think about it.  Dunno.

PR #353, opened 2022-04-14
feat(deploy): add spin deploy command

michelleN:
This is an MVP for the spin deploy command. Given
a bindle server url, hippo server url, and hippo
username/password, this command will create
(1) and push a bindle containing the app to the bindle server
(2) a Hippo App with the name provided in the Spin.toml
(3) a Hippo channel in the Hippo App called "spin-deploy"
and create a post request to register a revision in Hippo.
If an app with the same name already exists, this command will
return an error. If a channel with the name spin-deploy exists,
this will return an error. These errors will be handled more
gracefully with prompts/helpful error messages in subsequent PRs.

This commit it meant to lay the groundwork for the deploy command
so folks can start playing with it and collect opinions.

To play with this command:
1. Use the[ nomad-local-demo](https://github.com/fermyon/nomad-local-demo) to deploy nomad (+ Consul, Traefik, Vault) and bindle.
2. Clone Hippo and in `Infrastructure/Jobs/NomadJob.cs`, change the driver setting from "exec" to "raw_exec". Remove the artifact stanza so that Nomad uses your local Spin binary. Be sure to set the bindle url env variable given in the `./run_servers.sh` output 
3. Start hippo with the nomad scheduler and bindle url.
```
$ dotnet build
$ dotnet run \
--Scheduler:Driver=nomad \
  --Bindle:Url="${BINDLE_URL}"
```
4. Create a hippo account. Use these creds in the next step
5. Build the wasm module for an example app and then `spin deploy --bindle-server <bindle-url> --hippo-username <username> --hippo-password <password>`

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -1,0 +1,249 @@
+use anyhow::{Context, Result};
+use bindle::Id;
+use core::panic;
+use hippo_openapi::apis::{
+    account_api::api_account_createtoken_post,
+    app_api::{api_app_post, ApiAppPostError},
+    channel_api::api_channel_post,
+    configuration::{ApiKey, Configuration},
+    revision_api::{api_revision_post, ApiRevisionPostError},
+    Error,
+};
+use hippo_openapi::models::{
+    ChannelRevisionSelectionStrategy, CreateAppCommand, CreateChannelCommand, CreateTokenCommand,
+    RegisterRevisionCommand,
+};
+use reqwest::header;
+use std::path::PathBuf;
+use structopt::{clap::AppSettings, StructOpt};
+
+const APP_CONFIG_FILE_OPT: &str = "APP_CONFIG_FILE";
+const BINDLE_SERVER_URL_OPT: &str = "BINDLE_SERVER_URL";
+const BINDLE_URL_ENV: &str = "BINDLE_URL";
+const HIPPO_SERVER_URL_OPT: &str = "HIPPO_SERVER_URL";
+const HIPPO_URL_ENV: &str = "HIPPO_URL";
+const STAGING_DIR_OPT: &str = "STAGING_DIR";
+const JSON_MIME_TYPE: &str = "application/json";
+
+/// Package and upload Spin artifacts, notifying Hippo
+#[derive(StructOpt, Debug)]
+#[structopt(
+    about = "Deploy a Spin application",
+    global_settings = &[AppSettings::ColoredHelp, AppSettings::ArgRequiredElseHelp]
+)]
+pub struct DeployCommand {
+    /// Path to spin.toml
+    #[structopt(
+        name = APP_CONFIG_FILE_OPT,
+        short = "f",
+        long = "file",
+        default_value = "spin.toml"
+    )]
+    pub app: PathBuf,
+
+    /// URL of bindle server
+    #[structopt(
+        name = BINDLE_SERVER_URL_OPT,
+        long = "bindle-server",
+        env = BINDLE_URL_ENV,
+    )]
+    pub bindle_server_url: String,
+
+    /// URL of hippo server
+    #[structopt(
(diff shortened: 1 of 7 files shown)


src/commands/deploy.rs
```diff
@@ -0,0 +1,249 @@
...
+};
+use hippo_openapi::models::{
+    ChannelRevisionSelectionStrategy, CreateAppCommand, CreateChannelCommand, CreateTokenCommand,
+    RegisterRevisionCommand,
+};
+use reqwest::header;
+use std::path::PathBuf;
+use structopt::{clap::AppSettings, StructOpt};
+
+const APP_CONFIG_FILE_OPT: &str = "APP_CONFIG_FILE";
```
lann, line 20: I know you're following the existing pattern here, but do we know why we define these `const`s with basically matching names/strings to use them only once?

src/commands/deploy.rs
```diff
@@ -0,0 +1,249 @@
...
+        env = BINDLE_URL_ENV,
+    )]
+    pub bindle_server_url: String,
+
+    /// URL of hippo server
+    #[structopt(
+        name = HIPPO_SERVER_URL_OPT,
+        long = "hippo-server",
+        env = HIPPO_URL_ENV,
+        default_value = "https://localhost:5309",
```
radu-matei, line 57: I think we've had part of this conversation in #349, I don't think we want to have a default value for the Hippo URL.
If I'm using a remote Hippo instance (which we _think_ will be the common use case), this will fail because it's trying to connect to `localhost`.

src/commands/deploy.rs
itowlson: We've mostly used the consts where there are interdependencies between arguments e.g. `conflicts_with` or `requires`, so that we can reference the related arg in a checked way.  (There are cases where the arg itself doesn't have interdependencies but it's in a command with other args that do, and in those cases it's been "so things look the same."  E.g. the log directory in `UpCommand`.  Also cases where the same arg appears under multiple commands and we want to ensure they have the same name in both places.)  If there's no dependencies and the names are just UI elements then it's probably not needed.

This particular one might fall under "stay consistent" as the same arg appears in UpCommand and the Bindle commands.  We don't currently have a file of common names/args though so we just replicate the consts and arg settings around.  Learn more in episode 14 of my long-running podcast Why Structopt Is Wrong.

...but that is completely orthogonal to this PR and to whether we should use the const here.  Sorry Michelle.

src/commands/deploy.rs
```diff
@@ -0,0 +1,249 @@
...
+    )]
+    pub hippo_password: String,
+}
+
+impl DeployCommand {
+    pub async fn run(self) -> Result<()> {
+        let bindle_id = self
+            .create_and_push_bindle()
+            .await
+            .expect("Unable to create and push bindle from app {}");
```
radu-matei, line 93: There is an extra `{}` here.

radu-matei (approved):
I deployed the Nomad infra, started Hippo, and got an application up and running!
The only thing not working was routing using the default local domain, but I suspect that's an issue with my setup.

Creating the application, channel, and starting it using Nomad worked for me (with the mention that if running IPv6, both Bindle and the Nomad job definition from Hippo need `[]` around the IP address — ref https://github.com/fermyon/nomad-local-demo/issues/5).

(I have a few non-blocking comments related to error messages and default addresses).

I think this is in a great spot to be iterated on, LGTM.

src/commands/deploy.rs
michelleN: Removed. Thank you!

src/commands/deploy.rs
michelleN: Agree. Removed!

radu-matei:
:shipit:!

fibonacci1729 (approved):

PR #356, opened 2022-04-14
ref(sdk/go/http): Simplify code and free memory allocations

adamreese:


radu-matei (approved):
LGTM

At some point we should also make sure that not freeing memory doesn't cause issues, but this change LGTM.

Thanks!

Issue #340, opened 2022-04-10
Validate component IDs are unique


itowlson:
Spin currently permits non-unique IDs, e.g. this loads without complaint:

```toml
[[component]]
id = "hello"
source = "target/wasm32-wasi/release/spinhelloworld.wasm"
[component.trigger]
route = "/hello"

[[component]]
id = "hello"
source = "target/wasm32-wasi/release/spinhelloworld.wasm"
[component.trigger]
route = "/hello"
```

There is currently stuff in the assets code that assumes component IDs are unique - we could fix this, but it might be better to just fail on duplicate ID, (1) because future code might make the same assumption and (2) it would avoid confusion during diagnostics.

0xcodeboi:
Hi @itowlson, I was taking a look at this and I _think_ it makes sense to add a uniqueness constraint somewhere around this code:
https://github.com/fermyon/spin/blob/5b0a48ecb9ea47563ae35f9ea414487fbc5365c5/crates/loader/src/bindle/mod.rs#L67-L74

My initial thought was that it probably makes more sense to have a constraint check in `config` crate, maybe using this error:
https://github.com/fermyon/spin/blob/5b0a48ecb9ea47563ae35f9ea414487fbc5365c5/crates/config/src/lib.rs#L16-L19

but the validation logic appears to be operating on a per-key basis and does not seem to have the full picture (i.e. it does not have a reference to _all_ the components) -- does this sound right or am I missing something?

I think there's a clear path forward to add a uniqueness check somewhere in the `loader` crate, if that sounds good I can take a stab at this 🙂

radu-matei:
Fixed in #370

PR #370, opened 2022-04-18
Validate component IDs are unique

0xcodeboi:
The `spin.toml` configuration file currently loads without error when multiple components use the same `id`. This PR adds a uniqueness constraint in the `loader` crate which will fail to load a configuration file with multiple defined components sharing an identifier.

See https://github.com/fermyon/spin/issues/340 for more details.

### DIFF
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -10,7 +10,7 @@ pub mod config;
 #[cfg(test)]
 mod tests;
 
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use config::{RawAppInformation, RawAppManifest, RawAppManifestAnyVersion, RawComponentManifest};
 use futures::future;
 use path_absolutize::Absolutize;
@@ -63,6 +63,20 @@ async fn prepare_any_version(
     }
 }
 
+/// Iterates over a vector of RawComponentManifest structs and throws an error if any component ids are duplicated
+fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
+    let mut ids: Vec<String> = Vec::new();
+    for c in components {
+        let id = c.id;
+        if ids.contains(&id) {
+            bail!("cannot have duplicate component IDs: {}", id);
+        } else {
+            ids.push(id);
+        }
+    }
+    Ok(())
+}
+
 /// Converts a raw application manifest into Spin configuration.
 async fn prepare(
     mut raw: RawAppManifest,
@@ -70,6 +84,8 @@ async fn prepare(
     base_dst: impl AsRef<Path>,
 ) -> Result<Application<CoreComponent>> {
     let info = info(raw.info, &src);
+
+    error_on_duplicate_ids(raw.components.clone())?;
 
     let mut config_root = raw.config.unwrap_or_default();
     for component in &mut raw.components {
--- a/crates/loader/src/local/tests.rs
+++ b/crates/loader/src/local/tests.rs
@@ -139,3 +139,25 @@ fn test_wagi_executor_with_custom_entrypoint() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_duplicate_component_id_is_rejected() -> Result<()> {
+    const MANIFEST: &str = "tests/invalid-manifest-duplicate-id.toml";
+
+    let temp_dir = tempfile::tempdir()?;
+    let dir = temp_dir.path();
+    let app = from_file(MANIFEST, dir).await;
+
+    assert!(
+        app.is_err(),
+        "Expected component IDs to be unique, but there were duplicates"
+    );
+
+    let e = app.unwrap_err().to_string();
+    assert!(
+        e.contains("hello"),
+        "Expected error to contain duplicate component ID `hello`"
+    );
+
+    Ok(())
+}
--- a/crates/loader/tests/invalid-manifest-duplicate-id.toml
+++ b/crates/loader/tests/invalid-manifest-duplicate-id.toml
@@ -1,0 +1,16 @@
+spin_version = "1"
+name = "spin-hello-world-duplicate"
+version = "1.0.0"
+trigger = { type = "http", base = "/" }
+
+[[component]]
+id = "hello"
+source = "path/to/wasm/file.wasm"
+[component.trigger]
+route = "/hello"
+
+[[component]]
+id = "hello"
+source = "path/to/wasm/file.wasm"
+[component.trigger]
+route = "/hello"


crates/loader/src/local/tests.rs
```diff
@@ -139,3 +139,25 @@ fn test_wagi_executor_with_custom_entrypoint() -> Result<()> {
...
+    let temp_dir = tempfile::tempdir()?;
+    let dir = temp_dir.path();
+    let app = from_file(MANIFEST, dir).await;
+
+    assert!(
+        app.is_err(),
+        "Expected component IDs to be unique, but there were duplicates"
+    );
+
+    // let e = app.unwrap_err().to_string();
```
0xcodeboi, line 156: This call panics, basically want to test that the error message says which component ID is the duplicate

crates/loader/src/local/mod.rs
```diff
@@ -72,7 +72,11 @@ async fn prepare(
     let info = info(raw.info, &src);
 
     let mut config_root = raw.config.unwrap_or_default();
+    let component_ids: Vec<String> = raw.components.iter().map(|c| c.id.clone()).collect();
     for component in &mut raw.components {
+        if component_ids.contains(&component.id) {
+            panic!("cannot have duplicate component ID: {}", &component.id)
```
0xcodeboi, line 78: Probably don't want to use a `panic!` here?

crates/loader/src/local/mod.rs
```diff
@@ -72,7 +72,11 @@ async fn prepare(
     let info = info(raw.info, &src);
 
     let mut config_root = raw.config.unwrap_or_default();
+    let component_ids: Vec<String> = raw.components.iter().map(|c| c.id.clone()).collect();
```
0xcodeboi, line 75: instead of the `contains` call below, there's probably a way to just make sure these ids are unique, and error if they're not

crates/loader/src/local/mod.rs
radu-matei: We almost never want a library to be the source of a panic, which is why we try to return `Result` instead here.
(you can easily do that here with `bail!("<message>")`.

(we could do a much better job of returning actual typed errors instead, but for this PR, a `bail!` should be enough)

crates/loader/src/local/tests.rs
0xcodeboi: ah this is failing because the code is broken 😂 
```rust
component_ids.contains(&component.id)  // of course it does
```

radu-matei:
While you're iterating on the PR, keep an eye out for `cargo clippy` suggestions:

```
error: avoid using `collect()` when not needed
[19](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:19)
  --> crates/loader/src/local/mod.rs:75:82
[20](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:20)
   |
[21](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:21)
75 |     let component_ids: Vec<String> = raw.components.iter().map(|c| c.id.clone()).collect();
[22](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:22)
   |                                                                                  ^^^^^^^
[23](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:23)
76 |     for component in &mut raw.components {
[24](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:24)
77 |         if component_ids.contains(&component.id) {
[25](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:25)
   |            ------------------------------------- the iterator could be used here instead
[26](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:26)
   |
[27](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:27)
   = note: `-D clippy::needless-collect` implied by `-D warnings`
[28](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:28)
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_collect
[29](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:29)
help: check if the original Iterator contains an element instead of collecting then checking
[30](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:30)
   |
[31](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:31)
75 ~     
[32](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:32)
76 |     for component in &mut raw.components {
[33](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:33)
77 ~         if raw.components.iter().map(|c| c.id.clone()).any(|x| x == component.id) {
[34](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:34)
   |
[35](https://github.com/fermyon/spin/runs/6057820499?check_suite_focus=true#step:8:35)
```

crates/loader/src/local/mod.rs
```diff
@@ -71,6 +71,17 @@ async fn prepare(
 ) -> Result<Application<CoreComponent>> {
     let info = info(raw.info, &src);
 
+    let id_counts: HashMap<String, u8> =
+        raw.components.iter().fold(HashMap::new(), |mut counts, c| {
+            *counts.entry(c.id.clone()).or_insert(0) += 1;
+            counts
+        });
+    for (id, count) in id_counts {
```
0xcodeboi, line 79: This version works now, I'm going through some Rust docs to figure out a way to make this cleaner. Ideally, I'd avoid the extra `for` loop, and maybe avoid creating a new `HashMap` here.

crates/loader/src/local/mod.rs
```diff
@@ -71,6 +71,17 @@ async fn prepare(
 ) -> Result<Application<CoreComponent>> {
     let info = info(raw.info, &src);
 
+    let id_counts: HashMap<String, u8> =
```
itowlson, line 74: Suggestion: pull this out into a `contains_duplicate_ids(...) -> bool` function.  This makes it easier to understand at the point where it's called, and means that if people need to iterate on the implementation then it's guaranteed to be localised to that function.

crates/loader/src/local/mod.rs
```diff
@@ -71,6 +71,17 @@ async fn prepare(
...
+        raw.components.iter().fold(HashMap::new(), |mut counts, c| {
+            *counts.entry(c.id.clone()).or_insert(0) += 1;
+            counts
+        });
+    for (id, count) in id_counts {
+        if count > 1 {
```
itowlson, line 80: Since this bails at the first duplicate, rather than counting all occurrences, you could instead maintain a set or vector of "seen" IDs and bail as soon as you see the first duplicate, e.g.

```
let mut seen = FavouriteCollection::new();
for c in components {
  let id = c.id;
  if seen.contains(id) {
    bail!
  } else {
    seen.add(id);
  }
}
```

(I know it means using a mutable collection, but if, as suggested, you parcel it out to a separate function then the mutation is safely isolated from the rest of the program.)

itowlson (commented):
Nice one - thanks for tackling this.  I had just a couple of suggestions but otherwise looks good!

crates/loader/src/local/mod.rs
0xcodeboi: Makes sense, thanks for the suggestion.

0xcodeboi:
Applied the feedback and am happier with how the code looks now. Thanks again for feedback! I'll squash my commits after a review unless there are more changes requested.

crates/loader/src/local/mod.rs
```diff
@@ -63,6 +63,20 @@ async fn prepare_any_version(
     }
 }
 
+/// Iterates over a vector of RawComponentManifest structs and return an error if any component ids are duplicated
+fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
+    let mut ids: Vec<String> = Vec::new();
+    for c in components {
+        let id = c.id;
+        if ids.contains(&id) {
+            return Err(anyhow!("cannot have duplicate component IDs: {}", id));
```
radu-matei, line 72: nit: you can use `bail!` here as well.

crates/loader/src/local/mod.rs
```diff
@@ -71,6 +85,10 @@ async fn prepare(
 ) -> Result<Application<CoreComponent>> {
     let info = info(raw.info, &src);
 
+    if let Err(e) = error_on_duplicate_ids(raw.components.clone()) {
```
radu-matei, line 88: I think you can just `error_on_duplicate_ids(raw.components.clone())?` here?

crates/loader/src/local/mod.rs
0xcodeboi: I had originally written it that way but was running into some issues

basically:
```rust
if ids.contains(&id) {
    bail!("cannot have duplicate component IDs: {}", id);
}

...

error_on_duplicate_ids(raw.components.clone()); // warning: unused `Result` that must be used
``` 
I got rid of the warning by writing it as
```rust
_ = error_on_duplicate_ids(raw.components.clone());
```

The test case also did not work as is currently written, the `from_file(...).await` call panicked, and the following `app.is_err()` and the "error contains `hello`" couldn't be reached.

My thinking was for the constraint function to just return a result and let the caller handle it, as opposed to panic'ing inside the constraint function.

I think the _real_ issue I'm having is understanding how to structure the test if the main function `prepare` isn't the one making the `bail!` call.

crates/loader/src/local/mod.rs
0xcodeboi: Ah, ignore https://github.com/fermyon/spin/pull/370/files#r857180717

Adding the `?` makes my issues disappear 🙂 I _don't think_ I really understand why yet - but happy to push these changes up. Thanks!

radu-matei: Basically:
> The `?` placed after a `Result` value is defined to work in almost the same way as a `match` expressions defined to handle the `Result` values.

Here are a few resources on Rust errors:
- https://doc.rust-lang.org/book/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-the--operator
- https://blog.burntsushi.net/rust-error-handling/

0xcodeboi: That's super helpful, thanks! And thanks for guiding me through this PR. It's quite trivial, but I learned a lot!

radu-matei (approved):
Thank you so much for your contribution, @0xcodeboi!

LGTM

PR #372, opened 2022-04-18
fix(http): Make spin http headers optional

adamreese:
Spin http response headers are optional in the wit files.

https://github.com/fermyon/spin/blob/14603bb71f9592fbfe3d10f5deb961fddb89f58c/wit/ephemeral/http-types.wit#L47

### DIFF
--- a/crates/http/src/spin.rs
+++ b/crates/http/src/spin.rs
@@ -113,6 +113,13 @@ impl SpinHttpExecutor {
         })
         .await??;
 
+        if res.status < 100 || res.status > 600 {
+            log::error!("malformed HTTP status code");
+            return Ok(Response::builder()
+                .status(http::StatusCode::INTERNAL_SERVER_ERROR)
+                .body(Body::empty())?);
+        };
+
         let mut response = http::Response::builder().status(res.status);
         if let Some(headers) = response.headers_mut() {
             Self::append_headers(headers, res.headers)?;
--- a/crates/http/src/spin.rs
+++ b/crates/http/src/spin.rs
@@ -114,7 +114,9 @@ impl SpinHttpExecutor {
         .await??;
 
         let mut response = http::Response::builder().status(res.status);
-        Self::append_headers(response.headers_mut().unwrap(), res.headers)?;
+        if let Some(headers) = response.headers_mut() {
+            Self::append_headers(headers, res.headers)?;
+        }
 
         let body = match res.body {
             Some(b) => Body::from(b),


fibonacci1729 (approved):

crates/http/src/spin.rs
```diff
@@ -114,7 +114,9 @@ impl SpinHttpExecutor {
         .await??;
 
         let mut response = http::Response::builder().status(res.status);
-        Self::append_headers(response.headers_mut().unwrap(), res.headers)?;
+        if let Some(headers) = response.headers_mut() {
```
radu-matei, line 117: Has this actually ever failed?
Theoretically, we just created a new empty response, so the `headers_mut()` call only ever fail if the builder has an error — https://docs.rs/http/latest/http/response/struct.Builder.html#method.headers_mut


That being said, I don't have anything against an extra check here.

adamreese: I was getting a panic when not setting the headers with the go sdk

adamreese: I think If the wit says optional we should implement it as defined

radu-matei: But I don't think this is the right fix.
This checks that _the new Rust HTTP response we just created with the builder_ doesn't have an empty header map.

I'd love to have a look at the actual panic.

adamreese: Looks like I only get it when there is no status code and no headers.  Status code is required so an error is correct.  It just should be more useful.  I'll take another look

```
2022-04-18T16:25:36.837707Z TRACE prepare_component{component="tinygo-hello" env=None args=None}: spin_engine: Creating store.
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', crates/http/src/spin.rs:117:53
```

radu-matei: Ah, I assume this is because `res.status` is 0 when it's not set, which results in an error there.

Could you check for `res.status != 0` and for `None` on headers?

Thanks!

radu-matei (changes requested):
Could you please check for `res.status != 0` as well as for `None` on headers?

radu-matei (approved):

crates/http/src/spin.rs
adamreese: Can you check if this is what you meant?

adamreese: Never mind I didn't refresh

PR #374, opened 2022-04-18
Exit Spin when no Redis connection is available

### LINKED ISSUE #357: Stopping the Redis server while Spin is subscribed on a channel results in infinite loop
To reproduce this:

- start the Redis server
- start Spin using a Redis example (tested with both Rust and Go components)
- publish a few messages
- stop the server while Spin is still running

This results in an infinite loop of:

```
2022-04-14T18:51:59.596483Z TRACE spin_redis_engine: Empty message
2022-04-14T18:51:59.596495Z TRACE spin_redis_engine: Empty message
2022-04-14T18:51:59.596504Z TRACE spin_redis_engine: Empty message
2022-04-14T18:51:59.596513Z TRACE spin_redis_engine: Empty message
2022-04-14T18:51:59.596522Z TRACE spin_redis_engine: Empty message
2022-04-14T18:51:59.596531Z TRACE spin_redis_engine: Empty message
```

cc @fibonacci1729

fibonacci1729:
Fixes #357

### DIFF
--- a/crates/redis/src/lib.rs
+++ b/crates/redis/src/lib.rs
@@ -6,7 +6,7 @@ use crate::spin::SpinRedisExecutor;
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use futures::StreamExt;
-use redis::Client;
+use redis::{Client, ConnectionLike};
 use spin_engine::{Builder, ExecutionContextConfiguration};
 use spin_manifest::{
     Application, ComponentMap, CoreComponent, RedisConfig, RedisTriggerConfiguration,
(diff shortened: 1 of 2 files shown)


crates/redis/src/lib.rs
```diff
@@ -95,7 +95,13 @@ impl RedisTrigger {
         loop {
             match stream.next().await {
                 Some(msg) => drop(self.handle(msg).await),
-                None => log::trace!("Empty message"),
+                None => {
+                    log::trace!("Empty message");
+                    if !client.check_connection() {
+                        log::trace!("No Redis connection available");
```
radu-matei, line 101: Should this be an `info!` log, similar to the one informing about the subscription?

radu-matei (approved):
Manually tested, this works great.

LGTM, with one question related to the log level.

michelleN (approved):
Should we re-check the connection anytime a message comes in? So in the Some case too?

fibonacci1729:
@michelleN I don't believe so but I could be wrong. Do you have a compelling reason? To me, we really only care about checking the connection if we receive an empty message, otherwise if the `Some(_)` branch is executed, the connection being up seems implied.

Issue #376, opened 2022-04-18
Make Application clonable 


Mossaka:
We should make `Application` clonable, such that it could be used to constrct customized `ExecutionContext`.

```rust
  --> crates/manifest/src/lib.rs:23:5
   |
14 | #[derive(Clone, Debug)]
   |          ----- in this derive macro expansion
...
23 |     pub config_resolver: Option<Resolver>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Clone` is not implemented for `Resolver`
   |
   = note: required because of the requirements on the impl of `Clone` for `std::option::Option<Resolver>`
   = note: this error originates in the derive macro `Clone` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0277`.
error: could not compile `spin-manifest` due to previous error
```

One approach to do this is to change the type of config_resolver to `Option<Arc<Resolver>>`

Discussed with @radu-matei on this issue this morning.

radu-matei:
cc @lann

lann:
I'd really like to remove `Application` from trigger initialization, but unfortunately the HTTP trigger config is currently deeply woven into the component config.

In the meantime it should be fine to wrap `config_resolver` in an `Arc`. It does currently get mutated in the `spin up` init code (which I need to find a better way of dealing with), but it shouldn't be cloned prior to that so `Arc::get_mut` ought to work there. If you don't intend to make that change I can.

Mossaka:
@lann thanks for your response! I can get the commented code uncommented.

lann:
I happen to be working on some related code so I'll be making this change. If you've already done it I'll take care of conflicts later.

Mossaka:
Sure go ahead!

PR #377, opened 2022-04-19
Add execution context as a parameter to trigger constructor

Mossaka:
This PR fixes the issue #376 

1. Makes `Application` clonable
2. Adds a new parameter `execution_context` to triggers so that users can customize different wasmtime configurations. 

@radu-matei

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -1,5 +1,6 @@
 use anyhow::{bail, Result};
-use spin_http_engine::{HttpTrigger, TlsConfig};
+use spin_engine::{Builder, ExecutionContext, ExecutionContextConfiguration};
+use spin_http_engine::{spin_http::SpinHttpData, HttpTrigger, TlsConfig};
 use spin_manifest::{Application, ApplicationTrigger, CoreComponent};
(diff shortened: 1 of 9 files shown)


src/commands/up.rs
```diff
@@ -111,10 +111,10 @@ impl UpCommand {
...
         append_env(&mut app, &self.env)?;
 
-        if let Some(ref mut resolver) = app.config_resolver {
-            // TODO(lann): Make config provider(s) configurable.
-            resolver.add_provider(spin_config::provider::env::EnvProvider::default());
-        }
+        // if let Some(ref mut resolver) = app.config_resolver {
+        //     // TODO(lann): Make config provider(s) configurable.
+        //     resolver.add_provider(spin_config::provider::env::EnvProvider::default());
+        // }
```
lann, line 117: We can't just remove this...

lann (changes requested):

src/commands/up.rs
lann: Presumably related to #376

lann:
Good idea, I just did it a little more aggressively. :slightly_smiling_face:

PR #381, opened 2022-04-21
Add support for "host components"

### LINKED ISSUE #311: refactor: Host components
Currently, host-implemented interfaces have to be hard-coded in `engine` in a couple of places:
- the implementation state is included in `RuntimeContext`: https://github.com/fermyon/spin/blob/2be403465390f1a767181f16190aa4f14512c9f3/crates/engine/src/lib.rs#L49-L50
- a `link_*` method, which is called in `build_default`: https://github.com/fermyon/spin/blob/2be403465390f1a767181f16190aa4f14512c9f3/crates/engine/src/lib.rs#L106-L108

I'd like to invert this dependency by introducing a new `HostComponent` trait. After a little prototyping I think it can look something like:
```rust
pub trait HostComponent {
    type State: Default + 'static;
    fn add_to_linker<T>(id: String, linker: &mut wasmtime::Linker<crate::RuntimeContext<T>>);
    fn initial_state(id: String) -> State;
}
```
implemented like:
```rust
impl HostComponent for OutboundHttp {
    type State = Self;
    fn add_to_linker<T>(id: String, linker: &mut Linker<RuntimeContext<T>>) {
        wasi_outbound_http::add_to_linker(linker, |ctx| ctx.host_components.get_state(&id).unwrap())
    }
}
```
_Note: The separate `State` type isn't necessary here but is needed for interfaces that use reftypes, which changes their state data type._

The component state would be held in a mapping on `RuntimeContext` as e.g.:
```rust
/// Host components data,
pub host_components: HashMap<String, Box<dyn Any>>,
```

There might be a performance issue here as this HashMap would be referenced once for every host component function call, but if that is the case we could switch to a more efficient form of mapping (perhaps a Vec with pre-computed indexes or even something more exotic/unsafe :grimacing:). If performance is a real problem here I _think_ its a solvable one but its a real tradeoff vs the organizational benefits of splitting things up like this. Other ideas for how to do this very welcome.

lann:
Host components are a way to add host implementations of component
interfaces with a dependency injection pattern. Host components
implement the HostComponent trait which defines how the host
implementation is linked and how the component's state is constructed.

Fixes #311

### DIFF
--- a/crates/engine/src/host_component.rs
+++ b/crates/engine/src/host_component.rs
@@ -1,0 +1,89 @@
+use std::{any::Any, marker::PhantomData};
+
+use spin_manifest::CoreComponent;
+use wasmtime::Linker;
+
+use crate::RuntimeContext;
+
+/// Represents a host implementation of a Wasm interface.
+pub trait HostComponent: Send + Sync {
+    /// Host component runtime state.
+    type Data: Any + Send;
+
+    /// Add this component to the given Linker, using the given runtime state-getting closure.
+    fn add_to_linker<T>(
+        linker: &mut Linker<RuntimeContext<T>>,
+        data_handle: HostComponentsDataHandle<Self::Data>,
+    ) -> anyhow::Result<()>;
+
+    /// Build a new runtime state object for the given component.
+    fn build_data(&self, component: &CoreComponent) -> anyhow::Result<Self::Data>;
+}
+type HostComponentData = Box<dyn Any + Send>;
+
+type DataBuilder = Box<dyn Fn(&CoreComponent) -> anyhow::Result<HostComponentData> + Send + Sync>;
+
+#[derive(Default)]
+pub(crate) struct HostComponents {
+    data_builders: Vec<DataBuilder>,
+}
+
+impl HostComponents {
+    pub(crate) fn insert<'a, T: 'static, Component: HostComponent + 'static>(
+        &mut self,
+        linker: &'a mut Linker<RuntimeContext<T>>,
+        host_component: Component,
+    ) -> anyhow::Result<()> {
+        let handle = HostComponentsDataHandle {
+            idx: self.data_builders.len(),
+            _phantom: PhantomData,
+        };
+        Component::add_to_linker(linker, handle)?;
+        self.data_builders.push(Box::new(move |c| {
+            Ok(Box::new(host_component.build_data(c)?))
+        }));
+        Ok(())
+    }
+
+    pub(crate) fn build_data(&self, c: &CoreComponent) -> anyhow::Result<HostComponentsData> {
+        Ok(HostComponentsData(
+            self.data_builders
+                .iter()
+                .map(|build_data| build_data(c))
+                .collect::<anyhow::Result<_>>()?,
+        ))
+    }
+}
+
+/// A collection of host component data.
+#[derive(Default)]
+pub struct HostComponentsData(Vec<HostComponentData>);
+
+/// A handle to component data, used in HostComponent::add_to_linker.
+pub struct HostComponentsDataHandle<T> {
+    idx: usize,
+    _phantom: PhantomData<fn(T) -> T>,
+}
+
+impl<T: 'static> HostComponentsDataHandle<T> {
+    /// Get the component data associated with this handle from the RuntimeContext.
+    pub fn get_mut<'a, U>(&self, ctx: &'a mut RuntimeContext<U>) -> &'a mut T {
+        ctx.host_components_data
+            .0
+            .get_mut(self.idx)
+            .unwrap()
+            .downcast_mut()
+            .unwrap()
+    }
+}
+
+impl<T> Clone for HostComponentsDataHandle<T> {
+    fn clone(&self) -> Self {
+        Self {
+            idx: self.idx,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<T> Copy for HostComponentsDataHandle<T> {}
--- a/crates/object-store/src/file.rs
+++ b/crates/object-store/src/file.rs
@@ -7,11 +7,8 @@ use std::{
 use crate::wit::spin_object_store;
 use anyhow::Context;
 use cap_std::{ambient_authority, fs::OpenOptions};
-
-pub type FileObjectStoreData = (
-    FileObjectStore,
-    spin_object_store::SpinObjectStoreTables<FileObjectStore>,
-);
+use spin_engine::host_component::{HostComponent, HostComponentsDataHandle};
+use spin_manifest::CoreComponent;
 
 pub struct FileObjectStore {
     root: cap_std::fs::Dir,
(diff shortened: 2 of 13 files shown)


lann:
Keeping this a draft until dependency #382 merges.

lann:
I only updated outbound-redis and object-store in this PR to give an idea of what  `HostComponent` implementations look like. I think outbound-http will be a little more complicated to convert (but possible). I'm not sure about config yet; it might need to be kept separate if `HostComponents` end up using the config mechanism themselves. Wasi will stay in engine for now.

src/commands/up.rs
```diff
@@ -170,6 +170,7 @@ impl UpCommand {
         };
         let mut builder = Builder::new(config)?;
         builder.link_defaults()?;
+        builder.add_host_component(outbound_redis::OutboundRedis)?;
```
itowlson, line 173: This is really nice.

crates/outbound-redis/src/lib.rs
```diff
@@ -2,13 +2,29 @@ use outbound_redis::*;
...
+use spin_engine::host_component::HostComponent;
 
 wit_bindgen_wasmtime::export!("../../wit/ephemeral/outbound-redis.wit");
 
 /// A simple implementation to support outbound Redis commands.
 #[derive(Default, Clone)]
 pub struct OutboundRedis;
 
+impl HostComponent for OutboundRedis {
+    type Data = Self;
```
itowlson, line 14: I don't quite get why this would be `Self` rather than `()`.  Is the 'runtime state' always an instance of the component plus optional something else (config?)?  If so, would it be clearer to distinguish these, so that the Self is baked in and only the additional data needs to be in the `Data` type?

crates/object-store/src/file.rs
```diff
@@ -25,9 +22,29 @@ impl FileObjectStore {
...
+        spin_object_store::SpinObjectStoreTables<FileObjectStore>,
+    );
+
+    fn add_to_linker<T>(
+        linker: &mut wasmtime::Linker<spin_engine::RuntimeContext<T>>,
+        data_handle: HostComponentsDataHandle<Self::Data>,
+    ) -> anyhow::Result<()> {
+        crate::add_to_linker(linker, move |ctx| {
+            let (data, table) = data_handle.get_mut(ctx);
+            (data, table)
```
itowlson, line 41: Given this deconstructs and immediately reconstructs the tuple, why not just `move |ctx| data_handle.get_mut(ctx)`?

crates/engine/src/host_component.rs
```diff
@@ -0,0 +1,89 @@
...
+    type Data: Any + Send;
+
+    /// Add this component to the given Linker, using the given runtime state-getting closure.
+    fn add_to_linker<T>(
+        linker: &mut Linker<RuntimeContext<T>>,
+        data_handle: HostComponentsDataHandle<Self::Data>,
+    ) -> anyhow::Result<()>;
+
+    /// Build a new runtime state object for the given component.
+    fn build_data(&self, component: &CoreComponent) -> anyhow::Result<Self::Data>;
```
itowlson, line 20: `data` is such an open-ended term - I found it hard to understand what this member did at the point of implementation.  Can we be more specific e.g. `build_runtime_state`?

crates/engine/src/host_component.rs
```diff
@@ -0,0 +1,89 @@
...
+use wasmtime::Linker;
+
+use crate::RuntimeContext;
+
+/// Represents a host implementation of a Wasm interface.
+pub trait HostComponent: Send + Sync {
+    /// Host component runtime state.
+    type Data: Any + Send;
+
+    /// Add this component to the given Linker, using the given runtime state-getting closure.
```
itowlson, line 13: Not sure I understand the reference to a closure - does this refer to a previous implementation?

crates/engine/src/host_component.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+/// A collection of host component data.
+#[derive(Default)]
+pub struct HostComponentsData(Vec<HostComponentData>);
+
+/// A handle to component data, used in HostComponent::add_to_linker.
+pub struct HostComponentsDataHandle<T> {
+    idx: usize,
```
itowlson, line 64: This is an index into a specific `HostComponentsData`.  Is there a way to bundle this up so they travel together instead of having this slightly orphaned-looking index?  (Is that what the closure used to do?)

itowlson (approved):
I love the structure of this - I love that each component defines its own linkage and that the top level now injects what it wishes.  I struggled a bit to understand the implementation.  But from a usage point of view this is great.

crates/outbound-redis/src/lib.rs
lann: The closure passed to `add_to_linker` just below has to return a `&mut Data`; you have to have an instance of `Data` to borrow from even if its a unit struct (unless you want to do some `transmute` jiggery-pokery, but if your `Data` has state you'll need the ref anyway).

`Data` is just the type that implements the interface _unless_ that interface uses resources, in which case `Data` is `(Impl, <InterfaceName>Tables<Impl>)` (the "tables" track resource state).

I would be happy to have a way to simplify this, but I spent way too much time trying and not enough time succeeding.

lann: To clarify: "has to return a `&mut Data`" really means "`Data` has to match the wit-bindgen-generated signature of that closure".

crates/object-store/src/file.rs
lann: `get_mut` returns e.g. `&mut (A, B)`, but this closure needs (&mut A, &mut B). I also spent an unreasonable amount of time trying to fix this with no luck. Suggests welcome. If they work.

crates/engine/src/host_component.rs
lann: Woops! Yes indeed; I originally tried to pass a closure in rather than this "handle". I still don't understand why that wasn't essentially equivalent to this, but after much lifetime wrangling I gave up on it.

crates/engine/src/host_component.rs
lann: Maybe? This handle is created well before the data, so it can't reference it directly.

crates/object-store/src/file.rs
itowlson: Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaah.  That's cool - I just completely missed that nuance.

crates/engine/src/host_component.rs
radu-matei: Interesting, I would have expected to be able to use a closure similar to the one in the generated Wasmtime code.

crates/engine/src/host_component.rs
radu-matei: `data` is the term used by `wit-bindgen` to refer to the runtime state, so the name does make sense.

crates/engine/src/host_component.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Represents a host implementation of a Wasm interface.
+pub trait HostComponent: Send + Sync {
+    /// Host component runtime state.
+    type Data: Any + Send;
+
+    /// Add this component to the given Linker, using the given runtime state-getting closure.
+    fn add_to_linker<T>(
+        linker: &mut Linker<RuntimeContext<T>>,
+        data_handle: HostComponentsDataHandle<Self::Data>,
+    ) -> anyhow::Result<()>;
```
radu-matei, line 17: nit: could you just `use anyhow::Result` here?

crates/outbound-redis/src/lib.rs
```diff
@@ -2,13 +2,29 @@ use outbound_redis::*;
...
+
+    fn add_to_linker<T>(
+        linker: &mut wit_bindgen_wasmtime::wasmtime::Linker<spin_engine::RuntimeContext<T>>,
+        data_handle: spin_engine::host_component::HostComponentsDataHandle<Self::Data>,
```
radu-matei, line 18: nit: would you mind importing `Linker` and `HostComponentsDataHandle`?

radu-matei (approved):

crates/engine/src/host_component.rs
lann: If someone can make it work I'd be happy with the change :shrug:

radu-matei: Hah, no, I definitely wouldn't block this PR.

PR #382, opened 2022-04-21
Refactor ExecutionContext init out of triggers

### LINKED ISSUE #376: Make Application clonable

lann:
Move the ExecutionContext construction out of HttpTrigger/RedisTrigger
and into the `spin up` handler. This helps deduplicate this
initialization and allows custom trigger executors to customize the
ExecutionContext.

Make Application clone-able (again) by wrapping config_resolver in Arc.

Fixes #376

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3339,6 +3339,7 @@ dependencies = [
  "anyhow",
  "http 0.2.6",
  "hyper",
+ "spin-engine",
  "spin-http-engine",
  "spin-manifest",
 ]
--- a/crates/engine/src/lib.rs
+++ b/crates/engine/src/lib.rs
@@ -39,7 +39,7 @@ impl From<Application<CoreComponent>> for ExecutionContextConfiguration {
         Self {
             components: app.components,
             label: app.info.name,
-            config_resolver: app.config_resolver.map(Arc::new),
+            config_resolver: app.config_resolver,
             ..Default::default()
         }
     }
--- a/crates/engine/src/lib.rs
+++ b/crates/engine/src/lib.rs
@@ -139,6 +139,7 @@ impl<T: Default> Builder<T> {
     /// Builds a new instance of the execution context.
     #[instrument(skip(self))]
     pub async fn build(&mut self) -> Result<ExecutionContext<T>> {
+        let _sloth_warning = warn_if_slothful();
         let mut components = HashMap::new();
         for c in &self.config.components {
             let core = c.clone();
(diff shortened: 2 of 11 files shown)


lann:
Part of the rationale for this change is to allow this: https://github.com/fermyon/spin/pull/381/files#diff-5907a2d3818517e8e2f9619cd85b4578244a757255d6d6ce270f280ae35442a9R173

fibonacci1729 (approved):

fibonacci1729:
Looks great! LGTM

itowlson (approved):
I don't entirely get what's going on here in terms of how responsibilities are parcelled out - from the linked PR, I think what we are saying is that instead of the command just telling the trigger to run (and it being up to the trigger to figure out how), the command now _assembles_ an execution engine and a trigger (and tells the trigger to run using the engine).  This feels like a good decoupling.

There are still a couple of things that feel conceptually a bit murky, but they are very cosmetic:

1. Both bits have copies of the `app` object.  This makes me wonder if there is still subtle coupling - longer term should we be prising the app object into distinct trigger and execution engine bits?

2. To me, and this is very subjective, the fact that the trigger is created with its engine makes this feel like layering rather than putting the two bits together.  And perhaps that's true.  Or perhaps it's a distinction so subtle it only exists in my mind - that does seem more likely.

lann:
> Both bits have copies of the app object. This makes me wonder if there is still subtle coupling - longer term should we be prising the app object into distinct trigger and execution engine bits?

Agreed :slightly_smiling_face: https://github.com/fermyon/spin/issues/376#issuecomment-1102690677

itowlson:
Shoulda known you'd be way ahead of me!

lann:
I'm wondering if - as a pattern - the triggers should actually take a `spin_engine::Builder` to give them a chance to make their own changes before building the `ExecutionContext`. We just discussed in the spin meeting the possibility of updating the http trigger to incorporate a host-exported interface (e.g. `read_body(buf)`), which would be awkward to implement without `HttpTrigger::new` being able to link in that interface.

lann:
@fibonacci1729 @itowlson I made :point_up: that change: https://github.com/fermyon/spin/pull/382/commits/9adf7489b95151168f3e8a8ebc4464904808b4a0

itowlson:
I like it!

radu-matei (approved):
This looks great!

Thanks!

Mossaka:
Thanks a lot!

PR #384, opened 2022-04-21
docs(sip): add `spin build` SIP

radu-matei:

### DIFF
--- a/docs/content/sips/index.md
+++ b/docs/content/sips/index.md
@@ -4,6 +4,14 @@ date = "2022-03-22T14:53:30Z"
 
 ---
 
-- [Spin Deploy](001-spin-deploy.md)
-- [Application Configuration](002-app-config.md)
-- [Trigger Executors](003-trigger-executors.md)
+A SIP is a Spin Improvement Proposal. It is a design document that describes a
+new feature for Spin. It is intended to be the primary mechanism for proposing
+major work items in the Spin project, and for documenting the design decisions
+that have gone into the project.
+
+Here is a list of accepted SIPs:
+
+- [Spin Deploy](../sips/001-spin-deploy.md)
+- [Application Configuration](../sips/002-app-config.md)
+- [Trigger Executors](../sips/003-trigger-executors.md)
+- [Spin Build](../sips/004-spin-build.md)
--- a/docs/content/sips/001-spin-deploy.md
+++ b/docs/content/sips/001-spin-deploy.md
@@ -1,6 +1,6 @@
 title = "SIP 001 - Spin Deploy"
 template = "main"
-
+date = "2022-03-18T14:53:30Z"
 ---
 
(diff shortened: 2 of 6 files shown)


docs/content/sips/00x-spin-build.md
```diff
@@ -0,0 +1,46 @@
...
+field on the component configuration in a local `spin.toml` manifest file:
+
+```toml
+[[component]]
+id = "hello"
+source = "target/wasm32-wasi/release/spinhelloworld.wasm"
+
+[[component.build]]
+command = [ "cargo", "build", "--release", "--target", "wasm32-wasi" ]
+```
```
radu-matei, line 34: I think this represents a good starting point for the MVP.

radu-matei:
ref #352

docs/content/sips/00x-spin-build.md
lann: I wonder if `build` could be an alternative kind of component source. Then `spin up` could check for the existence of e.g. `component.build.output` and run the build if necessary.

lann: Could also have a `rebuild-if-changed = "src/lib.rs"`

itowlson: At that point we can add an option to email the author on build failures, and by Zawinski's law we'll be done.

Kidding aside, let's not try to fit some partial implementation of `make` into Spin.  I like `spin build` as a convenient alternative to running manual build commands across multiple components.  Beyond that, trust each component's build tooling do its thing.

lann: Fair enough, though I think it would be worthwhile to "bless" (via docs, examples) a particular set of tools that make a good dev experience.

radu-matei:
This PR also updates the SIP index page and individual SIPs to render properly.
(I can split this into a separate PR if we want, I just thought it might be easier to get away with it 😅)

docs/content/sips/index.md
```diff
@@ -4,6 +4,14 @@ date = "2022-03-22T14:53:30Z"
...
-- [Application Configuration](002-app-config.md)
-- [Trigger Executors](003-trigger-executors.md)
+A SIP is a Spin Improvement Proposal. It is a design document that describes a
+new feature for Spin. It is intended to be the primary mechanism for proposing
+major work items in the Spin project, and for documenting the design decisions
+that have gone into the project.
+
+Here is a list of accepted SIPs:
+
+- [Spin Deploy](../sips/001-spin-deploy.md)
```
lann, line 14: Is the `../sips` necessary here? :thinking:

lann (approved):

docs/content/sips/index.md
radu-matei: It is for how Bartholomew currently does relative paths that are not in the root of the `content/` directory.
(will open an issue to track that in Bartholomew)

PR #390, opened 2022-04-22
test(sdk/go): Add integration tests for go sdk

adamreese:


lann:
MacOS failure looks legit.

Issue #385, opened 2022-04-22
Feature: Include `trigger.base` in "Serving HTTPS on address https://..." message 


lann:
For apps that set it, it would be a helpful reminder of what URL to use for testing.

~Bonus points: if there is only a single http trigger component, append its `component.trigger.route`, too.~
Even better: :point_down:

radu-matei:
ref #177 

Even for multiple HTTP components, it would be helpful to see all routes as well, in something like a "summary".

PR #392, opened 2022-04-24
feat: print available routes at startup of spin up

toVersus:
I'm not sure this is a preferred way to print available routes to stdout, but implement `Display` for `RoutePattern` enum and combine it with base url.

Add some dummy routes to `examples/http-rust/spin.toml`:

```toml
spin_version = "1"
authors = ["Fermyon Engineering <engineering@fermyon.com>"]
description = "A simple application that returns hello."
name = "spin-hello-world"
trigger = { type = "http", base = "/v1" }
version = "1.0.0"

[[component]]
id = "hello"
[component.trigger]
route = "/hello"

[[component]]
id = "goodby"
[component.trigger]
route = "/goodby"

[[component]]
id = "goodnight"
[component.trigger]
route = "/goodnight"
```

Run `spin up`:

```sh
❯ ../../target/release/spin up
Serving HTTP on address http://127.0.0.1:3000
Available Routes:
  http://127.0.0.1:3000/v1/hello
  http://127.0.0.1:3000/v1/goodby
  http://127.0.0.1:3000/v1/goodnight
```

Fixes: #385

### DIFF
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -6,7 +6,7 @@ use anyhow::{bail, Result};
 use http::Uri;
 use indexmap::IndexMap;
 use spin_manifest::{Application, CoreComponent};
-use std::fmt::Debug;
+use std::fmt;
 use tracing::log;
 
 // TODO
@@ -61,7 +61,7 @@ impl Router {
     /// if no component matches.
     /// If there are multiple possible components registered for the same route or
     /// wildcard, this returns the last one in the components vector.
-    pub(crate) fn route<S: Into<String> + Debug>(&self, p: S) -> Result<CoreComponent> {
+    pub(crate) fn route<S: Into<String> + fmt::Debug>(&self, p: S) -> Result<CoreComponent> {
         let p = p.into();
 
         let matches = &self
@@ -143,6 +143,15 @@ impl RoutePattern {
         match s.strip_suffix('/') {
             Some(s) => s.into(),
             None => s,
+        }
+    }
+}
+
+impl fmt::Display for RoutePattern {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self {
+            RoutePattern::Exact(path) => write!(f, "{}", path),
+            RoutePattern::Wildcard(pattern) => write!(f, "{} (wildcard)", pattern),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


crates/http/src/routes.rs
```diff
@@ -6,6 +6,7 @@ use anyhow::{bail, Result};
 use http::Uri;
 use indexmap::IndexMap;
 use spin_manifest::{Application, CoreComponent};
+use std::fmt;
```
radu-matei, line 9: nit: could you please merge the two `std` imports?

crates/http/src/routes.rs
```diff
@@ -147,6 +148,15 @@ impl RoutePattern {
     }
 }
 
+impl fmt::Display for RoutePattern {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self {
+            RoutePattern::Exact(path) => write!(f, "{}", path),
+            RoutePattern::Wildcard(pattern) => write!(f, "{}", pattern),
```
radu-matei, line 155: I think there should be a way to indicate this is a wildcard pattern — how about `/...`? We use it already to define such a pattern anyway.

radu-matei (commented):
This is a really good approach and mostly looks good to me.

I have a question about representing wildcard patterns.
Thanks!

crates/http/src/routes.rs
toVersus: Yeah, I agree to distinguish wildcard pattern from exact path, but using `/...` identifier will cause it unclickable from the terminal. Is that fine?

crates/http/src/routes.rs
toVersus: Done.

crates/http/src/routes.rs
radu-matei: Ah, good point.
Perhaps something like: `write!(f, "{} (wildcard)", pattern)` ?

(I don't know what the best solution would be here either.)

toVersus: Oh, that's the exact same idea when I first implemented this functionality :)
I take this solution here and hope someone will come up with the best solution. Done.

radu-matei (approved):
This looks great, thank you so much for your contribution, @toVersus!

LGTM

PR #399, opened 2022-04-26
feat(sdk/go): Add SIP-002 "Application Configuration" to TinyGo SDK

adamreese:


examples/config-tinygo/readme.md
```diff
@@ -0,0 +1,76 @@
...
+content-type: text/plain; charset=utf-8
+content-length: 16
+date: Wed, 09 Mar 2022 17:23:08 GMT
+
+Hello, Fermyon!
+```
+
+## Notes
+
+- components built using TinyGo will be run in Spin using the Wagi executor
```
radu-matei, line 74: I don't think this readme is accurate anymore.

examples/config-tinygo/main.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+
+	spinconfig "github.com/fermyon/spin/sdk/go/config"
```
radu-matei, line 7: I'm torn between prefixing `spin*` before packages everywhere, or just where it clashes with standard library packages.
(nothing to change here, just wondering what your thoughts are)

examples/config-tinygo/main.go
```diff
@@ -0,0 +1,22 @@
...
+	spinhttp "github.com/fermyon/spin/sdk/go/http"
+)
+
+func init() {
+	spinhttp.Handle(func(w http.ResponseWriter, r *http.Request) {
+		if val, err := spinconfig.Get("message"); err != nil {
```
radu-matei, line 13: Would you mind adding a comment here describing where the value is coming from?
Thanks!

examples/config-tinygo/readme.md
adamreese: copy-pasta

examples/config-tinygo/main.go
adamreese: I think only where it clashes.

radu-matei (approved):
Thanks!

LGTM

PR #400, opened 2022-04-26
Basic template support

itowlson:
This is a baseline providing basic template support, on which we can iterate (indeed, @fibonacci1729 is already doing so).  Thanks and credit to Brian for the core work which got folded into this major refactor - the good bits are all him, the bad bits are all me.

What's in this drop:

* Scaffold new application with starter component
* The proposed new template layout
* Parameter substitution for strings
* Updated management commands (`spin templates install/list/uninstall`)
* Brian's templates
* ~A~ Two slightly grumpy messages about `cargo fmt`

What's not in this drop:

* ~Template authoring docs~ - there's now a rough cut document - it's going to need reworking but may be better to hold off until the feature is more ready?
* Crate docs (because the API is still in flux, but they need to be added back!)
* ~Brian's Go templates~
* File exclusions (e.g. if your template contains a Handlebars file we may attempt to substitute into it!)
* Non-string parameters
* Constraints other than regex
* Scaffold empty application
* Add component to existing application
* Some errors still lacks context
* ETA: Tests!!!
* Lots of other stuff

To install and run the Rust templates, check out and build this branch and:

```
spin templates install --dir .
spin new http-rust -o myspinapp
```


lann:
Should templates be a `feature` of the cli?

crates/templates/src/run.rs
```diff
@@ -0,0 +1,334 @@
...
+};
+
+use anyhow::{anyhow, Context};
+use itertools::Itertools;
+use path_absolutize::Absolutize;
+use walkdir::WalkDir;
+
+use crate::template::{Template, TemplateParameter};
+
+// TODO: better name, e.g. Gather???
```
lann, line 13: Render?

crates/templates/src/source.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[derive(Debug)]
+pub enum TemplateSource {
+    Git(Url),
+    File(PathBuf),
+}
+
+impl TemplateSource {
+    pub fn try_from_git(git_url: impl AsRef<str>) -> anyhow::Result<Self> {
+        let url_str = git_url.as_ref();
```
lann, line 20: style: I like shadowing function params when you're just `.as_ref`/`.into`ing them immediately, e.g. `let git_url = git_url.as_ref();`

crates/templates/src/source.rs
```diff
@@ -0,0 +1,105 @@
...
+            ))
+        }
+    }
+}
+
+async fn clone_local(url: &Url) -> anyhow::Result<LocalTemplateSource> {
+    let temp_dir = tempdir()?;
+    let path = temp_dir.path().to_owned();
+
+    let url_str = url.to_string();
```
lann, line 72: Seems unnecessary.

crates/templates/src/source.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let mut git = Command::new("git");
+    git.arg("clone");
+
+    // if let Some(b) = branch {
+    //     git.arg("--branch").arg(b);
+    // }
```
lann, line 79: Remove? Comment?

crates/templates/src/store.rs
```diff
@@ -0,0 +1,96 @@
...
+impl TemplateStore {
+    pub(crate) fn new(root: impl AsRef<Path>) -> Self {
+        Self {
+            root: root.as_ref().to_owned(),
+        }
+    }
+
+    pub(crate) fn default() -> anyhow::Result<Self> {
+        let data_dir = dirs::data_local_dir()
+            .or_else(|| dirs::home_dir().map(|p| p.join(".fermyon")))
```
lann, line 25: I think this should probably be `.spin`?

templates/http-go/metadata/manifest.toml
```diff
@@ -0,0 +1,9 @@
+template_version = "1"
```
lann, line 1: I think we need to somehow indicate that this is a "_Spin_ template manifest". Some options, in lann-preference-order:

1. rename the `metadata` dir to something like `spin-template`, maybe even `.spin-template`
1. rename `manifest.toml` to `spin-template.toml`
1. change this key to `spin_template_version`

crates/templates/src/template.rs
```diff
@@ -0,0 +1,164 @@
...
+        self.data_type.validate_value(value.as_ref().to_owned())
+    }
+}
+
+impl TemplateParameterDataType {
+    fn parse(raw: &RawParameter) -> anyhow::Result<Self> {
+        match &raw.data_type[..] {
+            "string" => Ok(Self::String(parse_string_constraints(raw)?)),
+            "number" => Ok(Self::Number),
+            "boolean" => Ok(Self::Boolean),
```
lann, line 146: If I'm interpretting correctly, it looks like we're allowing these types in the manifest, but then panicing at runtime? That doesn't seem ideal...

crates/templates/src/environment.rs
```diff
@@ -0,0 +1,89 @@
+use anyhow::Result;
+use git2::{Config as GitConfig, Repository as GitRepository};
```
lann, line 2: This is going to break OpenSSL again. We need to shell out to `git`

crates/templates/src/filters.rs
```diff
@@ -0,0 +1,88 @@
+use heck::{ToKebabCase, ToSnakeCase, ToUpperCamelCase};
+use liquid_core::{Filter, ParseFilter, Runtime, ValueView};
+use liquid_derive::FilterReflection;
+
+// Filters that are added to the Liquid parser and allow templates to specify
+// transformations of input strings
+// TODO: ^^ I THINK. Check with Brian
```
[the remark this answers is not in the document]
fibonacci1729: Yep. That comment is correct.

crates/templates/src/lib.rs
```diff
@@ -1,230 +1,20 @@
...
-                let v = String::from_utf8_lossy(&output.stdout).trim().to_string();
-                if v.is_empty() {
-                    None
-                } else {
-                    Some(v)
-                }
-            })
-    }
-}
+#![allow(missing_docs)]
```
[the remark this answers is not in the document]
itowlson: Yes, as noted I skipped these while things were in flux.

crates/templates/src/interaction.rs
```diff
@@ -0,0 +1,39 @@
...
+
+pub(crate) fn prompt_parameter(parameter: &TemplateParameter) -> Option<String> {
+    let prompt = parameter.prompt();
+    let default_value = parameter.default_value();
+
+    loop {
+        let input = match parameter.data_type() {
+            TemplateParameterDataType::String(_) => ask_free_text(&prompt, &default_value),
+            TemplateParameterDataType::Number => todo!(),
+            TemplateParameterDataType::Boolean => todo!(),
```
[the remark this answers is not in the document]
itowlson: Brian is iterating on these even as we speak, but I can take these cases out for now.

crates/templates/src/run.rs
```diff
@@ -0,0 +1,334 @@
...
+    pub name: Option<String>,
+    pub output_path: Option<PathBuf>,
+    pub values: HashMap<String, String>,
+}
+
+pub enum Cancellable<T, E> {
+    Cancelled,
+    Err(E),
+    Ok(T),
+}
```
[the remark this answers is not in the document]
itowlson: Cancelled is not an error - Cancelled is a valid choice to not proceed.

Though, as it turns out, I think cancelled can only happen by the user Ctrl+C-ing the program, so it's probably a non-issue...

crates/templates/src/run.rs
```diff
@@ -0,0 +1,334 @@
...
+    match std::str::from_utf8(bytes) {
+        Ok(s) => Some(s.to_owned()),
+        Err(_) => None, // TODO: try other encodings!
+    }
+}
+
+impl TemplateOutputs {
+    pub async fn write(&self) -> anyhow::Result<()> {
+        for (path, contents) in &self.files {
+            let dir = path.parent().unwrap(); // TODO: replace unwrap with error handler
```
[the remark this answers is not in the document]
itowlson: ooooooooooooooooooooooooooooooooh - TIL - thanks!

templates/http-go/metadata/manifest.toml
itowlson: Hah, I was thinking we should change `manifest.toml` to `spin-template.toml` or some such!

I was also thinking we might want to let authors version their templates (for diagnostic purposes).  This field is really the "schema version" - maybe there should be a true "template version" field as well...

crates/templates/src/lib.rs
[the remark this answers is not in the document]
lann: Ohhh it helps if I read the _entire_ description, doesn't it?

crates/templates/src/template.rs
itowlson: Well, these were placeholders for Brian.  I can remove them and he can bring them back though - probably safer!

crates/templates/src/run.rs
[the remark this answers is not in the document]
lann: I learned last week!

templates/http-go/metadata/manifest.toml
lann: Aha! I'm fine with no explicit schema version - you can always treat its absence as version "0" if you end up needing it.

crates/templates/src/source.rs
itowlson: Replaced with `as_str` which saves a clone (which is what this alluded to)

crates/templates/src/store.rs
itowlson: That would result in `~/.spin/spin` for folks who don't have a `data_local_dir` though.

This is probably a broader question on where we like to store stuff and merits some bikeshedding.

I now also notice that if the user does _not_ have a `data_local_dir()`, and then later does something that causes them to have one, then their installed templates will disappear.  Not sure whether this is an issue or what to do about it though.

lann: A quick perusal of the code suggests that `data_local_dir` won't return `None` for any major OS.

templates/http-go/metadata/manifest.toml
itowlson: * Changed file name to `spin-template.toml`
* Changed field name to `manifest_version`

Tentatively resolving this but happy to keep iterating!

itowlson:
@lann re templates as a `feature` flag.  There's probably a broader context around features and embedding.  Certainly it should be possible to build Spin without templates, but the solution to that may be to move it to the eventual plug-in model.  For now, nothing else depends on the template crate, so people embedding the runtime would not need to pick it up (at least, assuming they import the crates rather than the bin); and we don't yet have a structure in place for CLI builds than include and exclude features.

But I feel like I am probably missing your point...?!

lann (commented):
Gah, this is why I don't like GH reviews. I wrote these comments hours ago but forgot to submit...

lann (dismissed):

lann:
@itowlson 
> re templates as a `feature` flag

I'm mostly concerned with compilation times at this point, but being able to ship just the code for `spin up` to servers would be good too.

crates/templates/src/run.rs
```diff
@@ -186,12 +196,32 @@ impl Run {
...
 
+    fn populate_parameters_silent(&self) -> anyhow::Result<Option<HashMap<String, String>>> {
+        let mut values = HashMap::new();
+        for parameter in self.template.parameters() {
+            match self.populate_parameter_silent(parameter) {
+                Ok(v) => {
+                    values.insert(parameter.id().to_owned(), v);
+                }
+                Err(e) => return Err(e),
+            }
```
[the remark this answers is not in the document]
itowlson: Yes, you're right.  This was down to hasty copying of the `interactive` populator to the `silent` populator with a view to subsequent unification.  Thanks!

docs/content/template-authoring.md
```diff
@@ -0,0 +1,86 @@
...
+template = "main"
+date = "2022-03-14T00:22:56Z"
+[extra]
+url = "https://github.com/fermyon/spin/blob/main/docs/content/template-authoring.md"
+---
+
+Spin templates allow a Spin developer to quickly create the skeleton of an
+application or component, ready for the application logic to be filled in.
+
+A template consists of two directories, `content` and `metadata`.
```
lann, line 11: Somewhat :bike::derelict_house:y: what is the rationale for having both of these directories? It seems like one of them could be flattened into the root?

src/commands/new.rs
```diff
@@ -1,27 +1,75 @@
...
+                name: name.to_owned(),
+                value: value.to_owned(),
+            })
+        } else {
+            Err(anyhow!("'{}' should be in the form name=value", s))
+        }
     }
 }
+
+fn to_hash_map(values: &[ParameterValue]) -> HashMap<String, String> {
```
lann, line 70: minor: Could be `impl Into<HashMap<String, String>> for &[ParameterValue]`?

docs/content/template-authoring.md
itowlson: Because it _guarantees_ a clean separation.  In future a template might have scripts, optional files, IDE icons, who knows what; by hiving off the execution stuff from the payload we allow arbitrary extension without any risk of a future clash.

That said, we could say that the manifest goes in the root and the payload in `content`, and when we add other things they can go in a `metadata` or other more specifically named directory at that point.  It doesn't really much matter.  But I don't want to end up with a grab bag of stuff in the root and the `content` kind of lost among it all.

src/commands/new.rs
itowlson: image

itowlson: I have a feeling I tried that before and was thwarted in this exact way!

docs/content/template-authoring.md
lann: Having the manifest in the root feels more "familiar" to me, but I think its just an aesthetic preference.

src/commands/new.rs
lann: Ah yeah, I was thinking you could implement for `&[LocalType]`, but I guess it makes sense for coherence that you can't.

docs/content/template-authoring.md
fibonacci1729: I am critically _weary _of the cost of trying to "gaurantee" this clean separation. 

1, it constrains the flexibility a template author has over how they organize their repository and code they maintain. 
2. The assumptions here is that `spin-template.toml` must exist in the parent of any directory named `content` and `metadata` to qualify as a template. This implies that as an author I am required to duplicate parameter definitions, licenses into each tempate's metadata directory on a  *per* template basis. 
3. As a template author, I want to frame the interactive experience for a consumer (done so through `spin-template.toml`) however, given the current design, I am forced to develop a unique (or duplicate) my interactive experience for each template i offer.

fibonacci1729: I am heavily leaning towards the only assumption about a repository is it contains a (or many) `spin-template.toml` somewhere in the repo. Everything part of the template is relative to where the `spin-template.toml` is located. 

Basically, the way the original prototype worked

fibonacci1729 (approved):
This is a great first pass! Organized & clean code that we can build on to achieve some of our out-of-scope for this PR goals. Feel free to resolve my comments, I dont think this PR is the time/place to address them. Thanks alot Ivan!

lann (approved):

radu-matei (approved):
I think we've agreed that this is a really good starting point for the templates work that we can iterate on.
Could you please rebase this, and it should be good to merge?

Thanks!

LGTM

PR #401, opened 2022-04-26
Improve config error formatting

### LINKED ISSUE #397: Partially escaped error strings are returned from get-config
Errors from get-config are returned as `InvalidPath("no slot at path Path(\"spin_config_tinygo.noop\")")`

lann:
`impl Display for Path` and use it in errors.

Fixes #397


adamreese:
Looks great!  I think it would be easier to read if we separate the path from the message

```
InvalidPath("no slot at path: spin_config_tinygo.bad")
```

adamreese (approved):

itowlson:
If we are doing this, should we remove all the `path.display()` calls?  I have been using that - I assumed there was a reason Path had the method rather than implementing Display directly.

lann:
> If we are doing this, should we remove all the `path.display()` calls? I have been using that - I assumed there was a reason Path had the method rather than implementing Display directly.

This is `Path`, not `Path`... :neutral_face: 

...I really need to change that type name.

PR #404, opened 2022-04-26
Migrate outbound-http to a HostComponent

lann:
Add HostComponentsStateHandle::get to support wasi-experimental-http's
slightly different add_to_linker signature.


radu-matei (approved):
This is such an elegant solution, thanks!

LGTM

Issue #402, opened 2022-04-26
`trigger` crate


lann:
Spin currently has two built-in trigger types: `HttpTrigger` and `RedisTrigger`. There is also a `TimerTrigger` example and ongoing work towards new cron / "Schedule" (#365) and "CloudEvents" (#394) triggers.

These triggers - especially the two built-in types - share some common logic in how they prepare and invoke their handler components. In order to reduce duplication and future divergence, we should extract that commonality into a new `trigger` crate that exposes trait(s) that can be implemented by individual trigger types.

Mossaka:
I can create a trigger trait. This will help me to work on the CloudEvents trigger feature

Mossaka:
I am thinking about merging the two triggers, HTTP and Redis, to one crate called `spin-trigger`. The `spin-trigger` crate exposes a `Trigger` trait

```rust
#[async_trait]
pub trait Trigger<T> {
    pub async fn new(
        builder: Builder<T>,
        app: Application<CoreComponent>,
        config: TriggerConfig<T>,
    ) -> Result<Self>;
    async fn run(&self) -> Result<()>;
}
```

and changes our we import triggers:
```rust
use spin_trigger::{HttpTrigger, RedisTrigger, CloudEventTrigger};
use spin_engine::{Builder, ExecutionContextConfiguration};

fn main() {
  ...
  let builder = self.prepare_ctx_builder(app.clone()).await?;
  let trigger = HttpTrigger::new(builder, app, self.opts.address, tls).await?;
  trigger.run().await?;

}
```

The benefits of doing this are
1. easy to author a trigger
2. easy to import a trigger
3. author a trigger

Let me know your thoughts @lann @radu-matei

radu-matei:
I agree with having a crate that contains the common logic between the triggers, but not sure with merging our triggers into this crate.
Rather, I would expect the existing triggers to remain separate crates that implement the trait from the new one and import the common logic.

Unless you are thinking about something else with merging the existing triggers?

What do you think?

lann:
I haven't put a _ton_ of thought into this yet, but I was vaguely thinking that the `trigger` crate would consist of one or more traits to define the trigger interface plus some common logic to be shared between (most) triggers. Specifically, I'd be happy if we could move this:
https://github.com/fermyon/spin/blob/ad51cdde94517aa7d6824fdc53b420d993d26389/src/commands/up.rs#L158-L172

Ideally we could also come up with generic versions of some of these (and anything else that makes sense!):
[...]

Issue #413, opened 2022-04-29
High memory consumption when running a Spin application from Bindle


radu-matei:
For Spin applications with large number of components and a few hundred static
assets, we have seen significantly increased memory allocation when running from
Bindle compared to running locally. This is a meta-issue that attempts to track
down the root causes of that.

TL; DR: There are three main causes for increased memory consumption when
running a Spin application directly from Bindle (in no particular order):

1. The Bindle loader attempting to pull all static assets in parallel.
2. Expensive clones of core components (and applications).
3. Not caching compiled Wasm modules (applies to both modes of running Spin).

---
1. The Bindle loader attempting to pull all static assets in parallel

Each new Tokio thread created
[allocates by default 2MiB of memory on the stack](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.thread_stack_size).

When preparing a component from Bindle (preparing its assets),
[all components are handled in parallel](https://github.com/fermyon/spin/blob/main/crates/loader/src/bindle/mod.rs#L84),
and all parcels are again
[handled in parallel](https://github.com/fermyon/spin/blob/main/crates/loader/src/bindle/assets.rs#L55).

This results in N (the total number of static assets in an application) Tokio
threads, so a spike of N x 2MiB in memory when pulling the static assets. This
is memory that is allocated by default, but as far as I can see, not
necessarily used.

(This also results in errors because of too many open files / connections in multiple OSes.)

Attached is the stack trace for preparing the application (there are ~200 Tokio
threads that I can see spawned to handle static assets for Fermyon.com, which
seems to roughly match the number of assets we have):

```rust
std::sys::unix::thread::Thread::new::h56395654af139df9
std::thread::Builder::spawn::hcf8ef13595f288d8
tokio::runtime::blocking::pool::Spawner::spawn::he31007656ba14ed0
tokio::runtime::handle::Handle::spawn_blocking::hb7a1d001cd691f95
_$LT$hyper..client..connect..dns..GaiResolver$u20$as$u20$tower_service..Service$LT$hyper..client..connect..dns..Name$GT$$GT$::call::h85c1e5c145e062be
bindle::client::Client$LT$T$GT$::get_parcel_request::_$u7b$$u7b$closure$u7d$$u7d$::hd2dbfcf845ca9e75
[...]

PR #415, opened 2022-04-29
trigger crate prototype

lann:
Currently just for discussion in #402

### DIFF
--- a/crates/trigger/src/lib.rs
+++ b/crates/trigger/src/lib.rs
@@ -1,0 +1,68 @@
+use std::{convert::TryFrom, error::Error, path::PathBuf};
+
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use spin_engine::{Builder, ExecutionContext, ExecutionContextConfiguration};
+use spin_manifest::{Application, ApplicationTrigger, ComponentMap, CoreComponent, TriggerConfig};
+
+#[async_trait]
+pub trait Trigger: Sized {
+    type ContextData: Default + 'static;
+    type Config;
+    type ComponentConfig;
+    type RunConfig;
+
+    fn new(
+        execution_context: ExecutionContext<Self::ContextData>,
+        config: Self::Config,
+        component_configs: ComponentMap<Self::ComponentConfig>,
+    ) -> Result<Self>;
+
+    fn configure_execution_context(builder: &mut Builder<Self::ContextData>) -> Result<()> {
+        configure_execution_context_defaults(builder)
+    }
+
+    async fn run(self, config: Self::RunConfig) -> Result<()>;
+}
+
+pub fn configure_execution_context_defaults<T: Default + 'static>(
+    builder: &mut Builder<T>,
+) -> Result<()> {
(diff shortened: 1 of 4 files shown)


lann:
Note that this specific design would require a bit of refactoring in the http trigger's `Router::build`, but nothing too hard I think.

radu-matei:
The `Router` implementation is also one of the causes in #413, and solving it will require some refactoring anyway.

crates/trigger/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+pub trait Trigger: Sized {
+    type ContextData: Default + 'static;
+    type Config;
+    type ComponentConfig;
+    type RunConfig;
+
+    fn new(
+        execution_context: ExecutionContext<Self::ContextData>,
+        config: Self::Config,
+        component_configs: ComponentMap<Self::ComponentConfig>,
```
Mossaka, line 18: we can simply pass `Application<CoreComponent>` to new, instead of `config` and `component_configs`. Is this design intentionally trying to decouple Application and Trigger?

crates/trigger/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    <T::Config as TryFrom<ApplicationTrigger>>::Error: Error + Send + Sync + 'static,
+    <T::ComponentConfig as TryFrom<TriggerConfig>>::Error: Error + Send + Sync + 'static,
+{
+    let mut builder = Builder::<T::ContextData>::new(ExecutionContextConfiguration {
+        components: app.components,
+        label: app.info.name,
+        log_dir: opts.log_dir,
+        config_resolver: app.config_resolver,
+    })?;
+    T::configure_execution_context(&mut builder)?;
```
Mossaka, line 58: Maybe we should have one prepare trigger function that prepares builder, and then leave the `Trigger::new` and `trigger.run` to the up command. 

```rust
  pub async fn prepare_trigger<T: Trigger>(
    app: Application<CoreComponent>, 
    opts: RunOptions<T>,
  ) -> (Builder<ContextData>, Config, ComponentConfig) {}
```

This allows SDK users to customize `builder` and `configs`.

crates/trigger/src/lib.rs
lann: Yes it's intentionally decoupling, but it also avoids needing to `clone` the `Application`, which seems to be part of the issue in #413.

crates/trigger/src/lib.rs
lann: In this PR's design the `builder` can be customized up in `Trigger::configure_execution_context`; that is just a _default_ implementation of that trait function.

That said, I'm not opposed to leaving more the logic in `spin up` at this point. This prototype was partially my attempt at figuring out what in implementation of #417 might look like if we move forward with that.

crates/trigger/src/lib.rs
Mossaka: it seems like Http trigger will need application to build routers

lann:
Implementation in #418

PR #417, opened 2022-04-29
SIP: Trigger Executors

### LINKED ISSUE #302: pre-sip: Trigger "plugins"
Currently we have "http" and "redis" triggers hard-coded in the `spin` bin. We also have `examples/spin-timer` which implements a `TimerTrigger`, but it is missing a lot of the functionality of `spin up`. Third-party triggers would be in the same position.

What if triggers were run by `exec`ing them as separate binaries? We could define a "spin trigger plugin interface" with certain arguments, env vars, open file descriptors, etc. that an external rust binary could implement by linking `spin/engine` et al (like `spin-timer` does). Then `rust up` would be responsible for configuration, fetching/unpacking components, etc. and then `exec`ing that binary.

Git (among other tools) uses a similar model for several features like ["remote helpers"](https://git-scm.com/docs/gitremote-helpers).

We could even have built-in triggers like "http" use this same interface by "reexecing" them as e.g. `spin internal http-trigger` subcommands.

lann:
Fixes #302

### DIFF
--- a/docs/content/sips/xxx-trigger-executors.md
+++ b/docs/content/sips/xxx-trigger-executors.md
@@ -1,0 +1,68 @@
+title = "SIP xxx - Trigger Executors"
+template = "main"
+
+---
+
+Summary: A new architecture for Spin trigger executors.
+
+Owner: lann.martin@fermyon.com
+
+Created: April 29, 2022
+
+<!-- Updated: April 4, 2022 -->
+
+## Background
+
+Spin currently includes built-in implementations for two trigger types: HTTP, and Redis (PubSub).
+We want to allow for additional trigger types in the future, whether they are "internal" types developed and
+supported within the Spin codebase, or "external" types developed by the broader community.
+
+While it is currently possible to build external trigger types - as in the
+[TimerTrigger example](https://github.com/fermyon/spin/blob/main/examples/spin-timer/src/main.rs) - it requires replicating much of the
+functionality provided by the `spin up` command.
+
+## Proposal
+
+### Trigger types are implemented as "trigger executors"
+
+These executors are regular subcommands/binaries, but are not normally executed directly by
+users.
+
+* "Internal" triggers use `spin` subcommands, e.g. `spin trigger http`
+* "External" triggers use external binaries, e.g. `spin-trigger-timer`
+* Trigger executors share common code, including common CLI options in a `spin-trigger` crate
+
+### The `spin up` command changes
+
+It remains the primary entrypoint for starting a Spin app, but its behavior is changed to:
+
+1. Load the application manifest
+1. Find a trigger executor for the application (e.g. check the `PATH` for `"spin-trigger-${trigger.type}"`)
+1. Set some environment variables:
+   * `SPIN_MANIFEST_URL`: a URL pointing to the application manifest (e.g. `file:///path/to/spin.toml` or `bindle+https://bindle-server?id=bindle-id`)
+   * `SPIN_TRIGGER_TYPE` the spin trigger type (e.g. `http`) being executed
+1. `exec` the trigger executor, forwarding its own CLI arguments
+
+## Future design options
+
+_This section contains possible future features which are not fully defined here._
+
+### Multi-trigger orchestration
+
+If multiple trigger types were to be supported by a single application, the role of `spin up` could
+grow to include trigger orchestration:
+
+* Finding executors for _all_ required trigger types
+* Forking a child process for each trigger executor
+* Managing the child processes (restarts, graceful shutdowns, etc.)
+
+## Potential disadvantages
+
+If the "Multi-trigger orchestration" option is taken, having separate processes for different triggers
+might have some disadvantages:
+
+* Multiple triggers types would use more system resources implemented as multiple processes rather than as
+  multiple async tasks in the same process, especially if a single module implements multiple trigger type
+  handlers.
+* There has been discussion the possibility of trigger handlers making optimized "in-memory" calls to
+  other triggers. This design would make cross-trigger-type calls of this sort more complex at best.


docs/content/sips/xxx-trigger-executors.md
```diff
@@ -0,0 +1,60 @@
...
+* Find a trigger executor for that type (e.g. check the `PATH` for `"spin-trigger-${type}"`)
+* `exec` the trigger executor, forwarding its own CLI arguments
+
+## Future design options
+
+_This section contains possible future features which are not fully defined here._
+
+### Multi-trigger orchestration
+
+If multiple trigger types ere to be supported by a single application, the role of `spin up` could
```
radu-matei, line 44: typo: _ere_

radu-matei: Wow that was fast. 😂

lann: Says the guy who submitted a review 2 minutes after the PR was created...

Mossaka:
Like this proposal! Just one question: what does`spin trigger http` command do - start a server that listens to some host:port? 

I also think it will be cool to have a command plugin for 3rd party triggers so people can use `spin trigger <3rd party trigger>` just like internal triggers.

lann:
> what does`spin trigger http` command do - start a server that listens to some host:port?

Yep! It would basically function identically to the current `spin up` when applied to an HTTP application.

> I also think it will be cool to have a command plugin for 3rd party triggers so people can use spin trigger <3rd party trigger> just like internal triggers.

That would probably be fine, but I think the `spin trigger *` subcommands should be a semi-hidden implementation detail, at least initially. Users would generally continue to use `spin up`.

docs/content/sips/xxx-trigger-executors.md
```diff
@@ -0,0 +1,65 @@
...
+<!-- Updated: April 4, 2022 -->
+
+## Background
+
+Spin currently includes built-in implementations for two trigger types: HTTP, and Redis (PubSub).
+We want to allow for additional trigger types in the future, whether they are "internal" types developed and
+supported within the Spin codebase, or "external" types developed by the broader community.
+
+While it is currently possible to build external trigger types - as in the
+[TimerTrigger example](../../../examples/spin-timer/src/main.rs) - it requires replicating much of the
```
radu-matei, line 21: nit: Could you make this an absolute GitHub URL?
(when rendered in the docs website, the relative path won't work)

Mossaka (commented):
Does spin intend to make all triggers binaries, or just the "external triggers"?

lann:
> Does spin intend to make all triggers binaries, or just the "external triggers"?

This proposal would make "internal" and "external" triggers be executed in the same way: by forking a new process. While the "internal" triggers _could_ be separate binaries, I think for user-friendliness we would release them as part of the `spin` binary, where - for example - `spin up` would fork `spin trigger http` as a new process.

docs/content/sips/xxx-trigger-executors.md
```diff
@@ -0,0 +1,65 @@
...
+## Proposal
+
+### Trigger types are implemented as "trigger executors"
+
+These executors are regular subcommands/binaries, but are not normally executed directly by
+users.
+
+* "Internal" triggers use `spin` subcommands, e.g. `spin trigger http`
+* "External" triggers use external binaries, e.g. `spin-trigger-timer`
+* Trigger executors share common code, including common CLI options in a `spin-trigger` crate
```
itowlson, line 33: Do we need to say anything about versioning here - if `spin up` and `spin-trigger-foo` understand different sets of CLI options?

docs/content/sips/xxx-trigger-executors.md
```diff
@@ -0,0 +1,65 @@
...
+
+### The `spin up` command changes
+
+It remains the primary entrypoint for starting a Spin app, but its behavior is changed to:
+
+* Detect the trigger type for an application
+* Find a trigger executor for that type (e.g. check the `PATH` for `"spin-trigger-${type}"`)
```
itowlson, line 40: What are the ramifications of different strategies here?  E.g. also checking current directory, or directory that contains `spin` - might improve convenience but are there any safety issues?

docs/content/sips/xxx-trigger-executors.md
```diff
@@ -0,0 +1,65 @@
...
+## Potential disadvantages
+
+If the "Multi-trigger orchestration" option is taken, having separate processes for different triggers
+might have some disadvantages:
+
+* Multiple triggers types would use more system resources implemented as multiple processes rather than as
+  multiple async tasks in the same process, especially if a single module implements multiple trigger type
+  handlers.
+* There has been discussion the possibility of trigger handlers making optimized "in-memory" calls to
+  other triggers. This design would make cross-trigger-type calls of this sort more complex at best.
```
itowlson, line 65: We've also talked about a control plane-execution plane architecture in which the control plane might start applications/triggers in an already running process (subject to policy).  That may be too vague to think about in detail but might be worth a handwave. *grin*

docs/content/sips/xxx-trigger-executors.md
```diff
@@ -0,0 +1,65 @@
...
+* "External" triggers use external binaries, e.g. `spin-trigger-timer`
+* Trigger executors share common code, including common CLI options in a `spin-trigger` crate
+
+### The `spin up` command changes
+
+It remains the primary entrypoint for starting a Spin app, but its behavior is changed to:
+
+* Detect the trigger type for an application
+* Find a trigger executor for that type (e.g. check the `PATH` for `"spin-trigger-${type}"`)
+* `exec` the trigger executor, forwarding its own CLI arguments
```
itowlson, line 41: What new concerns exist with two processes collaborating?  If the parent (spin up) process crashes, what happens to the trigger?  E.g. can the user determine which trigger process corresponds to a crashed spin up in order to kill it?  What is the flow when the user runs `spin up` and then hits Ctrl+C - does that kill the trigger process and return to the spin process, or does it kill the spin process and we need to forward it to the trigger process?  If the former, can the spin process tell the difference between "user killed" and "program crashed" - or do we not need to care?

itowlson (commented):
Looks good!  A couple of things we could maybe expand on but no worries with the core design.

docs/content/sips/xxx-trigger-executors.md
lann: I was thinking that `spin up` would only accept a couple of options itself, mostly passing through arguments to the trigger without attempting to validate. That said, I had considered setting an env var which could be e.g. `SPIN_VERSION=0.3` and allow the trigger framework to bail out if necessary.

itowlson: What would this mean for someone doing `spin up --help`?

docs/content/sips/xxx-trigger-executors.md
lann: I couldn't think of any safety issues as long as the prefix (`spin-trigger-`) doesnt collide with other projects, but I think it's something that a prototype would help clarify.

docs/content/sips/xxx-trigger-executors.md
lann: In Linux, a parent process dying will normally kill all children automatically. Worth review on other platforms, but I would expect to be able to implement that behavior if it isn't the default.

docs/content/sips/xxx-trigger-executors.md
lann: Great question...I suppose `spin up` and the trigger could just print their options one after the other?

itowlson: You don't have a trigger at that point though...

docs/content/sips/xxx-trigger-executors.md
lann: I looked at this briefly when I was thinking there might be a more complex multiprocess interface between `spin up` and triggers: https://doc.servo.org/ipc_channel/index.html

docs/content/sips/xxx-trigger-executors.md
lann: It could! Though it would be weird for bindles...🤔

itowlson: Could you expand on that?  Are you thinking it would look to see if there was a spin.toml in the current directory?

(This would end up requiring a custom help implementation either way, so we can do what we like...)

lann: Yes, basically:

1. Print `up`'s own help
2. Run the "Detect the trigger type for an application" step (below)
3. If found, `exec` the trigger w/ `--help`

lann: This raises another good point, that `spin up` still needs to be responsible for fetching _at least_ `spin.toml` from bindles.

itowlson: I was assuming that `spin up` would still take care of all the asset loading, but yeah, maybe some of that needs to go to the trigger because we can't pass the abstract application object to the child process, and that means each trigger needs to know how to load all possible sources.

I wonder if we are better off defining an interface between `spin up` and external triggers, similar to the abstract application object, so that external triggers don't need to carry all that logic with them (and to update with things like performance fixes).

lann: I think this is another area where a prototype would help.

lann: I added a bit more to this section.

docs/content/sips/xxx-trigger-executors.md
```diff
@@ -0,0 +1,68 @@
...
+* Trigger executors share common code, including common CLI options in a `spin-trigger` crate
+
+### The `spin up` command changes
+
+It remains the primary entrypoint for starting a Spin app, but its behavior is changed to:
+
+1. Load the application manifest
+1. Find a trigger executor for the application (e.g. check the `PATH` for `"spin-trigger-${trigger.type}"`)
+1. Set some environment variables:
+   * `SPIN_MANIFEST_URL`: a URL pointing to the application manifest (e.g. `file:///path/to/spin.toml` or `bindle+https://bindle-server?id=bindle-id`)
```
lann, line 42: The `bundle+https` URL here is just one option to support future-compatibility with other loaders.

michelleN (approved):

docs/content/sips/xxx-trigger-executors.md
itowlson: There's no equivalent on Windows either - you always end up spawning a child process.

itowlson:
This looks good.  As you say, there are areas where a prototype would help clarify things, so the best way forward is probably to get trying things!

fibonacci1729 (approved):

docs/content/sips/xxx-trigger-executors.md
radu-matei: `spin up` could handle the core components (that is without trigger configuration), I wonder if splitting that between `spin up` itself and the trigger is worth it though.

Just something to think about, definitely not blocking this SIP, and I think a prototype would help a lot here.

radu-matei: `spin up` could handle the core components (that is without trigger configuration), I wonder if splitting that between `spin up` itself and the trigger is worth it though.

Just something to think about, definitely not blocking this SIP, and I think a prototype would help a lot here.

radu-matei (approved):

docs/content/sips/xxx-trigger-executors.md
radu-matei: So far we managed to dodge a _Spin_ configuration file (not the app manifest), but we might start have enough potential things to configure that we might need one.

One of the things would be setting the paths to the executables for triggers.
(other configurations could include template defaults, Wasmtime cache configuration..)

PR #418, opened 2022-04-30
feat: trigger crate

Mossaka:
Built on top of the draft PR #415 and discussed in issue #402 

Notable difference: added a type `TriggerExtra` that encapsulate `Router` for `HttpTrigger` or `subscriptions` for `RedisTrigger`

@lann @radu-matei


lann:
@Mossaka: Reviewing this PR inspired me to make a bunch of changes I've been thinking of doing anyway, one benefit of which being that I was able to simplify some of the new trigger crate code. Take a look: https://github.com/fermyon/spin/pull/418/files/2eaa0c48c09a2f16a4211a169dd047bc2fefc7c1..4b54f9cc1883f3b1e2bff0b2c438a1ad26f1b667

crates/http/src/routes.rs
```diff
@@ -2,52 +2,35 @@
...
-use anyhow::{bail, Result};
+use anyhow::{Context, Result};
 use http::Uri;
 use indexmap::IndexMap;
-use spin_manifest::{Application, CoreComponent};
+use spin_manifest::{ComponentMap, HttpConfig};
 use std::fmt;
-use tracing::log;
-
-// TODO
```
radu-matei, line 12: Yay!

Mossaka:
@lann I really like your approach that removes the dependency on `Application`

Mossaka:
I will have an update to resolve the conflicts

crates/manifest/src/lib.rs
```diff
@@ -101,33 +112,38 @@ pub struct HttpTriggerConfiguration {
...
     pub base: String,
 }
+
 impl Default for HttpTriggerConfiguration {
     fn default() -> Self {
         Self { base: "/".into() }
     }
 }
 
+impl TryFrom<ApplicationTrigger> for HttpTriggerConfiguration {
```
itowlson, line 122: Are these able to move into the trigger-specific crates at this point, or does that come later as part of the trigger executors work?

src/commands/up.rs
```diff
@@ -154,16 +155,33 @@ impl UpCommand {
             _ => unreachable!(),
         };
 
+        let mut wasmtime_config = wasmtime::Config::default();
```
itowlson, line 158: Let's still keep this in a separate function, even if that's a smaller function than it was before - the `run` function is getting very long and I'm finding it hard to navigate.

itowlson (approved):
LGTM - delighted to see this

crates/http/src/lib.rs
```diff
@@ -59,43 +54,56 @@ pub struct HttpTrigger {
...
-            .clone();
-
-        let component_triggers = app.component_triggers.try_map_values(|id, trigger| {
-            trigger
-                .as_http()
-                .cloned()
-                .ok_or_else(|| anyhow!("Expected HTTP configuration for component {}", id))
-        })?;
+#[derive(Clone)]
+pub struct HttpRuntimeConfig {
```
fibonacci1729, line 58: Nit: Anything *but* just "Runtime"......maybe _HttpTriggerRuntimeConfig_? ......pretty please

fibonacci1729: OR *HttpTriggerConfig*

crates/manifest/src/lib.rs
lann: They could be moved but they can't really be decoupled without more config refactoring

fibonacci1729 (approved):

crates/http/src/lib.rs
fibonacci1729: Whatya think?

Mossaka: there are just too many configs...
1. `HttpTriggerConfiguration`
2. `HttpConfig`
3. and this `HttpRuntimeConfig`

How about `HttpRuntimeConfig` -> `HttpTriggerExecutionConfig`?

crates/manifest/src/lib.rs
Mossaka: I will keep them in the manifest for now

Mossaka:
I had a commit https://github.com/fermyon/spin/pull/418/commits/3822f5829dae7e1b7b372199ca7e296b38930b95 that changes some struct's names 

mostly `Runtime` -> `Execution`

crates/http/src/lib.rs
itowlson: This might be one for a subsequent blitz on naming - I don't even know the difference between the first two let along the new one - we should do a pass for figure out their different responsibilities and give them clearly distinct names that clearly express those differences, but this PR might not be the place for that, not sure.

Mossaka: Agreed. My understanding is that

1. The first one is for top level trigger struct `"trigger = { type = "http", base = "/" }`
2. The second one is for component trigger struct 
  ```
  [component.trigger]
  route = "/hello"
  ```
3. The third one is for CLI aguments: http has `--listen address`, `--log-dir` and `--tls-cert/key`

lann: I agree that we shouldn't try to name everything perfectly in this PR. The names were already pretty confusing.

lann (approved):

lann:
I am going to resolve conflicts and merge shortly.

PR #419, opened 2022-05-01
added http OPTIONS

Mossaka:
Signed-off-by: Jiaxiao Zhou <jiazho@microsoft.com>

The `SpinHttpExecutor` misses a `OPTIONS` http method implementation. This PR adds that.


examples/cloudevents-abuse-protection/src/lib.rs
```diff
@@ -0,0 +1,29 @@
+use anyhow::Result;
+use spin_sdk::{
+    http::{Request, Response},
+    http_component,
+};
+
+/// A spin component that can be used to test the event-grid-validation.
+/// For more info, see here: https://github.com/cloudevents/spec/blob/v1.0/http-webhook.md#42-validation-response
+#[http_component]
+fn validation(req: Request) -> Result<Response> {
```
itowlson, line 10: This doesn't seem to check for the `OPTIONS` verb - is that intentional?  Is the options support because options messages need to get through, but we don't actually care what the verb is?

itowlson (changes requested):
All looks good, the only reason for requesting changes is that there seems to be an unrelated dump file in the `redis-rust` directory...?

examples/cloudevents-abuse-protection/src/lib.rs
Mossaka: Good question. Actually this example will need to use the OPTIONS verb to distinguish a validation request vs. normal request. I will push an update.

Mossaka:
> rust

interesting... not sure how that went through but I will delete it.

radu-matei:
Thanks a lot for catching the missing HTTP method!

Would it make sense to split the two contributions, as they are separate?
Perhaps the cloud events example would make more sense in the kitchensink examples repo?
https://github.com/fermyon/spin-kitchensink

Mossaka:
> Thanks a lot for catching the missing HTTP method!
> 
> Would it make sense to split the two contributions, as they are separate? Perhaps the cloud events example would make more sense in the kitchensink examples repo? https://github.com/fermyon/spin-kitchensink

@radu-matei I moved the example out of this PR. Now it only fixes the http options.

lann (approved):

Issue #412, opened 2022-04-28
Add `description` to component configuration


radu-matei:
Thanks to #392, we now have the routes printed at startup:

```
Available Routes:
  http://127.0.0.1:3000 (wildcard)
  http://127.0.0.1:3000/static (wildcard)
  http://127.0.0.1:3000/favicon.ico
```

This issue proposes two things:

- also printing the component ID that corresponds to that route
- adding an optional `description` or `summary` field to the component configuration

lianghanzhen:
This seems a good first issue, maybe I can pick it up?

radu-matei:
That would be awesome, @lianghanzhen!

Printing the component IDs happen here:
https://github.com/fermyon/spin/blob/8bac94fceba53fbde259f0520a934e425374a198/crates/http/src/lib.rs#L220-L223

There are three main places where a description would have to be updated:

- in the manifest definition https://github.com/fermyon/spin/blob/8bac94fceba53fbde259f0520a934e425374a198/crates/manifest/src/lib.rs#L60-L70
- in the local loader
https://github.com/fermyon/spin/blob/8bac94fceba53fbde259f0520a934e425374a198/crates/loader/src/local/config.rs#L55-L73
- in the Bindle loader
https://github.com/fermyon/spin/blob/8bac94fceba53fbde259f0520a934e425374a198/crates/loader/src/bindle/config.rs#L19-L35

Let me know if I can help further.

Thanks!

lianghanzhen:
I'm thinking about printing out the available routes in this format:
```
component id: URL (wildcard)
```

Do we need to print out the description(if applicable) as well?

radu-matei:
Implemented in #420

PR #420, opened 2022-05-01
feat: print component id when spin up

lianghanzhen:
The available routes will be printed out in the following format:
`Component Id: URL (wildcard)`

` (wildcard)` will be printed out iff applicable.

Examples:
- hello: http://127.0.0.1:3000/hello
- static: http://127.0.0.1:3000/static (wildcard)

Signed-off-by: Hanzhen Liang <handrenliang@gmail.com>

### DIFF
--- a/crates/manifest/src/lib.rs
+++ b/crates/manifest/src/lib.rs
@@ -65,6 +65,8 @@ pub struct CoreComponent {
     /// ID of the component. Used at runtime to select between
     /// multiple components of the same application.
     pub id: String,
+    /// Description of the component(Optional).
+    pub description: Option<String>,
     /// Per-component WebAssembly configuration.
     pub wasm: WasmConfig,
 }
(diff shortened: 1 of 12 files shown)


lianghanzhen:
Fix first point of #412

crates/manifest/src/lib.rs
```diff
@@ -65,6 +65,8 @@ pub struct CoreComponent {
     /// ID of the component. Used at runtime to select between
     /// multiple components of the same application.
     pub id: String,
+    /// Description of the component(Optional).
```
itowlson, line 68: Suggest removing `optional` here - it's already implicit in the `Option` type. 
```suggestion
    /// Description of the component.
```

itowlson (approved):
LGTM - thanks!

crates/manifest/src/lib.rs
lianghanzhen: Make sense. Resolved.

radu-matei:
Could you also add a reference to the new field in the component configuration here, please?
https://github.com/fermyon/spin/blob/main/docs/content/configuration.md

Thanks!

PR #424, opened 2022-05-02
feat(cli): add Wasmtime cache

### LINKED ISSUE #169: Investigate wasmtime compilation caching
Compilation caching can speed up spin startup a _lot_, which is important if we want to scale up from zero on-demand.

Its easy to enable caching; it just needs [`cache_config_load_default`](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.cache_config_load_default) called on the Config and the default config file created with `wasmtime config new`. However, we should understand cache invalidation, disk usage, etc. and whether we want to tune the config before doing so.

radu-matei:
close #169 
ref #413 

This updates the `spin up` command to always use a Wasmtime cache.
Specifically, if no option is passed, it will attempt to use the default
Wasmtime cache, or create a new `enabled` cache configuration on the first
run.

This behavior can be overridden with the `--disable-cache=true` setting, and the
cache file can be configured with the `--cache <path-to-cache-file>` flag.

See https://github.com/bytecodealliance/wasmtime/blob/main/docs/cli-cache.md
for a reference on the cache file.

Not exactly sure how to (or if we want to) just pass `--disable-cache` instead of `--disable-cache=true`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3150,6 +3150,7 @@ dependencies = [
  "tracing-subscriber 0.3.9",
  "vergen",
  "wasi-outbound-http",
+ "wasmtime",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ tracing = { version = "0.1", features = [ "log" ] }
 tracing-futures = "0.2"
 tracing-subscriber = { version = "0.3.7", features = [ "env-filter" ] }
 wasi-outbound-http = { path = "crates/outbound-http" }
+wasmtime = "0.34"
 
 [dev-dependencies]
 hyper = { version = "0.14", features = [ "full" ] }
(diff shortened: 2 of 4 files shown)


itowlson (approved):
Nice!

As a UI, I vaguely favour `--disable-cache` over `--disable-cache=true`.  But the latter can be handy when running programmatically because you can insert a true/false value instead of needing conditional logic to provide the flag or not.

I thought that, for booleans, structopt took the flag without a value anyway by default (so this would give `--disable-cache`).  That seems to be how Clap 3 works at least.  But maybe it's different in this version!

Anyway, looks good and will be nice to have!

itowlson:
I guess it must be different in this version - I see we are explicitly giving `takes_value = false` on things like the `--insecure` flag.

lann (approved):

PR #426, opened 2022-05-02
docs(*): fix links which result in 404: Not Found on GitHub

dicej:
While perusing the excellent documentation, I noticed that all the links of the
form (/page-name) were broken in both Emacs (file not found) and on GitHub
(404).  Changing them to (./page-name.md) fixed both cases.  Additionally, one
of the HTTPS links had an extra pair of parentheses, which I've removed.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -40,7 +40,7 @@ The following are the fields supported by the `spin.toml` manifest file:
 - `trigger` (REQUIRED): Trigger for the application. Currently, the two
 implemented trigger types are:
   - `http`: All components of the application are invoked as a result of
-  incoming HTTP requests. [The HTTP trigger](/http-trigger) configuration has
+  incoming HTTP requests. [The HTTP trigger](./http-trigger.md) configuration has
   the following fields:
     - `type` (REQUIRED): The application trigger type with the value `"http"`.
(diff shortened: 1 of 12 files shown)


michelleN (approved):
Thank you!

radu-matei:
Hi, Joel!
This is because of the way we render the documentation website — this change to a relative file path fixes navigating locally (and on GitHub), but breaks the URLs from the actual website.

I'd love to find a solution that addresses both.

dicej:
Hi Radu!

Yes, I expected there would be a good reason for it -- thanks for explaining.  What are you using the render the website?  [Bartholomew](https://github.com/fermyon/bartholomew)?  If so, perhaps I could open an issue there and take a closer look.

radu-matei:
Exactly, we are using Bartholomew to render the website.
That's exactly what I was thinking — fixing this in Bartholomew would be awesome!

radu-matei:
I suspect this is where we'd have to make a change: https://github.com/fermyon/bartholomew/blob/875106be47a4362c903bc40859cc3a750fab35dd/src/content.rs#L59-L66

dicej:
Thanks, Radu -- I've opened https://github.com/fermyon/bartholomew/issues/69 based on my current understanding.  Feel free to edit the description if it's not accurate.

PR #431, opened 2022-05-03
Follow logs

### LINKED ISSUE #380: Add option to follow the component logs when running `spin up`
When running `spin up`, you see the _Spin_ logs, and the logs of components are saved in the configured logs directory.

I would like to propose optionally following the component logs in the standard output / error of the Spin process.

itowlson:
Fixes #380.

For discussion and feedback.

The current UI is:

* `spin up ... --follow comp1 --follow comp2` to follow specifically those components
* `spin up ... --follow-all` to follow all components

The current behaviour is the followed component stdout gets redirected to Spin stdout and component stderr to Spin stderr, _unless_ the executor chooses to do things differently.  (WAGI, for example, follows only stderr.)

I feel like there should be scope for simplification, and I'm not quite sure whether I'm passing the follow selection around in an appropriate way.  Also there is a clippy lint that I'm not sure how to address.  Input eagerly requested.

Note: This needs a rebase but I was thinking I'd hold off on that until I've had feedback on whether this is along the right lines - no point ploughing through a rebase if we decide to redesign it!

### DIFF
--- a/crates/engine/src/io.rs
+++ b/crates/engine/src/io.rs
@@ -20,3 +20,25 @@ pub struct OutRedirect {
     /// Lock for writing.
     pub lock: Arc<RwLock<Vec<u8>>>,
 }
+
+/// Prepares WASI pipes which redirect a component's output to
+/// memory buffers.
+pub fn prepare_io_redirects() -> anyhow::Result<IoStreamRedirects> {
+    let stdin = ReadPipe::from(vec![]);
+
+    let stdout_buf: Vec<u8> = vec![];
+    let lock = Arc::new(RwLock::new(stdout_buf));
+    let stdout = WritePipe::from_shared(lock.clone());
+    let stdout = OutRedirect { out: stdout, lock };
+
+    let stderr_buf: Vec<u8> = vec![];
+    let lock = Arc::new(RwLock::new(stderr_buf));
+    let stderr = WritePipe::from_shared(lock.clone());
+    let stderr = OutRedirect { out: stderr, lock };
+
+    Ok(IoStreamRedirects {
+        stdin,
+        stdout,
+        stderr,
+    })
+}
--- a/crates/engine/src/io.rs
+++ b/crates/engine/src/io.rs
@@ -1,44 +1,215 @@
-use std::sync::{Arc, RwLock};
-use wasi_common::pipe::{ReadPipe, WritePipe};
+use std::{
+    collections::HashSet,
+    io::{LineWriter, Write},
+    sync::{Arc, RwLock, RwLockReadGuard},
+};
+use wasi_common::{
+    pipe::{ReadPipe, WritePipe},
+    WasiFile,
+};
 
-/// Input/Output stream redirects
-#[derive(Clone)]
-pub struct IoStreamRedirects {
-    /// Standard input redirect.
-    pub stdin: ReadPipe<std::io::Cursor<Vec<u8>>>,
-    /// Standard output redirect.
-    pub stdout: OutRedirect,
-    /// Standard error redirect.
-    pub stderr: OutRedirect,
+/// Which components should have their logs followed on stdout/stderr.
+#[derive(Clone, Debug)]
+pub enum FollowComponents {
+    /// No components should have their logs followed.
+    None,
+    /// Only the specified components should have their logs followed.
+    Named(HashSet<String>),
+    /// All components should have their logs followed.
+    All,
 }
 
-/// Output redirect and lock.
-#[derive(Clone)]
-pub struct OutRedirect {
-    /// Output redirect.
-    pub out: WritePipe<Vec<u8>>,
-    /// Lock for writing.
-    pub lock: Arc<RwLock<Vec<u8>>>,
+impl FollowComponents {
+    /// Whether a given component should have its logs followed on stdout/stderr.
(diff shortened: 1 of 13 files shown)


lann:
I wonder if it would make sense to push all this logging through `tracing`, then implement "following" as a custom tracing subscriber. I'm thinking forward to other things we might want to do with these component logs like exposing them through hippo; it seems likely that `tracing`'s [ecosystem](https://docs.rs/tracing/latest/tracing/index.html#related-crates) could simplify that for us.

crates/engine/src/io.rs
```diff
@@ -1,22 +1,215 @@
...
+    None,
+    /// Only those listed.
+    Named(Vec<String>),
+    /// All of them.
+    All,
+}
+
+impl FollowComponents {
+    /// Whether a given component should have its logs followed on stdout/stderr.
+    #[allow(clippy::ptr_arg)] // Clippy recommendation makes it not compile
```
lann, line 23: The [docs suggest](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.contains) using `any`:

```rust
ids.any(|id| id == component_id)
```

itowlson:
@lann re using the tracing crate - yes, I had been wondering whether to use tracing or to dump straight to the output streams.  My initial feeling was that the goal here was specifically "follow" so you wouldn't necessarily want an alternate subscriber, you want to see it in your UI; but it's definitely an open question and I should have flagged it.  Thanks for bringing it up!

lann:
> you wouldn't necessarily want an alternate subscriber, you want to see it in your UI

I think seeing things in your UI is _perfect_ example of a subscriber!

src/commands/up.rs
```diff
@@ -105,6 +106,21 @@ pub struct UpOpts {
...
+        multiple = true,
+        )]
+    pub follow_components: Vec<String>,
+
+    /// Print all component output to stdout/stderr
+    #[structopt(
+        long = "follow-all",
+        conflicts_with = FOLLOW_LOG_OPT,
+        )]
+    pub follow_all_components: bool,
```
radu-matei, line 123: Do we want an environment variable for this?

radu-matei:
Potential follow-up.

How would we feel about prepending the component name in the followed logs?

radu-matei (approved):
This looks great, thank you!

LGTM

itowlson:
This, like Lann's tracing question, is interesting.  At the moment we avoid having to do anything with the output byte stream.  Using `tracing` or prepending component IDs would need us to parse the stream and deal with it on a line-by-line basis.  The PR as it stands already uses a `LineWriter` to buffer and emit as lines (so that concurrent output doesn't get mixed up on the same line): we could certainly see whether we can wrap that further to modify the line as it's written, or to explicitly detect things as lines and emit them to trace.

It's certainly attractive to be able to distinguish logs in this way.  But I can also see a case for not messing with the output of code that doesn't belong to us and we have no idea what it does!

This would be good to resolve before we merge this, because it might mean changes at the structural level.  We'd probably also want to run it past the commenters on #438 to make sure it wouldn't affect the way they want to do things (I think they plan to inject their own streams so it shouldn't be an issue but).

crates/engine/src/io.rs
```diff
@@ -1,22 +1,215 @@
...
+};
+use wasi_common::{
+    pipe::{ReadPipe, WritePipe},
+    WasiFile,
+};
+
+/// Which components should have their logs followed on stdout/stderr.
+#[derive(Clone, Debug)]
+pub enum FollowComponents {
+    /// None of them.
```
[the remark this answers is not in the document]
itowlson: *chuckle*  Apologies.  I was getting a bit lazy because I wasn't sure whether I was going in the right direction.  You are 100% right!

radu-matei:
Ah, good point, I will reference the comment in #438.

src/commands/up.rs
itowlson: We've tended to use environment variables for things that are "this is what my setup looks like and it will never change."  But if someone is running Spin repeatedly trying to track down an issue, then they may want to leave this on for the duration of that session; but maybe that's what shell history is for.  On the whole I worry that an EV could lead to surprises when you accidentally spewed logs during a demo or when you did `--follow` for one component and it conflicted with the presence of the EV.  But it's no effort to add one if people reckon it would be good!

radu-matei: Ah, good point, I agree that it could be confusing.
Happy to defer this conversation to a later time (say to when we have a global configuration file).

Thanks!

crates/http/src/lib.rs
```diff
@@ -391,6 +400,9 @@ pub(crate) fn compute_default_headers<'a>(
 /// All HTTP executors must implement this trait.
 #[async_trait]
 pub(crate) trait HttpExecutor: Clone + Send + Sync + 'static {
+    // TODO: allowing this lint because I want to gather feedback before
+    // investing time in reorganising this
+    #[allow(clippy::too_many_arguments)]
```
lann, line 405: We ran into this before and punted by...removing an unnecessary argument. :neutral_face:

I'm fine leaving this for now since these `HttpExecutor`s are ripe for a broader refactoring. Should probably reword the comment though.

lann (approved):
I'm fine punting on `tracing` integration for now. This PR addresses the problem well.

fibonacci1729 (approved):

PR #432, opened 2022-05-03
fix(docs): update Bartholomew with fix for relative links

radu-matei:
This commit updates the Bartholomew module used to include the fix for relative
links in markdown.
The new Bartholomew now targets the Spin executor, so `spin.toml` is updated
accordingly.

Thanks to @dicej for the Bartholomew fix!


vdice (approved):
🎉

Issue #78, opened 2022-02-14
Add Spin HTTP C++ template


radu-matei:
The templates directory should contain a template for writing C++ HTTP components.

An example on how to write a C++ module based on a WIT definition can be found here — https://github.com/fermyon/wasi-experimental-toolkit/tree/main/tests/modules/cache-cpp

dicej:
@radu-matei I could take a crack at this.  Can you clarify what you mean by "the templates directory", though?  I don't see such a directory in the spin repo.  I do see a crates/templates directory, which I assume is related but not what you're referring to here.

radu-matei:
Back when the issue was created, there used to be one..
Now we have an `examples` directory, which this would go into.

See https://github.com/fermyon/wasi-experimental-toolkit/tree/main/tests/modules/cache-cpp for an example of how this would look like.

radu-matei:
Added in #433

PR #433, opened 2022-05-03
feat: add C++ spin-http example

dicej:
This is a simple "hello, world" example analogous to http-rust.  Since it relies
on the generated C bindings, it isn't idiomatic C++ -- just C with a smattering
of C++ syntax.  An idiomatic C++ example would require native C++ bindings which
use e.g. std::string, std::tuple, std::optional, etc.

Issue: https://github.com/fermyon/spin/issues/78

### DIFF
--- a/examples/http-cpp/README.md
+++ b/examples/http-cpp/README.md
@@ -1,0 +1,39 @@
+## Spin HTTP Example in C++
+
+This is a simple example of a [Spin HTTP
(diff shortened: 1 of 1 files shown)


radu-matei (commented):
Could you please write a short readme documenting the commands required to build and run the application?

Thank you for your contribution!

lann (approved):
Looks good. Thanks!

radu-matei (approved):
Thank you for your contribution!

LGTM

PR #435, opened 2022-05-03
fix(http/router): update router to contain an ordered component map

### LINKED ISSUE #434: HTTP router is now non-deterministic
ref #418 
cc @vdice, @Mossaka 

Consider the following two component Spin application:

```toml
[[component]]
source = "modules/spin_bartholomew.wasm"
id = "bartholomew"
files = [ "content/**/*" , "templates/*", "scripts/*", "config/*"]
[component.trigger]
route = "/..."

[[component]]
source = "modules/spin_static_fs.wasm"
id = "fileserver"
files = [ { source = "static/", destination = "/" } ]
[component.trigger]
route = "/static/..."
```

For overlapping HTTP routes (like in the example above), the HTTP router relies on the order in which they have been defined in `spin.toml`.

Running `spin up` multiple times results in the routes order being non-deterministic:

```
run 1:   spin_http_engine: Constructed router for application fermyon.com: {Exact("/favicon.ico"): "favicon", Wildcard(""): "bartholomew", Wildcard("/static"): "fileserver"}    

run 2:  spin_http_engine: Constructed router for application fermyon.com: {Wildcard("/static"): "fileserver", Wildcard(""): "bartholomew", Exact("/favicon.ico"): "favicon"}    
``` 

The implementation relied on the router being constructed from a `Vec<CoreComponent>` (which had the component order preserved), which changed to a `ComponentMap` (backed by a `HashMap`):

https://github.com/fermyon/spin/pull/418/files#diff-d7033a1fe9a4798b8b561185baca64b8b4fddb02552a9edf0c6ae1162638d6b6R23

radu-matei:
fix #434 
cc @vdice, @Mossaka 

This commit updates `ComponentMap` to be backed by an `IndexMap` as opposed to
a `HashMap` to preserve the deserialization order from `spin.toml`, which is important
when constructing the HTTP router.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3276,6 +3276,7 @@ name = "spin-manifest"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "indexmap",
  "serde",
  "spin-config",
  "thiserror",
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -34,16 +34,13 @@ impl Router {
         Ok(Self { routes })
     }
 
-    // This assumes the order of the components in the app configuration vector
-    // has been preserved, so the routing algorithm, which takes the order into
-    // account, is correct. This seems to be the case with the TOML deserializer,
-    // but might not hold if the application configuration is deserialized in
-    // other ways.
-
+    // This assumes the order of the components in the manifest has been
+    // preserved, so the routing algorithm, which takes the order into account,
+    // is correct.
     /// Returns the component ID that should handle the given path, or an error
     /// if no component matches.
     /// If there are multiple possible components registered for the same route or
-    /// wildcard, this returns the last one in the components vector.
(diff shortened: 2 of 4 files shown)


fibonacci1729 (approved):

lann:
Dang, I think that was my fault.

Mossaka:
Will the use of indexmap degrade the performance, especially if there are plenty of routes? Maybe we can just sort the routes in spin cli when we show routes to the user

radu-matei:
We already have an `IndexMap` in the router, and I doubt our using of an `IndexMap` will be the bottleneck, really looking forward to when that will be the case though!

The main point is not sorting them, but quite the opposite — preserving the order defined in `spin.toml`.
(see #434 for a bit of context around that).

Mossaka:
ah that makes sense! thanks

Issue #436, opened 2022-05-04
Feature: allow user to bring it's own wasmtime Engine


Mossaka:
`spin_engine::Builder` currently builds a wasmtime engine by the user. Would like to have a function to bring the engine. 

## Proposal API:

<s>removed
```rust
let exec_ctx = spin_engine::Builder::new()
  .engine(engine.clone()) // note: engine.clone() is a shadow copy
  .link_defaults()?
  ...
  .build()
  .await?
```
</s>

<s>spin could add necessary engine configs in `build()` function
```rust
let mut config = self.engine.config();
config
  .wasm_multi_memory(true)
  .wasm_module_linking(true)
```
</s>

```rust
let engine = wasmtime::Engine::default();
let builder = spin_engine::Builder::with_wasmtime_engine(config, engine.clone());
builder
  .link_defaults()?
  .add_component(...)
  .build()
  .await?
```

## Why do we want this?
Some application might have already built a wasmtime engine, and shared across different threads. They might just want to pass a reference of the engine to spin to run multiple triggers, instead of having to create a new wasmtime engine for every trigger.

lann:
I don't think this approach works over the long term. The `ExecutionContext` depends on certain `wasmtime::Config` settings, and I expect those settings to change as the `ExecutionContext` behavior changes.

Could you give more information on your requirements? There may be another design that allows `Engine` reuse.

lann:
@Mossaka It looks like you are only wanting to share a `wasmtime::Engine` between multiple `ExecutionContext`s, is that right? If so, I think we can get what you are looking for by extracting just the wasmtime `Config` and `Engine` construction from `Builder` into a new struct. I will take a closer look later today.

Mossaka:
But `wasmtime::Engine` depends on `wasmtime::Config`, so it's unavoidable when I bring an engine, the configs are already inside of the engine. Could you please elaborate more on what you mean by extracting just the Config and Engine from `Builder`?

lann:
Something like (untested):

`engine/src/lib.rs`
```rust
#[derive(Clone)]
pub struct Engine(wasmtime::Engine);

impl Engine {
  pub fn new(config: wasmtime::Config) -> Result<Self> {
    config.wasm_multi_memory(true);
    config.wasm_module_linking(true);
    Ok(Self(wasmtime::Engine::new(&config)?))
  }
}

impl Builder {
  ...
[...]

PR #437, opened 2022-05-04
add with_wasmtime_engine() to Builder

Mossaka:
1. added a getter for wasmtime::Engine
2. extract wasmtime config and engine to a struct 
3. moved `data` and `store` to `build()`

Related to issue #436 

Signed-off-by: Jiaxiao Zhou <jiazho@microsoft.com>


lann (changes requested):
Pending discussion: https://github.com/fermyon/spin/issues/436#issuecomment-1117304647

Mossaka:
Made an update to add the Engine struct and a getter function @lann

crates/engine/src/lib.rs
```diff
@@ -61,43 +61,55 @@ pub struct RuntimeContext<T> {
...
+        // See https://github.com/bytecodealliance/wit-bindgen/blob/main/crates/wasmlink.
+        config.wasm_multi_memory(true);
+        config.wasm_module_linking(true);
+        Ok(Self(wasmtime::Engine::new(&config)?))
+    }
+
+    /// Get a clone of the internal `wasmtime::Engine`.
+    /// WARNING: The configuration of this Engine is likely to change in the future, and
+    /// will not be covered by any future stability guarantees.
+    pub fn inner(&self) -> wasmtime::Engine {
```
lann, line 82: It doesn't matter much for a `wasmtime::Engine`, but in _general_ you would return a reference here and let the caller `.clone`

lann (commented):
Woops, I just realized I had pending comments I never submitted, sorry!

crates/engine/src/lib.rs
```diff
@@ -61,43 +61,55 @@ pub struct RuntimeContext<T> {
...
+        self.0.clone()
+    }
+}
+
 /// An execution context builder.
 pub struct Builder<T: Default> {
     config: ExecutionContextConfiguration,
     linker: Linker<RuntimeContext<T>>,
-    store: Store<RuntimeContext<T>>,
```
lann, line 68: Per discussion with @radu-matei, we'd like to keep this on the Builder for now. Would you mind reverting this part of the change?

lann (changes requested):

crates/engine/src/lib.rs
Mossaka: good tip!

crates/engine/src/lib.rs
```diff
@@ -135,7 +151,7 @@ impl<T: Default + 'static> Builder<T> {
             let core = c.clone();
             let module = match c.source.clone() {
                 ModuleSource::FileReference(p) => {
-                    let module = Module::from_file(&self.engine, &p).with_context(|| {
+                    let module = Module::from_file(&self.engine.0, &p).with_context(|| {
```
lann, line 154: Sorry for the noise. I reverted my own feedback to use `.inner` here; doesn't make as much sense with `inner` doing that `clone`.

lann (approved):

Mossaka:
let's merge this in?

PR #439, opened 2022-05-04
doc: fix instructions in distributing-apps.md

dicej:
`bindle-server` v0.8 requires an `--unauthenticated` option in order to run
without authentication.  Also, the port used for `BINDLE_URL` didn't match what
was specified in the `bindle-server` command.


radu-matei (approved):
Good catch, LGTM.

PR #440, opened 2022-05-04
perf: reduce peak memory usage and open files when loading bindle

dicej:
This patch does two things:

1. Use `bindle::client::Client::get_parcel_stream` instead of `get_parcel`.  The
latter loads the whole asset into memory, while the former allows us to stream
into a local file chunk-by-chunk.

2. Limit parallel I/O in `spin_loader::bindle::assets::Copier` to avoid
hammering the Bindle server and also avoid running out of file descriptors.

This addresses https://github.com/fermyon/spin/issues/413.

### DIFF
--- a/crates/loader/src/bindle/assets.rs
+++ b/crates/loader/src/bindle/assets.rs
@@ -6,11 +6,14 @@ use crate::{
 };
 use anyhow::{anyhow, bail, Context, Result};
 use bindle::{Id, Label};
-use futures::future;
+use futures::{future, stream, StreamExt, TryStreamExt};
 use spin_manifest::DirectoryMount;
 use std::path::Path;
-use tokio::fs;
+use tokio::{fs, io::AsyncWriteExt};
 use tracing::log;
+
+/// Maximum number of assets to download in parallel
+const MAX_PARALLEL_COPIES: usize = 16;
 
 pub(crate) async fn prepare_component(
     reader: &BindleReader,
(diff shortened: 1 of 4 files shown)


dicej:
This reduces peak memory usage from ~200MB to ~12MB for [this test case](https://github.com/dicej/spin-test).

crates/loader/src/bindle/assets.rs
```diff
@@ -76,18 +79,29 @@ impl Copier {
...
-        let buf =
-            self.reader.get_parcel(&p.sha256).await.with_context(|| {
-                anyhow!("Failed to fetch asset parcel '{}@{}'", self.id, p.sha256)
+        let mut stream = self
+            .reader
+            .get_parcel_stream(&p.sha256)
+            .await
+            .with_context(|| anyhow!("Failed to fetch asset parcel '{}@{}'", self.id, p.sha256))?;
+
+        let mut file = fs::File::create(&to).await?;
```
itowlson, line 88: Could we add a `with_context(...)` here so that the user can see where the failure was and which file/parcel it occurred on?

itowlson (approved):
This looks great, and I learned a ton.  Thank you!  (The one suggestion is minor.)

crates/loader/src/bindle/assets.rs
dicej: Good catch -- done.

Clippy also caught a bug, so I fixed that, too.  Note to self: `for x in stream.try_next().await? { .. }` almost certainly does not do what you want it to do.

crates/loader/src/bindle/assets.rs
```diff
@@ -52,12 +55,12 @@ impl Copier {
     }
 
     async fn copy_all(&self, parcels: &[Label], dir: impl AsRef<Path>) -> Result<()> {
-        match future::join_all(parcels.iter().map(|p| self.copy(p, &dir)))
-            .await
-            .into_iter()
-            .filter_map(|r| r.err())
+        match stream::iter(parcels.iter().map(|p| self.copy(p, &dir)))
+            .buffer_unordered(MAX_PARALLEL_COPIES)
```
lann, line 59: @adamreese I _knew_ this was implemented _somewhere_!

dicej: [StackOverflow FTW](https://stackoverflow.com/questions/43247212/join-futures-with-limited-concurrency)

lann (approved):
This is great! Thank you!

radu-matei (approved):
This is great, thank you so much!
Manually tested and profiled again, this is in line with what we are seeing when running Spin with local assets.

Thank you again, LGTM.

PR #442, opened 2022-05-05
Fix Go template to get SDK from Git

itowlson:
In order to get the template stuff landed, I compromised on the Go template.  This is slightly less compromised in that it uses a hardwired commit instead of a tag or anything, but it does build and run without needing the Spin repo locally or any further intervention from the user.


radu-matei (approved):

adamreese:
I fat fingered a merge @itowlson.  LGTM anyways

PR #444, opened 2022-05-05
Trigger executors

lann:
ref #417

### DIFF
--- a/crates/loader/src/bindle/assets.rs
+++ b/crates/loader/src/bindle/assets.rs
@@ -1,16 +1,20 @@
 #![deny(missing_docs)]
 
-use crate::{
-    assets::{create_dir, ensure_under},
-    bindle::utils::BindleReader,
-};
+use std::path::Path;
+use std::path::PathBuf;
+
 use anyhow::{anyhow, bail, Context, Result};
 use bindle::{Id, Label};
 use futures::{future, stream, StreamExt, TryStreamExt};
 use spin_manifest::DirectoryMount;
-use std::path::Path;
 use tokio::{fs, io::AsyncWriteExt};
 use tracing::log;
+
+use crate::file_sha256_digest_string;
+use crate::{
+    assets::{create_dir, ensure_under},
+    bindle::utils::BindleReader,
+};
 
 pub(crate) async fn prepare_component(
     reader: &BindleReader,
@@ -69,6 +73,14 @@ impl Copier {
 
         ensure_under(&dir, &to)?;
 
+        if to.exists() {
+            match check_existing_file(to.clone(), p).await {
+                Ok(true) => return Ok(()),
+                Ok(false) => (),
+                Err(err) => tracing::error!("Error verifying existing parcel: {}", err),
+            }
+        }
+
         log::trace!(
             "Copying asset file '{}@{}' -> '{}'",
             self.id,
@@ -107,4 +119,10 @@ impl Copier {
 
         Ok(())
     }
+}
+
+async fn check_existing_file(path: PathBuf, label: &Label) -> Result<bool> {
+    let sha256_digest =
+        tokio::task::spawn_blocking(move || file_sha256_digest_string(path)).await??;
+    Ok(sha256_digest == label.sha256)
 }
--- a/src/bin/spin.rs
+++ b/src/bin/spin.rs
@@ -1,10 +1,11 @@
 use anyhow::Error;
-use clap::Parser;
+use clap::{Parser, Subcommand};
 use lazy_static::lazy_static;
 
 use spin_cli::commands::{
     bindle::BindleCommands, build::BuildCommand, deploy::DeployCommand, new::NewCommand,
-    templates::TemplateCommands, up::UpCommand,
+    templates::TemplateCommands, trigger_http::TriggerHttpCommand,
+    trigger_redis::TriggerRedisCommand, up::UpCommand,
 };
 
 #[tokio::main]
@@ -42,6 +43,14 @@ enum SpinApp {
     Bindle(BindleCommands),
     Deploy(DeployCommand),
     Build(BuildCommand),
+    #[clap(subcommand, hide = true)]
+    Trigger(TriggerCommands),
+}
+
+#[derive(Subcommand, Debug)]
+enum TriggerCommands {
+    Http(TriggerHttpCommand),
+    Redis(TriggerRedisCommand),
 }
 
 impl SpinApp {
@@ -54,6 +63,8 @@ impl SpinApp {
             Self::Bindle(cmd) => cmd.run().await,
             Self::Deploy(cmd) => cmd.run().await,
             Self::Build(cmd) => cmd.run().await,
+            Self::Trigger(TriggerCommands::Http(cmd)) => cmd.run().await,
+            Self::Trigger(TriggerCommands::Redis(cmd)) => cmd.run().await,
         }
     }
 }
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -1,22 +1,18 @@
-use std::path::PathBuf;
+use std::{ffi::OsString, path::PathBuf};
 
 use anyhow::Result;
 use clap::Parser;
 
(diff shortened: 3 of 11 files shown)


crates/loader/src/bindle/assets.rs
```diff
@@ -69,6 +76,11 @@ impl Copier {
 
         ensure_under(&dir, &to)?;
 
+        // Validate existing file if it exists
+        if check_existing_file(&to, p).await? {
+            return Ok(());
+        }
```
lann, line 82: Hopefully this can be removed/replaced with #486

src/commands/up.rs
```diff
@@ -40,96 +41,42 @@ pub struct UpCommand {
...
+    /// All other args, to be passed through to the trigger
+    pub trigger_args: Vec<OsString>,
 }
 
 impl UpCommand {
     pub async fn run(self) -> Result<()> {
-        let working_dir_holder = match &self.opts.tmp {
+        let help = self.help;
+        let res = self.run_inner().await;
+        if help && res.is_err() {
```
fibonacci1729, line 65: Per your comment: `The help/usage stuff is a little weird but I'm not really sure what the ideal behavior should be. Currently in the PR if you spin up --help, it will attempt to load a manifest and call e.g. spin trigger http --help, but if that fails you'll just get the non-trigger-specific spin up --help usage`

This is tricky. Would it make sense to print the failed command before the general `help`?

fibonacci1729:
Manually tested. LGTM.

src/commands/up.rs
lann: _I'm not sure._ On one hand, its confusing that `--help` shows different things depending on `spin.toml`. On the other, you don't normally expect to ever see errors with `--help`.

lann: Updated to log that error as a warning.

crates/loader/src/bindle/assets.rs
```diff
@@ -108,3 +120,24 @@ impl Copier {
         Ok(())
     }
 }
+
+async fn check_existing_file(path: &PathBuf, label: &Label) -> Result<bool> {
+    if let Ok(mut existing) = fs::File::open(path).await {
+        let mut buf = [0; 0xffff];
+        let mut hasher = Sha256::new();
+        let hex_sum = loop {
```
itowlson, line 128: The publish crate already has a function for the hash string (`file_digest_string`).  We probably don't want this crate to depend on `publish` but I'd suggest we copy it rather than have a separate inline implementation.  (I'm happy if you want to replace the implementation with this one, as long as we keep the same implementation in both copies.)

src/bin/spin.rs
```diff
@@ -54,6 +63,8 @@ impl SpinApp {
             Self::Bindle(cmd) => cmd.run().await,
             Self::Deploy(cmd) => cmd.run().await,
             Self::Build(cmd) => cmd.run().await,
+            Self::Trigger(TriggerCommands::Http(cmd)) => cmd.run().await,
```
itowlson, line 66: How will this extend to external triggers or is that out of scope for now?

src/commands/build.rs
```diff
@@ -28,6 +24,9 @@ pub struct BuildCommand {
     /// Run the application after building.
     #[clap(name = BUILD_UP_OPT, short = 'u', long = "up")]
     pub up: bool,
+
+    #[clap(requires = BUILD_UP_OPT)]
+    pub up_args: Vec<OsString>,
```
itowlson, line 29: Is this intended to be visible to users running `spin build --help`, or only used programmatically?  I'm not quite sure how it fits in, sorry.

src/commands/trigger.rs
```diff
@@ -0,0 +1,119 @@
...
+            let followed = self.follow_components.clone().into_iter().collect();
+            FollowComponents::Named(followed)
+        }
+    }
+
+    pub async fn app_from_env(&self) -> Result<Application> {
+        let working_dir = std::env::var("SPIN_WORKING_DIR").context("SPIN_WORKING_DIR")?;
+        let manifest_url = std::env::var("SPIN_MANIFEST_URL").context("SPIN_MANIFEST_URL")?;
+
+        let mut app = if let Some(manifest_file) = manifest_url.strip_prefix("file://") {
```
itowlson, line 89: It might be useful to document the permitted/expected formats - I was finding this a bit hard to read.  Also possibly worth pulling it out into some sort of "parse and load app" function?

src/commands/trigger_http.rs
```diff
@@ -0,0 +1,73 @@
...
+            name = TLS_KEY_FILE_OPT,
+            long = "tls-key",
+            env = TLS_KEY_ENV_VAR,
+            requires = TLS_CERT_FILE_OPT,
+        )]
+    pub tls_key: Option<PathBuf>,
+}
+
+impl TriggerHttpCommand {
+    pub async fn run(&self) -> Result<()> {
```
itowlson, line 42: I like this - much simpler than when one command was doing both triggers.

src/commands/up.rs
```diff
@@ -143,85 +93,48 @@ impl UpCommand {
...
+        let mut cmd = std::process::Command::new(std::env::current_exe().unwrap());
+        cmd.arg("trigger")
+            .env("SPIN_WORKING_DIR", working_dir)
+            .env("SPIN_MANIFEST_URL", manifest_url)
+            .env("SPIN_TRIGGER_TYPE", trigger_type)
+            .arg(trigger_type)
+            .args(trigger_args);
+
+        if let Some(bindle_server) = self.server {
+            cmd.env(BINDLE_URL_ENV, bindle_server);
```
itowlson, line 128: If you are passing it this way, do you need the server in the URL passed to the trigger executor?  Which one should a trigger treat as authoritative?

src/commands/build.rs
radu-matei: This change was introduced by #352 — basically, `spin build --up` needs the same options as `spin up`.

src/commands/up.rs
radu-matei: It feels like we would need the information in both?

crates/loader/src/bindle/assets.rs
```diff
@@ -1,17 +1,24 @@
...
+    fs,
+    io::{AsyncReadExt, AsyncWriteExt},
+};
 use tracing::log;
 
+use crate::{
+    assets::{create_dir, ensure_under},
+    bindle::utils::BindleReader,
+};
+
```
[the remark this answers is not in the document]
radu-matei: I will do a rebase to #460 _at some point_ 😅

radu-matei (approved):
This is AWESOME, thanks!

LGTM

crates/loader/src/bindle/assets.rs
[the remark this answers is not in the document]
lann: I actually think it might be less painful to do it incrementally, at least until the option stabilizes.

src/bin/spin.rs
lann: Out of scope for this PR. We may want to support `spin trigger <custom-type>` at some point but I don't think it is necessary immediately; `spin up` can resolve those custom types directly.

src/commands/build.rs
lann: Yes, its just a different implementation of the existing behavior that made the split in `spin up` easier.

src/commands/trigger.rs
lann: I'd like to discuss whether this is even a good idea. It might make more sense to reuse/forward loader args from `spin up`, its just very awkward without more refactoring (which I'm working on).

Longer-term we should think about whether/how we want to split up the "application resolution" (fetching all modules, assets) and "execution" phases of `spin up`, which starts getting into production deployment questions.

src/commands/up.rs
lann: This turns the clap field (which can be passed as an arg or env var) into always an env var. As above, I'm not sure this is a good long term solution, but its awkward to deduplicate without more refactoring.

lann:
I'm going to merge this one since its approved, but I already have a major refactor of this along with the trigger crate: #513

PR #445, opened 2022-05-05
test(sdk/go): Add integration tests for go sdk

adamreese:
- Adds integration test for Go SDK.
- Builds TinyGo examples

CI will only run these tests on ubuntu because it's platform agnostic.


radu-matei:
Could we add a single Rust integration test in `integration.rs` that runs this `make` target here Unix (non-Windows) on `cargo test`, please?

lann:
Now the linux CI job is _almost_ as slow as the windows job! :sweat_smile:

This looks good. AFAICT these tests don't depend on the rest of the workflow (i.e. they don't use the built `spin` binary); could they be moved to a separate job so it can run in parallel? This would also let it run [only on changes to `sdk/go`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths) if desired.

lann (approved):
Given that this doesn't seem to actually make the full CI process slower (because Windows), I don't think a separate job is a blocker if its too much to deal with now.

PR #446, opened 2022-05-05
Switch from `structopt` to `clap`

lann:


src/bin/spin.rs
```diff
@@ -27,18 +28,17 @@ fn version() -> &'static str {
 }
 
 /// The Spin CLI
-#[derive(Debug, StructOpt)]
-#[structopt(
+#[derive(Parser, Debug)]
+#[clap(
     name = "spin",
     version = version(),
```
fibonacci1729, line 34: my favorites you can have if you like them: 

```
#[clap(author, version, about, long_about = None)]
#[clap(propagate_version = true)]
```

fibonacci1729 (approved):
SUPERB. LGTM

src/bin/spin.rs
lann: I'm just shooting for a straight upgrade in this PR.

fibonacci1729: SHIPIT

Mossaka:
Not intended to review the code, but I have a question: what's the motivation to change `structopt` to `clap`. I know little about them, and they seem to have the effects to me.

lann:
> Not intended to review the code, but I have a question: what's the motivation to change `structopt` to `clap`. I know little about them, and they seem to have the effects to me.

It is essentially just an upgrade. The `structopt` crate was deprecated and its functionality was merged into `clap` v3. The specific reason to upgrade now is that `clap` has a new feature that will help with https://github.com/fermyon/spin/pull/444

itowlson (approved):
Good times

michelleN (approved):

src/commands/build.rs
```diff
@@ -1,33 +1,32 @@
 use std::path::PathBuf;
 
+use anyhow::Result;
+use clap::Parser;
+
```
radu-matei, line 5: Official request for bikeshedding (that should be ignored in this PR).

Is there a convention on order imports that we want to follow?

radu-matei (approved):

PR #448, opened 2022-05-06
Add update flag to `spin templates install`

itowlson:
This adds a new flag `spin templates install (--git|--dir} --update`, which updates existing templates from the source, instead of skipping them.

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -123,7 +123,8 @@ impl TemplateManager {
         reporter.report(&message);
 
         let dest_dir = self.store.get_directory(&id);
-        if dest_dir.exists() {
+
+        let template = if dest_dir.exists() {
             match options.exists_behaviour {
                 ExistsBehaviour::Skip => {
                     return Ok(InstallationResult::Skipped(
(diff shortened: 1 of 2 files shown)


crates/templates/src/manager.rs
```diff
@@ -91,10 +124,23 @@ impl TemplateManager {
...
-                SkippedReason::AlreadyExists,
-            ));
+            match options.exists_behaviour {
+                ExistsBehaviour::Skip => {
+                    return Ok(InstallationResult::Skipped(
+                        id.to_owned(),
+                        SkippedReason::AlreadyExists,
+                    ))
+                }
+                ExistsBehaviour::Update => tokio::fs::remove_dir_all(&dest_dir)
```
lann, line 134: This is somewhat unpleasant if the install ends up failing...

itowlson: Yeah, by this point everything is ready locally so it's just a local file copy which _should_ be reasonably safe... but you're right, we should try to make failure non-impactful.  I'll take a look.

itowlson: I've pushed something that reduces the window of error but I'm not sure if there's a better way to offer the guarantee.  One possibility is to indirect through an index file: then we can simply install the new version to a completely fresh directory and update the index (and then garbage collect any unreferenced directories).  There may be value in an index file anyway (e.g. to simplify searching) - let me know if you think this is the way to go, or if you have other suggestions.  Thanks!

lann: I was going to suggest printing a description of what went wrong to allow manual cleanup, but this works too!

lann (approved):

crates/templates/src/manager.rs
```diff
@@ -166,6 +181,127 @@ impl TemplateManager {
     }
 }
 
+async fn copy_template_over_existing(
+    id: &str,
+    source_dir: &Path,
+    dest_dir: &Path,
+) -> Result<Template, anyhow::Error> {
```
radu-matei, line 188: nit: `use anyhow::Result`

itowlson: darn dirty extract method refactoring

PR #450, opened 2022-05-06
fix(vscode): add templates directory to gopls ignore list

radu-matei:
This commit adds the templates directory to the list of directories that should
be ignored by the Go language server.
This is done in order to avoid errors from the templates filling the error pane
every time the root Spin directory is opened in VS Code.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,7 +8,8 @@
     "gopls": {
         "build.directoryFilters": [
             "-sdk",
-            "-examples"
+            "-examples",
+            "-templates"
         ],
     },
     "rust-analyzer.files.excludeDirs": [


lann (approved):

.vscode/settings.json
```diff
@@ -8,7 +8,8 @@
     "gopls": {
         "build.directoryFilters": [
             "-sdk",
-            "-examples"
+            "-examples",
+            "-templates",
```
lann, line 12: Err...I assume VSCode is OK with the trailing comma here?

radu-matei: Normally it should be fine, but removed it just in case, thanks!

itowlson (approved):
Oops, thanks for the catch and sorry for missing this!

PR #456, opened 2022-05-09
Allow parcel ref in component source

### LINKED ISSUE #454: Implement Bindle module source in application manifest
The local configuration allows defining a remote reference to a bindle as the module source:

https://github.com/fermyon/spin/blob/a09fb1f2b6f9a52dee7dbcbe54e6f9ff41a72ea0/crates/loader/src/local/config.rs#L123-L131

However, that is currently not implemented:

https://github.com/fermyon/spin/blob/a09fb1f2b6f9a52dee7dbcbe54e6f9ff41a72ea0/crates/loader/src/local/mod.rs#L142-L144

It would be extremely helpful to have this implemented in order to more easily reuse modules across applications.

itowlson:
Fixes #454.

This duplicates the `BindleConnectioninfo` infrastructure from the Hippo CLI and the `bindle push` command.  It looks like the `publish` crate depends on `loader` so we can probably remove it from `publish`, though it feels a bit weird for `loader` to be the one to export it - it's not the obvious place a consumer would look for it.  We should discuss where we want things like this to live.

It may be that there's a nicer way of threading the connection info through the loader - very much open to suggestions here.

~TO DO~ NOW DONE:
* Write some tests
* Beef up the error handling

### DIFF
--- a/crates/loader/src/bindle/connection.rs
+++ b/crates/loader/src/bindle/connection.rs
@@ -1,0 +1,66 @@
+use std::sync::Arc;
+
+use bindle::client::{
+    tokens::{HttpBasic, NoToken, TokenManager},
+    Client, ClientBuilder,
+};
+
+/// BindleConnectionInfo holds the details of a connection to a
+/// Bindle server, including url, insecure configuration and an
+/// auth token manager
+#[derive(Clone)]
+pub struct BindleConnectionInfo {
+    base_url: String,
+    allow_insecure: bool,
+    token_manager: AnyAuth,
+}
+
+impl BindleConnectionInfo {
+    /// Generates a new BindleConnectionInfo instance using the provided
+    /// base_url, allow_insecure setting and optional username and password
+    /// for basic http auth
+    pub fn new<I: Into<String>>(
+        base_url: I,
+        allow_insecure: bool,
+        username: Option<String>,
+        password: Option<String>,
+    ) -> Self {
+        let token_manager: Box<dyn TokenManager + Send + Sync> = match (username, password) {
+            (Some(u), Some(p)) => Box::new(HttpBasic::new(&u, &p)),
+            _ => Box::new(NoToken::default()),
+        };
+
+        Self {
+            base_url: base_url.into(),
+            allow_insecure,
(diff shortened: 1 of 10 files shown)


lann:
In the longer term - if we wanted loaders and publishers to be pluggable - all the bindle code could move into a `spin-bindle` crate that provides both of those implementations.

crates/loader/src/bindle/connection.rs
```diff
@@ -0,0 +1,66 @@
...
+
+impl BindleConnectionInfo {
+    /// Generates a new BindleConnectionInfo instance using the provided
+    /// base_url, allow_insecure setting and optional username and password
+    /// for basic http auth
+    pub fn new<I: Into<String>>(
+        base_url: I,
+        allow_insecure: bool,
+        username: Option<String>,
+        password: Option<String>,
```
lann, line 26: How about this, to allow easier extension to other auth types:
```rust
pub fn basic_auth(&mut self, username: String, password: String) ...
```

lann: I see this is copied from `publish`, so this doesn't need to be in scope for this PR.

crates/loader/src/bindle/connection.rs
```diff
@@ -0,0 +1,66 @@
...
+            token_manager: AnyAuth {
+                token_manager: Arc::new(token_manager),
+            },
+        }
+    }
+
+    /// Returns a client based on this instance's configuration
+    pub fn client(&self) -> bindle::client::Result<Client<AnyAuth>> {
+        let builder = ClientBuilder::default()
+            .http2_prior_knowledge(false)
```
lann, line 45: This appears to be the default...

itowlson: I think when this code was written it was not the default (in fact I think I had to add the option so I could set it to false), so this is historical I guess... but unless the default is contractual (that is, it would be documented as a breaking change if they changed it), I'm inclined to set this explicitly.  (Ultimately it should be under config control.)

lann (approved):

PR #459, opened 2022-05-09
Check if bindle already exists before pushing

### LINKED ISSUE #458: `spin bindle push` does not error if invoice exists
Because of https://github.com/deislabs/bindle/issues/335.

We can probably work around this for now by checking explicitly if the invoice exists.  🤞

itowlson:
Fixes #458.


radu-matei (approved):
Thanks, tested, and it now works as expected.

LGTM

PR #460, opened 2022-05-10
chore(rustfmt): standardize the order of Rust imports

radu-matei:
This commit updates the order of the Rust imports throughtout the project to be
consistent with — specifically, standard library, external, and crate imports.

The `group_imports=StdExternalCrate` rustfmt setting only works for the nightly
toolchain, so this commit does not enforce it in CI (which would also require
all contributors to run the nightly toolchain to format).

To run the formatter using this setting:

`cargo +nightly fmt --all -- --emit files --config group_imports=StdExternalCrate`

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -2,10 +2,11 @@
 
 //! A library for building Spin components.
 
+use std::path::Path;
+
 use anyhow::{bail, Context, Result};
 use path_absolutize::Absolutize;
 use spin_loader::local::config::{RawAppManifest, RawComponentManifest};
-use std::path::Path;
 use subprocess::{Exec, Redirection};
 use tracing::log;
 
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -7,9 +7,8 @@ mod tree;
 use std::fmt::Debug;
 
 pub use provider::Provider;
-pub use tree::{Tree, TreePath};
-
 use template::{Part, Template};
+pub use tree::{Tree, TreePath};
 
 /// A config resolution error.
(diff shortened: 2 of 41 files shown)


itowlson:
What is the expectation here for maintenance, given no CI and the nightly toolchain requirement?  Is this something that we will do as and when the spirit moves us, or is it on contributors to keep it straight day by day?

radu-matei:
As you mentioned, because there is not way to enforce this without forcing contributors to use the nightly toolchain, I don't think it is reasonable to expect CI to break either, and given that we already have strict formatting rules for everything else, I would be ok with doing this _as and when the spirit moves us_.

We could add a Make target for running the nightly command above, if you think that helps.

itowlson:
Nah, "as and when" was the arrangement I assumed and I'm happy with it.  Thanks!

itowlson (approved):

Mossaka:
I am a bit concenred that if we do not enforce this to CI and all developers, pr like this will generate lots of conflicts for local developing branches.

radu-matei:
Yeah, that can definitely happen.
But that is the case with any large PR.

My hope is that when such PRs generate conflicts, you could run the the local formatting command:

```
$ cargo +nightly fmt --all -- --emit files --config group_imports=StdExternalCrate
```

But I get your concern here, not sure what the best way to address this.

lann (approved):

lann:
We can also do this incrementally as we are doing "major work" on individual crates/files. Until the config option is stable its going to drift over time anyway.

radu-matei:
Closing this for now.
Once the formatter is stable, we could revisit this.

PR #461, opened 2022-05-10
Doc comments for templates crate

### LINKED ISSUE #447: Doc comments for templates crate
During development I turned off docs for the templates crate because I was experimenting with designs and didn't want to write docs for something I would fling away in disgust 10 minutes later.

Unfortunately, I merged it without reinstating the docs.  So I now need to turn on the error and write the docs.

itowlson:
Fixes #447.

For some reason the compiler was not identifying some types as part of the public interface and needing comments.  I'm not quite sure what was going on there and it would be good to understand.  In the meantime, I've hopefully found everything!

Writing this does confirm that some of the names are not right, but it's also plausible that the API surface isn't quite right yet, so I've deferred any renaming to tackle in a future iteration.


crates/templates/src/run.rs
```diff
@@ -10,19 +10,24 @@ use walkdir::WalkDir;
 
 use crate::template::{Template, TemplateParameter};
 
-// TODO: better name, e.g. Gather???
+/// Executes a template to the point where it is ready to generate
+/// artefacts.
```
lann, line 14: I new difference in commonwealth spelling I hadn't been aware of!

lann (approved):

PR #467, opened 2022-05-11
Confirm before writing template into directory with existing content

itowlson:
Because it's usually not what you intended.


radu-matei (approved):
Manually tested, things LGTM.

Thanks!

Issue #438, opened 2022-05-04
stdio in spin


Mossaka:
How does application wire it's own stdin/out/err path to spin such that spin pass them to `WasiCtxBuilder`? 

I found the relevent implementation here: https://github.com/fermyon/spin/blob/a9181f7b7ed290ee0b724ccb266f7d8d00f28431/crates/http/src/spin.rs#L217-L235

lann:
We don't currently expose stdio streams, we just save them to logs. PR #431 aims to extend this behavior to optionally forward guest stdio to host stdio. Could you expand on what you'd like to do? If you have another use case it might change the design of #431.

radu-matei:
ref #380

Mossaka:
> Could you expand on what you'd like to do?

I am using the Http Trigger in an application, which has inherited stdio (they are path to stdio files). I want to wire them to spin's stdio, or more specifically, `WasiCtx`'s stdio

radu-matei:
As @lann points out, #380 and #431 address redirecting component stdio to Spin's standard output/error.
So if the goal is to capture logs, #380 is the place we are trying to figure out the user stories for that.

(see https://github.com/fermyon/spin/pull/431/files#diff-e009a66ed5d35b924eb37f0c5fc5c5abd0abf68521a94d316a4dba88c97410a9R1)

lann:
I was mistaken - we _do_ expose stdio, but only as buffers (`Vec<u8>`), so you can pass in a complete buffer to stdin and get the stdout/stderr buffers. What we don't support is a _streaming_ interface to either of these.

For your use-case, can you clarify if you want stdio for logging or some other purpose? The Wagi HTTP executor, for example, uses stdin to provide the request to the guest, and parses stdout into the response:
https://github.com/fermyon/spin/blob/0f51ad0b86b798ae18cb4f211f0c3fcfbbe9a156/crates/http/src/wagi.rs#L54
https://github.com/fermyon/spin/blob/0f51ad0b86b798ae18cb4f211f0c3fcfbbe9a156/crates/http/src/wagi.rs#L116

danbugs:
[...]

PR #471, opened 2022-05-11
Implement `CustomLogPipes` for Spin Logging/IO

danbugs:
> That's exactly right. If you have your WasiFiles already, you can call `ModuleIoRedirects::new` directly, and pass that to `prepare_component`. (At least, you'll be able to if that PR is accepted.)
> 
> WAGI does this in a very simple form because it needs different behaviour for stdin, and a slightly different interaction with the buffer for stdout. It's really valuable to know that this is more broadly useful as a public customisation point though!

To provide an update on this from #438 (and maybe close that issue soon), it turns out that I couldn't really make the current architecture use our own pipes — As previously implied, the meat and potatoes of logging is done in `execute`:
https://github.com/fermyon/spin/blob/c3ff7d90fe115daa3fa85ce3e626c738e7baef76/crates/http/src/spin.rs#L19-L59

It's there that `capture_io_to_memory`, `prepare_component` (that receives redirects), and, subsequently, `redirect_to_mem_buffer` are called.

Considering our builder is set w/ `with_engine()`, as per:
https://github.com/fermyon/spin/blob/1fbfb23b4e96a580d6d4f4b6d617e916974493f1/crates/engine/src/lib.rs#L103-L119

I tried emulating the work done in `execute` in the `build` stage similar to how the `store` does it:
https://github.com/fermyon/spin/blob/0f51ad0b86b798ae18cb4f211f0c3fcfbbe9a156/crates/engine/src/lib.rs#L304-L316

... but, nothing came of it as `execute` would eventually run and override/never access my `RuntimeContext` WASI context data.

All and all, this PR allows users to set `CustomLogPipes` when creating their `ExecutionContextConfiguration` — these `CustomLogPipes` are later accessed by `execute` and passed in to `capture_io_to_memory` (which has a side-effect on `prepare_component`) providing a good and needed override of Spin's default settings.


Mossaka:
@lann @itowlson @radu-matei for review~

crates/engine/src/io.rs
```diff
@@ -38,6 +40,63 @@ pub trait OutputBuffers {
...
+                    .unwrap();
+        Self(f, self.1.clone())
+    }
+}
+
+/// CustomIoPipes that can be passed to `ExecutionContextConfiguration`
+/// to direct out and err
+#[derive(Clone, Debug)]
+pub struct CustomLogPipes {
+    pub(crate) stdout_pipe: PipeFile,
```
Mossaka, line 83: what's the reason for leaving stdin_pipe behind?

danbugs: I've just added it!

danbugs:
_Note: I had to do a force push to fix commits that weren't signed-off~ I was under the impression that having a `git config --global` set w/ `user.email` and `user.name` made it so that `-s` in `git commit` wasn't necessary 🤷_

danbugs:
⚠️I think my rebases made this PR a bit messy, so I remade it here at #482 for ease of review~

PR #485, opened 2022-05-12
feat(Makefile): add make doc command

michelleN:
+ adds a `doc` make target that generates a new doc
from template
+ can take name of document on command line optionally
+ usage: SPIN_DOC_NAME=mydocument.md make doc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 LOG_LEVEL ?= spin=trace
 CERT_NAME ?= local
+SPIN_DOC_NAME ?= new-doc.md
 
 .PHONY: build
 build:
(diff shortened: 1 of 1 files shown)


radu-matei:
Rather than a bespoke script, I'd rather add the missing features to `bart`, the CLI for Bartholomew:

```
❯ bart new post --help
bart-new-post 0.1.0
Create a new post

USAGE:
    bart new post [OPTIONS] <dir> [file]

FLAGS:
    -h, --help    Prints help information

OPTIONS:
        --author <author>              Author of the post [default: John Doe]
        --description <description>    Description for the post
        --type <post-type>             Type of the post [default: post]
        --template <template>          Template for the post [default: blog]
        --title <title>                Title for the post [default: Untitled]

ARGS:
    <dir>     Path to the directory where to create the new post
    <file>    Name of the file [default: untitled.md]
```

It seems like it could be useful to either have a template `.md` file, or have the ability to add `extra` key/value pairs.
What do you think?

michelleN:
Just wanted to make it easier to write docs without another dependency/tool in the mix. I think a template.md file would be great.

michelleN:
closing in favor of #488

michelleN:
@radu-matei re-opened this for more discussion. The thought occurred to me last night that the whole reason I wanted this make command is so I didn't have to use bart to generate the date for me. Otherwise, I would just copy/paste. Also, I kinda like this date format a little better because it's easier to understand.

Do you think the process should be create a template, copy/paste, insert the date manually and then put your content in? I just feel like the make target takes so much thought out of the whole thing like it puts the file in the place it needs to be, generates the date for you, etc. 

Why do we need to stick to using bart? What is the dependency here? Maybe I am missing something.

radu-matei:
We definitely don't _need_ to use `bart`!
My only intention was to understand what was missing from the Bartholomew CLI so we can improve the experience when using it, and not to force us to take a dependency on it.

For transparency, I've been copy-pasting the page headers in all docs I've written a while back, so having this target helps for sure, happy to have it merged. 

I still think adding this to the CLI would be a good idea, and I should have been a bit more clear that this PR and improving Bart are not mutually exclusive!

Thanks!

radu-matei (approved):

michelleN:
I see! Thanks for explaining @radu-matei !

PR #486, opened 2022-05-12
Option to cache files from bindle

### LINKED ISSUE #455: Global cache
There are a few scenarios that would benefit from having a global Spin cache:

- applications run from Bindle are currently pulled every time, and having a cache would reduce network traffic significantly
- in the context of #454, local applications with a Bindle module source, the module cache would be first checked before pulling the module
- the Wasmtime module cache could also be part of this global cache

itowlson:
Fixes #455.

This feels pretty untidy in places.  If anyone has thoughts on how it could be tidied, I'd really appreciate it!

I did wonder about sticking the cache option into the `BindleConnectionInfo` so it flowed without changing so many things, but I'm not sure that feels quite right...

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -168,6 +168,73 @@ mod integration_tests {
     }
 
     #[tokio::test]
+    async fn test_can_cache_bindle_data() -> Result<()> {
+        // start the Bindle registry.
+        let config = BindleTestControllerConfig {
+            basic_auth_enabled: false,
+        };
+        let b = BindleTestController::new(config).await?;
+        let url = b.url.clone();
+
+        // push the application to the registry using the Spin CLI.
+        run(
+            vec![
+                SPIN_BINARY,
+                "bindle",
+                "push",
+                "--file",
+                &format!(
+                    "{}/{}",
+                    RUST_HTTP_STATIC_ASSETS_TEST, DEFAULT_MANIFEST_LOCATION
+                ),
+                "--bindle-server",
+                &b.url,
+            ],
+            None,
+        )?;
+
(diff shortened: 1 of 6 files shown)


lann:
Have you seen bindle's (apparent) caching support? https://docs.rs/bindle/latest/bindle/cache/index.html

itowlson:
@lann RED FACE HERE

I had not - thanks!  That would certainly be a more desirable way to do this.  I'll take a look when I've finished kicking myself.

itowlson:
@lann I had a go with the Bindle cache module.  Although it claims to be for both client and server side use, it currently seems to accept only server providers as sources.  So, sadly, I think this is a non-starter for now.

lann:
> This feels pretty untidy in places. If anyone has thoughts on how it could be tidied, I'd really appreciate it!

How about a new `BindleLoader` struct with e.g. `url`, `base_dst`, `cache_dir` fields, refactoring some of these functions into methods?

tests/integration.rs
```diff
@@ -167,6 +167,73 @@ mod integration_tests {
...
+        assert_status(&s, "/static/thisshouldbemounted/1", 200).await?;
+        assert_status(&s, "/static/thisshouldbemounted/2", 200).await?;
+        assert_status(&s, "/static/thisshouldbemounted/3", 200).await?;
+
+        assert_status(&s, "/static/donotmount/a", 404).await?;
+
+        drop(s);
+        drop(b);
+
+        // The Bindle server is now unavailable but Spin should still be able to use the cache.
```
radu-matei, line 217: ❤️

PR #398, opened 2022-04-26
SIP: CloudEvents HTTP for spin SDK

Mossaka:
Signed-off-by: Jiaxiao Zhou <jiazho@microsoft.com>

Issue #394, opened 2022-04-24
Feature: CloudEvents component


Mossaka:
Enabling [CloudEvents](https://cloudevents.io/) component could make spin a great solution to serverless and event-driven applications. I am thinking about an event trigger that subscribes to a addressible broker, and whenever the broker emits an event, the event trigger will instantiate a wasm module to hanlde. Or the trigger itself can emits an event. 

For example the trigger can subscribe to a [knative event source](https://knative.dev/docs/eventing/sources/).

I am proposing the following component sdk. Let me know your thoughts. @radu-matei 
```rust
use anyhow::Result;
use spin_sdk::{
    event::{Event},
    event_component,
};

/// A simple Spin event component.
#[event_component]
fn trigger(event: Event) -> Result<()> {
    println!("{}", event);
    Ok(())
}
```

radu-matei:
@Mossaka this looks like a *great* idea!

For larger work items, we try to write a Spin Improvement Proposal (SIP) - see a few [examples](https://github.com/fermyon/spin/pulls?q=is%3Apr+sort%3Aupdated-desc+label%3Asip+).

radu-matei:
Thanks a lot for yesterday's presentation, @Mossaka!

ref #398 

After going through the CloudEvents [core spec](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md),
the [HTTP webhook spec](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md),
and the [protocol bindings](https://github.com/cloudevents/spec/tree/main/cloudevents/bindings),
here is roughly how I am thinking about implementing support for CloudEvents in Spin.

## Responding to HTTP webhooks

The CloudEvents [HTTP webhook specification](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md)
describes the pattern to deliver notifications through HTTP endpoints.

In this scenario, the _producer_ sends a valid HTTP request with a CloudEvents
payload (potentially with support for the [validation and abuse protection](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md#4-abuse-protection)
protocol), and expects a valid HTTP response, with an optional CloudEvents payload.

In handling the incoming request, a handler component can use external services
(and potentially send new CloudEvents payloads), but the interaction is
limited to the request-response model of HTTP/1.1.
Because of that, it seems like this can be implemented with the current HTTP
trigger — the incoming webhook is an HTTP request, with a CloudEvent payload,
[...]

PR #502, opened 2022-05-16
feat: added HTTP webhook validation to spin SDK

Mossaka:
this PR adds the http webhook validation function to Spin SDK; The detailed specification for validation is written [here](https://github.com/cloudevents/spec/blob/main/cloudevents/http-webhook.md). 

This is the first PR to implement the Spin SIP #398 . See related discussion here #394


sdk/rust/src/webhook.rs
```diff
@@ -0,0 +1,110 @@
...
+use crate::outbound_http::send_request;
+use anyhow::Result;
+use std::str::FromStr;
+
+static ORIGIN_REQUEST_HEADER: &str = "webhook-request-origin";
+static CALLBACK_REQUEST_HEADER: &str = "webhook-request-callback";
+static RATE_RESPONSE_HEADER: &str = "webhook-allowed-rate";
+static ORIGIN_RESPONSE_HEADER: &str = "webhook-allowed-origin";
+
+/// validate the webhook according to this [spec](https://github.com/cloudevents/spec/blob/v1.0/http-webhook.md#4-abuse-protection).
```
radu-matei, line 11: nit: could you please use capital letters for doc comments?

sdk/rust/src/webhook.rs
```diff
@@ -0,0 +1,110 @@
...
+pub fn validate_webhook(req: Request, callback_allowed: bool, rate: &str) -> Result<Response> {
+    if req.method() != http::Method::OPTIONS {
+        return Err(anyhow::anyhow!("invalid method"));
```
radu-matei, line 14: This can be simplified with `bail!("invalid method")`.

lann:
We have been discussing package organization and I think it would make sense for this to live in an outside repository. @michelleN suggests that a new `deislabs/spin-cloudevents` repo might be a good home for this code plus any future cloudevents trigger executor(s) (which should become much easier with #513!).

lann:
Oh! We were also discussing adding a "related projects" section to this repo's README, which would include a link to your repo.

Mossaka:
@lann okay I will try #513 out on a new repo.

lann:
Is this spec used anywhere apart from CloudEvents?

Mossaka:
> Is this spec used anywhere apart from CloudEvents?

so far I haven't seen it being used in other places.

radu-matei:
Thanks a lot for the contribution, @Mossaka!

It seems like it would make sense for this functionality to be part of a repo that was focused on adding CloudEvents support for Spin, rather than in the Spin repo itself.
As we were discussing in the developer meeting, that repo could evolve to include support for subscription managers and other things.

Then we can link the repo in the documentation and in the readme.
What do you think?

Mossaka:
@radu-matei yep I will move to spin-cloudevents repo.

Issue #490, opened 2022-05-13
Template "project name" parameter


fibonacci1729:
It's probably safe to assume this is a required option that needs to be set via cli or through prompt before expansion.

We should disallow the redefining this parameter in a template manifest

itowlson:
I think this is done now - @fibonacci1729 was there anything else that needed to be done for this?

fibonacci1729:
Just verified this works per our discussions in the linked/merged PRs. Thank you! I'll close.

Issue #479, opened 2022-05-12
`spin new` should create a new directory by default


lann:
Options:

1. ~~Existing behavior (don't create a new directory). Somewhat surprising and very annoying when a bunch of files get dropped into your top-level "projects" directory~~ #467 mitigates this
2. Create a new directory by default (if `--output` isn't set)
3. Magic: if the current directory is empty, use it; otherwise, do option 2
4. Ask: prompt for the output directory, maybe with a default value from option 3

Mossaka:
@itowlson and I discussed similar issue in https://github.com/fermyon/spin/issues/463.

lann:
#467 makes this less annoying, but I think this would still be a nice improvement.

itowlson:
I have a feeling that 'default to current directory' is inherited from the old 'copy an example' behaviour.  We certainly don't need to keep it.  The reason for it, I assume, was so people could create a directory, cd into it and do all their spinning there.

I'm not wildly keen on creating a new directory because it may not be obvious if it gets created into a busy place like Lann's project directory or Mossaka's root directory _grin_.  So I would definitely want to prompt rather than just create something!  Perhaps offer a choice of 'cancel, yes I really meant generate into busy directory, or pick new destination"...?

lann:
What about something like:
```console
$ spin new http-rust
...
Project name: foobar
...
# (current directory empty)
Output directory [.]:
# (current directory full)
Output directory [./foobar]:
```

itowlson:
The easiest thing is just to make `-o` mandatory.  If the user wants to generate into the current directory they can go `-o .`...?

Mossaka:
I personally think `Cargo new my-project` is pretty neutral and I don't have problems with it. How about we make `spin new` to have the same semantics?

itowlson:
So... make the output dir mandatory but make it positional rather than a flag?

Mossaka:
A flag is a feature that allows user to specify which path to create spin project, but the default value should be current_dir/proj_name

itowlson:
Now I'm confused, sorry.  Is this saying there should be a new positional parameter which doubles as a "project name" (when creating a new Spin app, the application name) _and_ the default for the directory, but there _should also_ be a flag parameter to override the output directory?  Apologies for being slow to understand!

fibonacci1729:
Following what Cargo does is my preferred option as well. "Project name" if not supplied on CLI should be prompted, the output of the template is then put into a new directory named "project name"

[...]

PR #510, opened 2022-05-18
Require a name when newing a tamplate, and use it as default output dir

### LINKED ISSUE #491: Spin new ignores project-name flag
When creating a new project from a template Spin prompts for a project name ignoring the non interactive "--name/-n" option. If the flag is set we shouldn't prompt

itowlson:
New syntax is:

```
spin new http-rust wondersite  # creates it in ./wondersite
spin new http-rust wondersite ~/projects/wondersite  # creates it in ~/projects/wondersite
```

The well-known parameter `project-name` is set to the name given on the CLI.  Templates no longer define this parameter.

Fixes #491, #479, #490.


lann (approved):

Mossaka:
This is awesome, thanks!

fibonacci1729 (approved):
Thanks Ivan!

Mossaka (approved):

PR #511, opened 2022-05-18
ref(.gitignore): ignore .vscode config

michelleN:


itowlson:
Won't this result in excluding things like the Rust and Go language server settings?

(It's frustrating because one wants to exclude settings like colour schemes but include actual workspace settings but VS Code doesn't seem to have a way to split those files.)

Also we don't currently use it but `./vscode` is where `launch.json` and `tasks.json` live and if we ever have those they should definitely travel with the project.

itowlson (approved):
LGTM!

PR #516, opened 2022-05-20
SIP: Application Manifest Redesign

lann:
[Rendered](https://github.com/lann/spin/blob/manifest-redesign-sip/docs/content/sips/xxx-manifest-redesign.md)

### DIFF
--- a/docs/content/sips/xxx-manifest-redesign.md
+++ b/docs/content/sips/xxx-manifest-redesign.md
@@ -1,0 +1,126 @@
+title = "SIP xxx - Application Manifest Redesign"
+template = "main"
+date = "2022-05-20T13:22:30Z"
+---
+
+Summary: A new design for Application Manifests (`spin.toml`)
+
+Owner: lann.martin@fermyon.com
+
+Created: May 20, 2022
+
+## Background
+
+Spin's Application Manifest file (usually named `spin.toml`) describes a Spin application's metadata,
+components, and trigger configurations. The current design supports a single trigger type and combines
+component and trigger configuration with an implicit one-to-one relationship:
+
+```toml
+name = "hello-world"
+trigger = { type = "http", base = "/hello" }
+
+[[component]]
+id = "world"
+# ...
+[component.trigger]
+route = "/world"
+```
+
+In order to support future features, we would like to enable more flexible trigger configurations:
+
+- Multiple trigger types in an application
+  - an HTTP trigger that sends asynchronous jobs to a Redis trigger
+- Multiple triggers associated with a single component
+  - a file server component serving multiple HTTP trigger routes
+- One trigger associated with multiple components
+  - an HTTP trigger with middleware components
+
+## Proposal
+
+### Global trigger config
+
+In order to allow for multiple trigger types, we remove the top-level `trigger.type` field.
+We also move all application metadata from the top-level config to a new `[application]`
+section, and all "global trigger config" to new `[application.trigger.<type>]` sections:
+
+```toml
+[application]
+name = "hello-world"
+[application.trigger.http]
+base = "/hello"
+```
+
+> In the short term we can continue to enforce the use of exactly one trigger type in an
+> application as a manifest validation step.
+
+### Trigger config
+
+In order to decouple triggers and components, we move triggers to new top-level
+`[[trigger.<type>]]` sections:
+
+```toml
+[[trigger.http]]
+route = "/world"
+component = "world"
+
+[[component]]
+id = "world"
+# ...
+```
+
+### Manifest version
+
+Since this proposal represents a backward-incompatible change to the manifest format, the manifest
+version changes. The existing `spin_version = "1"` field name leaves some room for misinterpretation
+as "compatible with Spin project version 1". As an optional part of this proposal, change that name:
+
+```toml
+spin_manifest_version = "2"
+```
+
+## Backward compatibility
+
+Existing "version 1" manifests can be transformed into this "version 2" data structure losslessly:
+
+```toml
+trigger = { type = "http", base = "/" }
+[[component]]
+id = "hello"
+[component.trigger]
+route = "/hello"
+```
+
+is equivalent to:
+
+```toml
+[application.trigger.http]
+base = "/"
+[[trigger.http]]
+route = "/hello"
+component = "hello"
+[[component]]
+id = "hello"
+```
+
+## Design options
+
+### Don't update the manifest version
+
+Instead of updating the manifest version we could not do that. It would be nice to include
+error message guidance and/or a `spin doctor` tool to upgrade the manifest format. This
+would be easy to detect if we switch the `spin_version` key.
+
+### Trigger IDs
+
+We might want to be able to reference individual triggers, for example to implement a
+`spin up --trigger=world`. It would save some compatibility headache if we require IDs now:
+
+```toml
+[[trigger.http]]
+id = "world"
+```
+
+Mandatory IDs would require a translation from the current "v1" manifest, which could either
+simply copy the component ID (if we're fine with component and trigger IDs being separate
+namespaces), or e.g. add a suffix to the component ID (`world` -> `world-trigger`), which
+could technically cause conflicts but doesn't seem likely in practice.


docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Manifest version
+
+Since this proposal represents a backward-incompatible change to the manifest format, the manifest
+version changes. The existing `spin_version = "1"` field name leaves some room for misinterpretation
+as "compatible with Spin project version 1". As an optional part of this proposal, change that name:
+
+```toml
+spin_manifest_version = "2"
+```
```
lann, line 76: Per discussion with @fibonacci1729 we don't _have_ to do this. See "Don't update the manifest version" below.

fibonacci1729: I'm in favor of renaming spin_version to spin_manifest_version but I think we should wait to officially freeze version 1 manifest changes until spin 1.0

I'm hesitant toward bumping the manifest version before our first major release.

fibonacci1729 (approved):
I really like this model!

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+[application]
+name = "hello-world"
+[application.trigger.http]
+base = "/hello"
+```
+
+> In the short term we can continue to enforce the use of exactly one trigger type in an
+> application as a manifest validation step.
+
+### Trigger config
```
Mossaka, line 53: how will manifest define triggers that are outside of the spin repo?

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+id = "world"
+# ...
+[component.trigger]
+route = "/world"
+```
+
+In order to support future features, we would like to enable more flexible trigger configurations:
+
+- Multiple trigger types in an application
+- Multiple triggers associated with a single component
```
Mossaka, line 32: Isn't the difference of trigger function signature prevents one component associating with multiple triggers? 

For example, a Redis trigger and a HTTP trigger can't associae with a single component since their require linking to different .wit files.

Mossaka (commented):
enjoyed this proposal! Had some minor questions

docs/content/sips/xxx-manifest-redesign.md
itowlson: Agree on this from an aesthetic point of view, but without some sort of schema ID, a user who downloads a new Spin build that changes the schema is going to get cryptic parse errors working with existing manifests.  Can we make sure that we do at least enough to give a meaningful error like "this version of the manifest is no longer supported"?

itowlson: Says the man who just changed templates.

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+### Don't update the manifest version
+
+Instead of updating the manifest version we could not do that. It would be nice to include
+error message guidance and/or a `spin doctor` tool to upgrade the manifest format. This
+would be easy to detect if we switch the `spin_version` key.
+
+### Trigger IDs
+
+We might want to be able to reference individual triggers, for example to implement a
+`spin up --trigger=world`. It would save some compatibility headache if we require IDs now:
```
itowlson, line 113: This feels a bit intrusive in the simple case (which is likely to be common especially amongst the vital tyre-kicker demographic).  Could we, say, plan to default the ID to the trigger type, and require IDs only if there are multiple triggers of the same type?

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+[[component]]
+id = "hello"
+```
+
+## Design options
+
+### Don't update the manifest version
+
+Instead of updating the manifest version we could not do that. It would be nice to include
+error message guidance and/or a `spin doctor` tool to upgrade the manifest format. This
```
itowlson, line 107: I like this - super helpful!

docs/content/sips/xxx-manifest-redesign.md
itowlson: Oh wait is this meant to refer to individual components under a trigger?  (or should I say trigger-config-to-component maps)  This is what the example syntax seems to show, but if so I fear we may have two meanings for the term 'trigger'.

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+### Global trigger config
+
+In order to allow for multiple trigger types, we remove the top-level `trigger.type` field.
+We also move all application metadata from the top-level config to a new `[application]`
+section, and all "global trigger config" to new `[application.trigger.<type>]` sections:
+
+```toml
+[application]
+name = "hello-world"
+[application.trigger.http]
```
itowlson, line 46: This is a table rather than a table array, which means that (at some point) we can have multiple triggers of different types, but not multiple triggers of the same type, is that right?  Or do we envisage revving the version at that point?

(Of course I might just be misunderstanding my TOML syntax here!)

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+> In the short term we can continue to enforce the use of exactly one trigger type in an
+> application as a manifest validation step.
+
+### Trigger config
+
+In order to decouple triggers and components, we move triggers to new top-level
+`[[trigger.<type>]]` sections:
+
+```toml
+[[trigger.http]]
```
itowlson, line 59: I'm not sure I see the value of this over putting the trigger-related config under the component.  It adds a layer of indirection which means:

1. Two places to edit the component ID
2. When I look at the component it's not obvious what it's hooked up to (e.g. HTTP route)

itowlson (commented):
I like this!  The compatibility story is good, which minimises the number of people who will go "eh, too hard" and stop kicking the tyres, and I like the improved extensibility and cleaner organisation.  A few comments/gripes but I definitely like the direction!

docs/content/sips/xxx-manifest-redesign.md
lann: A component can implement multiple interfaces.

docs/content/sips/xxx-manifest-redesign.md
lann: This is "application-wide HTTP trigger config". I'm not entirely happy with the syntax, but didn't come up with anything better.

docs/content/sips/xxx-manifest-redesign.md
lann: They are identified by string `type`, which maps to a binary `spin-trigger-<type>`.

docs/content/sips/xxx-manifest-redesign.md
lann: I think my "trigger" is equivalent to your "trigger config". Each trigger has a "type", and a "trigger executor" can serve multiple triggers of a given type.

itowlson: So, in the HTTP case, each route is considered a separate trigger?  (And for Redis it would be each channel, etc.?)

docs/content/sips/xxx-manifest-redesign.md
lann: A couple of motivations:

- I think it makes more sense conceptually for triggers to be a top-level entity. Another design I considered was just inverting the component/trigger relationship, i.e. turning `component.trigger` into `trigger.component`, but:
- I want to enable HTTP middleware (and other kinds of trigger-component relationships), where a single trigger may reference multiple components (one "handler" and zero+ "interceptors"). I like the consistency of referencing them all by ID.

It's true that it introduces some duplication and potential non-locality, but in practice you can group a trigger and its "main" component together in the file, e.g.:

```toml
[[trigger.http]]
route = "/hello"
component = "world"
[[component]]
id = "world"
```

docs/content/sips/xxx-manifest-redesign.md
lann: Yep, and a single component may handle multiple triggers.

docs/content/sips/xxx-manifest-redesign.md
radu-matei: Theoretically, you can do that today if you implement and export more than one interface (both the HTTP and the Redis entrypoint) and use the same `.wasm` file.

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+[[component]]
+id = "world"
+# ...
+[component.trigger]
+route = "/world"
+```
+
+In order to support future features, we would like to enable more flexible trigger configurations:
+
+- Multiple trigger types in an application
```
radu-matei, line 31: I have been wanting to have an application with multiple triggers for a long time.
However, one of the advantages of this separation is that you can clearly decide if your engine can support running the application or not — whereas this way you can be in a situation where the engine can run some components in the app manifest, but not others.

I don't have a solution, and I'm not proposing we don't do this, just opening the discussion to see what others are thinking here.

docs/content/sips/xxx-manifest-redesign.md
radu-matei: ref #519

docs/content/sips/xxx-manifest-redesign.md
lann: I'm not sure I understand what you mean by:

> you can clearly decide if your engine can support running the application or not

If you can decide whether you can run _one_ type then surely you can decide whether you can run _all_ the types? :thinking:

michelleN: @radu-matei would you mind giving a concrete example?

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+- Multiple triggers associated with a single component
+- One trigger associated with multiple components
```
michelleN, line 33: Based on the convo, it'd be nice if we could put a simple use case under the bullet for each of these 3 scenarios to help clarify why we would want the functionality.

docs/content/sips/xxx-manifest-redesign.md
michelleN: @itowlson's perspective resonates with me because I have been thinking about Spin as a framework for building components and so components are the central subject in my head and everything revolves around them. However, when I tell myself that the spin.toml's job is to wire up triggers to components, then making the triggers a top level entity makes much more sense to me.

docs/content/sips/xxx-manifest-redesign.md
```diff
@@ -0,0 +1,123 @@
...
+id = "hello"
+[component.trigger]
+route = "/hello"
+```
+
+is equivalent to:
+
+```toml
+[application.trigger.http]
+base = "/"
```
michelleN, line 94: Can this top level `application.trigger.http` be optional/omitted if folks don't need to change base = "/"? The concept of having a base is a little distracting anyway when first looking at a spin.toml file.

docs/content/sips/xxx-manifest-redesign.md
michelleN: +1 ... I'm such a sucker for cute names like "spin doctor"

docs/content/sips/xxx-manifest-redesign.md
michelleN: This seems really useful especially for testing purposes but it would be really nice if we could make this optional so we can make "the mean time to dopamine" to understand a spin.toml file and spin up as low as possible.

docs/content/sips/xxx-manifest-redesign.md
lann: Yes. I suspect we could drop `base` entirely, in favor of custom config e.g. `route = "${base}/route"`, but I didn't want to drop the "global trigger config" entirely.

docs/content/sips/xxx-manifest-redesign.md
lann: Yeah that's why this idea is down here and not up there :point_up_2:.  :slightly_smiling_face:

docs/content/sips/xxx-manifest-redesign.md
lann: Done.

michelleN: thank you

michelleN (approved):

docs/content/sips/xxx-manifest-redesign.md
radu-matei: We have briefly talked about having specialized engines for single triggers that would be able to handle a single component type — this might be relevant in that scenario.

(node A can only handle HTTP components, while node B pub/sub components; not an issue right now, since we our current CLI implements both, but we will have to be mindful of how we handle multiple component types and nodes if we ever have such specialized engines.)

Definitely not a blocker for this PR, resolving.

Issue #517, opened 2022-05-22
Point to specific branch of templates in Spin Quickstart documentation


kate-goldenring:
**Issue**
The Spin [Quickstart](https://spin.fermyon.dev/quickstart/) walks through first installing the latest release of spin and then using one of the templates from `main` with `spin templates install --git https://github.com/fermyon/spin`.

If someone is risk averse (like me 😅) and opts to install a stable release of spin (`v0.2.0`) and still references the templates from `main`, the templates may be incompatible with Spin.

For example, right now, Spin `v0.2.0` is incompatible with the templates state after [this PR merged](https://github.com/fermyon/spin/pull/510), which adds an extra parameter to `spin new` for project name. This is because Spin `v0.2.0` does not require the name parameter while the templates assume it will be provided:
```
kagold@ubuntu:~/spin-quickstart$ spin new http-rust
Project description: Some description
HTTP base: /
HTTP path: /...
Error: liquid: Unknown variable
  with:
    requested variable=project-name
```

**Potential solutions**
Modify the quickstart document to either:
1. Explicitly state that the canary release should be used in the quickstart guide -- maybe most folks are already doing this when instructed to "install latest release", and I am an outlier here. In that case, `quickstart.md` could quickly be updated to match `main` CLI with a `<NAME>` parameter.
2. Add a note, suggesting adding `--branch $RELEASE_TAG` if not installing canary. 
3. Instruct installing a specific stable release and the associated branch of templates (i.e. `spin templates install --git https://github.com/fermyon/spin/ --branch v0.2.0`). Then, the quickstart demo can be updated each release cycle to point to the new templates and contain any updates to the CLI i.e the addition of `<NAME>` in `spin-canary new [OPTIONS] <TEMPLATE_ID> <NAME>`

itowlson:
Thanks for this catch - you are absolutely right and this is something we'll need to plan for in the future and in other contexts.  Apologies for the inconvenience.

itowlson:
I believe #556 provides a longer-term solution to this and we can now close it.  @kate-goldenring is that okay?

kate-goldenring:
#556 is great! Thanks @itowlson. Closing

PR #518, opened 2022-05-22
Install stable templates if using stable build

itowlson:
This is a partial fix for #517.

It's partial because we need a longer term strategy for things like this, whether that's keeping `HEAD` on the 'stable' branch or auto inferring the matching tag or what.  Perhaps a topic for the Spin meeting.  It's also partial because it's not sustainable, we don't want to have to rev this as part of every release!

Is it possible to get this fix rolled out to the 0.2.0 docs site, please?

### DIFF
--- a/docs/content/quickstart.md
+++ b/docs/content/quickstart.md
@@ -60,7 +60,7 @@ to install a starter set.
 We first need to configure the [templates from the Spin repository](https://github.com/fermyon/spin/tree/main/templates):
 
 ```console
-$ spin templates install --git https://github.com/fermyon/spin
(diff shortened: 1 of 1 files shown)


michelleN (approved):
Manually tested and works! Although I also like @lann's suggestion. Maybe we could call it `latest-release`?

itowlson:
*chuckle* agree that `stable` is stretching it right now, but something along those lines seems like a nice simple solution.

Shall I merge this now so we can push new docs, and then get it to use the default tag when we've decided what to call it?

radu-matei (approved):

PR #523, opened 2022-05-24
test: add integration test for `spin build`

mooori:
WIP. Description and tests using `workdir` coming up.


mooori:
The tests of `spin build` might be better off in a separate file, for instance
`tests/cli_build.rs`. However, that requires code duplication (for example
[`fn run`](https://github.com/fermyon/spin/blob/131c0ae973296ea244d11efe79994e0ee06598c1/tests/integration.rs#L481-L507)) or placing common code in a module like `tests/common/mod.rs`.

PR #524, opened 2022-05-25
wip

lann:

### DIFF
--- a/crates/engine/src/io.rs
+++ b/crates/engine/src/io.rs
@@ -1,356 +1,342 @@
-use cap_std::fs::File as CapFile;
-use std::{
-    collections::HashSet,
-    fmt::Debug,
-    fs::{File, OpenOptions},
-    io::{LineWriter, Write},
-    path::PathBuf,
-    sync::{Arc, RwLock, RwLockReadGuard},
-};
+use std::io::{Read, Write};
+
+use spin_manifest::CoreComponent;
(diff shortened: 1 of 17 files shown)


crates/engine/src/io.rs
```diff
@@ -1,356 +1,342 @@
...
-    pub fn should_follow(&self, component_id: &str) -> bool {
-        match self {
-            Self::None => false,
-            Self::All => true,
-            Self::Named(ids) => ids.contains(component_id),
+impl Default for ComponentStdio {
+    fn default() -> Self {
+        Self {
+            stdin: StdInput::Null,
+            stdout: StdOutput::Null,
```
itowlson, line 24: Is this null as in /dev/null?  Suggest using `Discard` if so

itowlson (commented):
This looks great!

crates/engine/src/io.rs
lann: I stole from https://doc.rust-lang.org/std/process/struct.Stdio.html#method.null but `Discard` would be fine too.

itowlson: Oh, if it's an existing convention then happy to stick with it!

lann:
OK I'm calling this one a failure for now. Hooking stdio into tracing worked fine but reproducing the logging and console follow features in the CLI was shockingly difficult.

PR #526, opened 2022-05-26
ci(*): add docs site content check

vdice:
* adds a `check-content` make target (and corresponding step in the GH build.yml workflow) using [bart](https://github.com/fermyon/bartholomew#the-bartholomew-cli) to check the markdown content/templating for the spin docs


radu-matei (approved):
Thanks!

PR #529, opened 2022-05-27
chore(*): address clippy errors

vdice:
* Addresses two clippy errors seen while running `make test` recently

1.  Multiple instances of:
```
error: you are deriving `PartialEq` and can implement `Eq`
   --> crates/manifest/src/lib.rs:278:24
    |
278 | #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
    |                        ^^^^^^^^^ help: consider deriving `Eq` as well: `PartialEq, Eq`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq
```

2. One instance of the following.  Please double-check my work here to make sure it is the right fix.
```
error: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
  --> crates/engine/src/host_component.rs:33:5
   |
33 | /     pub(crate) fn insert<'a, T: 'static, Component: HostComponent + 'static>(
34 | |         &mut self,
35 | |         linker: &'a mut Linker<RuntimeContext<T>>,
36 | |         host_component: Component,
37 | |     ) -> Result<()> {
   | |___________________^
   |
   = note: `-D clippy::needless-lifetimes` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
```

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -123,7 +123,7 @@ impl Resolver {
 }
 
 /// A config key.
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Eq, PartialEq)]
 pub struct Key<'a>(&'a str);
 
 impl<'a> Key<'a> {
--- a/crates/config/src/template.rs
+++ b/crates/config/src/template.rs
@@ -4,7 +4,7 @@ use crate::{Error, Result};
 
 /// Template represents a simple string template that allows expressions in
 /// double curly braces, similar to Mustache or Liquid.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub(crate) struct Template(Vec<Part>);
 
 impl Template {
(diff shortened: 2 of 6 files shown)


itowlson:
I don't see these warnings, whether I run `make test` or `cargo clippy --all-targets --all-features -- -D warnings` on its own.  I've tried updating to latest Rust but still no joy.

vdice:
@itowlson thank you for attempting to repro.  It looks like my default rust channel was config'd to nightly for a bit (was investigationg https://github.com/fermyon/bartholomew/issues/75).  I switched back to `stable` and, indeed, am not hitting these errors.

I imagine the progression is that changes/linter updates/etc migrate from `nightly` into `stable` as they are vetted/hardened?

In any case, I'll close this PR and we can revisit if/when the errors show up on `stable` and/or in CI.

PR #533, opened 2022-06-03
feat: accept a values file for `spin new`

### LINKED ISSUE #492: Accept a values file for spin new
Right now a user has to either work through the prompt or manually set KEY=VALUE strings through the CLI to satisfy the parameters defined in a template. 

It'd be nice to pass a values.toml in the non interactive experience to avoid long CLI incantations

mooori:
Makes `spin new` accept a file which sets parameters defined in the template.
The file can be provided via `--values-file`.

- If a parameter is set in *both* the file and as CLI option, the value provided
  as option takes precedence.
- If a value is not specified (neither in the file nor as CLI option), the user
  will be asked to enter the missing value in a prompt. I.e. not every
  parameter must be set in the file.

# Example usage
```toml
# ./values.toml
project-description = "This is a project."
http-base = "/"
http-path = "/foo"
```

After building spin with these changes, the following command scaffolds a new
app with parameters read from the file:

```bash
spin new --values-file values.toml http-rust foo
```

Fixes #492

cc @fibonacci1729

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -1,7 +1,13 @@
-use std::{collections::HashMap, path::PathBuf, str::FromStr};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
 
 use anyhow::{anyhow, Context, Result};
 use clap::Parser;
+use path_absolutize::Absolutize;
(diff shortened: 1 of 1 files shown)


fibonacci1729:
Thanks @mooori ! This is excellent

fibonacci1729 (approved):

Issue #216, opened 2022-03-23
Update tests to use `sdk` 


fibonacci1729:
The integration tests take a direct dependency on the relevant`wit` files; We should update these tests to use the `sdk`

ejmg:
Hey there, I might claim this issue. I started working with Spin and have at least one test going with the SDK instead of directly relying on the wit files & `wit_bindgen`. I'll PR my existing fork. What I'll need to know is how to enforce the actual intended behavior behind the integration tests given I'm not completely familiar with what is actually being enforced with respect to those wit files.

radu-matei:
That is great, thanks a lot, @ejmg!
Let us know if there is anything we can do to help.

vdice:
I revisited our tests today and as far as I can tell the only remaining test still using wit files is the [headers-env-routes-test](https://github.com/fermyon/spin/tree/main/tests/http/headers-env-routes-test).  It looks like there is an outstanding Draft PR in this area (#565), though it would need to be rebased or perhaps created anew.  Since we don't have an assignee at this time, I'm moving this into the Backlog.

PR #534, opened 2022-06-03
Begin replacing wit_bindgen with Rust SDK in integration tests

ejmg:
This PR addresses #216.

I'll need some feedback on the intended behavior for these tests to ensure the integrations are still checking for the desired behavior and functionality.

### DIFF
--- a/tests/http/simple-spin-rust/src/lib.rs
+++ b/tests/http/simple-spin-rust/src/lib.rs
@@ -1,37 +1,27 @@
-// Generate Rust bindings for interfaces defined in WIT files
-wit_bindgen_rust::export!("../../../wit/ephemeral/spin-http.wit");
+use anyhow::Result;
+use spin_sdk::{http::{Request, Response}, http_component};
 wit_bindgen_rust::import!("../../../wit/ephemeral/spin-config.wit");
 
-// Import the HTTP objects from the generated bindings.
-use spin_http::{Request, Response};
 
-struct SpinHttp {}
-impl spin_http::SpinHttp for SpinHttp {
-    // Implement the `handler` entrypoint for Spin HTTP components.
-    fn handle_http_request(req: Request) -> Response {
-        let path = req.uri;
+#[http_component]
+fn hello_world(req: Request) -> Result<Response> {
 
-        if path.contains("test-placement") {
-            match std::fs::read_to_string("/test.txt") {
-                Ok(text) => Response {
-                    status: 200,
-                    headers: None,
-                    body: Some(text.as_bytes().to_vec()),
-                },
-                Err(e) => Response {
-                    status: 500,
-                    headers: None,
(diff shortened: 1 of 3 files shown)


fibonacci1729 (approved):
Thanks for the PR @ejmg ! This all looks great. I think we can open a new issue regarding the note you left to make the logic of these tests (in the component) a little more robust. Would you mind dropping the note you left into an issue linking this PR?

ejmg:
Sounds good to me. I am getting into the car to take off for a small
camping trip until this Wednesday so I’ll file it myself if it can wait
until then; I planned on seeing through the PR either way.

fibonacci1729:
Sounds great! Most of the Spin folks are returning from an offsite this week so no rush! Happy camping! 🏕

ejmg:
Welp, camping plans changed. I'll open up a new issue with my comment and let me know if there is anything else I need to do on this PR before it gets merged.

tests/http/simple-spin-rust/src/lib.rs
```diff
@@ -1,37 +1,16 @@
...
+//       - whether URI path contains "test-placement"
+//         + if so, whether or not /test.txt exists and returning its contents, throwing an error otherwise...
+//       - if it doesn't contain test-placement, checking whether the config contains "message" and returning that
+//         + similar to test.txt, return error if config doesn't have message
+#[http_component]
+fn hello_world(_req: Request) -> Result<Response> {
+    Ok(http::Response::builder()
+        .status(200)
+        .body(Some("Hello, Fermyon!".into()))?)
+}
```
radu-matei, line 16: nit: missing new line

ejmg:
Sorry for redundant commit, I missed the newline comment! I've signed off and verified my commits along with the suggested edits.

ejmg:
@radu-matei and @fibonacci1729, I've updated this single test case to replicate the previous behavior of the test binary used by the integrations tests. To actually assert correctness, we would probably need to check the contents of the body but this is closer to correct behavior.

The one thing I'm wondering about right now (and as you'll see as a some `NOTE`s in the code) is that the return type of the component is no longer just a Response, but `anyhow::Result`. The error case for when the test asset file is not located is very different from the original code and I don't know how significant this is.

radu-matei:
Re: your note — you're not actually changing the response of the _Wasm component_ itself.
Rather, the Rust SDK takes care of that, and checks the `anyhow::Result<Response>` and returns the right value.

(for a bit more context on the WIT design for the HTTP component, which we might want to change eventually — the `Response` is directly sent to the client — so if we ever have a Rust error, we automatically send a 500 status back, without the user manually doing that.)

In terms of the commits, would you please squash to a single commit that is signed?

Thanks, and apologies for the git friction here..

ejmg:
@radu-matei good to go?

radu-matei (approved):
Thank you for your contribution, @ejmg!

LGTM

PR #541, opened 2022-06-07
Application manifest redesign

lann:
Implementing https://github.com/fermyon/spin/blob/main/docs/content/sips/005-manifest-redesign.md

### DIFF
--- a/crates/manifest/src/resolved.rs
+++ b/crates/manifest/src/resolved.rs
@@ -1,0 +1,219 @@
+use std::{collections::BTreeMap, path::PathBuf};
+
+use serde::{Deserialize, Serialize};
+use serde_value::Value;
+use serde_with::{hex::Hex, As};
+
+use crate::ManifestVersion;
+
+/// ResolvedManifest is a "resolved" application manifest, ready to be used by an executor.
+//
+// NOTE: The serialization of this data may be used for application de-duplication, so it
+// should be kept reasonably stable, at least for a particular Spin release.
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ResolvedManifest {
+    /// Manifest schema version.
+    pub spin_lock_version: ManifestVersion<0>,
+
+    /// Application metadata, for logging, debugging, etc.
+    pub metadata: ApplicationMetadata,
+
+    /// Custom configuration variables.
+    pub variables: BTreeMap<String, Variable>,
+
+    /// Per-trigger-type configuration. Must have an entry for every trigger type used by the application.
+    #[serde(default)]
+    pub trigger_types: BTreeMap<String, BTreeMap<String, Value>>,
+
+    /// Trigger definitions.
+    pub triggers: Vec<Trigger>,
+
+    /// Component definitions.
+    pub components: Vec<Component>,
+}
+
+/// Application metadata; used for logging debugging, etc.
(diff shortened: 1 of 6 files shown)


crates/manifest/src/resolved.rs
```diff
@@ -0,0 +1,219 @@
+use std::{collections::BTreeMap, path::PathBuf};
+
+use serde::{Deserialize, Serialize};
+use serde_value::Value;
+use serde_with::{hex::Hex, As};
+
+use crate::ManifestVersion;
+
+/// ResolvedManifest is a "resolved" application manifest, ready to be used by an executor.
```
lann, line 9: @itowlson

itowlson: Not sure what I'm being pinged on here but this looks like it is used both as the internal ready-to-use format _and_ a serialisation format, and I have never had good experiences of mixing those two... was that what you were flagging me for?

lann: Just flagging as this is my draft of a "spin.lock" which we have discussed. _Not_ reusing serialization structures for application logic is one of my major goals with all this refactoring noise.

crates/manifest/src/resolved.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    /// Application metadata, for logging, debugging, etc.
+    pub metadata: ApplicationMetadata,
+
+    /// Custom configuration variables.
+    pub variables: BTreeMap<String, Variable>,
+
+    /// Per-trigger-type configuration. Must have an entry for every trigger type used by the application.
+    #[serde(default)]
+    pub trigger_types: BTreeMap<String, BTreeMap<String, Value>>,
```
radu-matei, line 27: Will this be "slightly more typed" in the future?

crates/manifest/src/resolved.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// Component definition.
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Component {
+    /// Component ID.
+    id: String,
+
+    /// Component Wasm content reference.
+    wasm: ContentRef,
```
radu-matei, line 82: The pair of field name and struct are slightly confusing here.

crates/manifest/src/v1.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    fn kitchen_sink() -> Manifest {
+        toml::toml! {
+            spin_manifest_version = 1
+
+            [application]
+            name = "kitchen sink"
+            version = "9999.9.9"
+            description = "Test All Features"
+            authors = ["Charles Xavier <prof.x@xmen.io>"]
```
radu-matei, line 224: We lost the chance to own `xmen.io`...

crates/manifest/src/resolved.rs
lann: Nope, but this would get forwarded to the TriggerExecutor to deserialize into its own config model.

crates/manifest/src/resolved.rs
lann: I'm just making this stuff up as I go... `source` seems weird since at this point it is resolved.

lann: Also I'm just tricking you into reviewing this now; I'm going to pull this out for any PR that would land before 0.3.

Issue #528, opened 2022-05-26
Expose an Admin API


ecumene:
Spin keeps its logs and manifest on the filesystem, which makes them opaque to anything outside system its hosted on. Could we expose an API for administration of Spin apps? I think it would be an aliasable route, and flagged off by default. Maybe on a path like `http://localhost:3000/_spin/`?

I don't know what the shape of this API, but for my purposes something like
1. `/logs/:id` - Exposes the logs of any component as a service, something like [HTTPAppender](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/appender/HttpAppender.html)
2. `/` -  Exposes the manifest as JSON

fibonacci1729:
I like this @ecumene ! I wonder how we handle AuthN/AuthZ on these privileged endpoints. Perhaps to expose something like this we should start a discussion for API security for Spin? What do you think?

ecumene:
> Perhaps to expose something like this we should start a discussion for API security for Spin? 

[Hasura exposes an admin console for their service, but it's a flag. Maybe that's something we could do?](https://hasura.io/docs/latest/graphql/core/hasura-cli/hasura_console/)

fibonacci1729:
I wouldn't be opposed to that! Would you be interested in putting together a SIP so we can gather broader feedback? Or do you feel more comfortable chatting this out in this issue for a bit?

ecumene:
I'll take a crack at a SIP 😁

radu-matei:
This sounds great!
I think a flag to expose that initially sounds great — at some point we will have to integrate authz for the components / paths..

Looking forward to the SIP!

PR #542, opened 2022-06-07
SIP: Spin State WASI

ecumene:
Ref: #528

My first SIP! Please let me know how it looks.

[Rendered](https://github.com/fermyon/spin/blob/3f4ed4b4141c2b27ba035b82ffec6cab76b26d78/docs/content/sips/005-admin-api.md)

Signed-off-by: Mitchell Hynes <ecumene@users.noreply.github.com>


docs/content/sips/005-admin-api.md
```diff
@@ -0,0 +1,92 @@
+title = "SIP 005 - The Admin API
```
fibonacci1729, line 1: Minor thing, but the naming convention for SIPs in draft phase are SIP XXX - <Title>. Once merged, you can name it to its concrete index.

docs/content/sips/xxx-admin-api.md
```diff
@@ -0,0 +1,92 @@
...
+- it must have access to the file system which users may not have access to
+- being a server itself it must be forwarded out of whatever network users are
+  running it in, along with Spin itself
+- the file system is itself limited as new logs won't notify the user.
+  system file events could be used but they're notoriously platform dependent.
+
+If Spin exposed an "administrator's API" that is handled by the same route that
+Spin apps are served on we could open the door to many hypervisor-like tools.
+
+## Proposal
```
lann, line 36: Do you see anything that would prevent us from exposing any of this data as a host-provided Wasm interface (i.e. host component)? All of this API would still be relevant, but could be implemented as http trigger(s). One advantage of this approach would be that other trigger types - notably the upcoming "clock trigger" - could access the same info.

ecumene: No, in fact I think that might be better for my use case entirely.

I'm going to rewrite the Spin to account for this :)

ecumene:
I rewrote the Spin Admin API SIP to be based around a new WASI based FS interface instead. 

I'm not happy with the design, but if it was implemented in its current state it would serve my purposes.

Overall, I'd like to keep it as simple as possible. The interface I have for logs and the manifest suits me fine, maybe the persisting to disk part could be kicked down the road for later.

ecumene:
I left the original commit for posterities sake :)

docs/content/sips/xxx-spin-state-wasi.md
```diff
@@ -0,0 +1,131 @@
...
+
+Exposing `~/.spin/` to components doesn't work because all files and folders
+remain in the same state they were mounted at. Any new logs won't be included
+when read again.
+
+If instead of mounting a `~/.spin/` directory on the host machine, we allowed
+users to mount a directory internal to Spin in-memory this would be updated
+on subsequent reads while having the access control of a mounted directory.
+
+#### Mount: `/monitor/`
```
lann, line 66: I think we should namespace this _within_ the WASI FS to avoid conflicting with a component's own mounts. I think `/.spin/monitor/` would be sensible and give us a place to put future host-provided files; I have thought about exposing config variables in a similar way and those could go in `/.spin/config/`.

docs/content/sips/xxx-spin-state-wasi.md
```diff
@@ -0,0 +1,131 @@
...
+> B. `spin_source` is confusing unless you grok'd the docs for it
+
+```diff
+-files = [{ source = "~/.spin/[...]/logs", destination = "/logs" }]
++files = [{ spin_source = "/monitor/logs/", destination = "/monitor/logs" }]
+```
+
+The `/monitor` path is a special path in Spin where logs are accessed from. If
+users provide this path to their components under the `spin_source` source key
+then the component's logs would be accessible to the guest module.
```
lann, line 87: Rather than making this a magic mount point, I think we can add explicit configuration for "enable the monitor interface, please". Sort of like the existing `allowed_http_hosts` configures the outbound-http interface. Something like `spin_monitoring = true` on the component.

lann (commented):
A couple of comments, but I like the approach overall!

ecumene:
Thanks for the thoughtful input, updated!

michelleN (commented):
This is real neat @ecumene ! Could you describe what the Spin workflow/experience would look like for a user to access their logs? Once a user has used the `enable_monitoring` flag, will we stream the logs in the `spin up` and `spin deploy` output?

ecumene:
> Could you describe what the Spin workflow/experience would look like for a user to access their logs? 

Sure! When the flag is enabled, this directory is mounted to the host component:

```sh
tree /.spin/
└── logs
    └── [componentname]
        ├── stderr.txt
        └── stdout.txt
```

You can access this through WASI File Descriptors, which is just like reading any other file. This directory differs from if you were to mount your app's directory (`~/.spin/[appname]`) because the files are updated in realtime.

Issue #539, opened 2022-06-07
Error running spin on Ubuntu 22.04


scotty-c:
I found this issue running spin on my laptop once I updated to Ubuntu 22.04. It seems like there is an issue with loading the shared Openssl shared libraries with openssl3 package. I installed the packages in the readme `build-essential libssl-dev pkg-config`  downloaded the latest stable version of spin `0.2.0` and got the following error `./spin: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory`

I was able to get spin working by adding the package from the impish repos 
```
echo "deb http://security.ubuntu.com/ubuntu impish-security main" | sudo tee /etc/apt/sources.list.d/impish-security.list
sudo apt-get update 
sudo apt-get install libssl1.1
```

I thought I would report this as others might be having the same issue on Ubuntu 22.04

dicej:
@scotty-c I believe this has been addressed.  When you have a moment, would you please try [the latest canary release](https://github.com/fermyon/spin/releases/tag/canary)?  You'll probably want to `sudo apt-get remove --purge libssl1.1` first to make it a valid test.

bacongobbler:
This appears to be a duplicate of #498. I tested the latest canary and it appears to work fine now.

```
><> cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

```
><> spin --version
spin 0.2.0 (1cc8af7 2022-06-09)
```

This can be closed. Thanks @dicej!

PR #546, opened 2022-06-08
statically link OpenSSL using the vendored feature

dicej:
This is shamelessly stolen from https://github.com/fermyon/spin/pull/247, with
trivial differences:

* Dependencies in Cargo.toml are alphabetized
* The openssl-src version in Cargo.lock is slightly more recent

For context: we had closed https://github.com/fermyon/spin/pull/247 due to
security concerns, but https://github.com/fermyon/spin/issues/539 has forced us
to re-evaluate that decision and conclude that static linking is the least-bad
option available for the time being.


radu-matei:
❤️ 
> Dependencies in Cargo.toml are alphabetized

dicej:
Oh, I just remembered I was going to add `--features=openssl/vendored` to the GitHub `cargo build` action as well -- stand by for an update.

lann (approved):

itowlson:
I don't fully get features but will this _always_ statically link, or is static linking something we need to opt into at build time, or it is a default that we can opt out of at build time?

dicej:
> I don't fully get features but will this _always_ statically link, or is static linking something we need to opt into at build time, or it is a default that we can opt out of at build time?

See my comment about `--features=openssl/vendored` above.  It will still dynamically link by default without that flag.

fibonacci1729 (approved):
LGTM!

itowlson:
Sweet, that's what I thought - got confused because I hadn't seen the update at that time - thanks!

dicej:
On second thought, maybe the GitHub action config doesn't need to change.  @lann @vdice Does [this](https://github.com/fermyon/spin/blob/22b0fe6191d6d165c96b7f6bc0cb6bd5e4e0f79a/.github/workflows/release.yml#L99) mean the openssl/vendored feature will be enabled automatically?

lann:
Yes...though we really shouldn't have that turned on there...

dicej:
> Yes...though we really shouldn't have that turned on there...

Yeah, it feels a little blunt.  Will want to revisit if we add any other optional features.

dicej:
Per @vdice's excellent advice, I'm going to wait until the release action finishes on my fork and test the Linux binary before merging this.

lann:
I just realized this will enable the `openssl` dependency for all platforms. Not sure if that will have any net effect since we aren't using it directly... 🤔

dicej:
> I just realized this will enable the `openssl` dependency for all platforms. Not sure if that will have any net effect since we aren't using it directly... thinking

I just pushed a change to avoid that.  Based on the logs for the Windows build, openssl was indeed being downloaded but not actually built.  Hopefully the latest change will ensure it is neither downloaded nor built except on Linux.

lann:
> I just pushed a change to avoid that. Based on the logs for the Windows build, openssl was indeed being downloaded but not actually built. Hopefully the latest change will ensure it is neither downloaded nor built except on Linux.

Thanks. The mac build at least _was_ building openssl, which it normally shouldn't.

dicej:
Sorry for all the force-pushes -- I wanted to pull in https://github.com/fermyon/spin/pull/550, but forgot to pass `-S` to `git rebase`.

lann:
Oh right it's public now :laughing:

PR #547, opened 2022-06-08
implement `spin up -k`

### LINKED ISSUE #545: Feature: Support -k/insecure on spin up
There is a need to allow users to ignore cert errors on a `spin up` command, for example when invoking `spin up --bindle <bindle id> --bindle-url <bindle url>` wherein the bindle url is accessible via an https endpoint but cert errors may occur (and such errors may be acceptable to the user).

Note that Spin supports an equivalent flag for the `spin bindle push` command already.

bacongobbler:
closes #545

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -50,6 +50,15 @@ pub struct UpCommand {
         )]
     pub server: Option<String>,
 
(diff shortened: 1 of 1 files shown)


itowlson (approved):
LGTM - thanks!

PR #551, opened 2022-06-09
ci: Build Linux releases with Ubuntu 18.04

lann:
This fixes runtime glibc errors on 18.04.

Signed-off-by: Lann Martin <lann.martin@fermyon.com>

PR #552, opened 2022-06-09
Remove `--all-features` from build scripts

lann:
@dicej I wanted to get the flag out of `make test`, so I figured I'd just clean it up everywhere while I was at it.

@vdice Any good way to test release jobs?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,11 +77,11 @@ jobs:
           cargo clippy --workspace --all-targets --all-features -- -D warnings
 
       - name: Cargo Build
-        run: cargo build --workspace --all-targets --all-features
+        run: cargo build --workspace --all-targets
 
       - name: Cargo Test
         run:
-          cargo test --workspace --all-targets --all-features --no-fail-fast -- --nocapture
+          cargo test --workspace --all-targets --no-fail-fast -- --nocapture
         env:
           RUST_LOG: spin=trace
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,9 @@ jobs:
               os: "ubuntu-latest",
               arch: "amd64",
               extension: "",
-              extraArgs: "",
+              # Ubuntu 22.04 no longer ships libssl1.1, so we statically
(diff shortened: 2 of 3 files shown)


lann:
Will rebase after #546 merges.

dicej:
> > @vdice Any good way to test release jobs?
> 
> Yes, merge this branch to `main` on your fork an push. This will trigger the action.

But make sure Settings->Actions->General->Workflow permissions is set to "Read and write permissions" on your fork first.

lann:
Oh right it's public now :laughing:

lann:
Release builds passed here: https://github.com/lann/spin/actions/runs/2469517288 (those builds test this PR along with #551 since release builds take so long)

itowlson:
Should we keep the `--all-features` in the clippy and fmt stuff?  Just be more specific in the actual build?

lann:
> Should we keep the `--all-features` in the clippy and fmt stuff? Just be more specific in the actual build?

I'm not sure. It isn't ideal to to have to build openssl just for clippy. Since that is currently the only feature I think the answer _today_ is "no", but if we do add more features the answer could change. It might be that even then the right solution would be to list relevant features explicitly.

dicej:
I'd lean towards leaving `--all-features` in for clippy.  Clippy is just so valuable for code quality, and I don't want its warnings to start slipping past CI because we forgot to include a `--features` flag.

lann:
Alright, `--all-features` returns for `cargo clippy`

lann:
> BTW, do we really need to run `cargo clippy` separately for every platform in the matrix? Won't `--all-targets` cover everything if we only run it e.g. on Linux? Likewise for `cargo fmt`.

We don't for PR builds. `fmt` and `clippy` are guarded by `if: ${{ matrix.config.platformAgnosticChecks }}`, which is only set for the Linux job.

lann:
> I don't know if spin even has platform-specific code

It doesn't today. There are nice crates that abstract platform differences for most things.

dicej:
Now I'm wondering about `cargo test`.  I also don't want failing, feature-specific tests to slip by CI.  What do you think?

lann:
We found a better approach.

PR #556, opened 2022-06-10
Template version affinity

### LINKED ISSUE #555: Template version affinity
When you do a `spin templates list` and there are no templates, it suggests you run `spin templates install --git ...`.  This hits HEAD, which causes problems if you are on a release build and there have been breaking changes to the templating stuff in `main` since that release.

Our original thought was to change the suggestion to specify a branch.  This can be a bit tricky because it requires Spin to know it is a tagged release build, which is hard to do without contacting the server, which seems heavyweight for a message.

An alternative approach is to smartly default the branch.  That is, if the user runs `spin templates install --git ...`, we look for `spin/templates/<version>` instead of HEAD.  If that doesn't exist, we fall back to HEAD.  (Of course, if a user wants to use the HEAD templates from a tagged release, they can still write `spin templates install --branch HEAD`.)

This is nice because it allows third-party repos that contain templates to also mark version compatibility, though it does require coordination on their part.  I _think_ the fail-forward strategy mostly works correctly, though it's problematic if the third party doesn't backfill on a change.  (Consider the following: Bob tagged at 0.6. There is a breaking change in 0.8, which Bob updates and tags. If Alice is on 0.7, she will get HEAD, which is broken for her.)

**Important:** This requires us to update the version in Cargo.toml _immediately_ after doing a release rather than when we get to the next one.  (Otherwise we will keep installing the last released version of the templates, even when we are on a local or canary build.)  Are we okay with that?

Thanks to @lann for discussion around this!

itowlson:
Fixes #555.

This makes `spin template install --git` attempt to locate compatible templates instead of defaulting to HEAD.  The intended logic is as follows:

* If the user runs `install --git ... --branch ...` then the given branch or tag is used.  No inference occurs.
* If the user runs `install --git ...` without `--branch`:
  * If the repo _contains_ a tag `spin/templates/v1.2.3` (where `1.2.3` is the version of the Spin binary), that tag is used.
  * If _not_, then HEAD is used.

**NOTE:** #555 raises some difficulties and considerations for this logic.  We should not merge this PR without agreeing that this is a reasonable approach to take.

**CONFESSION:** This PR includes an unrelated change to clone the remote repo with depth 1 instead of hauling in all the history.  I can move this to a separate PR if preferred!

### DIFF
--- a/crates/templates/src/source.rs
+++ b/crates/templates/src/source.rs
@@ -8,6 +8,7 @@ use url::Url;
 use crate::directory::subdirectories;
 
 const TEMPLATE_SOURCE_DIR: &str = "templates";
+const TEMPLATE_VERSION_TAG_PREFIX: &str = "spin/templates/v";
 
 /// A source from which to install templates.
 #[derive(Debug)]
@@ -18,12 +19,7 @@ pub enum TemplateSource {
     ///
     /// Templates much be in a `/templates` directory under the root of the
     /// repository.
-    Git {
-        /// The URL of the Git repository from which to install templates.
-        url: Url,
-        /// The branch or tag from which to install templates; HEAD if omitted.
-        branch: Option<String>,
-    },
+    Git(GitTemplateSource),
     /// Install from a directory in the file system.
     ///
     /// Templates much be in a `/templates` directory under the specified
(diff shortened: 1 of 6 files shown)


crates/templates/src/source.rs
```diff
@@ -18,30 +18,42 @@ pub enum TemplateSource {
...
+    spin_version: String,
+}
+
 impl TemplateSource {
     /// Creates a `TemplateSource` referring to the specified Git repository
     /// and branch.
-    pub fn try_from_git(git_url: impl AsRef<str>, branch: &Option<String>) -> anyhow::Result<Self> {
+    pub fn try_from_git(
+        git_url: impl AsRef<str>,
+        branch: &Option<String>,
```
lann, line 46: Not really relevant to this PR but from our #rust discussion: since this function requires ownership of `branch` (except in the failure case, which probably isn't worth optimizing), this doesn't need to be a ref.

lann: Ditto `spin_version`: https://rust-lang.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

lann:
Would it be sensible to strip the semver patch version for this? It seems like "don't break templates" is a reasonable goal if we need to do a bugfix release at some point, especially if we need to do a security fix backport.

crates/templates/src/source.rs
```diff
@@ -86,20 +98,28 @@ impl LocalTemplateSource {
...
 
-    let url_str = url.as_str();
+    let url_str = git.url.as_str();
+
+    let actual_branch = match &git.branch {
+        Some(b) => Some(b.clone()),
+        None => version_matched_branch(url_str, &git.spin_version).await,
+    };
 
     let mut git = Command::new("git");
```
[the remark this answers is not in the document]
itowlson: Ouch, yes. Thanks for catching!

crates/templates/src/source.rs
itowlson: The trouble is that `branch` gets initialised from `self.branch` in the `Install` command, so we can't move it - we could clone on the caller side.  `spin_version` comes from a macro which evaluates to a `&'static str` so again would need to `to_owned` on the caller side.  I can do these, it makes no odds to me, but they feel like bothering the caller with an implementation detail (e.g. the caller doesn't care whether `TemplateSource` owns a `branch` option, or composes the git command objects and owns them, or what) - but let me know what you prefer, I'm happy either way.

itowlson:
@lann that's a good question.  I had expecting we would add something to the release CI to automatically apply the template and Go SDK tags, in which case it doesn't make much odds to us.  But it would reduce the churn for sure, which is an advantage for third party repos that likely have to tag manually.  Anyone else got views on this?

lann:
LGTM w/ or w/o version truncation.

fibonacci1729 (approved):

lann:
👍

PR #557, opened 2022-06-10
ref(deploy): allows user to re-deploy after changing version

michelleN:
This commit allows users to redeploy their app after changing the version specified in the spin.toml file. It simply deletes the app if it already exists and re-creates it and the relevant channel.

This PR is blocked by a change that needs to be made in the hippo-cli where we need to update the hippo-openapi dep. https://github.com/deislabs/hippo-cli/issues/135

The first route I tried was to simply keep the app and update the channel configuration but was unsuccessful in doing so. I'll need to debug why the hippo channel put api is not working as expected.

Co-authored-by: Ivan Towlson <ivan.towlson@fermyon.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1520,8 +1520,8 @@ dependencies = [
 
 [[package]]
 name = "hippo"
-version = "0.12.0"
-source = "git+https://github.com/deislabs/hippo-cli#5a7654e6fa9d063d934b078e5216a39ea2f91147"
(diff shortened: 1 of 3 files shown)


michelleN:
Also tried the route of bumping the revision in #561 but need some changes in hippo (listing channels not working properly) to make that route work.

fibonacci1729 (approved):

PR #558, opened 2022-06-10
Remove vendored-openssl feature

lann:
It turns out this can be replaced with `--features openssl/vendored`,
which allows us to toggle the feature independently from --all-features.

Use `openssl/vendored` for Linux builds to maintain static linking for
releases.


Cargo.toml
```diff
@@ -47,7 +44,8 @@ tracing-subscriber = { version = "0.3.7", features = [ "env-filter" ] }
 wasi-outbound-http = { path = "crates/outbound-http" }
 wasmtime = "0.34"
 
-[target.'cfg(target_os = "linux")'.dependencies]
+# This optional dependency isn't used directly, but must be here to allow
+# --features 'openssl/vendored', which is used for Linux releases.
```
dicej, line 48: When I said "remove the `[target.'cfg(target_os = "linux")'.dependencies]` section" in Slack, I meant remove the whole section, including the `openssl = { version = "0.10", optional = true }` line.  I don't know if that actually works, but I figured maybe Cargo would let you specify features on the command line for transitive dependencies as well as direct ones.  Could be totally wrong about that, though.

If I am wrong about that, might as well leave the `[target.'cfg(target_os = "linux")'.dependencies]` qualifier in, right?

lann: I was thinking it might be nice to allow for openssl on mac, but it looks like `--all-features` does enable the optional dependency which isn't great. I'll add it back.

lann: (and yes, the dependency is required to allow for `openssl/vendored`)

PR #560, opened 2022-06-10
update "deploy to hippo" documentation

bacongobbler:
Based on feedback/testing with recent changes to nomad-local-demo, spin,
and hippo.

Couple of notes on the changes:

1. There is no need to run a local version of Hippo - nomad-local-demo runs an instance for you on nomad
2. http://hippo.local.fermyon.link is set up with Traefik so that routes will resolve
3. No need for HTTPS or the `-k` flag because everything is over HTTP
4. The `nomad dashboard` lookup isn't necessary - the domain will be `http://<channel_name>.<app_name>.hippo.local.fermyon.link/hello`, which is routable via Traefik (see http://localhost:8081 for more details)

Signed-off-by: Matthew Fisher <matt.fisher@fermyon.com>

PR #563, opened 2022-06-10
doc: update deploying-to-hippo.md to reflect recent changes

dicej:
This addresses a few issues I ran into trying to follow these instructions:

* Node is now a prerequisite
* Several of the prerequisites must be installed in `sudo`'s PATH, so we call that out now
* Hippo listens on HTTP by default
* Bindle listens on port 80 by default

### DIFF
--- a/docs/content/deploying-to-hippo.md
+++ b/docs/content/deploying-to-hippo.md
@@ -27,10 +27,15 @@ coming soon.
 - [Vault >= v1.9.3](https://www.vaultproject.io/)
 - [Traefik >= 2.6.1](https://github.com/traefik/traefik/releases)
 - [Dotnet 6.0 CLI](https://dotnet.microsoft.com/en-us/download)
+- [Node >= 16.10](https://nodejs.org/en/)
+
+*Please note:* Nomad, Spin, Bindle, and Traefik must be installed such that they
+appear in the root user's `PATH` since Nomad is run using `sudo`. The
+`run_servers.sh` script used below may complain if this is not the case.
 
 ### Getting Set Up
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hey @dicej. Some of these changes conflict with https://github.com/fermyon/spin/pull/560. I'll try to explain some of the recent developments here...

nomad-local-demo [deploys Hippo for you](https://github.com/fermyon/nomad-local-demo/blob/2f73d18e2e83b5a3f8038a2d5b008ef9e383aa50/run_servers.sh#L92-L105) within Nomad. To that end, the user should not need to fetch, compile, and run Hippo. Node.js nor `dotnet` should be required on the user's system UNLESS they wanted to run a local copy (and in that case, they should be following [Hippo's developer guide](https://docs.hippofactory.dev/developers/developer-guide/)). I've reflected these changes in #560.

> * Bindle listens on port 80 by default

To clarify this point, Bindle does not listen on port 80. Bindle [listens on port 8080](https://github.com/fermyon/nomad-local-demo/blob/2f73d18e2e83b5a3f8038a2d5b008ef9e383aa50/job/bindle.nomad#L16). Traefik [listens on port 80](https://github.com/fermyon/nomad-local-demo/blob/2f73d18e2e83b5a3f8038a2d5b008ef9e383aa50/job/traefik.nomad#L11). Bindle [tells Traefik to forward requests](https://github.com/fermyon/nomad-local-demo/blob/2f73d18e2e83b5a3f8038a2d5b008ef9e383aa50/job/bindle.nomad#L25-L26) from `bindle.local.fermyon.link` to Bindle [using host header lookups](https://doc.traefik.io/traefik/getting-started/concepts/).

I'd suggest we start by converging on #560 first, then see what falls out from there. What do you think?

dicej:
@bacongobbler That works for me.  I'll close this for now.

PR #568, opened 2022-06-14
Print HTTP routes on deploy

### LINKED ISSUE #566: return routes available in spin deploy output
maybe we could also add a link to the hippo dashboard to the output as well.

itowlson:
Fixes #566.

@michelleN I'm afraid I can't yet test this because I can't get the Hippo-Nomad stuff working correctly yet.  I will try to get that sorted in the morning.  If you have time do feel free to give this a quick test - I _think_ it is right but am not sure in exactly what form Hippo returns the channel domain, so that at least is quite likely to be wrong!  Sorry!

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -3,8 +3,9 @@ use bindle::Id;
 use clap::Parser;
 use hippo::{Client, ConnectionInfo};
 use hippo_openapi::models::ChannelRevisionSelectionStrategy;
+use spin_http_engine::routes::RoutePattern;
 use spin_loader::local::config::RawAppManifestAnyVersion;
-use spin_manifest::HttpTriggerConfiguration;
+use spin_manifest::{HttpTriggerConfiguration, TriggerConfig};
 use std::path::PathBuf;
 
 use crate::opts::*;
@@ -95,7 +96,6 @@ impl DeployCommand {
     pub async fn run(self) -> Result<()> {
         let cfg_any = spin_loader::local::raw_manifest_from_file(&self.app).await?;
         let RawAppManifestAnyVersion::V1(cfg) = cfg_any;
-        let is_http_config = HttpTriggerConfiguration::try_from(cfg.info.trigger).is_ok();
 
         let bindle_id = self.create_and_push_bindle().await?;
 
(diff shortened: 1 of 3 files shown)


fibonacci1729:
Seems to work! Just verified against the local installer:

```console
$ spin deploy
Successfully deployed application spin-hello-world version 1.0.0!
Application is running at spin-hello-world.spin-hello-world.local.fermyon.link
Available Routes:
  hello: http://spin-hello-world.spin-hello-world.local.fermyon.link/hello
    A simple component that returns hello.
```

src/commands/deploy.rs
```diff
@@ -203,3 +204,24 @@ impl DeployCommand {
...
+    if cfg.components.is_empty() {
+        return;
+    }
+
+    println!("Available Routes:");
+    for component in &cfg.components {
+        if let TriggerConfig::Http(http_cfg) = &component.trigger {
+            let route = RoutePattern::from(base, &http_cfg.route);
+            println!("  {}: http://{}{}", component.id, address, route);
+            if let Some(description) = &component.description {
```
michelleN, line 222: I think we removing the description might make the output cleaner.

src/commands/deploy.rs
```diff
@@ -153,11 +153,12 @@ impl DeployCommand {
             name.clone(),
             bindle_id.version_string()
         );
-        if is_http_config {
+        if let Ok(http_config) = HttpTriggerConfiguration::try_from(cfg.info.trigger.clone()) {
             let channel = Client::get_channel_by_id(&hippo_client, &channel_id)
                 .await
                 .context("Problem getting channel by id")?;
             println!("Application is running at {}", channel.domain);
```
michelleN, line 160: What do you think about removing this line entirely?

michelleN (approved):
I just tested and it works! I have a few minor comments. Thank you!
```
examples/http-rust [itowlson-deploy-print-routes] $ spin deploy
Successfully deployed application spin-hello-world version 1.0.0!
Application is running at spin-hello-world.spin-hello-world.local.fermyon.link
Available Routes:
  hello: http://spin-hello-world.spin-hello-world.local.fermyon.link/hello
    A simple component that returns hello.
```

src/commands/deploy.rs
itowlson: I don't think we can.  We need to type the trigger/config so that we can access the `.base` field.

src/commands/deploy.rs
itowlson: I did this for consistency with the local printing of routes but can remove it if folks prefer!

src/commands/deploy.rs
lann: Perhaps move it into an `else` clause of this `if`?

itowlson: Oh, duh, I misread which line Michelle was referring to!

lann (approved):

src/commands/deploy.rs
lann: Yeah if we don't like the format it seems reasonable to change both (in a separate PR).

michelleN: sounds good

PR #570, opened 2022-06-14
Enable re-deploy without changing app version

michelleN:


src/commands/deploy.rs
```diff
@@ -1,13 +1,20 @@
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Context, Result};
 use bindle::Id;
 use clap::Parser;
+use data_encoding::HEXLOWER;
 use hippo::{Client, ConnectionInfo};
 use hippo_openapi::models::ChannelRevisionSelectionStrategy;
-use spin_loader::local::config::RawAppManifestAnyVersion;
+use ring::digest::{Digest, SHA256};
```
lann, line 7: Can we switch to the `sha2` crate here? We don't currently have a direct dependency on `ring` but we do use `sha2` in a few places.

src/commands/deploy.rs
```diff
@@ -89,15 +96,41 @@ pub struct DeployCommand {
...
+    /// Build metadata to append to the bindle version
+    #[clap(
+        name = BUILDINFO_OPT,
+        long = "buildinfo",
+        parse(try_from_str = parse_buildinfo),
+    )]
+    pub buildinfo: Option<BuildMetadata>,
+
+    /// Re-deploy bindle if it already exists in bindle server
+    #[clap(short = 'r', long = "re-deploy")]
```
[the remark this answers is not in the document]
fibonacci1729: Updated

src/commands/deploy.rs
fibonacci1729: Updated

src/commands/deploy.rs
```diff
@@ -163,6 +194,31 @@ impl DeployCommand {
...
+                    let _ = copy(&mut r, &mut sha256)?;
+                }
+                config::RawModuleSource::Bindle(_b) => {}
+            }
+        }
+
+        let mut r = File::open(&self.app)?;
+        let _ = copy(&mut r, &mut sha256)?;
+
+        let mut final_digest = format!("q{:x}", sha256.finalize());
```
itowlson, line 213: I think this could produce a repeat result if the `spin.toml` and the modules were all the same, but the set of assets had changed e.g. a component still has the same Wasm module and still has `*.jpg` for its files but the user has created a new `.jpg` file in the source directory...?

fibonacci1729: We addressed this by using `assets::collect` to write the digests of component assets into the build info. Verified using the local installer.

lann (approved):

src/commands/deploy.rs
```diff
@@ -90,14 +95,39 @@ pub struct DeployCommand {
...
+
+    /// Build metadata to append to the bindle version
+    #[clap(
+        name = BUILDINFO_OPT,
+        long = "buildinfo",
+        parse(try_from_str = parse_buildinfo),
+    )]
+    pub buildinfo: Option<BuildMetadata>,
+
+    /// Re-deploy bindle if it already exists in bindle server
```
itowlson, line 111: Is this doc string accurate?  It looks like all the `redeploy` option does is suppress the error report if the bindle ID already exists.  If I make a change to my application and do a `deploy --redeploy`, it won't actually upload the changes will it?  (Forgive me if I'm wrong, I'm having to go on just reading the code so may be missing something!)

(sorry wrote this yesterday but forgot to push the button)

itowlson (commented):
sorry wrote a comment yesterday but forgot to commit it

src/commands/deploy.rs
itowlson: The bindle publisher already collects and hashes all the assets so we could do this as part of that pass rather than revisiting all the files again.  But we can look at whether that's worth doing after the 0.3 crunch. - it's definitely not needed.

src/commands/deploy.rs
lann: My read was that it allows `create_and_push_bindle` to succeed on duplicate bindle rather than failing and short-circuiting the deploy.

PR #575, opened 2022-06-15
ref(deploy): update redeploy flag name

michelleN:
+ redeploy -> deploy-existing-bindle
+ updates error message for existing bindle
+ adds env variable SPIN_DEPLOY_NO_BUILDINFO
which enables folks using a version number without
hash.

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -97,7 +97,11 @@ pub struct DeployCommand {
     pub hippo_password: String,
 
     /// Disable attaching buildinfo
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

itowlson (approved):
Thanks heaps @michelleN, this makes it much clearer (at least to me) what the behaviour is.  Appreciate you taking the time to make this late change.

radu-matei (approved):
Thanks, LGTM!

Issue #362, opened 2022-04-15
Publish crates to crates.io


Mossaka:
The spin engine and triggers are highly reusuable crates. It will be great to publish them to crates.io so that we can use spin engine as a SDK.

radu-matei:
ref #316

melissaklein24:
We now have `spin-sdk` and `spin-macro` on crates.io.  Are there other crates you would like to see on crates.io?

Mossaka:
Well, that's fine. I was looking for all of the spin crates to be published but it's not a high priority thing.

PR #577, opened 2022-06-16
Update quickstart.md to include `cargo install --path .` installation option

joepio:
Installing spin and adding it to path can be a little easier if the user already has `cargo` installed. So I added the steps to the quickstart.

Ideally, users could simply `cargo install spin-cli` but that requires publishing to crates.io #362


vdice (approved):
LGTM, thank you for the added docs!

PR #578, opened 2022-06-16
Give hint in `spin new` for `spin templates list`

joepio:
I was using `spin` for the first time, and was kind of frustrated that I didn't know what to fill in the template field. Adding this could help a little.


vdice (approved):
Very helpful addition, LGTM!

PR #587, opened 2022-06-16
doc: add CHANGELOG.md

dicej:
This summarizes the releases made so far, stealing shamelessly from
https://github.com/fermyon/spin/releases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,25 @@
+# Changelog
+
+## [0.3.0] - 2022-06-16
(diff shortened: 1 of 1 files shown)


vdice (approved):

dicej:
Given that this currently just duplicates info on the GitHub releases page, I don't think this is adding a lot of value in its current state.  I'll close this, and we can revisit if/when we have a clearer idea of what purpose a change log might serve beyond what the releases page already provides.

PR #589, opened 2022-06-17
Expose Spin Application Directory to WASI FS

ecumene:
Ref: https://github.com/fermyon/spin/pull/542
Ref: https://github.com/fermyon/spin/issues/528

 - [ ] Handle `log-dir` case

### DIFF
--- a/crates/engine/src/lib.rs
+++ b/crates/engine/src/lib.rs
@@ -7,7 +7,12 @@ pub mod host_component;
 /// Input / Output redirects.
 pub mod io;
 
-use std::{collections::HashMap, io::Write, path::PathBuf, sync::Arc};
+use std::{
+    collections::HashMap,
+    io::Write,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
 use anyhow::{bail, Context, Result};
 use host_component::{HostComponent, HostComponents, HostComponentsState};
@@ -246,6 +251,17 @@ impl<T: Default> ExecutionContext<T> {
         Ok((store, instance))
     }
 
+    /// Sanitizes the component's label and returns the component's directory in
+    /// Spin home or simply just the component's directory depending on if a
+    /// home directory is present
+    fn app_dir(&self) -> PathBuf {
+        let sanitized_label = sanitize(&self.config.label);
+        match dirs::home_dir() {
+            Some(h) => h.join(SPIN_HOME).join(&sanitized_label),
+            None => PathBuf::from(&sanitized_label),
+        }
+    }
+
     /// Save logs for a given component in the log directory on the host
     pub fn save_output_to_logs(
         &self,
@@ -254,15 +270,11 @@ impl<T: Default> ExecutionContext<T> {
         save_stdout: bool,
         save_stderr: bool,
     ) -> Result<()> {
-        let sanitized_label = sanitize(&self.config.label);
         let sanitized_component_name = sanitize(&component);
 
         let log_dir = match &self.config.log_dir {
             Some(l) => l.clone(),
-            None => match dirs::home_dir() {
-                Some(h) => h.join(SPIN_HOME).join(&sanitized_label).join("logs"),
-                None => PathBuf::from(&sanitized_label).join("logs"),
-            },
+            None => self.app_dir().join("log"),
         };
 
         let stdout_filename = log_dir.join(sanitize(format!(
@@ -328,6 +340,16 @@ impl<T: Default> ExecutionContext<T> {
             let host = dir.host;
             wasi_ctx =
                 wasi_ctx.preopened_dir(Dir::open_ambient_dir(host, ambient_authority())?, guest)?;
+        }
+
+        if component.core.wasm.enable_monitoring {
+            let app_dir = self.app_dir();
+            std::fs::create_dir_all(&app_dir)?;
+            log::trace!("Pre-opening app dir {:?}", app_dir);
+            wasi_ctx = wasi_ctx.preopened_dir(
+                Dir::open_ambient_dir(&app_dir, ambient_authority())?,
+                Path::new("/.spin/"),
+            )?;
         }
 
         if let Some(resolver) = &self.config.config_resolver {
--- a/crates/manifest/src/lib.rs
+++ b/crates/manifest/src/lib.rs
@@ -158,6 +158,8 @@ pub struct WasmConfig {
     pub mounts: Vec<DirectoryMount>,
     /// Optional list of HTTP hosts the component is allowed to connect.
     pub allowed_http_hosts: Vec<String>,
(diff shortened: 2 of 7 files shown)


crates/engine/src/lib.rs
```diff
@@ -330,6 +336,18 @@ impl<T: Default> ExecutionContext<T> {
...
 
+        if component.core.wasm.enable_monitoring {
+            let spin_dir = match dirs::home_dir() {
+                Some(h) => h.join(SPIN_HOME),
+                None => env::current_dir()?,
+            };
+
+            wasi_ctx = wasi_ctx.preopened_dir(
+                Dir::open_ambient_dir(&spin_dir, ambient_authority())?,
+                Path::new("/.spin/"),
```
ecumene, line 347: How should this be configured?

crates/engine/src/lib.rs
```diff
@@ -330,6 +336,18 @@ impl<T: Default> ExecutionContext<T> {
...
         }
 
+        if component.core.wasm.enable_monitoring {
+            let spin_dir = match dirs::home_dir() {
+                Some(h) => h.join(SPIN_HOME),
+                None => env::current_dir()?,
+            };
+
+            wasi_ctx = wasi_ctx.preopened_dir(
+                Dir::open_ambient_dir(&spin_dir, ambient_authority())?,
```
lann, line 346: This should only mount the current application within ~/.spin, right?

crates/engine/src/lib.rs
lann: I think initially it doesn't need to be.

crates/engine/src/lib.rs
lann: ...hmmm now that I think about it, maybe I had a different assumption than you here :)

crates/engine/src/lib.rs
```diff
@@ -330,6 +335,21 @@ impl<T: Default> ExecutionContext<T> {
...
         }
 
+        if component.core.wasm.enable_monitoring {
+            let sanitized_label = sanitize(&self.config.label);
+            let app_dir = match dirs::home_dir() {
+                Some(h) => h.join(SPIN_HOME).join(&sanitized_label),
+                None => PathBuf::from(&sanitized_label),
+            };
+
+            std::fs::create_dir_all(&app_dir)?;
```
lann, line 345: It would be nice to dedupe this w/ `save_output_to_logs` into an e.g. `fn ensure_app_dir(label: &str) -> Result<PathBuf>`.

I'd be fine doing that in a separate PR though.

lann (commented):
Implementation looks good! I'm not sure how I feel about adding another example project; we have so many as it is...

crates/engine/src/lib.rs
ecumene: I can do that here, certainly. :)

crates/manifest/src/lib.rs
```diff
@@ -158,6 +158,8 @@ pub struct WasmConfig {
     pub mounts: Vec<DirectoryMount>,
     /// Optional list of HTTP hosts the component is allowed to connect.
     pub allowed_http_hosts: Vec<String>,
+    /// Flag to enable access to the Spin home directory
+    pub enable_monitoring: bool,
```
itowlson, line 162: The name seems out of sync with the documentation...?  (I don't fully understand the implementation though so I'm probably missing something!)

ecumene: No you're not missing anything, I've been too lazy and haven't updated the SIP

ecumene: All updated now :) It should be nice and clear.

crates/engine/src/lib.rs
```diff
@@ -254,15 +271,16 @@ impl<T: Default> ExecutionContext<T> {
...
         let log_dir = match &self.config.log_dir {
             Some(l) => l.clone(),
-            None => match dirs::home_dir() {
-                Some(h) => h.join(SPIN_HOME).join(&sanitized_label).join("logs"),
-                None => PathBuf::from(&sanitized_label).join("logs"),
-            },
+            None => {
+                let app_dir = self.ensure_app_dir()?;
+                let log_dir = app_dir.join("log");
+                std::fs::create_dir_all(&log_dir)?;
```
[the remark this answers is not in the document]
ecumene: Good catch. I think I moved that like around a couple of times then put it back in the wrong place.

lann:
LGTM; since we didn't get much discussion on the SIP I'd like others to review.

dicej (approved):
LGTM.  I nitpicked passing an explicit reference to the `trace!` macro, but it's certainly not a blocker.

crates/engine/src/lib.rs
```diff
@@ -330,6 +342,16 @@ impl<T: Default> ExecutionContext<T> {
                 wasi_ctx.preopened_dir(Dir::open_ambient_dir(host, ambient_authority())?, guest)?;
         }
 
+        if component.core.wasm.enable_monitoring {
+            let app_dir = self.app_dir();
+            std::fs::create_dir_all(&app_dir)?;
+            log::trace!("Pre-opening app dir {:?}", &app_dir);
```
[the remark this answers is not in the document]
ecumene: Good point!

crates/manifest/src/lib.rs
itowlson: @ecumene just pinging on this, I'm still seeing a mismatch between the flag name and flag doc string...?

ecumene: I didn't see your reply:

I bounced the name `monitoring` back and forth with @lann. I suggested it but I don't think it's an accurate description either. I kept it hoping someone would come up with something better.

The flag name should be changed, I agree. How about something like `enable_home` or `mount_home`? I think someone with more experience than I could chime in

itowlson: `mount_home_dir` or `access_home_dir` perhaps?  I'm going by the doc string, though, I don't really understand the feature, so I could be way off beam!  From the screenshot... this is allowing the hosted modules access to each others' asset directories?  In which case `access_other_component_dirs` might be clearer than "home"...?

itowlson: (Many apologies if I am being obtuse here... I haven't followed enough and should have avoided suggesting things from a position of ignorance!)

lann: The current iteration of this feature has a very straightforward effect: the `~/.spin/<your-app>/` dir is mounted in the component's WASI env as `/.spin/`. Currently the only useful files we expect there are the `logs/<component>_std{err,out}.txt`s, but anything that Spin wants to put in that dir would be available, which is why I'm fine with the more general "monitoring" term.

fibonacci1729 (approved):

PR #592, opened 2022-06-17
ref(deploy): add correct scheme to routes

michelleN:
+ At the moment, the default scheme is http so if
you deploy Hippo to AWS for example with ssl enabled
the output would show http as the scheme for the
available routes instead of https.
+ This change makes it so that the routes that get
displayed in the output of spin deploy match the
scheme of the hippo url.
+ In the future, it'd be nice if we could get the routes
with the scheme from the Hippo API.

Co-authored-by: Brian Hardock <brian.hardock@fermyon.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3539,6 +3539,7 @@ dependencies = [
  "tracing",
  "tracing-futures",
  "tracing-subscriber 0.3.9",
+ "url 2.2.2",
  "vergen",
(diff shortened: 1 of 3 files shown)


vdice (approved):
This worked well for me!

itowlson:
Oh awesome, thanks for figuring this out Michelle and Brian, it had been bugging me!

PR #593, opened 2022-06-18
fix(deploy): use absolute path to load component source

### LINKED ISSUE #590: Deploy: uninformative error if run from another directory
```
~/github/fermyon-installer/aws/terraform [main|✔]
12:16 $ ls ~/github/spin/examples/http-rust/
Cargo.lock  Cargo.toml  spin.toml  src  target
~/github/fermyon-installer/aws/terraform [main|✔]
12:17 $ ~/github/spin/target/debug/spin deploy -f ~/github/spin/examples/http-rust/spin.toml
Error: No such file or directory (os error 2)
```

I added some logging and the issue appears to be that it is using the relative path to the module source from the `spin.toml` and trying to resolve it relative to the _current_ directory rather than the application directory.  As I was in the terraform directory, this didn't work.

However, it was hard to diagnose the underlying issue because the error message didn't say _which_ file or directory was missing.  This is a vexing habit of the Rust stdlib!

FrankYang0529:
fixes #590

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -200,11 +200,19 @@ impl DeployCommand {
 
     async fn compute_buildinfo(&self, cfg: &RawAppManifest) -> Result<BuildMetadata> {
         let mut sha256 = Sha256::new();
+        let app_folder = self.app.parent().with_context(|| {
+            anyhow!(
+                "Cannot get a parent directory of manifest file {}",
+                &self.app.display()
+            )
+        })?;
 
         for x in cfg.components.iter() {
             match &x.source {
                 config::RawModuleSource::FileReference(p) => {
-                    let mut r = File::open(p)?;
+                    let full_path = app_folder.join(p);
+                    let mut r = File::open(&full_path)
+                        .with_context(|| anyhow!("Cannot open file {}", &full_path.display()))?;
                     copy(&mut r, &mut sha256)?;
                 }
                 config::RawModuleSource::Bindle(_b) => {}


michelleN (approved):
Thank you @FrankYang0529 for fixing this bug. Manually tested. Also was nice to see the `push` method on `PathBuf` being used. I usually use `join` on `Path` so I got to learn something new. 🎉 

We might also want to absolutize the app dir reference as a follow up.

Would you mind also signing your commit so we can merge?

michelleN:
@FrankYang0529  https://spin.fermyon.dev/contributing/

FrankYang0529:
> We might also want to absolutize the app dir reference as a follow up.

Do you mean also put app_folder in `DeployCommand`? If yes, I can help to do the follow up.

> Would you mind also signing your commit so we can merge?

I sign it. Thank you.

src/commands/deploy.rs
```diff
@@ -200,11 +200,15 @@ impl DeployCommand {
...
     async fn compute_buildinfo(&self, cfg: &RawAppManifest) -> Result<BuildMetadata> {
         let mut sha256 = Sha256::new();
+        let mut app_folder = PathBuf::from(&self.app);
+        app_folder.pop();
 
         for x in cfg.components.iter() {
             match &x.source {
                 config::RawModuleSource::FileReference(p) => {
-                    let mut r = File::open(p)?;
+                    let mut absolute_path = PathBuf::from(&app_folder);
```
itowlson, line 209: Nit: this isn't necessarily an absolute path.  `full_path` maybe?

src/commands/deploy.rs
```diff
@@ -200,11 +200,15 @@ impl DeployCommand {
...
+                    absolute_path.push(p);
```
itowlson, line 210: Nit: rather than separately creating a `PathBuf` and appending to it, consider `let full_path = app_folder.join(p);`

src/commands/deploy.rs
```diff
@@ -200,11 +200,15 @@ impl DeployCommand {
...
+                    let mut r = File::open(absolute_path)?;
```
itowlson, line 211: Please add a `with_context`, specifically to include the file name - that way the user gets a more informative error message if anything goes wrong - unfortunately the stdlib errors are a bit woeful for this.

src/commands/deploy.rs
```diff
@@ -200,11 +200,15 @@ impl DeployCommand {
 
     async fn compute_buildinfo(&self, cfg: &RawAppManifest) -> Result<BuildMetadata> {
         let mut sha256 = Sha256::new();
+        let mut app_folder = PathBuf::from(&self.app);
+        app_folder.pop();
```
itowlson, line 204: Super nit: elsewhere we have used `parent()` for this to avoid the two-step and to pick up edge case errors (e.g. `let app_folder = &self.app.parent().with_context(...)?;`.

itowlson (changes requested):
This looks good - thanks!  Comments are mostly minor nits; the only important thing is that when we try to open a file it's really important to provide a `with_context` with the file name because the stdlib error gives you _no clue at all_ which file the operation has failed on - this was the issue I hit in #590!  Thanks again!

src/commands/deploy.rs
FrankYang0529: Updated it. Thank you.

src/commands/deploy.rs
FrankYang0529: Updated it. Thank you.

src/commands/deploy.rs
FrankYang0529: Updated it. Thanks for the suggestion! I am new to Rust and I was wondering how to show error messages with more content. I know how to do it now 👍

src/commands/deploy.rs
FrankYang0529: Updated it. Thank you.

itowlson (approved):
LGTM - thanks for the fixes!

PR #607, opened 2022-06-28
Add deploy integration test

### LINKED ISSUE #583: e2e test for spin deploy

FrankYang0529:
resolves #583

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,8 +122,9 @@ jobs:
         run: cargo build --workspace --all-targets --all-features ${{ matrix.config.extraArgs }}
 
       - name: Cargo Test
-        run:
-          cargo test --workspace --all-targets --all-features --no-fail-fast -- --nocapture
+        run: |
+          make test-unit
+          make test-integration
         env:
           RUST_LOG: spin=trace
 
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,27 @@
 LOG_LEVEL ?= spin=trace
 CERT_NAME ?= local
 SPIN_DOC_NAME ?= new-doc.md
-NOMAD := nomad
-ifeq ($(OS),Windows_NT)
-	NOMAD = nomad.exe
-endif
 
 .PHONY: build
 build:
 	cargo build --release
 
 .PHONY: test
-test: nomad-version
-	RUST_LOG=$(LOG_LEVEL) cargo test --all --no-fail-fast -- --nocapture --include-ignored
+test: lint test-unit test-integration
+
+.PHONY: lint 
+lint:
 	cargo clippy --all-targets --all-features -- -D warnings
 	cargo fmt --all -- --check
+
+.PHONY: test-unit
+test-unit:
+	RUST_LOG=$(LOG_LEVEL) cargo test --no-fail-fast -- --skip integration_tests --nocapture --include-ignored
+
+.PHONY: test-integration
+test-integration:
+	RUST_LOG=$(LOG_LEVEL) cargo test --test integration --no-fail-fast  -- integration_tests::test_dependencies --nocapture
+	RUST_LOG=$(LOG_LEVEL) cargo test --test integration --no-fail-fast -- --skip integration_tests::test_dependencies --nocapture --include-ignored
 
 .PHONY: test-sdk-go
 test-sdk-go:
@@ -36,7 +43,3 @@ doc:
 .PHONY: check-content
 check-content:
 	cd docs && bart check content/* && bart check content/**/*
-
-.PHONY: nomad-version
-nomad-version:
-	$(NOMAD) --version
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -15,6 +15,7 @@ mod integration_tests {
     };
     use tempfile::TempDir;
     use tokio::{net::TcpStream, time::sleep};
+    use which::which;
 
     const RUST_HTTP_INTEGRATION_TEST: &str = "tests/http/simple-spin-rust";
     const RUST_HTTP_INTEGRATION_TEST_REF: &str = "spin-hello-world/1.0.0";
@@ -29,13 +30,30 @@ mod integration_tests {
 
     const SPIN_BINARY: &str = "./target/debug/spin";
     const BINDLE_SERVER_BINARY: &str = "bindle-server";
+    const NOMAD_BINARY: &str = "nomad";
+    const HIPPO_BINARY: &str = "Hippo.Web";
 
     const BINDLE_SERVER_PATH_ENV: &str = "SPIN_TEST_BINDLE_SERVER_PATH";
     const BINDLE_SERVER_BASIC_AUTH_HTPASSWD_FILE: &str = "tests/http/htpasswd";
     const BINDLE_SERVER_BASIC_AUTH_USER: &str = "bindle-user";
     const BINDLE_SERVER_BASIC_AUTH_PASSWORD: &str = "topsecret";
 
+    const HIPPO_BASIC_AUTH_USER: &str = "hippo-user";
+    const HIPPO_BASIC_AUTH_PASSWORD: &str = "topsecret";
+
     // This assumes all tests have been previously compiled by the top-level build script.
+
+    #[tokio::test]
+    async fn test_dependencies() -> Result<()> {
+        which(get_process(BINDLE_SERVER_BINARY))
+            .with_context(|| format!("Can't find {}", get_process(BINDLE_SERVER_BINARY)))?;
+        which(get_process(NOMAD_BINARY))
+            .with_context(|| format!("Can't find {}", get_process(NOMAD_BINARY)))?;
+        which(get_process(HIPPO_BINARY))
+            .with_context(|| format!("Can't find {}", get_process(HIPPO_BINARY)))?;
+
+        Ok(())
+    }
 
     #[tokio::test]
     async fn test_simple_rust_local() -> Result<()> {
@@ -269,6 +287,44 @@ mod integration_tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_spin_deploy() -> Result<()> {
+        // start the Bindle registry.
+        let config = BindleTestControllerConfig {
+            basic_auth_enabled: false,
+        };
+        let _nomad = NomadTestController::new().await?;
+        let bindle = BindleTestController::new(config).await?;
+        let hippo = HippoTestController::new(&bindle.url).await?;
+
+        // push the application to the registry using the Spin CLI.
+        run(
+            vec![
+                SPIN_BINARY,
+                "deploy",
+                "--file",
+                &format!(
+                    "{}/{}",
+                    RUST_HTTP_HEADERS_ENV_ROUTES_TEST, DEFAULT_MANIFEST_LOCATION
+                ),
+                "--bindle-server",
+                &bindle.url,
+                "--hippo-server",
+                &hippo.url,
+                "--hippo-username",
+                HIPPO_BASIC_AUTH_USER,
+                "--hippo-password",
+                HIPPO_BASIC_AUTH_PASSWORD,
+            ],
+            None,
+        )?;
+
+        let apps_vm = hippo.client.list_apps().await?;
+        assert_eq!(apps_vm.apps.len(), 1, "hippo apps: {apps_vm:?}");
+
+        Ok(())
+    }
+
     async fn verify_headers(
         s: &SpinTestController,
         absolute_uri: &str,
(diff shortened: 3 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -28,10 +28,15 @@ jobs:
...
               pathInBindleArchive: "bindle-server",
+              nomadUrl: "https://releases.hashicorp.com/nomad/1.3.1/nomad_1.3.1_linux_amd64.zip",
+              nomadBinary: "nomad",
+              pathInNomadArchive: "nomad",
+              hippoUrl: "https://github.com/deislabs/hippo/releases/download/v0.17.0/hippo-server-linux-x64.tar.gz",
               wasmtimeUrl: "https://github.com/bytecodealliance/wasmtime/releases/download/v0.36.0/wasmtime-v0.36.0-x86_64-linux.tar.xz",
               wasmtimeBinary: "wasmtime",
               pathInWasmtimeArchive: "wasmtime-v0.36.0-x86_64-linux/wasmtime",
               platformAgnosticChecks: true,
+              isNotWindows: true
```
vdice, line 39: Can we use the pre-existing `isWindows` key instead?  eg `isWindows: false` here and below?

tests/integration.rs
```diff
@@ -29,12 +29,17 @@ mod integration_tests {
 
     const SPIN_BINARY: &str = "./target/debug/spin";
     const BINDLE_SERVER_BINARY: &str = "bindle-server";
+    const NOMAD_BINARY: &str = "nomad";
```
vdice, line 32: With these two added dependencies for running `make test` locally, do you have thoughts on how best to alert users on their necessity?  Perhaps some combination of docs, obvious error/failure messages when a dep is missing and/or checks in the Makefile (e.g. as a first task when running `make test`, check for needed binaries and return non-zero with a helpful error message if any are missing)?

tests/integration.rs
```diff
@@ -482,6 +525,99 @@ mod integration_tests {
...
+            let url = format!("http://127.0.0.1:{}", get_random_port()?);
+
+            let mut hippo_handle = Command::new(get_process(HIPPO_BINARY))
+                .env("ASPNETCORE_URLS", &url)
+                .env("Nomad__Driver", "raw_exec")
+                .env("Nomad__Datacenters__0", "dc1")
+                .env("Database__Driver", "sqlite")
+                .env(
+                    "ConnectionStrings__Database",
+                    "Data Source=hippo.db;Cache=Shared",
```
vdice, line 573: On repeated runs of this integration test, I'm seeing:
```
      Executed endpoint 'Hippo.Web.Api.AccountController.Register (Hippo.Web)'
Error: One or more validation errors occurred. {"UserName": ["The specified username already exists."]}
thread 'integration_tests::test_spin_deploy' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/test/src/lib.rs:186:5
test integration_tests::test_spin_deploy ... FAILED

failures:

failures:
    integration_tests::test_spin_deploy

test result: FAILED. 9 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.25s
```

Do we need to also be sure to clean up or delete `hippo.db` after (or prior to) each run?

vdice (commented):
This looks great, @FrankYang0529!  Had a few initial comments and some notes from running the added test locally.  Once I had dependencies installed, I had a passing run.  For subsequent runs, we may need to look into cleaning up Hippo's db...

Also, have you been able to run the workflow on your fork with success?

tests/integration.rs
FrankYang0529: Yes, I also wonder about this. After I read https://github.com/fermyon/installer/issues/36#issuecomment-1155234416, I realize that we can use `inmemory` for testing, so we can rerun it.

.github/workflows/build.yml
FrankYang0529: I update the if condition in steps, so we don't need to introduce both `isWindows` and `isNotWindows`.

tests/http/simple-spin-rust/spin.toml
```diff
@@ -5,14 +5,11 @@ name = "spin-hello-world"
...
-
 [[component]]
 id = "hello"
 source = "target/wasm32-wasi/release/spinhelloworld.wasm"
 files = [ { source = "assets", destination = "/" } ]
 [component.trigger]
 route = "/hello/..."
 [component.config]
-message = "I'm a {{object}}"
+message = "I'm a teapot"
```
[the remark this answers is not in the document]
FrankYang0529: This config still works for `spin up`, but not for `spin deploy`. The issue toml `ToModel` is in hippo.

```
fail: Hippo.Application.Revisions.Commands.ImportRevisionsCommand[0]
      Hippo Request: Unhandled Exception for Request ImportRevisionsCommand Hippo.Application.Revisions.Commands.ImportRevisionsCommand
      Tomlyn.TomlException: (4,2) : error : Unable to set the property config on object type Hippo.Application.Common.Interfaces.BindleService.RevisionSpinToml.

         at Tomlyn.Toml.ToModel[T](DocumentSyntax syntax, TomlModelOptions options)
         at Hippo.Infrastructure.Services.BindleService.ParseSpinTomlParcel(HttpContent parcelResponse) in /home/runner/work/hippo/hippo/src/Infrastructure/Services/BindleService.cs:line 62
         at Hippo.Infrastructure.Services.BindleService.GetSpinTomlParcel(Invoice invoice, String revisionId) in /home/runner/work/hippo/hippo/src/Infrastructure/Services/BindleService.cs:line 50
         at Hippo.Infrastructure.Services.BindleService.GetRevisionDetails(String revisionId) in /home/runner/work/hippo/hippo/src/Infrastructure/Services/BindleService.cs:line 27
         at Hippo.Application.Revisions.EventHandlers.UpdateRevisionDetails.Handle(CreatedEvent`1 notification, CancellationToken cancellationToken) in /home/runner/work/hippo/hippo/src/Application/Revisions/EventHandlers/UpdateRevisionDetails.cs:line 33
         at MediatR.Mediator.PublishCore(IEnumerable`1 allHandlers, INotification notification, CancellationToken cancellationToken)
         at MediatR.MediatorExtensions.DispatchDomainEvents(IMediator mediator, DbContext context) in /home/runner/work/hippo/hippo/src/Infrastructure/Common/MediatorExtensions.cs:line 22
         at Hippo.Infrastructure.Data.ApplicationDbContext.SaveChangesAsync(CancellationToken cancellationToken) in /home/runner/work/hippo/hippo/src/Infrastructure/Data/ApplicationDbContext.cs:line 60
         at Hippo.Application.Revisions.Commands.ImportRevisionsCommandHandler.Handle(ImportRevisionsCommand request, CancellationToken cancellationToken) in /home/runner/work/hippo/hippo/src/Application/Revisions/Commands/ImportRevisionsCommand.cs:line 49
         at Hippo.Application.Common.Behaviours.PerformanceBehaviour`2.Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate`1 next) in /home/runner/work/hippo/hippo/src/Application/Common/Behaviours/PerformanceBehaviour.cs:line 31
         at Hippo.Application.Common.Behaviours.ValidationBehaviour`2.Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate`1 next) in /home/runner/work/hippo/hippo/src/Application/Common/Behaviours/ValidationBehaviour.cs:line 35
         at Hippo.Application.Common.Behaviours.AuthorizationBehaviour`2.Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate`1 next) in /home/runner/work/hippo/hippo/src/Application/Common/Behaviours/AuthorizationBehaviour.cs:line 78
         at Hippo.Application.Common.Behaviours.UnhandledExceptionBehaviour`2.Handle(TRequest request, CancellationToken cancellationToken, RequestHandlerDelegate`1 next) in /home/runner/work/hippo/hippo/src/Application/Common/Behaviours/UnhandledExceptionBehaviour.cs:line 19
```

tests/integration.rs
FrankYang0529: I would like to use `type nomad` in `make test`. However, should we also consider Windows? If yes, it will be `type nomad.exe` on Windows. Not sure what is a proper way to do that on different platforms.

tests/integration.rs
vdice: Oh, good point.  Yes, we should try using `inmemory` here.

vdice: Ah, you already made the change.  Thanks!

tests/integration.rs
vdice: I think we can use something like the following in the Makefile:

```Makefile
NOMAD := nomad
ifeq ($(OS),Windows_NT)
	NOMAD = nomad.exe
endif

nomad-version:
	$(NOMAD) --version
```

(`nomad-version` just example usage.)

tests/http/simple-spin-rust/spin.toml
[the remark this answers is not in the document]
FrankYang0529: I file an issue for it.

https://github.com/deislabs/hippo/issues/978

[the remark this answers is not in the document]
vdice: I wonder if we should leave this test project as-is (so that it is continued to be tested via the other integration tests) and create a separate test project that is compatible with hippo/`spin deploy` today?  We could have a link to the issue in deislabs/hippo in the new test project for context.  WDYT?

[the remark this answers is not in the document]
vdice: 👍 Yup, that sounds good to me.

tests/integration.rs
FrankYang0529: I added this part in `make test`, but not sure what is a proper way to test `Hippo.Web`. I tried `-h`, `--help`, `-v` and `--version`, but all show error.

vdice: Hmmm, ok.  I don't have a Windows host right now to test, but would something like this work?

```Makefile
WHICH := which
EXT   :=
ifeq ($(OS),Windows_NT)
	WHICH = where
	EXT   = .exe
endif

.PHONY: ensure-bindle-server
ensure-bindle-server:
	@$(WHICH) bindle-server$(EXT) &>/dev/null || \
		(echo "bindle-server$(EXT) must be installed to run the integration tests" && exit 1)

.PHONY: ensure-hippo
ensure-hippo:
	@$(WHICH) Hippo.Web &>/dev/null || \
		(echo "Hippo.Web must be installed to run the integration tests" && exit 1)

.PHONY: ensure-nomad
ensure-nomad:
	@$(WHICH) nomad$(EXT) &>/dev/null || \
		(echo "nomad$(EXT) must be installed to run the integration tests" && exit 1)
```

I'm also open to ideas on where we should put these checks for the integration test dependencies.  Would it be better to place the checks in `integration.rs` itself?

Also, do we like continuing to run all tests (unit and integration) during `make test`?  Or do we want separate targets, e.g. `make test-unit`, `make test-integration`, etc.?  cc @michelleN

FrankYang0529: I remove `$(OS)` in Makefile, so we don't need to test it in two OS. I add the dependency test to `integration.rs` and separate unit and integration test in Makefile.

Makefile
```diff
@@ -7,11 +7,19 @@ build:
 	cargo build --release
 
 .PHONY: test
-test:
-	RUST_LOG=$(LOG_LEVEL) cargo test --all --no-fail-fast -- --nocapture --include-ignored
+test: test-unit test-integration
```
vdice, line 10: Let's also break out the clippy and fmt checks into their own target, maybe `lint`, then `test` can run these first (eg `test: lint test-unit test-integration`)

tests/integration.rs
vdice: Nice, I like that the checks are in `integration.rs` now.  I did a quick search -- do we want to try using something like https://crates.io/crates/which which appears to also support Windows?

Also, can we add a helpful context message for when one of the dependencies isn't found?

vdice (commented):
A few more comments - thanks for the great updates so far.

Makefile
FrankYang0529: Updated it. Thank you.

tests/integration.rs
FrankYang0529: I change to use `which` and give it `with_context`. Thank you.

vdice: Thanks!

vdice (approved):
LGTM!  CI tested on my fork [here](https://github.com/vdice/spin/runs/7254005105?check_suite_focus=true).  Thanks for adding this @FrankYang0529 !

src/commands/deploy.rs
```diff
@@ -231,8 +231,8 @@ impl DeployCommand {
                 let source_dir = crate::app_dir(&self.app)?;
                 let fm = assets::collect(files, &source_dir)?;
                 for f in fm.iter() {
-                    let src_path = source_dir.join(&f.src);
-                    let mut r = File::open(src_path)?;
+                    let mut r = File::open(&f.src)
```
michelleN, line 234: Hey @FrankYang0529 - I think if we remove line 34, it might introduce a regression in that running deploy from outside of the app directory may not work as expected. Would you mind explaining this change?

Makefile
```diff
@@ -7,11 +7,22 @@ build:
...
+lint:
 	cargo clippy --all-targets --all-features -- -D warnings
 	cargo fmt --all -- --check
 
+.PHONY: test-unit
+test-unit:
+	RUST_LOG=$(LOG_LEVEL) cargo test --no-fail-fast -- --skip integration_tests --nocapture --include-ignored
+
+.PHONY: test-integration
+test-integration:
```
michelleN, line 22: It looks like `make test-integration` is also running unit tests:
```
src/spin [add-deploy-integration-test] $ make test-integration
RUST_LOG=spin=trace cargo test integration_tests::test_dependencies --no-fail-fast -- --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.29s
     Running unittests src/lib.rs (target/debug/deps/spin_cli-9890c9e4ab67bd17)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 4 filtered out; finished in 0.00s

     Running unittests src/bin/spin.rs (target/debug/deps/spin-8e3437ebb5e6a5ea)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/integration.rs (target/debug/deps/integration-c72b827f23a47925)

running 1 test
[...]
```

Makefile
```diff
@@ -7,11 +7,22 @@ build:
...
+	RUST_LOG=$(LOG_LEVEL) cargo test integration_tests::test_dependencies --no-fail-fast -- --nocapture
+	RUST_LOG=$(LOG_LEVEL) cargo test integration_tests --no-fail-fast -- --skip integration_tests::test_dependencies --nocapture --include-ignored
```
michelleN, line 24: What is the reason we're using this flag: `--skip integration_tests::test_dependencies`?

michelleN (changes requested):
`make test-integration` gives me the following error:
```
Error: Can't find Hippo.Web

Caused by:
    cannot find binary path
thread 'integration_tests::test_dependencies' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`: the test returned a termination value with a non-zero status code (1) which indicates a failure', /rustc/546c826f0ccaab36e897860205281f490db274e6/library/test/src/lib.rs:184:5
test integration_tests::test_dependencies ... FAILED

failures:

failures:
    integration_tests::test_dependencies

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 10 filtered out; finished in 0.00s

error: test failed, to rerun pass '--test integration'
```

Would you mind adding documentation around how to install dependencies locally for these tests?

fibonacci1729:
Can we feature gate the newly added tests? Something like,
    ```rust
    #[cfg(feature = "e2e-tests")]
    mod e2e {
         ... all new test code goes here ...
    }
    ```

    Then adding the feature to Cargo.toml:
    ```toml
    ...
    [features]
    e2e-tests = []
    ```

    Add a make target for `test-e2e` that invokes the test with the e2e-tests feature enabled.
    
    Why? This way the original integration tests can be run locally until we figure how to
    ensure the installation of e2e prerequisites.

src/commands/deploy.rs
FrankYang0529: Hi @michelleN, if we don't update the file path, we may have an error like this https://github.com/fermyon/spin/issues/590#issuecomment-1167611846.

```console
# before updating the file path, but adding with_context to show informative error message.
# let mut r = File::open(&src_path).with_context(|| anyhow!("Cannot open file {}", &src_path.display()))?;

./target/debug/spin deploy --file tests/http/simple-spin-rust/spin.toml --bindle-server=http://localhost:8080/v1 --hippo-server=http://localhost:5309 --hippo-username=admin --hippo-password=password --deploy-existing-bindle
Error: Cannot open file tests/http/simple-spin-rust/tests/http/simple-spin-rust/assets/test.txt

Caused by:
    No such file or directory (os error 2)
```

```console
# after updating the file path, although there is error from hippo, it has been fixed in https://github.com/deislabs/hippo/issues/978
./target/debug/spin deploy --file tests/http/simple-spin-rust/spin.toml --bindle-server=http://localhost:8080/v1 --hippo-server=http://localhost:5309 --hippo-username=admin --hippo-password=password --deploy-existing-bindle
Error: Unable to create Hippo app

Caused by:
    {"type":"https://tools.ietf.org/html/rfc7231#section-6.6.1","title":"An error occurred while processing your request.","status":500}
```

Makefile
FrankYang0529: Updated `cargo test` command to fix it. Thank you.

```console
$ make test-integration
RUST_LOG=spin=trace cargo test --test integration --no-fail-fast  -- integration_tests::test_dependencies --nocapture
    Finished test [unoptimized + debuginfo] target(s) in 0.33s
     Running tests/integration.rs (target/debug/deps/integration-02d2ab06f0b17d5a)

running 1 test
test integration_tests::test_dependencies ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 10 filtered out; finished in 0.00s

RUST_LOG=spin=trace cargo test --test integration --no-fail-fast -- --skip integration_tests::test_dependencies --nocapture --include-ignored
    Finished test [unoptimized + debuginfo] target(s) in 0.40s
     Running tests/integration.rs (target/debug/deps/integration-02d2ab06f0b17d5a)

running 10 tests
[...]
```

FrankYang0529:
Hi @fibonacci1729, I add `README.md` in the `tests` folder to ask users to install `bindle-server`, `nomad`, and `Hippo.Web` before running `make test-integration`. Besides `test_spin_deploy`, most integration tests also have `bindle-server` dependence, so probably we still can put all integration tests in one module.

michelleN (approved):

vdice:
Thanks again for the excellent work here @FrankYang0529 !

Issue #393, opened 2022-04-24
Grab `.wasm` binaries from the web


ecumene:
I want to be able to share modules with others but not require them to download/compile them themselves

## Spin.toml

My intuition says this URL would be written in the `spin.toml` like so:

```
[[component]]
id = "key-value"
source = "https://github.com/ecumene/components/.../hello-world.wasm"
```

But how would we get the modules? I think there are generally two ways to achieve this

1. Build a tool that saves them to a cache based on the `spin.toml` and change nothing here
1. Have Spin download these at runtime

We could also do both! Maybe we could include a mapping of these modules to their place on the drive, and that would leave it up to something built into Spin or some manager to resolve them.

Mossaka:
We could combine the two ways: having spin download these wasm files at runtime and cache them to a `spin.lock` file, which includes the checksums. In this way when the next time spin is running, if the checksum of the `spin.toml` is not changed, then can use the cached wasm files. And of course if the lock file is deleted, then spin will always fetch the newest remote wasm modules.

ecumene:
Would the lockfile map the downloaded modules to their place on disk too?

radu-matei:
I would decouple the two conversations as two distinct ones:

1. using HTTP URLs as module sources in `spin.toml` files
1. caching pulled modules (and assets), which also includes when running from Bindle, and I think it makes sense to have a common caching implementation for various sources.

For 1, do we need to consider any authentication mechanisms when performing HTTP requests, or are we ok with public sources for now?
(I think for the first iteration on this, public sources, like GitHub releases, are totally fine).

ecumene:
> For 1, do we need to consider any authentication mechanisms when performing HTTP requests, or are we ok with public sources for now?

Like HTTP basic auth? Or something else like a cookie?

For my purposes public sources are fine

lann:
I think we should default to verifying the integrity of remote-sourced binaries to encourage good security practices. That might look something like:

```toml
[component.source]
url = "https://raw.githubusercontent.com/..."
integrity = "sha256:abc123..."
```

_If necessary_, we could allow explicit integrity opt-out, e.g.: `integrity = "insecure"`

This could also (later) be expanded nicely to code signing, e.g. `integrity = "ed25519:<pubkey>"`.

itowlson:
[...]

PR #610, opened 2022-06-29
SIP: URLs in spin.toml sources

technosophos:
SIP for supporting `source` that uses HTTP/HTTPS or `file` schema.

Example:

```toml
[[component]]
source = "https://github.com/fermyon/bartholomew/releases/download/v0.3.0/bartholomew.wasm#sha256:723ade692c4f8b047b6e90b7c9625a57c9819e0f1bd29bd75eae5b79dd436c4b"
```

Signed-off-by: Matt Butcher <matt.butcher@fermyon.com>

### DIFF
--- a/docs/content/sips/XXX-wasm-url.md
+++ b/docs/content/sips/XXX-wasm-url.md
@@ -1,0 +1,165 @@
+title = "SIP XXX - Spin Wasm HTTP URLs"
+template = "main"
+date = "2022-06-29T02:55:31Z"
+---
+
+Summary: A standard way of defining an HTTP(S) URL to a Wasm module
+
+Owner: matt.butcher@fermyon.com
+
+Created: June 29, 2022
+
+## Background
+
+Currently, when a WebAssembly component is loaded, it can only be loaded from a local file source.
+
+For example, to load `scoreboard.wasm`, the current `spin.toml` specifies `source` as a path:
+
+```toml
+[[component]]
+id = "scoreboard"
+source = "components/scoreboard.wasm"
+environment = {REDIS_ADDRESS = "redis://localhost:6379/"}
+[component.trigger]
+route = "/score"
+```
+
+In this case, any WebAssembly module to be used by Spin must be present on the filesystem prior to Spin starting.
+Moreover, there is no indication of what the version of said object is,
+nor any way to ensure that the file located at that path is the one intended by the developer.
+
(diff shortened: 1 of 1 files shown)


radu-matei:
Today you _can_ reference a remote parcel from a remote bindle:

```toml
[component.source]
reference = "fermyon/bartholomew/0.3.0"
parcel = "bartholomew.wasm"
```

But this assumes an always running Bindle server with anonymous pulls enabled — which we can definitely do for our projects, but this does not immediately address the needs of people who don't want to run their own.


We have [explored in #393](https://github.com/fermyon/spin/issues/393#issuecomment-1109093546) using a URL pointing to a module and its digest as the component source:

```
[component.source]
url = "https://raw.githubusercontent.com/..."
integrity = "sha256:abc123..."
```

Similar to the Bindle component source, where we have `reference` and `parcel`, I think having separate `url` and `hash`/`integrity` fields is a bit more clear.


TL; DR — this looks good to me, and I'd prefer separate fields for the new component source option.

lann:
ref #393

vdice:
@technosophos now that we have support for fetching wasm modules via the web (courtesy https://github.com/fermyon/spin/pull/890), should we close this PR?  There is one piece of functionality mentioned in this SIP that isn't yet in Spin, caching functionality, but we do have the following open issue around this: https://github.com/fermyon/spin/issues/455.  As caching is broader than the particular mechanism for fetching modules (eg bindle, web URL, OCI reference), it makes sense to me to track separately anyhow.  WDYT?

itowlson:
@vdice This SIP also asks that we support SHA512, redirects, `file:` URLs, and omitting the digest, though of course we can raise separate issues for those (or may feel that they aren't immediate enough to need capturing now).

technosophos:
When you read a PR and have NO RECOLLECTION of writing it....

Yeah, capturing the few related items as issues is probably all that is necessary before closing this. Thanks!

vdice:
Alright, I believe I've tracked all of the follow-ups, linking back to this issue and/or specific sections of the SIP.  These should be bite-sized enough that we wouldn't need to break out their implementation details into their own SIPs (or a revised SIP based on this PR), but I'll defer to @itowlson and @technosophos.  Thanks for adding all of the high-level details on these features here, @technosophos !

itowlson:
Pretty sure we've banned @technosophos from GitHub by this point so I'll use my powers to close it. :trollface:

PR #612, opened 2022-07-01
Spin go sdk error return statement for nil request

elewis787:
Currently the spin go sdk is checking for an err when creating a go internal http request. However, the error handling only prints the error message and the code continues on. This change adds a return statement to ensure we do not attempt to access/use a nil request object. That said, this might not be something that needs to be handled due to the pass through with the host env checks. 

I tested this change does not break using the go-spin examples in spin-kitchen sink. If there is advice on how this might be trigger to cause the error capture and access of the nil pointer I am happy implement a more holistic test for it.

### DIFF
--- a/sdk/go/http/internals.go
+++ b/sdk/go/http/internals.go
@@ -23,6 +23,7 @@ func handle_http_request(req *C.spin_http_request_t, res *C.spin_http_response_t
 	r, err := http.NewRequest(sr.Method(), sr.URL(), bytes.NewReader(sr.Body()))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
+		res.status = C.uint16_t(http.StatusInternalServerError)
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


elewis787:
Rebased to pick up latest changes

itowlson:
@elewis787 I think something might have gone awry in the rebase?  There seem to be a lot of commits and diffs that appear unrelated to your fix?

elewis787:
> @elewis787 I think something might have gone awry in the rebase? There seem to be a lot of commits and diffs that appear unrelated to your fix?

I agree - let me see if I can undo it

elewis787:
@itowlson - not sure what happened with the rebase - sorry about that.  I reset back to my original commit.

adamreese:
I think this could be hard to debug if no http response is returned.  Can we return a 500 error rather than calling `handler()`?

elewis787:
> I think this could be hard to debug if no http response is returned.  Can we return a 500 error rather than calling `handler()`?  

Good idea and makes sense to me - I'll update the code later this evening.

elewis787:
@adamreese I have updated the code to return a 500 error on the resp if we fail to create a request object. Let me know if this is what you had in mind. 

I also stumbled into another issue with golang/tinygo version that I have outlined here https://github.com/fermyon/spin/issues/624. 

Lastly, running the spin make file in the root directory currently builds using `cargo build --release` which does not produce the debug / unoptimized spin binary. The go integration tests are currently looking for `../../target/debug/` to start spin. Should we update this ?

 I got around it by rebuilding using cargo build.

adamreese (approved):
Looks great!  Thanks

PR #613, opened 2022-07-05
chore: remove unneeded explicit lifetime

dicej:
This should appease clippy and fix CI.

### DIFF
--- a/tests/http/simple-spin-rust/Cargo.lock
+++ b/tests/http/simple-spin-rust/Cargo.lock
@@ -148,7 +148,7 @@ dependencies = [
 
 [[package]]
 name = "spin-sdk"
-version = "0.2.0"
+version = "0.3.0"
 dependencies = [
  "anyhow",
  "bytes",
--- a/crates/engine/src/host_component.rs
+++ b/crates/engine/src/host_component.rs
@@ -30,9 +30,9 @@ pub(crate) struct HostComponents {
 }
 
 impl HostComponents {
-    pub(crate) fn insert<'a, T: 'static, Component: HostComponent + 'static>(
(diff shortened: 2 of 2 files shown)


lann (approved):
I did a few iterations of this code; I'm sure I just left these in accidentally.

tests/http/simple-spin-rust/Cargo.lock
```diff
@@ -148,7 +148,7 @@ dependencies = [
 
 [[package]]
 name = "spin-sdk"
-version = "0.2.0"
+version = "0.3.0"
```
lann, line 151: Is this intentional (in this PR?)

dicej: Seems to be a side effect of building.  Perhaps someone made an earlier change but didn't check in Cargo.lock?

I can remove that from the commit and force-push, but I'm guessing it will pop up the next time anyone builds.

lann: Yeah seems fine to keep it, just making sure it wasn't accidentally pulled in from another change.

lann:
> This should appease clippy and fix CI.

Oh right we just got a new rust release :+1:

dicej:
> > This should appease clippy and fix CI.
> 
> Oh right we just got a new rust release +1

What's slightly confusing is that the [referenced lint](https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes) was apparently added in Rust 1.29.0 or earlier.  Not sure why it's only complaining now.  Perhaps the lint level was raised or the lint was expanded to cover more cases.

Issue #615, opened 2022-07-06
Invalid discriminant for `option`


Dueen:
Hey all,
Just tried spin out for the first time and tried to build the hello-rust example and I'm running into the following situation:

## Expected Behavior

```sh
HTTP/1.1 200 OK
foo: bar
content-length: 15

Hello, Fermyon!
```

## Steps to Reproduce the Problem

  1. Follow installation guide using cargo 
  2. Build new project using: spin new http-rust spin-hello-world
  3. Run:  spin build --up
  4. Make a request with: curl -i localhost:3000/hello

## Actual Behavior
I get a response of:
```sh
HTTP/1.1 500 Internal Server Error
content-length: 0
```

With  a log of:
```sh
 ERROR spin_http_engine: Error processing request: invalid discriminant for `option`
```

## Specifications

  - Version: 0.3.0
  - Platform: Intel Mac running MacOS 12.4

As I said this is the first time I've built spin so not sure where to debug this

radu-matei:
Hi, @Dueen, thanks for reporting the issue!
Would you mind first exporting the `RUST_LOG=spin=trace` environment variable, so we get a bit more detailed output?

The error is really confusing. So first things first:

- what happens if you run the `cargo build` manually? (the one executing when running `spin build`)
- is there a Wasm binary in `/target/wasm32-wasi/release/`?
- I assume you are seeing this error when running `spin up?

Thanks!

Dueen:
@radu-matei  Yes, running cargo build does give me a `/target/wasm32-wasi/release/` directory. And I see the error when making a request to `/hello` , however testing the `/healthz` path gives me a success response.

This is what my trace looks like with RUST_LOG=spin=trace:
```sh
TRACE spin_cli::commands::up: Running trigger executor: "/Users/usr/.cargo/bin/spin" "trigger" "http"
2022-07-06T08:01:13.014532Z TRACE build: spin_engine: Created module for component spin-hello-world from file "/Users/usr/Repos/spin-hello-world/target/wasm32-wasi/release/spin_hello_world.wasm"    
2022-07-06T08:01:13.022350Z TRACE build: spin_engine: Created pre-instance from module for component spin-hello-world.    
2022-07-06T08:01:13.022377Z TRACE build: spin_engine: Execution context initialized.    
2022-07-06T08:01:13.023528Z TRACE spin_http_engine: Constructed router for application spin-hello-world: {Exact("/hello"): "spin-hello-world"}    
[...]

PR #616, opened 2022-07-06
doc(quickstart): specify tag in `git clone` command

dicej:
Per #615, the HEAD of the main branch of the repo is not guaranteed to be
compatible with the most recent SDK release, so it's better to have users use
the latest release tag.

We also need better error reporting for SDK compatibility issues; I'll open a
new work item to track that.

### DIFF
--- a/docs/content/quickstart.md
+++ b/docs/content/quickstart.md
@@ -24,7 +24,7 @@ $ ./spin --help
 If you have [`cargo`](https://doc.rust-lang.org/cargo/getting-started/installation.html), you can clone the repo and install it to your path:
 
 ```bash
-$ git clone https://github.com/fermyon/spin
+$ git clone https://github.com/fermyon/spin -b v0.3.0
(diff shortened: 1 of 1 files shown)


lann (approved):

fibonacci1729 (approved):

michelleN (approved):
In the release process document, there is a line
`Check the docs for hard-coded version strings` that covers updating this but we should make the release process doc better in the future to ensure we don't miss updating this version reference.

dicej:
> In the release process document, there is a line `Check the docs for hard-coded version strings` that covers updating this but we should make the release process doc better in the future to ensure we don't miss updating this version reference.

Agreed.  Hopefully it will be pretty obvious given that the version number also appears just above in the `wget` command to get the release tarball.

lann:
I think version affinity can really only handle tagged releases, not running from `main`

Issue #605, opened 2022-06-27
Make `files` mounts read-only by default


lann:
They are currently writable, but not persistent. It might be less confusing to make them read-only by default with an explicit `allow_transient_writes = true` or something if necessary.

lann:
Context: https://github.com/fermyon/spin/pull/147#discussion_r822268913

FrankYang0529:
For this issue, if users give `allow_transient_writes = true`, do we need to watch files change and copy them to `temp` directory again?

lann:
> For this issue, if users give `allow_transient_writes = true`, do we need to watch files change and copy them to `temp` directory again?

Not for _this_ issue, no, but that is something we have discussed before: https://github.com/fermyon/spin/issues/148, https://github.com/fermyon/spin/issues/324#issuecomment-1090169857

FrankYang0529:
I would like to take this issue, so I want to confirm change area. For both local and bindle loaders, files in the `temp` directory are read-only by default. If users give `allow_transient_writes = true`, we will allow them to update files like the current implementation. Do I misunderstand anything? Thank you.

PR #618, opened 2022-07-07
feat(up): add --allow-transient-write

FrankYang0529:
resolve https://github.com/fermyon/spin/issues/605

## Test cases

Please run the following script at the end of each case.
```console
for entry in $TEMP/assets/**/*.txt
do
  test -w $entry && echo "Writable" || echo "Not Writable"
done
```

### Local

**Setup**
```console
cargo build
cd tests/http/simple-spin-rust
```

**Case 1: new file can be set to read-only**
Should show Not Writable.
```console
TEMP=$(mktemp -d)
../../../target/debug/spin up --temp $TEMP
```

**Case 2: same temporary folder can be changed to writable**
Follow case 1. Should show Writable.
```console
../../../target/debug/spin up --temp $TEMP --allow-transient-write
```

**Case 3: new file can be set to writable**
Should show Writable.
```console
TEMP=$(mktemp -d)
../../../target/debug/spin up --temp $TEMP --allow-transient-write
```

### Bindle

**Setup**
```console
BINDLE_TEMP=$(mktemp -d)
bindle-server --unauthenticated -d $BINDLE_TEMP
cd tests/http/simple-spin-rust
../../../target/debug/spin bindle push
```

**Case 4: new file can be set to read-only**
Should show Not Writable.
```console
TEMP=$(mktemp -d)
./target/debug/spin up --temp $TEMP --bindle spin-hello-world/1.0.0
```

**Case 5: same temporary folder content can be updated**
Stop the bindle-server and start a new bindle-server with a new temporary folder.
```console
BINDLE_TEMP=$(mktemp -d)
bindle-server --unauthenticated -d $BINDLE_TEMP
```
Update content in `spin/tests/http/simple-spin-rust/assets/test.txt`. Run spin up again. The file in the temporary folder should be updated and it is still Not Writable after the update.
```console
cd tests/http/simple-spin-rust
../../../target/debug/spin bindle push
cd ../../../
./target/debug/spin up --temp $TEMP --bindle spin-hello-world/1.0.0
```

**Case 6: same temporary folder can be changed to writable**
Follow case 5. Should show Writable.
```console
./target/debug/spin up --temp $TEMP --bindle spin-hello-world/1.0.0 --allow-transient-write
```

**Case 7: new file can be set to writable**
Should show Not Writable.
```console
TEMP=$(mktemp -d)
./target/debug/spin up --temp $TEMP --bindle spin-hello-world/1.0.0 --allow-transient-write
```


crates/loader/src/local/assets.rs
```diff
@@ -208,13 +213,25 @@ async fn copy(file: &FileMount, dir: impl AsRef<Path>) -> Result<()> {
...
+    }
+
     let _ = tokio::fs::copy(&from, &to)
         .await
         .with_context(|| anyhow!("Error copying asset file  '{}'", from.display()))?;
 
-    let perms = tokio::fs::metadata(&to).await?.permissions();
-    // perms.set_readonly(true);
-    tokio::fs::set_permissions(&to, perms).await?;
+    let mut perms = tokio::fs::metadata(&to).await?.permissions();
```
itowlson, line 230: The `let mut perms = ...; perms.set_readonly(...); set_permissions(..., perms).await.with_context(...)` sequence seems to happen in quite a few places.  Could we create a utility function (e.g. `set_writeable` or `allow_write`)  to tidy and reveal intent?

FrankYang0529: Yes, I add `change_file_permission` function for it. Thank you.

crates/loader/src/bindle/assets.rs
```diff
@@ -50,33 +54,56 @@ impl Copier {
...
         if to.exists() {
             match check_existing_file(to.clone(), p).await {
-                Ok(true) => return Ok(()),
-                Ok(false) => (),
+                Ok(true) => {
+                    change_file_permission(&to, allow_transient_write).await?;
+                    return Ok(());
+                }
+                Ok(false) => {
+                    // if destination file is read-only, set it to writable first
```
itowlson, line 101: This comment sent me in the wrong direction a bit (I thought there was a bug where you forgot to consult `allow_transient_writes`), partly because `check_existing_file` is already confusing!  I think this branch is "file exists but digest doesn't match so it's like corrupted.  In which case this is "set it writeable because we need to overwrite it" I think?  Sort of a "comment says what the code is doing but needs to say *why* instead" issue if that makes sense!

src/commands/up.rs
```diff
@@ -81,6 +81,10 @@ pub struct UpCommand {
     #[clap(long = "temp")]
     pub tmp: Option<PathBuf>,
 
+    /// Temporary directory for the static assets of the components.
```
itowlson, line 84: This doc string doesn't look right?

itowlson (changes requested):
Looks good to me - thanks!  Just two comments - the doc string is must-fix, but the comment one is just nice to have.

crates/loader/src/bindle/assets.rs
FrankYang0529: I change it to `file exists but digest doesn't match, set it to writable first for further writing`. Do you think it's more clear? Thank you.

src/commands/up.rs
FrankYang0529: Fixed it. Thank you.

crates/loader/src/assets.rs
```diff
@@ -90,6 +90,19 @@ pub fn file_sha256_digest_string(path: impl AsRef<Path>) -> std::io::Result<Stri
     Ok(digest_string)
 }
 
+/// Return an error if a path is not under a given directory.
```
itowlson, line 93: Apologies for missing this before but I think this doc string is incorrect?

itowlson (approved):
This looks good to me - thanks for the tidying!  There is a doc string that I missed before (sorry) that I think is wrong but it's not user-facing so it's not a blocker - but if you have time to fix it then that would be awesome.  Otherwise happy to merge and fix it up later.

crates/loader/src/assets.rs
FrankYang0529: Hi @itowlson, no worries, it's my fault that I didn't check it carefully. Thanks for catching it!

itowlson (approved):
Looks good - thanks and sorry for the multiple go-arounds!

PR #623, opened 2022-07-09
ci: replace makefile with xtask

csmoe:
Current development flow depends on `make`, makes it cumbersome to contribute on windows and hard to maintain as more steps to add in the future.

This commit introduce [xtask](https://github.com/matklad/cargo-xtask) with the tiny fast cli crate [`xflags`](https://crates.io/crates/xflags)(we should never use the `clap` like crates, as they are slow to compile)


lann:
Thanks for the PR. After some internal discussion I don't think we want to make this change at this time as it introduces breaking changes to developer environments and some external build processes. It looks like `make` is available [via Chocolatey](https://community.chocolatey.org/packages/make) for Windows.

PR #640, opened 2022-07-14
Fix crate unit tests not being run

itowlson:
The makefile was reorganised to provide nice clearly separated unit and integration test targets, but in the process the `--all` fell on the cutting room floor.  So we are currently running only the tests for the bin, not the crates.

I assume the `test-integration` target doesn't need the flag.

(Also fixes a test failure that sneaked through when the tests weren't being fully run.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ lint:
 
 .PHONY: test-unit
 test-unit:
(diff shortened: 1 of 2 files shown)


fibonacci1729 (approved):

radu-matei (approved):
Good catch, thanks!

LGTM

radu-matei:
Oh, DCO bot is unhappy.

itowlson:
_bumping DCO to Assistant Principal Arch-Nemesis_

PR #642, opened 2022-07-14
fix: Terminate trigger handler on `spin up` termination

### LINKED ISSUE #638: Integration tests for go sdk does not terminate the spin process
Integration test for Golang starts a spin instance that doesn't get shutdown.

https://github.com/fermyon/spin/blob/main/sdk/go/integration_test.go#L75

lann:
The `Ctrl-C` handler works because it is handled in the trigger handler
child process, but if the parent `spin up` process is terminated
instead - as with e.g. Nomad - the child is left running.

Fix this by adding a signal handler to the parent process which sends
SIGTERM to the child.

I expect this to fix #638

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -157,7 +157,20 @@ impl UpCommand {
 
         tracing::trace!("Running trigger executor: {:?}", cmd);
 
-        let status = cmd.status().context("Failed to execute trigger")?;
+        let mut child = cmd.spawn().context("Failed to execute trigger")?;
+
+        #[cfg(not(windows))]
+        {
+            // https://github.com/nix-rust/nix/issues/656
+            let pid = nix::unistd::Pid::from_raw(child.id() as i32);
+            ctrlc::set_handler(move || {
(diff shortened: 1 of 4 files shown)


src/commands/up.rs
```diff
@@ -157,7 +157,20 @@ impl UpCommand {
 
         tracing::trace!("Running trigger executor: {:?}", cmd);
 
-        let status = cmd.status().context("Failed to execute trigger")?;
+        let mut child = cmd.spawn().context("Failed to execute trigger")?;
+
+        #[cfg(not(windows))]
```
itowlson, line 162: What happens if it is on Windows?

fibonacci1729:
<a href="https://imgflip.com/i/6msmi5"></a><div><a href="https://imgflip.com/memegenerator">from Imgflip Meme Generator</a></div>

src/commands/up.rs
lann: What happens on windows today? 🤷

PR #645, opened 2022-07-15
feat: add exclude-files

### LINKED ISSUE #554: Add a way to exclude patterns from `files` patterns
Per https://github.com/fermyon/spin-fileserver/issues/11, it is easy to accidentally include unwanted files in a `files` glob pattern. We should think of a mechanism to exclude some files from a glob, perhaps something like a `exclude_files` pattern. It might be nice for this to even default to `["**/.*"]` or similar.

FrankYang0529:
resolves #554

### DIFF
--- a/crates/loader/src/local/assets.rs
+++ b/crates/loader/src/local/assets.rs
@@ -6,7 +6,10 @@ use crate::assets::{
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use futures::{future, stream, StreamExt};
 use spin_manifest::DirectoryMount;
-use std::path::{Path, PathBuf};
+use std::{
+    path::{Path, PathBuf},
+    vec,
+};
 use tracing::log;
 use walkdir::WalkDir;
 
@@ -20,6 +23,7 @@ pub(crate) async fn prepare_component(
     base_dst: impl AsRef<Path>,
     id: &str,
     allow_transient_write: bool,
+    exclude_files: &[String],
 ) -> Result<Vec<DirectoryMount>> {
     log::info!(
         "Mounting files from '{}' to '{}'",
@@ -27,7 +31,7 @@ pub(crate) async fn prepare_component(
         base_dst.as_ref().display()
     );
 
-    let files = collect(raw_mounts, src)?;
+    let files = collect(raw_mounts, exclude_files, src)?;
     let host = create_dir(&base_dst, id).await?;
     let guest = "/".to_string();
     copy_all(&files, &host, allow_transient_write).await?;
@@ -45,11 +49,7 @@ pub struct FileMount {
 }
 
 impl FileMount {
-    fn from(
-        src: Result<impl AsRef<Path>, glob::GlobError>,
-        relative_to: impl AsRef<Path>,
-    ) -> Result<Self> {
-        let src = src?;
+    fn from(src: impl AsRef<Path>, relative_to: impl AsRef<Path>) -> Result<Self> {
         let relative_dst = to_relative(&src, &relative_to)?;
         let src = src.as_ref().to_path_buf();
         Ok(Self { src, relative_dst })
@@ -63,13 +63,19 @@ impl FileMount {
 }
 
 /// Generate a vector of file mounts for a component given all its file patterns.
-pub fn collect(raw_mounts: &[RawFileMount], rel: impl AsRef<Path>) -> Result<Vec<FileMount>> {
+pub fn collect(
+    raw_mounts: &[RawFileMount],
+    exclude_files: &[String],
+    rel: impl AsRef<Path>,
+) -> Result<Vec<FileMount>> {
     let (patterns, placements) = uncase(raw_mounts);
 
     let pattern_files = collect_patterns(&patterns, &rel)?;
     let placement_files = collect_placements(&placements, &rel)?;
     let all_files = [pattern_files, placement_files].concat();
-    Ok(all_files)
+
+    let exclude_patterns = convert_strings_to_glob_patterns(exclude_files, &rel)?;
+    Ok(get_included_files(all_files, &exclude_patterns))
 }
 
 fn collect_placements(
@@ -172,7 +178,7 @@ fn collect_pattern(pattern: &str, rel: impl AsRef<Path>) -> Result<Vec<FileMount
     let matches = glob::glob(&abs.to_string_lossy())?;
     let specifiers = matches
         .into_iter()
-        .map(|path| FileMount::from(path, &rel))
+        .map(|path| FileMount::from(path?, &rel))
         .collect::<Result<Vec<_>>>()?;
     let files: Vec<_> = specifiers.into_iter().filter(|s| s.src.is_file()).collect();
     ensure_all_under(&rel, files.iter().map(|s| &s.src))?;
@@ -259,4 +265,48 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
     // can tell - with Unix-style guest paths. So we have to check these
     // paths specifically using Unix logic rather than the system function.
     path.as_ref().to_string_lossy().starts_with('/')
+}
+
+/// Convert strings to glob patterns
+fn convert_strings_to_glob_patterns<T: AsRef<str>>(
+    files: &[T],
+    rel: impl AsRef<Path>,
+) -> Result<Vec<glob::Pattern>> {
+    let file_paths = files
+        .iter()
+        .map(|f| match rel.as_ref().join(f.as_ref()).to_str() {
+            Some(abs) => Ok(abs.to_owned()),
+            None => Err(anyhow!(
+                "Can't join {} and {}",
+                rel.as_ref().display(),
+                f.as_ref()
+            )),
+        })
+        .collect::<Result<Vec<_>>>()?;
+    file_paths
+        .iter()
+        .map(|f| {
+            glob::Pattern::new(f).with_context(|| format!("can't convert {} to glob pattern", f))
+        })
+        .collect::<Result<Vec<glob::Pattern>>>()
+}
+
+/// Remove files which match excluded patterns
+fn get_included_files(files: Vec<FileMount>, exclude_patterns: &[glob::Pattern]) -> Vec<FileMount> {
+    files
+        .into_iter()
+        .filter(|f| {
+            for exclude_pattern in exclude_patterns {
+                if exclude_pattern.matches_path(Path::new(&f.src)) {
+                    tracing::info!(
+                        "file: {} is excluded by pattern {}",
+                        f.src.display(),
+                        exclude_pattern
+                    );
+                    return false;
+                }
+            }
+            true
+        })
+        .collect::<Vec<_>>()
 }
--- a/crates/publish/src/expander.rs
+++ b/crates/publish/src/expander.rs
@@ -174,7 +174,8 @@ fn collect_assets(
     base_dir: impl AsRef<Path>,
 ) -> Result<Vec<(spin_loader::local::assets::FileMount, String)>> {
     let patterns = component.wasm.files.clone().unwrap_or_default();
-    let file_mounts = spin_loader::local::assets::collect(&patterns, &base_dir)
+    let exclude_files = component.wasm.exclude_files.clone().unwrap_or_default();
+    let file_mounts = spin_loader::local::assets::collect(&patterns, &exclude_files, &base_dir)
         .with_context(|| format!("Failed to get file mounts for component '{}'", component.id))?;
     let annotated = file_mounts
         .into_iter()
--- a/crates/loader/src/local/config.rs
+++ b/crates/loader/src/local/config.rs
@@ -98,6 +98,9 @@ pub struct RawWasmConfig {
     /// is either a file path or glob relative to the spin.toml file, or a
     /// mapping of a source path to an absolute mount path in the guest.
     pub files: Option<Vec<RawFileMount>>,
+    /// Optional list of file path or glob relative to the spin.toml that don't mount to wasm.
+    /// When exclude_files conflict with files config, exclude_files take precedence.
+    pub exclude_files: Option<Vec<String>>,
     /// Optional list of HTTP hosts the component is allowed to connect.
     pub allowed_http_hosts: Option<Vec<String>>,
 }
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -196,7 +196,16 @@ async fn core(
     let description = raw.description;
     let mounts = match raw.wasm.files {
         Some(f) => {
-            assets::prepare_component(&f, src, &base_dst, &id, allow_transient_write).await?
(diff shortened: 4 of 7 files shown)


itowlson (commented):
Apologies for the premature review - I didn't spot it was WIP until I'd written half a dozen comments and I didn't want to throw them away!  Hope they're useful all the same!

crates/loader/src/local/assets.rs
```diff
@@ -129,52 +138,66 @@ fn collect_placement(
...
     let abs = rel.as_ref().join(pattern);
     log::trace!("Resolving asset file pattern '{:?}'", abs);
 
-    let matches = glob::glob(&abs.to_string_lossy())?;
-    let specifiers = matches
-        .into_iter()
-        .map(|path| FileMount::from(path, &rel))
-        .collect::<Result<Vec<_>>>()?;
-    let files: Vec<_> = specifiers.into_iter().filter(|s| s.src.is_file()).collect();
+    let mut files = Vec::new();
```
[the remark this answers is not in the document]
FrankYang0529: I update `is_exlcluded_file` to return `Result<bool>`, so probably we can't just use `filter` for it. I separate for loop to multiple iterators so that we can use the `?` operator in each `Result`.

crates/loader/src/local/assets.rs
```diff
@@ -260,3 +283,22 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
     // paths specifically using Unix logic rather than the system function.
     path.as_ref().to_string_lossy().starts_with('/')
 }
+
+/// Convert strings to glob patterns
+pub fn convert_strings_to_glob_patterns<T: AsRef<str>>(files: &[T]) -> Result<Vec<glob::Pattern>> {
+    Ok(files
+        .into_iter()
+        .map(|f| glob::Pattern::new(f.as_ref()))
+        .collect::<Result<Vec<glob::Pattern>, _>>()?)
```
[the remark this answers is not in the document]
FrankYang0529: I add `with_context` after `glob::Pattern::new(f.as_ref())`. Thank you.

crates/publish/src/expander.rs
```diff
@@ -174,7 +177,9 @@ fn collect_assets(
     base_dir: impl AsRef<Path>,
 ) -> Result<Vec<(spin_loader::local::assets::FileMount, String)>> {
     let patterns = component.wasm.files.clone().unwrap_or_default();
-    let file_mounts = spin_loader::local::assets::collect(&patterns, &base_dir)
+    let exclude_files = component.wasm.exclude_files.clone().unwrap_or_default();
+    let exclude_patterns = convert_strings_to_glob_patterns(&exclude_files)?;
```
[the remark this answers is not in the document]
FrankYang0529: Update it in `convert_strings_to_glob_patterns` function. Thank you.

crates/loader/src/local/assets.rs
```diff
@@ -20,14 +23,17 @@ pub(crate) async fn prepare_component(
...
+    exclude_files: &[String],
 ) -> Result<Vec<DirectoryMount>> {
     log::info!(
         "Mounting files from '{}' to '{}'",
         src.as_ref().display(),
         base_dst.as_ref().display()
     );
 
-    let files = collect(raw_mounts, src)?;
+    let exclude_patterns = convert_strings_to_glob_patterns(exclude_files)?;
```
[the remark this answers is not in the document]
FrankYang0529: Update it in `convert_strings_to_glob_patterns` function. Thank you.

crates/loader/src/local/assets.rs
```diff
@@ -260,3 +283,22 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
...
+pub fn convert_strings_to_glob_patterns<T: AsRef<str>>(files: &[T]) -> Result<Vec<glob::Pattern>> {
+    Ok(files
+        .into_iter()
+        .map(|f| glob::Pattern::new(f.as_ref()))
+        .collect::<Result<Vec<glob::Pattern>, _>>()?)
+}
+
+/// Return whether the file path in in excluded patterns
+pub fn is_excluded_file(file: impl AsRef<Path>, exclude_patterns: &[glob::Pattern]) -> bool {
+    tracing::info!("file: {:?}", file.as_ref());
```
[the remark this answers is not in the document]
FrankYang0529: Sorry, this is just for tracing some information when I developed. Removed it. Thank you.

crates/loader/src/local/assets.rs
```diff
@@ -129,52 +138,66 @@ fn collect_placement(
...
+        }
+        let relative_to_match_root_dst = to_relative(match_path, &abs).with_context(|| {
+            format!(
+                "Failed to establish relative path for '{}' relative to '{}'",
+                match_path.display(),
+                abs.display()
+            )
+        })?;
+        let guest_dst = relative_guest_path.join(relative_to_match_root_dst);
+        files.push(FileMount::from_exact(match_path, &guest_dst)?);
```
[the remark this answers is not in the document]
FrankYang0529: I separate for loop to multiple iterators. Each iterator will resolve `Result` with the `?` operator so that the next iterator can get data directly.

FrankYang0529:
Hi @itowlson, I should set it as draft PR first. Thanks for having suggestions! I update an integration test, and it can pass on my computer.

crates/loader/src/local/assets.rs
```diff
@@ -63,21 +64,26 @@ impl FileMount {
...
+    raw_mounts: &[RawFileMount],
+    exclude_patterns: &[glob::Pattern],
+    rel: impl AsRef<Path>,
+) -> Result<Vec<FileMount>> {
     let (patterns, placements) = uncase(raw_mounts);
 
-    let pattern_files = collect_patterns(&patterns, &rel)?;
-    let placement_files = collect_placements(&placements, &rel)?;
+    let pattern_files = collect_patterns(&patterns, exclude_patterns, &rel)?;
+    let placement_files = collect_placements(&placements, exclude_patterns, &rel)?;
```
itowlson, line 75: I'm a bit dubious about doing the filtering in both `collect_patterns` and `collect_placements`.  It significantly increases the complexity of those functions (which are already fairly complicated), and we seem to be doing very much the same filtering in both (although I may be missing some nuance here).

How viable is it to perform the filtering on the `all_files` collection, after we have done all the accumulation of files from patterns and placements?  For me this would be clearer if practical.

crates/loader/src/local/assets.rs
```diff
@@ -260,3 +289,36 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
...
+    rel: impl AsRef<Path>,
+    exclude_patterns: &[glob::Pattern],
+) -> Result<bool> {
+    let relative_path = to_relative(&file, &rel)?;
+    let relative_path = Path::new(&relative_path);
+
+    for exclude_pattern in exclude_patterns {
+        if exclude_pattern.matches_path(relative_path) {
+            tracing::info!(
+                "file: {:?} is exluded by pattern {:?}",
```
itowlson, line 316: tiny typo
```suggestion
                "file: {:?} is excluded by pattern {:?}",
```

crates/loader/src/local/assets.rs
```diff
@@ -129,52 +137,73 @@ fn collect_placement(
         bail!("Cannot place {}: source must be a directory", abs.display());
     }
 
+    // let mut files = Vec::new();
```
itowlson, line 140: Think this can now be deleted

crates/loader/src/local/assets.rs
```diff
@@ -260,3 +289,36 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
...
+        .iter()
+        .map(|f| {
+            glob::Pattern::new(f.as_ref())
+                .with_context(|| format!("can't convert {} to glob pattern", f.as_ref()))
+        })
+        .collect::<Result<Vec<glob::Pattern>>>()
+}
+
+/// Return whether the file path in in excluded patterns
+pub fn is_excluded_file(
```
itowlson, line 305: Does this need to be public?  It feels like nothing outside the loader should care about it...?

crates/loader/src/local/assets.rs
```diff
@@ -260,3 +289,36 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
     // paths specifically using Unix logic rather than the system function.
     path.as_ref().to_string_lossy().starts_with('/')
 }
+
+/// Convert strings to glob patterns
+pub fn convert_strings_to_glob_patterns<T: AsRef<str>>(files: &[T]) -> Result<Vec<glob::Pattern>> {
```
itowlson, line 294: Could this conversion be an internal concern of `assets::collect`?  It feels like we always need to call them together and I'm wondering if we'd be better off passing the raw strings to collect and having it do the conversion...?  (I do like that the input has been validated before entry though, so I'm in two minds.)

itowlson (changes requested):
Thanks for this!  My 'request changes' is more a 'request viability of a change' - moving exclusion filtering to after accumulation, rather than doing it as we go - and if there's a reason that won't work then no worries.  Everything else is minor.  Thanks again!

crates/loader/src/local/assets.rs
FrankYang0529: Thanks for the suggestion. I remove `exclude_patterns` from `collect_patterns` and `collect_placements`. The function is simpler now.

crates/loader/src/local/assets.rs
FrankYang0529: Updated it. Thank you.

crates/loader/src/local/assets.rs
FrankYang0529: Removed it. Thank you.

crates/loader/src/local/assets.rs
FrankYang0529: I remove the function. Thank you.

crates/loader/src/local/assets.rs
FrankYang0529: Yes, I updated it. Other components only need to call `assets::collect` without getting `exclude_patterns` first. Thanks for the suggestion!

crates/loader/src/local/assets.rs
```diff
@@ -130,37 +152,45 @@ fn collect_placement(
     }
 
     let walker = WalkDir::new(&abs);
-    let files = walker
+    let dir_entries = walker
```
itowlson, line 155: I feel like this block should no longer need to change except for adding the relative path to the FileMount constructor?  (Which, as mentioned elsewhere, I suspect it not needed - but even if it is...)

crates/loader/src/local/assets.rs
```diff
@@ -130,37 +152,45 @@ fn collect_placement(
...
+            if let Err(e) = relative_to_rel {
+                return Some(Err(e).with_context(|| {
+                    format!(
+                        "Failed to establish relative path for '{}' to '{}'",
+                        match_path.display(),
+                        rel.as_ref().display(),
+                    )
+                }));
+            }
+            let guest_dst = relative_guest_path.join(relative_to_abs.unwrap());
```
itowlson, line 186: Although these unwraps are currently infallible because of the `if let` statements above, I'm anxious they would prevent the compiler from flagging it if the previous code changed in a way that made them fallible.

I wonder if we could reformulate the error checking along the lines of

```
let relative_to_rel_definitely_for_realz = match relative_to_rel {
  Ok(r) => r,
  Err(e) => return Some(...,)
};
```

crates/loader/src/local/assets.rs
```diff
@@ -40,36 +44,54 @@ pub(crate) async fn prepare_component(
 pub struct FileMount {
     /// The source
     pub src: PathBuf,
+    /// The relative path to spin.toml parent folder.
+    pub relative_src: String,
```
itowlson, line 48: Do you need to keep this in the FileMount struct?  The excluded files filter should be able to work on absolute paths - just prepend the base source directory (which we call `rel`, but is absolute) before resolving each pattern (as we do in `collect_pattern`).

crates/loader/src/local/assets.rs
```diff
@@ -260,3 +290,35 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
     // paths specifically using Unix logic rather than the system function.
     path.as_ref().to_string_lossy().starts_with('/')
 }
+
+/// Convert strings to glob patterns
+fn convert_strings_to_glob_patterns<T: AsRef<str>>(files: &[T]) -> Result<Vec<glob::Pattern>> {
```
itowlson, line 295: If you pass the base source directory in here, you can make the glob patterns absolute, and then shouldn't need to keep the relative source around in the FileMount.

crates/loader/src/local/assets.rs
```diff
@@ -260,3 +290,35 @@ fn is_absolute_guest_path(path: impl AsRef<Path>) -> bool {
...
+                .with_context(|| format!("can't convert {} to glob pattern", f.as_ref()))
+        })
+        .collect::<Result<Vec<glob::Pattern>>>()
+}
+
+/// Remove files which match excluded patterns
+fn get_included_files(files: Vec<FileMount>, exclude_patterns: &[glob::Pattern]) -> Vec<FileMount> {
+    files
+        .iter()
+        .cloned()
```
itowlson, line 309: I am not sure, but since you are passing `files` by value rather than by reference, you can probably to `into_iter` and save the clone?

itowlson (changes requested):
This is looking good - I think there is one major simplification you can make that will allow you to back out all the changes to `FileMount` and the collect functions, but other than that I think this is good to go.

crates/loader/src/local/assets.rs
FrankYang0529: Thanks for the suggestion. I remove `relative_src` and updates in `collect_placement`. It looks much better now!

crates/loader/src/local/assets.rs
FrankYang0529: Yeah, added `rel` here. Thank you.

crates/loader/src/local/assets.rs
FrankYang0529: Removed related updates in `collect_placement`.

crates/loader/src/local/assets.rs
FrankYang0529: Yes, removed it in `collect_placement`. Thank you.

itowlson (approved):
Looks good!  Thanks for the feature and for putting up with my nitpicking!

itowlson:
Oh no, it looks like the signatures got lost from some of the commits - could you re-sign please?  We need it for the DCO, sorry.

FrankYang0529:
Hmm, I tried to update commits but not sure why CI can't pass. It keeps showing `error Connection refused (os error 61)`. However, I can run the integration test without error on my computer.

FrankYang0529:
I update `wait_tcp` to wait for 180 seconds and integration tests can pass.

Issue #614, opened 2022-07-06
Deploy: raw JSON in output if Hippo login fails


itowlson:
```
# URLs etc set via EVs
13:21 $ ../../target/debug/spin deploy --hippo-password NOTMYPASSWORD
Error: {"type":"https://tools.ietf.org/html/rfc7231#section-6.5.1","title":"Login failed.","status":400,"detail":"Login failed: "}
```

Our other errors are more plain textish (it would be good to have JSON output options but then you wouldn't prefix the JSON with the `Error:` text, and that's a different and much broader concern anyway!)

itowlson:
Would be useful to identify it as originating from Hippo as well (for when Bindle has auth and we can fail to log into that too _whee_)

lidongjies:
Hi @itowlson, can I pick this up?

itowlson:
@lidongjies That would be great - thanks!

radu-matei:
The fix for this issue was merged — thank you so much for your contribution, @lidongjies!

PR #647, opened 2022-07-16
feat(deploy): Output plain text when Hippo login fails

lidongjies:
Signed-off-by: lidongjies <lidongjies@gmail.com>

With this PR, if it fails to obtain the Hippo token when executing the "spin deploy" command, the following error message will be output

> Error:  Problem login Hippo.
> 
> Caused by:
>           {"type":"https://tools.ietf.org/html/rfc7231#section-6.5.1","title":"Login failed.","status":400,"detail":"Login failed: "}

Related issue #614

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,6 @@ url = "2.2.2"
 wasi-outbound-http = { path = "crates/outbound-http" }
 wasmtime = "0.35.3"
 
-
 [target.'cfg(target_os = "linux")'.dependencies]
 # This needs to be an explicit dependency to enable
 # '--features openssl/vendored', which is used for Linux releases.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ regex = "1.5.5"
 reqwest = { version = "0.11", features = ["stream"] }
 semver = "1.0"
 serde = { version = "1.0", features = [ "derive" ] }
+serde_json = "1.0.82"
 sha2 = "0.10.2"
 spin-build = { path = "crates/build" }
 spin-config = { path = "crates/config" }
@@ -47,6 +48,7 @@ tracing-subscriber = { version = "0.3.7", features = [ "env-filter" ] }
 url = "2.2.2"
 wasi-outbound-http = { path = "crates/outbound-http" }
 wasmtime = "0.35.3"
+
 
 [target.'cfg(target_os = "linux")'.dependencies]
 # This needs to be an explicit dependency to enable
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -1,9 +1,10 @@
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use bindle::Id;
 use clap::Parser;
 use hippo::{Client, ConnectionInfo};
 use hippo_openapi::models::ChannelRevisionSelectionStrategy;
 use semver::BuildMetadata;
+use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use spin_http_engine::routes::RoutePattern;
 use spin_loader::local::config::{RawAppManifest, RawAppManifestAnyVersion};
@@ -118,6 +119,12 @@ pub struct DeployCommand {
     pub redeploy: bool,
 }
 
+#[derive(Deserialize, Serialize, Debug)]
+struct LoginHippoError {
+    title: String,
+    detail: String,
+}
+
 impl DeployCommand {
     pub async fn run(self) -> Result<()> {
         let cfg_any = spin_loader::local::raw_manifest_from_file(&self.app).await?;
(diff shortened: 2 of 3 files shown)


src/commands/deploy.rs
```diff
@@ -147,7 +147,8 @@ impl DeployCommand {
             self.hippo_username.clone(),
             self.hippo_password.clone(),
         )
-        .await?
+        .await
+        .context("Problem login Hippo")?
```
itowlson, line 151: A couple of things here:

* First, and apologies for nit-picking the English here, but this should be phrased as something like "Problem logging into Hippo"
* Second, it would be nice to extract the reason from the Hippo failure response and present that.  What this does is address the second part of #614 (say that the error comes from Hippo rather than from Bindle), which is good, but the reason is still buried in the raw JSON.  What I'd suggest is that we parse the JSON, look at either the `title` or `detail` elements (@bacongobbler can probably recommend which) and output that as part of the message.  We can still dump the JSON, but this way we get a top-level message which is still precise but is also more readable.

Hope this makes sense, happy to discuss further!

itowlson (changes requested):
Thanks for tacking this!  Unfortunately, we are ideally looking for a bit more than this - what I'm hoping for is to have the top level message reflect the error detail as well as indicating that it's Hippo.  That said, I'm not sure what messages the Hippo client returns, and maybe we can't get anything more than "Login failed", in which case fair enough.

bacongobbler:
`format_response_error` is meant more for API-level response errors, like when you're adding an App or a Channel to the API.

I'd look at [HandleLoginFailedException](https://github.com/deislabs/hippo/blob/66b1bfe3a8f3b9a52d565fa8e20049026a2a1ce9/src/Web/Filters/ApiExceptionFilterAttribute.cs#L65) for implementation details as well as what's being returned in this scenario.

lidongjies:
@itowlson 

I have parsed JSON and displayed the 'detailed' field, because neither the 'title' nor the 'detailed' field indicates that the problem occurs in Hippo, so I keep 'problem login into Hippo'.

When using an unregistered Hippo username.
> Error: Problem login into Hippo: Login failed: Account does not exist.

When Hippo password is incorrect.
> Error: Problem login into Hippo: Login failed:

src/commands/deploy.rs
```diff
@@ -147,11 +154,13 @@ impl DeployCommand {
...
-        .await?
-        .token
+        .await
         {
-            Some(t) => t,
-            None => String::from(""),
+            Ok(token_info) => token_info.token.unwrap_or_default(),
+            Err(err) => {
+                let error: LoginHippoError = serde_json::from_str(err.to_string().as_str())?;
+                bail!("Problem login into Hippo: {}", error.detail);
```
itowlson, line 162: Minor grammar tweak:

```suggestion
                bail!("Problem logging into Hippo: {}", error.detail);
```

Cargo.toml
```diff
@@ -48,6 +49,7 @@ url = "2.2.2"
 wasi-outbound-http = { path = "crates/outbound-http" }
 wasmtime = "0.35.3"
 
+
```
itowlson, line 52: Super pedantic: I don't think the extra line is needed?

src/commands/deploy.rs
itowlson: You note that if the password is incorrect, the `detail` is blank, so you end up with a trailing colon here (`Problem logging into Hippo: `).  I'd suggest checking for the empty string and dropping the colon (`Problem logging into Hippo`) so it looks a bit tidier in this case.

To avoid this making the `Err` clause so big that it derails the reading flow, you could hive this off into a helper function e.g.

```
Err(err) => bail!(format_login_error(err)?),

// ...

fn format_login_error(err: &Error) -> anyhow::Result<String> {
  let error: LoginHippoError = /* as you have it now */;
  if error.detail.is_empty() {
    // message without colon
  } else {
    // message with colon and detail
  }
}
```

itowlson (changes requested):
This looks good - thanks for addressing that!  There's still one persnickety amendment to the grammar (for which I apologise, but it's user facing), and I also had a suggestion regarding the case where there's no detail.  In substance it's spot on though - thanks again!

Cargo.toml
lidongjies: 👌 Of course, I'll delete this blank line

src/commands/deploy.rs
lidongjies: The 'detail' field is not blank, and the trailing colon is part of the' detail' field. 

When using an unregistered Hippo username.
> { "title": "Login failed.", "detail": "Login failed: Account does not exist." }

When Hippo password is incorrect.  
> { "title": "Login failed.", "detail": "Login failed:" }

lidongjies:
```rust
fn format_login_error(err: &anyhow::Error) -> anyhow::Result<String> {
    let error: LoginHippoError = serde_json::from_str(err.to_string().as_str())?;
    if error.detail.ends_with(": ") {
        Ok(format!(
            "Problem logging into Hippo: {}",
            error.detail.replace(": ", ".")
        ))
    } else {
        Ok(format!("Problem logging into Hippo: {}", error.detail))
    }
}
```

When using an unregistered Hippo username.

> Error: Problem login into Hippo: Login failed: Account does not exist.

When Hippo password is incorrect.

> Error: Problem login into Hippo: Login failed.

itowlson: Oops, sorry, my misunderstanding - we should tweak this on the Hippo side.  I like your solution though!

itowlson: (that is to say.... we should do what you suggest even if it's only needed temporarily)

itowlson (approved):
Looks good - thanks for your patience and your contribution!

PR #648, opened 2022-07-16
opt: remove needless clone

csmoe:


itowlson (changes requested):
I'm not very au fait with more sophisticated Rust idioms, but I found the use of Borrow to make a HashSet behave like a HashMap a bit disorienting (although it's ingenious and I learned something from it, so thanks!).  If this is a well-known technique then I'm okay with it; otherwise, my preference would be to go back to the HashMap-based implementation and see if there are other ways we can avoid the clone of `source` (or optimise it e.g. using an Arc).

The other clarity improvements around parameter renaming are very welcome though!

crates/engine/src/lib.rs
```diff
@@ -140,16 +147,15 @@ impl<T: Default + 'static> Builder<T> {
     #[instrument(skip(self))]
     pub async fn build(mut self) -> Result<ExecutionContext<T>> {
         let _sloth_warning = warn_if_slothful();
-        let mut components = HashMap::new();
+        let mut components = HashSet::with_capacity(self.config.components.len() / 2);
```
dicej, line 150: Why the `/ 2` here?

dicej (commented):
I tend to agree with @itowlson on this: it's a really cool technique (which I'll definitely consider using next time I'm looking to optimize a scenario like this), but I'm not sure it's worthwhile in this case given that this code is only run once at startup for a given app.  Also, any performance gain here is likely to be tiny compared to the work done in `Linker::instantiate_pre`.

For cold path code like this, I'd say simpler is better, even if it involves a few extra allocations.

csmoe:
This patch is trivial on performance, close.(Thanks for your reviewing)

PR #649, opened 2022-07-18
Update README.md

Paul-Bird:
Signed-off-by: Paul-Bird <34755807+Paul-Bird@users.noreply.github.com>


itowlson:
Kia ora @Paul-Bird and thanks for the contribution.  Could you say a few words about why you'd like to see this change?  As far as I can see the existing `download` URL matches the GitHub "releases" page and I'm not sure of the motivation for changing it to use the `tag` URL.  Thanks!

PR #652, opened 2022-07-19
fix: Rename top-level config section to variables

fibonacci1729:
This PR renames the top-level config section in the spin.toml to variables.
Updates the examples and documentation.

NOTE: This is a breaking change.


lann:
Update docs?

lann (approved):

Issue #658, opened 2022-07-25
Missing protocol in `allowed_http_hosts` gives misleading error at runtime


itowlson:
Create a `spin.toml` with the following:

```
allowed_http_hosts = [ "spin.fermyon.dev" ]
```

Now make an outbound HTTP call to `https://spin.fermyon.dev`.

It fails with error 2, "invalid URL."

This happens because the allow checks attempts to parse all allowed hosts as URLs, and in this case fails with "relative URL without base".  This is mapped to the InvalidUrl error from `http-types.wit` at https://github.com/fermyon/spin/blob/b54bb8639aff95636580b9431c692b0068bf7f26/crates/outbound-http/src/lib.rs#L48.  So of course the user tries to figure out what was wrong with _their_ URL.

We should perform any required validation of `allowed_http_hosts` when loading the `spin.toml` file and fail immediately, rather than waiting until runtime and then returning an error that implies the request was bad.

lidongjies:
Need to do the same check for allowed_http_hosts before pushing to bindle server?

itowlson:
That's a good idea.  We could (and probably should) do it when loading from a bindle, but at that point it's unfixable...

Mossaka:
Do we have any forms of manifest verification in load time? I am happy to take this task.

Mossaka:
Figuring out at which layer to validate would be hard. I would improve the error messages for this issue and leave a more proper validation infras to be designed later.

PR #659, opened 2022-07-25
feat(loader): validate wasm config

FrankYang0529:
resolves https://github.com/fermyon/spin/issues/658

### DIFF
--- a/crates/loader/src/bindle/mod.rs
+++ b/crates/loader/src/bindle/mod.rs
@@ -10,9 +10,12 @@ mod connection;
 /// Bindle helper functions.
 mod utils;
 
-use crate::bindle::{
-    config::{RawAppManifest, RawComponentManifest},
-    utils::{find_manifest, parcels_in_group},
+use crate::{
+    bindle::{
+        config::{RawAppManifest, RawComponentManifest},
+        utils::{find_manifest, parcels_in_group},
+    },
+    validation::validate_allowed_http_hosts,
 };
 use anyhow::{anyhow, Context, Result};
 use bindle::Invoice;
@@ -67,6 +70,7 @@ async fn prepare(
 
     let mut config_root = raw.config.take().unwrap_or_default();
     for component in &mut raw.components {
+        validate_allowed_http_hosts(&component.wasm.allowed_http_hosts)?;
         if let Some(config) = component.config.take() {
             let path = component.id.clone().try_into().with_context(|| {
                 format!("component ID {:?} not a valid config path", component.id)
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -21,7 +21,7 @@ use spin_manifest::{
 use std::{path::Path, str::FromStr, sync::Arc};
 use tokio::{fs::File, io::AsyncReadExt};
 
-use crate::bindle::BindleConnectionInfo;
+use crate::{bindle::BindleConnectionInfo, validation::validate_allowed_http_hosts};
 
 /// Given the path to a spin.toml manifest file, prepare its assets locally and
 /// get a prepared application configuration consumable by a Spin execution context.
@@ -38,6 +38,7 @@ pub async fn from_file(
         .absolutize()
         .context("Failed to resolve absolute path to manifest file")?;
     let manifest = raw_manifest_from_file(&app).await?;
+    validate_raw_app_manifest(&manifest)?;
 
     prepare_any_version(
         manifest,
@@ -88,6 +89,20 @@ fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
             bail!("cannot have duplicate component IDs: {}", id);
         } else {
             ids.push(id);
+        }
+    }
+    Ok(())
+}
+
+/// Validate fields in raw app manifest
+pub fn validate_raw_app_manifest(raw: &RawAppManifestAnyVersion) -> Result<()> {
+    match raw {
+        RawAppManifestAnyVersion::V1(raw) => {
+            let _ = raw
+                .components
+                .iter()
+                .map(|c| validate_allowed_http_hosts(&c.wasm.allowed_http_hosts))
+                .collect::<Result<Vec<_>>>()?;
         }
     }
     Ok(())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3767,6 +3767,7 @@ dependencies = [
  "tracing-futures",
  "tracing-subscriber 0.3.9",
  "walkdir",
+ "wasi-outbound-http",
 ]
 
 [[package]]
--- a/crates/loader/Cargo.toml
+++ b/crates/loader/Cargo.toml
@@ -30,3 +30,4 @@ tracing = { version = "0.1", features = [ "log" ] }
 tracing-futures = "0.2"
 tracing-subscriber = { version = "0.3.7", features = [ "env-filter" ] }
 walkdir = "2.3.2"
+wasi-outbound-http = { path = "../outbound-http" } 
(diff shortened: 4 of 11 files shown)


crates/loader/src/local/mod.rs
```diff
@@ -93,6 +98,37 @@ fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
     Ok(())
 }
 
+fn validate_raw_app_manifest(raw: &RawAppManifestAnyVersion) -> Result<()> {
+    match raw {
+        RawAppManifestAnyVersion::V1(raw) => {
```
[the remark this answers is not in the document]
itowlson: I'm also wondering if we should set up more general validation infrastructure..

crates/loader/src/local/mod.rs
```diff
@@ -93,6 +98,37 @@ fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
...
+        if domains
+            .iter()
+            .any(|domain| domain == wasi_outbound_http::ALLOW_ALL_HOSTS)
+        {
+            return Ok(());
+        }
+        let _ = domains
+            .iter()
+            .map(|d| {
+                Url::parse(d).with_context(|| format!("can't parse {} in allowed_http_hosts", d))
```
Mossaka, line 125:
```suggestion
                Url::parse(d).with_context(|| format!("Can't parse {} in allowed_http_hosts", d))
```

crates/loader/src/local/mod.rs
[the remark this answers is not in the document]
Mossaka: I agree that this would not be the best place to put validator, as each new version would require some copy and paste of previous version's validations. Building a general validation infras would be hard, but I like the idea. It needs to be versioned too.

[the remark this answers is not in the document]
FrankYang0529: Hi @itowlson and @Mossaka, thanks for the suggestion. Currently, if we want to set validation on the `Application` object, only `up` and `trigger` commands will be validated. Other commands like `build` or `deploy` use `RawAppManifest` directly. For validating `spin.toml` before runtime, I think adding validation in the `Application` object is enough. If you all agree with this, I can update this PR later.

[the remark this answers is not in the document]
itowlson: `build` shouldn't care about configuration validity, unless used with `--up`, and I'd expect it that part to go through `Application`.  Good point about `deploy` (and `publish`) though, we'd want to catch it there and they don't build `Application` abstractions.

Given that, the best course is, I think, to go ahead with the validation where you have it, and raise an issue to revisit the consolidation issue as part of a general validation infrastructure, which might require more invasive changes.  @Mossaka does that make sense to you?

crates/loader/src/bindle/mod.rs
```diff
@@ -169,3 +171,21 @@ fn info(raw: &RawAppManifest, invoice: &Invoice, url: &str) -> ApplicationInform
...
+        if domains
+            .iter()
+            .any(|domain| domain == wasi_outbound_http::ALLOW_ALL_HOSTS)
+        {
+            return Ok(());
+        }
+        let _ = domains
+            .iter()
+            .map(|d| {
+                Url::parse(d).with_context(|| format!("can't parse {} in allowed_http_hosts", d))
```
itowlson, line 186:
```suggestion
                Url::parse(d).with_context(|| format!("Can't parse {} in allowed_http_hosts", d))
```

crates/loader/src/local/mod.rs
```diff
@@ -93,6 +98,37 @@ fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
...
+                .components
+                .iter()
+                .map(|c| validate_raw_wasm_config(&c.wasm))
+                .collect::<Result<Vec<_>>>()?;
+        }
+    }
+    Ok(())
+}
+
+fn validate_raw_wasm_config(raw: &RawWasmConfig) -> Result<()> {
```
itowlson, line 114: This is the same function as in the bindle loader.  I know the `RawWasmConfig` types are different between the two, but the `allowed_http_hosts` types are the same.  Could we have a helper function that takes `&Option<Vec<String>>` (or similar) and has all the logic in it, and call that function from both places?

crates/loader/src/local/mod.rs
```diff
@@ -60,6 +64,7 @@ pub async fn raw_manifest_from_file(app: &impl AsRef<Path>) -> Result<RawAppMani
         .with_context(|| anyhow!("Cannot read manifest file from {:?}", app.as_ref()))?;
 
     let manifest: RawAppManifestAnyVersion = toml::from_slice(&buf)?;
+    validate_raw_app_manifest(&manifest)?;
```
itowlson, line 67: Placing this here will cause it to run on `build`, which normally only cares about compilation... I'm not sure of a good solution to this.  We could place this call in the `from_file` and `from_bindle` wrappers used by `up`, and call it manually from `deploy`.  But then it starts to feel a bit like whack-a-mole...

itowlson (changes requested):
This looks good apart from a couple of minor changes, but it does hit a snag in our local loader flow.  If you put the validation in the raw manifest deserialiser, it gets run during `build`, which feels iffy.  But if you _don't_ put it in the deserialiser, it _doesn't_ get run during `deploy` or `bindle prepare/publish`, which is even worse.

We can bodge around this by manually calling it from `expand_manifest`, and I think that would work.  But then we have multiple places where validation happens, and as the number of entry points grows it gets hard to make sure we're catching it everywhere.  So we're probably going to need to do some rework on the loading pipeline.

My suggestion for now would be to remove the validation call from `raw_manifest_from_file` (which really is meant to do nothing more than deserialise the TOML) and call it instead from `from_file` and `expand_manifest`.  I think that would pick up all the cases.

As always I'm very open to discussion - I am really not entirely sure what the right thing to do is!

crates/loader/src/local/mod.rs
FrankYang0529: Updated it. Thank you.

crates/loader/src/bindle/mod.rs
FrankYang0529: Updated it. Thank you.

crates/loader/src/local/mod.rs
FrankYang0529: Thanks for your suggestion. I add a `validate_allowed_http_hosts` function and remove `validate_raw_wasm_config` in both `local` and `bindle` loader.

crates/loader/src/local/mod.rs
FrankYang0529: Yeah, I agree with it. We should have a place to transfer `RawAppManifest` to `Application.` Also, other functions/commands should rely on the `Application` object, so we don't need to put the validating function everywhere.

crates/loader/src/assets.rs
```diff
@@ -103,6 +104,24 @@ pub(crate) async fn change_file_permission(
     Ok(())
 }
 
+pub fn validate_allowed_http_hosts(http_hosts: &Option<Vec<String>>) -> Result<()> {
```
itowlson, line 107: nit: I don't think this belongs in the `assets` module; if there's no existing home for it then maybe create a new `validation` module?

itowlson (changes requested):
One final request but it's an easy one I hope - conceptually HTTP host validation is not part of the assets subsystem - I can't see an existing home so maybe create a new file/module for it?  Thanks!  Otherwise looks good to go.

crates/loader/src/assets.rs
FrankYang0529: Yeah, add a `validation` module for it. Thank you.

itowlson (approved):
Awesome, thanks!  LGTM

PR #677, opened 2022-08-07
Postgres data types and error reporting

itowlson:
This PR rounds out the initial Postgres implementation in two main ways:

1. Support data types other than strings
2. Categorise errors and provide error messages

As an implementation detail, it also proposes a `rdbms-types.wit` for types that are likely to be the same across multiple relational databases.  This is arguably premature, possibly even undesirable, though it could easily be reversed without a breaking change (bindgen folds includes, so things from this WIT are surfaced as they had been declared in "outbound-pg").

Feedback very much sought!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2474,6 +2474,7 @@ dependencies = [
  "postgres",
  "spin-engine",
  "spin-manifest",
+ "tracing",
  "wit-bindgen-wasmtime",
 ]
 
--- a/crates/outbound-pg/Cargo.toml
+++ b/crates/outbound-pg/Cargo.toml
@@ -11,4 +11,5 @@ anyhow = "1.0"
 postgres = { version = "0.19.3" }
 spin-engine = { path = "../engine" }
 spin-manifest = { path = "../manifest" }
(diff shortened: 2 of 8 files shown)


itowlson:
Adding the do-not-merge tag because it looks like parameters need to be typed too.

itowlson:
Okay, parameters are now working.  (The reason for having a separate parameter value type, by the way, is that even if they're the same, `wit-bindgen` generates two different types, at least in Rust - I assume because in-position and out-position.  It felt cleaner to make that distinction in the WIT where we could control the naming.)

And apologies for all the build failure noise spam.

lann:
Someone else's take: https://github.com/theduke/wasi-sql/blob/main/schema/sql_v1_alpha1.wit

itowlson:
Awesome, thanks @lann!

itowlson:
How do we want to take this forward?  Do we want to switch to the alpha proposal, or go with something like this for the time being and adopt the more complex and capable proposal once it's closer to standardisation?

fibonacci1729 (approved):

fibonacci1729:
LGTM

PR #681, opened 2022-08-10
feat: feature gate the e2e tests that require dependencies

### LINKED ISSUE #626: Test: Feature gate the e2e tests
(Where e2e tests refer to all of the currently-named integration tests that rely on dependencies such as the `bindle-server`, `nomad` and `Hippo.Web` binaries.)

Can we feature gate the newly added tests? Something like,
  ```rust
  #[cfg(feature = "e2e-tests")]
  mod e2e {
       ... all new test code goes here ...
  }
  ```

  Then adding the feature to Cargo.toml:
  ```toml
  ...
  [features]
  e2e-tests = []
  ```

  Add a make target for `test-e2e` that invokes the test with the e2e-tests feature enabled.
  
  Why? This way the original integration tests can be run locally until we figure how to
  ensure the installation of e2e prerequisites.

_Originally posted by @fibonacci1729 in https://github.com/fermyon/spin/issues/607#issuecomment-1179282331_

kate-goldenring:
Moves e2e integration tests that require dependencies behind an `e2e-tests` feature, so the dependencies (`bindle-server`, `nomad `and `Hippo.Web`) do not need to be installed to run `make test` or `make test-integration`. 

Running `make test-e2e` will run all tests including ones with dependencies (after checking the dependencies exist).

fixes #626 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,7 +124,7 @@ jobs:
       - name: Cargo Test
         run: |
           make test-unit
-          make test-integration
+          make test-e2e
         env:
           RUST_LOG: spin=trace
 
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1,7 +1,7 @@
 #[cfg(test)]
 mod integration_tests {
     use anyhow::{Context, Result};
-    use hyper::{header::HeaderName, Body, Client, Response};
+    use hyper::{Body, Client, Response};
     use spin_loader::local::{
         config::{RawAppManifestAnyVersion, RawModuleSource},
         raw_manifest_from_file,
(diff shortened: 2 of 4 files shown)


vdice (commented):
This works great for me when I run locally!  My only suggested change is to add a `make test-e2e` line to https://github.com/fermyon/spin/blob/main/.github/workflows/build.yml#L124-L129 now that we have a new target for these.

lann:
The clippy errors are due to the rust toolchain being updated. I fixed it as part of #684

fibonacci1729 (approved):
Thanks Kate, this looks great! 🎉

kate-goldenring:
> This works great for me when I run locally! My only suggested change is to add a `make test-e2e` line to https://github.com/fermyon/spin/blob/main/.github/workflows/build.yml#L124-L129 now that we have a new target for these.

Thank you @vdice. Essential step i missed. Replaced `make test-integration` with `make test-e2e` in 3981763

.github/workflows/build.yml
```diff
@@ -124,7 +124,7 @@ jobs:
       - name: Cargo Test
         run: |
           make test-unit
-          make test-integration
+          make test-e2e
```
vdice, line 127: Do we have a gap in test coverage if we don't run `make test-integration` along with these two others?  Or do they run now as part of `test-e2e`?

tests/integration.rs
```diff
@@ -1,314 +1,602 @@
...
-    const NOMAD_BINARY: &str = "nomad";
-    const HIPPO_BINARY: &str = "Hippo.Web";
 
-    const BINDLE_SERVER_PATH_ENV: &str = "SPIN_TEST_BINDLE_SERVER_PATH";
-    const BINDLE_SERVER_BASIC_AUTH_HTPASSWD_FILE: &str = "tests/http/htpasswd";
-    const BINDLE_SERVER_BASIC_AUTH_USER: &str = "bindle-user";
-    const BINDLE_SERVER_BASIC_AUTH_PASSWORD: &str = "topsecret";
+    // This module consist of all integration tests that require dependencies such as bindle-server, nomad, and Hippo.Web to be installed.
+    #[cfg(feature = "e2e-tests")]
+    mod e2e_tests {
```
lann, line 27: Would it make sense to break these out into a separate file?

.github/workflows/build.yml
kate-goldenring: They run as part of `test-e2e`. This make handle just also turns on the e2e-tests feature:
```
.PHONY: test-integration
test-integration:
	RUST_LOG=$(LOG_LEVEL) cargo test --test integration --no-fail-fast -- --nocapture --include-ignored

.PHONY: test-e2e
test-e2e:
	RUST_LOG=$(LOG_LEVEL) cargo test --test integration --features e2e-tests --no-fail-fast  -- integration_tests::test_dependencies --nocapture 
	RUST_LOG=$(LOG_LEVEL) cargo test --test integration --features e2e-tests --no-fail-fast -- --skip integration_tests::test_dependencies --nocapture --include-ignored
```

vdice: Nice, thank you!

tests/integration.rs
kate-goldenring: I chose to use a nested module since they are still integration tests and from my understanding our goal is for this to be a temporary gating until we have a setup were the dependencies can be installed at the beginning of integration testing (maybe as in the Fermyon installer): https://github.com/fermyon/spin/issues/626. Open to separating out for cleanliness if that is more desired.

kate-goldenring: There is also a lot of shared structures (`SpinTestController`) and helper functions (`assert_status` etc) that the e2e_tests need that would need to be exported

lann: Okey dokey :+1:.

> from my understanding our goal is for this to be a temporary gating until we have a setup were the dependencies can be installed at the beginning of integration testing

I think we'll always want to keep tests that require external infrastructure separate from self-contained integration tests. We've even discussed splitting them out into a separate repo, but I think that is out of scope for this PR.

Issue #599, opened 2022-06-23
CI: Bump template versions and commit changes on tag pushes


vdice:
In https://github.com/fermyon/spin/pull/586 we updated the [release-process.md](https://github.com/fermyon/spin/blob/main/docs/content/release-process.md) doc with steps around bumping the SDK versions on a new Spin release and then following up with bumps to the template projects and committing the resulting changes.

Let's see if we can just have this be another automated step in the [release.yml](https://github.com/fermyon/spin/blob/main/.github/workflows/release.yml) GH workflow.

FrankYang0529:
I would like to take this, but I have one question: which GitHub account should we use to commit in [step 7](https://spin.fermyon.dev/release-process)? Thank you.

vdice:
@FrankYang0529 Excellent!  Which GH action are you thinking of using?  Can we utilize the usual/default `secrets.GITHUB_TOKEN` for creating a Pull Request from an action?

FrankYang0529:
I would like to use [git-auto-commit-action](https://github.com/stefanzweifel/git-auto-commit-action). Let me check whether I can use `GITHUB_TOKEN`.

radu-matei:
We have @fermybot for this if we need an account to commit to the repository.

PR #685, opened 2022-08-11
feat(ci): create sdk and template tag in release action

FrankYang0529:
resolve https://github.com/fermyon/spin/issues/599

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -212,3 +212,77 @@ jobs:
           body: |
             This is a "canary" release of the most recent commits on our main branch. Canary is **not stable**.
             It is only intended for developers wishing to try out the latest features in Spin, some of which may not be fully implemented.
+
+  create-go-sdk-tag:
+    name: create tag sdk/go/v*
+    runs-on: ubuntu-latest
+    needs: build
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: set the tag sdk/go/v*
+        shell: bash
+        run: echo "GO_SDK_TAG=sdk/go/${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+
+      - name: Tag sdk/go/v* and push it
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git tag ${{ env.GO_SDK_TAG }}
+          git push origin ${{ env.GO_SDK_TAG }}
+
+  commit-and-create-template-tag:
+    name: Change sdk version in templates and create a new tag spin/templates/v*
+    runs-on: ubuntu-latest
+    needs: create-go-sdk-tag
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: set the spin tag
+        shell: bash
+        run: echo "SPIN_TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+
+      - name: set the tag spin/templates/v*
+        shell: bash
+        run: |
+          IFS=. read -r major minor patch <<< "${{ env.SPIN_TAG }}"
+          echo "TEMPLATE_TAG=spin/templates/$major.$minor" >> $GITHUB_ENV
+
+      - name: Change sdk version
+        shell: bash
+        run: |
+          cd templates
+          SDK_VERSION=${{ env.SPIN_TAG }} make
+
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v5
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "feat(templates): update sdk to ${{ env.SPIN_TAG }}"
+          commit_options: '-s -S'
+          commit_user_name: fermybot
+          commit_user_email: 103076628+fermybot@users.noreply.github.com
+          commit_author: fermybot <103076628+fermybot@users.noreply.github.com>
+          file_pattern: templates/*
+          branch: main
+          skip_dirty_check: true
+          skip_fetch: true
+          skip_checkout: true
+
+      - name: Tag spin/templates/v* and push it
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git tag ${{ env.TEMPLATE_TAG }} -f
+          git push origin ${{ env.TEMPLATE_TAG }} -f
--- a/docs/content/release-process.md
+++ b/docs/content/release-process.md
@@ -17,23 +17,7 @@ To cut a release of Spin, you will need to do the following:
 1. Before proceeding, verify that the merge commit on `main` intended to be
    tagged is green, i.e. CI is successful
 1. Create a new tag with a `v` and then the version number (`v0.3.1`)
-1. Create a tag for the Go SDK on the same commit as above, e.g.
-   `sdk/go/v0.3.1`
-1. Push the tags to origin on GitHub
-   - The `v0.3.1` tag push will trigger a
-   [release build](https://github.com/fermyon/spin/actions/workflows/release.yml)
-1. While waiting for the release, bump the version for the project templates
-   under the `templates` directory via the `make` target:
-   ```console
-      cd templates
-      SDK_VERSION=v0.3.1 make
-   ```
-   - Create a follow-up PR with the updated files
-1. Async step: After the follow-up PR from the previous step is merged, create
-   a tag for the templates at the merge commit.  Tag should be named e.g.
-   `spin/templates/v0.3`
-   - Note the `v0.3` - the tag only needs to be created on minor releases and
-   this step should be skipped on patch releases.
+1. The Go SDK tag `sdk/go/v0.3.1` and template tag `spin/templates/v0.3` will be created in `release` [action](https://github.com/fermyon/spin/actions/workflows/release.yaml).
 1. When the `release`
    [action](https://github.com/fermyon/spin/actions/workflows/release.yaml)
    completes, binary artifacts and checksums will be automatically uploaded.


FrankYang0529:
I will check how to commit with GPG tomorrow.

ref: https://gist.github.com/swinton/03e84635b45c78353b1f71e41007fc7c

.github/workflows/release.yml
```diff
@@ -212,3 +212,54 @@ jobs:
...
+
+  commit-and-create-template-tag:
+    name: Change sdk version in templates and create a new tag spin/templates/v*
+    runs-on: ubuntu-latest
+    needs: create-go-sdk-tag
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: set the tag sdk/go/v*
```
vdice, line 244:
```suggestion
      - name: set the tag spin/templates/v*
```

vdice: The only thing I'm unsure about is if we're okay pushing the `spin/templates/v*` tag for all `v*` releases, since the process mentions: 

```
Note the `v0.3` - the tag only needs to be created on minor releases and
   this step should be skipped on patch releases.
```

Let's double-check with @itowlson.  Also, it would be good to double-check that all of the details of this process are still needed for template versioning -- not sure if things have changed.

vdice (commented):
Liking the automation improvements already included!

.github/workflows/release.yml
FrankYang0529: Fixed the error name.

FrankYang0529:
> I will check how to commit with GPG tomorrow.
> 
> ref: https://gist.github.com/swinton/03e84635b45c78353b1f71e41007fc7c

Hi @vdice, I use another way to commit templates change and it can have gpg key of `github-actions[bot]` like the following.


Screen Shot 2022-08-17 at 8 42 30 PM

FrankYang0529:
Sorry, I found that I can't use `FILE_TO_COMMIT: templates/*`. It will create a new file `*`, so I add `WIP` to the title.

FrankYang0529:
If we want to sign templates change commit with GPG key, I think the easiest way is to give `gpg_private_key` in secrets. Other ways like using `gh api` will need to create a tree first to commit multiple files.

https://github.com/crazy-max/ghaction-import-gpg
https://dev.to/bro3886/create-a-folder-and-push-multiple-files-under-a-single-commit-through-github-api-23kc

vdice:
@itowlson can you take a look at this PR when convenient?  I'd like to double-check that we're automating the sdk and template tags correctly (see also https://github.com/fermyon/spin/pull/685#discussion_r943806935)

vdice:
@FrankYang0529 Indeed, it does look cleaner to supply the gpg key rather than constructing the tree.  Are you able to test the gpg key approach on your fork?

FrankYang0529:
@vdice I add "Import GPG key" step and it works fine. The following pictures are results on my fork repo.

Screen Shot 2022-08-18 at 9 40 34 PM

Screen Shot 2022-08-18 at 9 40 43 PM

Could you help to setup `GPG_PRIVATE_KEY` and `PASSPHRASE `? Also, I am unsure what is the email of @fermybot?

Ref: https://github.com/crazy-max/ghaction-import-gpg

FrankYang0529:
The failed workflow checks will be fixed here https://github.com/fermyon/spin/pull/687#discussion_r949181230.

vdice:
@FrankYang0529 Excellent!  Yes, I'll help add the secrets once back next week.  Perhaps by then we can get @itowlson's eyes on the SDK/template tagging bits to make sure all looks good.  This automation is really helpful!

.github/workflows/release.yml
```diff
@@ -212,3 +212,65 @@ jobs:
...
+    name: Change sdk version in templates and create a new tag spin/templates/v*
+    runs-on: ubuntu-latest
+    needs: create-go-sdk-tag
+    if: startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: set the tag spin/templates/v*
+        shell: bash
+        run: echo "TEMPLATE_TAG=spin/templates/${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
```
itowlson, line 246: Does this drop the patch field?

.github/workflows/release.yml
```diff
@@ -212,3 +212,65 @@ jobs:
...
+          git_commit_gpgsign: true
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "feat(templates): update sdk to ${{ env.SPIN_TAG }}"
+          commit_options: '-s -S'
+          commit_user_name: fermybot
+          commit_user_email: bot@fermyon.com
+          commit_author: femybot <bot@fermyon.com>
```
itowlson, line 273: fermybot?

docs/content/release-process.md
```diff
@@ -17,23 +17,7 @@ To cut a release of Spin, you will need to do the following:
...
-      cd templates
-      SDK_VERSION=v0.3.1 make
-   ```
-   - Create a follow-up PR with the updated files
-1. Async step: After the follow-up PR from the previous step is merged, create
-   a tag for the templates at the merge commit.  Tag should be named e.g.
-   `spin/templates/v0.3`
-   - Note the `v0.3` - the tag only needs to be created on minor releases and
-   this step should be skipped on patch releases.
+1. The Go SDK tag `sdk/go/v0.3.1` and template tag `spin/templates/v0.3.1` will be created in `release` [action](https://github.com/fermyon/spin/actions/workflows/release.yaml).
```
itowlson, line 20: nope templates tag needs to be spin/templates/0.3

itowlson (changes requested):

.github/workflows/release.yml
```diff
@@ -212,3 +212,65 @@ jobs:
...
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "feat(templates): update sdk to ${{ env.SPIN_TAG }}"
+          commit_options: '-s -S'
+          commit_user_name: fermybot
+          commit_user_email: bot@fermyon.com
```
vdice, line 272: We can use `103076628+fermybot@users.noreply.github.com` for both emails

.github/workflows/release.yml
FrankYang0529: @fermybot is from https://github.com/fermyon/spin/issues/599#issuecomment-1216102631

.github/workflows/release.yml
FrankYang0529: Updated it. Thank you.

docs/content/release-process.md
FrankYang0529: Updated it. Thank you.

.github/workflows/release.yml
FrankYang0529: Fixed it. Thank you.

vdice:
@FrankYang0529 Thanks so much for addressing all of the requested changes.

I've created a branch from this PR and am testing on my fork.  However, on the `create-go-sdk-tag` step from the [release workflow](https://github.com/vdice/spin/runs/7983617470?check_suite_focus=true), I'm seeing the following:

```
To https://github.com/vdice/spin
 ! [remote rejected] sdk/go/v0.4.1 -> sdk/go/v0.4.1 (refusing to allow a GitHub App to create or update workflow `.github/workflows/release.yml` without `workflows` permission)
 ```

The permissions for the GitHub Action for my fork are already set to read and write:
Screen Shot 2022-08-23 at 4 01 33 PM

Did you encounter the same behavior?  Is a separate GITHUB_TOKEN with more advanced permissions necessary?

**Comment by FrankYang0529** [comment]
@vdice, I didn't change any setting about permission. Do you set any rule in Protected tags?
Screen Shot 2022-08-24 at 10 48 39 PM
Screen Shot 2022-08-24 at 10 49 04 PM
Screen Shot 2022-08-24 at 10 51 26 PM

**Comment by vdice** [comment]
@FrankYang0529 Thanks for comparing settings -- ours are the same.  My first attempt whereupon I hit the aforementioned error, I had created the tag on a branch other than `main`.  This time, I merged the branch to `main` on my fork and then created the tag -- and it worked! https://github.com/vdice/spin/runs/8024077723?check_suite_focus=true

File: .github/workflows/release.yml [hunk 6]
```diff
@@ -212,3 +212,77 @@ jobs:
...
+        shell: bash
+        run: echo "GO_SDK_TAG=sdk/go/${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+
+      - name: Tag sdk/go/v* and push it
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git tag ${{ env.GO_SDK_TAG }}
+          git push origin ${{ env.GO_SDK_TAG }} -f
```
vdice, line 234: I wonder if we want to _not_ use `-f`/`--force`.  I think I'd feel more comfortable erroring out if we're unable to push to origin without forcing.  WDYT?

.github/workflows/release.yml
```diff
@@ -212,3 +212,77 @@ jobs:
...
+          skip_dirty_check: true
+          skip_fetch: true
+          skip_checkout: true
+
+      - name: Tag spin/templates/v* and push it
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git tag ${{ env.TEMPLATE_TAG }} -f
```
vdice, line 287: Same note as above re: `-f`.

vdice (commented):
A question around forcing the tag creation and push...

.github/workflows/release.yml
FrankYang0529: Yeah, sorry, I used `v0.3.0` for testing, so I need to force push to overwrite the tag. I think we don't use the same tag for `sdk/go/v*`, so it's good to remove `-f`. Updated it. Thank you.

.github/workflows/release.yml
FrankYang0529: We may need to keep `-f` here because we drop the patch version. If we push tags like `v0.3.0` and `v0.3.1`, their template tag will be the same `spin/templates/v0.3`.

vdice: Oh, good point 👍

vdice (approved):
LGTM!  Thanks for the big effort on this complex bit of automation @FrankYang0529 🙏

vdice:
Checking in with @itowlson to be sure the requested changes have been made.

itowlson (approved):
Looks right to me (though I admit I don't fully understand all of it!).  Thanks @FrankYang0529 for slogging through all the fiddly details of this - it's going to be a real quality of life improvement.

Issue #672, opened 2022-08-04
Cannot run simple unit test with default Rust template


radu-matei:
The default Rust template for a Spin application sets the default build target in `.cargo/config.toml` as `wasm32-wasi`:

```toml
[build]
target = "wasm32-wasi"
```

This make sense in order to ensure `cargo build` creates a module that Spin can run, but this means that a simple unit test that purely wants to test some logic in the handler cannot be run:

- with a regular `cargo test`, this fails because it would try to instantiate a Wasm binary and run the test:
```
.../target/wasm32-wasi/debug/deps/rust_api-a5d8eced053c53aa.wasm: cannot execute binary file
```

- with `cargo wasi test`, this fails because of host imports that are not available in Wasmtime directly (but are in Spin):

```
wasm32-wasi/debug/deps/rust_api-a5d8eced053c53aa.wasm`
Error: failed to run main module `.../rust-api/target/wasm32-wasi/debug/deps/rust_api-a5d8eced053c53aa.wasm`

Caused by:
    0: failed to instantiate ".../target/wasm32-wasi/debug/deps/rust_api-a5d8eced053c53aa.wasm"
    1: unknown import: `wasi_experimental_http::req` has not been defined
```

The workaround for this is to remove the default target in `.cargo/config.toml` and run `cargo test` — assuming this only tests logic that does not have anything to do with Wasm or Spin functionality.

Because in the default Rust template we have a build command that sets the target to `wasm32-wasi`, do we want to remove the default target in the Cargo configuration file so testing with `cargo test` works for simple unit tests?

FrankYang0529:
This is similar to what I encountered in https://github.com/fermyon/spin-fileserver/pull/15. Probably, we can use `cargo test --target=$$(rustc -vV | sed -n 's|host: ||p')` as @lann mentioned in https://github.com/fermyon/spin-fileserver/pull/15#issuecomment-1190355203.

radu-matei:
Given that all our Rust templates explicitly build to `--target wasm32-wasi`, could we remove `.cargo/config.toml` from the templates?

vdice:
It sounds like we've converged on https://github.com/fermyon/spin/issues/672#issuecomment-1216103487 as being the approved fix here?  Shall we move into the Backlog?

PR #691, opened 2022-08-16
this target is already set in .cargo/config.toml

tshepang:
I suspect the docs would have to be updated if this change is accepted.


radu-matei:
This is something we have talked about in #672 as well — and there, we explored removing `.cargo/config.toml` instead.

What do you think?

tshepang:
After reading that #672, I agree that removing `.cargo/config.toml` is a better change.

PR #695, opened 2022-08-17
feat: add accept-defaults to new templates

### LINKED ISSUE #591: spin new: flag to accept defaults where available
At the moment, spin new prompts for fields like description. If the fields it prompts for are not required by the manifest file, it'd be nice to disable the prompt and leave those optional fields empty. I think this will give us a more streamlined spin new experience.

karthik2804:
Add optional argument `--accept-defaults` to `spin new` to accept defaults where available from the template. The values file and value mentioned in the CLI still take precedence.

closes #591.

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -387,7 +387,7 @@ impl InstallationResults {
 
 #[cfg(test)]
 mod tests {
-    use std::{fs, path::PathBuf};
+    use std::{collections::HashMap, fs, path::PathBuf};
 
     use tempfile::tempdir;
 
@@ -629,6 +629,7 @@ mod tests {
             output_path: output_dir.clone(),
             name: "my project".to_owned(),
             values,
+            accept_defaults: false,
         };
 
         template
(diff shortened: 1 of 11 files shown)


src/commands/new.rs
```diff
@@ -52,6 +57,18 @@ impl NewCommand {
                 Some(file) => values_from_file(file.as_path()).await?,
                 None => HashMap::new(),
             };
+            if template.is_some() && self.accept_defaults {
+                let parameters = template.as_ref().unwrap().parameters();
```
radu-matei, line 61: I know we are checking for `.is_some`, but I would still like to avoid calling `.unwrap()` here if possible.

itowlson (changes requested):
Thanks for taking this on.  I'd strongly encourage you to dive a little deeper into the templates crate - although it requires more code reading, I believe you'll end up at an existing function where the code has both the parameter object and designated value to hand, and is performing the existing "value > prompt" logic.  Extending that to "value > accepted default > prompt" should (I hope) be a couple of lines.

Happy to chat or pair I'm not explaining myself well...

src/commands/new.rs
```diff
@@ -34,6 +34,11 @@ pub struct NewCommand {
     /// file.
     #[clap(long = "values-file")]
     pub values_file: Option<PathBuf>,
+
+    /// An optional argument that allows to skip prompts for the manifest file
+    /// by accepting the defaults if avaiblable on the template
```
itowlson, line 39: typo 
```suggestion
    /// by accepting the defaults if available on the template
```

crates/templates/src/manager.rs
```diff
@@ -629,6 +629,47 @@ mod tests {
...
+        manager
+            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .await
+            .unwrap();
+
+        let template = manager.get("http-rust").unwrap().unwrap();
+
+        let dest_temp_dir = tempdir().unwrap();
+        let output_dir = dest_temp_dir.path().join("myproj");
+        let values = [("http-path".to_owned(), "/path/...".to_owned())]
```
itowlson, line 665: Suggestion: _don't_ send in a value for `http-path`, and assert that the `spin.toml` correctly gets the default value.

crates/templates/src/run.rs
```diff
@@ -282,7 +284,12 @@ impl Run {
     fn populate_parameter_interactive(&self, parameter: &TemplateParameter) -> Option<String> {
         match self.options.values.get(parameter.id()) {
             Some(s) => Some(s.clone()),
-            None => crate::interaction::prompt_parameter(parameter),
+            None => {
+                if let (Some(s), true) = (parameter.default_value(), self.options.accept_defaults) {
```
itowlson, line 288: I think a `match` here might be Rustier:

```
None => match (self.options.acccept_defaults, parameter.default_value()) {
    (true, Some(v)) => Some(v.to_string()),
    _ => crate::interaction::prompt_parameter(parameter),
}
```

crates/templates/src/run.rs
```diff
@@ -298,7 +305,12 @@ impl Run {
     fn populate_parameter_silent(&self, parameter: &TemplateParameter) -> anyhow::Result<String> {
         match self.options.values.get(parameter.id()) {
             Some(s) => Ok(s.clone()),
-            None => Err(anyhow!("Parameter '{}' not provided", parameter.id())),
+            None => {
+                if let (Some(s), true) = (parameter.default_value(), self.options.accept_defaults) {
```
itowlson, line 309: Again suggest using a `match`

itowlson (changes requested):
This looks good - thanks!  Couple of minor recommendations, but otherwise good to go!

itowlson (approved):
LGTM!

radu-matei (approved):

Issue #700, opened 2022-08-19
`spin build` should say if there is nothing to build


lann:
```console
$ spin build
Successfully ran the build command for the Spin components.
```

...is a little confusing when there are no build commands to run.

jim4067:
Would love to try hacking on this.

radu-matei:
@jim4067 — that would be great!
Here is the part in the build logic that you can start looking at — https://github.com/fermyon/spin/blob/d81a37b4b0349f7286c43c27ee5b60d202ac3426/crates/build/src/lib.rs#L12-L31

Let us know if there is anything we can help with.

Thanks!

PR #703, opened 2022-08-20
feat: show when there are not builds to run

jim4067:
For #700 

here's how I implemented this.
 
I check this directory `target/wasm32-wasi/release`. If 10 seconds have elapsed after it was modified, then I assume that there are no builds happening.

### DIFF
--- a/crates/build/Cargo.toml
+++ b/crates/build/Cargo.toml
@@ -12,3 +12,4 @@ tokio = { version = "1.11", features = [ "full" ] }
 spin-loader = { path = "../loader" }
 subprocess = "0.2.8"
 tracing = { version = "0.1", features = [ "log" ] }
+toml = "0.5.9"
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -6,6 +6,7 @@ use anyhow::{bail, Context, Result};
 use path_absolutize::Absolutize;
 use spin_loader::local::config::{RawAppManifest, RawComponentManifest};
 use std::path::{Path, PathBuf};
+use std::{fs, time::Duration};
 use subprocess::{Exec, Redirection};
 use tracing::log;
 
@@ -26,7 +27,22 @@ pub async fn build(app: RawAppManifest, src: &Path) -> Result<()> {
         }
     }
 
-    println!("Successfully ran the build command for the Spin components.");
+    let mut path = src.to_str().unwrap().split('/').collect::<Vec<&str>>();
+    path.pop();
+    path.push("target");
+    path.push("wasm32-wasi");
+    path.push("release");
+    let path = path.join("/");
+
+    let modified = fs::metadata(Path::new(&path)).unwrap().modified();
+    let ten_secs = Duration::from_secs(10);
+
+    if modified.unwrap().elapsed().unwrap() >= ten_secs {
+        println!("nothing to build. Everything up to date");
+    } else {
+        println!("Successfully ran the build command for the Spin components.");
+    }
+
     Ok(())
 }
 
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -5,26 +5,16 @@
 use anyhow::{bail, Context, Result};
 use path_absolutize::Absolutize;
 use spin_loader::local::config::{RawAppManifest, RawComponentManifest};
-use std::fs::read_to_string;
 use std::path::{Path, PathBuf};
 use subprocess::{Exec, Redirection};
-use toml::Value;
 use tracing::log;
 
 /// If present, run the build command of each component.
 pub async fn build(app: RawAppManifest, src: &Path) -> Result<()> {
     let src = src.absolutize()?;
-    let mut path = src.to_str().unwrap().split('/').collect::<Vec<&str>>();
-    path.pop();
-    path.push("spin.toml");
-    let path = path.join("/");
 
-    let spin_manifest = read_to_string(&path).unwrap();
-    let manifest_info: Value = toml::from_str(&spin_manifest).unwrap();
-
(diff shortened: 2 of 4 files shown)


radu-matei (changes requested):
Hi, @jim4067, thanks for your contribution!

This approach _might_ make sense for Rust modules, but Spin components can be from any programming language that compiles to WASI — so we can't rely on the intended path for a Rust module.

I think the intended request for #700 was to print a message _IF_ there are no build commands set for any component in `spin.toml`.

Let me know if that makes sense.
Thanks!

jim4067:
Hello @radu-matei. Thank for the update. 

So what I was to do was check if 

```toml
[component.build]
``` 

exists in the toml file?

radu-matei:
Yes, but there could be N Spin components defined in `spin.toml`, so we should check if there is at least one perhaps?

jim4067:
I'm not sure I follow here. Could ypu add more info.

radu-matei:
A `spin.toml` file can have any number of components. For example, this is the `spin.toml` for the Spin documentation website:

```toml
name = "spin-docs"
spin_version = "1"
version = "0.1.0"
description = "The Spin documentation website running on... Spin."
authors = [ "Fermyon Engineering <engineering@fermyon.com>" ]
trigger = { type = "http", base = "/" }

[[component]]
source = "modules/bartholomew.wasm"
id = "bartholomew"
files = [ "content/**/*" , "templates/*", "scripts/*", "config/*"]
[component.trigger]
route = "/..."

[[component]]
source = "modules/spin_static_fs.wasm"
id = "fileserver"
files = [ { source = "static/", destination = "/" } ]
[component.trigger]
route = "/static/..."
```

When developing a Spin application, you could set the `build` command for zero, one, or more of the components of the application.

The intended behavior #700 was asking for is: if no component defined a `build` command, let the user know about it as opposed to saying "build successful".

jim4067:
I don't think I still follow.

Let's say you have many components, you only need one `build` command to say "build successful", otherwise if there is no build command for any of the components "alert the user that there is no build command set, right?

crates/build/src/lib.rs
```diff
@@ -5,13 +5,29 @@
...
 /// If present, run the build command of each component.
 pub async fn build(app: RawAppManifest, src: &Path) -> Result<()> {
     let src = src.absolutize()?;
+    let mut path = src.to_str().unwrap().split('/').collect::<Vec<&str>>();
+    path.pop();
+    path.push("spin.toml");
+    let path = path.join("/");
+
+    let spin_manifest = read_to_string(&path).unwrap();
+    let manifest_info: Value = toml::from_str(&spin_manifest).unwrap();
```
itowlson, line 23: This looks like it re-parses the TOML file.  You shouldn't need to do this because the `RawAppManifest` passed into this function contains this pre-parsed.

crates/build/src/lib.rs
```diff
@@ -5,13 +5,29 @@
...
+
+    //return early if build command not found
+    if !manifest_info.to_string().contains("component.build") {
```
itowlson, line 26: I'm a bit confused that this appears to parse the TOML then re-serialise it - is this to normalise table layouts so you can guarantee a literal string match perhaps?  But rather than working with TOML strings, you can use the `RawAppManifest` directly - something along the lines of

```
if app.components.iter().all(|c| c.build.is_none()) {
  // Nothing to do, print message and return
}
```

should do the trick I think.  (Though I could be missing something for sure!)

itowlson (changes requested):
Thanks for this!  I think you can simplify it by using the pre-parsed manifest though, and I'd recommend that approach as it cuts out a lot of the TOML management.

jim4067:
I see. thanks.

crates/build/src/lib.rs
jim4067: thanks.

crates/build/src/lib.rs
jim4067: I should have checked if you were already parsing the `.toml` files. sorry.

crates/build/Cargo.toml
```diff
@@ -12,3 +12,4 @@ tokio = { version = "1.11", features = [ "full" ] }
 spin-loader = { path = "../loader" }
 subprocess = "0.2.8"
 tracing = { version = "0.1", features = [ "log" ] }
+toml = "0.5.9"
```
itowlson, line 15: nit: this can probably be removed now?  (This would allow you to revert the Cargo.lock file as well.)

tests/http/simple-spin-rust/Cargo.lock
```diff
@@ -31,18 +31,22 @@ version = "1.1.0"
...
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
-
 [[package]]
 name = "fnv"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
+[[package]]
```
itowlson, line 40: This lockfile is prone to changing on rebuilds for some reason, but it would be great if you could revert it so it doesn't get this PR in its history - apologies for the inconvenience.

itowlson (approved):
Thanks, looks good!  There are a couple of changes that it would be good to back out, but nothing blocking.

itowlson:
Hi @jim4067, looks like we need the commits signed for the DCO please?  Thanks!

jim4067:
Hello @itowlson. I'm closing this PR and creating a new one with the changes requested.

PR #712, opened 2022-08-23
SIP: Spin Plugins

kate-goldenring:
SIP for supporting Spin Plugins

[rendered](https://github.com/kate-goldenring/spin/blob/spin-plugins-sip/docs/content/sips/006-spin-plugins.md)

Scoped down revision of @karthik2804's https://github.com/fermyon/spin/pull/682.

### DIFF
--- a/docs/content/sips/006-spin-plugins.md
+++ b/docs/content/sips/006-spin-plugins.md
@@ -1,0 +1,233 @@
+title = "SIP 006 - Spin Plugins"
+template = "main"
+date = "2022-08-23T14:53:30Z"
+---
+
+Summary: A Spin CLI command that will enable plugging in additional functionality and subcommands to Spin.
+
+Owners: karthik.ganeshram@fermyon.com and kate.goldenring@fermyon.com 
+
+Created: August 23, 2022
+
+## Background
+
+The realm of possibilities with Spin continues to grow. However, not every new feature is desired by every user. Instead of needing to modify the Spin codebase, contributors should be able to plug in new functionality or subcommands to Spin via the Spin CLI. This makes Spin easily extensible while keeping it lightweight.
+
+## Proposal
+
+Create a `spin plugin` command, which can be used to install a subcommand that can later be invoked via the Spin CLI. 
+
+For the initial proposal, all Spin plugins are expected to be packaged as an executable that will be executed by Spin when the plugin subcommand is invoked.
+
+A [`spin-plugins` repository](#centralized-plugin-manifest-repository) will act as an inventory of available plugins, made by both Spin maintainers and the community. In the repository, a plugin will be defined by a [JSON Spin plugin manifest](#spin-plugin-manifest). Spin will pull down this manifest during installation, which will instruct it on where to find the plugin binary, version, platform compatibility, and more.
+
+### Usage
+
+The `spin plugin` command will have three sub-commands.
+
+```bash
+Commands for working with Spin plugins
+
+USAGE:
+    spin plugin <SUBCOMMAND>
+
+SUBCOMMANDS:
+    install      Install plugin as described by a remote or local plugin manifest
+    uninstall    Uninstall a plugin
+	update		 Update one or all plugins to the latest or specified version 
+```
+
+**`spin plugin install`**
+
+The `spin plugin install` subcommand installs a plugin named `$name`. By default, it will look for a plugin manifest named `$name.json` in the `spin-plugin` repository; however, it can be directed to use a local manifest or one at a different remote location using the `--file` or `--url` flag, respectively.
+
+```bash
+Install a Spin plugin using a plugin manifest file. 
+By default, looks for the plugin manifest named <name>.json
+in the Spin plugins repository https://github.com/fermyon/spin-plugins
+
+USAGE:
+    spin plugin install <name>
+
+OPTIONS:
+    -f, --file                       Path to local plugin manifest
+	-u, --url                        Address of remote plugin manifest
+    -v, --version                    Desired version to update plugin to. Defaults to latest.
+    -s, --skip-version-compat-check  Instructs Spin to install the plugin even 
+                                     if its manifest specifies that it is 
+                                     incompatible with the current version of Spin
+```
+
+If the manifest is found, Spin will check that the plugin is compatible with the current OS, platform, and version of Spin. If so, before installing the plugin, Spin will prompt the user as to whether to trust the source. For example, the following prompt would be displayed for a plugin named `deploy` with an Apache 2 license and hosted at [`https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz`](https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz):
+
+```bash
+Installing plugin deploy with license Apache 2.0 from https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz
+For more information, reference the plugin metadata at `https://github.com/fermyon/spin-plugins/plugin-manifests/deploy.json`.
+Are you sure you want to proceed? ('yes'/'no') (default: no)
+```
+
+The plugin will only be installed if a user enters `yes`. Otherwise, the command exits. 
+
+Spin will reference the plugin manifest in order to fetch the plugin binary and install it into the user’s local data directory under a Spin-managed `plugins` subdirectory. The plugin manifest will be stored within a `manifests` subdirectory. 
+
+After installing a plugin, it can be executed directly from the Spin CLI. For example, a plugin named `$name` would be executed by running `spin $name <args>`. Any additional arguments supplied will be passed when executing the associated binary.
+
+ 
+
+**`spin plugin uninstall`**
+
+The `spin plugin uninstall` command uninstalls a plugin named `$name`. 
+
+```bash
+Uninstall a Spin plugin.
+
+USAGE:
+    spin plugin uninstall <name>
+```
+
+**`spin plugin update`**
+
+The `spin plugin update` command updates one or all plugins. If updating a single plugin, the desired version can be specified. By default, plugins are updated to the latest version in the plugins repository. As with `spin plugin install`, the local path or remote addresses to a plugin manifest can be specified.
+
+```bash
+Update one or all installed Spin plugins.
+
+USAGE:
+    spin plugin update [OPTIONS]
+
+OPTIONS:
+	-a, --all        Update all installed plugins (cannot be used with any other option)
+    -p, --plugin     Name of plugin
+    -v, --version    Desired version to update the plugin to. Defaults to latest. 
+    -f, --file       Path to local manifest (mutex with `-u`)
+    -u, --url        Address of remote manifest (mutex with `-f`)
+    -d, --downgrade  Enables downgrading a plugin to an older specified version.
+    -s, --skip-version-compat-check  Instructs Spin to install the plugin even 
+                                     if its manifest specifies that it is 
+                                     incompatible with the current version of Spin
+```
+
+The update will fail if the latest or user-specified version of the plugin is not [compatible with the current version of Spin](#plugin-compatibility). This failure can be overridden with `--skip-version-compat-check`.
+
+### Spin Plugin Manifest
+
+A Spin plugin is defined by a Spin Plugin Manifest which is a JSON file that conforms with the following [JSON Schema](https://json-schema.org/):
+
+```json
+{
+    "$schema": "https://json-schema.org/draft/2019-09/schema",
+    "$id": "https://github.com/fermyon/spin-plugins/json-schema/spin-plugin-manifest-schema-0.1.json",
+    "type": "object",
+    "title": "spin-plugin-manifest-schema-0.1",
+    "required": [
+        "name",
+        "description",
+        "version",
+        "spinCompatibility",
+        "license",
+        "packages"
+    ],
+    "properties": {
+        "name": {
+            "type": "string"
+        },
+        "description": {
+            "type": "string"
+        },
+        "homepage": {
+            "type": "string"
+        },
+        "version": {
+            "type": "string"
+        },
+        "spinCompatibility": {
+            "type": "string",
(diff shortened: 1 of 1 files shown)


docs/content/sips/006-spin-plugins.md
```diff
@@ -0,0 +1,233 @@
...
+
+If the manifest is found, Spin will check that the plugin is compatible with the current OS, platform, and version of Spin. If so, before installing the plugin, Spin will prompt the user as to whether to trust the source. For example, the following prompt would be displayed for a plugin named `deploy` with an Apache 2 license and hosted at [`https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz`](https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz):
+
+```bash
+Installing plugin deploy with license Apache 2.0 from https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz
+For more information, reference the plugin metadata at `https://github.com/fermyon/spin-plugins/plugin-manifests/deploy.json`.
+Are you sure you want to proceed? ('yes'/'no') (default: no)
+```
+
+The plugin will only be installed if a user enters `yes`. Otherwise, the command exits.
```
radu-matei, line 69: Thoughts on defaulting to `yes` vs. `no` here?

bacongobbler: I think this is the right approach, IMO. Safer to ask the user to explicitly check that they are indeed installing the correct third-party plugin. Much like how you have to say yes before connecting to a host via SSH. Make sure you trust the owner before connecting/installing something.

However I'd suggest accepting "y" and "n" as acceptable input.

docs/content/sips/006-spin-plugins.md
```diff
@@ -0,0 +1,233 @@
...
+                                     if its manifest specifies that it is 
+                                     incompatible with the current version of Spin
+```
+
+If the manifest is found, Spin will check that the plugin is compatible with the current OS, platform, and version of Spin. If so, before installing the plugin, Spin will prompt the user as to whether to trust the source. For example, the following prompt would be displayed for a plugin named `deploy` with an Apache 2 license and hosted at [`https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz`](https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz):
+
+```bash
+Installing plugin deploy with license Apache 2.0 from https://github.com/fermyon/spin-plugin-deploy/releases/download/v0.1.0/spin-plugin-deploy-v0.1.0-macos-aarch64.tar.gz
+For more information, reference the plugin metadata at `https://github.com/fermyon/spin-plugins/plugin-manifests/deploy.json`.
+Are you sure you want to proceed? ('yes'/'no') (default: no)
```
bacongobbler, line 66: A more compact form would be to capitalize the default value:

```suggestion
Are you sure you want to proceed? y/N
```

docs/content/sips/006-spin-plugins.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+The plugin will only be installed if a user enters `yes`. Otherwise, the command exits. 
+
+Spin will reference the plugin manifest in order to fetch the plugin binary and install it into the user’s local data directory under a Spin-managed `plugins` subdirectory. The plugin manifest will be stored within a `manifests` subdirectory. 
+
+After installing a plugin, it can be executed directly from the Spin CLI. For example, a plugin named `$name` would be executed by running `spin $name <args>`. Any additional arguments supplied will be passed when executing the associated binary.
```
bacongobbler, line 73: Can plugins override existing commands? For example, could I potentially write a custom `spin up` plugin that replaces the built-in version of `spin up`?

docs/content/sips/006-spin-plugins.md
```diff
@@ -0,0 +1,233 @@
...
+    -v, --version    Desired version to update the plugin to. Defaults to latest. 
+    -f, --file       Path to local manifest (mutex with `-u`)
+    -u, --url        Address of remote manifest (mutex with `-f`)
+    -d, --downgrade  Enables downgrading a plugin to an older specified version.
+    -s, --skip-version-compat-check  Instructs Spin to install the plugin even 
+                                     if its manifest specifies that it is 
+                                     incompatible with the current version of Spin
+```
+
+The update will fail if the latest or user-specified version of the plugin is not [compatible with the current version of Spin](#plugin-compatibility). This failure can be overridden with `--skip-version-compat-check`.
```
bacongobbler, line 110: When would skipping the version compatibility check be useful? In theory, if a plugin is incompatible with a particular version of spin, and you install it with `--skip-version-compat-check`, the plugin would be completely useless because it should not be able to work against that version of spin.

If the plugin is not compatible with that version of spin, what scenarios would require a user to skip the compatibility check?

docs/content/sips/006-spin-plugins.md
```diff
@@ -0,0 +1,233 @@
...
+- Manifests for older versions of the plugin can be retained in the Spin Plugins repository named `$name@$version.json` where `$version` is the value of the `version` field of the manifest. These specific versions can be installed using the `--version` flag.
+- The binary of the plugin must be named `$name`
+- The latest plugin manifest file must be named `$name.json`
+- The license for the plugin must be named `$name.license`
+
+#### Plugin Compatibility
+
+Spin plugins must specify compatible versions of Spin in the `spinCompatibility` field of the manifest. The field is expected to be a [comparison operator](https://docs.rs/semver/1.0.13/semver/enum.Op.html) (`=, >, >=, <, <=, ~, ^, *`) along with the compatible version of Spin. The JSON schema validates that the `spinCompatibility` field is a string that matches the following regular expression: `^[><~^*]?[=]?v?(0|[1-9]\d*)(\.(0|[1-9]\d*))?(\.(0|[1-9]\d*))?(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$`. 
+For example, specifying `=0.4` means that the plugin is compatible with versions equivalent to `>=0.4.0, <0.5.0`.
```
bacongobbler, line 210: Do we want to accept multiple rules to the field? For example setuptools allows the form `Django>=1.4.2,<1.8.99,!=1.5,!=1.6` which means

- install Django
- ALLOW `>=1.4.2, <= 1.8.99`
- EXCEPT `>=1.5.0, <1.7.0`

Which could be useful if it is incompatible with few specific versions of spin.

docs/content/sips/006-spin-plugins.md
```diff
@@ -0,0 +1,233 @@
...
+
+**`spin plugin update`**
+
+The `spin plugin update` command updates one or all plugins. If updating a single plugin, the desired version can be specified. By default, plugins are updated to the latest version in the plugins repository. As with `spin plugin install`, the local path or remote addresses to a plugin manifest can be specified.
+
+```bash
+Update one or all installed Spin plugins.
+
+USAGE:
+    spin plugin update [OPTIONS]
```
bacongobbler, line 96: Is it worth introducing a third command called `spin plugin upgrade` to handle this? Kinda like how you have to `apt update` your repositories before you `apt upgrade`. Wondering what might be the implications involved by compressing both steps into one command.

docs/content/sips/006-spin-plugins.md
```diff
@@ -0,0 +1,233 @@
...
+- Plugin manifests can be hosted elsewhere and installed via the `--file` or `--url` fields of `spin plugin install`.
+
+## Future design considerations
+
+### Larger scope for Spin Plugins
+
+The concept of Spin plugins is to allow both new subcommands and functionality to be added to Spin. This SIP focuses on the former, enabling users to both install and execute subcommands from the Spin CLI; however, there are cases where it may be useful to install a new Spin feature that is executed by Spin rather than the user. An example of this is Spin triggers. A user may wish to [extend Spin to support a timer trigger](https://spin.fermyon.dev/extending-and-embedding/) that executes components at a configured time interval. Instead of having to understand, modify, and grow the spin codebase, a user could package the trigger as a plugin. After installing the trigger via `spin plugin install`. Spin could invoke it when a Spin manifest references the trigger. 
+
+While for now plugins are assumed to be executables, in the future, support for plugging in WebAssembly modules may be desirable.
```
bacongobbler, line 227: Should this be a separate topic, i.e. `### WebAssembly Plugin Support`?

docs/content/sips/006-spin-plugins.md
karthik2804: In a scenario where Spin has been updated where the version of Spin does not satisfy the compatibility specified on the plugin manifest but does not break anything for the plugin. 

An example would be of a plugin whose manifest states compatibility as `=0.4` and a new spin version `0.5.0` is released. Now even if the spin version change does not break the plugin, the version compatibility check would render it broken. In such cases, if the plugin is provided by the community, the author must be very involved in updating the manifests. Having this as a flag would allow users to override the check and still install it.

bacongobbler: Wouldn't this be a non-issue if users made plugins use the syntax `>=1.0` instead of `=1.0`? IMO this seems more like "protecting" users from plugin developers not following the semantic versioning convention, which seems like a misstep.

IMO, as a plugin developer, If I explicitly state that my plugin is compatible with `=0.4`, then I expect it to NOT be usable against any version of spin other than spin `>=0.4<0.5`. Full stop.

kate-goldenring: These are good points. I am fine removing the flag, having it declares something as "okay" behavior when it isnt a best practice. The workaround for users would be:
```
wget github.com/fermyon/spin-plugins/plugin_name.json
# locally modify spinCompatibility
spin plugin install --file plugin_name.json
```

docs/content/sips/006-spin-plugins.md
kate-goldenring: Are you proposing breaking this down into a "check if updates exist" and "do update" step? We aren't caching available plugin information so i don't see the need for the former. What benefit to you see just the update side having?

bacongobbler: Yes.

Some form of caching/update lookup mechanism may be necessary if we are to use GitHub repos for indexing/searching. I know homebrew and other package managers got into trouble from causing tons of strain upstream due to commands like `brew search` cloning github.com/Homebrew/homebrew-core on every call...

If the plan is to implement a `spin plugin search` from the spin-plugins repo, it may be best to store a local cache to prevent unnecessary load upstream.

Might be putting the cart before the horse a bit, but, food for thought :)

bacongobbler: We could just cache the repo locally on `spin plugin update`, which calls to `spin plugin search` could use. But then users wouldn't be able to update their search results until they `spin plugin update` which may cause their plugins to update. That may or may not be intended behaviour, but I could see both arguments:

1. I just wanted to search for new plugins, not update them!
2. Oh hey! New updates came out and now I'm on the latest. Whee!

karthik2804: I think it would make sense to split it into `update` and `upgrade` if we are going to be caching the plugins repo to avoid unexpected upgrades for the user.

docs/content/sips/006-spin-plugins.md
kate-goldenring: I would say no. We can add a GitHub action to the plugins repository that prevents a naming conflict of that sort for plugins that live there and then have a check during install. Do you think a note on checking and denying installs of a plugin with the same name as an internal plugin should be added to the SIP?

karthik2804: I think even if installed it cannot be invoked as the priority of the subcommands will be higher for the internal subcommands over the external ones but in any case it might be good to make sure such installations cannot take place.

docs/content/sips/006-spin-plugins.md
kate-goldenring: These are all great points! And now i understand why cloning the entire repo is the optimal path of fetching plugins. Looks like we have 3 options (i may be missing some):
1. Add `spin plugin update` that fetches latest spin repo contents and rename current command to `spin plugin upgrade`
1. Just do latter of above, renaming `spin plugin update` to `spin plugin upgrade`. Always fetch latest before upgrades. Later, can add a separate `spin plugin update` that fetches the latest spin-repo contents. 
1. Add a `--fetch-latest` flag to `spin plugin update` (and later `spin plugin search`) that adds a preliminary step of pulling the latest spin-repo contents. This will avoid the scenario you gave of having to upgrade in order to search latest.

docs/content/sips/006-spin-plugins.md
itowlson, line 69: Very tactical detail, but there should be an option to bypass the prompt for CI scenarios.

kate-goldenring: Good point. Maybe a `-y` flag as in `apt`?

kate-goldenring: added in d75cfed287ce

docs/content/sips/006-spin-plugins.md
kate-goldenring: added in 24adf4a8de2a7627d

docs/content/sips/006-spin-plugins.md
kate-goldenring: removed in c775ecdb5617ac8b3e1

docs/content/sips/006-spin-plugins.md
kate-goldenring: added in 1cd903d28425e0

radu-matei (approved):
This looks great, @kate-goldenring and @karthik2804, thank you for the work here!

docs/content/sips/006-spin-plugins.md
bacongobbler: I like option 2 the best. Seems like the simplest option (just rename the command) while keeping the door open for future enhancements (`spin plugin update`).

kate-goldenring: Thanks for this recommendation @bacongobbler. Renamed in 76aa5da02f74b

docs/content/sips/006-spin-plugins.md
bacongobbler: Okay! Can we add a comment in here stating that plugins matching an existing command name will not be accepted into spin-plugins? Or did we want to go even further and add a check to `spin plugin install` itself?

kate-goldenring: I would think add a check to `spin plugin install` since plugins can exist outside of spin. I can add a note in the SIP in the install section.

kate-goldenring: Note added in dbea6bf9a6b. I also changed the install example from `deploy` to `test` to not make the example contradictory to the note.

bacongobbler (approved):

PR #719, opened 2022-08-29
Add HTTP TooManyRequests error

fibonacci1729:

### DIFF
--- a/sdk/rust/macro/wit/spin-http.wit
+++ b/sdk/rust/macro/wit/spin-http.wit
@@ -58,4 +58,5 @@ enum http-error {
     invalid-url,
     request-error,
     runtime-error,
+    too-many-requests,
 }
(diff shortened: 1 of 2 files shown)


sdk/rust/macro/wit/spin-http.wit
```diff
@@ -58,4 +58,5 @@ enum http-error {
     invalid-url,
     request-error,
     runtime-error,
+    too-many-requests,
 }
```
lann, line 62: Do you remember why we had to duplicate the `.wit`s over here?

wit/ephemeral/http-types.wit
```diff
@@ -55,4 +55,5 @@ enum http-error {
     invalid-url,
     request-error,
     runtime-error,
+    too-many-requests,
```
lann, line 58: Just a reminder that this _can be_ a breaking change. I don't see us doing explicit matching on these anywhere in our repos tho, so probably fine.

Aside: it would nice if wit-bindgen had a "non-exhaustive" annotation to emit e.g. Rust's `#[non_exhaustive]`

lann (approved):

Issue #663, opened 2022-07-26
Outbound HTTP component ignores `params` field


itowlson:
`http-types.wit` defines the `request` type to have a member `params` which is documented as `// The HTTP parameter queries, represented as a list of (name, value) pairs.`

If you set this on an outbound request, it is ignored - no query string is included the URL.

We should either respect `params`, or remove it from the outbound request type.

lann:
> remove it from the outbound request type

We should consider removing it entirely (moving it into SDKs); param parsing can be subtly different in different languages.

radu-matei:
One thing to consider if removing is that we are reusing the HTTP request and response type definitions across the HTTP trigger and the outbound HTTP implementation.

lann:
Yeah, I'm suggesting removing it from everywhere. Using url form encoding for params is "just" a (very common) convention.

rajatjindal:
this is also impacting tweet embedding in `bartholomew`. 

if i understand correctly, the consensus is to remove the params parameter from wit types file and make corresponding changes in sdk's?

lann:
If we're removing `params` this is _not_ a good first issue.

lann:
For outbound HTTP the `params` field is currently non-functional but it looks like the query part of the request URI should be working; at least, I don't see in the code that it is being removed.

itowlson:
Current behaviour from testing:

* Inbound
  * The original query string is included in `url`
  * The `params` field is populated
* Outbound
  * If you include a query string in `url` it is preserved verbatim (as Lann predicted)
  * If you set `params` it is ignored

lann:
> * Inbound
>   * The original query string is included in url

Well that's good. I propose we:

1. Switch our SDKs to parse the query
1. Document that `params` is deprecated
1. Drop `params` the next time we make breaking changes to the http interfaces (after streams support lands, if not sooner)

itowlson:
I will send a PR for 1 and 2.

itowlson:
Slight correction: if you include a query string on an outbound request, _Spin_ leaves it alone.  But if you're using the Rust SDK, _that_ strips it out (before passing the request to Spin).

itowlson:
After discussion, we believe we can parallel-run a new WIT with the existing one by using the Joel Trick TM at the ABI level and hiding it in the SDKs at the application level.  It will need some planning but we think it can be a non-breaking change.

rylev:
[...]

PR #723, opened 2022-08-30
Stop using WASI HTTP params in preparation for deprecation

itowlson:
Relates to #663.

What this PR does:

* HTTP inbound (trigger):
  * The `params` field is no longer populated when passed to the guest module. (It is set to an empty vector instead.) As far as I can tell, this should have no impact since the Rust and Go SDKs both appear to consume only the `uri`, and any query string APIs operate on that not on the `params` collection.
* HTTP outbound:
  * Spin will log a warning if an outbound request has any parameters set.
  * The Rust SDK no longer splits the `http::Request` URI into WASI `uri` and `params`.  The URI is stringised and preserved intact, and `params` is set to an empty vector.  This should mean the Rust SDK now behaves correctly when the request URI contains a query string but I am not sure where I can test this.
  * I don't believe any changes are required to the Go SDK, as this appears not to refer to `params` anywhere in the code - I assume it already passed through URIs with query strings unaltered.  Again not sure where to put a test for this.

@adamreese can you take a look and confirm that no changes are needed to the Go SDK please?  Thanks!

### DIFF
--- a/crates/http/src/spin.rs
+++ b/crates/http/src/spin.rs
@@ -4,7 +4,6 @@ use crate::{
 };
 use anyhow::Result;
 use async_trait::async_trait;
-use http::Uri;
 use hyper::{Body, Request, Response};
 use spin_engine::io::ModuleIoRedirects;
 use std::{net::SocketAddr, str, str::FromStr};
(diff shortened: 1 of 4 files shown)


karthik2804:
I have tested and confirmed that the Rust SDK now behaves correctly when there is a URI containing query strings. I tested by building Bartholomew & handlebars-sprig against the SDK to test the tweeting functionality.

crates/http/src/spin.rs
```diff
@@ -85,11 +84,7 @@ impl SpinHttpExecutor {
                 .map(|(k, v)| (k.as_str(), v.as_str()))
                 .collect();
 
-            let params = &Self::params(&parts.uri)?;
-            let params: Vec<(&str, &str)> = params
-                .iter()
-                .map(|(k, v)| (k.as_str(), v.as_str()))
-                .collect();
+            let params = vec![];
```
radu-matei, line 87: Would you mind adding a comment pointing to the issue, please?

radu-matei (approved):

lann (approved):

radu-matei (approved):

Issue #722, opened 2022-08-30
Panic if you specify `localhost:port` in `allowed_http_hosts`


itowlson:
If you write `allowed_http_hosts = ["localhost:3001"]` (forgetting the protocol), then make an outbound call, the Spin worker thread panics:

```
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', crates/outbound-http/src/lib.rs:52:47
```

This occurs because it tries to parse `localhost:3001` as `{ scheme: "localhost", host: None, port: None, path: "3001" }`.  This passes load-time validation as a syntactically legitimate (a la `urn:blah-blah-blah` scheme) though absurd URI, but crashes when we try to `unwrap` the host part.

itowlson:
Fixed via #727.

PR #727, opened 2022-08-31
Rework allowed HTTP host validation and checking

### LINKED ISSUE #720: `allowed_http_hosts` requires protocol but ignores it
If I write `allowed_http_hosts = ["https://localhost:3001"]` I can still make a request to `http://localhost:3001/` and it works.

If we are to require a protocol, shouldn't that be enforced?  I assumed this was why we required a protocol rather than just a hostname.

itowlson:
Fixes #720 and #722.

This now allows `allowed_http_hosts` to contain:

* Raw host name (`example.com`) - allows the default port for the HTTP or HTTPS schemes, whichever the guest uses
* Host plus port (`example.com:4444`) - allows that port (and the usage must be explicit - if the config is `example.com:80` then `http://example.com/foo` will be banned)
* Either of the above prefixed with `http://` or `https://` - the actual scheme is ignored - this is for back compat

This is a breaking change if:

* A manifest specifies a port, but the component uses a different port (or a default port)
* A manifest specifies a non-HTTP scheme
* A manifest specifies a port that is on the new deny list (common SMTP ports)

Validation failures are reported at `spin up` and look like this:

```
# spin.toml
# allowed_http_hosts = ["http://localhost:3001/foo/bar", "ftp://example.com", "google.com", "example.com:25", "localhost:3001"]

16:59 $ spin up
Error: One or more allowed_http_hosts entries was invalid:
http://localhost:3001/foo/bar contains a path, should be host and optional port only
ftp://example.com isn't a valid host or host:port string
Can't allow example.com:25 because the port is reserved for non-HTTP protocols
```

Happy to improve messages!

One ugly detail is that the permission logic has moved from the outbound HTTP crate into the manifest crate, because it goes on a type that represents the parsed and validated list (which has to be constructed at manifest time).  I can't think of a better way to organise this but would welcome ideas!

There is probably a bit more testing to be done on this but it's ready for review, just please don't merge until I've had the chance to whale on it a bit more.

### DIFF
--- a/crates/loader/src/bindle/mod.rs
+++ b/crates/loader/src/bindle/mod.rs
@@ -15,7 +15,7 @@ use crate::{
         config::{RawAppManifest, RawComponentManifest},
         utils::{find_manifest, parcels_in_group},
     },
-    validation::validate_allowed_http_hosts,
+    validation::{parse_allowed_http_hosts, validate_allowed_http_hosts},
 };
 use anyhow::{anyhow, Context, Result};
 use bindle::Invoice;
@@ -68,9 +68,10 @@ async fn prepare(
         toml::from_slice(&reader.get_parcel(&find_manifest(&invoice)?).await?)?;
     log::trace!("Recreated manifest from bindle: {:?}", raw);
 
+    validate_raw_app_manifest(&raw)?;
+
     let mut config_root = raw.config.take().unwrap_or_default();
     for component in &mut raw.components {
-        validate_allowed_http_hosts(&component.wasm.allowed_http_hosts)?;
         if let Some(config) = component.config.take() {
             let path = component.id.clone().try_into().with_context(|| {
                 format!("component ID {:?} not a valid config path", component.id)
@@ -104,6 +105,12 @@ async fn prepare(
         component_triggers,
         config_resolver,
     })
+}
+
+fn validate_raw_app_manifest(raw: &RawAppManifest) -> Result<()> {
+    raw.components
+        .iter()
+        .try_for_each(|c| validate_allowed_http_hosts(&c.wasm.allowed_http_hosts))
 }
 
 /// Given a raw component manifest, prepare its assets and return a fully formed core component.
@@ -140,7 +147,7 @@ async fn core(
         None => vec![],
     };
     let environment = raw.wasm.environment.unwrap_or_default();
-    let allowed_http_hosts = raw.wasm.allowed_http_hosts.unwrap_or_default();
+    let allowed_http_hosts = parse_allowed_http_hosts(&raw.wasm.allowed_http_hosts)?;
     let wasm = WasmConfig {
         environment,
         mounts,
--- a/crates/loader/src/validation.rs
+++ b/crates/loader/src/validation.rs
@@ -1,23 +1,246 @@
 #![deny(missing_docs)]
 
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Result};
 use reqwest::Url;
+use spin_manifest::{AllowedHttpHost, AllowedHttpHosts};
 
-// Check whether http host can be parsed by Url
+// Checks a list of allowed HTTP hosts is valid
 pub fn validate_allowed_http_hosts(http_hosts: &Option<Vec<String>>) -> Result<()> {
-    if let Some(domains) = http_hosts.as_deref() {
-        if domains
-            .iter()
-            .any(|domain| domain == wasi_outbound_http::ALLOW_ALL_HOSTS)
-        {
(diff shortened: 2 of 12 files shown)


itowlson:
Oh I need to update the docs too to show that we now expect host names not schemed URLs.

crates/loader/src/validation.rs
```diff
@@ -1,23 +1,276 @@
...
+                    ))
+                }
+            }
+            _ => Err(format!("{} isn't a valid host or host:port string", text)),
+        },
+        _ => Err(format!("{} isn't a valid host or host:port string", text)),
+    }
+}
+
+fn is_valid_host_name(text: &str) -> bool {
```
lann, line 78: Is this check necessary? It will reject IPv6 addrs as-is.

crates/loader/src/validation.rs
```diff
@@ -1,23 +1,276 @@
...
+    return !text.is_empty()
+        && text.len() <= 253
+        && text
+            .chars()
+            .all(|c| c.is_ascii_lowercase() || c.is_ascii_digit() || c == '-' || c == '.');
+}
+
+fn parse_allowed_http_host_as_schemed_url(text: &str) -> Result<AllowedHttpHost, String> {
```
lann, line 86: I think this would be easier to follow in straight procedural style, e.g.

```rust
let url = Url::parse(text)
    .map_err(|e| format!("{text} isn't a valid HTTP host URL: {e}"))?;

let host = url.host_str();
if host.is_none() {
    return Err(format!("{text} doesn't contain a host name"));
}
...
```

Perhaps just personal preference. :shrug:

crates/loader/src/validation.rs
```diff
@@ -1,23 +1,276 @@
...
-                Url::parse(d).with_context(|| format!("Can't parse {} in allowed_http_hosts", d))
-            })
-            .collect::<Result<Vec<_>, _>>()?;
-    }
-    Ok(())
+    }
+}
+
+fn is_allowed_port(port: u16) -> bool {
+    let forbidden_ports: &[u16] = &[25, 587, 2525];
```
lann, line 116: Top-level const? (or just drop this check per discussion w/ Mikkel)

crates/loader/src/validation.rs
```diff
@@ -1,23 +1,276 @@
...
+                        "One or more allowed_http_hosts entries was invalid:\n{}",
+                        errors.join("\n")
+                    ))
+                }
+            }
+        }
+    }
+}
+
+fn parse_allowed_http_host(text: &str) -> Result<AllowedHttpHost, String> {
```
lann, line 44: Would this make sense as `impl FromStr for AllowedHttpHost`?

crates/loader/src/validation.rs
itowlson: Hmm, and the way I'm checking for the presence of a port will also misfire on IPv6 addresses.

_mutters darkly_

crates/loader/src/validation.rs
itowlson: Yeah, I like Mikkel's suggestion.  I'll remove this and leave it to hosts to forbid these ports.

crates/loader/src/validation.rs
itowlson: It would, but that would mean putting the parse logic into the manifest crate, which is generally concerned with execution-ready forms rather than stringised formats in the TOML file.  The loader crate is, I think, the right place for this code, but it can't add new trait implementations.

lann: I guess I'll be updating this as part of the spin-core migration anyway, which should move this logic into `outbound-http`.

crates/loader/src/bindle/mod.rs
```diff
@@ -106,6 +107,15 @@ async fn prepare(
     })
 }
 
+fn validate_raw_app_manifest(raw: &RawAppManifest) -> Result<()> {
+    let _ = raw
+        .components
+        .iter()
+        .map(|c| validate_allowed_http_hosts(&c.wasm.allowed_http_hosts))
+        .collect::<Result<Vec<_>>>()?;
+    Ok(())
```
lann, line 116: Minor shortening; feel free to ignore:

```suggestion
    raw
        .components
        .iter()
        .try_for_each(|c| validate_allowed_http_hosts(&c.wasm.allowed_http_hosts))?
```

lann (approved):

crates/loader/src/bindle/mod.rs
itowlson: Ooh, TIL. Thanks!

PR #735, opened 2022-09-02
feat: Add support for Spin plugins

kate-goldenring:
## Adds support for Spin Plugins -- [SIP 006](https://spin.fermyon.dev/sips/006-spin-plugins.md).
Spin plugins are additional functionality and subcommands that can be added to Spin via this new `spin plugin` CLI command. 

A Spin plugin is defined by a JSON plugin manifest. A plugin manifest defines a plugin’s name, version, license, homepage (i.e. GitHub repo), compatible Spin version, and gives a short description of the plugin. It also points to the plugin source for various operating systems and platforms. This manifest is reference for plugin installs and upgrades.

The [`spin-plugins` repository](https://github.com/fermyon/spin-plugins) defines the plugin manifest [JSON schema](https://github.com/fermyon/spin-plugins/tree/main/json-schema) and contains an index of installable Spin and community maintained plugins.

**_These changes were equally contributed and co-authored by @karthik2804 -- thanks for all fabulous 🍐ing!_**

## Testing locally
Test the flow of installing an `example` plugin that echos `This is an example Spin plugin!` when run. It lives in the centralized Spin plugins repository [here](https://github.com/fermyon/spin-plugins/tree/main/manifests/example).

Clone this branch, build spin with `make build`, and install spin `cargo install --locked --path .` or replace `spin` below with the path to the binary `./target/release/spin`.

**Installing a plugin from the `spin-plugins` repo**
```
# install latest version of example plugin (to ignore prompting add --yes)
spin plugin install example 
# execute plugin
spin example
# uninstall plugin
spin uninstall example
```

**Upgrading a plugin**
```
spin plugin install example  --yes --version 0.1.0
spin upgrade example
# now could downgrade with: spin plugin upgrade example -d --version 0.1.0
```

**Installing plugin from remote manifest**
```
spin plugin install --url https://github.com/fermyon/spin-plugins/blob/main/manifests/example/example.json
```

**Installing plugin from local manifest**
See `test_spin_plugin_install_command` in `integration.rs`.


src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+}
+
+/// Install plugins from remote source
+#[derive(Parser, Debug)]
+pub struct Install {
+    /// Name of Spin plugin.
+    #[clap(
+        name = "PLUGIN_NAME",
+        conflicts_with = "REMOTE_PLUGIN_MANIFEST",
```
itowlson, line 47: Use consts for argument names; that way, if a name changes, it doesn't have to be kept in sync across multiple places.

src/commands/external.rs
```diff
@@ -0,0 +1,56 @@
...
+            "SPIN_BIN_PATH".to_string(),
+            env::current_exe()?
+                .to_str()
+                .ok_or_else(|| anyhow!("Could not convert binary path to string"))?
+                .to_string(),
+        ),
+        (
+            "SPIN_PLUGIN_PATH".to_string(),
+            path.to_str()
+                .ok_or_else(|| anyhow!("Could not convert plugin path to string"))?
```
itowlson, line 49: Can you do `path.display().to_string()` and avoid the `Result`?  Or should we use the underlying OS string (`as_os_str`) for this and the Spin binary path and pass OsStrings to `envs`?

src/commands/external.rs
```diff
@@ -0,0 +1,56 @@
...
+use tokio::process::Command;
+use tracing::log;
+
+/// Executes a Spin plugin as a subprocess, expecting the first argument to
+/// indicate the plugin to execute. Passes all subsequent arguments on to the
+/// subprocess.
+pub async fn execute_external_subcommand(args: Vec<String>) -> Result<()> {
+    let plugin_name = args.first().ok_or_else(|| anyhow!("Expected subcommand"))?;
+    let plugins_dir = get_spin_plugins_directory()?;
+    check_plugin_spin_compatibility(plugin_name, env!("VERGEN_BUILD_SEMVER"), &plugins_dir)?;
```
itowlson, line 15: It's a bit non-obvious to me where the existence check happens - it clearly does because we get a nice message, but it's not obvious in the code flow.

src/commands/external.rs
```diff
@@ -0,0 +1,56 @@
...
+use anyhow::{anyhow, Result};
+use spin_plugins::version_check::check_plugin_spin_compatibility;
+use std::{collections::HashMap, env, path::Path};
+use tokio::process::Command;
+use tracing::log;
+
+/// Executes a Spin plugin as a subprocess, expecting the first argument to
+/// indicate the plugin to execute. Passes all subsequent arguments on to the
+/// subprocess.
+pub async fn execute_external_subcommand(args: Vec<String>) -> Result<()> {
```
itowlson, line 12: If I mistype a command I now get a message about plugins:

```
$ ./target/debug/spin biuld
Error: The plugin does not exist
```

Could we say something like "unknown command" and print the help, which is what I think we used to do?

src/bin/spin.rs
```diff
@@ -44,8 +44,12 @@ enum SpinApp {
     Bindle(BindleCommands),
     Deploy(DeployCommand),
     Build(BuildCommand),
+    #[clap(subcommand)]
+    Plugin(PluginCommands),
     #[clap(subcommand, hide = true)]
     Trigger(TriggerCommands),
+    #[clap(external_subcommand)]
```
itowlson, line 51: Is there a way to surface these to the help system?  When I run `cargo`, it lists all known commands including those from plugins - is it possible for us to do the same?  (Probably future enhancement, don't think it was discussed in the SIP.)

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+        conflicts_with = "LOCAL_PLUGIN_MANIFEST",
+        requires("PLUGIN_NAME")
+    )]
+    pub version: Option<Version>,
+}
+
+impl Install {
+    pub async fn run(self) -> Result<()> {
+        let manifest_location = match (self.local_manifest_src, self.remote_manifest_src, self.name) {
```
itowlson, line 91: Could move this lengthy match out to a `self.manifest_location()` function?

itowlson: Using a `ManifestLocation` enum is really nice.  Much cleaner than carting all those options around.

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+            _ => return Err(anyhow::anyhow!("Must provide plugin name for plugin look up xor remote xor local path to plugin manifest")),
+        };
+        PluginInstaller::new(
+            manifest_location,
+            get_spin_plugins_directory()?,
+            self.yes_to_all,
+            env!("VERGEN_BUILD_SEMVER"),
+        )
+        .install()
+        .await?;
```
itowlson, line 104: Again, I'm a little surprised at the indentation here.  I thought a vertical chain of methods was usually indented from the code around it.

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+        let manifest_dir = plugins_dir.join(PLUGIN_MANIFESTS_DIRECTORY_NAME);
+
+        // Check if no plugins are currently installed
+        if !manifest_dir.exists() {
+            println!("No currently installed plugins to update.");
+            return Ok(());
+        }
+
+        if self.all {
+            // Install the latest of all currently installed plugins
```
itowlson, line 200: Strongly suggest moving this block out to a `upgrade_all()` function.  The `else` is so far away, and there is so much indentation, it's hard for the reader to match it with the `if`.

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+            for plugin in std::fs::read_dir(manifest_dir)? {
+                let path = plugin?.path();
```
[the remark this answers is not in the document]
itowlson: This probably applies elsewhere too... Rust file system error messages in particular are absolutely awful for not telling you which file the error is on...

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+                    spin_version,
+                )
+                .install()
+                .await
+                {
+                    // Ignore plugins that were not installed from the central
+                    // plugins repository
+                    if e.to_string().contains("Could not find plugin") {
+                        log::info!(
+                            "Could not update {} plugin as DNE in central repository",
```
itowlson, line 226: DNE?

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+            // Install the latest of all currently installed plugins
+            for plugin in std::fs::read_dir(manifest_dir)? {
+                let path = plugin?.path();
+                let name = path
+                    .file_stem()
+                    .ok_or_else(|| anyhow!("expected directory for plugin"))?
+                    .to_str()
+                    .ok_or_else(|| anyhow!("Cannot convert directory to String"))?
+                    .to_string();
+                if let Err(e) = PluginInstaller::new(
```
itowlson, line 209: I found this hard to read as the operation being checked (`.install()`) was a long way from the check.  Can we introduce a variable for the installer object?

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+                    plugins_dir.clone(),
+                    self.yes_to_all,
+                    spin_version,
+                )
+                .install()
+                .await
+                {
+                    // Ignore plugins that were not installed from the central
+                    // plugins repository
+                    if e.to_string().contains("Could not find plugin") {
```
itowlson, line 224: How do you know that this string represents "it wasn't installed from the central registry"?  Is there a safer way to manage this?  (The answer may well be no.)

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+                        log::info!(
+                            "Could not update {} plugin as DNE in central repository",
+                            name
+                        );
+                    } else {
+                        return Err(e);
```
itowlson, line 230: Correct to bail on error rather than upgrading remaining plugins?  This could mean one bad plugin blocking others from upgrading.

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+                    }
+                }
+            }
+        } else {
+            let name = self
+                .name
+                .ok_or_else(|| anyhow!("plugin name is required for upgrades"))?;
+            // If downgrade is allowed, first uninstall the plugin
```
itowlson, line 238: Why is this needed for downgrades if it isn't for upgrades?

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+            let manifest_location = match (self.local_manifest_src, self.remote_manifest_src) {
+                (Some(path), None) => ManifestLocation::Local(path),
+                (None, Some(url)) => ManifestLocation::Remote(url),
+                _ => ManifestLocation::PluginsRepository(PluginInfo::new(
+                    &name,
+                    Url::parse(SPIN_PLUGINS_REPO)?,
+                    self.version,
+                )),
+            };
+            PluginInstaller::new(
```
itowlson, line 251: This is very subjective, but my eyes keep speed-bumping on the `Foo::new(several-lines-of-stuff).operation()` pattern.  I want `new` to return a value and keep being surprised that it doesn't, particularly when the "oh it's just to call the method" realisation is so far away.

If we really aren't interested in these values as stateful, how about having a `PluginInstaller::install(...)` function for when you have no interest in keeping the installer object around?  (Just wrapping the `new/install` combo.)

itowlson (changes requested):
A few thoughts on the command handlers - I've not had chance to look at the actual crate yet but will do so tomorrow.  Most of what I have is stylistic issues rather than functionality, and some of it is subjective I know.  Anyway, this is fantastic to see - thanks!

crates/plugins/src/plugin_manifest.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Checksum to verify the plugin before installation.
+    pub sha256: String,
+}
+
+/// Describes the compatible OS of a plugin
+#[derive(Serialize, Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) enum Os {
+    Linux,
+    Osx,
```
itowlson, line 51: Is this called MacOS now?  Makes no odds really though!

crates/plugins/src/plugin_manifest.rs
```diff
@@ -0,0 +1,168 @@
...
+            "description": "Some description.",
+            "homepage": "www.example.com",
+            "version": "1.0",
+            "spinCompatibility": "=0.4",
+            "license": "Mit",
+            "packages": [
+                {
+                    "os": "linux",
+                    "arch": "amd64",
+                    "url": "www.example.com/releases/1.0/binary.tgz",
```
itowlson, line 97: Worth having different URLs and SHAs so you can have a test that the selector picks the right one for a given OS/arch?

crates/plugins/src/prompt.rs
```diff
@@ -0,0 +1,63 @@
...
+/// want to proceed with installation.
+pub(crate) struct Prompter {
+    plugin_name: String,
+    plugin_license: String,
+    source_url: String,
+}
+
+impl Prompter {
+    /// Creates a new prompter
+    pub fn new(plugin_name: &str, plugin_license: &str, source_url: &str) -> Result<Self> {
```
itowlson, line 14: As ever, this is subjective, but this design feels a bit odd to me.  A Prompter feels like it would encapsulate a _strategy_ for getting input from the user (e.g. console prompter), and that things like the names would be passed in each time it was called.  With the current design the Prompter object has no useful public role because the only thing you can do is call `run` on it, and `run` returns the result directly - there's no point keeping the Prompter itself around.  Why not have a `prompt(...)` function and avoid the extra characters and concepts of `Prompter:new(...).run()`?

(This isn't me saying "don't do it this way," but me asking for more clarity on _why_ do it this way.)

crates/plugins/src/prompt.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    fn are_you_sure(&self) -> Result<bool> {
+        let mut resp = String::new();
+        io::stdin().read_line(&mut resp)?;
+        Ok(self.parse_response(&mut resp))
+    }
+
+    fn parse_response(&self, resp: &mut str) -> bool {
+        let resp = resp.trim().to_lowercase();
+        resp.eq("yes") || resp.eq("y")
```
itowlson, line 38: This will accept any non-Y input as "no", e.g. if I aim for the 'y` key but accidentally hit the key next to it, that's treated as "no".  Is that usual?  Or should we re-prompt if we don't get a recognised input or the "accept default" key?

itowlson, line 38: Why use `eq` instead of `==`?  (not saying it's wrong, more hoping to learn something!)

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
...
+            .find(|p| p.os == os && p.arch.to_string() == arch)
+            .ok_or_else(|| anyhow!("This plugin does not support this OS or architecture"))?;
+        let target = plugin_package.url.to_owned();
+
+        // Ask for user confirmation to install if not overridden with CLI
+        // option
+        if !self.yes_to_all
+            && !Prompter::new(&plugin_manifest.name(), &plugin_manifest.license, &target)?.run()?
+        {
+            // User has requested to not install package, returning early
```
itowlson, line 197: If the user says no, and this was a downgrade, will this result in the user's existing plugin being deleted rather being a no-op?

crates/plugins/src/uninstall.rs
```diff
@@ -0,0 +1,46 @@
...
+        let manifest_file = self
+            .plugins_dir
+            .join(PLUGIN_MANIFESTS_DIRECTORY_NAME)
+            .join(get_manifest_file_name(&self.plugin_name));
+        let plugin_exists = manifest_file.exists();
+        match plugin_exists {
+            // Remove the manifest and the plugin installation directory
+            true => {
+                fs::remove_file(manifest_file)?;
+                fs::remove_dir_all(self.plugins_dir.join(&self.plugin_name))?;
```
itowlson, line 34: I think I'm clicking what the layout is, but the information seems decentralised.  Is it worth having a `PluginsStore` object that could provide methods like `manifest_file_path(name)` and `plugin_directory(name)` so we're not having to make sure to join to the right parent every time we construct a path?

crates/plugins/src/uninstall.rs
```diff
@@ -0,0 +1,46 @@
...
+                println!("Uninstalled plugin {} successfully", &self.plugin_name);
+            }
+            false => {
+                println!(
+                    "The following plugin  \"{}\" does not exist, therefore cannot be uninstalled",
```
itowlson, line 39: Probably don't need "following" e.g. 

```suggestion
                    "Plugin \"{}\" isn't present, so no changes were made",
```

By the way, this message isn't using the square brackets convention used in the installer - which is preferred?

crates/plugins/src/version_check.rs
```diff
@@ -0,0 +1,67 @@
+use crate::{
+    get_manifest_file_name, plugin_manifest::PluginManifest, PLUGIN_MANIFESTS_DIRECTORY_NAME,
+};
+use anyhow::{anyhow, Result};
+use semver::{Version, VersionReq};
+use std::{fs::File, path::Path};
+
+/// Checks whether the plugin supports the currently running version of Spin.
+pub fn assert_supported_version(spin_version: &str, supported: &str) -> Result<()> {
```
itowlson, line 9: "assert" is a very "should never happen" kind of name *grin*

crates/plugins/src/version_check.rs
```diff
@@ -0,0 +1,67 @@
...
+    let supported = VersionReq::parse(supported).map_err(|e| {
+        anyhow!(
+            "could not parse manifest compatibility version {} as valid semver -- {:?}",
+            supported,
+            e
+        )
+    })?;
```
itowlson, line 16: Is there a way for the user to go "okay there's a typo in the version spec but I want it anyway"?

crates/plugins/src/version_check.rs
```diff
@@ -0,0 +1,67 @@
...
+    let version = Version::parse(spin_version)?;
+    match supported.matches(&version) {
+        true => Ok(()),
+        false => Err(anyhow!(
+            "plugin is compatible with Spin {} but running Spin {}",
+            supported,
+            spin_version
+        )),
```
itowlson, line 24: Again, is there or should there be a way for the user to go YOLO SHIP IT ANYWAY?

Generally, should we distinguish between "are you really sure you want to do this" warnings (the author hasn't updated their version spec yet) vs "we can't do this at all" (the upstream site is gone) errors?

crates/plugins/src/version_check.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+pub(crate) fn get_plugin_manifest(plugin_name: &str, plugins_dir: &Path) -> Result<PluginManifest> {
```
itowlson, line 28: Does this belong in a `version_check` module?

crates/plugins/src/version_check.rs
```diff
@@ -0,0 +1,67 @@
...
+    let manifest_path = plugins_dir
+        .join(PLUGIN_MANIFESTS_DIRECTORY_NAME)
+        .join(get_manifest_file_name(plugin_name));
+    log::info!("Reading plugin manifest from {:?}", manifest_path);
```
itowlson, line 32: consider `manifest_path.display()` and not using the debug format

itowlson (changes requested):
Super happy to see this - I have a lot of nits, but they are largely just nits.

The main things that I _would_ like to see are:

* Name more things - create more functions and variables
* Define some sort of Store or Layout type to manage all the paths - there's a lot of duplication and at some point someone is going to join a path to the wrong parent
* Decouple the installer from the specifics of the CLI environment
* Ensure errors give good actionable data e.g. file names (or that users can get this info somehow)
* Ensure that changing your mind (saying no at  the confirmation prompt) is a no-op (I think it might not be in a downgrade scenario)
* Distinguish errors (we can't or won't do this) from warnings where the user may want to go ahead anyway even knowing the risks

That said, even these could be iterated on if we want to get this in quickly - the functionality looks great.

crates/plugins/src/plugin_manifest.rs
```diff
@@ -0,0 +1,168 @@
+use serde::{Deserialize, Serialize};
+
+/// Expected schema of a plugin manifest. Should match the latest Spin plugin
+/// manifest JSON schema:
```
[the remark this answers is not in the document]
kate-goldenring: We went with JSON because of the large ecosystem around JSON schema documentation and JSON parsing for schema matching. However, I did not realize that TOML also can be [parsed with JSON schema](https://json-schema-everywhere.github.io/toml); though, I cannot see examples. Have you seen examples of TOML schemas?

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
+use crate::{
+    get_manifest_file_name, get_manifest_file_name_version,
+    version_check::{assert_supported_version, get_plugin_manifest},
+    PLUGIN_MANIFESTS_DIRECTORY_NAME, SPIN_INTERNAL_COMMANDS,
+};
+
+use super::git::GitSource;
```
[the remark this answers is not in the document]
kate-goldenring: Sadly not a valid one. For some reason, I associated crate importing with public modules and super with ones that are private to the crate. However, i think the standard is to use crate when you are specifying specific paths (like we are for all these imports in `install.rs`) and use `super` when there are nested modules in a file (such as adding a tests module to the bottom of a file). I'll switch all to `crate` thanks!

crates/plugins/src/plugin_manifest.rs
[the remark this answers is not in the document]
itowlson: The "Even Better TOML" VS Code extension does some stuff with TOML schemas.  But yeah, it's not as widely used.  It's a definite advantage of JSON.

Should we migrate to JSON for our other manifests to take advantage of schema support, or does the plugin manifest have specific considerations (e.g. higher complexity, less human authoring) that make schema support more important (or human readability less important) than in other manifests?

src/commands/external.rs
```diff
@@ -0,0 +1,56 @@
...
+        binary.set_extension("exe");
+    }
+    let mut command = Command::new(binary);
+    if args.len() > 1 {
+        command.args(&args[1..]);
+    }
+    command.envs(&get_env_vars_map(&path)?);
+    log::info!("Executing command {:?}", command);
+    // Allow user to interact with stdio/stdout of child process
+    let status = command.status().await?;
```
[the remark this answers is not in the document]
karthik2804: The exit code is logged.

src/commands/external.rs
karthik2804: Yes, will fix this.

src/commands/external.rs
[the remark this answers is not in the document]
itowlson: Yeah, I was thinking if a script wanted to use it.  In particular, a script will want to know if the plugin succeeded or failed; at the very least, we should return a nonzero exit code if the plugin does, which I don't think happens at the moment?  (do correct me if I'm missing it!)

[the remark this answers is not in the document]
karthik2804: Do you mean in the case where someone has a script that executes `spin <plugin>` and wants to check on its status to verify that this step succeeded before moving on to the next? If so, you are correct. We are currently not doing this but I agree this must be added. Thanks for pointing it out!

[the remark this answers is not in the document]
itowlson: Exactly, or e.g. imagine a `spin test` plugin being run as part of a CI script.

src/bin/spin.rs
karthik2804: I think cargo lists it when using the argument `--list`. It is a command that can be potentially added later or if we want the default `help` to display all the known commands including plugins, it can be accomplished by modifying the default help text.

crates/plugins/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+pub(crate) const SPIN_INTERNAL_COMMANDS: [&str; 9] = [
+    "templates",
+    "up",
+    "new",
+    "bindle",
+    "deploy",
+    "build",
+    "plugin",
+    "trigger",
+    "external",
```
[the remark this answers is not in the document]
karthik2804: if both a `test` internal command and plugin exist, it will be parsed as an internal subcommand and therefore the plugin cannot be invoked until the binary name is changed. Therefore, plugins cannot squat on internal subcommand names.

crates/plugins/src/prompt.rs
karthik2804: I think most package managers default to no. At least `pacman` and `apt` share this behaviour.

crates/plugins/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+mod git;
+pub mod install;
+mod plugin_manifest;
+mod prompt;
+pub mod uninstall;
+pub mod version_check;
+
+/// List of Spin internal subcommands
+pub(crate) const SPIN_INTERNAL_COMMANDS: [&str; 9] = [
```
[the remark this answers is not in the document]
karthik2804: Thanks for pointing it out! Yes, we can implement tests using [get_subcommands](https://docs.rs/clap/latest/clap/builder/struct.App.html#method.get_subcommands) method provided by `clap`. We could even potentially replace the generation of the list of commands to be dynamic such that it does not need to be updated manually.

crates/plugins/src/lib.rs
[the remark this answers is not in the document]
itowlson: Thanks, that's good to understand.  Unfortunately, I didn't explain my concern clearly (and then muddied the waters with that second thought).  What I was thinking of when I mentioned squatting was something like:

* Alice creates a `spin test` plugin
* Bob likes it and uses it in his CI system
* Now we can't create a built-in `spin test` command because it would break Bob's CI, which would be rude

So if there are commands we're tossing up having, we might want to reserve the names.  Hope that makes more sense now!

Also, might be worth having a way to force mapping to a plugin, e.g. `spin external test` or `spin plugin run test`, so that at least Bob can fix his CI without having to badger Alice to rename her plugin.  That can be retrofitted if the plugin ecosystem merits it though!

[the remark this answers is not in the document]
karthik2804: That makes sense to me. I would assume that if we are releasing a new internal subcommand, the version of spin would change thus not allowing the external plugin to run in the first place without an update to the plugin. This would depend on the version compatibility specified by the plugin.

Although, I like the solution of an `external` subcommand to force run plugins in such cases.

src/bin/spin.rs
itowlson: I stand corrected!  I saw `add` in the `cargo` output, but forgot that `add` no longer came from a plugin!  My apologies.  And agree later in a distinct command is fine.

crates/plugins/src/lib.rs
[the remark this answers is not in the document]
karthik2804: My only concern with that would be that we want to avoid plugins having the same name as the internal subcommands. Adding this option would make that test pointless. It might lead to confusion among users due to commands of the same name doing different things.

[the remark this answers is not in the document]
itowlson: We do!  And it might!  And the test is good because it prevents plugins having the same names as _known_ commands.  The trouble is, we can't stop a plugin having the same name as a command we haven't thought of yet.  So: when we want to use a command name that is already used by an existing plugin, what happens?

As I say... we can probably punt on this for now.  We may need to confront it if the plugin ecosystem takes off, and it would be good to have the shape of a solution ready.  But we don't need to implement it yet.

crates/plugins/src/prompt.rs
kate-goldenring: I must have gotten a clippy error in the past that made me add this instinctually. this did set me on a learning path. `.eq()` will ensure that the types being compared are both references if that is not the case. Seems preferable to stick with `==`.

src/commands/plugins.rs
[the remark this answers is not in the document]
kate-goldenring: Good point. Added in 8da5853c0e844

src/commands/external.rs
```diff
@@ -0,0 +1,56 @@
...
+pub async fn execute_external_subcommand(args: Vec<String>) -> Result<()> {
+    let plugin_name = args.first().ok_or_else(|| anyhow!("Expected subcommand"))?;
+    let plugins_dir = get_spin_plugins_directory()?;
+    check_plugin_spin_compatibility(plugin_name, env!("VERGEN_BUILD_SEMVER"), &plugins_dir)?;
+    let path = plugins_dir.join(plugin_name);
+    let mut binary = path.join(plugin_name);
+    if cfg!(target_os = "windows") {
+        binary.set_extension("exe");
+    }
+    let mut command = Command::new(binary);
```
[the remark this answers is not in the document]
kate-goldenring: hid it behind a `installed_binary_path` function

src/commands/external.rs
```diff
@@ -0,0 +1,56 @@
...
+                .to_string(),
+        ),
+        (
+            "SPIN_PLUGIN_PATH".to_string(),
+            path.to_str()
+                .ok_or_else(|| anyhow!("Could not convert plugin path to string"))?
+                .to_string(),
+        ),
+    ]
+    .into_iter()
```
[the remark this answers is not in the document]
kate-goldenring: I believe I've been using `cargo fmt` before all commits :/ maybe i forgot this past time

src/commands/external.rs
kate-goldenring: broke it out into steps to make it clearer

src/commands/external.rs
kate-goldenring: I am having a hard time replicating the current experience where it will also tell you if you may have mistyped:
```
error: The subcommand 'biuld' wasn't recognized

        Did you mean 'build'?

If you believe you received this message in error, try re-running with 'spin -- biuld'

USAGE:
    spin <SUBCOMMAND>

```
I wonder if we do want to do `spin plugin run $plugin_name`? This would also resolve the internal plugin name conflict. It might be worth doing for the time being as i am not sure how to dynamically configure clap with the plugin subcommands.

src/commands/plugins.rs
```diff
@@ -0,0 +1,275 @@
...
+            return Ok(());
+        }
+
+        if self.all {
+            // Install the latest of all currently installed plugins
+            for plugin in std::fs::read_dir(manifest_dir)? {
+                let path = plugin?.path();
+                let name = path
+                    .file_stem()
+                    .ok_or_else(|| anyhow!("expected directory for plugin"))?
```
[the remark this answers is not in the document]
kate-goldenring: tried to clear it up

src/commands/plugins.rs
kate-goldenring: made it so upgrading continues. Good call out!

src/commands/plugins.rs
kate-goldenring: got rid of the `PluginInstaller` in favor of a `PluginManager` with `install` and `uninstall` functions. Thank you for pointing this out!

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    /// Installs a Spin plugin. First attempts to retrieve the plugin manifest.
+    /// If installing a plugin from the centralized Spin plugins repository, it
+    /// fetches the latest contents of the repository and searches for the
+    /// appropriately named and versioned plugin manifest. Parses the plugin
+    /// manifest to get the appropriate source for the machine OS and
+    /// architecture. Verifies the checksum of the source, unpacks and installs
+    /// it into the plugins directory.
+    pub async fn install(&self) -> Result<()> {
```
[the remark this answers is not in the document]
kate-goldenring: I am now calling the prompt mechanisms from spin CLI. Are you also suggesting that it is best to move that functionality out of the plugins crate? I can move the `prompt` function to `external.rs.

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
...
+    ) -> Self {
+        Self {
+            manifest_location,
+            plugins_dir,
+            yes_to_all,
+            spin_version: spin_version.to_string(),
+        }
+    }
+
+    /// Installs a Spin plugin. First attempts to retrieve the plugin manifest.
```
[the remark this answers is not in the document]
kate-goldenring: I made it chunky :)

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// If installing a plugin from the centralized Spin plugins repository, it
+    /// fetches the latest contents of the repository and searches for the
+    /// appropriately named and versioned plugin manifest. Parses the plugin
+    /// manifest to get the appropriate source for the machine OS and
+    /// architecture. Verifies the checksum of the source, unpacks and installs
+    /// it into the plugins directory.
+    pub async fn install(&self) -> Result<()> {
+        let plugin_manifest: PluginManifest = match &self.manifest_location {
```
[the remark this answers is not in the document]
kate-goldenring: I agree that the remote repo pulling could be separated out. this motivated me creating a `PluginLookup` type that focuses on the repo fetching and lookups

src/commands/plugins.rs
itowlson: Our defaulting to `anyhow::Error` bites us once again... one possibility would be for `install()` to return an error type a la `enum InstallError { ConnectionFailed, NotFound, InvalidManifest, ... }` (made conveniently stringable with `thiserror`) and then you could check whether it was a `NotFound` error directly?

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
...
+        {
+            bail!(
+                "Trying to install a plugin with the same name '{}' as an internal plugin",
+                plugin_manifest.name()
+            );
+        }
+
+        // Disallow downgrades and reinstalling identical plugins
+        if let Ok(installed) = get_plugin_manifest(&plugin_manifest.name(), &self.plugins_dir) {
+            if installed.version > plugin_manifest.version || installed == plugin_manifest {
```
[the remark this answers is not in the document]
kate-goldenring: Good point. Made it so downgrade optionality is passed to the `install` function and evaluated here

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
...
+        // Ask for user confirmation to install if not overridden with CLI
+        // option
+        if !self.yes_to_all
+            && !Prompter::new(&plugin_manifest.name(), &plugin_manifest.license, &target)?.run()?
+        {
+            // User has requested to not install package, returning early
+            println!("Plugin {} will not be installed", plugin_manifest.name());
+            return Ok(());
+        }
+        let target_url = Url::parse(&target)?;
```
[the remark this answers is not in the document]
kate-goldenring: I could move this into `check_manifest` but there will be duplicate parsing of the url due to the target_url not being retained. Is this still ideal?

crates/plugins/src/install.rs
```diff
@@ -0,0 +1,288 @@
...
+            }
+        };
+        self.verify_checksum(&plugin_tarball_path, &plugin_package.sha256)?;
+
+        self.untar_plugin(&plugin_tarball_path, &plugin_manifest.name())?;
+
+        // Save manifest to installed plugins directory
+        self.add_to_manifest_dir(&plugin_manifest)?;
+        println!(
+            "Plugin [{}] was installed successfully!",
```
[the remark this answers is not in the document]
kate-goldenring: I am trying to distinguish it but yes we are inconsistent. I removed brackets

src/commands/external.rs
itowlson: I don't know if we need to full-on replicate the current experience, but if we could at least say "unknown command" or "the command '..' wasn't recognised" (and "for more information..."?) that would suffice.  Just trying to avoid sending them in the wrong direction.

crates/plugins/src/version_check.rs
kate-goldenring: changed it to a nice kind `check_`

src/commands/plugins.rs
itowlson: Fair enough, sorry for the red herring.  Just another reason for me to rant about `cargo fmt`.

crates/plugins/src/plugin_manifest.rs
kate-goldenring: looks like macos is the latest and greatest. updated here and in `spin-plugins`

crates/plugins/src/prompt.rs
kate-goldenring: I agree. This can just be a function. changed it!

crates/plugins/src/install.rs
kate-goldenring: Good point. I fixed this by removing an uninstall for downgrade scenarios and just installing over it.

crates/plugins/src/uninstall.rs
kate-goldenring: We have a `PluginStore`! 🛍️

kate-goldenring:
Thank you @itowlson for this incredible review! This should be in much better shape now. The following still need to be addressed:
- Executing plugins as subcommands: Should it be `spin example` or `spin plugin run example. Force mapping to plugins would resolve `clap` issues on mistyped supported commands and help with conflicts with plugin naming and internal commands (https://github.com/fermyon/spin/pull/735#discussion_r964371372)
- Should we add `update` that will conditionally pull the latest from `spin-plugins` repo. Or pull during every install
- Should we allow plugins to be installed if they do not pass the version check? This was the original idea but removed from the SIP after more conversation: https://github.com/fermyon/spin/pull/712#discussion_r952981227
- I am not sure I have appropriately moved the parsing component to the CLI as you suggested @itowlson

src/commands/external.rs
kate-goldenring: I added the above, so if its unrecognized, it says so and prints help info. However, i think force mapping plugins would lead to an even better ux

crates/plugins/src/install.rs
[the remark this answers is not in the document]
itowlson: I'm not too fussed about whether the `plugins` crate _offers_ a way of interaction; I was only bothered that it _mandated_ a way of interaction.

crates/plugins/src/install.rs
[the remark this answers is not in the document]
itowlson: Ah, I was thinking it could be parsed once and then passed around in parsed form.  Never mind then!

src/commands/external.rs
itowlson: I do like the convenience of being able to write `spin test` etc. - from the user point of view it would be nice not to have to care if a command is built in or contributed by a plugin.  I would suggest leaving it as it is, and if the third-party plugin ecosystem takes off, we can add a `spin plugin run` command (and/or `spin plugin alias` or whatever else we come up with after lengthier thought) at that point.  I realise this risks storing up compatibility issues that force-mapping would avoid, though.  Happy to discuss further and more methodically!

itowlson:
@kate-goldenring for me the "override version check" is a way to bridge a mismatch between developers and users.  Consider:

* Alice writes a plugin and tests it against Spin 0.5, the latest at that time.  Alice doesn't know whether a future version of Spin will break her plugin, so wisely sets its compatibility to `0.5`.
* Bob installs Alice's plugin and loves it.
* Carol adds MySQL support and releases it as Spin 0.6, and Bob (who loves MySQL almost as much as he loves Alice's plugin) upgrades.  Oh no!  Bob's favourite plugin stops working.
* Bob raises an issue on the plugin repo.  But Alice has gone on a well-deserved holiday, and won't be back for two months...

The trouble is partly that we're not on useful semver yet, so the change from 0.5 to 0.6 communicates little - they might be backward compatible, or might not.  The author has to be conservative and assume minor or patch releases could be breaking, but that means the user gets hosed when a release is only a bugfix or new feature, yet their plugins stop working anyway.

Hope that makes sense.  I do take the point that the user could hose themselves by running an incompatible plugin, so a dire warning is fine - but blocking it completely feels too strict for a project where users are on the bleeding edge anyway!  As always, happy to discuss further.

(And apologies for not flagging this at the SIP stage - I realise how frustrating it is to have it relitigated.)

kate-goldenring:
@itowlson that does make sense. My only concern is what do we do when plugins are executed? We currently are also checking version there in `external.rs` because what if:
- install spin `v0.4.0`
- install plugin `example` which supports `=0.4.0`
- upgrade to spin `v0.5.0` 
- run `spin example`

Currently we throw an error (and should also recommend upgrading). I am not sure how we could pass an `override-version-check` flag to an external command. We could just log here that are running a version that is incompatible and continue. But that doesnt seem ideal either.

src/commands/external.rs
```diff
@@ -0,0 +1,75 @@
...
+    match plugin_store.read_plugin_manifest(plugin_name) {
+        Ok(manifest) => {
+            let spin_version = env!("VERGEN_BUILD_SEMVER");
+            if let Err(e) = check_supported_version(&manifest, spin_version) {
+                eprintln!("{}", e);
+                eprintln!(
+                    "Try running `spin plugin upgrade {}` to get the latest version of the plugin.",
+                    manifest.name()
+                );
+                process::exit(1);
```
kate-goldenring, line 23: TODO: status should be `2` for incorrect usage

src/commands/external.rs
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+        Err(e) => {
+            // If a manifest file cannot be found for a plugin with the given name
+            // error that the command does not exist.
+            if let Some(io_e) = e.downcast_ref::<std::io::Error>() {
+                if io_e.kind() == std::io::ErrorKind::NotFound {
+                    eprintln!("Unknown command.");
+                    app.clone().print_help()?;
+                    process::exit(1);
```
kate-goldenring, line 33: TODO: status code should be 2 for incorrect use

karthik2804:
We could pass the `override-version-check` to the `external` command as a positional argument. The argument can be more descriptive to ensure the user knows the risk. 

The positional argument can be either structured as 

```
spin <version_check_flag> <plugin> <args> 
```

or 

```
spin <plugin> <args> <version_check_flag>
```

itowlson:
@kate-goldenring Another possibility is to prompt if a version mismatch is detected, with options of run, don't run, and "I know this is safe, don't ask again."  If the user chooses "don't ask again," it would apply only to the exact combination of plugin and Spin versions.  E.g. if the user upgraded to Spin 0.5.1 then we'd prompt for `spin example` again.

However, we'd still likely need @karthik2804's "skip version check" option as well, so that people could run "incompatible but known safe" plugins in CI.  So this is strictly more effort, and we'd need to decide if we felt the added convenience was worth it.

We'd also need to pay attention to how any prompt worked in CI environments.  You don't want to upgrade your CI environment to 0.5.0 and find all your workflows stall without meaningful error!  Maybe this is a can of worms...  (Or maybe this wouldn't be an issue because without intervention they'd fail at the plugin install stage.)

kate-goldenring:
@itowlson prompting seems feasible. A few follow up thoughts/questions:
1. Is the current way prompting done sufficient? I want to make sure i am addressing that initial concern of where that lived before extending it for incompatibility prompting.
2. would the `-y` / `yes-to-all` flag also auto yes for this (on install)?
3. For `dont-ask-again`, we'd need to store this state. We could store this state by writing to the json manifests some metadata that demarcates this.

crates/plugins/src/plugin_manifest.rs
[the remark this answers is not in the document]
kate-goldenring: I like how JSON schemas enable specifying field options (ie OS enum), verify names (regex and enums), can specify manditory and optional fields, and act as documentation on how to construct a manifest. However, the Spin manifest tomls are refreshingly easy to read. I wonder if we can make manifests TOML but when validating them, convert them to JSON and check them against a schema. That could give us the best of both worlds. (Unfortunately the TOML schema parser is unmaintained: https://github.com/json-schema-everywhere/pajv, which is which the initial convert to JSON step may be needed)

itowlson:
1. Yes, the current way of prompting is okay.

2. Ooh... that's a good question.  We'd be combining two things if it did - skip "are you sure" prompts on the happy path, and skip "wait are you REALLY sure" on the not so happy path.  The latter feels like a distinct decision, yet would be needed so that CI environments didn't blindly install incompatible plugins - this is @karthik2804's override option.  But having multiple `yes` flags might be overkill - would we be able to document it clearly?  And how would we handle CI when the flag *wasn't* passed - would they block?

Maybe this is too much of a can of worms.  I am struggling rather with how to offer an experience that would be safe in CI but forgiving to interactive users.  Perhaps always failing on incompatibility (and allowing Karthik's flag to override) is safer...

itowlson:
I am conscious the whole override thing is relitigating the SIP so I'm happy to drop it for this PR if you prefer - nothing stops us coming back to it later.

crates/plugins/src/plugin_manifest.rs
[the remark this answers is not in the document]
kate-goldenring: I played around with this in this [branch of `spin-plugins`](https://github.com/kate-goldenring/spin-plugins/blob/toml-manifests/README.md#validating-plugin-schemas)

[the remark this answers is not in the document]
kate-goldenring: It feels a little messy. In the end, @itowlson, I think your note summarizes it: spin plugin manifests are less frequently authored and can use the benefits of json tooling (schema validation). However, spin app manifests are validated when they are run, so they dont need json schema validation. So i think using json for spin plugin manifests and toml for app manifests is appropriate

kate-goldenring:
@bacongobbler what are your thoughts on adding an `--override-version-check` flag: https://github.com/fermyon/spin/pull/735#issuecomment-1241448659? It opposes the original logic of the SIP that was motivated in part by your comment here https://github.com/fermyon/spin/pull/712#discussion_r952981227.

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,220 @@
...
+        spin_version: &str,
+        allow_downgrades: bool,
+    ) -> Result<()> {
+        // Disallow installing plugins with the same name as spin internal subcommands
+        if SPIN_INTERNAL_COMMANDS
+            .iter()
+            .any(|&s| s == plugin_manifest.name())
+        {
+            bail!(
+                "Can't install a plugin with the same name ({}) as an internal plugin",
```
itowlson, line 116: I do think `internal command` rather than `internal plugin`.  If it's internal, users won't think of it as a plugin, even if we implement it that way.

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,220 @@
...
+                plugin_manifest.name()
+            );
+        }
+
+        // Disallow reinstalling identical plugins and downgrading unless permitted.
+        if let Ok(installed) = self.store.read_plugin_manifest(&plugin_manifest.name()) {
+            if &installed == plugin_manifest
+                || (installed.version > plugin_manifest.version && !allow_downgrades)
```
itowlson, line 124: nit: Suggest distinguishing the "you already have that" and "you have a newer version of that" cases in the error.

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,220 @@
...
+                    .await?
+                    .json::<PluginManifest>()
+                    .await?
+            }
+            ManifestLocation::Local(path) => {
+                log::info!("Pulling manifest for plugin from {}", path.display());
+                let file = File::open(path).with_context(|| {
+                    format!("The local manifest {} could not be opened", path.display())
+                })?;
+                serde_json::from_reader(file)?
```
itowlson, line 157: Throughout this function it would be really really good to get some context on the various possible error returns (unless that's provided by whatever calls this, I'm not sure).  Like if JSON deserialisation fails, will the user have any more guidance than a serde "expected integer, got biscuits" kind of error?

crates/plugins/src/prompt.rs
```diff
@@ -0,0 +1,44 @@
+use crate::manifest::{PluginManifest, PluginPackage};
+use anyhow::Result;
+use std::io;
+
+/// Prompts user as to whether they trust the source of the plugin and
+/// want to proceed with installation.
+fn print_prompt(name: &str, license: &str, source_url: &str) {
```
itowlson, line 7: Suggest either renaming this to something like "print_confirm_install_prompt" or having the prompt text as a parameter - at the moment it is a very generic name for a very specific prompt.

crates/plugins/src/prompt.rs
```diff
@@ -0,0 +1,44 @@
...
+    Ok(parse_response(&mut resp))
+}
+
+fn parse_response(resp: &mut str) -> bool {
+    let resp = resp.trim().to_lowercase();
+    resp == "yes" || resp == "y"
+}
+
+/// Returns whether or not the user would like to proceed with the installation of a plugin.
+pub fn prompt(manifest: &PluginManifest, package: &PluginPackage) -> Result<bool> {
```
itowlson, line 27: Again, suggest e.g. `prompt_confirm_install` so readers at the call site know what is being prompted for.

crates/plugins/src/prompt.rs
```diff
@@ -0,0 +1,44 @@
...
+    print_prompt(&manifest.name(), &manifest.license, &package.url);
+    let sure = are_you_sure()?;
+    if !sure {
+        println!("Plugin {} will not be installed", manifest.name());
+    }
+    Ok(sure)
+}
```
itowlson, line 34: super nit: blank line before start of test code

src/commands/plugins.rs
```diff
@@ -0,0 +1,295 @@
...
+            &manager,
+            self.yes_to_all,
+            self.downgrade,
+        )
+        .await?;
+        Ok(())
+    }
+}
+
+async fn update() -> Result<()> {
```
itowlson, line 265: Suggest saying *what* it updates

itowlson (changes requested):
@kate-goldenring @karthik2804 Thanks for this - to me at least this iteration is much easier to follow, and it partitions responsibilities in a way I really liked.  There are a few things I think still need a bit of tweaking, but they are local (e.g. more info on errors, renaming here and there) rather than structural.  Great stuff!

crates/plugins/src/lookup.rs
```diff
@@ -0,0 +1,104 @@
...
+        let manifest: PluginManifest = serde_json::from_reader(file)?;
+        Ok(manifest)
+    }
+}
+
+pub async fn fetch_plugins_repo(plugins_dir: &Path, update: bool) -> Result<()> {
+    let repo_url = Url::parse(SPIN_PLUGINS_REPO)?;
+    let git_root = plugin_manifests_repo_path(plugins_dir);
+    let git_source = GitSource::new(&repo_url, None, &git_root)?;
+    if !git_root.join(".git").exists() {
```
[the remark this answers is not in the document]
kate-goldenring: I am not sure inverted works here given there is no else condition. If the git directory exists, nothing is to be done unless update is specified. The following adheres to the if-else but is more nested:
```
if git_root.join(".git").exists() {
    if update {
        git_source.pull().await?;
    }
} else {
    git_source.clone().await?;
}
```

Is this clearer?

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,220 @@
...
+            _ => download_plugin(&plugin_manifest.name(), &temp_dir, &target).await?,
+        };
+        verify_checksum(&plugin_tarball_path, &plugin_package.sha256)?;
+
+        self.store
+            .untar_plugin(&plugin_tarball_path, &plugin_manifest.name())?;
+
+        // Save manifest to installed plugins directory
+        self.store.add_manifest(plugin_manifest)?;
+        println!(
```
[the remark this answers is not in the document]
kate-goldenring: To make sure I'm following, value returned could be name of the plugin that was installed/uninstalled so then the calling function can choose to print a confirmation.

crates/plugins/src/lookup.rs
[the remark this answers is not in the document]
kate-goldenring: I am not sure inverted works here given there is no else condition. If the git directory exists, nothing is to be done unless update is specified. The following adheres to the if-else but is more nested:

if git_root.join(".git").exists() {
    if update {
        git_source.pull().await?;
    }
} else {
    git_source.clone().await?;
}
Is this clearer?

crates/plugins/src/lookup.rs
```diff
@@ -0,0 +1,104 @@
...
+        ))
+        .with_context(|| {
+            format!(
+                "{} {} {} in centralized repository",
+                self.name,
+                self.version_string(),
+                PLUGIN_NOT_FOUND_ERROR_MSG,
+            )
+        })?;
+        let manifest: PluginManifest = serde_json::from_reader(file)?;
```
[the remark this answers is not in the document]
kate-goldenring: Thanks! Added context on all serde deserialization and on reqwest parsing: d9d9417a53ff6a22

crates/plugins/src/lookup.rs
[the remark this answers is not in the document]
kate-goldenring: added the above in d9d9417a53ff6a22

crates/plugins/src/manager.rs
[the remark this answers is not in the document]
kate-goldenring: made install return name of plugin that was installed and uninstall return a bool as to whether was installed. Love this addition of keeping the ui printing outside of the lib: d9d9417a53ff6a22

crates/plugins/src/lookup.rs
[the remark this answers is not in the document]
itowlson: I like that more - in fact, as you may have guessed, my eyes saw the `else` and completely slid over the `if update` part, so having it more visible is helpful, even with the clutter of the extra nesting.  Of course, this is subjective!

crates/plugins/src/manager.rs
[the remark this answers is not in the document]
itowlson: Yeah, that was the sort of thing I meant.  Thanks!

itowlson (approved):
I really like how this is looking!  And I'm excited to see where this leads us in the future!  :shipit:

bacongobbler:
> @bacongobbler what are your thoughts on adding an `--override-version-check` flag: [#735 (comment)](https://github.com/fermyon/spin/pull/735#issuecomment-1241448659)?

@itowlson raises a good point. Because Spin is not yet 1.0, we cannot guarantee compatibility for plugin developers between Spin 0.x and Spin 0.x+1. It is very likely that we'll have plugin developers pin their plugins to a specific release for the time being until we can guarantee backwards compatibility (1.0.0)

Post-1.0, the same situation could occur if Spin 2.0.0 was released and Alice was no longer maintaining the plugin. Bob would likely want to use Alice's plugin against Spin 2.0, but because Alice stepped away from the project, nobody is available to update it. Though in that case, perhaps Bob should consider forking Alice's plugin to ensure compatibility with 2.0.0...

Given that

1. This is an opt-in flag that will help plugin users in the event of a "break in case of emergency" event, and
2. It still provides the plugin author some form of protection from letting their users install plugins that were incompatible with this version of Spin

I'm okay with the feature flag.

One point of feedback that you're welcome to disregard: have you thought about putting this behind an interactive prompt window instead? Say for example Bob was installing Alice's plugin with Spin 0.6:

```
$ spin plugin install plugin-by-alice
WARNING: This plugin has not been tested for compatibility with Spin v0.6.0. Proceed? (y/N) y
Installing...
```

With `--override-version-check` bypassing the prompt.

kate-goldenring:
@bacongobbler @itowlson sounds like we've reached consensus on adding an `--override-version-check` (or maybe better `--override-compatibility-check`) flag. For the sake of stabilizing the state of this PR, let's add that to a subsequent PR after this.

tests/integration.rs
```diff
@@ -870,6 +888,113 @@ mod integration_tests {
...
+                {
+                    "os": "macos",
+                    "arch": "amd64",
+                    "url": "file:PATH_TO_TESTS/tests/plugin/example.tar.gz",
+                    "sha256": "f7a5a8c16a94fe934007f777a1bf532ef7e42b02133e31abf7523177b220a1ce"
+                }
+            ]
+        }"#;
+        let plugin_manifest_json = plugin_manifest_json_template
+            .replace("PATH_TO_TESTS", path_to_test_dir.to_str().unwrap());
```
lann, line 938: You might like `serde_json::json!` for this, which lets you write things like:

```rust
let plugin_manifest_json = serde_json::to_vec(serde_json::json!({
  // ...
  "url": format!("file:{PATH_TO_TESTS}/tests/plugin/example.tar.gz")
  // ...
})).unwrap();
```

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub async fn install(
+        &self,
+        plugin_manifest: &PluginManifest,
+        plugin_package: &PluginPackage,
+    ) -> Result<String> {
+        let target = plugin_package.url.to_owned();
+        let target_url = Url::parse(&target)?;
+        let temp_dir = tempdir()?;
+        let plugin_tarball_path = match target_url.scheme() {
+            URL_FILE_SCHEME => PathBuf::from(target_url.path()),
```
lann, line 63: Something like this might behave better on Windows:

```suggestion
            URL_FILE_SCHEME => target_url.to_file_path()
                .map_err(|_| anyhow!("Invalid file URL: {target_url:?}")),
```

https://docs.rs/url/2.3.1/url/struct.Url.html#method.to_file_path

Note the oddity here that `to_file_path` returns `Result<PathBuf, ()>` and since `()` doesn't `impl Error` it means `context` doesn't work... :shrug:

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,219 @@
...
+        let manifest_file = plugin_store.installed_manifest_path(plugin_name);
+        match manifest_file.exists() {
+            // Remove the manifest and the plugin installation directory
+            true => {
+                fs::remove_file(manifest_file)?;
+                fs::remove_dir_all(plugin_store.plugin_subdirectory_path(plugin_name))?;
+                Ok(true)
+            }
+            false => Ok(false),
+        }
```
lann, line 90:
```suggestion
        let exists = manifest_file.exists()
        if exists {
            // Remove the manifest and the plugin installation directory
            fs::remove_file(manifest_file)?;
            fs::remove_dir_all(plugin_store.plugin_subdirectory_path(plugin_name))?;
        }
        Ok(exists)
```

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,219 @@
...
+        &self,
+        manifest_location: &ManifestLocation,
+    ) -> Result<PluginManifest> {
+        let plugin_manifest = match manifest_location {
+            ManifestLocation::Remote(url) => {
+                log::info!("Pulling manifest for plugin from {}", url);
+                reqwest::get(url.as_ref())
+                    .await?
+                    .error_for_status()
+                    .with_context(|| format!("Manifest endpoint {} is unhealthy", url))?
```
lann, line 146: "unhealthy" may be a little confusing here; it really just means the request failed, right?

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,219 @@
...
+                    .json::<PluginManifest>()
+                    .await
+                    .with_context(|| format!("Failed to deserialize remote plugin at {}", url))?
+            }
+            ManifestLocation::Local(path) => {
+                log::info!("Pulling manifest for plugin from {}", path.display());
+                let file = File::open(path).with_context(|| {
+                    format!("The local manifest {} could not be opened", path.display())
+                })?;
+                serde_json::from_reader(file).with_context(|| format!("Failed to deserialize local manifest at {}. Ensure it has the expected plugin manifest format", path.display()))?
```
lann, line 156: Unless I'm missing something, `.display()` should be redundant here, implied by `{}`. Also you can (optionally) inline simple variables if you like:

```suggestion
                log::info!("Pulling manifest for plugin from {path}");
                let file = File::open(path).with_context(|| {
                    format!("The local manifest {path} could not be opened")
                })?;
                serde_json::from_reader(file).with_context(|| format!("Failed to deserialize local manifest at {path}. Ensure it has the expected plugin manifest format"))?
```

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,219 @@
...
+    plugin_manifest
+        .packages
+        .iter()
+        .find(|p| p.os.to_string() == os && p.arch.to_string() == arch)
+        .ok_or_else(|| {
+            anyhow!(
+                "This plugin does not support this OS ({}) or architecture ({}).",
+                os,
+                arch
+            )
```
lann, line 181: You could bring symbols into function-local scope instead of assigning here:

```suggestion
    use std::env::consts::{ARCH, OS};
    plugin_manifest
        .packages
        .iter()
        .find(|p| p.os.to_string() == OS && p.arch.to_string() == ARCH)
        .ok_or_else(|| {
            anyhow!(
                "This plugin does not support this OS ({OS}) or architecture ({ARCH}).")
```

crates/plugins/src/manifest.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl ToString for Architecture {
+    fn to_string(&self) -> String {
+        match self {
+            Self::Amd64 => "x86_64".to_string(),
+            Self::Aarch64 => "aarch64".to_string(),
+            Self::Arm => "arm".to_string(),
+        }
+    }
+}
```
lann, line 86: I wouldn't use `ToString` for this; generally `ToString` should match `Display`, which is output for humans to read. That's confusing here because the output won't match the serialized input.

Instead, pick a method name that reflects what these are used for, e.g. `rust_arch`.

crates/plugins/src/manifest.rs
```diff
@@ -0,0 +1,221 @@
...
+// https://doc.rust-lang.org/std/env/consts/constant.OS.html
+impl ToString for Os {
+    fn to_string(&self) -> String {
+        match self {
+            Self::Linux => "linux".to_string(),
+            Self::Macos => "macos".to_string(),
+            Self::Windows => "windows".to_string(),
+        }
+    }
+}
```
lann, line 67: See comment below.

While this one is less of a problem because it does match the serialization, I would still move it to a separate method for consistency.

crates/plugins/src/store.rs
```diff
@@ -0,0 +1,117 @@
...
+
+/// Houses utilities for getting the path to Spin plugin directories.
+pub struct PluginStore {
+    root: PathBuf,
+}
+
+impl PluginStore {
+    pub fn new(root: impl AsRef<Path>) -> Self {
+        Self {
+            root: root.as_ref().to_owned(),
```
lann, line 22: As a general pattern:

```suggestion
    pub fn new(root: impl Into<PathBuf>) -> Self {
        Self {
            root: root.into(),
```

crates/plugins/src/store.rs
```diff
@@ -0,0 +1,117 @@
...
+            binary.set_extension("exe");
+        }
+        binary
+    }
+
+    /// Returns the PluginManifest for an installed plugin with a given name.
+    /// Looks up and parses the JSON plugin manifest file into object form.
+    pub fn read_plugin_manifest(&self, plugin_name: &str) -> Result<PluginManifest> {
+        let manifest_path = self.installed_manifest_path(plugin_name);
+        log::info!("Reading plugin manifest from {}", manifest_path.display());
```
lann, line 70: As with `format!`, you can inline simple variables and `.display()` is redundant.

```suggestion
        log::info!("Reading plugin manifest from {manifest_path}");
```

src/commands/external.rs
```diff
@@ -0,0 +1,75 @@
...
+                }
+            }
+            return Err(e);
+        }
+    }
+
+    let mut command = Command::new(plugin_store.installed_binary_path(plugin_name));
+    if args.len() > 1 {
+        command.args(&args[1..]);
+    }
```
lann, line 43: (if the `plugin_name` arg suggestion above is taken)

```suggestion
    command.args(args);
```

src/commands/plugins.rs
```diff
@@ -0,0 +1,305 @@
...
+            return Ok(());
+        }
+
+        if self.all {
+            self.upgrade_all(manifests_dir).await
+        } else {
+            let plugin_name = self
+                .name
+                .clone()
+                .ok_or_else(|| anyhow!("plugin name is required for upgrades"))?;
```
lann, line 214: As a nice convenience, `anyhow::Context` is also implemented for `Option`.
```suggestion
                .context("plugin name is required for upgrades")?;
```

lann (commented):
Mostly minor style nits / suggestions. Nothing impacting correctness except potentially the `to_string_lossy` stuff, though that's pretty _unlikely_ to cause problems.

No blockers from my perspective.

crates/plugins/src/manager.rs
kate-goldenring: Interesting point! I looked into this because casting to `Path` display type does not work and gets an error that path does not implement Display. Rather it still has a `display()` method that can be used that works around potential Unicode errors. `"{path}"` is disallowed and `path.display()` is the method encouraged by the docs: https://doc.rust-lang.org/stable/std/path/struct.Display.html

src/commands/external.rs
```diff
@@ -0,0 +1,75 @@
...
+use spin_plugins::{manifest::check_supported_version, PluginStore};
+use std::{collections::HashMap, env, process};
+use tokio::process::Command;
+use tracing::log;
+
+/// Executes a Spin plugin as a subprocess, expecting the first argument to
+/// indicate the plugin to execute. Passes all subsequent arguments on to the
+/// subprocess.
+pub async fn execute_external_subcommand(args: Vec<String>, app: App<'_>) -> Result<()> {
+    let plugin_name = args.first().ok_or_else(|| anyhow!("Expected subcommand"))?;
```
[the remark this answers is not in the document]
kate-goldenring: clap doesn't allow defining complex types for `external_subcommand`: `The type must be either `Vec<String>` or `Vec<OsString>` to be used with `external_subcommand`.` It would be preferable to have the cli input immediately parsed by clap as a command and vector of args but clap doesnt support this.
I did try to make this cleaner by clearly separating the command/plugin name from the rest of the arguments in the first line:
```
let (plugin_name, args) = cmd
        .split_first()
        .ok_or_else(|| anyhow!("Expected subcommand"))?;
```

crates/plugins/src/git.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub async fn clone(&self) -> Result<()> {
+        let mut git = Command::new("git");
+        git.args([
+            "clone",
+            self.source_url.as_ref(),
+            "--branch",
+            &self.branch,
+            "--single-branch",
+            &self.git_root.to_string_lossy(),
+        ]);
```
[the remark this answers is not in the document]
kate-goldenring: Thanks for pointing this out! This is because some inportant information may be dropped in the lossy conversion, correct? So data used by functions shouldnt undergo this but if its for logging information, it isnt risky?

crates/plugins/src/store.rs
kate-goldenring: same as above ... cannot display `Path` this way unfortunately

crates/plugins/src/manifest.rs
kate-goldenring: It matched at first and then I switched the mappings -- thanks for catching this! I made them functions on the `PluginManifest` struct. Let me know if you think they would be better suited as independent functions.

crates/plugins/src/manager.rs
kate-goldenring: Good point. This now maps to a custom `Error::ConnectionFailed`.

src/commands/external.rs
```diff
@@ -0,0 +1,75 @@
...
+                    "Try running `spin plugin upgrade {}` to get the latest version of the plugin.",
+                    manifest.name()
+                );
+                process::exit(1);
+            }
+        }
+        Err(e) => {
+            // If a manifest file cannot be found for a plugin with the given name
+            // error that the command does not exist.
+            if let Some(io_e) = e.downcast_ref::<std::io::Error>() {
```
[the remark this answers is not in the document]
kate-goldenring: Thanks for the bump here. Implemented some Error options with `thiserror`

tests/integration.rs
kate-goldenring: This was much cleaner. Thanks!

kate-goldenring:
@lann Thank you for the great pointers here -- learned a lot! I think I addressed your comments -- but happy to iterate if you have more feedback.

crates/plugins/src/manager.rs
lann: >  Unless I'm missing something

I was missing something!

lann: Looking at some of my own code I realized why I was confused; I almost always use Debug output for logging/errors, e.g. `{path:?}` which _is_ supported by `Path`. The Debug formatter adds quotes and escapes special chars which may or not be desirable in different situations.

crates/plugins/src/error.rs
```diff
@@ -0,0 +1,89 @@
+pub type PluginLookupResult<T> = std::result::Result<T, Error>;
```
lann, line 1: For better or worse these `Result` aliases are typically just called `Result`, e.g. `std::{fmt,io,thread}::Result`.

On the other hand its also common to e.g. `use std::io::Result as IoResult};`... :thinking:

I guess I don't actually have a strong opinion on this one, but something to be aware of.

crates/plugins/src/error.rs
```diff
@@ -0,0 +1,89 @@
...
+    NotFound(NotFoundError),
+
+    #[error("{0}")]
+    ConnectionFailed(ConnectionFailedError),
+
+    #[error("{0}")]
+    InvalidManifest(InvalidManifestError),
+
+    #[error("URL parse error {0}")]
+    UrlParseError(url::ParseError),
```
lann, line 16: A convenience from `thiserror`:

```suggestion
    UrlParseError(#[from] url::ParseError),
```

This generates a `impl From<url::ParseError> for Error` for you, which allows `?` to convert types for you without an explicit `map_err`.

crates/plugins/src/error.rs
```diff
@@ -0,0 +1,89 @@
...
+    pub fn new(name: Option<String>, addr: String, err: String) -> Self {
+        Self { name, addr, err }
+    }
+}
+
+impl std::fmt::Display for NotFoundError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&format!(
+            "plugin '{}' not found at expected location {}: {}",
+            self.name.clone().unwrap_or_default(),
```
lann, line 37: Makes no practical difference in this specific case, but as a general pattern that comes up pretty frequently:

```suggestion
            self.name.as_deref().unwrap_or_default(),
```

`as_deref` borrows an `Option<&str>` from the `Option<String>`.

crates/plugins/src/manager.rs
```diff
@@ -60,7 +61,9 @@ impl PluginManager {
         let target_url = Url::parse(&target)?;
         let temp_dir = tempdir()?;
         let plugin_tarball_path = match target_url.scheme() {
-            URL_FILE_SCHEME => PathBuf::from(target_url.path()),
+            URL_FILE_SCHEME => target_url
+                .to_file_path()
+                .map_err(|_| anyhow!("Invalid file URL: {target_url:?}"))?,
```
lann, line 66: I _almost_ just suggested `.context` here, which goes to show how confusing `to_file_path` is :sweat_smile:

crates/plugins/src/manifest.rs
```diff
@@ -30,6 +30,26 @@ impl PluginManifest {
     pub fn name(&self) -> String {
         self.name.to_lowercase()
     }
+
+    // Maps manifest Architecture options to associated Rust ARCH strings
+    // https://doc.rust-lang.org/std/env/consts/constant.ARCH.html
+    pub(crate) fn rust_arch(arch: &Architecture) -> String {
```
lann, line 36: I would expect these to be methods on either `PluginPackage` or on `Architecture`/`Os` (in which case they could be named e.g. `rust_name`).

crates/plugins/src/lookup.rs
```diff
@@ -33,49 +29,53 @@ impl PluginLookup {
...
-                "{} {} {} in centralized repository",
-                self.name,
-                self.version_string(),
-                PLUGIN_NOT_FOUND_ERROR_MSG,
-            )
+    pub async fn get_manifest_from_repository(
+        &self,
+        plugins_dir: &Path,
+    ) -> PluginLookupResult<PluginManifest> {
+        let url = plugins_repo_url().map_err(Error::UrlParseError)?;
```
lann, line 36: With the `thiserror` `#[from]` suggestion above:

```suggestion
        let url = plugins_repo_url()?;
```

lann (approved):
Looks good! A few more style suggestions on the changes, but again nothing blocking.

crates/plugins/src/manifest.rs
```diff
@@ -0,0 +1,206 @@
...
+}
+
+impl Os {
+    // Maps manifest OS options to associated Rust OS strings
+    // https://doc.rust-lang.org/std/env/consts/constant.OS.html
+    pub(crate) fn rust_name(os: &Os) -> String {
+        match os {
+            Os::Linux => "linux".to_string(),
+            Os::Macos => "macos".to_string(),
+            Os::Windows => "windows".to_string(),
```
lann, line 64: Enums can have methods just like structs. Also, because these are static strings, you can return `&'static str` instead of `String`:

```suggestion
    pub(crate) fn rust_name(&self) -> &'static str {
        match os {
            Os::Linux => "linux",
            Os::Macos => "macos",
            Os::Windows => "windows",
```

crates/plugins/src/manager.rs
```diff
@@ -0,0 +1,238 @@
...
+    }
+}
+
+/// Gets the appropriate package for the running OS and Arch if exists
+pub fn get_package(plugin_manifest: &PluginManifest) -> Result<&PluginPackage> {
+    use std::env::consts::{ARCH, OS};
+    plugin_manifest
+        .packages
+        .iter()
+        .find(|p| Os::rust_name(&p.os) == OS && Architecture::rust_name(&p.arch) == ARCH)
```
lann, line 202: With above:

```suggestion
        .find(|p| p.os.rust_name() == OS && p.arch.rust_name() == ARCH)
```

lann (approved):
Code LGTM. Unfortunately I didn't participate as much as others in the SIP review so I don't have much to say on functionality.

radu-matei (approved):

PR #740, opened 2022-09-07
Validate name when instantiating template

### LINKED ISSUE #739: Using a digit as the starting character of a rust component fails to build
```
🐟 > spin new http-rust 34
Project description: description
HTTP base: /
HTTP path: /...

mikkel on laptop ~/code
🐟 > cd 34

mikkel on laptop ~/code/34 is 📦 v0.1.0 via 🦀 v1.63.0
🐟 > spin build
Executing the build command for component 34: cargo build --target wasm32-wasi --release
error: failed to parse manifest at `/Users/mikkel/code/34/Cargo.toml`

Caused by:
  the name `34` cannot be used as a package name, the name cannot start with a digit
Error: Build command for component 34 failed with status Exited(101).
```

itowlson:
Fixes #739.

Output:

```
$ spin new http-rust 123test
error: Invalid value "123test" for '<NAME>': Name must start with a letter
```

Notes for reviewers:

1. This validates in the CLI, rather than in the template crate itself. Is this reasonable or should we push the validation into that shared core?

2. This makes no consideration for language-specific naming rules.  E.g. if a language used the name only for directory and component names, this restriction would be unnecessary.  Is this okay or should we move the validation somewhere that understands per-language or per-template constraints?

3. This forbids problematic names.  An alternative approach would be to transform them into legal ones, which could be done via Liquid filters in the templates (again allowing more language-specific understanding) - e.g. transform `34` to `spinapp34`.  What do we prefer?

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -18,6 +18,7 @@ pub struct NewCommand {
     pub template_id: String,
 
     /// The name of the new application or component.
+    #[clap(value_parser = validate_name)]
     pub name: String,
 
     /// The directory in which to create the new application or component.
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):
Looks good to me.

PR #753, opened 2022-09-09
Remove double slash from file URIs

lann:
While valid for absolute URIs, the double slashes cannot work with relative URIs and are usually redundant.


fibonacci1729 (approved):
Good catch.

PR #756, opened 2022-09-11
Minor change: Changed 'API' to 'SDK'.

coderoflagos:
Signed-off-by: Opemipo Disu <46439130+coderoflagos@users.noreply.github.com>

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -32,7 +32,7 @@ route = "/hello"
 
 The following are the fields supported by the `spin.toml` manifest file:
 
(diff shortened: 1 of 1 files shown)


itowlson (changes requested):
This field doesn't refer to the SDK version.  I definitely agree that 'API' is ambiguous and would benefit from being clearer, but what it refers to here is the Spin manifest format, i.e. how to interpret the rest of the `spin.toml` file.

coderoflagos:
Oh, true!

PR #729, opened 2022-09-01
Install script for Spin

mikkelhegn:
Signed-off-by: Mikkel Mørk Hegnhøj <mikkel@fermyon.com>

This is the first commit for an installer script for Spin, to make the acquisition story easier.

Current progress on testing:

- [x] Tested on MacBook with Apple Silicon
- [ ] Tested on MacBook with Intel
- [x] Tested on Linux with Intel
- [x] Tested on Linus with ARM processor (should fail)
- [x] Tested on WSL
- [x] Get the latest version from GitHub - it's currently hard-coded in the script
- [x] Support installing a specific version (including canary), through an argument (e.g., install.sh -v canary)

A subsequent PR could have the equivalent for PowerShell.

Relates to #641

PR #757, opened 2022-09-12
Set up with the Intel-Based Macs release.

coderoflagos:
Signed-off-by: Opemipo Disu <46439130+coderoflagos@users.noreply.github.com>


lann:
It seems strange to have _two_ specific examples here, when there are only 4 supported platforms. Having one examples with a link to the rest makes sense, and I can imagine listing all 4 links.

mikkelhegn:
Superseded by #729

PR #759, opened 2022-09-12
Rename `spin-http-engine` to `spin-http`

lann:
This always confused me...


fibonacci1729 (approved):

fibonacci1729:
LGTM, but missed a spot.

PR #761, opened 2022-09-13
Allow templates to define custom Liquid filters

itowlson:
This is infrastructure which will allow us to address the .NET namespacing bug (and similar bugs in future) without requiring custom support in Spin for every language convention.

The idea here is that a template can define custom Liquid filters for its own use.  Custom filters are registered in a new `[[custom_filters]]` section in `spin-template.toml`.  An entry maps a name to a Wasm implementation.

```toml
[[custom_filters]]
name = "studly_caps"
wasm = "studly_caps.wasm"
```

This can then be invoked as if it were a built-in filter like `snake_case`:

```
// I've thought of a great new capitalisation convention
namespace {{ project-name | studly_caps }};
```

Filters are shipped in binary (`.wasm`) form in the templates `metadata/filters` directory.  A filter must export the interface defined in `wit/custom-filter.wit` - for now, custom filters process only strings, and cannot be parameterised.  Filters are executed with no permissions.

Questions:

1. Is this a reasonable way to structure how filters are included in a template?  E.g. are there better ways of expressing them in the TOML?  Is the filesystem design overly restrictive or insufficiently expressive?  (Specifically thinking about the possibility of writing filters in other ways in future.)
2. I really want filters to be pure functions, but had to link WASI because when I built a filter in Rust it required those imports.  Is the implementation safe and reasonable?
3. How should we approach versioning?  I have left the manifest version at 1, which I think is safe, but I am wondering if I should version the interface so that when we need to change it (say, to accept a variant of string or number) it doesn't become an undetectable ABI mismatch resulting in weird memory errors.


kate-goldenring:
> This is infrastructure which will allow us to address the .NET namespacing bug (and similar bugs in future) without requiring custom support in Spin for every language convention.

@itowlson can you provide a little more context on the namespacing bug? Is it an issue with the rendering of the namespace name syntax when/if a .NET template were added?

kate-goldenring:
Looks like this is the bug: https://github.com/fermyon/spin-dotnet-sdk/issues/36
And this is the example implementation that would build off this pr: https://github.com/fermyon/spin-dotnet-sdk/pull/40

itowlson:
@kate-goldenring That's correct - apologies for not including the links in the first place.

kate-goldenring (approved):
LGTM!

PR #763, opened 2022-09-13
Spin core refactor

lann:
This refactors Spin to use the new `spin-core` and `spin-app` crates that we've been developing in an internal prototype.

With a few small exceptions I have attempted to not change the visible behavior of `spin up`, but with so many updates I'm sure I've missed some things.

There are thousands of lines of changes here, so I've tried to split things up into reasonable commits for those who prefer to review that way.


crates/app/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+    app: App<'this>,
+}
+
+impl std::ops::Deref for OwnedApp {
+    type Target = App<'static>;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe {
+            // We know that App's lifetime param is for AppLoader, which is owned by self here.
+            std::mem::transmute::<&App, &App<'static>>(self.borrow_app())
```
dicej, line 96: This doesn't seem sound to me.  For example:

```rust
async fn foo(loader: impl Loader + Send + Sync + 'static, uri: String) -> Result<&'static AppLoader> {
  Ok(AppLoader::new(loader).load_owned_app(uri).await?.deref().loader)
}
```
That would return a dangling reference, wouldn't it?

crates/app/src/values.rs
```diff
@@ -0,0 +1,57 @@
...
+        &mut self,
+        key: impl Into<String>,
+        value: impl Serialize,
+    ) -> serde_json::Result<&mut Self> {
+        let value = serde_json::to_value(value)?;
+        self.0.insert(key.into(), value);
+        Ok(self)
+    }
+
+    pub fn build(&mut self) -> ValuesMap {
```
dicej, line 54: Minor nit: I generally expect the `build` method of a builder to take ownership of `self`, which makes it obvious that I can't use the builder anymore.  Here I see I can reuse the builder, but that's not super useful given that it's been cleared out, plus I wouldn't know it is cleared out without looking at the code.

So I'd like to see this either take ownership of `self` or a doc comment describing the behavior.

crates/app/src/lib.rs
lann: That wouldn't compile would it? `.deref()` returns a `&App<'static>` which would be borrowing from the intermediate `OwnedApp` returned from `load_owned_app`.

dicej: Right, but then I'm immediately getting the `loader` field from the `&App<'static>`, which is of type `&'static AppLoader`, i.e. does not borrow from the temporary `OwnedApp`.  So I think it would compile, but give me a minute and I'll find out for sure.

lann: Ohh yeah it could be used unsafely _in this crate_.

lann: I should probably just drop this. It's just a bit of convenience anyway.

dicej: My suggested change would be to replace the `Deref` impl with an inherent method in `OwnedApp` with a signature like:

```rust
fn as_app<'this>(&'this self) -> &'this App<'this>
```

lann: Yeah the first iteration just used `AsRef` but I got annoyed typing it while knowing that it deref _should_ be possible. I agree that it isn't worth it.

dicej: I wonder if the [recently-stabilized GAT feature](https://github.com/rust-lang/rust/pull/96709) might help here, assuming `Deref::Target` could be generic over a lifetime.  Might be worth experimenting with sometime.

crates/app/src/values.rs
lann: I moved this behavior into a new `take` method and updated `build` to consume self.

crates/app/src/lib.rs
lann: I forgot that you can't `impl AsRef` with this generic lifetime. Ended up calling the method `borrowed` since it reads a little better than `app.as_app()...`

dicej (approved):
Looks good to me.  Thanks for taking on this huge effort!

Mossaka:
This looks great to me! I especially love the design of DynamocHostComponent that can update host components' data in runtime (whene http request comes in), and the design of LockedComponent. 

I do wonder why do you decide to impl `allowed_http_hosts` as dynamic data in http. I thought this variable is statically decalred in `spin.toml` manifest file, and thus can be added to HostComponent Data before runtime. No? 

In addition, what is the use cases of `spin.lock`?

lann:
@Mossaka:
> I do wonder why do you decide to impl `allowed_http_hosts` as dynamic data in http. I thought this variable is statically decalred in `spin.toml` manifest file, and thus can be added to HostComponent Data before runtime. No?

HostComponent Data is currently constructed per-request, not at startup. Preinitialization of some data would be possible but I don't see much benefit for the current set of host components.

> In addition, what is the use cases of `spin.lock`?

One purpose is to provide a more stable interface between `spin up` and trigger executors, which should simplify the implementation of trigger executor plugins (see also #735).

crates/core/src/host_component.rs
```diff
@@ -0,0 +1,217 @@
...
+        (**self).build_data()
+    }
+}
+
+/// An opaque handle returned by [`crate::EngineBuilder::add_host_component`]
+/// which can be passed to [`HostComponentsData`] to access or set associated
+/// [`HostComponent::Data`].
+pub struct HostComponentDataHandle<HC: HostComponent> {
+    idx: usize,
+    _phantom: PhantomData<fn() -> HC::Data>,
```
kate-goldenring, line 80: I have not used `PhantomData` before so i am trying to get a sense of what it is being used for here. Is this helping map the data types for different languages for a host component?

crates/core/src/host_component.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    struct TestHC;
+
+    impl HostComponent for TestHC {
```
kate-goldenring, line 184: If you are just trying to test code flow here, you could use mockall instead of creating a test struct: https://docs.rs/mockall/latest/mockall/

crates/app/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+
+#![deny(missing_docs)]
+
+mod host_component;
+pub mod locked;
+pub mod values;
+
+use ouroboros::self_referencing;
+use serde::Deserialize;
+use spin_core::{wasmtime, Engine, EngineBuilder, StoreBuilder};
```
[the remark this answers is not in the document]
lann: The block above is from external crates and the one below is from modules from this crate. Just personal preference though we've discussed standardizing on it: https://github.com/fermyon/spin/pull/460 (the conclusion there being that it isn't worth the churn until `rustfmt` supports automating it)

crates/core/src/host_component.rs
lann: This would be good to discuss with the team. Mocking isn't _super_ common in the Rust ecosystem and there is always some learning overhead in new frameworks like that, but it would certainly be helpful in some cases like this.

crates/core/src/host_component.rs
lann: Rust requires that generic params be used in type definitions. In this case the struct represents a handle to some `HostComponent::Data` but doesn't actually hold any; `PhantomData` is a zero-sized placeholder to appease the type checker.

radu-matei:
Have switched to this branch as the Spin binary I am using on a daily basis, and I am hitting the error below when trying to run https://github.com/coderoflagos/bartholomew-sample.

> Rendering "/content/index.md": Capabilities insufficient (os error 76)

lann:
> Have switched to this branch as the Spin binary I am using on a daily basis, and I am hitting the error below when trying to run https://github.com/coderoflagos/bartholomew-sample.
> 
> > Rendering "/content/index.md": Capabilities insufficient (os error 76)

Bartholomew tries to create a cache file: https://github.com/fermyon/bartholomew/blob/c602220ee4da6b2a22b17f0b04ff8eba1b8adc75/src/content.rs#L133

Looks like Spin currently allows writing _new_ files even without `--allow-transitent-write`

lann:
Bartholomew issue should be fixed in https://github.com/fermyon/bartholomew/pull/135

radu-matei (approved):
I've been using a build from this branch as my main Spin binary for a while now and everything appears to be in order.

As we start working on various parts of the codebase, I am sure we will start finding small bits that we can improve, but overall, the improvements in this PR are great!

Thank you!

LGTM

PR #777, opened 2022-09-20
MySQL support

itowlson:
This adds MySQL in a similar manner to Postgres.

For now, I kept them in the same crate and did not yet change the name.  It's probably worth having a discussion on how we want to take database integrations forward - e.g. should we move to a connection resource model, should we make the WIT database-agnostic (specifying the database type as part of connection creation) instead of parallel APIs, how do we surface database-specific features, etc.

Underlying all this is the lurking issue of how we enable people to use their familiar, language-specific database libraries and idioms when those are often tied to things like socket implementations - but that may be more than we can tackle ourselves.


itowlson:
Superseded by #864.

Issue #776, opened 2022-09-20
Migrate spin-outbound-pg to use `tokio-postgres`


lann:
Now that Spin is using Wasmtime's async support, it would be good to switch the outbound PG crate to use an async library. It currently uses the `postgres` crate which is just a blocking wrapper around `tokio-postgres`, so presumably migrating to just using `tokio-postgres` would make sense.

PR #779, opened 2022-09-21
use `tokio-postgres` in `outbound-pg`

dicej:
Since `postgres` just wraps the async `tokio-postgres` library in a synchronous wrapper, and `outbound-pg` exports async methods, it makes more sense to use `tokio-postgres` directly.

Addresses #776 

### DIFF
--- a/crates/outbound-pg/src/lib.rs
+++ b/crates/outbound-pg/src/lib.rs
@@ -1,6 +1,10 @@
 use anyhow::anyhow;
 use outbound_pg::*;
-use postgres::{types::ToSql, types::Type, Client, NoTls, Row};
+use tokio_postgres::{
+    tls::NoTlsStream,
+    types::{ToSql, Type},
+    Connection, NoTls, Row, Socket,
+};
 
 pub use outbound_pg::add_to_linker;
 use spin_engine::{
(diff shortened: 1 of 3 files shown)


lann (approved):
Well that wasn't so bad. I don't really get the "separate connection Future you have to spawn" design. :shrug:

radu-matei (approved):

crates/outbound-pg/src/lib.rs
```diff
@@ -41,23 +45,24 @@ impl outbound_pg::OutboundPg for OutboundPg {
...
+        let params: Vec<&(dyn ToSql + Sync)> = params
+            .iter()
+            .map(to_sql_parameter)
+            .collect::<anyhow::Result<Vec<_>>>()
+            .map_err(|e| PgError::QueryFailed(format!("{:?}", e)))?;
+
+        let nrow = client
+            .execute(statement, params.as_slice())
+            .await
+            .map_err(|e| PgError::ValueConversionFailed(format!("{:?}", e)))?;
```
itowlson, line 63: This might be an error in the original code - this seems more like it should be `QueryFailed` than `ValueConversionFailed`.

crates/outbound-pg/src/lib.rs
```diff
@@ -238,3 +245,11 @@ fn convert_entry(row: &Row, index: usize) -> Result<DbValue, postgres::Error> {
     };
     Ok(value)
 }
+
+fn spawn(connection: Connection<Socket, NoTlsStream>) {
+    tokio::spawn(async move {
+        if let Err(e) = connection.await {
```
kate-goldenring, line 251: is there a reason that we dont want to bubble up this error?

itowlson (approved):
Spotted one thing that looks iffy but it was iffy before, not anything introduced in this PR - not fussed if you want to leave it.

crates/outbound-pg/src/lib.rs
dicej: We can't bubble it up since `tokio::spawn` is the moral equivalent of spawning a thread -- it will continue to run after our `spawn` function returns.  And the future that `tokio::spawn` accepts returns `()`, so it can't return a `Result`.  Technically we could panic, but that would just leave Tokio to deal with the panic.

crates/outbound-pg/src/lib.rs
lann: Yeah, in fact it looks like this error is swapped with the one right above at line 58...

crates/outbound-pg/src/lib.rs
kate-goldenring: Could we store a handle to the futures and then await them both?
```
let spawn_handle = spawn_handle(connection);
let client_handle = client
            .execute(statement, params.as_slice());
let nrow, _ = futures::future::try_join_all(vec![spawn_handle, client_handle]).await?;
```

crates/outbound-pg/src/lib.rs
dicej: Swapped them: https://github.com/fermyon/spin/pull/779/commits/bdbb3eed6d6d882d2a7e8585398dc625e1cbafb7

crates/outbound-pg/src/lib.rs
itowlson: To an extent, do we care?  If the connection has failed, the user will get the error from the database request anyway, won't they?

crates/outbound-pg/src/lib.rs
itowlson: Yikes, good catch @lann, and sorry folks, this has been wrong since the year dot.  Thanks for fixing @dicej

crates/outbound-pg/src/lib.rs
kate-goldenring: true. I'm fine leaving it as. More of a question of whether we want this error to error out the application or not.

dicej: Yes, based on prior experience with this I can confirm that any connection error that interrupts sending queries or statements or receiving results will throw an error at that level, so no need to process both errors.

kate-goldenring (approved):

PR #780, opened 2022-09-21
feat: add support to override Spin compatibility checks for plugins

kate-goldenring:
Spin plugins state which versions of Spin they are compatible with in their plugin manifest. Currently, Spin checks if the plugin is compatible before allowing it to be installed or executed. If incompatible, the operation fails. This allows overriding that failure with an `--override-compatibility-check` flag. When specified during Spin plugin installs, upgrades and executions, the action will continue even if the plugin is incompatible with the current version of Spin.

More context on the decision to add this override can be found in this thread: https://github.com/fermyon/spin/pull/735#issuecomment-1246356616

### DIFF
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -88,23 +88,32 @@ impl Architecture {
 }
 
 /// Checks whether the plugin supports the currently running version of Spin.
-pub fn check_supported_version(manifest: &PluginManifest, spin_version: &str) -> Result<()> {
+pub fn check_supported_version(
+    manifest: &PluginManifest,
+    spin_version: &str,
+    override_compatibility_check: bool,
+) -> Result<()> {
     let supported_on = &manifest.spin_compatibility;
-    inner_check_supported_version(supported_on, spin_version)
+    inner_check_supported_version(supported_on, spin_version, override_compatibility_check)
 }
 
-fn inner_check_supported_version(supported_on: &str, spin_version: &str) -> Result<()> {
+fn inner_check_supported_version(
+    supported_on: &str,
+    spin_version: &str,
+    override_compatibility_check: bool,
+) -> Result<()> {
     let comparator = VersionReq::parse(supported_on).with_context(|| {
-        format!(
-            "Could not parse manifest compatibility version {} as valid semver",
-            &supported_on,
-        )
+        format!("Could not parse manifest compatibility version {supported_on} as valid semver")
     })?;
     let version = Version::parse(spin_version)?;
     if !comparator.matches(&version) {
-        return Err(anyhow!(
-            "Spin version compatibility check failed (supported: {supported_on}, actual: {spin_version}). Try running `spin plugin update` to get latest."
+        if override_compatibility_check {
+            eprintln!("Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Check overridden ... continuing to install or execute plugin.");
+        } else {
+            return Err(anyhow!(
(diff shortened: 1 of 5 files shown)


crates/plugins/src/manifest.rs
```diff
@@ -88,23 +88,32 @@ impl Architecture {
...
+    supported_on: &str,
+    spin_version: &str,
+    override_compatibility_check: bool,
+) -> Result<()> {
     let comparator = VersionReq::parse(supported_on).with_context(|| {
-        format!(
-            "Could not parse manifest compatibility version {} as valid semver",
-            &supported_on,
-        )
+        format!("Could not parse manifest compatibility version {supported_on} as valid semver")
```
itowlson, line 106: Suggestion, though probably an irrelevant one... if `override_compatibility_check` is set, a parse error should _not_ be fatal (the user has already said they don't care about versions).  That said, it's hard to imagine how any plugin with a broken version spec gets this far!

crates/plugins/src/manifest.rs
```diff
@@ -88,23 +88,32 @@ impl Architecture {
...
     })?;
     let version = Version::parse(spin_version)?;
     if !comparator.matches(&version) {
-        return Err(anyhow!(
-            "Spin version compatibility check failed (supported: {supported_on}, actual: {spin_version}). Try running `spin plugin update` to get latest."
+        if override_compatibility_check {
+            println!("Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Check overridden ... continuing to install or execute plugin.");
```
itowlson, line 111: Suggest printing to stderr instead of stdout, particularly because we don't know if something (a script, or maybe Spin itself if we one day use stdin-stdout as an interface to plugins) is consuming stdout. 
```suggestion
            eprintln!("Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Check overridden ... continuing to install or execute plugin.");
```

src/commands/plugins.rs
```diff
@@ -104,7 +108,14 @@ impl Install {
         // Downgrades are only allowed via the `upgrade` subcommand
         let downgrade = false;
         let manifest = manager.get_manifest(&manifest_location).await?;
-        try_install(&manifest, &manager, self.yes_to_all, downgrade).await?;
+        try_install(
+            &manifest,
+            &manager,
+            self.yes_to_all,
```
itowlson, line 114: Wondering if the time has come to put the yes, OCC and downgrade flags into an InstallOptions type... (definitely not a must-do at this point though!)

src/commands/plugins.rs
```diff
@@ -176,6 +187,10 @@ pub struct Upgrade {
     #[clap(short = 'y', long = "yes", takes_value = false)]
     pub yes_to_all: bool,
 
+    /// Overrides a failed compatibility check of the plugin with the current version of Spin.
+    #[clap(long = "override-compatibility-check", takes_value = false)]
```
itowlson, line 191: In the Install options you were using a const for this.  (I don't have strong feelings either way, just suggesting consistency, unless of course there is a reason not to.)

itowlson (approved):
Sorry for letting this slip through the cracks.  I had a few suggestions, but the only one I feel strongly about is printing the override message to stderr instead of stdout, and even that isn't a blocker.  Otherwise it looks great - thank you!

src/commands/plugins.rs
kate-goldenring: Good point -- this would ensure consistency across subcommands. Let's save this for another PR.

itowlson (approved):
LGTM!

karthik2804 (approved):

PR #781, opened 2022-09-22
Add option to use --follow-all when running Redis Trigger (to see messages interactively, as apposed to via logs)

tpmccallum:
Signed-off-by: tpmccallum <tim.mccallum@fermyon.com>

### DIFF
--- a/docs/content/rust-components.md
+++ b/docs/content/rust-components.md
@@ -194,8 +194,15 @@ message on the `messages` channel, the `echo-message` component will be executed
 $ redis-server --port 6379
 # then, start the Spin application
 $ spin up --file spin.toml
+# the logs will output the following
 INFO spin_redis_engine: Connecting to Redis server at redis://localhost:6379
 INFO spin_redis_engine: Subscribed component 0 (echo-message) to channel: messages
+```
+
+**Please note**: By default, `stdout` goes only to the logs. This is because Spin is intended to run as a server. If you would like to see the published messages interactively please use the `--follow-all` option, when starting the spin application. For example:
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/content/rust-components.md
```diff
@@ -194,10 +194,17 @@ message on the `messages` channel, the `echo-message` component will be executed
 $ redis-server --port 6379
 # then, start the Spin application
 $ spin up --file spin.toml
+# the logs will output the following
 INFO spin_redis_engine: Connecting to Redis server at redis://localhost:6379
 INFO spin_redis_engine: Subscribed component 0 (echo-message) to channel: messages
 ```
 
+**Please note**: By default, `stdout` goes only to the logs. This is because Spin is intended to run as a server. If you would like to see the published messages interactively please use the `--follow-all` option, when starting the spin application. For example:
```
itowlson, line 202: I wonder if this should be moved down to be with the "Spin will instantiate..." line and the logging of the `println!`.  It feels a bit out of place up here with the Spin system logs themselves.

E.g. change line 215 to "Spin will instantiate and execute the component we just built, which will emit the `println!` message to the application log file:"  Then after line 220 something like "If you'd also like to see the `println!` messages echoed to the console as they happen... (stuff about `--follow-all`)"...?

itowlson (changes requested):

docs/content/rust-components.md
tpmccallum: Thanks Ivan, I have added an additional commit which implements those suggestions.
Reads well now.

itowlson (approved):
LGTM

PR #785, opened 2022-09-23
Swap wget for curl to be consistent with docs

ipedrazas:
The [quickstart](https://spin.fermyon.dev/quickstart/) has install instructions using `curl`, however, the install script is using `wget`.

Quickstart:
```
curl https://spin.fermyon.dev/downloads/install.sh | bash
```

Install script:
```
# Download file, exit if not found - e.g. version does not exist
fancy_print 0 "Step 1: Downloading: ${URL}"
wget -q $URL || (fancy_print 1 "The requested file does not exist: ${FILE}"; exit 1)
fancy_print 0 "Done...\n"
```

This PR replaces the `wget` instructions for the equivalent `curl` ones, to keep the tool used in the documentation and in the script consistent.

### DIFF
--- a/docs/downloads/install.sh
+++ b/docs/downloads/install.sh
@@ -66,7 +66,7 @@ while [[ $# -gt 0 ]]; do
 done
 
 # Check all required utilities are available
-require wget
+require curl
 require tar
 require uname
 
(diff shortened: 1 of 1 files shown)


docs/downloads/install.sh
```diff
@@ -115,7 +115,7 @@ URL="https://github.com/fermyon/spin/releases/download/${VERSION}/${FILE}"
 
 # Download file, exit if not found - e.g. version does not exist
 fancy_print 0 "Step 1: Downloading: ${URL}"
-wget -q $URL || (fancy_print 1 "The requested file does not exist: ${FILE}"; exit 1)
+curl -sOL $URL || (fancy_print 1 "The requested file does not exist: ${FILE}"; exit 1)
```
mikkelhegn, line 118: This doesn't capture the "file does not exists" scenario. I'm getting an invalid archive downloaded.

E.g., `./install.sh -v rubbish`

I think you would have to add `--fail` to curl to have it return a non-zero exit code.

mikkelhegn (changes requested):

docs/downloads/install.sh
ipedrazas: You're right :) added the `fail` flag

mikkelhegn (approved):

mikkelhegn:
Cool - thx!

PR #788, opened 2022-09-24
build: rely on bash from 'PATH'

endocrimes:
/bin/bash isn't guaranteed to exist on any platform (unlike /bin/sh), here we switch to relying on bash from the path instead, this lets us build on systems like NixOS without breaking other systems.

Signed-off-by: Danielle Lancashire <dani@builds.terrible.systems>

---

Tested on NixOS (amd64) and macOS Monterey (arm)


radu-matei (approved):
Thank you for catching this!

LGTM

Issue #667, opened 2022-07-30
Reuse connection in outbound PosgreSQL implementation


radu-matei:
The outbound Redis implementation currently creates a new connection for each operation: https://github.com/fermyon/spin/blob/360ad620543baa26bf96f76c4ed596894fb73024/crates/outbound-redis/src/lib.rs#L45-L57

Ideally, when using the same connection multiple times in the same module, the underlying connection would be reused. This means (I suspect) introducing a resource field for the connection, and maintaining that as state in the host component.

A guest module would be able to connect to two different servers by creating two different connections.

A similar improvement could be made for the outbound PostgreSQL implementation.

radu-matei:
The outbound Redis implementation has been patched thanks to @FrankYang0529.

Updating the issue title to reflect that the PostgreSQL implementation is the one needing a patch to reuse connections.

radu-matei:
The outbound Redis implementation has been patched thanks to @FrankYang0529.

Updating the issue title to reflect that the PostgreSQL implementation is the one needing a patch to reuse connections.

etehtsea:
This one can be closed I think.

PR #791, opened 2022-09-25
feat(outbound-pg): Reuse connection during request lifecycle

etehtsea:
Also, pg_backend_pid endpoint into outbound-pg example.

Before the fix:

```
$ curl -i localhost:3000/pg_backend_pid
HTTP/1.1 500 Internal Server Error
content-length: 0
date: Tue, 27 Sep 2022 14:03:50 GMT

thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `592913`,
  right: `592914`', src/lib.rs:112:5
```

After the fix:

```
$ curl -i localhost:3000/pg_backend_pid
HTTP/1.1 200 OK
content-length: 23
date: Tue, 27 Sep 2022 14:07:14 GMT

pg_backend_pid: 595194
```

Ideally, this fix has to be covered by an integration test rather than
manual testing through examples, but the testing environment has to be
set up first.

Refs: https://github.com/fermyon/spin/issues/667.


itowlson (approved):
Thanks @etehtsea, looks good to me!  Really appreciate the extra info and polish in the sample, and nice to have the connection reuse.  Are you still pushing updates or is this okay to merge?

etehtsea:
@itowlson yep, it should be ready to go. The only thing I was thinking about is the difference with redis connections map implementation - https://github.com/fermyon/spin/blob/main/crates/outbound-redis/src/lib.rs#L19-L21

My understanding is that as the state is built per request then it's single-threaded. Am I missing something?

lann:
> My understanding is that as the state is built per request then it's single-threaded. Am I missing something?

The difference with the outbound-redis example is that it is reusing connections _between_ requests. Since this PR only caches within a single request it does not need to lock; individual requests are handled on a single thread.

itowlson:
@etehtsea Given Lann's edit to his comment, it looks like the pattern we should be adopting is the one you already have, so I'll merge this as it stands.  Thanks for the contribution and for taking the time to cross-check!

PR #794, opened 2022-09-29
deploy to the Cloud

bacongobbler:
This feature enhancement allows Spin users to deploy their applications to the Cloud.

```
><> cd examples/http-rust

><> spin build
Executing the build command for component hello: cargo build --target wasm32-wasi --release
    Finished release [optimized] target(s) in 0.01s
Successfully ran the build command for the Spin components.

><> spin deploy --hippo-server ***************
Open the Cloud's device authorization URL in your browser: ***********/device-authorization and enter the code: ******
The code has been copied to your clipboard for convenience.
Waiting for device authorization...
Device authorized!
Deployed spin-hello-world version 1.0.0
Application is running at spin-hello-world-61fr9ys5.*************

><> curl https://spin-hello-world-61fr9ys5.*************/hello
Hello, Fermyon!
```

A couple of open comments/questions for the reviewer:

1. Because of the reliance on `copypasta`, Spin developers will need to install `xorg-dev` to compile against Linux. Where should we document this? Docs? README?
2. On every invocation to `spin deploy`, the command will prompt for a new device code. This is because `spin deploy` does not store any configuration. Is this something we should track, or is the current behaviour okay?
3. There's a bunch of code duplication - specifically surrounding `crates/cloud/src/registry/bindle` and `crates/publish`. This is an area worth looking at a little closer. (cc @radu-matei)
4. All of the `spin deploy` flags are prefixed with `--hippo-*`. Is this worth re-naming these? (he says, as he asks the most difficult issue in the engineering world... Naming things)

TODO:

- [x] Have `spin login --status` print as JSON
- [x] Implement non-interactive login flow


itowlson:
I thought we were going to do something like `spin login` for environments with persistent login tokens.  Then `spin deploy` would gain a `--token` option to go alongside `--hippo-password`.

I feel like we could do with a SIP to figure out things like this and the clipboard overwrite issue and the persistence model and all that... what do you reckon?

bacongobbler:
```
./target/release/spin login --status
You are logged into http://localhost:5301
Your session will expire on 2022-10-12 19:31:21 +00:00.
```

bacongobbler:
Okay. Everything except for shunting back to `spin login` when the file is not found or the token has expired has been implemented. Please feel free to give this a test.

itowlson:
This is really nice and super smooth!

One niggle I spotted.  In the GH auth scenario we proxy Bindle; if this is far away then you can get this:

```
ivan@hecate:~/testing/deploytest$ ~/github/spin/target/debug/spin deploy
<URL> is responding slowly or not responding...

<URL>/api/registry is responding slowly or not responding...
```

Could we spot the proxy case and emit only one sloth warning?  Sorry to nitpick.

itowlson:
Not sure if the PR is ready for feedback at this level, but the error mode if the Hippo server doesn't exist is that it does nothing for about 15 (?) seconds then prints `Error: expected value at line 1 column 1`.

(This was hitting a URL that had previously existed as a AWS Hippo instance, but had been terraform destroyed.  Hitting a completely made-up domain gives the same message but a lot quicker!)

bacongobbler:
To reproduce that issue, did you stand up a cluster, run `spin login`, shut down the cluster, then ran `spin deploy`?

itowlson:
My specific scenario - and the weird and dev-specific way it happened may mean it's not a real issue - was:

* Build this PR
* `spin login` using GH auth and do a deployment
* Switch back to Spin main
* Create an AWS cluster using the installer, export the environment variables (all as per installer instructions), do a deployment
* Destroy the AWS cluster (`terraform destroy`)
* Switch back to this PR
* `spin login` which this time picked up the dangling Hippo environment variables

itowlson:
To be clear, it was that last _login_ which gave the error, not any deploy.

bacongobbler:
I managed to reproduce https://github.com/fermyon/spin/pull/794#issuecomment-1270643405:

```
><> env | grep HIPPO
HIPPO_PASSWORD=bar
HIPPO_USERNAME=foo
HIPPO_URL=http://localhost:5309
><> env | grep BINDLE
BINDLE_URL=http://localhost:8080/v1
><> spin login
What authentication method does this server support?

1. Sign in with GitHub
2. Sign in with a username and password

Enter a number: 2
Error: expected value at line 1 column 1
```

The actual error is this:

```
error in reqwest: error sending request for url (http://localhost:5309/api/auth-tokens): error trying to connect: tcp connect error: Connection refused (os error 111)
```

This is because the call to `format_login_error` on line 182 does not account for tcp connection errors. It always assumes the input error string will be JSON.

This code was present on main as well... I'll see about fixing the error message.

src/commands/login.rs
```diff
@@ -0,0 +1,400 @@
...
+    let timeout_seconds = 15 * 60;
+
+    // Loop while waiting for the device code to be authorized by the user
+    loop {
+        if seconds_elapsed > timeout_seconds {
+            bail!("Timed out waiting to authorize the device. Please execute `spin login` again and authorize the device with GitHub.");
+        }
+
+        match client.login(device_code.device_code.clone().unwrap()).await {
+            Ok(response) => {
```
radu-matei, line 291: This needs to check whether `response.token` is `None` and continue to wait if so.

src/commands/login.rs
```diff
@@ -0,0 +1,400 @@
...
+
+        let login_connection: LoginConnection;
+        let mut url = DEFAULT_CLOUD_URL.to_owned();
+        let mut auth_method = AuthMethod::Github;
+
+        if let Some(u) = self.hippo_server_url {
+            url = u;
+            // prompt the user for the authentication method
+            // TODO: implement a server "feature" check that tells us what authentication methods it supports
+            auth_method = prompt_for_auth_method();
```
itowlson, line 151: Sorry to be a broken record but there needs to be a way to bypass this for programmatic environments (e.g. so they can ask via a GUI and pass the setting down to `spin login`) please?

Also a couple of issues that you may already have in hand:

1. it's currently being shown even when `--get-device-code` or `--check-device-code` is present
2. currently if you provide a Hippo URL you're still required to provide username, password and Bindle server, even though you may be planning to answer "GitHub" at the prompt

src/commands/deploy.rs
```diff
@@ -238,15 +238,121 @@ impl DeployCommand {
...
+                    CloudChannelRevisionSelectionStrategy::UseRangeRule,
+                    range_rule,
+                    None,
+                )
+                .await
+                .context("Problem creating a channel")?
+            }
+        };
+
+        println!(
```
itowlson, line 343: The last rebase introduced a call to `wait_for_ready` on the Hippo flow - flagging this up for the Cloud flow too.  (Sorry, I should have looped you in on the PR where that was introduced.)

crates/cloud/src/client.rs
```diff
@@ -196,25 +179,20 @@ impl Client {
...
         range_rule: Option<String>,
         active_revision_id: Option<Uuid>,
-        certificate_id: Option<Uuid>,
         environment_variables: Option<Vec<UpdateEnvironmentVariableDto>>,
     ) -> anyhow::Result<()> {
         let command = PatchChannelCommand {
             channel_id: Some(id),
             name: name.map(|n| Box::new(StringField { value: Some(n) })),
-            domain: domain.map(|d| Box::new(StringField { value: Some(d) })),
             revision_selection_strategy: revision_selection_strategy
```
gogeterobert, line 190: Does ChannelRevisionSelectionStrategyField still exist on the new specification of the could swagger file (asking this because when I run locally the revision_selection_strategy is a string, as seen in the picture below)?
image

bacongobbler: Yes

bacongobbler: https://github.com/fermyon/cloud-openapi/blob/350f67d7abf3cab9da83a44ba93ff5bfc2ad8b29/swagger.json#L1351-L1357

bacongobbler: Let me try to re-generate the openapi against main and see if I missed something.

src/commands/login.rs
bacongobbler: resolved:

- set to `AuthMethod::Github` if either `--get-device-code` or `--check-device-code` is set
- set to `AuthMethod::Github` if `--method github`
- set to `AuthMethod::UsernameAndPassword` if `--method username`

src/commands/login.rs
bacongobbler: fixed

crates/cloud/src/client.rs
bacongobbler: I just re-generated the new specification of the cloud swagger file and I received the same file as before.

As demonstrated in https://github.com/fermyon/spin/pull/794#discussion_r992673079, the type is a String, but it is still generating the "ChannelRevisionSelectionStrategyField" as before.

bacongobbler: https://github.com/fermyon/cloud-openapi/commit/350f67d7abf3cab9da83a44ba93ff5bfc2ad8b29 demonstrates what changed between updates.

src/commands/login.rs
```diff
@@ -0,0 +1,424 @@
...
+    )]
+    pub insecure: bool,
+
+    /// URL of hippo server
+    #[clap(
+        name = HIPPO_SERVER_URL_OPT,
+        long = "hippo-server",
+        env = HIPPO_URL_ENV,
+        requires = BINDLE_SERVER_URL_OPT,
+        requires = HIPPO_USERNAME,
```
itowlson, line 77: At the moment if you pass a URL then you have to pass a username and password, even if you plan to use GH auth.

(sorry, thought I'd flagged this but apparently not)

bacongobbler: Good point. This is a holdover from prior iterations. Will change

itowlson (approved):
This is working great for me!  Thanks for battling through all the issues.

PR #795, opened 2022-09-30
Deployment auth SIP

itowlson:
WIP but getting discussion under way!

[Rendered](https://github.com/itowlson/spin/blob/sip-device-flow-auth/docs/content/sips/007-deployment-auth.md)


docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+
+## `spin login` detailed considerations
+
+`spin login` is a new command and a relatively complex user experience.  This section describes the behaviour in more detail.
+
+### Browser interaction
+
+As part of device auth flow, the user needs to go to the verification URL in their browser and paste in the code.
+
+> QUESTION: Should we attempt to open the browser automatically?
```
bacongobbler, line 100: I think we should. Most CLI with OAuth integration use common tools like `xopen` to open the default web browser when the user is asked to open a browser window.

We could provide an opt-out flag, though I'm not sure how useful that would be to users.

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+> * Pro: it's more convenient
+> * Con: it may be unexpected (we have not asked for consent or warned it will happen)
+> * Con: it may work differently across browsers and operating systems, which may require explanation in the documentation
+
+> QUESTION: Should we copy the user code to the clipboard?
```
bacongobbler, line 105: This is a fairly contentious point for me. On one hand, this is incredibly convenient when I want to log in. On the other hand, if I accidentally execute this command, it could wipe my clipboard's content with junk.

Say that I have some important snippet in my clipboard, and I accidentally fat-finger `spin login`. As a result, I've lost the contents of my clipboard.

IMO a `--copy-to-clipboard=false` flag may make sense here.

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+> * Pro: it's convenient
+> * Con: the user may already have stuff on their clipboard
+> * Con: again, documentation needs to consider differences across OSes
+
+(For both of these questions, is there prior art to which we can refer?)
```
bacongobbler, line 110: Commands like `az login` and `gh auth login` would be good places to reference.

https://cli.github.com/manual/gh_auth_login

https://learn.microsoft.com/en-us/cli/azure/authenticate-azure-cli

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+### Verification polling
+
+The device auth flow involves the client polling the server using the device code as the auth token. Each poll can have three outcomes:
+
+* 200 and the response contains a token: user has authorised this device
+* 200 and the response doesn't contain a token: user has not yet authorised this device
+* Server error
+* Network error
+
+We propose to poll for two minutes. If the user has not entered the code by then, `spin login` fails and exits.
```
bacongobbler, line 121: A device code is provided with an expiration date. The expiration date determines how long the user can take to enter the device code until the code expires. The server is the one that controls how long the device code is valid.

We could check the expiration date to determine how long we can poll for. I believe it's currently available for 900 seconds (15 minutes).

https://github.com/fermyon/cloud-openapi/blob/896a21187ff72a1e779f13be47e8a1685833df85/clients/rust/src/models/device_code_item.rs#L23

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Token storage
+
+When the user logs into a server, we need to store:
+
+* The server URL
+* Any server connection options, such as the `--insecure` flag
+* The token
+
+A simple solution is to store these in a file under a non-roaming directory. This is not fantastically secure for tokens, but may be enough for the preview.
```
bacongobbler, line 135: what does "non-roaming" mean?

bacongobbler: > This is not fantastically secure for tokens

Yeah. This is something we need to fix on the server side via refresh tokens.

In the current system, we only issue an auth token. If an attacker manages to get a hold of that auth token, they can use it for as long as it's valid (currently 7 days from date of issue). Even if you change your password. That’s because the server will trust the token if its signature is valid. The only way to invalidate the token is to change the key that was used to sign it, and that has the consequence of invalidating everyone else’s tokens.

With refresh tokens, you can send both the auth token as well as the refresh token to fetch a fresh auth token. This means the user will not need to re-enter their username/password or re-enter a device code when it expires, and the server can shorten the time window that the auth token is valid down to a more reasonable limit (generally between 1-8 hours). If the auth token is compromised, a server administrator can invalidate the refresh token, forcing the user to log back in. If a user does not use the app within the time limit of the refresh token (usually between 7 days), then they are forced to log back in.

bacongobbler (changes requested):
Looking great! Thanks for taking the time to write this up.

docs/content/sips/007-deployment-auth.md
bacongobbler: A good crash course on auth vs refresh tokens can be found here: https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+
+> QUESTION: If a user is logged into multiple servers, should we:
+> * Default the URL to the _most recent_ login
+> * Provide a way for the user to say which URL is their default
+> * Require the user to provide a URL
+> * Provide an abstraction (e.g. name, config file path) by which the user can switch between servers
+
+> QUESTION: The last section represents a change in behaviour when no flags are provided. Are we happy with this, or should we direct people to `spin login` and not attempt it automatically?
+
+> QUESTION: Is it correct to reject a username/password if the user is logged in to the chosen server, or should we attempt username/password login in this case?
```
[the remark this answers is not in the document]
itowlson: @bacongobbler Can you clarify the comment "The first time the user runs spin deploy, they will be logged in" please?  It sounds like this is postulating:

1. There will be a concept of "logging into" a username-password server (as well as a device token server)
2. This will happen _implicitly_ if the username-password set of options are provided during `deploy` (rather than controlling the current deployment but not affecting state).  [Token login is an explicit step, even if triggered from `deploy`, and is currently specced not to happen if username-password options are present.]

I am not keen on 2, though we could do it if we ask the user "do you want to remain logged in" or some such wording.

But I agree, if 1 is true then we need to rework this to express that if URL/username/password are present and the URL is _not_ logged in as a token server, then URL/username/password should override and behave as they do now.

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Deploy command changes
+
+The `spin deploy` command will now act as follows:
+
+* The Hippo URL, username and password are now **optional** at the implementation level.
+* If the Hippo URL is specified:
+  * If the user is logged into that server, use the stored token for auth.  The username and password are not required and _must not be present_.
+  * If the user is not logged into that server, and a username or password is provided, follow the current behaviour.  Both username and password are required.
+  * If the user is not logged into that server, and no username or password is provided, attempt a login using the same behaviour as `spin login`.  If that succeeds, continue to deployment.  If it fails, exit.
```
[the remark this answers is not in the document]
itowlson: Again, this sounds like you envisage a logged-in state for username-password servers, as well as token servers.  (This was an outstanding question for me, I'm not disagreeing, I just want to make sure I'm understanding correctly so I can specify it.)

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+> * Default the URL to the _most recent_ login
+> * Provide a way for the user to say which URL is their default
+> * Require the user to provide a URL
+> * Provide an abstraction (e.g. name, config file path) by which the user can switch between servers
+
+> QUESTION: The last section represents a change in behaviour when no flags are provided. Are we happy with this, or should we direct people to `spin login` and not attempt it automatically?
+
+> QUESTION: Is it correct to reject a username/password if the user is logged in to the chosen server, or should we attempt username/password login in this case?
+
+> QUESTION: What happens if the user is logged in to one server and `$HIPPO_URL` is set to another?  What happens if the user is logged in but the `$HIPPO_USERNAME` is set?  (Thinking here of situations where the user is switching between a username/password server and a device-auth one.)  Should we provide a way to "log into" username/password servers, and use that stored state instead of the environment variables?
```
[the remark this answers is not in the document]
itowlson: Aha... your second remark makes me suspect that you have a mental model of how `HIPPO_USERNAME` and `HIPPO_PASSWORD` map onto token servers, which would explain some of your other remarks that I hadn't quite followed.  (Whereas I have been treating the username-password UI as largely distinct from the token UI.)  Let's maybe have a chat and see if we can capture and document that understanding?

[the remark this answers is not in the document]
itowlson: Re your first comment, I think you maybe have a mental model where any deployment is a login?  Like if I deploy to server A, that logs me into server A; then if I set `HIPPO_URL` and deploy to server B, that logs me into server B?  There is no (or less) sense of command options applying just to the current deploy command?  Have I understood this right?  Again, happy to chat and get a sense for how you see this.

docs/content/sips/007-deployment-auth.md
itowlson: If copy to clipboard is opt _out_, then fat-fingering will still wipe your clipboard.  Only if it's opt _in_ will users know they're about to overwrite their current clip.  And that greatly reduces the convenience because you have to write a whole extra option.

A possible approach would be to have the "please open your browser to..." be interactive, e.g.:

```
Open the Platform device authorization URL in your browser: https://example.com/device-authorization and enter the code: tWOJCFhY
[c] Copy code to clipboard
[b] Open browser
[r] Ready
# user hits "c"
Copied
[b] Open browser
[r] Ready
# user does the dance then hits "r"
Device authorised!
Let the revels commence
```

But that feels like a very heavyweight solution.

On the positive side, we can at least all agree that it's "a contentious issue."

docs/content/sips/007-deployment-auth.md
itowlson: "Non-roaming" means it lives only on this machine and is not synced to other machines.  (As I assume a "device code" works only on the machine it came from.)  E.g. on Windows the %USERPROFILE%/AppData/Roaming directory follows the user around but %USERPROFILE%/AppData/Local is specific to the user's account _on that machine_.  Not sure of analogues on other OSes.

docs/content/sips/007-deployment-auth.md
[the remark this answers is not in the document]
itowlson: To clarify, my mental model was that:

1. Command line options (and hence EVs) affected _that one execution of the command_ and overrode defaults
2. Only an explicit login step (whether via `spin login` or prompted from `spin deploy`) would set new defaults

(This is not to argue for either model... I just realised I hadn't explicitly written mine down!)

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+Login successful.
+```
+
+The URL and token generated by the device authorisation flow is stored on the machine.
+
+A user may log into multiple Hippo servers.  The (URL, token) pairs for each will be stored.  That is, logging into one server does not log you out of others.
+
+> QUESTION: Or should you only be logged into one at a time?
+
+> QUESTION: Should we provide an explicit `logout` command to destroy the token?
```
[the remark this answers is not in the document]
itowlson: Between this and the "only logged into one system at a time" comment, I worry about creating friction for people who are switching between multiple deployment environments (likely including our own developer relations team _grin_).

Probably the SIP needs to explicitly identify concepts of "the default deployment target" and "deployment targets for which you have saved credentials."  So if I have, say, a Spin 0.6 install and a Spin canary install, I can work in one or the other [change which one if default], without needing to re-log-in [re-establish credentials] when I switch.

Do those sound like clarifying notions?

docs/content/sips/007-deployment-auth.md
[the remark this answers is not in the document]
lann: I think if credentials are explicitly set via CLI arg or env var, stored credentials should be ignored. It could make sense to include the username with stored credentials and warn a user if they are overriding that saved username's credential, i.e. if I login as `lann` and later `HIPPO_USERNAME=lann HIPPO_PASSWORD=... spin deploy`, I might get a message `You are setting a password for already logged-in user "lann".` to help contextualize any auth errors.

I would expect credentials to only be saved if 1) they are entered interactively or 2) they are explicitly passed to `spin login`.

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+
+> QUESTION: What happens if the user is logged in to one server and `$HIPPO_URL` is set to another?  What happens if the user is logged in but the `$HIPPO_USERNAME` is set?  (Thinking here of situations where the user is switching between a username/password server and a device-auth one.)  Should we provide a way to "log into" username/password servers, and use that stored state instead of the environment variables?
+
+## Error flows
+
+### Auth failure on deploy
+
+Deployment would fail if the token was incorrect or expired.
+
+> QUESTION: In this case should we simply fail, or should we attempt login?  _Note:_ if we decide to attempt login we _must_ provide a flag to suppress this (for programmatic consumers who have their own login flows).
```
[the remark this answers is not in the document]
lann: > we _must_ provide a flag to suppress this

It's good to have the flag for edge cases, but you can generally detect this with e.g. [`atty`](https://docs.rs/atty/latest/atty/)

docs/content/sips/007-deployment-auth.md
lann: `gh` flow for ref:

```
? How would you like to authenticate GitHub CLI? Login with a web browser

! First copy your one-time code: AE4A-1435
Press Enter to open github.com in your browser... 
```

docs/content/sips/007-deployment-auth.md
lann: > Not sure of analogues on other OSes.

https://docs.rs/dirs/4.0.0/dirs/fn.data_local_dir.html

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+
+When the user logs into a server, we need to store:
+
+* The server URL
+* Any server connection options, such as the `--insecure` flag
+* The token
+
+A simple solution is to store these in a file under a non-roaming directory. This is not fantastically secure for tokens, but may be enough for the preview.
+
+> QUESTION: What priority should we assign to storing tokens securely?  (E.g. DPAPI on Windows, not sure of the equivalent across other OSes.)
```
lann, line 137: https://docs.rs/keyring/1.2.0/keyring/

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+
+> QUESTION: Should we provide a way for the user to control where things are stored, opening up the possibility of multiple configurations? (This also provides an approach to the default/override URL question on deploy.)
```
lann, line 139: This seems like the simplest way to allow for multiple logins for "power users".

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,158 @@
...
+$ spin login --url foo --check-device-code bar
+{ "status" : "waiting" }
+
+# user enters code into browser
+
+$ spin login --url foo --check-device-code bar
+{ "status" : "authorized", "token" : "..." }
+```
+
+> QUESTION: Should we present the token to the caller for it to store wherever it finds convenient, or store it ourselves in the same way as interactive login?  If the former, we will need a way for a caller to pass it to `spin deploy`, either via a new flag or by overloading `--hippo-password`.
```
lann, line 158: If we allow the credential store path to be overridden as above, the caller can point it at a temp location and read out from there.

docs/content/sips/007-deployment-auth.md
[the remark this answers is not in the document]
bacongobbler: I had considered the point of view where `spin deploy` does not accept "login" feature flags at all. That is to say, `HIPPO_USERNAME`, `HIPPO_PASSWORD`, etc. are all localized to `spin login`.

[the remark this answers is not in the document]
bacongobbler: Sorry, had to jump in a meeting so I had to leave mid-thought.

> There will be a concept of "logging into" a username-password server (as well as a device token server)

Yes.

> This will happen implicitly if the username-password set of options are provided during `deploy`

Not entirely. I expect the user to run `spin login` prior to running `spin deploy`. All login-related feature flags like `--hippo-username` will be removed from `spin deploy` and moved to `spin login`.

My mental model looks like this:

1. An explicit login step (via `spin login`) is required. Envvars like `HIPPO_USERNAME` are read during `spin login`.
2. Follow-up commands that require login session information (like `spin deploy`) will read session information that is stored in the keyring.
3. Commands like `spin deploy` do NOT have an option to pass a username/password/login URL.

This mental model maps closely with `docker`:

1. `docker login`: log into the server via username/password, storing session information into the keyring
2. `docker push`: pull session information from the keyring; push docker image to the server
3. `docker pull`: pull session information from the keyring; pull docker image from the server
4. `docker logout`: remove session information from the keyring.

This keeps commands like `spin deploy` focused on feature flags related to deploying the application. Login-related feature flags are strictly handled by `spin login` and ONLY `spin login`.

Here's what `docker push --help` looks like:

```
><> docker push --help

Usage:  docker push [OPTIONS] NAME[:TAG]

Push an image or a repository to a registry

Options:
  -a, --all-tags                Push all tagged images in the repository
      --disable-content-trust   Skip image signing (default true)
  -q, --quiet                   Suppress verbose output
```

Notice how there's nothing related to logging into the registry,

docs/content/sips/007-deployment-auth.md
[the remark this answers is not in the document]
bacongobbler: > Command line options (and hence EVs) affected that one execution of the command and overrode defaults

Following up from https://github.com/fermyon/spin/pull/795#discussion_r986028506. If we were to match that mental model, then there would not be a way to have these options affect a single execution because they would not exist as part of `spin deploy`.

I would argue removing the flags and moving them to `spin login`, forcing the user to log in, is the better approach. Consider the current implementation of `spin deploy` in #794:

```
$ export HIPPO_URL=https://***************
$ spin deploy
$ vim src/lib.rs
$ vim spin.toml
$ spin deploy
$ sleep 30d # wait until the session token expires
$ spin deploy
```

If the server implements Github OAuth sign-in, and EVs affect the single execution of the command, the user will be forced to enter a device code three times - twice in short succession of each other, and once more a week later. I feel this would be poor/clunky design from a user's perspective.

My mental model looks more like this:

```
$ export HIPPO_URL=https://***************
$ spin login
$ spin deploy
$ vim src/lib.rs
$ vim spin.toml
$ spin deploy
$ sleep 30d # wait until the session token expires
$ spin deploy
```

The user will be asked to enter a device code once (`spin login`). After the session token expires, `spin deploy` will check the session token and notice it has expired. It will either error out ("Your session has expired. Please use `spin login` to log in"), or it will invoke `spin login` before continuing with `spin deploy`. Kinda like how `helm upgrade --install` [would invoke `helm install`](https://github.com/helm/helm/blob/45161a5366ca9dd5de2a070336d4bdc191448d6b/cmd/helm/upgrade.go#L95-L126) if it was the first time the user was installing a chart.

docs/content/sips/007-deployment-auth.md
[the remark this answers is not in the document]
lann: Let's say I have a CI/CD pipeline that wants to deploy a staging app, with a auth token in some external secret management like a Github actions secret. I don't _really_ want that secret persisted anywhere on disk. Ideally I could do something like `export HIPPO_TOKEN=...` at the beginning of the script and expect it to just work.

docs/content/sips/007-deployment-auth.md
bacongobbler: I like that flow quite a bit. I'll look and see if we can implement this in time.

docs/content/sips/007-deployment-auth.md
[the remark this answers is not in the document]
bacongobbler: We could make an exception for accepting `HIPPO_TOKEN` as input to `spin deploy`. We've talked about supporting personal access tokens for CI/CD pipeline scenarios where the login flow doesn't make much sense.

That being said, I _really_ want to keep the separation of concerns between "logging in" and "deploying an app" as separate as possible. Accepting a one-off auth token would be OK, but supporting the entire login workflow will be difficult to maintain over time. We convoluted the concerns of "logging in" and "pushing to a server" with [Draft](https://github.com/Azure/draft-classic) and it was an absolute nightmare to maintain as we introduced new features for `draft up`.

itowlson:
Thanks folks for all the feedback.  It sounds like the model we want to espouse is:

* `spin login` both selects the deployment server and records credentials for it.
  * Even for username-password servers.
  * You can only be logged into one server at a time - logging out, or logging into another server, not only deselects it but erases the stored credentials.
  * We need a solution for folks who often switch between multiple deployment servers - one possibility is to have them manage file locations.
* `spin deploy` _always_ deploys to the target specified in the last login (and we remove _all_ deployment server options).
  * Possible exception for CI environments?
  * If you're not logged in then it either fails entirely, or hands you off to the login flow (from Fisher's example it looks like if you're not logged in it just fails, but if your credentials fail it hands you off to the login flow?).  But in either case there is a clear handoff - we do not entangle auth and deployment.
  * We are willing to add an extra step to the `spin new -> spin build -> spin deploy` pipeline because it allows us to vastly simplify `deploy`.

Is that correct?

docs/content/sips/007-deployment-auth.md
radu-matei: I'm all in favor of dropping the clipboard copying "feature" altogether.

bacongobbler:
> Is that correct?

Yes. Your assessment matches my own mental model.

The last step seems like it is the most impactful to the UX.

> You can only be logged into one server at a time - logging out, or logging into another server, not only deselects it but erases the stored credentials.

In order to minimize impact to the UX, this makes sense for now. But we should circle back on this one ASAP to see if we can add multiple login profile support as this will be a big QoL improvement to the UX

mikkelhegn:
`spin deploy always deploys to the target specified in the last login (and we remove all deployment server options)`
Without a `--target` (or similar) option, I would probably at any given time deploy to the wrong target, so is the endpoint you are signed in to stored somewhere? I'm asking as a prompt (e.g. Starship) could show that info if that's the case.

itowlson:
@bacongobbler Hope this now accurately reflects our discussions!

itowlson:
@bacongobbler one thing not covered here... there probably needs to be a way to tell what you're logged into... I'm reluctant to have a top level command for that though... would something like `spin login --status` make sense?

bacongobbler:
Yeah that makes sense to me!

bacongobbler:
@itowlson see https://github.com/fermyon/spin/pull/794#issuecomment-1268868957 - `spin login --status` has now been implemented.

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,162 @@
...
+`spin login --url https://example.com`
+
+This launches an interactive login experience as follows.
+
+**1. Determine auth type.**
+
+At the moment, Spin doesn't know whether a URL uses device flow or username-password, so `spin login` will prompt.  If we can identify particular URLs or families of URLs whose auth mode is known, those can skip this step.
+
+```
+How would you like to authenticate Spin?  [Username and password|Log in with a web browser]
```
bacongobbler, line 52: This is a good way to handle this issue for the time being.

bacongobbler: This is how it looks as written in #794:

```suggestion
What authentication method does this server support?

1. Sign in with GitHub
2. Sign in with a username and password

Enter a number: 
```

bacongobbler (changes requested):

itowlson:
@bacongobbler thanks!  Updated the sign-in method prompt; I also added a note about needing a flag to suppress it (which can be hidden).

docs/content/sips/007-deployment-auth.md
```diff
@@ -0,0 +1,168 @@
...
+At the moment, Spin doesn't know whether a URL uses device flow or username-password, so `spin login` will prompt.  If we can identify particular URLs or families of URLs whose auth mode is known, those can skip this step.
+
+```
+What authentication method does this server support?
+1. Sign in with GitHub
+2. Sign in with a username and password
+Enter a number:
+```
+
+Programmatic consumers need a way to override this prompt, e.g. `--method=[username|github]`.
```
bacongobbler, line 58: I think this would be a good option. We can also check for the presence of `--check-device-code` to determine the auth mode, since the presence of that flag automatically short-circuits it to be Github auth.

bacongobbler (approved):

docs/content/sips/007-deployment-auth.md
itowlson: Yep, and `--get-device-code` too of course.  I'll add this note and merge.  Thanks!

PR #796, opened 2022-09-30
Teach outbound-http to use connection pool

lann:
`reqwest::Client` manages an internal connection pool. For executions that make multiple requests to the same host we can take advantage of that pool.


dicej (approved):

radu-matei (approved):
This is excellent, thank you!

LGTM

PR #800, opened 2022-10-02
Fix redis/http triggers

### LINKED ISSUE #799: Redis triggered template for rust fails
Using the canary build version `spin 0.5.0 (dfb3a90 2022-09-29)` and the template for redis-rust results in the following error on `spin up`:

```
🐟 > spin up
Error: Failed to configure Redis trigger

Caused by:
    metadata error: missing required "redis_address"
Error: exit status: 1
```

etehtsea:
[chore: Drop trigger-new leftovers](https://github.com/fermyon/spin/commit/ac4dcbadc1d63b0116fe5b55f529b719f27b1ba4)

If this is intentional, I'll drop this commit.

[fix(spin-redis-engine): Fix error during startup](https://github.com/fermyon/spin/commit/fdf446686d9d51982e37a1933c7c89ffb239242d)

Steps to reproduce:

```shell
spin main $ cargo build
$ cd ..
$ spin/target/debug/spin new redis-rust spin-redis-rust-example
Project description: 
Redis address: redis://localhost:6379
Redis channel: messages
$ cd spin-redis-rust-example
spin-redis-rust-example $ ../spin/target/debug/spin build --up --follow-all
Executing the build command for component spin-redis-rust-example: cargo build --target wasm32-wasi --release
Successfully ran the build command for the Spin components.
Error: Failed to configure Redis trigger

Caused by:
    metadata error: missing required "redis_address"
Error: exit status: 1
```
Closes #799.

[chore(spin-http): Fix base mapping](https://github.com/fermyon/spin/commit/30f0893462a0df9f085fed90b401c8aff053a8fa)

Set base path in spin-http rather than joining it as part of the route.
Split test configs in spin-testing because the trigger's configuration
is different for http and redis.

The latter isn't a user-facing bug, but the `base` is set internally to `/` (because there is no such key in metadata) and all routes have a `base` prefix in their paths. This commit makes the behavior consistent with the redis trigger.

Also, out of the current testing crate, it looks like it's possible to have different triggers (`http` and `redis`) simultaneously in one spin app. According to the current templates, I couldn't figure out how to do this, so I split testing configs for now.


mikkelhegn:
I can confirm the redis-rust trigger template works with a build from this PR.

lann:
> [chore: Drop trigger-new leftovers](https://github.com/fermyon/spin/commit/ac4dcbadc1d63b0116fe5b55f529b719f27b1ba4)

Thanks. This must have been a bad rebase conflict resolution from right before merging.

crates/redis/src/lib.rs
```diff
@@ -50,8 +57,8 @@ impl TriggerExecutor for RedisTrigger {
     fn new(engine: TriggerAppEngine<Self>) -> Result<Self> {
         let address = engine
             .app()
-            .require_metadata("redis_address")
-            .context("Failed to configure Redis trigger")?;
+            .require_metadata::<TriggerMetadata>("trigger")?
+            .address;
```
lann, line 61: It's only been three weeks and I can't even remember switching back to `trigger` here, let alone _why_. :thinking: 

This fix looks correct so I think its good to merge, but I'll probably be reverting this and fixing the matching metadata generation later (which should be fine as the metadata is an unstable interface).

crates/testing/src/lib.rs
```diff
@@ -37,13 +37,18 @@ macro_rules! from_json {
 }
 
 #[derive(Default)]
-pub struct TestConfig {
+pub struct HttpTestConfig {
```
lann, line 40: Good change, thanks. :+1:

crates/trigger/src/locked.rs
```diff
@@ -92,9 +92,8 @@ impl LockedAppBuilder {
 
                 let trigger_type;
                 match (app_trigger, config) {
-                    (ApplicationTrigger::Http(HttpTriggerConfiguration{base}), TriggerConfig::Http(HttpConfig{ route, executor })) => {
+                    (ApplicationTrigger::Http(HttpTriggerConfiguration{base: _}), TriggerConfig::Http(HttpConfig{ route, executor })) => {
                         trigger_type = "http";
-                        let route = base.trim_end_matches('/').to_string() + &route;
```
lann, line 97: Ugh, missed this in a meta-refactor of the refactor PR.

lann (approved):
Thanks @etehtseal; clearly we don't have enough test coverage... :grimacing:

PR #802, opened 2022-10-03
Deduplicate some spin-testing code

lann:
The test configs were just split into http and redis variants. This follow-up reduces duplicated code between those.


lann:
@etehtsea FYI

etehtsea:
> @etehtsea FYI

I wasn't sure that my PR is "the right way" to fix it.
Thanks for cleaning it up!

radu-matei (approved):

Issue #804, opened 2022-10-03
`healthz` endpoint is not relative to HTTP base


itowlson:
I deployed a test app using a spin.toml which set a non-root HTTP base:

```
spin_version = "1"
name = "deploytest15"
### note the base ###
trigger = { type = "http", base = "/b" }
version = "0.15.0"

[[component]]
id = "deploytest"
source = "target/wasm32-wasi/release/deploytest2.wasm"
[component.trigger]
route = "/r"
```

After `spin up`, I tried to curl the health check endpoint:

```
# content is correctly based at /b
$ curl -i http://127.0.0.1:3000/b/r
HTTP/1.1 200 OK
content-length: 16
date: Mon, 03 Oct 2022 23:47:47 GMT

# ...output omitted...

# healthz based at root...
$ curl -i http://127.0.0.1:3000/healthz
HTTP/1.1 200 OK
content-length: 2
date: Mon, 03 Oct 2022 23:47:58 GMT

OK

# ...but not at /b
$ curl -i http://127.0.0.1:3000/b/healthz
HTTP/1.1 404 Not Found
content-length: 0
date: Mon, 03 Oct 2022 23:48:04 GMT
```

Not sure if this is expected?  We can't currently have two applications on the same port so maybe it's not an issue for now...?

lann:
`/healthz` is a Spin health check, not an application health check. I would be in favor of moving it to a better namespaced path, like `/.spin/health`.

itowlson:
Endpoint no longer exists. Also, `base` no longer exists. 💥

PR #805, opened 2022-10-04
Wait for deployment to be healthy

### LINKED ISSUE #619: Deploy: is it possible to wait for application ready?
At the moment, `spin deploy` returns as soon as Hippo accepts the deployment request.  If the user is excited and quickly clicks on the link (or already has a previous version open, and refreshes), they may get a 502 gateway error or 404, as Traefik is provisioning and the Spin server job is starting up.  This dims their excitement, and can even be misleading, as the user (me. It was me) thinks there is a problem with their application - particularly if, for example, they are already in the middle of a fault investigation (YES OKAY YES IT WAS ME).

Could we detect when the application is actually ready to serve, and print the URLs only at that point?

cc @bacongobbler as I don't know if Hippo is best placed to assess the status of Nomad/Traefik, or the Spin client should do this by poking an application health endpoint

itowlson:
Fixes #619.

At the moment, if you do a `spin deploy` and then eagerly click on the link for your hello-world dopamine, you get a 404 and it can feel like it hasn't worked.

This PR polls the deployed app's `/healthz` endpoint before printing the app's URL.  It will poll for up to 60 seconds (configurable).  If the app doesn't come ready in that time, it still prints the URL anyway so the user can keep trying or see if real URLs work etc.

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -11,7 +11,7 @@ use spin_loader::local::config::{RawAppManifest, RawAppManifestAnyVersion};
 use spin_loader::local::{assets, config};
 use spin_manifest::{HttpTriggerConfiguration, TriggerConfig};
 use std::fs::File;
-use std::io::copy;
+use std::io::{copy, Write};
 use std::path::PathBuf;
 use url::Url;
 use uuid::Uuid;
@@ -120,6 +120,12 @@ pub struct DeployCommand {
     /// Deploy existing bindle if it already exists on bindle server
     #[clap(short = 'e', long = "deploy-existing-bindle")]
     pub redeploy: bool,
+
+    /// How long in seconds to wait for a deployed HTTP application to become
+    /// ready. The default is 60 seconds. Set it to 0 to skip waiting
+    /// for readiness.
+    #[clap(long = "readiness-timeout", default_value = "60")]
+    pub readiness_timeout_secs: u16,
 }
 
 impl DeployCommand {
@@ -140,7 +146,7 @@ impl DeployCommand {
 
         let bindle_id = self.create_and_push_bindle(buildinfo).await?;
 
-        let _sloth_warning = warn_if_slow_response(&self.hippo_server_url);
+        let sloth_warning = warn_if_slow_response(&self.hippo_server_url);
 
         let token = match Client::login(
             &Client::new(ConnectionInfo {
@@ -218,6 +224,9 @@ impl DeployCommand {
             }
         };
 
+        // Hippo has responded - we don't want to keep the sloth timer running.
+        drop(sloth_warning);
+
         println!(
             "Deployed {} version {}",
             name.clone(),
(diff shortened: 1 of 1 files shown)


itowlson:
This assumes current `/healthz` behaviour.  If #804 is deemed a bug and fixed, then the logic here will need updating too

src/commands/deploy.rs
```diff
@@ -398,6 +415,74 @@ impl DeployCommand {
     }
 }
 
+async fn wait_for_ready(
+    address: &str,
```
lann, line 419: It's a bit confusing to me to call this "address", which sounds like a `SocketAddr`. "host" perhaps?

src/commands/deploy.rs
```diff
@@ -398,6 +415,74 @@ impl DeployCommand {
...
+
+    if cfg.components.is_empty() {
+        return;
+    }
+
+    let url_result = Url::parse(hippo_url);
+    let scheme = match &url_result {
+        Ok(url) => url.scheme(),
+        Err(_) => "http",
+    };
```
lann, line 436: What's the rationale here? It doesn't seem like the hippo URL scheme and app host scheme would have anything to do with each other.

src/commands/deploy.rs
```diff
@@ -398,6 +415,74 @@ impl DeployCommand {
...
+            println!("Application deployed, but Spin could not establish readiness");
+            return;
+        }
+        tokio::time::sleep(poll_interval).await;
+    }
+}
+
+fn readiness_poll_interval(readiness_timeout_secs: u16) -> tokio::time::Duration {
+    let millis = std::cmp::min(u64::from(readiness_timeout_secs) * 1000 / 60, 2000);
+    tokio::time::Duration::from_millis(millis)
```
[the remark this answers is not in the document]
lann: Actually, I'm a little skeptical of this approach; maybe better to just hard-code a 2s interval?

src/commands/deploy.rs
itowlson: This is the logic we use in `print_available_routes` - if it's wrong we need to update it there too.  I'll see if there's a way for Hippo to report the scheme, but I think the idea was that local platforms were typically non-TLS for both Hippo and apps, and remote platforms typically TLS for both.

src/commands/deploy.rs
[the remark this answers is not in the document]
itowlson: Yeah, I was a bit wary of printing a dot every 2s if the user chose a really long timeout, but I think you're right.  ON THEIR OWN HEADS BE IT

src/commands/deploy.rs
itowlson: Changed to `app_domain` which better fits the calling site.

src/commands/deploy.rs
lann: Ah, so its hacks then.

I think it would be less confusing if both this and `print_available_routes` took an e.g. `app_base_url` which could be built separately with the hack documented.

Given that this is copying an existing pattern it doesn't need to block this PR.

PR #807, opened 2022-10-04
More flexibility and guidance in the `spin new` experience

itowlson:
Currently, a developer new to Spin and wanting to create an application needs to manually install templates, and needs to know the ID of the template they want to use.

This PR makes it optional to pass the template ID on the command line.  If the template ID is omitted, `spin new` prompts with a list of installed templates:

image

Additionally, if `spin new` needs to prompt but _no_ templates are installed, `spin new` offers to install the "default" set:

image

If the template ID and project name _are_ passed on the command line, `spin new` continues to behave as it did before.

**Note:** Because the ID and project name are positional args, this PR necessarily also makes the name optional.  It will be prompted for if not provided.  This creates an additional experience of `spin new <template_id>` where you will be prompted for the name but not the template.  I don't think this is vastly valuable but it's benign so I haven't worried about it.

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -9,17 +9,17 @@ use clap::Parser;
 use path_absolutize::Absolutize;
 use tokio::{fs::File, io::AsyncReadExt};
 
-use spin_templates::{RunOptions, TemplateManager};
+use spin_templates::{RunOptions, Template, TemplateManager};
 
 /// Scaffold a new application or component based on a template.
 #[derive(Parser, Debug)]
 pub struct NewCommand {
     /// The template from which to create the new application or component. Run `spin templates list` to see available options.
-    pub template_id: String,
+    pub template_id: Option<String>,
 
     /// The name of the new application or component.
     #[clap(value_parser = validate_name)]
-    pub name: String,
+    pub name: Option<String>,
 
     /// The directory in which to create the new application or component.
     /// The default is the name argument.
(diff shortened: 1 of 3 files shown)


src/commands/new.rs
```diff
@@ -124,6 +134,77 @@ fn merge_values(from_file: &mut HashMap<String, String>, from_cli: &[ParameterVa
...
+        .await
+        .context("Failed to install the default templates")?;
+    Ok(())
+}
+
+async fn prompt_name() -> anyhow::Result<String> {
+    let mut input = dialoguer::Input::<String>::new();
+    input.with_prompt("Enter a name for your new project");
+    let result = input.interact_text()?;
+    Ok(result)
```
lann, line 205: I wonder if this should loop on empty input, printing something like:

```
Project name is required. (ctrl-c to cancel)
```

src/commands/new.rs
```diff
@@ -124,6 +134,77 @@ fn merge_values(from_file: &mut HashMap<String, String>, from_cli: &[ParameterVa
...
+const DEFAULT_TEMPLATES_INSTALL_PROMPT: &str =
+    "You don't have any templates yet. Would you like to install the default set?";
+const DEFAULT_TEMPLATE_REPO: &str = "https://github.com/fermyon/spin";
+
+async fn get_or_install_templates(
+    template_manager: &TemplateManager,
+) -> anyhow::Result<Option<Vec<Template>>> {
+    let templates = template_manager.list().await?.templates;
+    if templates.is_empty() {
+        let should_install = dialoguer::Confirm::new()
```
kate-goldenring, line 168: Love the use of the dialoguer! We should use it for plugins, too.

src/commands/new.rs
kate-goldenring: Right now, looks like the dialoguer doesnt continue until something is entered, which seems like proper behavior

kate-goldenring (approved):
Love this experience, especially the choosing the template scrollbar!

src/commands/new.rs
itowlson: Looks like the `dialoguer` component already rejects empty input, though it does accept "all spaces."  (Which, to my surprise, we actually cope with.)  I'll trim and loop for that case just in case (though realistically we're more interested in protecting against Murphy vs. against Machiavelli).

fibonacci1729 (approved):

lann (approved):

radu-matei (approved):
The new experience is GREAT!

LGTM

Issue #576, opened 2022-06-16
How to design the WIT layer for pg host component


miketang84:
Hi, I'm designing the feature of supporting the postgres as a host component, the first problem is:

The client.query() returns `Vec<Row>` which is defined in https://docs.rs/tokio-postgres/0.7.6/tokio_postgres/row/struct.Row.html.

And the outbound-redis sample just uses a plain string as parameters.

I wonder that if I need to map the definition of Row and its inner types into the wit types one by one (maybe it's difficult, and it has two generics in its' method parameters), or is there some simple way to just use the string as input and output parameters.

Thank you.

miketang84:
Oh, I found the item of `resource` identifier.

https://github.com/bytecodealliance/wit-bindgen/blob/main/WIT.md#item-resource

Wish it helps.

dicej:
I agree that representing a `Row` as a WIT `resource` is the most natural option.  Regarding `get`/`try_get` methods: since WIT does not support arbitrary generics, it may be necessary to split each of those methods separate versions for each index and value type.  For example:

```
resource row {
  columns: func() -> list<column>
  is-empty: func() -> bool
  len: func() -> u64
  get-i64-by-index: func(index: u32) -> expected<i64, error>
  get-string-by-index: func(index: u32) -> expected<string, error>
  get-bytes-by-index: func(index: u32) -> expected<list<u8>, error>
  ...
  get-i64-by-name: func(name: string) -> expected<i64, error>
  get-string-by-name: func(name: string) -> expected<string, error>
  get-bytes-by-name: func(name: string) -> expected<list<u8>, error>
  ...
}
```

Since PostgreSQL has _a lot_ of types, that would end up being a lot of methods, and even then it wouldn't support all the user-defined `FromSql` types `tokio-postgres` does.  An alternative would be to only support `get-bytes-by-index` and `get-bytes-by-name`, using some language-agnostic serialization format (e.g. JSON or CBOR) to encode the data.

miketang84:
I found that the redis outbound sample is too simple, on every publish/set/get, it open and close the connection in the host function implementation. This design avoids the definition of something like `resource` to keep the redis connection across the border of the wasmtime and the wasm sandbox. But for the prototype, I think it makes sense.

miketang84:
[...]

PR #813, opened 2022-10-11
feat(spin_sdk::pg): Extract type conversions into sdk

etehtsea:
This has to be the starting point on the way to providing the API that is able to automatically map a DB response into the user-defined struct.


itowlson (commented):
I'm happy for this to go into the SDK, but I suggest we have to make it more comprehensive first.  A sample can get away with just doing the things needed for the sample (and to show the pattern), but a blessed library needs to be systematic.  We'd also need to manage things like SQL NULL handling, which the sample dodged by making all of the columns NOT NULL.  **grin**

We do have fuller conversion coverage in the .NET SDK (https://github.com/spin/spin-dotnet-sdk) -- if we have parity with that then that would be fine I reckon.  Thoughts?

etehtsea:
@itowlson I've added decoding support for the following `DbValue`s so as their NULLable variants:
- `boolean(bool)`
- `int16(s16)`
- `int32(s32)`
- `int64(s64)`
- `floating32(float32)`
- `floating64(float64)`
- `str(string)`
- `binary(list<u8>)`

Initially, I wanted to review/test types mapping inside the `outbound-pg` crate first because, during experimenting, I've got a few thoughts/concerns:
- `convert_data_type` doesn't seem to handle all types available in the wit (at least `INT2` seems to be missing) (https://github.com/fermyon/spin/blob/main/crates/outbound-pg/src/lib.rs#L133-L149);
- `tokio-postgres` maps several pg types into one rust type (https://docs.rs/postgres-types/latest/postgres_types/trait.FromSql.html#types) but `convert_entry` [maps only *some* types from the list](https://github.com/fermyon/spin/blob/main/crates/outbound-pg/src/lib.rs#L159-L235) (e.g. `INT4`, but not `INT` and `SERIAL`);

The rest of the list is probably out of the scope of this PR but is related to:
- https://github.com/fermyon/spin/pull/777
- https://github.com/fermyon/spin/issues/576
- https://github.com/fermyon/spin-dotnet-sdk/tree/main/wit/ephemeral

So here are my random thoughts:
* Regarding other SQL DBs support: rather than adding support for different SQL DBs using various crates from different maintainers wouldn't it be easier to utilize [sqlx](https://github.com/launchbadge/sqlx) for this matter?
Originally, I was thinking that `sqlx` is just the layer above different DB drivers, but during my experiments last week I realized that this is not the case.
Besides checking the `sqlx` itself another goal of this experiment was to check the ability to get information about the column's `NOT NULL` constraint from the column info. Unfortunately, there is no way to do this for dynamic SQL queries (at least in PG).
* Regarding WIT format: at this moment I'm not sure that generic db values types are convenient because:
  - what to do with db-specific types?
  - there is no way to figure out the original (DB) type. For example, how to (de)serialize timestamps? They could be encoded as `DbValue::Str(RFC 3339/ISO 8601)` but this way it's not clear how to deserialize them. So it makes sense to add a separate enum variant for them.
  - current `DbValue`'s format seems to be modeled after Rust's types but different languages might approach DB types in a different way (https://pkg.go.dev/github.com/lib/pq#hdr-Data_Types)
All in all, it might be easier to introduce different mappings for different SQL DBs that are close to their types.
* At this moment, it's not clear how to handle breaking WIT format changes (https://github.com/fermyon/spin-dotnet-sdk/tree/main/wit/ephemeral). Any change to the format will break C# SDK. When it gets more stable, it wouldn't be a big problem, but for now, it might become an issue.

itowlson:
Thanks for the great feedback.  `sqlx` certainly looks like it might be a great four-for-the-price-of-one solution - I didn't know about it before so thanks for the pointer!

I agree we need a better story for DB-specific types and generally on how to represent DB-specific operations and options.  There's a reason every database gets its own library and we can't afford to smooth that over too much.  One option longer term is to not even have database WIT specs, and provide infrastructure whereby source languages can use their own database crates / packages / modules / etc.  But that's hard to do at the moment because most of them sit over raw sockets, and that sockets code doesn't yet compile to Wasm.

The current situation is a sighting shot at what we can do in the meantime, and there's certainly a good case for moving the data types out of a common WIT and into database-specific ones.  But at the WIT level we'd probably still end up with a big honkin' `variant`, which would require languages like Go and C# to do some manual unpicking to make it idiomatic.  As noted in the MySQL POC, there are other things we should look at when we approach this, such as whether to go for a more resource-oriented view of connections and transactions and so on.  In any case, we're very happy to iterate on this and we'd welcome your (and other contributors') thoughts and advice.

Regarding how we evolve WIT files - this (and contract versioning more generally) is a big concern of mine, and folks within Fermyon are actively debating when to stabilise various contracts and how to version them thereafter.  We have muddled along so far by batching up breaking changes to limit the number of times we have to perform breaks, but we do need a more robust and systemic solution.  Unfortunately, we don't yet know what that looks like, and it likely needs to align to broader Wasm Component Model standards which are still in flux.  Sorry I can't give you a more satisfying answer for now, but it's very much on our radar.

Thanks again for the detailed thoughts.  This is a very early-stage area and it's so helpful to have wider conversations about it!

etehtsea:
> such as whether to go for a more resource-oriented view of connections and transactions and so on
Yeah, I was looking into more resource-oriented approach but wit-bindgen [is currently experiencing a big rewrite](https://github.com/bytecodealliance/wit-bindgen/issues/314), so resource/handle support is dropped for now.
https://github.com/bytecodealliance/wit-bindgen/commit/25596c799b3bbe85f2b5939e8538a524fa4efdf2
So it looks like the current approach is the only way for the short-term.

P.S. async support is also temporary dropped
https://github.com/bytecodealliance/wit-bindgen/commit/f1682df9adacb515227d061a4601b2b4c997e7a2

etehtsea:
I've added basic integration test to test outbound-pg. It's currently excluded from the default integration test because it expects running postgresql server.
The test itself is a bit awkward but I couldn't find other easy way to test it atm.

etehtsea:
### UPD
- Added mapping table between Rust, WIT, and Postgres types
- Introduced `spin_sdk::pg::Error` and replaced `anyhow` with it in `db-value` decoding errors.

Is there something else to cover?

`int8(s8)` support wasn't added into conversions because:
- the only type that fits is `"char"` which is:

>  The type "char" (note the quotes) is different from char(1) in that it only uses one byte of storage, and therefore can store only a single ASCII character. It is used in the system catalogs as a simplistic enumeration type.
https://www.postgresql.org/docs/current/datatype-character.html

- `"char"` isn't supported in `outbound-pg`;

sdk/rust/src/pg.rs
```diff
@@ -0,0 +1,207 @@
...
+            | PgError::BadParameter(err_msg)
+            | PgError::QueryFailed(err_msg)
+            | PgError::ValueConversionFailed(err_msg)
+            | PgError::OtherError(err_msg) => write!(f, "Postgres error: {}", err_msg),
+            PgError::Success => panic!("Unexpected error: Success isn't supposed to be used"),
+        }
+    }
+}
+
+// A pg error
```
itowlson, line 36: Not sure if I'm misreading this but I think `pub` stuff should have doc strings?  The module `deny(missing-docs)` should be catching this though so perhaps I'm misunderstanding...

itowlson (changes requested):
Wow, this looks tremendous.  Thank you so much for slogging through all the types, and for putting in the hard yards to produce an properly self-contained integration test.

I feel a bit guilty for raising a nit about doc strings now... everything else looks great!

sdk/rust/src/pg.rs
etehtsea: SDK allows `missing_docs` https://github.com/fermyon/spin/blob/main/sdk/rust/src/lib.rs#L41-L60 because, otherwise it fails with:

```
error: missing documentation for an enum
  --> sdk/rust/src/lib.rs:54:5
   |
54 |     wit_bindgen_rust::import!("../../wit/ephemeral/spin-config.wit");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `wit_bindgen_rust::import` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not document `spin-sdk`
```

But you're right, this isn't the reason for not adding the docs :)
Fixed, thanks!

itowlson (approved):
LGTM - thanks!

PR #818, opened 2022-10-13
Break login into smaller functions

itowlson:
This should have no observable effect on the UX* - it just breaks up the LoginCommand::run() function into smaller bits that hopefully make the various flows more visible and tractable.

`*` oh okay, it does have one observable effect - the `--get-device-code` and `--check-device-code` arguments now override `--auth-method` whereas previously it was the other way round - but mixing those flags is a bit iffy anyway

### DIFF
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -202,22 +202,11 @@ impl LoginCommand {
     }
 
     async fn run_interactive_basic_login(&self) -> Result<LoginConnection> {
-        let username = match &self.hippo_username {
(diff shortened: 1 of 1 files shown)


itowlson:
I was doing some further testing on this and hit a display issue with interactive prompts (prompts weren't displaying until a newline character arrived).  I've included a fix for that here, as a separate commit, but can split it into a separate PR if we want.

radu-matei (approved):
Reviewed and tested as part of #819.

LGTM

PR #824, opened 2022-10-18
Improve deployment experience

itowlson:
This PR:

* modifies the sloth warnings to be "checking status" (for the Hippo health check) and "uploading application" (for Bindle) instead of warning about a slow response
* adds the "wait for application ready" UI when deploying using a GH token, so the user is less likely to see the 404 of application not ready

It also removes what seems to have been a duplicate sloth warning for the bindle upload, probably a spurious leftover from an earlier merge.

### DIFF
--- a/src/commands/bindle.rs
+++ b/src/commands/bindle.rs
@@ -180,7 +180,7 @@ impl Push {
             .await
             .with_context(|| crate::write_failed_msg(bindle_id, dest_dir))?;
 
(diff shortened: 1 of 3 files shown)


mikkelhegn:
This is my experience, and it's much better now - Thx @itowlson!

radu-matei (approved):

karthik2804 (approved):

PR #826, opened 2022-10-18
command(login): validate that the provided hippo URL is valid

endocrimes:
I have no idea what the "idomatic rust" way to do this is, so I'm going vaguely based on how we return errors elsewhere in the `login` command.

Previous behavior:
One would be prompted to choose an auth method before then receiving a `reqwest` error.

New behavior:

```
[spin(dani/validate-hippo-url)] $ ./target/debug/spin login --url some-nonsnese
Error: Invalid Hippo URL, should be in the format "https://my-hippo-instance.com". "some-nonsnese": relative URL without a base
```

Happy to workshop on the error message itself (and if you point me to where integration testing happens I'm happy to add a regression test).

Signed-off-by: Danielle Lancashire <dani@builds.terrible.systems>

### DIFF
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -14,6 +14,7 @@ use serde::Serialize;
 use serde_json::json;
 use tokio::fs;
 use tracing::log;
+use url::Url;
 use uuid::Uuid;
 
 use crate::opts::{
@@ -71,8 +72,10 @@ pub struct LoginCommand {
         name = HIPPO_SERVER_URL_OPT,
         long = "url",
         env = HIPPO_URL_ENV,
+        default_value = DEFAULT_CLOUD_URL,
+        value_parser = parse_url,
     )]
-    pub hippo_server_url: Option<String>,
+    pub hippo_server_url: url::Url,
 
     /// Hippo username
     #[clap(
@@ -153,6 +156,16 @@ pub struct LoginCommand {
         conflicts_with = "check-device-code"
     )]
     pub list: bool,
+}
+
+fn parse_url(arg: &str) -> Result<url::Url> {
+    let u = arg.trim().trim_end_matches('/');
+    Url::parse(u)
+        .map_err(|error| {
+            anyhow::format_err!(
+                "URL should be fully qualified in the format \"https://my-hippo-instance.com\". Error: {}", error
+            )
+        })
 }
 
 impl LoginCommand {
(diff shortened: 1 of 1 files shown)


src/commands/login.rs
```diff
@@ -367,6 +372,18 @@ impl LoginCommand {
         }
     }
 
+    fn validate_hippo_url(&self) -> Result<(), anyhow::Error> {
```
lann, line 375: I would say this should probably happen in `fn url(&self)`, i.e. changing its signature to return `Result<Url>`

src/commands/login.rs
```diff
@@ -367,6 +372,18 @@ impl LoginCommand {
...
+        let url = self.url();
+        match Url::parse(url) {
+            Ok(_) => Result::Ok(()),
+            Err(error) => Result::Err(anyhow::format_err!(
+                "Invalid Hippo URL, should be in the format \"https://my-hippo-instance.com\". \"{}\": {}",
+                url,
+                error
+            )),
+        }
```
lann, line 384: There are a couple of nicer options here, depending on exactly what you want the errors to look like.

The easiest option (which we use all over the place) is the `.with_context` method, which attaches context to another error, e.g.:

```suggestion
        Url::parse(url).with_context(|| {
            format!("Invalid Hippo URL, should be in the format \"https://my-hippo-instance.com\". {url:?}")
        })
```

That will give you a chained "caused by" list of errors, which might not be what you want here.

A shorter way to get (almost) exactly the error above (untested, hopefully right):

```suggestion
        Url::parse(url)
            .map_err(|err| anyhow::anyhow!(
                "Invalid Hippo URL, should be in the format \"https://my-hippo-instance.com\". {url:?}: {err}"
            )
```

src/commands/login.rs
endocrimes: We would then need to randomly call `self.url()?;` to validate early which feels a _bit_ weird conceptually, but otherwise seems fine to me.

lann: Another option would be to let `clap` (the CLI arg parser) validate this. That is super easy if you don't care about precise error messaging but a little trickier if you want to customize errors (as in I don't remember how to do it off the top of my head :sweat_smile:).

radu-matei: nit: I think we now call the flag just `--url` — could be `validate_url` (or something similar) instead?

lann (approved):
Nice :+1:

PR #835, opened 2022-10-19
do not deploy redis triggers on cloud

bacongobbler:
Redis triggers are not supported by `spin deploy` to the Cloud.

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -12,6 +12,7 @@ use sha2::{Digest, Sha256};
 use spin_http::routes::RoutePattern;
 use spin_loader::local::config::{RawAppManifest, RawAppManifestAnyVersion};
 use spin_loader::local::{assets, config};
+use spin_manifest::ApplicationTrigger;
 use spin_manifest::{HttpTriggerConfiguration, TriggerConfig};
 use spin_publish::BindleConnectionInfo;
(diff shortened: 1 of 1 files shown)


src/commands/deploy.rs
```diff
@@ -712,3 +717,10 @@ fn print_available_routes(
         }
     }
 }
+
+fn check_trigger_is_supported(cfg: &RawAppManifest) -> Result<()> {
+    match cfg.info.trigger {
+        ApplicationTrigger::Redis(_) => bail!("Redis triggers are not supported"),
```
itowlson, line 723: Should we allow-list rather than deny-list?  For example if and when we implement a cron or queue trigger, allow-list might result in more correct behaviour if we forget to update the list...?

bacongobbler: Good call.

bacongobbler: Note that I just tried this and I get an "unreachable pattern" warning after exhausting all other possible options. So I think this is fine, because new trigger types will require an explicit addition to the pattern matcher.

itowlson: Ah!  For some reason I thought they were not all listed (and there was a default clause) but no, it's all explicit so 100% safe.  Sorry about that.

radu-matei (approved):

PR #840, opened 2022-10-21
Use random buildinfo in cloud_deploy

lann:
This is a temporary workaround to some bad UX from partial uploads to bindle.


lann:
Tested locally

fibonacci1729 (approved):

endocrimes (approved):
lgtm

PR #844, opened 2022-10-21
Improve error message on missing manifest content path

### LINKED ISSUE #838: `spin up`: No such file or directory (os error 2)
I am trying to run the finicky-whiskers app with `spin up`, but I am hitting a missing file/directory error. The question I ask when using the command is - what file or directory am I missing? Can we improve the error message to tell the user what's missing?

```
><> spin build
No build command found!
><> spin up
Error: No such file or directory (os error 2)
```

I think this is because the directory is missing the `session` component:

```
# Initialize session data
[[component]]
id = "session"
source = "session/session.wasm"
[component.trigger]
executor = { type = "wagi", argv = "${SCRIPT_NAME} -v /lib/session.rb ${SCRIPT_NAME} ${ARGS}" }
route = "/session"
```

```
><> ls session/
Gemfile  Gemfile.lock  Makefile  README.md  lib  spin.toml
```

But I had to look at spin.toml to figure out what might've been missing.

lann:
Fixes #838

### DIFF
--- a/crates/trigger/src/locked.rs
+++ b/crates/trigger/src/locked.rs
@@ -177,7 +177,9 @@ impl LockedAppBuilder {
 
 fn content_ref_path(path: &Path) -> Result<ContentRef> {
     Ok(ContentRef {
-        source: Some(file_uri(path)?),
+        source: Some(
+            file_uri(path).with_context(|| format!("failed to resolve content at {path:?}"))?,
+        ),
         ..Default::default()
     })
 }


lann:
```
Error: failed to resolve content at "/[SNIP]/finicky-whiskers/session/session.wasm"

Caused by:
    No such file or directory (os error 2)
```

fibonacci1729 (approved):

PR #845, opened 2022-10-21
Bump version to 0.6.0

fibonacci1729:


vdice (commented):
Shall we update the appearances of `v0.5.0` in the https://github.com/fermyon/spin/blob/main/docs/downloads/install.sh script as well?

Otherwise, it looks like we've got the initial steps covered in https://spin.fermyon.dev/release-process.

vdice (approved):

PR #864, opened 2022-10-27
Add/enhance MySQL support

etehtsea:
Rebased https://github.com/fermyon/spin/pull/777 against current master, fixed a few conflicts/incompatibilities.
@itowlson, will be happy to rebase against your branch if you prefer to rebase/sign it yourself.

- Extract outbound-mysql into separate crate
- Add integration test
- Add connections reusing during the request
- Improve MySQL types conversion


etehtsea:
@itowlson this one is also ready for review!

itowlson (approved):
Thank you!  This looks great and I'm 100% happy with it.

But...

I want to hold off a bit on this because I think we need to have a couple of broader discussions:

1. Do we want to invest more in this API for databases, or do we want to switch to a more resource-oriented approach which better models providers, connections, transactions, etc., and could have a unified-ish interface instead of multiple parallel ones?

2. How do we manage new host components in a way that doesn't require big-bang updates to client, SDK and cloud?  That's not technically a blocker for this because it's purely additive, but it makes me wary of throwing in new interfaces (and yeah we want to do key-value stores soon so maybe I should just suck it up for another few months).

I hope this makes sense.  We can always kick the can down the road: there's nothing about this PR that makes it a specific sticking point.  So if you or anyone is looking to use this, we can go ahead.  But otherwise I'd like to broaden the discussion a bit first...

itowlson:
@etehtsea I tagged you in on the kick-off draft of `wasi-sql`.  At the moment, what they have there is close enough to what you have here that we could consider pushing ahead on this basis.  Or we could try to align more closely, though I'm not sure if the `resource` stuff is properly available yet, and the draft will certainly change.  What do you reckon?

etehtsea:
@itowlson thanks! I left a comment there. As you said earlier, if there is somebody interested in MySQL support, it should be good to go (for now).

I was curious about the differences between pg and MySQL implementations. And my conclusion is if the goal is to provide more or less full support for specific DB, it's hard to do this using a shared wit interface between them.

P.S. Also, I've checked wasi preview2 today and it seems that sockets support is somewhere near. So, if there be an ability to use the "native" library from the guest module, it probably be the solution for those who needs full DB support.

itowlson:
@radu-matei (and anyone else interested in this!)  Despite earlier reservations (expressed upthread), I'd like to merge this.  The interface is well aligned with the current `wasi-sql` proposal, and I think comparative feedback from users of Postgres and MySQL would help inform the discussions there.  It does, of course, mean another interface to maintain and, eventually, deprecate as `wasi-sql` evolves (and as the goals for it become clearer).

Let me know if you object to going ahead with this.

Issue #367, opened 2022-04-17
Track code coverage


radu-matei:
We should consider tracking code coverage for Spin.

A possible solution could be https://github.com/xd009642/tarpaulin and https://about.codecov.io.

jdbohrman:
This is kinda relevant to #343 also

lann:
I find tarpaulin + [this extension](https://marketplace.visualstudio.com/items?itemName=markis.code-coverage) pretty helpful.

FrankYang0529:
I tried tarpaulin on Ubuntu 22.04. My command is `cargo tarpaulin -v --skip-clean -t 600 --out html` and it uses `Ptrace` engine to track the coverage. The coverage result is 24.47%. I will add it to the CI flow with codecov.io.

GitHub can't upload HTML files, so I add the extension `.txt`.
[tarpaulin-report.html.txt](https://github.com/fermyon/spin/files/9894122/tarpaulin-report.html.txt)

PR #866, opened 2022-10-31
ci: add codecov

FrankYang0529:
resolve https://github.com/fermyon/spin/issues/367

Need to setup `CODECOV_TOKEN`. Thanks


.github/workflows/build.yml
```diff
@@ -122,12 +122,40 @@ jobs:
...
+          use-tool-cache: true
+
+      - name: Cargo Tarpaulin (ubuntu)
+        if: ${{ fromJSON(matrix.config.os == 'ubuntu-latest') }}
+        run: cargo tarpaulin --follow-exec --skip-clean -t 6000 --out xml --features openssl/vendored,default,e2e-tests
+        env:
+          RUST_LOG: spin=trace
+
+      - name: Upload to codecov.io (ubuntu)
+        uses: codecov/codecov-action@v1.0.2
```
vdice, line 147: While reading up on this action, I noticed the following disclaimer: https://github.com/codecov/codecov-action#%EF%B8%8F--deprecation-of-v1

Can we bump to a more recent (3.x?) version?

.github/workflows/build.yml
```diff
@@ -122,12 +122,40 @@ jobs:
...
+        if: ${{ fromJSON(matrix.config.os == 'ubuntu-latest') && github.ref == 'refs/heads/main' }}
+        with:
+          token: ${{secrets.CODECOV_TOKEN}}
+
+      - name: Archive code coverage results (ubuntu)
+        uses: actions/upload-artifact@v1
```
vdice, line 153: In the continued spirit of bumping to latest versions, can we bump this one as well?  Looks like v3.x is the current one...

vdice (commented):
Thanks @FrankYang0529!

Re: the CODECOV_TOKEN... it looks like a separate token won't be necessary for this public spin repo, if I'm understanding correctly: https://github.com/codecov/codecov-action#usage

Have you been able to test this on your fork, by chance?

kate-goldenring:
Once we have the token figured out/removed, I am curious whether installing Tarpaulin on the GitHub runner or using the Tarpaulin containers to execute the tests (like here: https://github.com/project-akri/akri/blob/main/.github/workflows/run-tarpaulin.yml) is faster. Either way it can take a long time to run

.github/workflows/build.yml
FrankYang0529: Updated it. Thanks!

.github/workflows/build.yml
FrankYang0529: Updated it. Thanks.

FrankYang0529:
> Re: the CODECOV_TOKEN... it looks like a separate token won't be necessary for this public spin repo, if I'm understanding correctly: https://github.com/codecov/codecov-action#usage

Yeah, it looks like we don't need it. I updated a version for it.

> Have you been able to test this on your fork, by chance?

Yes, this one https://github.com/FrankYang0529/spin/actions/runs/3369212681/jobs/5588606673

FrankYang0529:
> Re: the CODECOV_TOKEN... it looks like a separate token won't be necessary for this public spin repo, if I'm understanding correctly: https://github.com/codecov/codecov-action#usage

Yeah, it looks like we don't need it. I updated a version for it.

> Have you been able to test this on your fork, by chance?

Yes, this one https://github.com/FrankYang0529/spin/actions/runs/3369212681/jobs/5588606673

FrankYang0529:
> Once we have the token figured out/removed, I am curious whether installing Tarpaulin on the GitHub runner or using the Tarpaulin containers to execute the tests (like here: https://github.com/project-akri/akri/blob/main/.github/workflows/run-tarpaulin.yml) is faster. Either way it can take a long time to run

I am not sure whether it's faster, but I would prefer not to use containers here, because:
* We need some dependencies like bindle, nomad, and hippo to run e2e-tests.
* [Default runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) only have 14G SSD. If we use volume with containers, it may take extra space on the host.

vdice:
Thanks for the updates @FrankYang0529.  I see we upload the codecov report as an artifact on the workflow run; what do we envision the current flow for tracking these reports over time?

Would it be possible to have the codecov results be a separate check in CI?  I'm thinking it would be handy to see it alongside the other checks (currently the `Rust / Build Spin` and DCO checks).  And maybe eventually we'd have a threshold where if code coverage dips below a certain amount, the check would fail. Perhaps this would involve authorizing a Codecov GH App?  Assuming it would take a non-trivial amount of config/work, it can certainly be a follow-up (if it's even something the Spin team wants).

FrankYang0529:
> Would it be possible to have the codecov results be a separate check in CI?

Yes, I think it's good to separate these in another check so that we can unify test commands in the `Build Spin` check. I will update it tomorrow.

> And maybe eventually we'd have a threshold where if code coverage dips below a certain amount, the check would fail.

I would suggest we don't do this now because the result from the Ptrace engine with `--follow-exec` is not very accurate. We have a test case for `spin deploy`, and I added `e2e-tests` in the `--features`, but the coverage rate for [`deploy.rs`](https://app.codecov.io/gh/FrankYang0529/spin/tree/main/src/commands) is 0.

> Perhaps this would involve authorizing a Codecov GH App?

I will do some surveys and test in my repo first.

vdice:
Hi @FrankYang0529, checking in to see if I can help on this one.  There may be a bit of refactoring needed now that the build.yml workflow has been broken up into multiple jobs.  If it isn't too complex, it would still be great to have the codecov checks run as their own job/check (ref https://github.com/fermyon/spin/pull/866#issuecomment-1305683732).  Otherwise, I'm sure we could also get a version of this in and start refining/interating from there.

FrankYang0529:
Hi @vdice, sorry for the late update. I will update the PR this weekend. Thanks.

.github/workflows/build.yml
```diff
@@ -250,3 +250,83 @@ jobs:
...
+
+      - name: Install cargo-tarpaulin binary crate
+        uses: actions-rs/install@v0.1
+        with:
+          crate: cargo-tarpaulin
+          version: latest
+          use-tool-cache: true
+
+      - name: Cargo Tarpaulin
+        run: cargo tarpaulin --follow-exec --skip-clean -t 6000 --out xml --features openssl/vendored,default,e2e-tests
```
vdice, line 316: ping @lann or @radu-matei on if this `cargo tarpaulin` command suits our needs per the original issue https://github.com/fermyon/spin/issues/367

.github/workflows/build.yml
```diff
@@ -250,3 +250,83 @@ jobs:
...
+        env:
+          RUST_LOG: spin=trace
+
+      - name: Upload to codecov.io
+        uses: codecov/codecov-action@v3
+        if: ${{ github.ref == 'refs/heads/main' }}
```
vdice, line 322: Are we only interested in codecov results for `main` at this time?  cc @lann @radu-matei

vdice (approved):
Thank you @FrankYang0529!

Added a few comments/questions for the broader team but otherwise the automation looks great to me.

.github/workflows/build.yml
lann: Yep that's what I use.

.github/workflows/build.yml
lann: I would imagine we'd like it for PRs too, but its easy enough to change :shrug:

vdice: Great, we can see how it works for us for `main` and then potentially create a follow-up.

vdice:
Thanks again for this contribution @FrankYang0529!

PR #867, opened 2022-10-31
Use dialoguer crate to prompt for Spin plugin installs

kate-goldenring:
Remove custom `prompt` code and use `dialoguer` crate that is used in `spin new`.


karthik2804 (approved):

crates/plugins/src/manifest.rs
```diff
@@ -30,6 +30,9 @@ impl PluginManifest {
     pub fn name(&self) -> String {
         self.name.to_lowercase()
     }
+    pub fn license(&self) -> String {
```
itowlson, line 33: Our usual convention for functions like this (and I can't remember where it came from) has been to skip the clone and return `&str`, as often callers don't need to own a copy, and if they do they can do the clone.  In this case, of course, it makes no odds, and we're certainly not consistent about doing it...!

itowlson (approved):
LGTM!

crates/plugins/src/manifest.rs
kate-goldenring: Definitely an unnecessary clone. Thanks for pointing this out @itowlson

PR #870, opened 2022-11-02
Fix redeploy readiness

lann:
Currently, a redeploy of an existing app will be seen as "ready" immediately, since the previous version is ready. This means that after a successful deploy a user might still see the old version active for a while.

This fixes the issue by introducing a new `/.well-known/spin/info` endpoint that returns some app info including the currently deployed bindle version. That version will be checked in the `spin deploy` readiness check loop to ensure that the currently deploying version is active before returning.

Additionally, refactor a bunch of stuff in here that has been bothering me :sweat_smile:

### DIFF
--- a/crates/http/src/lib.rs
+++ b/crates/http/src/lib.rs
@@ -362,11 +362,11 @@ impl HttpTrigger {
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct AppInfo {
-    name: String,
+    pub name: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    version: Option<String>,
+    pub version: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    bindle_version: Option<String>,
+    pub bindle_version: Option<String>,
 }
 
 fn parse_listen_addr(addr: &str) -> anyhow::Result<SocketAddr> {
--- a/crates/http/src/lib.rs
+++ b/crates/http/src/lib.rs
@@ -265,11 +265,8 @@ impl HttpTrigger {
     fn app_info(&self) -> Result<Response<Body>> {
         let info = AppInfo {
             name: self.engine.app_name.clone(),
-            version: self
-                .engine
-                .app()
-                .get_metadata("version")?
-                .unwrap_or_default(),
+            version: self.engine.app().get_metadata("version")?,
+            bindle_version: self.engine.app().get_metadata("bindle_version")?,
         };
         let body = serde_json::to_vec_pretty(&info)?;
         Ok(Response::builder()
@@ -366,7 +363,10 @@ impl HttpTrigger {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct AppInfo {
     name: String,
-    version: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    version: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    bindle_version: Option<String>,
 }
 
 fn parse_listen_addr(addr: &str) -> anyhow::Result<SocketAddr> {
--- a/crates/http/src/lib.rs
+++ b/crates/http/src/lib.rs
@@ -207,6 +207,7 @@ impl HttpTrigger {
         if let Some(well_known) = path.strip_prefix(WELL_KNOWN_PREFIX) {
             return match well_known {
                 "health" => Ok(Response::new(Body::from("OK"))),
+                "info" => self.app_info(),
                 _ => Self::not_found(),
             };
         }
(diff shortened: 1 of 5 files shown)


src/commands/login.rs
```diff
@@ -474,7 +478,7 @@ async fn create_device_code(client: &Client) -> Result<DeviceCodeItem> {
 
 #[derive(Clone, Serialize, Deserialize)]
 pub struct LoginConnection {
-    pub url: String,
+    pub url: Url,
```
lann, line 481: @endocrimes

crates/http/src/lib.rs
```diff
@@ -201,57 +201,79 @@ impl HttpTrigger {
             req.uri()
         );
 
-        match req.uri().path() {
```
lann, line 204: Hide whitespace
2022-11-02T10:34:41,133882335-04:00

endocrimes (approved):
SGTM as a first pass at this - we eventually might want this to go through hippo as an auth layer, but there's nothing sensitive in there for now.

src/commands/deploy.rs
```diff
@@ -687,26 +687,31 @@ async fn wait_for_ready(
...
-    let (msg, ready) = match resp {
-        Err(e) => (format!("error {}", e), false),
-        Ok(r) => {
-            let status = r.status();
-            let ok = status.is_success();
-            let desc = if ok { "ready" } else { "not ready" };
-            (format!("{desc}, code {status}"), ok)
+#[instrument(level = "debug")]
+async fn is_ready(app_info_url: &str, bindle_version: &str) -> Result<bool> {
+    // If the request fails, the app isn't ready
```
itowlson, line 692: ...yet.  Looking at the call site I think this would cause the wait-for-ready loop to print a failure message and terminate, but couldn't this happen if, say, the domain is still being brought up?  Or there's a transient network error?  I'd prefer to err on the side of being forgiving here.

src/commands/deploy.rs
```diff
@@ -687,26 +687,31 @@ async fn wait_for_ready(
...
+        tracing::debug!("App not ready: {}", resp.status());
+        return Ok(false);
+    }
+    if let Some(active_version) = resp
+        .json::<AppInfo>()
+        .await
+        .ok()
+        .and_then(|info| info.bindle_version)
+    {
+        if active_version != bindle_version {
```
itowlson, line 704: Suggest renaming `bindle_version` to `expected_version` to make this check clearer.

src/commands/deploy.rs
```diff
@@ -687,26 +687,31 @@ async fn wait_for_ready(
...
-            (format!("{desc}, code {status}"), ok)
+#[instrument(level = "debug")]
+async fn is_ready(app_info_url: &str, bindle_version: &str) -> Result<bool> {
+    // If the request fails, the app isn't ready
+    let resp = reqwest::get(app_info_url).await?;
+    if !resp.status().is_success() {
+        tracing::debug!("App not ready: {}", resp.status());
+        return Ok(false);
+    }
+    if let Some(active_version) = resp
```
itowlson, line 698: For me, this chain of combinators took a bit of mental parsing.  Could we extract it into a function such as `parse_bindle_version(response)`?

src/commands/deploy.rs
```diff
@@ -640,43 +644,39 @@ async fn check_healthz(url: &str) -> Result<()> {
...
-    let url_result = Url::parse(hippo_url);
-    let scheme = match &url_result {
-        Ok(url) => url.scheme(),
-        Err(_) => "http",
-    };
-
-    let path = WELL_KNOWN_HEALTH_PATH;
-    let health_check_url = format!("{}://{}{}", scheme, app_domain, path);
+    let app_info_url = app_base_url
+        .join(WELL_KNOWN_PREFIX.trim_start_matches('/'))
```
itowlson, line 653: I'm a bit puzzled that you diligently add the trailing slash in `parse_url` "so join works" and then need to remove the leading slash of `WELL_KNOWN_PREFIX` to compensate for that - it makes me wonder why not just declare WKP without the leading slash.  Is it because other consumers of WKP presume that the URL does _not_ have the trailing slash?

itowlson (commented):
Delighted to see this - it had been bugging me too.  A couple of small, non-blocking nits/suggestions but otherwise LGTM.  Thanks!

src/commands/deploy.rs
lann: Since we've just successfully used the network it seems much more likely for a failure here to be permanent(ish) than transient. _I_ would rather see an error immediately than a timeout in that situation, but we could probably come up with more nuanced error handling than either of those two options.

lann: I reverted to making failures non-fatal with additional WARN level logging

src/commands/deploy.rs
lann: Reworked.

src/commands/deploy.rs
lann: > Is it because other consumers of WKP presume that the URL does not have the trailing slash?

Yes, it is used directly in a `.strip_prefix(WELL_KNOWN_PREFIX)` in `spin-http`. The code is either awkward here or awkward there.

itowlson (approved):

fibonacci1729 (approved):

lann:
Waiting on cloud release.

Issue #708, opened 2022-08-22
GitHub's Ubuntu 18 image is being deprecated


lann:
https://github.com/actions/runner-images/issues/6002

We saw an intentional "brown-out" today which broke the release build. It looks like they are ["revising the schedule" on those due to push-back](https://github.com/actions/runner-images/issues/6002#issuecomment-1222477506), but we need to keep track of this as it impacts the Linux distro compatibility of our binary releases.

radu-matei:
We use that for the release job — https://github.com/fermyon/spin/blob/main/.github/workflows/release.yml

Do we need to create the release binary in a container instead?

lann:
> GirHub
:facepalm: 

> Do we need to create the release binary in a container instead?

Yeah it might be that simple. I just saw the error and wanted to make sure this was captured.

vdice:
I've moved this from Investigating to the Backlog in an attempt to bump the priority a bit.  Looks like we're still using the `ubuntu-18.04` image in our [release.yml](https://github.com/fermyon/spin/blob/main/.github/workflows/release.yml#L17) and it looks like the target date for dropping support is 4/1/23.

vdice:
@FrankYang0529 thanks for investigating and testing out the udpate to 22.04!  I think what we need to check here is that the spin artifact built from the updated pipeline runs as intended on each platform.

lann:
I don't know of anything blocking an upgrade now.

FrankYang0529:
I tested spin with ubuntu 22.04 on EC2 and it looks fine, so I create a PR for it.

```bash
$ sudo apt update
$ sudo apt install build-essential -y
$ curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
$ rustup target add wasm32-wasi
$ git clone https://github.com/fermyon/spin && cd spin
$ cargo build --all-features --release --features openssl/vendored
$ sudo mv target/release/spin /usr/local/bin/
$ cd ~
$ spin --version
spin 0.6.0 (3b482ba 2022-10-27)
$ spin templates install --git https://github.com/fermyon/spin
$ spin new http-rust myapp && cd myapp
$ spin build
$ spin up
$ curl -i http://localhost:3000
HTTP/1.1 200 OK
foo: bar
content-length: 14
date: Wed, 02 Nov 2022 15:09:58 GMT

Hello, Fermyon

```

lann:
[...]

PR #871, opened 2022-11-02
ci: update release workflow ubuntu version

FrankYang0529:
resolve https://github.com/fermyon/spin/issues/708

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         config:
           - {
-              os: "ubuntu-18.04",
+              os: "ubuntu-22.04",
               arch: "amd64",
               extension: "",
               # Ubuntu 22.04 no longer ships libssl1.1, so we statically


vdice (approved):
Thank you @FrankYang0529!  LGTM.

lann:
I'm pretty sure this will have broken release binaries on older Linux distros like Ubuntu 18.04 (which is still under LTS). We can talk about what kind of support we want but this was part of the reason for remaining pinned to the old version.

FrankYang0529:
Oh! Sorry! I didn't know this doesn't work on Ubuntu 18.04. I moved the binary from 22.04 to 18.04. It doesn't work. @vdice do we want to revert this? Thanks.

```
# Ubuntu 18.04
$ spin --version
spin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by spin)
spin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by spin)
spin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by spin)
spin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by spin)
spin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by spin)
```

PR #884, opened 2022-11-07
Improve error message when app needs to be built

etehtsea:
### Context
- Replaces `anyhow` with `thiserror` inside `spin_publish` library. Tried to preserve the original behavior.
- Improves UX when app needs to be built. Please see the first example in the comparison list.

#### Project needs to be built

Before:

```
spin/examples/http-rust $ /tmp/spin bindle prepare --staging-dir /tmp/http-rust
Error: Failed to expand 'spin.toml' to a bindle

Caused by:
    0: Failed to convert components to Bindle format
    1: Failed to get parcel id for '<redacted>spin/examples/http-rust/target/wasm32-wasi/release/http_rust.wasm'
    2: No such file or directory (os error 2)
```

After (`spin bundle prepare`):

```
spin/examples/http-rust 859-publish $ ../../target/release/spin bindle prepare --staging-dir /tmp/http-rust
Error: Missing build artifact: '<redacted>/spin/examples/http-rust/target/wasm32-wasi/release/http_rust.wasm'

Please try to run `spin build` first
```

After (`spin deploy`):

```
 ../../target/release/spin deploy
Error: Missing build artifact: '<redacted>/spin/examples/http-rust/target/wasm32-wasi/release/http_rust.wasm'

Please try to run `spin build` first

Learn more at https://developer.fermyon.com/cloud/faq
```

After (`spin bindle push`):
```
 ../../target/release/spin bindle push --bindle-server http://localhost:8080
Error: Missing build artifact: '<redacted>/spin/examples/http-rust/target/wasm32-wasi/release/http_rust.wasm'

Please try to run `spin build` first
```
#### Bindle server is unavailable

Before:

```
/tmp/spin bindle push --bindle-server http://localhost:8000/v1
Error: Failed to push bindle to server

Caused by:
    0: Failed to push bindle from '/tmp/.tmpaDtdJm' to server at 'http://localhost:8000/v1'
    1: Error creating request
    2: error sending request for url (http://localhost:8000/v1/_i): error trying to connect: tcp connect error: Connection refused (os error 111)
    3: error trying to connect: tcp connect error: Connection refused (os error 111)
    4: tcp connect error: Connection refused (os error 111)
    5: Connection refused (os error 111)
```

After:

```
../../target/release/spin bindle push --bindle-server http://localhost:8000/v1 
Error: Failed to push bindle from '/tmp/.tmpxJ6f3N' to the server at 'http://localhost:8000/v1'

Caused by:
    0: Error creating bindle client
[...]

PR #890, opened 2022-11-09
Get Wasm modules from Web URLs

### LINKED ISSUE #393: Grab `.wasm` binaries from the web

itowlson:
Fixes #393.

Currently a proof of concept, needs tidying, testing and error handling, plus re-assessing against the requirements in #610 (e.g. should support SHA512, redirects, caching, `file:` URL support, making the digest optional).

Also note this uses a `url/digest` object instead of overloading the string form of `component.source` as proposed in #610.  This isn't a breaking change and is similar to the way we specify parcel sources.

I ported the `bartholomew` component of `fermyon/developer` to exercise it and this is what it ends up looking like:

```
[[component]]
source = { url = "https://github.com/fermyon/bartholomew/releases/download/v0.6.0/bartholomew.wasm", digest = "sha256:b64bc17da4484ff7fee619ba543f077be69b3a1f037506e0eeee1fb020d42786" }
id = "bartholomew"
files = [ "content/**/*" , "templates/*", "scripts/*", "config/*", "shortcodes/*"]
[component.trigger]
route = "/..."
```

I also deployed it to cloud so it looks like it works correctly with bindling, but again, that needs more (and more methodical) testing.

### DIFF
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -10,20 +10,27 @@ pub mod config;
 #[cfg(test)]
 mod tests;
 
-use std::{path::Path, str::FromStr};
+use std::{
+    path::{Path, PathBuf},
+    str::FromStr,
+};
 
 use anyhow::{anyhow, bail, Context, Result};
 use futures::future;
+use itertools::Itertools;
 use outbound_http::allowed_http_hosts::validate_allowed_http_hosts;
 use path_absolutize::Absolutize;
+use reqwest::Url;
 use spin_manifest::{
     Application, ApplicationInformation, ApplicationOrigin, CoreComponent, ModuleSource,
     SpinVersion, WasmConfig,
 };
 use tokio::{fs::File, io::AsyncReadExt};
 
-use crate::bindle::BindleConnectionInfo;
+use crate::{bindle::BindleConnectionInfo, digest::bytes_sha256_string};
 use config::{RawAppInformation, RawAppManifest, RawAppManifestAnyVersion, RawComponentManifest};
+
+use self::config::FileComponentUrlSource;
 
 /// Given the path to a spin.toml manifest file, prepare its assets locally and
 /// get a prepared application configuration consumable by a Spin execution context.
@@ -188,6 +195,17 @@ async fn core(
             let name = format!("{}@{}", bindle_id, parcel_sha);
             ModuleSource::Buffer(bytes, name)
         }
+        config::RawModuleSource::Url(us) => {
+            let source = UrlSource::new(&us)
+                .with_context(|| format!("Can't use Web source in component {}", id))?;
+
+            let bytes = source
+                .get()
+                .await
+                .with_context(|| format!("Can't use source {} for component {}", us.url, id))?;
+
+            ModuleSource::Buffer(bytes, us.url)
+        }
     };
 
     let description = raw.description;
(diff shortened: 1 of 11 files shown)


crates/loader/src/assets.rs
```diff
@@ -90,6 +90,15 @@ pub fn file_sha256_digest_string(path: impl AsRef<Path>) -> std::io::Result<Stri
     Ok(digest_string)
 }
 
+/// Return the hex-encoded SHA256 digest of the given bytes.
+pub fn sha256_digest_string(data: &[u8]) -> String {
+    let mut sha = sha2::Sha256::new();
+    sha.update(data);
+    let digest_value = sha.finalize();
+    let digest_string = format!("{:x}", digest_value);
+    digest_string
```
lann, line 99: This doesn't seem any less clear to me:

```suggestion
    format!("{:x}", Sha256::digest(data))
```

crates/loader/src/assets.rs
```diff
@@ -90,6 +90,15 @@ pub fn file_sha256_digest_string(path: impl AsRef<Path>) -> std::io::Result<Stri
     Ok(digest_string)
 }
 
+/// Return the hex-encoded SHA256 digest of the given bytes.
+pub fn sha256_digest_string(data: &[u8]) -> String {
```
lann, line 94: This could be used in this file above at line 80 (and a few other places if the dependencies made sense).

At what point do we add a `spin-util` crate? :grimacing:

crates/loader/src/local/mod.rs
```diff
@@ -188,6 +189,45 @@ async fn core(
...
+        config::RawModuleSource::Url(us) => {
+            let url = reqwest::Url::parse(&us.url)
+                .with_context(|| format!("Invalid source URL {} in component {}", us.url, id))?;
+            if url.scheme() != "https" {
+                anyhow::bail!(
+                    "Invalid URL scheme {} in component {}: must be HTTPS",
+                    url.scheme(),
+                    id
+                );
+            }
```
lann, line 201: I'm a bit ambivalent about this one, but its there :shrug: 
```suggestion
              anyhow::ensure!(
                  url.scheme() == "https",
                  "Invalid URL scheme {} in component {}: must be HTTPS",
                  url.scheme(),
                  id
              );
```

crates/loader/src/local/mod.rs
```diff
@@ -188,6 +189,45 @@ async fn core(
...
+
+            let (digest_format, digest_text) = us.digest.split_once(':').ok_or_else(|| {
+                anyhow!(
+                    "Invalid digest string '{}' in component {}: must be sha256:...",
+                    us.digest,
+                    id
+                )
+            })?;
+            if digest_format != "sha256" {
```
lann, line 210: It's a bit messy to have this conditional format logic split between here and `verify_digest`. Perhaps simplify both to only support sha256 (YAGNI) or move all the logic to an e.g. `enum ContentDigest { Sha256(String) }` with a `TryFrom<String>`?

crates/publish/src/expander.rs
```diff
@@ -113,6 +113,24 @@ fn bindle_component_manifest(
...
+                if format == "sha256" {
+                    text.to_owned()
+                } else {
+                    anyhow::bail!(
+                        "Invalid digest format '{}' for '{}': must be sha256",
+                        format,
+                        local.id
+                    )
+                }
+            }
```
lann, line 132: We really ought to share logic between `loader` and `publish`...

lann (commented):
We really need a way to share this digest logic. Much as we may hate to see it, maybe a `spin-common` crate would be worthwhile, with modules for `digest`, `sloth`, and a couple other things we have duplicated around?

crates/loader/src/assets.rs
etehtsea: This is funny coincidence, but I'm currently trying to extract similar stuff from my draft [PR](https://github.com/fermyon/spin/pull/884).

https://github.com/etehtsea/spin/commit/d9e0a5498e4e82ebdea930dcf3c15b880526d9c5

My two cents about spin-util:
there is an opinion that util-like libraries is a sort of "anti-pattern" where all the code "I don't know where to put it" is living, but I don't have any better idea at the moment.

Other candidates might be:
- `app_dir` which gets parent of spin.toml and fails with an error or panics when there is none.
- `absolutize`/`canonicalize` (wrappers around std/path-absolutize/dunce)

crates/loader/src/assets.rs
itowlson: TIL (or maybe I already L-ed it but forgot) - thanks!

crates/publish/src/expander.rs
itowlson: Totally agree.  There's a _lot_ of tidying to do here, some of which will slot in more neatly with the improvements @etehtsea already has in progress.

itowlson:
Looking at the stuff suggested in #610:

* SHA512 digests - can add later, nonbreaking
* follow redirects - can add later, nonbreaking
* caching - can add later, nonbreaking
* file: URLs - can add later, nonbreaking
* making the digest optional - can add later ~BUT WON'T~, nonbreaking

So my suggestion is we consider going with this as it stands, and iterate on it in main rather than holding up the PR (if and when the PR is otherwise acceptable of course!).

Note this will still need a rebase on top of #893.

itowlson:
Integration test failing on Windows, think it worked before so will check back tomorrow to see if it's a flake.

lann (approved):

PR #892, opened 2022-11-09
Require raw app manifest to be valid

etehtsea:
Previously, raw app manifest was validated explicitly during the app publishing/load. Running validations automatically in the loader looks like a safer approach.

From the user's perspective, from now on, the invalid manifest will make the build to fail (as well as `spin up`, `spin bindle prepare/push`, `spin deploy`):

```
$ spin build
Error: One or more allowed_http_hosts entries was invalid: ftp://some-random-api.ml
isn't a valid host or host:port string example.com/wib/wob contains a path, should
be host and optional port only
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4258,7 +4258,6 @@ dependencies = [
  "futures",
  "itertools",
  "mime_guess",
- "path-absolutize",
  "reqwest",
  "semver 1.0.14",
  "serde",
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -10,9 +10,11 @@ pub mod config;
 #[cfg(test)]
 mod tests;
 
-use std::{path::Path, str::FromStr};
+use std::{borrow::Cow, path::Path, str::FromStr};
 
+use crate::bindle::BindleConnectionInfo;
 use anyhow::{anyhow, bail, Context, Result};
+use config::{RawAppInformation, RawAppManifest, RawAppManifestAnyVersion, RawComponentManifest};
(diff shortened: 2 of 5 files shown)


itowlson:
I think we used to have some manifest validations that ran during build, and it was quite annoying, because you wanted to check that stuff compiled and Spin would go "no not until you fix this thing you don't need yet."  I can't find the issue now though!

itowlson:
That said I would _really really_ like nothing above the loader to have to deal with the raw on-disk format.  And I do want the fully loaded form to be known valid.  So I'm not sure how to deal with these conflicting impulses.  I guess we could have a `BuildSettings` form that we could parse into that ignored everything else, but maybe that feels kinda heavyweight...?

etehtsea:
> I think we used to have some manifest validations that ran during build, and it was quite annoying, because you wanted to check that stuff compiled and Spin would go "no not until you fix this thing you don't need yet." I can't find the issue now though!

Currently, `allowed_http_hosts` is the only validation, so it should be relatively easy to comment it out when you don't need it yet.

itowlson:
It's relatively easy but it _breaks flow_.  Especially because many developers (me. It was me) will, rather than comment it out, get distracted with "oh I'll just quickly figure out what's wrong, it won't take more than a minute."  And those strong-willed people who _do_ comment it out will then forget (it's amazing what an hour of wrestling with the borrow checker can do to one's mental stack), run their project, and get weird errors when their HTTP requests are blocked.

I get the attraction of validating right up front, I really do, but there's a reason we don't do the runtime validation(s) on build.  Sorry.

PR #893, opened 2022-11-09
chore: Reuse sha256 digest from spin_loader

etehtsea:
Extracted from #884.
Also, it might worth checking the discussion - https://github.com/fermyon/spin/pull/890#discussion_r1018013724


crates/loader/src/assets.rs
```diff
@@ -76,18 +76,22 @@ pub(crate) fn component_dir(id: &str) -> String {
...
-    let mut file = std::fs::File::open(path)?;
-    let mut sha = sha2::Sha256::new();
-    std::io::copy(&mut file, &mut sha)?;
-    let digest_value = sha.finalize();
-    let digest_string = format!("{:x}", digest_value);
-    Ok(digest_string)
+pub async fn file_sha256_digest_string(path: impl AsRef<Path>) -> std::io::Result<String> {
+    let bytes = tokio::fs::read(&path).await?;
+
+    Ok(bytes_sha256_digest_string(&bytes))
```
itowlson, line 94: The reason for using `copy` was, I think, that the data could be streamed from the file into the hash algorithm, without needing to load a potentially large file into memory.  Unless `copy` does a full load, I'd suggest reverting to the earlier implementation even if it does result in a minor divergence of the bytes and file functions.

crates/loader/src/assets.rs
```diff
@@ -76,18 +76,22 @@ pub(crate) fn component_dir(id: &str) -> String {
     // if the asset paths are also long. Longer term, consider an alternative approach where
     // we use an index or something for disambiguation, and/or disambiguating only if a clash is
     // detected, etc.
-    let id_sha256 = format!("{:x}", Sha256::digest(id));
+    let id_sha256 = bytes_sha256_digest_string(id.as_bytes());
     format!("{}_{}", UNSAFE_CHARACTERS.replace_all(id, "_"), id_sha256)
 }
 
+/// Return the hex-encoded SHA256 digest of the given bytes.
```
itowlson, line 83: Would it be worth moving these to a `digest.rs` or `sha256.rs` file instead of having them in `assets.rs`?  As they are used for bindling manifests and modules, they're not specific to assets.

itowlson (approved):
LGTM with a couple of suggestions - thanks for this much-needed piece of tidying!

crates/loader/src/assets.rs
etehtsea: You're right, I've checked it and it seems that hash algo doesn't need the whole file in memory. Reverted, thanks!

crates/loader/src/assets.rs
etehtsea: Done, thanks!

crates/loader/src/digest.rs
```diff
@@ -0,0 +1,25 @@
...
+pub fn file_sha256_string(path: impl AsRef<Path>) -> std::io::Result<String> {
+    let mut file = std::fs::File::open(path)?;
+    let mut sha = sha2::Sha256::new();
+    std::io::copy(&mut file, &mut sha)?;
+    let digest_value = sha.finalize();
+
+    Ok(to_hex_string(digest_value))
+}
+
+fn to_hex_string(digest_value: impl std::fmt::LowerHex) -> String {
```
itowlson, line 23: Oh nice!

itowlson (approved):
Nice - thanks!

PR #894, opened 2022-11-10
chore: Use workspace inheritence

lann:
This uses the "workspace inheretance" features introduced in Cargo 1.64 to deduplicate dependency specifications and package info across the workspace.
    
Add package 'license' and 'rust-version' to the workspace root.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,14 +1,21 @@
 [package]
 name = "spin-cli"
+version = { workspace = true }
+authors = { workspace = true }
+edition = { workspace = true }
+license = "Apache-2.0 WITH LLVM-exception"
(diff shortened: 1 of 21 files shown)


Cargo.toml
```diff
@@ -1,14 +1,21 @@
 [package]
 name = "spin-cli"
+version = { workspace = true }
+authors = { workspace = true }
+edition = { workspace = true }
+license = "Apache-2.0 WITH LLVM-exception"
```
lann, line 6: Lifted from wasmtime, which uses the same license we do.

dicej (approved):
Thanks for doing this.  Now I won't have to use `sed` so much when upgrading dependencies.

PR #895, opened 2022-11-11
Bartholomew template

itowlson:
My thought here is to provide an "add component" version of this so that people can easily add a content-managed component to their Spin applications.  This will hopefully make a nice demo of how Spin facilitates composing applications out of Wasm modules, whether custom or off-the-shelf.

Known issues:

* Does not play nicely with HTTP bases other than root.
* The static fileserver URL currently points to a test release in my fork.

Open question:

* Does this even belong in the Spin default template set, or should it go in the Bartholomew repo and require separate installation?  Or should we avoid having a Spin template that would need to be maintained alongside the GitHub template?  (If we don't want this here or at all, I would at least like to offer a static file server component in the default set - it's a widely useful thing.)

### DIFF
--- a/templates/bartholomew/content/README.md
+++ b/templates/bartholomew/content/README.md
@@ -1,0 +1,31 @@
+# {{ site-title }}
+
+_Built with (Bartholomew)[https://bartholomew.fermyon.dev/], (Fermyon)[https://www.fermyon.com]'s open source micro-CMS._
+
+## Directory Structure:
+
+- `config/site.toml`: The main configuration file for the site. You should edit this.
+- `content/`: Your markdown files go in here.
+- `scripts/` (advanced): If you want to write your owh Rhai scripts, they go here.
+- `spin.toml`: The configuration file for the Spin application.
+- `static/`: Static assets like images, CSS, and downloads go in here.
+- `templates/`: Your handlebars templates go here. 
+
+## Running the site
+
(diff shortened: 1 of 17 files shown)


templates/bartholomew/content/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Running the site
+
+To start this site, run `spin up`` from this directory.  (If you don't have Spin, you can install it from (here)[https://developer.fermyon.com/spin/install].)
+
+```console
+$ spin up --follow-all
+spin up --follow-all
+Serving HTTP on address http://127.0.0.1:3000
+Available Routes:
+  bartholomew: http://127.0.0.1:3000 (wildcard)
```
itowlson, line 23: TODO: update these with the HTTP base

templates/bartholomew/content/static/example.txt
```diff
@@ -0,0 +1,4 @@
+This is an example of a static text file.
+
+If you can see this in your browser at `http://localhost:3000/static/example.txt`, then you have
```
itowlson, line 3: TODO: show the HTTP base here (but might need some magic to manage the right number of slashes in all circumstances).

itowlson:
Closing as stale: the GitHub template repository does everything this does that people are likely to care about, and has the huge advantage of working correctly.

PR #887, opened 2022-11-08
Introduce spin_publish::prepare_bindle

etehtsea:
Extracted the straightforward part from https://github.com/fermyon/spin/pull/884 to reduce the scope of it.

PR #896, opened 2022-11-11
Introduce spin_loader::local::{absolutize, parent_dir}

etehtsea:
Extract repetitive patterns into fns. Probably this isn't the best
place to put them, but there is no decision about the introduction
of some util-like library.

Rebased against https://github.com/fermyon/spin/pull/887 which is approved but not merged yet
Please review these commits:
 - c0a9b8db25bb42d89311735c554a8df1c012435d
 - e51367d88b2178d2219c64c2b753e6aa138f80d8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -535,9 +535,6 @@ dependencies = [
  "semver 1.0.14",
  "serde",
  "serde_json",
- "spin-loader",
- "spin-publish",
- "tempfile",
  "tokio",
  "tokio-util 0.7.4",
  "toml",
@@ -4263,7 +4260,6 @@ dependencies = [
  "semver 1.0.14",
  "serde",
  "spin-loader",
- "tempfile",
  "tokio",
  "toml",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -529,7 +529,6 @@ dependencies = [
  "itertools",
  "log",
  "mime_guess",
- "path-absolutize",
  "regex",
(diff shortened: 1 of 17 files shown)


itowlson (commented):
I have mixed feelings about this!  I like getting all the ceremony out of the application code.  But I'm worried that the error messages provide less direction to help the user fix a problem.  For example, I think this changes "can't get parent of manifest file '/'" to "can't get parent of /", which no longer tells the user why Spin was looking for '/'.  Can we continue to provide that info somehow?  Or are these cases so pathological that they are unlikely ever to turn up in real life?

etehtsea:
Rationale for this one was to enhance the error message to always provide the path and handle this consistently throughout the code. In the majority of cases the path wasn't in the error msg at all (https://github.com/fermyon/spin/pull/896/files#diff-b18d9ef02a8395df7dcf8cc2d999cc16fe1fefe5b2151dfbd763f8b65e0c778dL87, https://github.com/fermyon/spin/pull/896/files#diff-4380556bac0572cbce80ec3a061ef8c1855cf7068b8c5130d4817cfd41e45d0fL47)

itowlson:
That sounds like a worthy goal.  Certainly having the path consistently in the message is a high priority, and these aren't things that should fail under normal circumstances.  Let's go ahead then, and if we run into issues where things are failing and are hard to diagnose, we can revisit.  Thanks!

itowlson (approved):

Issue #869, opened 2022-11-01
CI: Runner disk usage


vdice:
With the growing list of test dependencies, it appears we're running into the default 14GB disk size limit of the default GH runner.

Ref https://github.com/fermyon/spin/pull/798#issuecomment-1298440711

One initial question: Are we getting bit by our use of the rust cache?  Are we able to clear it?  Even with the growing list of dependencies for e2e testing (bindle, hippo, soon Vault), I wouldn't expect to exceed 14GB of disk.

Failing that, do we want to try to get creative in our tests to just-in-time install deps and then remove them after the corresponding tests have completed?  (eg install bindle, run tests that require bindle, remove bindle, move on to the next tests that require external deps)  Or would this get unwieldy/fragile?

If other methods like the above don't help, do we want to:
1. Pay $$ for [larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#machine-specs-for-larger-runners)
2. Use self-hosted runners with deps pre-installed
3. Other?

itowlson:
This is becoming urgent - PRs are blocking on CI runs that are failing due to out of space.

Is there a short-term fix we can get in place?  E.g. could we pay $$ for a month and buy some time to reorganise/refine?  Or is the cost prohibitive/would it require significant setup?

vdice:
I've cleaned up the saved caches (ranging from 1.5-2.5GB), any one of which may have been mounted into the build agent.  Let's see if builds can make further progress.

As a potential next step, what do we think about breaking up the current do-all build.yml into more workflows?  Each workflow we break out means a new runner host to spread the disk usage load.  I could see having a single compile/build workflow, a unit-test workflow and an integration/e2e workflow.  Thoughts?

itowlson:
Nice one @vdice - things looking back on track!  Thanks!

Do you have a sense for how long this buys us - e.g. days, weeks, months?

vdice:
Yay!

Alas, I don't... and I see that we've already gone over the cache max with the recent CI (re-)runs 😄.  We can view a list here: https://github.com/fermyon/spin/actions/caches

It doesn't look like GitHub automagically deletes caches for PR merge branches once the PR itself is merged; so to clear up space, we'd need to manually delete those.

I'm still unclear on the interaction between the caches and runner disk usage... needs more researching.

rajatjindal:
Note: following numbers were captured when running the action with ubuntu only on a fork in my user account. 

[...]

PR #902, opened 2022-11-15
chore: Upgrade wasmtime to 2.0.2

lann:
This required forking wit-bindgen to upgrade it's wasmtime dependency while maintaining the existing Wasm interface ABI.


dicej (approved):

radu-matei:
ref #869 for the build error

PR #903, opened 2022-11-16
feat: allow component selector during spin up

rawkode:
I find myself often wanting to iterate on a single component within my Spin application, instead of building them all; this can take considerable time as my applicationg grows.

This selector allows a single, or multiple, uses of --component/-c to limit the components specified by the user.


src/commands/up.rs
```diff
@@ -90,6 +90,10 @@ pub struct UpCommand {
     /// All other args, to be passed through to the trigger
     #[clap(hide = true)]
     pub trigger_args: Vec<OsString>,
+
+    /// Only run a subset of the components
+    #[clap(long = "component", short = 'c')]
+    pub components: Vec<OsString>,
```
itowlson, line 96: This should probably be `String` rather than `OsString` - I don't think it's going to end up being passed back to the OS?.

crates/loader/src/local/mod.rs
```diff
@@ -40,12 +41,13 @@ pub async fn from_file(
     app: impl AsRef<Path>,
     base_dst: impl AsRef<Path>,
     bindle_connection: &Option<BindleConnectionInfo>,
+    components: Vec<OsString>,
```
itowlson, line 44: `components` is a bit undescriptive here - aren't the "components" in the app file?  I realise it's more verbose but maybe a name like `include_components` or `include_only` would give the reader a clearer idea of what the parameter does?

crates/loader/src/local/mod.rs
```diff
@@ -125,26 +130,38 @@ pub fn validate_raw_app_manifest(raw: &RawAppManifestAnyVersion) -> Result<()> {
     Ok(())
 }
 
+fn filter_components(components: &Vec<OsString>, c: &RawComponentManifest) -> bool {
```
itowlson, line 133: Suggest `include_component`.  `filter` implies a set operation, such as filtering the list of RawComponentManifests to just the included ones..

itowlson, line 133: Maybe just pass the component id here?  It doesn't really need to know the whole manifest.

src/commands/up.rs
```diff
@@ -127,7 +131,13 @@ impl UpCommand {
                     .as_deref()
                     .unwrap_or_else(|| DEFAULT_MANIFEST_FILE.as_ref());
                 let bindle_connection = self.bindle_connection();
-                spin_loader::from_file(manifest_file, &working_dir, &bindle_connection).await?
+                spin_loader::from_file(
+                    manifest_file,
+                    &working_dir,
+                    &bindle_connection,
+                    self.components,
```
itowlson, line 138: What is the error behaviour if all components are excluded?  For example, if I have an app with `hello` and `goodbye` components, and pass `-c goobdye`?  Does it tell me there are no components, or does it report specifically that no components matched the include arguments?

Thinking about it, passing an id that doesn't exist in the manifest should probably be an error.  I can't think of any situation in which that would be what the user wanted...?

crates/loader/src/local/tests.rs
```diff
@@ -252,7 +252,7 @@ async fn test_invalid_url_in_allowed_http_hosts_is_rejected() -> Result<()> {
 
     let temp_dir = tempfile::tempdir()?;
     let dir = temp_dir.path();
-    let app = from_file(MANIFEST, dir, &None).await;
+    let app = from_file(MANIFEST, dir, &None, vec![]).await;
```
itowlson, line 255: Should we have a test for when the include list is non-empty?  (And for error cases.)

src/commands/up.rs
```diff
@@ -90,6 +90,10 @@ pub struct UpCommand {
     /// All other args, to be passed through to the trigger
     #[clap(hide = true)]
     pub trigger_args: Vec<OsString>,
+
+    /// Only run a subset of the components
+    #[clap(long = "component", short = 'c')]
```
itowlson, line 95: Looks like bindle loading ignores this flag, which makes sense given this is aimed at the local dev inner loop, but means we should have a clap attribute to prevent it.

itowlson (changes requested):

src/commands/up.rs
```diff
@@ -98,6 +98,10 @@ pub struct UpCommand {
     /// All other args, to be passed through to the trigger
     #[clap(hide = true)]
     pub trigger_args: Vec<OsString>,
+
+    /// Only run a subset of the components
+    #[clap(long = "include_components", short = 'c')]
+    pub include_components: Vec<String>,
```
itowlson, line 104: I may be misreading this but it looks like this doesn't apply if the app is being loaded from a bindle, only if it's being loaded from a local file.  Could we add a `conflicts_with = BINDLE_ID_OPT` to the clap attribute please?

itowlson (changes requested):
Looks good - thanks in particular for improving the error messages.  Just one change to prevent surprises (in a little-used path, but still) and then we are good to go!

src/commands/up.rs
rawkode: Ah. I couldn’t find that attribute. No problem.

itowlson (approved):

PR #904, opened 2022-11-16
 chore(GitHub): Refactor build/test workflow 

endocrimes:
Our current builds are currently unstable because we run out of disk
space on the default GitHub Actions builders.

While we could start trying to be clever and purge various resources
between steps - it's simpler to refactor the build into mutliple
independent steps.

For now we're going to use a "shared-for-the-platform" cache based on
the hash of the Cargo.lock. I plan to follow up with a little more
rigerous of a cache-key system that can also naturally expire a little
sooner to ensure everything is still buildable - but for now, lets get
spin building consistently again.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,9 +14,82 @@ on:
 env:
   CARGO_TERM_COLOR: always
 jobs:
-  build:
+  lint-rust:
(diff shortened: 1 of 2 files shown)


radu-matei (approved):
Yay, thank you for the fix!

LGTM

PR #908, opened 2022-11-17
Fail if template references unknown filter

itowlson:
The template system is intentionally forgiving of parse errors, because a template might contain text that confuses the Liquid parser but which an author wants to be included anyway.  However, this can disguise genuine authoring errors.  This particular one bit me when accidentally using the redirect template with a previous version of Spin - it appeared to succeed but actually failed on a new filter and dumped the unsubstituted template into my project.  With this fix, I need never be confused again.

Longer term, we perhaps need more robust mechanisms for template authors to exclude files from parsing, after which we can banhammer actual parse failures.

### DIFF
--- a/crates/templates/src/renderer.rs
+++ b/crates/templates/src/renderer.rs
@@ -1,4 +1,5 @@
 use anyhow::anyhow;
+use lazy_static::lazy_static;
 use std::{collections::HashMap, path::PathBuf};
 
 use crate::writer::{TemplateOutput, TemplateOutputs};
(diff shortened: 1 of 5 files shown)


crates/templates/src/renderer.rs
```diff
@@ -99,3 +112,31 @@ fn string_from_bytes(bytes: &[u8]) -> Option<String> {
...
 }
+
+enum TemplateParseFailure {
+    UnknownFilter(String),
+    Other(liquid::Error),
+}
+
+lazy_static! {
+    static ref UNKNOWN_FILTER: regex::Regex =
+        regex::Regex::new("requested filter=(\\S+)").expect("Invalid unknown filter regex");
```
kate-goldenring, line 123: is this `requested filter` string a guaranteed prefix of `liquid::Error` messages?

itowlson: It's what it currently reports when it finds a filter that isn't registered in the parser.  Hopefully no other errors contain that string.  Unfortunately I think this is the best we can do given Liquid's stringy error API.

crates/templates/src/renderer.rs
```diff
@@ -99,3 +112,31 @@ fn string_from_bytes(bytes: &[u8]) -> Option<String> {
...
+}
+
+fn understand_liquid_error(e: liquid::Error) -> TemplateParseFailure {
+    let err_str = e.to_string();
+    let err_lines = err_str.lines().collect_vec();
+
+    // They should use typed errors like we, er, don't
+    match err_lines.first() {
```
dicej, line 131: Nit: no need to collect the lines into a `Vec` if you only care about the first one.
```suggestion
    match err_str.lines().next() {
```

dicej (approved):

crates/templates/src/renderer.rs
itowlson: ooh thanks!

crates/templates/src/renderer.rs
itowlson: Oh also not a prefix, this is just the substring to look for to pull out the unrecognised name.

fibonacci1729 (approved):

PR #911, opened 2022-11-20
Restore sloth warning around module build/load

### LINKED ISSUE #909: Progress indicator when Wasm modules take a long time to load
The old engine module used to have a 'sloth warning'  so that people didn't have a blank screen when modules took a long time to load (particularly an issue in debug builds).  It would be good to bring this back, though better placed in `spin up`  (if that gives us enough visibility).

itowlson:
Fixes #909.

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -3,6 +3,10 @@ use std::path::PathBuf;
 use anyhow::{Context, Result};
 use clap::{Args, IntoApp, Parser};
 use serde::de::DeserializeOwned;
+use tokio::{
+    task::JoinHandle,
+    time::{sleep, Duration},
+};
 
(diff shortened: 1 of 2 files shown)


crates/trigger/src/cli.rs
```diff
@@ -167,3 +173,35 @@ where
...
+
+#[cfg(debug_assertions)]
+async fn warn_slow_wasm_build() {
+    sleep(Duration::from_millis(SLOTH_WARNING_DELAY_MILLIS)).await;
+    println!("This is a debug build - preparing Wasm modules might take a few seconds");
+    println!("If you're experiencing long startup times please switch to the release build");
+    println!();
+}
+
+#[cfg(not(debug_assertions))]
```
lann, line 202: These can apply to code blocks instead of whole functions.

lann (approved):

PR #913, opened 2022-11-22
Skip validating unused manifest fields in spin build

### LINKED ISSUE #473: `spin build` trigger validation messages are not explained
I fouled up the Redis template so that `trigger` had a `base` field instead of `address`.

When I ran `spin build`, it said `Error: missing field `address``.

As my brain was in the "I am compiling" mindset, I assumed there was something wrong in the code, though I was puzzled by the lack of compiler context.  It didn't occur to me for a while that Spin was actually validating the trigger configuration.

Could we either ignore irrelevant fields during build, or provide a contextful message such as "Invalid trigger configuration" or "Invalid manifest file", ideally saying more about where the error occurred than just the name of what in this case was a nested field?  Even calling it `trigger.address` would have helped.

itowlson:
Fixes #473.

The approach taken here is that we don't care if other parts of the manifest are valid, as long as we have the info we need to do the build.  In terms of the original issue, the build won't fail because you have a missing field in your trigger.  The downside of this is that it results in some duplication of the parts of the manifest schema.

An alternative approach would be to keep the existing loader but just improve the error reporting around manifest load failures.  Despite the duplication, I prefer skipping validation, because if I'm in coding flow I don't want to have to take time out to fix some random other field.  But happy to have that discussion.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4006,8 +4006,11 @@ version = "0.6.0"
 dependencies = [
  "anyhow",
  "futures",
+ "serde",
  "spin-loader",
  "subprocess",
+ "tokio",
+ "toml",
  "tracing",
 ]
 
--- a/crates/build/Cargo.toml
+++ b/crates/build/Cargo.toml
@@ -7,6 +7,9 @@ edition = { workspace = true }
 [dependencies]
 anyhow = "1.0.57"
 futures = "0.3.21"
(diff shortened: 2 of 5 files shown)


dicej (approved):
The duplication is annoying, but not egregious.

itowlson:
Will accept any review with the word 'egregious' in it.  Thanks!

PR #914, opened 2022-11-23
chore: Cleanup unused deps

etehtsea:
I've been curious about why spin is depending on `wayland-*` libs, so when I checked it I found that these are dependencies of `copypasta` lib which isn't used. Then I ran [cargo-udeps](https://github.com/est31/cargo-udeps) to clean some more.


itowlson (approved):
Hah, funny - I was looking at build output just yesterday and thinking "why the heck do we have Wayland deps in here."  Thanks for going through this!

PR #915, opened 2022-11-23
add support for outbound Redis set operations

dicej:
This adds support for the SADD, SREM, and SMEMBERS Redis commands, useful for apps which need to manage groups of unique identifiers (e.g. chat apps).

### DIFF
--- a/crates/abi-conformance/src/outbound_redis.rs
+++ b/crates/abi-conformance/src/outbound_redis.rs
@@ -49,6 +49,33 @@ pub struct RedisReport {
     /// function with the arguments \["127.0.0.1", \["foo"\]\] and expect `ok(0)` as the result.  The host will assert
     /// that said function is called exactly once with the specified arguments.
     pub del: Result<(), String>,
+
+    /// Result of the Redis `SADD` test
+    ///
+    /// The guest module should expect a call according to [`super::InvocationStyle`] with \["outbound-redis-sadd",
+    /// "127.0.0.1", "foo", "bar", "baz"\] as arguments. The module should call the host-implemented
+    /// `outbound-redis::sadd` function with the arguments \["127.0.0.1", "foo", \["bar", "baz"\]\] and expect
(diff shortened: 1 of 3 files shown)


itowlson (approved):
I'm no Redis expert but you have my LGTM for what it's worth...!

dicej:
@lann FYI

crates/abi-conformance/src/outbound_redis.rs
```diff
@@ -49,6 +49,33 @@ pub struct RedisReport {
...
     pub del: Result<(), String>,
+
+    /// Result of the Redis `SADD` test
+    ///
+    /// The guest module should expect a call according to [`super::InvocationStyle`] with \["outbound-redis-sadd",
+    /// "127.0.0.1", "foo", "bar", "baz"\] as arguments. The module should call the host-implemented
+    /// `outbound-redis::sadd` function with the arguments \["127.0.0.1", "foo", \["bar", "baz"\]\] and expect
+    /// `ok(2)` as the result.  The host will assert that said function is called exactly once with the specified
+    /// arguments.
+    pub sadd: Result<(), String>,
```
lann, line 60: :sob:

lann:
This seems fine, but I have to wonder if we'd be better off with a generic "exec" function rather than a function per redis command...

dicej:
> This seems fine, but I have to wonder if we'd be better off with a generic "exec" function rather than a function per redis command...

Yeah, I was thinking the same thing.  Redis a _lot_ of operations.  It's nice to have a bit of static type and arity checking, though.  Maybe there's some kind of machine-readable Redis IDL we could generate WIT files and tests from.

dicej:
I've run `make test-outbound-redis` and manually tested the `tinygo-outbound-redis` example to make sure there are no regressions.  I'm going to go ahead and merge this since set operations are broadly useful.

I do agree that having a separate function for every operation is not sustainable unless it can be automated somehow, so a good follow-up PR would be to add a general-purpose `exec` function to cover the myriad other ops, transactions, etc.

PR #918, opened 2022-11-27
Templates tidying

itowlson:
Reorganising the templates crate so that:

* Modules are less bulky and the code overall is easier to navigate
* Things are more distinctively and specifically named
* The core run flow is clearer
* It is harder to express invalid states

Leaving the individual commits for now in case some of it is deemed unwelcome and we need to cherry-pick the good bits, but should be squashed before merging.

### DIFF
--- a/crates/templates/src/cancellable.rs
+++ b/crates/templates/src/cancellable.rs
@@ -1,0 +1,50 @@
+pub(crate) enum Cancellable<T, E> {
+    Cancelled,
+    Err(E),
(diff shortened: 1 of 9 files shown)


dicej (approved):
I don't have any history with this code, so I'm probably not the best one to review this, but it all looks reasonable to me.

itowlson:
Thanks @dicej - pinged a couple of people who might be more familiar!

PR #923, opened 2022-11-29
make `spin_publish::expander::expand_manifest` public again

dicej:
An earlier commit changed this from `pub` to `pub(crate)`, making it inaccessible to third-party crates.  This makes it public again.


lann (approved):

itowlson:
I believe the intent here was to route all bindle generation through `prepare_bindle`.  What's using `expand_manifest` and should those consumers migrate?

(I'm not against making `expand_manifest` public again, just checking that this is the right solution rather than consumers updating to the consolidated API.  Obviously if all they want is the expanded manifest, or we can't rely on them updating, then good to go.)

PR #925, opened 2022-11-30
Build is not async

itowlson:
Some internals of the `build` crate were declared as async and had machinery to run in parallel.  From what I can tell, they never have been async, so the parallelism/join stuff wasn't doing anything - the builds have always been sequential.  The presence of `async` may be a hangover from during initial development (perhaps the author started out by using `tokio::process::Command`?).

This PR removes the `async` modifier and strips out the machinery for joining after parallel builds.

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -25,13 +25,10 @@ pub async fn build(manifest_file: &Path) -> Result<()> {
         return Ok(());
     }
 
-    let results = futures::future::join_all(
-        app.components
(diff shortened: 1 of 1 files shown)


radu-matei (approved):
Good catch!

I think the desired behavior _is_ for component build steps to run sequentially anyway, so this makes sense.

LGTM

itowlson:
Yes, they print build output as they go, and it would be super confusing to have multiple component build outputs overlapping...!

Issue #901, opened 2022-11-15
Add support for `linux/arm64` releases


mikesir87:
I'm working on creating a Docker container that can use spin to development, but wanting to leverage native architectures as much as possible. On Macs with Docker Desktop, we can leverage both `linux/amd64` and `linux/arm64`. You're building `linux/amd64` as part of your releases, but it would be great to have `linux/arm64` support as well to prevent qemu emulation. Thanks!

itowlson:
Does the `linux/aarch64` build from #926 meet this need, or is there a distinct `linux/arm64` that we'd need to ship?

Issue #293, opened 2022-03-30
ARM Linux release builds


lann:
We are disabling builds that are failing  due to cross-compiling openssl for ARM:

https://github.com/fermyon/spin/runs/5759249651

 I think we can fix that by merging #247 and using `--features=vendored-openssl` _for cross-compiled builds only_. In order for this to work I think we need to remove `--all-features` from various places as well.

mreferre:
@lann is there an ETA for when you will be starting re-producing Arm builds? Not putting pressure, just curious. Thanks.

lann:
I spent a little time looking into this today, but even with the above openssl vendoring and using [`cross`](https://github.com/cross-rs/cross) I hit some pretty nasty-looking bugs cross-compiling `wasmtime-fiber`. I would be interested to hear if anyone is able to get this working by compiling on a native (or emulated) ARM box.

mreferre:
Is there an update on this? I see Spin ships for linux/amd64, windows/amd64, mac/amd64, mac/arm. Looks like linux/arm is the only miss (with the big trend towards more arm in the cloud it looks like a good opportunity for Spin).

lann:
No update yet. Currently the release builds are really more geared toward development environments where ARM Linux is less common. That said, this is mostly blocked on finding a good way to actually perform the cross-arch builds in Github Actions. For some reason cross-compiling for aarch64-apple-darwin "just works" where aarch64-unknown-linux-gnu doesn't.

lann:
```console
$ cargo build --release --target aarch64-unknown-linux-gnu --features openssl/vendored
[...many copies of line below...]
/usr/bin/ld: <PATH>/spin/target/aarch64-unknown-linux-gnu/release/deps/spin-50cb93f1b1f7d4bc.spin.d8cbf558-cgu.0.rcgu.o: Relocations in generic ELF (EM: 183)
/usr/bin/ld: <PATH>/spin/target/aarch64-unknown-linux-gnu/release/deps/spin-50cb93f1b1f7d4bc.spin.d8cbf558-cgu.0.rcgu.o: error adding symbols: file in wrong format
collect2: error: ld returned 1 exit status

mreferre:
Thanks @lann. Out of curiosity I was trying to compile Spin on a Graviton instance but I ended up in a rat-hole where one of the requirements (i.e. bind) is not available on Linux/Arm either and trying to compile bindle gave me a problem etc. etc.

PR #926, opened 2022-11-30
Enable aarch64 in release workflow

jpflueger:
I've been using the `fermyon/installer` repository to run spin applications on a raspberry pi. It looks like there isn't an official release binary for aarch64 yet but there is an open issue #901 to add support for arm64. I found the notes in the release workflow file and was able to generate a release successfully in my fork, [here's the workflow result](https://github.com/jpflueger/hippo/actions/runs/3569728099). Maybe I'm missing something but this does seem to work now.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,16 +23,14 @@ jobs:
               target: "",
               targetDir: "target/release",
             }
-          # TODO: openssl is still a nested dep here.
-          # Re-enable when updated to use vendored openssl or cross-compile setup is updated
-          # - {
-          #     os: "ubuntu-latest",
-          #     arch: "aarch64",
-          #     extension: "",
-          #     extraArgs: "--target aarch64-unknown-linux-gnu",
-          #     target: "aarch64-unknown-linux-gnu",
-          #     targetDir: "target/aarch64-unknown-linux-gnu/release",
-          #   }
+          - {
+              os: "ubuntu-latest",
+              arch: "aarch64",
+              extension: "",
+              extraArgs: "--features openssl/vendored --target aarch64-unknown-linux-gnu",
+              target: "aarch64-unknown-linux-gnu",
+              targetDir: "target/aarch64-unknown-linux-gnu/release",
+            }
           - {
               os: "macos-latest",
               arch: "amd64",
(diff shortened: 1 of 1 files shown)


itowlson (approved):
I have no way of testing it but it looks fine to me (and I know you've tested it).  Unless anyone else is in a position to test then I'd say merge it.

itowlson:
I think this also relates to #293 which talks more about the issues that were hit first time round.

radu-matei (approved):
Have you tested a binary generated from this action on a Linux ARM device, @jpflueger?
I don't have immediate access to one, but if you tried it and it works, the changes LGTM.

jpflueger:
Thanks for that link @itowlson, I wasn't able to find that when looking for prior work. I did make one change from just the commented code to add `--features openssl/vendored` but cargo/rust is pretty new to me so I'm not exactly sure of the implications here. 

Testing more thoroughly now but yesterday I was able to run some commands locally without error. Is there a suite of commands that I should run in my testing? Going through each of the templates now to run them and ensure they work and also going to deploy them to fermyon cloud.

radu-matei:
cc @lann @dicej @fibonacci1729, as this touches OpenSSL and you previously worked on this.

.github/workflows/release.yml
```diff
@@ -86,13 +84,13 @@ jobs:
...
-      #     sudo apt install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
-      #     echo '[target.aarch64-unknown-linux-gnu]' >> ${HOME}/.cargo/config.toml
-      #     echo 'linker = "aarch64-linux-gnu-gcc"' >> ${HOME}/.cargo/config.toml
+      - name: setup for cross-compiled linux aarch64 build
+        if: matrix.config.target == 'aarch64-unknown-linux-gnu'
+        run: |
+          sudo apt update
+          sudo apt install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
+          echo '[target.aarch64-unknown-linux-gnu]' >> ${HOME}/.cargo/config.toml
+          echo 'linker = "aarch64-linux-gnu-gcc"' >> ${HOME}/.cargo/config.toml
```
lann, line 93: It doesn't _seem_ like this should be necessary with `--target` set, but I don't know for sure.

lann (approved):
As long as you test `spin up` (to exercise wasmtime) I think this is pretty straightforward.

fibonacci1729 (approved):

jpflueger:
@itowlson I did just run `spin up` and `spin deploy` with rust and C languages successfully on a linux/arm64 device.

lann:
Feel free to merge. If it breaks CI for `main` we can just revert

.github/workflows/release.yml
jpflueger: I removed these two lines and started seeing what look like similar errors from #293. Here is the latest [workflow run](https://github.com/jpflueger/spin/actions/runs/3597379547/jobs/6059072395).

jpflueger: It looks like this is still necessary. In the failed workflow run the logs seem to indicate that is is still finding part of the x86_64 toolchain. Could we remove this echo by putting this configuration in the repository's `.cargo/config`? Not sure if that's a good idea or not, just thinking out loud.

lann: Let's just leave what we know works for now

lann:
Looks like it worked: https://github.com/fermyon/spin/actions/runs/3633357583/jobs/6130276904

rajatjindal:
> @jpflueger I don't know if there is an existing suite of tests other than the existing integration/e2e tests. I agree with Lann, if `spin up` is working (and I guess `spin deploy` for SSL?), that's fine, and hopefully @rajatjindal's broader test automation can exercise it more systematically in the long term.

added this to the epic where I am tracking the effort.

PR #929, opened 2022-12-01
wagi: Bail early if nothing is written to stdout

lann:
This provides a nicer error message if a user configures a component to use the WAGI executor when it isn't actually using WAGI.

### DIFF
--- a/crates/http/src/wagi.rs
+++ b/crates/http/src/wagi.rs
@@ -2,7 +2,7 @@ mod util;
 
 use std::{io::Cursor, net::SocketAddr};
 
-use anyhow::Result;
+use anyhow::{ensure, Result};
 use async_trait::async_trait;
 use hyper::{
     body::{self},
@@ -143,7 +143,14 @@ impl HttpExecutor for WagiHttpExecutor {
             .or_else(ignore_successful_proc_exit_trap)?;
         tracing::info!("Module execution complete");
 
-        util::compose_response(&stdout_buffer.take())
+        let stdout = stdout_buffer.take();
+        ensure!(
+            !stdout.is_empty(),
+            "The {component:?} component is configured to use the WAGI executor \
+             but did not write to stdout. Check the `executor` in spin.toml."
+        );
+
+        util::compose_response(&stdout)
     }
 }
 


crates/http/src/wagi.rs
```diff
@@ -143,7 +143,13 @@ impl HttpExecutor for WagiHttpExecutor {
             .or_else(ignore_successful_proc_exit_trap)?;
         tracing::info!("Module execution complete");
 
-        util::compose_response(&stdout_buffer.take())
+        let stdout = stdout_buffer.take();
+        ensure!(
+            !stdout.is_empty(),
+            "nothing was written to stdout; are you using a WAGI component?"
```
radu-matei, line 149: Could this be a bit more specific re: executors?

lann: You mean like "this component is using the WAGI executor but nothing was written to stdout"?

itowlson, line 149: As a reader I think this might suggest to me I _should_ be using a WAGI component but am not.  Perhaps "This module is configured to use the WAGI protocol, but did not write to stdout. Check the executor line in the manifest."

crates/http/src/wagi.rs
```diff
@@ -143,7 +143,13 @@ impl HttpExecutor for WagiHttpExecutor {
             .or_else(ignore_successful_proc_exit_trap)?;
         tracing::info!("Module execution complete");
 
-        util::compose_response(&stdout_buffer.take())
+        let stdout = stdout_buffer.take();
+        ensure!(
```
itowlson, line 147: Thought: put this check in `compose_response` since that's where other "insufficient output" validation is done?  It would couple `compose_response` more to the Spin hosting model vs being purely validating the WAGI protocol though.

crates/http/src/wagi.rs
radu-matei: Yes — specifically,it should guide me to look at my `executor` field in `spin.toml`.

crates/http/src/wagi.rs
lann: In the first version - and now the current version - the message references the `component` var in this scope.

crates/http/src/wagi.rs
```diff
@@ -143,7 +143,14 @@ impl HttpExecutor for WagiHttpExecutor {
...
+            !stdout.is_empty(),
+            "The {component:?} component is configured to use the WAGI executor \
```
itowlson, line 149: super nit: I think `component` is `&str` at this point so doesn't need the debug formatter

itowlson (approved):

crates/http/src/wagi.rs
lann: I like the debug formatter for error messages and logging because it quotes and escapes which can reveal surprise characters.

Issue #815, opened 2022-10-11
--log-dir has disappeared 


mreferre:
Hi there. I am experimenting with Spin in a (very) constrained environment where the only writable folder is /tmp. I used to be able to play with Spin easily by launching it with `--log-dir /tmp` and that worked just fine. The latest Spin release seems to have dropped that flag? I was able to workaround that problem by setting the `$HOME` variable to `/tmp` (because I noticed Spin trying to create folders/files in the $HOME folder. It works (but it's a bit extreme). 

Is the `--log-dir` gone for good? It seemed to be useful. 

Thanks!

mreferre:
Thanks @itowlson. 

How do you get that list of flags? I am using Spin 0.5.0 and this is what I see when I explore it with the `--help` flag:

```
ubuntu@ip-172-31-43-115:~$ spin up --help
spin-up 
Start the Spin application

USAGE:
    spin up [OPTIONS]

OPTIONS:
        --allow-transient-write
            Set the static assets of the components in the temporary directory as writable

    -b, --bindle <BINDLE_ID>
            ID of application bindle

        --bindle-password <BINDLE_PASSWORD>
            Basic http auth password for the bindle server [env: BINDLE_PASSWORD=]

        --bindle-server <BINDLE_SERVER_URL>
            URL of bindle server [env: BINDLE_URL=]

        --bindle-username <BINDLE_USERNAME>
            Basic http auth username for the bindle server [env: BINDLE_USERNAME=]

    -f, --file <APP_CONFIG_FILE>
            Path to spin.toml

    -h, --help
            

    -k, --insecure
            Ignore server certificate errors from bindle server

        --temp <TMP>
            Temporary directory for the static assets of the components

ubuntu@ip-172-31-43-115:~$ 
```

I did not spend a considerable amount of time on it but I concluded what I concluded because the script I was using (which included the `--log-dir` flag didn't work anymore - as in Spin would run but would just ignore that flag) and when I explored the syntax (above) the flag was no longer there. 

I am mildly confused but glad this could be a user error on my part with a path to happiness ;)

mreferre:
Very interesting. Thanks for the engagement on this. 

[...]

PR #931, opened 2022-12-04
feat!: Make log output to stdout default

### LINKED ISSUE #928: Make log output to stdout default
When running Spin locally in a development loop, I always use `--follow-all` to have my logs written out to the console.

From conversations I've heard others not discovering this option, e.g.:

> Debugging in spin was an initial pain point until finding the log directory. Did not find about follow-all until pretty late

I'd like to propose that writing logs to stdout becomes the default behavior, and that a new option to only output to log files gets introduced, e.g. `--silent` or `--no-stdout`.

etehtsea:
- spin writes to stdout/stderr by default (only)
- Drops `--follow-all` cmd arg is it no longer needed.
- `--follow` logs only the output for the provided components list
- `-L/--log-dir` w/o value makes spin to log in the default directory
- `-L/--log-dir <dir>` logs into the provided directory
- spin prints output to both stdout/stderr instead of stderr only
- Update `--follow <FOLLOW_ID>` to `--follow <COMPONENT_ID>`

```
  --follow <COMPONENT_ID>
        Print output for given component(s) to stdout/stderr. Can be provided multiple times to
        follow several components
```

- Mention default log location in `-L/--log-dir` help output.

Fixes #928.

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -9,7 +9,11 @@ use tokio::{
 };
 
 use crate::stdio::StdioLoggingTriggerHooks;
-use crate::{config::TriggerExecutorBuilderConfig, loader::TriggerLoader, stdio::FollowComponents};
+use crate::{
+    config::TriggerExecutorBuilderConfig,
+    loader::TriggerLoader,
+    stdio::{FollowComponents, LogDestination},
+};
 use crate::{TriggerExecutor, TriggerExecutorBuilder};
 
 pub const APP_LOG_DIR: &str = "APP_LOG_DIR";
@@ -30,12 +34,13 @@ where
     Executor::RunConfig: Args,
 {
     /// Log directory for the stdout and stderr of components.
+    /// Default location: ${HOME}/.spin/<APP_NAME>/logs
     #[clap(
             name = APP_LOG_DIR,
             short = 'L',
             long = "log-dir",
             )]
-    pub log: Option<PathBuf>,
+    pub log: Option<Option<PathBuf>>,
 
     /// Disable Wasmtime cache.
     #[clap(
@@ -56,20 +61,15 @@ where
     )]
     pub cache: Option<PathBuf>,
 
-    /// Print output for given component(s) to stdout/stderr
+    /// Print output for given component. Can be provided multiple times to
+    /// follow several components.
     #[clap(
         name = FOLLOW_LOG_OPT,
         long = "follow",
         multiple_occurrences = true,
+        value_name = "COMPONENT_ID",
         )]
     pub follow_components: Vec<String>,
-
-    /// Print all component output to stdout/stderr
-    #[clap(
-        long = "follow-all",
-        conflicts_with = FOLLOW_LOG_OPT,
-        )]
-    pub follow_all_components: bool,
 
     /// Set the static assets of the components in the temporary directory as writable.
     #[clap(long = "allow-transient-write")]
@@ -125,7 +125,8 @@ where
             let mut builder = TriggerExecutorBuilder::new(loader);
             self.update_wasmtime_config(builder.wasmtime_config_mut())?;
 
-            let logging_hooks = StdioLoggingTriggerHooks::new(self.follow_components(), self.log);
+            let logging_hooks =
+                StdioLoggingTriggerHooks::new(self.follow_components(), self.log_dest());
             builder.hooks(logging_hooks);
 
             builder.build(locked_url, trigger_config).await?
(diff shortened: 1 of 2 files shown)


crates/trigger/src/cli.rs
```diff
@@ -35,7 +39,7 @@ where
             short = 'L',
             long = "log-dir",
             )]
-    pub log: Option<PathBuf>,
+    pub log: Option<Option<PathBuf>>,
```
itowlson, line 42: If we are going with this UI, we should at least rename this option.  `--log-dir` with no value doesn't connote "log to the default place."  Seems like this also changes the behaviour to not write a log file by default, which doesn't seem to fit what Mikkel was asking for in the original issue.

crates/trigger/src/stdio.rs
```diff
@@ -15,16 +11,13 @@ pub enum FollowComponents {
...
-    All,
 }
 
 impl FollowComponents {
     /// Whether a given component should have its logs followed on stdout/stderr.
     pub fn should_follow(&self, component_id: &str) -> bool {
         match self {
-            Self::None => false,
-            Self::All => true,
+            Self::None => true,
```
itowlson, line 20: I should maybe wait for the refactoring but it seems like this gets rid of "all" and changes the semantics of "none" to "all" (and implements "none" via an empty list?).  If it results in following all components then call it "all"... *grin*

crates/trigger/src/stdio.rs
```diff
@@ -36,52 +29,84 @@ impl Default for FollowComponents {
...
+        if self.follow_components.should_follow(component_id) {
+            match &self.log {
+                LogDestination::Std => {
+                    builder.stdout_pipe(std::io::stdout());
+                    builder.stderr_pipe(std::io::stderr());
+                }
+                LogDestination::Dir(log_path) => {
+                    let log_path = log_path
+                        .as_deref()
+                        .expect("log should have been initialized in app_loaded");
```
itowlson, line 68: Seems like `LogDestination::Dir` should take a `PathBuf` rather than an `Option<PathBuf>` then, if the `None` case can never be valid?  (Again maybe I should be waiting for the refactoring.)

itowlson (commented):
Sorry, I meant to just leave a comment about the UI, but may have jumped the gun with a couple of other comments - possibly these are already on your radar for the refactoring - apologies if so.

crates/trigger/src/cli.rs
etehtsea: I was thinking about renaming it, but I couldn't find any option better than this.
When the default output is stdout, this option means to me `log to directory`, and no value corresponds the default location.

If it's made explicit in the help output like this:

```
 -L, --log-dir <APP_LOG_DIR>
            Log directory for the stdout and stderr of components [default: $HOME/.spin/<app-name>/logs]
```

it should be an acceptable solution (at least for now) with an ability to re-iterate it later.

Another option is to introduce separate argument, and use both together like `--log-destination file --log-dir <..>` but often this ends up with dozens of different options and confuses the user.

Regarding not writing to the log file by default:
- the fact that spin is implicitly writing logs somewhere was quite surprising to me. I couldn't find anything about it in the docs or help output (there is a chance that I missed it);
- another issue is that that at this moment there is no way to disable logging into the file which might be an issue in some restricted environments (https://github.com/fermyon/spin/issues/815);
- currently, with `--follow-all` options spin writes to both files and *stderr*, what surprised me until I looked into the source code (https://github.com/fermyon/spin/blob/main/crates/trigger/src/stdio.rs#L39).
When I ran spin like:

```
$ spin build --up --follow-all 2> /tmp/error.log
```

`println!` output (from the spin app) also appeared in the error log.

What I like about writing to std{out|err} by default is that this gives the flexibility to consume logs the way you like it:
- redirect into the file;
- put everything into the `/dev/null` because YOLO;
- redirect it into the log collector of your choice;

crates/trigger/src/stdio.rs
etehtsea: This one preserves current behavior when `self.log` is `None` it gets assigned to default log directory inside `app_loaded` hook:
- https://github.com/fermyon/spin/blob/main/crates/trigger/src/stdio.rs#L73-L81
- https://github.com/fermyon/spin/blob/main/crates/trigger/src/stdio.rs#L59-L63

The only change here is that arg got updated from [`Option<T>` to `Option<Option<T>>`](https://docs.rs/clap/latest/clap/_derive/index.html#arg-types) to allow `-L/--log-dir` to be called w/o value (to use the default).

I would prefer to refactor it in the follow-up (if possible).

PR #933, opened 2022-12-04
Set up a devcontainer as part of the Rust template

itowlson:
This is very closely based on the Spin codespace/devcontainer work by @craiglpeters, and it's possible that some bits could be refined (e.g. the Dockerfile uses the Rust baseline but then there is a Rust feature in the devcontainer.json as well).

With this change, when a user creates a Spin application using the `http-rust` templates, they will be able to build it in a devcontainer right off the bat, or in a GitHub codespace once they have committed it to GH.  `spin up` and `spin deploy` work fine, though `spin up` now auto-opens the browser (I assume based on the port forwarding settings) which I'm a bit doubtful about.  Also the user would need to tweak the port settings if they wanted to listen on a non-default port, but hopefully that wouldn't be an issue in a devcontainer.

With the current setup, the GH codespace takes a while to build - I do wonder if it would be faster if we created images with Spin and the appropriate tools - but that's harder to scale across multiple languages.

One thing I haven't figured out yet is how to make this play nicely with multiple components.  Like if I do a `spin new http-empty` and then add two components, I should be able to have a codespace with both of them in it.  And if one component is in Go and the other in Rust, the devcontainer declaration should be modified to include both languages' toolsets.

But anyway here's something to play with and provide feedback on.

### DIFF
--- a/templates/http-rust/content/.devcontainer/Dockerfile
+++ b/templates/http-rust/content/.devcontainer/Dockerfile
@@ -1,0 +1,16 @@
+ARG VARIANT="bullseye"
+FROM rust:1-${VARIANT}
+
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    # Remove imagemagick due to https://security-tracker.debian.org/tracker/CVE-2019-10131
+    && apt-get purge -y imagemagick imagemagick-6-common
+
+# [Optional] Uncomment this section to install additional OS packages.
+# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+#     && apt-get -y install --no-install-recommends <your-package-list-here>
+
+# Install additional rust components (rust-analyzer)
+ARG RUST_ANALYZER_URL="https://github.com/rust-lang/rust-analyzer/releases/download/2022-04-11/rust-analyzer-x86_64-unknown-linux-gnu.gz"
+RUN curl -sL "$RUST_ANALYZER_URL" | gunzip -c - > /usr/local/bin/rust-analyzer && chmod +x /usr/local/bin/rust-analyzer
+
+RUN rustup target add wasm32-wasi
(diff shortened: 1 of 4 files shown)


mikkelhegn:
I can't seem to get the GitHub UI to choose the devcontainer configuration in this PR, only the one in the root, but I assume it can run locally.

I love the idea of making it easy to get a machine with all the dependencies configured, so you only have to do `spin new`, `spin build` and `spin up`. And to easy build time, we could consider a common image, which could be the same we would use for regression testing of the templates.

I think there's a larger conversation about how to organize all of this, as there's a direct dependency between SDKs, templates and devcontainer definitions.

itowlson:
Closing as pretty much stale at this point.

craiglpeters:
@itowlson I would appreciate a chat to help me understand better how we can better meet your use case

itowlson:
@craiglpeters I think part of the issue is that we're not sure what our use case is.  This was an exploration of having our template system generate devcontainers as part of new apps (when the user ran `spin new`).  It stalled on:

1. You couldn't use the devcontainer as a Codespace until you pushed the code to GitHub, so it didn't help people getting over the initial "how do I build this thing without weird errors" hump which is important to us.  (I know they can run devcontainers locally.)
2. @mikkelhegn raised a the issue of a devcontainer that would contain _Spin itself_ together with the required toolchain so the user could do `spin new` _within_ the devcontainer instead of doing `spin new` on the host and then entering an app-specific devcontainer (which is probably more economical).

Then we got distracted onto other things and hit the holiday break, and this PR languished and grew stale.

I think we need to clarify in our own minds what a good use case is for this, and then we can pick this up again in a more "here is a thing we know will make users happier" manner, rather than this PR which, in retrospect, was a bit "OMG CODESPACES ARE SO COOL LET ME TRY."

PR #934, opened 2022-12-05
Fix `spin up --help` omitting many options if app is not ready to lock

### LINKED ISSUE #932: `spin up --help` doesn't display all options unless application is fully ready to load
If I do a `spin up --help` in a directory with a `spin.toml`, I normally get a full set of options including those that (internally) happen to be implemented through the trigger such as `--log-dir` and `--follow`.

However, if I `spin new` a directory and do `spin up --help`, I get only the subset that (internally) happen to be implemented diirectly on the `Up` command type itself.

But if I do a `spin build` then it starts working again.  It was the inability to load the component Wasm source that was causing the options to disappear.

So it looks like `spin up --help` is trying to load the application to the point where it is ready for the relevant `spin trigger foo` command to execute, even though all we actually need to be able to extract is the trigger type.

itowlson:
Partially fixes #932.

There remains an outstanding issue where `spin up --help` will skip "trigger" arguments if the app contains no components.  This happens because the serialisation objects make `component` a mandatory field, so we cannot even load the file in this case.  This could be addressed by making `component` optional, and validating that at least one component exists during the locking process (or otherwise after the initial load).  This is probably a bit of a niche case though.

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -148,32 +148,20 @@ impl UpCommand {
             ApplicationTrigger::Redis(_) => "redis",
         };
 
-        // Build and write app lock file
-        let locked_app = spin_trigger::locked::build_locked_app(app, &working_dir)?;
-        let locked_path = working_dir.join("spin.lock");
-        let locked_app_contents =
-            serde_json::to_vec_pretty(&locked_app).context("failed to serialize locked app")?;
-        std::fs::write(&locked_path, locked_app_contents)
-            .with_context(|| format!("failed to write {:?}", locked_path))?;
-        let locked_url = Url::from_file_path(&locked_path)
-            .map_err(|_| anyhow!("cannot convert to file URL: {locked_path:?}"))?
-            .to_string();
-
-        // For `spin up --help`, we just want the executor to dump its own argument usage info
-        let trigger_args = if self.help {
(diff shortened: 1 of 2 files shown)


src/commands/up.rs
```diff
@@ -148,32 +148,21 @@ impl UpCommand {
...
         cmd.arg("trigger")
-            .env(SPIN_WORKING_DIR, working_dir)
-            .env(SPIN_LOCKED_URL, locked_url)
             .arg(trigger_type)
-            .args(trigger_args);
+            .env(SPIN_WORKING_DIR, &working_dir);
+
+        if self.help {
+            let help_args = vec![OsString::from("--help-args-only")];
+            cmd.args(help_args);
```
lann, line 160: Can simplify since it doesn't have to match the type of `trigger_args`:

```suggestion
            cmd.arg("--help-args-only");
```

src/commands/up.rs
```diff
@@ -199,6 +188,25 @@ impl UpCommand {
         }
     }
 
+    fn build_locked_app(
```
lann, line 191: Maybe?

```suggestion
    fn write_locked_app(
```

lann (approved):
Minor style suggestions

PR #938, opened 2022-12-06
fix: Get rid of log/env_logger leftovers

etehtsea:
If I'm not mistaken, `env_logger` is no longer initialized, so the code that is still relying on it doesn't log anything.

Fixes https://github.com/fermyon/spin/issues/945.


itowlson (approved):
LGTM - thanks!

itowlson:
@radu-matei tagging you for review because it looks you were the one who removed the `env_logger::init` call, in https://github.com/fermyon/spin/commit/b6eb2f2d9dcd61265449b52f73111dc937a872c6#diff-be5c1b613147e092e70bc6ff1aa68de4be91ebe6102884d53ff040b9a6e9bb93 - can you sanity check please?  Thanks!

PR #951, opened 2022-12-08
Error if user follows a non-existent component

### LINKED ISSUE #944: --follow doesn't validate provided values
### Current behaviour
Spin accepts invalid component ids as an input.

### Expected behaviour
Spin fails with error if provided component id is invalid.

### Example

```sh
spin/examples/http-rust-outbound-http main $  ../../target/debug/spin build --up --follow rust-outbound-http,rust-outbound-http-wildcard
<..>
Serving http://127.0.0.1:3000
Available Routes:
  rust-outbound-http: http://127.0.0.1:3000/outbound
  rust-outbound-http-wildcard: http://127.0.0.1:3000/wildcard
```

The example about should fail because `--follow rust-outbound-http --follow rust-outbound-http-wildcard` is expected instead.

Other examples:

```
--follow ""
--follow ust-outbound-http
```

itowlson:
Fixes #944.

If a non-existent component is specified in `--follow`, Spin now exits with this message:

```
$ spin up --follow godby --follow helo
Error: The following component(s) specified in --follow do not exist in the application:
  - godby
  - helo
The following components exist:
  - gogogo
  - goodbyte
  - hello
Error: exit status: 1
```

(I don't know where the "exit status 1" comes from - is that an artefact of the `spin trigger` command failing and that output being reported by `spin up` @lann?)

(cc @etehtsea - it won't let me add you as a reviewer)

### DIFF
--- a/crates/trigger/src/stdio.rs
+++ b/crates/trigger/src/stdio.rs
@@ -65,11 +65,33 @@ impl StdioLoggingTriggerHooks {
         ComponentStdioWriter::new(&log_path, follow)
             .with_context(|| format!("Failed to open log file {log_path:?}"))
     }
+
+    fn validate_follows(&self, app: &spin_app::App) -> anyhow::Result<()> {
+        match &self.follow_components {
+            FollowComponents::Named(names) => {
+                let component_ids: HashSet<_> =
+                    app.components().map(|c| c.id().to_owned()).collect();
+                let unknown_names: Vec<_> = names.difference(&component_ids).collect();
+                if unknown_names.is_empty() {
+                    Ok(())
+                } else {
+                    let unknown_list = bullet_list(&unknown_names);
+                    let actual_list = bullet_list(&component_ids);
+                    let message = anyhow::anyhow!("The following component(s) specified in --follow do not exist in the application:\n{unknown_list}\nThe following components exist:\n{actual_list}");
+                    Err(message)
+                }
+            }
+            _ => Ok(()),
+        }
+    }
 }
 
 impl TriggerHooks for StdioLoggingTriggerHooks {
     fn app_loaded(&mut self, app: &spin_app::App) -> anyhow::Result<()> {
         let app_name: &str = app.require_metadata("name")?;
+
+        self.validate_follows(app)?;
+
         // Set default log_dir (if not explicitly passed)
         let log_dir = self.log_dir.get_or_insert_with(|| {
             let parent_dir = match dirs::home_dir() {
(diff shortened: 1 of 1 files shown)


crates/trigger/src/stdio.rs
```diff
@@ -65,11 +65,36 @@ impl StdioLoggingTriggerHooks {
         ComponentStdioWriter::new(&log_path, follow)
             .with_context(|| format!("Failed to open log file {log_path:?}"))
     }
+
+    fn validate_follows(&self, app: &spin_app::App) -> anyhow::Result<()> {
+        match &self.follow_components {
+            FollowComponents::Named(names) => {
+                let component_ids: HashSet<_> =
+                    app.components().map(|c| c.id().to_owned()).collect();
+                let unknown_names: Vec<_> = names
```
etehtsea, line 74:
```suggestion
                let unknown_names: Vec<_> = names.difference(&component_ids).collect();
```

etehtsea:
I've checked it locally and it works great!
Thanks!

crates/trigger/src/stdio.rs
```diff
@@ -126,3 +151,11 @@ impl std::io::Write for ComponentStdioWriter {
         Ok(())
     }
 }
+
+fn bullet_list<S: std::fmt::Display>(items: impl IntoIterator<Item = S>) -> String {
+    items
+        .into_iter()
+        .map(|item| format!("  - {item}"))
+        .collect::<Vec<_>>()
+        .join("\n")
```
lann, line 160: _Almost_ equivalent (but with extra trailing \n):

```suggestion
        .map(|item| format!("  - {item}\n"))
        .collect()
```

lann (approved):

crates/trigger/src/stdio.rs
```diff
@@ -65,11 +65,36 @@ impl StdioLoggingTriggerHooks {
...
+
+    fn validate_follows(&self, app: &spin_app::App) -> anyhow::Result<()> {
+        match &self.follow_components {
+            FollowComponents::Named(names) => {
+                let component_ids: HashSet<_> =
+                    app.components().map(|c| c.id().to_owned()).collect();
+                let unknown_names: Vec<_> = names
+                    .iter()
+                    .filter(|name| !component_ids.contains(name.as_str()))
+                    .collect();
```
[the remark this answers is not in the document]
itowlson: Bah, I knew that had to exist _somewhere_.  Thank you!

crates/trigger/src/stdio.rs
itowlson: That's nice, but because I return the error message rather than fully controlling the printing it means, I end up with a blank line between the error and the exit code message:

```
$ spin up --follow godby --follow helo
Error: The following component(s) specified in --follow do not exist in the application:
  - helo
  - godby
The following components exist:
  - gogogo
  - hello
  - goodbyte

Error: exit status: 1
```

...which is neither here nor there, I'm sure, but I slightly prefer the more compact form.  If we can get rid of the exit status message then this would be a good change.  (And thanks for mentioning it, I didn't know collecting to String did this.  Though now I do I'm not sure how I feel about it!)

PR #952, opened 2022-12-09
Don't return error status if plugin is already installed

### LINKED ISSUE #948: running `spin plugin install` again should not error out
if we run `spin plugin install js2wasm` twice in a row, the 2nd one fails with exit code 1.

can we make it so that 2nd run exit with a warning and exit code 0.

itowlson:
Fixes #948.

This should not result in any change in how the UI appears to the user, but if a `spin plugin install` is a no-op because the correct version of the plugin is already active, the `spin` process exit code will now be 0 instead of 1.

### DIFF
--- a/crates/plugins/src/manager.rs
+++ b/crates/plugins/src/manager.rs
@@ -100,7 +100,7 @@ impl PluginManager {
         spin_version: &str,
         override_compatibility_check: bool,
         allow_downgrades: bool,
-    ) -> Result<()> {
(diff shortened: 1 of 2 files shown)


kate-goldenring (approved):

kate-goldenring:
LGTM! Thanks for the change @itowlson!

michelleN (approved):

karthik2804 (approved):

PR #961, opened 2022-12-12
chore: update contributing link in README.md

morristai:
Since the [link](./docs/content/contributing.md) in the README.md is broken, the contributing link should now direct to the [Fermyon contributing page](https://developer.fermyon.com/spin/contributing).
image


itowlson (approved):
Well, that's embarrassing.  Thanks for tracking it down and taking the time to fix it!

morristai:
> Sorry, I'm afraid we need the commit to be signed, both with a Git "signed off by" and a GPG key (verified commit). Could you do those and re-push please? Thanks!

Done!

radu-matei (approved):
This is great, thank you for catching this!

LGTM

PR #967, opened 2022-12-12
add `--direct-mounts` option to `spin up`

### LINKED ISSUE #964: Add `spin up` option to mount static asset directory directly
Currently, `spin up` always copies static assets into a temporary directory and mounts that into the host.  This creates friction when iteratively developing static assets (e.g. SPAs) since `spin up` must be restarted each time any asset changes in order for the updated asset to be served.

I'm proposing that we add an option, e.g. `--direct-mount` to mount the asset directory directly instead of using a temporary directory, so that changes on the host filesystem are immediately visible to subsequent app invocations, with no restart required.

dicej:
This option tells `spin up` to directly mount any asset directories specified in the manifest file instead of copying the contents to a temporary directory and mounting that.  This is useful when iteratively developing the static assets in the app, since any modifications to assets are immediately visible to subsequent invocations -- no restart required.

Closes #964.

### DIFF
--- a/crates/loader/src/local/assets.rs
+++ b/crates/loader/src/local/assets.rs
@@ -20,22 +20,49 @@ use super::config::{RawDirectoryPlacement, RawFileMount};
 pub(crate) async fn prepare_component(
     raw_mounts: &[RawFileMount],
     src: impl AsRef<Path>,
-    base_dst: impl AsRef<Path>,
+    base_dst: Option<impl AsRef<Path>>,
     id: &str,
     exclude_files: &[String],
 ) -> Result<Vec<DirectoryMount>> {
-    tracing::info!(
-        "Mounting files from '{}' to '{}'",
-        src.as_ref().display(),
-        base_dst.as_ref().display()
-    );
+    if let Some(base_dst) = base_dst {
+        tracing::info!(
+            "Mounting files from '{}' to '{}'",
+            src.as_ref().display(),
+            base_dst.as_ref().display()
+        );
 
-    let files = collect(raw_mounts, exclude_files, src)?;
-    let host = create_dir(&base_dst, id).await?;
-    let guest = "/".to_string();
-    copy_all(&files, &host).await?;
+        let files = collect(raw_mounts, exclude_files, src)?;
+        let host = create_dir(&base_dst, id).await?;
+        let guest = "/".to_string();
+        copy_all(&files, &host).await?;
 
-    Ok(vec![DirectoryMount { guest, host }])
+        Ok(vec![DirectoryMount { guest, host }])
+    } else {
+        tracing::info!("directly mounting local asset directories into guest");
+
+        if !exclude_files.is_empty() {
+            bail!("exclusions not permitted when mounting directories directly")
+        }
+
+        raw_mounts
+            .iter()
+            .map(|mount| match mount {
+                RawFileMount::Placement(RawDirectoryPlacement {
+                    source,
+                    destination,
+                }) => Ok(DirectoryMount {
+                    guest: destination
+                        .to_str()
+                        .context("unable to parse mount destination as UTF-8")?
+                        .to_owned(),
+                    host: source.clone(),
+                }),
+                RawFileMount::Pattern(_) => Err(anyhow!(
+                    "patterns not permitted when mounting directories directly"
+                )),
+            })
+            .collect()
+    }
 }
 
 /// A file that a component requires to be present at runtime.
--- a/crates/loader/src/local/assets.rs
+++ b/crates/loader/src/local/assets.rs
@@ -25,44 +25,60 @@ pub(crate) async fn prepare_component(
     exclude_files: &[String],
 ) -> Result<Vec<DirectoryMount>> {
     if let Some(base_dst) = base_dst {
-        tracing::info!(
-            "Mounting files from '{}' to '{}'",
-            src.as_ref().display(),
-            base_dst.as_ref().display()
-        );
-
-        let files = collect(raw_mounts, exclude_files, src)?;
-        let host = create_dir(&base_dst, id).await?;
-        let guest = "/".to_string();
-        copy_all(&files, &host).await?;
-
-        Ok(vec![DirectoryMount { guest, host }])
+        prepare_component_with_temp_dir(raw_mounts, src, base_dst, id, exclude_files).await
     } else {
         tracing::info!("directly mounting local asset directories into guest");
 
         if !exclude_files.is_empty() {
-            bail!("exclusions not permitted when mounting directories directly")
+            bail!("this component cannot be run with `--direct-mount` because some files are excluded from mounting")
         }
 
-        raw_mounts
-            .iter()
-            .map(|mount| match mount {
-                RawFileMount::Placement(RawDirectoryPlacement {
-                    source,
-                    destination,
-                }) => Ok(DirectoryMount {
-                    guest: destination
-                        .to_str()
-                        .context("unable to parse mount destination as UTF-8")?
-                        .to_owned(),
-                    host: source.clone(),
-                }),
-                RawFileMount::Pattern(_) => Err(anyhow!(
-                    "patterns not permitted when mounting directories directly"
-                )),
-            })
-            .collect()
+        prepare_component_with_direct_mounts(raw_mounts)
     }
+}
+
+async fn prepare_component_with_temp_dir(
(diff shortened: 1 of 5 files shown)


src/commands/up.rs
```diff
@@ -87,6 +87,14 @@ pub struct UpCommand {
     #[clap(long = "temp")]
     pub tmp: Option<PathBuf>,
 
+    /// For local apps with directory mounts and no excluded files, mount them directly instead of using a temporary
+    /// directory.
+    ///
+    /// This allows you to update the assets on the host filesystem such that the updates are visible to the guest
+    /// without a restart.  This will raise an error if used with bindle apps or apps using file patterns.
+    #[clap(long, takes_value = false)]
+    pub direct_mounts: bool,
```
etehtsea, line 96: Does it make sense to provide `conflicts_with` for `--temp` and `--bindle` opts?

crates/loader/src/local/assets.rs
```diff
@@ -20,22 +20,49 @@ use super::config::{RawDirectoryPlacement, RawFileMount};
...
+        let host = create_dir(&base_dst, id).await?;
+        let guest = "/".to_string();
+        copy_all(&files, &host).await?;
+
+        Ok(vec![DirectoryMount { guest, host }])
+    } else {
+        tracing::info!("directly mounting local asset directories into guest");
+
+        if !exclude_files.is_empty() {
+            bail!("exclusions not permitted when mounting directories directly")
```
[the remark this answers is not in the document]
dicej: This line means that if you ask for direct mounts for an app that specifies on or more files to be excluded, we'll raise an error and exit without running the app.

Per [Ivan's comment](https://github.com/fermyon/spin/issues/964#issuecomment-1347186000), we'll need a more general "hot reload" solution to handle exclusions correctly.  It will probably involve using a temporary directory and simultaneously monitoring the original directories and files, copying the non-excluded items over to the temporary directory as they change.

crates/loader/src/local/assets.rs
```diff
@@ -20,22 +20,49 @@ use super::config::{RawDirectoryPlacement, RawFileMount};
...
+                    destination,
+                }) => Ok(DirectoryMount {
+                    guest: destination
+                        .to_str()
+                        .context("unable to parse mount destination as UTF-8")?
+                        .to_owned(),
+                    host: source.clone(),
+                }),
+                RawFileMount::Pattern(_) => Err(anyhow!(
+                    "patterns not permitted when mounting directories directly"
```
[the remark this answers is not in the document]
dicej: Yes, I think that's doable using WASI preopens -- not sure how much work would be required, though.  As I mentioned above, a general-purpose "hot reload" feature could handle this using a dynamically-updated temporary directory.

Personally, I prefer the tidiness of directory mounts anyway: you designate a directory for assets, put the ones you need in it (possibly as part of `spin build`) and omit the ones you don't.  You don't need to fuss with patterns or exclusions that way.

src/commands/up.rs
```diff
@@ -127,10 +135,25 @@ impl UpCommand {
                     .as_deref()
                     .unwrap_or_else(|| DEFAULT_MANIFEST_FILE.as_ref());
                 let bindle_connection = self.bindle_connection();
-                spin_loader::from_file(manifest_file, &working_dir, &bindle_connection).await?
+                spin_loader::from_file(
+                    manifest_file,
+                    if self.direct_mounts {
+                        None
```
etehtsea, line 141: Just a random thought about further refactoring - does it make sense to introduce `DirectMounts` for `WorkingDirectory` enum which `.path()` resolves to `None`?

https://github.com/fermyon/spin/blob/8dfc4d6fbb31d624eb034f78640cb207bec1f682/src/commands/up.rs#L221-L233

(Probably, `WorkingDirectory` could be provided instead of `base_dst`, but I haven't checked for does it fit or not.)

src/commands/up.rs
dicej: Good idea.  I originally had it conflict with `--temp`, but then realized that Spin uses the temporary directory for other things besides file and directory mounts, so it's still useful to specify `--temp` and `--direct-mounts` simultaneously.  I agree that it should conflict with the bindle opts, though.

etehtsea: Oh, right, sorry, I missed that.

I misunderstood the term "static assets" in the opt description:

https://github.com/fermyon/spin/blob/main/src/commands/up.rs#L86-L88

dicej: Yeah, I think that description is misleading -- I was also confused. I suspect it was originally just for static assets, but then ended up being used for other things.

crates/loader/src/local/assets.rs
```diff
@@ -20,22 +20,49 @@ use super::config::{RawDirectoryPlacement, RawFileMount};
...
 
-    Ok(vec![DirectoryMount { guest, host }])
+        let files = collect(raw_mounts, exclude_files, src)?;
+        let host = create_dir(&base_dst, id).await?;
+        let guest = "/".to_string();
+        copy_all(&files, &host).await?;
+
+        Ok(vec![DirectoryMount { guest, host }])
+    } else {
+        tracing::info!("directly mounting local asset directories into guest");
```
itowlson, line 41: I think the if and else parts of this function are entirely distinct.  As both are quite long and detailed, I'd be keen to split them into two separate functions (with this function just dispatching to one or the other) so that the reader has clearer breaks and less indentation!

crates/loader/src/local/assets.rs
itowlson, line 44: I'd flip this error around - if there are exclusions, it's because that's appropriate for the application, and it should be `--direct-mount` that yields, not the exclusions.  E.g. "this component cannot be run with --direct-mount because some files are excluded from mounting" or something like that?

crates/loader/src/local/assets.rs
itowlson, line 61: Same comment about flipping the error message

crates/loader/src/local/assets.rs
```diff
@@ -20,22 +20,49 @@ use super::config::{RawDirectoryPlacement, RawFileMount};
...
+
+        if !exclude_files.is_empty() {
+            bail!("exclusions not permitted when mounting directories directly")
+        }
+
+        raw_mounts
+            .iter()
+            .map(|mount| match mount {
+                RawFileMount::Placement(RawDirectoryPlacement {
```
itowlson, line 50: Might be able to get rid of some of the verticalness here by matching as `RawFileMount::Placement(dp)` and then using `dp.source` and `dp.destination`.  Or maybe if we manage to reduce the indenting then `cargo fmt` will be more merciful!

src/commands/up.rs
```diff
@@ -127,10 +135,25 @@ impl UpCommand {
                     .as_deref()
                     .unwrap_or_else(|| DEFAULT_MANIFEST_FILE.as_ref());
                 let bindle_connection = self.bindle_connection();
-                spin_loader::from_file(manifest_file, &working_dir, &bindle_connection).await?
+                spin_loader::from_file(
+                    manifest_file,
+                    if self.direct_mounts {
```
itowlson, line 140: I found the inlining here really threw off my reading - this is a simple function call with three arguments, but the inlining spreads it over 10 lines and three levels of indentation.  Can we not create a variable (e.g. `asset_dest`) on the line before and replace `&working_dir` in the original call with that?

src/commands/up.rs
```diff
@@ -127,10 +135,25 @@ impl UpCommand {
...
+                        Some(&working_dir)
+                    },
+                    &bindle_connection,
+                )
+                .await?
             }
             (None, Some(bindle)) => match &self.server {
-                Some(server) => spin_loader::from_bindle(bindle, server, &working_dir).await?,
+                Some(server) => {
+                    if self.direct_mounts {
```
itowlson, line 151: If we get clap to block the combination as @etehtsea suggested, then could we remove this check?  Belt and braces, yes, ideally, but there's already a lot of nesting of conditions (combined with other indents that obscure the structure, THANKS CARGO FMT), and this additional level starts to make the code very in-and-out-y if that makes sense...

itowlson (changes requested):
The functionality looks fine, but I have a few code organisation/formatting requests, and some suggestions on how to frame the "you can't do that here" error messages.  Otherwise good to go!

itowlson (approved):
Looks great now - thanks!

PR #970, opened 2022-12-13
use current version during first time deployment

### LINKED ISSUE #955: delete and redeploy of same app sometimes uses old revision
i tried deleting (from the UI) and redeploying my app on Fermyon cloud and sometimes it seems to be serving old revision (until I deploy the app 2nd time).

rajatjindal:
fixes #955 

Signed-off-by: Rajat Jindal <rajatjindal83@gmail.com>

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -384,17 +384,24 @@ impl DeployCommand {
                 existing_channel_id
             }
             Err(_) => {
-                let range_rule = Some(bindle_id.version_string());
                 let app_id = CloudClient::add_app(&client, &name, &name)
                     .await
                     .context("Unable to create app")?;
+
+                // When creating the new app, InitialRevisionImport command is triggered
+                // which automatically imports all revisions from bindle into db
+                // therefore we do not need to call add_revision api explicitly here
+                let active_revision_id = self
+                    .get_revision_id_cloud(&client, bindle_id.version_string().clone(), app_id)
+                    .await?;
+
                 CloudClient::add_channel(
                     &client,
                     app_id,
                     String::from(SPIN_DEPLOY_CHANNEL_NAME),
-                    CloudChannelRevisionSelectionStrategy::UseRangeRule,
-                    range_rule,
+                    CloudChannelRevisionSelectionStrategy::UseSpecifiedRevision,
                     None,
+                    Some(active_revision_id),
                 )
                 .await
                 .context("Problem creating a channel")?


michelleN:
@rajatjindal Thanks for this. I see there is a `[WIP]` in the title. Is this still a WIP or is it ready for review?

michelleN:
@rajatjindal Thanks for this. I see there is a [WIP] in the title. Is this still a WIP or is it ready for review?

rajatjindal:
Hi @michelleN , i just edited out the `WIP` tag, and this PR should be ready for review. IMO we still need to fix #971 for this fix to be complete though

src/commands/deploy.rs
```diff
@@ -384,17 +384,21 @@ impl DeployCommand {
                 existing_channel_id
             }
             Err(_) => {
-                let range_rule = Some(bindle_id.version_string());
                 let app_id = CloudClient::add_app(&client, &name, &name)
                     .await
                     .context("Unable to create app")?;
+
+                let active_revision_id = self
```
itowlson, line 391: We don't seem to be calling `CloudClient::add_revision` on this path (as we do on the "app exists" path.  Do we understand why it is unnecessary on this path but necessary on the other?  Could we maybe add a comment for the benefit of a maintainer who comes across this and is puzzled like me?  Thanks!

itowlson (changes requested):
This looks good but I had one question - and even if my question is a no-op I think it would benefit from a clarifying comment - thanks!

rajatjindal:
Hi Ivan,

Thanks for your question, it totally makes sense to add it to doc. Will do it in morning.

Recording here as well

What i noticed is that during create app, "import revisions" event is triggered which reconciles bindle revisions in our storage for the given user, so add revision is not required here.

I think one more testcase i should verify explicitly is creation of 2nd app, however because i tested with Fermyon cloud, that is probably covered already. 

I will update here on how it goes

src/commands/deploy.rs
rajatjindal: https://github.com/deislabs/hippo/blob/e83bda6858c5d5b8db234f6f07bbf4b1afb5f8d2/src/Application/Apps/EventHandlers/InitialRevisionImport.cs#L9  - on app creation, we import all existing revisions into db. therefore we don't need add_revision to be called explicitly (I did added that initially, but it was creating a duplicate revision entry into db).

added this as a comment in code as well. 

thank you for pointing it out

PR #972, opened 2022-12-14
Implement `spin plugin list`

### LINKED ISSUE #937: request to add `spin plugin list` command
it will be nice to list the `plugins` available to install (or already installed?)

itowlson:
Fixes #937.

I'd really value comments on the structure of this - even more than usual, because I kind of felt that a bunch of this must already be in there but I wasn't sure where to find it.  So please do flag up if logic is in the wrong places or are already done somewhere else.

The output with this code is loosely based on how `apt` looks on my WSL Ubuntu system, and looks like this:

```
$ spin plugin list
arm-only 0.1.0 [incompatible]
example 0.1.0
example 0.2.0
js2wasm 0.1.0 [installed]
not-spin-ver 10.2.0 [requires other Spin version]

$ spin plugin list --installed
js2wasm 0.1.0 [installed]
```

It could alternatively be printed in tabular format, and/or with colour highlighting, and/or with emoticons - the `PluginDescriptor` hopefully decouples the printing logic enough that this would be easy to fiddle with.

At the moment, this glides silently over plugins that are in such a corrupted state that we can't even load them.  This is preferable to being unable to list plugins _at all_ just because one got mangled while I was trying to exit `vi`.  It would be nice to print a warning if any were skipped, but it's a bit fiddly to do without losing the friendly `?` syntax - I can have a noodle on it if people think it's worth it though.

### DIFF
--- a/crates/plugins/src/store.rs
+++ b/crates/plugins/src/store.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, Result};
 use flate2::read::GzDecoder;
 use std::{
+    ffi::OsStr,
     fs::{self, File},
     path::{Path, PathBuf},
 };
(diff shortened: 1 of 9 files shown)


crates/plugins/tests/README.md
```diff
@@ -0,0 +1,8 @@
+## Fake plugin manifests
```
kate-goldenring, line 1: Do we want to create a test or integration test for these, whether in this PR or a future one? Or is this good to keep around mainly for manual testing?

crates/plugins/src/store.rs
```diff
@@ -62,6 +63,63 @@ impl PluginStore {
...
+            .filter(|p| p.is_dir())
+            .collect();
+        let manifest_paths: Vec<_> = plugin_dirs
+            .iter()
+            .flat_map(|d| Self::json_files_in(d))
+            .collect();
+        let manifests: Vec<_> = manifest_paths
+            .iter()
+            .filter_map(|path| Self::try_read_manifest_from(path))
+            .collect();
```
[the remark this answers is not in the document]
kate-goldenring: I like the steps being separated as you've done for readability. Could we remove the `collect` calls but still save the iterator progress in intermediate results?

kate-goldenring (approved):
Just some minor comments but this is great! I love how you implemented this to look similar with apt and make it clear the "status" of a plugin (compatible with host, compatible with Spin, installed, etc) and how easy it will be to add the different print styles in future if desired.

I think glossing over invalidly formatted plugins is okay. They shouldn't be able to be merged into the plugins repo as they would fail the json schema workflow. Something could get through but as you mention, it is likely not worth the syntactical disruption.

src/commands/plugins.rs
```diff
@@ -285,6 +289,122 @@ impl Upgrade {
     }
 }
 
+/// Install plugins from remote source
+#[derive(Parser, Debug)]
+pub struct List {
+    /// Skips prompt to accept the installation of the plugin.
```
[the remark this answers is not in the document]
itowlson: Yikes, thanks for catching that!

crates/plugins/tests/README.md
itowlson: I wasn't sure how to fit this stuff into the existing integration tests (not helped by my general woes with running the integration tests I admit).  I'd like to use them in a set of more unit-ish tests like the ones in the templates manager but not sure how - that might be a good thing to pick up in a later PR.  I would still say keep them to save maintenance programmers having to recreate them next time round!

kate-goldenring: Sounds good to me!

kate-goldenring (approved):
Looks like a Rust fmt fix and this is good to go

itowlson:
Huh.  `cargo fmt` isn't requiring any changes for me...

itowlson:
OH IT'S A NEW CLIPPY THANKS FOR THE BREAKING CHANGE ON A MINOR VERSION RUST

PR #973, opened 2022-12-14
Handle paginated responses when looking for revision and channel IDs

### LINKED ISSUE #971: support pagination in list_revision fn
in https://github.com/fermyon/spin/blob/8dfc4d6fbb31d624eb034f78640cb207bec1f682/crates/cloud/src/client.rs#L271, we do not do pagination, so probably defaulting to `50 per page` and first page only.  (per api spec)

I believe this is the reason for occasional failure when deploying to  the cloud:

```
no revision with version 0.1.0+r257179w and app id 9eb2d7df-30ad-4a05-83b7-313fbd455811
```

itowlson:
Fixes #971.

@rajatjindal This is an attempt to resolve the issue you've been seeing with the "no revision" error, but I don't have a test case to run it on.  Would you mind checking it out and seeing if it behaves correctly in all the various cases?  Thanks!

(I did it this way rather than modifying the cloud client to return all revisions so that we didn't have to hold a potentially large number of revisions in memory at once, but it does make both the client and the consumer a bit more fiddly.  Totally open to doing it the other way.)

### DIFF
--- a/crates/cloud/src/client.rs
+++ b/crates/cloud/src/client.rs
@@ -273,6 +273,19 @@ impl Client {
             .await
             .map_err(format_response_error)
     }
+
+    pub async fn list_revisions_next(
+        &self,
+        previous: &RevisionItemPage,
+    ) -> anyhow::Result<RevisionItemPage> {
+        api_revisions_get(
+            &self.configuration,
+            Some(previous.page_index + 1),
+            Some(previous.page_size),
+        )
+        .await
+        .map_err(format_response_error)
+    }
 }
 
 #[derive(Deserialize, Debug)]
--- a/crates/cloud/src/client.rs
+++ b/crates/cloud/src/client.rs
@@ -153,6 +153,19 @@ impl Client {
         .map_err(format_response_error)
     }
 
+    pub async fn list_channels_next(&self, previous: &ChannelItemPage) -> Result<ChannelItemPage> {
+        api_channels_get(
+            &self.configuration,
+            Some(""),
+            Some(previous.page_index + 1),
+            Some(previous.page_size),
+            Some("Name"),
+            None,
+        )
+        .await
+        .map_err(format_response_error)
+    }
+
     #[allow(clippy::too_many_arguments)]
     pub async fn add_channel(
         &self,
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -553,15 +553,29 @@ impl DeployCommand {
         name: String,
         app_id: Uuid,
     ) -> Result<Uuid> {
-        let channels_vm = CloudClient::list_channels(cloud_client).await?;
-        let channel = channels_vm
-            .items
-            .iter()
-            .find(|&x| x.app_id == app_id && x.name == name.clone());
-        match channel {
-            Some(c) => Ok(c.id),
-            None => bail!("No channel with app_id {} and name {}", app_id, name),
+        let mut channels_vm = cloud_client.list_channels().await?;
+
+        loop {
+            if let Some(channel) = channels_vm
+                .items
+                .iter()
+                .find(|&x| x.app_id == app_id && x.name == name.clone())
+            {
+                return Ok(channel.id);
+            }
+
+            if channels_vm.is_last_page {
+                break;
+            }
+
+            channels_vm = cloud_client.list_channels_next(&channels_vm).await?;
         }
+
+        Err(anyhow!(
+            "No channel with app_id {} and name {}",
+            app_id,
+            name
+        ))
     }
 
     async fn create_and_push_bindle(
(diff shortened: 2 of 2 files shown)


bacongobbler (approved):

rajatjindal:
i compiled a local version of `spin` from this branch and occasionally it seems to be running into infinite loop.

```
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
total: 65 page_index: 0 page_size: 50 is_last_page: false
.
.
.
Error: response status code: 429 Too Many Requests

```

I am double checking to ensure nothing else is wrong, but giving you a heads up here incase you can figure out what might be happening here

itowlson:
Ah, sorry, I guess page index is the page desired not the previous.  Pushed a candidate fix, please try again?

src/commands/deploy.rs
```diff
@@ -505,20 +505,30 @@ impl DeployCommand {
...
+                .find(|&x| x.revision_number == bindle_version && x.app_id == app_id)
+            {
+                return Ok(revision.id);
+            }
+
+            if revisions.is_last_page {
+                break;
+            }
+
+            revisions.page_index = revisions.page_index + 1;
```
rajatjindal, line 523: should we probably do `fn list_revision_page(page_index)` instead? changing the previously fetched response does not feel right. But then i am a total noob in Rust, so please ignore me if it does not make sense

src/commands/deploy.rs
```diff
@@ -505,20 +505,30 @@ impl DeployCommand {
...
+            revisions = cloud_client.list_revisions_next(&revisions).await?;
```
rajatjindal, line 524: should we probably do fn list_revision_page(page_index) instead? changing the previously fetched response does not feel right. But then i am a total noob in Rust, so please ignore me if it does not make sense

rajatjindal: or may be move the responsibility to increment the page_index within the function

itowlson: The idea of passing the revisions object back was that it could the index and page size and whatever else might needed, the caller didn't need to know.  But if we have to mess with the object then that becomes less attractive.  I think you're right, I called it `previous` so let the callee sort it.

itowlson: Update pushed, could you give it yet another go please?  Thanks!

crates/cloud/src/client.rs
```diff
@@ -273,6 +273,19 @@ impl Client {
...
     }
+
+    pub async fn list_revisions_next(
+        &self,
+        previous: &RevisionItemPage,
+    ) -> anyhow::Result<RevisionItemPage> {
+        api_revisions_get(
+            &self.configuration,
+            Some(previous.page_index + 1),
+            Some(previous.page_size),
```
rajatjindal, line 284: one more small suggestion, do you think we should pass `None` as `page_size` and let server use  the default `page_size`

src/commands/deploy.rs
rajatjindal:
```
Uploading test-app version 0.1.0+rb9f6446c...
Deploying...
total: 107 page_index: 0 page_size: 50 is_last_page: false
total: 107 page_index: 1 page_size: 50 is_last_page: false
total: 107 page_index: 2 page_size: 50 is_last_page: true
```

i think this is working fine now.

crates/cloud/src/client.rs
itowlson: We could, but doing it this way completely decouples the "next" function from how it got started.  For example, suppose we later modified the starter function to allow the caller to control the page size - we'd then need to modify this function as well (and might not realise until someone got burned by it!).  Doing it this way costs us nothing and breaks a possible future coupling.

rajatjindal:
we probably need to  add  this for `list_channels` also. I ran into that today during my testing  :)

itowlson:
Added channel IDs.

bacongobbler:
Manually tested with the latest commit. LGTM!

bacongobbler (approved):

PR #977, opened 2022-12-15
chore(*): update versions to v0.7.0

michelleN:
+ for v0.7.0 release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -555,7 +555,7 @@ dependencies = [
 
 [[package]]
 name = "cloud"
-version = "0.6.0"
+version = "0.7.0"
 dependencies = [
  "anyhow",
  "cloud-openapi",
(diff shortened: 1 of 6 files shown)


fibonacci1729 (approved):

lann (approved):

etehtsea:
Should this one also be updated - https://github.com/fermyon/spin/blob/main/templates/Makefile#L1?

michelleN:
Good catch @etehtsea - let me check if the release action updates the Makefile or if we need to update the release document with this step.

michelleN:
Looks like the SDK_VERSION is getting set correctly through the environment in the release action but I went ahead and updated templates/Makefile for consistency. We should explore removing what it is set to entirely in the Makefile or updating it via the release action after this release. Thanks @etehtsea .

etehtsea:
Also, `make check-rust-examples` generates some changes in the example's lock files.

PR #982, opened 2022-12-15
They said Stonehenge would stand forever

itowlson:
The Ubuntu 18.04 GitHub runner image is being deprecated and is being randomly browned out.  Including today.  This ~~removes it from~~ replaces it with 20.04 in the release action.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         config:
           - {
-              os: "ubuntu-18.04",
+              os: "ubuntu-20.04",
               arch: "amd64",
               extension: "",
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -13,16 +13,6 @@ jobs:
     strategy:
       matrix:
         config:
-          - {
-              os: "ubuntu-18.04",
```
lann, line 17: This is our main linux release. Bump to 20.04 instead?

itowlson: Oof, thanks for the catch.  For some reason I thought we released off latest.  Done!

lann (approved):

michelleN (approved):

PR #984, opened 2022-12-16
If config key isn't in Vault, try other providers or default

### LINKED ISSUE #983: Variable default value ignored and error thrown when Vault Config Provider is used
In my Spin app I've configured three variables (**it's important to have at least one variable configured which is not stored in Vault as secret**)

```toml
[variables]
a = { default = "foo" }
b = { required = true }
c = { required = true }
```
Variable `b` exists as secret in Vault

```bash
vault kv get secret/b
======== Secret Path ========
secret/data/b

======= Metadata =======
Key                Value
---                -----
created_time       2022-12-15T22:11:11.791431757Z
custom_metadata    <nil>
deletion_time      n/a
destroyed          false
version            1

==== Data ====
Key      Value
---      -----
value    Baz
```

Component configuration is just forwarding the variables:

```toml
[[component]]
id = "some"
source = "some.wasm"
[component.config]
a = "{{ a }}"
b = "{{ b }}"
c = "{{ c }}"
```

When trying to access the configuration data in my component as shown here:

```rust
let a = config::get("a")?;
let b = config::get("b")?;
let c = config::get("c")?;
```
Before starting the Spin app, variable `c` is set as env var.

```bash
export SPIN_APP_C="Baz"
spin build --up --follow-all --runtime-config-file runtime_config.toml
```
When triggering the component, the following error is thrown:

```bash
The Vault server returned an error (status code 404)
2022-12-15T22:25:09.024940Z ERROR rustify::client: error=Server returned error
2022-12-15T22:25:09.024972Z ERROR rustify::endpoint: error=Server returned error
2022-12-15T22:25:09.024982Z ERROR vaultrs::api: Detected errors in API response: []
2022-12-15T22:25:09.024988Z ERROR vaultrs::kv2: error=The Vault server returned an error (status code 404)
```

Although `a` has a default value, I have to provide `a` explicitly by either providing `SPIN_APP_A` as environment variable or by storing `a` in Vault.

The app works as expected **only** if all variables are provided by a **config data provider**.

IMO the `config::get` should not throw when the underlying variable has a default value.

itowlson:
Fixes #983.

cc @FrankYang0529 (can't add you as a reviewer, sorry)

This was manually tested by modifying `examples/config-rust` with a runtime-config file.  If we have an automated test for Vault then I can add it to that.

For the time being I have replicated the current error behaviour when the error is anything other than "not found."  This will mean that if, say, Vault is down, then _all_ config lookups will fail (including ones with defaults).  This seems reasonable because we don't know if it's important to security that a value come from Vault - but the same argument applies to the 404 case.  So we might need to have think about whether we need some finer granularity longer term.

### DIFF
--- a/crates/config/src/provider/vault.rs
+++ b/crates/config/src/provider/vault.rs
@@ -1,8 +1,9 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
 use async_trait::async_trait;
 use serde::{Deserialize, Serialize};
 use vaultrs::{
     client::{VaultClient, VaultClientSettingsBuilder},
+    error::ClientError,
     kv2,
 };
 
@@ -51,7 +52,12 @@ impl Provider for VaultProvider {
             Some(prefix) => format!("{}/{}", prefix, key.0),
             None => key.0.to_string(),
         };
-        let secret: Secret = kv2::read(&client, &self.mount, &path).await?;
-        Ok(Some(secret.value))
+        match kv2::read::<Secret>(&client, &self.mount, &path).await {
+            Ok(secret) => Ok(Some(secret.value)),
+            // Vault doesn't have this entry so pass along the chain
+            Err(ClientError::APIError { code: 404, .. }) => Ok(None),
+            // Other Vault error so bail rather than looking elsewhere
+            Err(e) => Err(e).context("Failed to check Vault for config"),
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


FrankYang0529:
LGTM, thanks.

We have an automated test
https://github.com/fermyon/spin/blob/1b06bf08a3fa946bd1a168ed02813b4b0fe4d790/tests/integration.rs#L736-L847

lann (approved):

fibonacci1729 (approved):

itowlson:
Thanks @FrankYang0529 - added something to the test for it.  (Also I fixed what looks like a glitch in the test - do let me know if I did wrong and I will revert!)

PR #991, opened 2022-12-16
chore: Clean up some 'allow' attrs

lann:
Some were no longer necessary, a couple were easy to fix, and one was moved to a more specific scope.

### DIFF
--- a/crates/cloud/src/client.rs
+++ b/crates/cloud/src/client.rs
@@ -3,11 +3,9 @@ use cloud_openapi::{
     apis::{
         self,
         apps_api::{api_apps_get, api_apps_id_delete, api_apps_post},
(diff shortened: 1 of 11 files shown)


crates/cloud/src/client.rs
```diff
@@ -3,11 +3,9 @@ use cloud_openapi::{
     apis::{
         self,
         apps_api::{api_apps_get, api_apps_id_delete, api_apps_post},
-        auth_tokens_api::api_auth_tokens_post,
```
lann, line 6: These were all unmasked by https://github.com/fermyon/spin/pull/991/files#diff-7ef37faa600100d1f893e4befbc8ce6bf7a0b555affb0bbefff2879af967601cL1

fibonacci1729 (approved):
LGTM

PR #992, opened 2022-12-17
Initial add of http-php template

technosophos:
This adds a PHP template that uses the Wagi executor.

It uses the latest PHP build from VMware Wasm Labs: https://github.com/vmware-labs/webassembly-language-runtimes/releases

Signed-off-by: Matt Butcher <matt.butcher@fermyon.com>

### DIFF
--- a/templates/http-php/metadata/spin-template.toml
+++ b/templates/http-php/metadata/spin-template.toml
@@ -1,6 +1,6 @@
 manifest_version = "1"
 id = "http-php"
-description = "HTTP PHP environment"
+description = "HTTP request handler using PHP"
 
 [parameters]
 project-description = { type = "string",  prompt = "Project description", default = "" }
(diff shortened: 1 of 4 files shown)


itowlson:
The reason for the test failure is that at the moment the template unit tests have a constant for how many templates they expect to find in the Spin repo.  So it needs to be updated every time we add a new template.  It's a bit of a maintenance pain but it is what it is for now.

https://github.com/fermyon/spin/blob/1439b95fea89a4863c05ae489f1a9c0de17f27b9/crates/templates/src/manager.rs#L417

templates/http-php/metadata/spin-template.toml
```diff
@@ -0,0 +1,8 @@
+manifest_version = "1"
+id = "http-php"
+description = "HTTP PHP environment"
```
itowlson, line 3: For other templates like this we've used the form "HTTP request handler using [language]"
```suggestion
description = "HTTP request handler using PHP"
```

itowlson (approved):
LGTM

technosophos:
Okay, I think I got it all fixed up. Thanks @itowlson.

technosophos:
I will leave it to someone else to merge, since I have given up the pretense of being a maintainer. 😀

Issue #942, opened 2022-12-06
Add aliases to `plugin` and `templates` subcommands for plurality


kate-goldenring:
Having aliases for the singular and plural of `plugin/s` and `template/s` subcommands would help reduce user error.
See https://docs.rs/clap/latest/clap/struct.Arg.html#method.alias

raymundovr:
Hi, can I work on this one?

itowlson:
@raymundovr Sure - thanks!

raymundovr:
@itowlson have submitted a PR, let me know if that's a correct approach.

itowlson:
I think it depends.  Were you thinking of sending another PR to change the 'default' plurality @raymundovr?  If so, then there's little point raising a new issue for 'part 2' (unless that would be useful to you).  But if we're parking this as 'done for now' - which is 100% reasonable, as your merged contribution fulfils what was asked for - then we should close.

raymundovr:
hi @itowlson, just wanted to corroborate after the PR comments, up to you.

itowlson:
Cool, I'll close this - thanks!

PR #994, opened 2022-12-19
feat: add aliases for plugin and templates subcommands for plurality

raymundovr:
Add aliases to `plugin` and `templates` subcommands
Issue #942 

Signed-off-by: Raymundo Vásquez Ruiz <raymundo.vr@protonmail.com>

### DIFF
--- a/src/bin/spin.rs
+++ b/src/bin/spin.rs
@@ -43,7 +43,7 @@ fn version() -> &'static str {
     version = version(),
 )]
 enum SpinApp {
-    #[clap(subcommand, alias="template")]
+    #[clap(subcommand, alias = "template")]
     Templates(TemplateCommands),
     New(NewCommand),
     Add(AddCommand),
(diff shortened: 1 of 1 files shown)


itowlson:
Thanks for this!  Could you run `cargo fmt` and `cargo clippy` please?  We require all Rust code to be clean with these two checkers.  Thanks!

itowlson (approved):
LGTM - thanks!

@kate-goldenring Do you think we should rationalise it so that both commands have the same "default" plurality (with the aliases so both forms are accepted regardless)?  I don't have a particular view, just raising it for discussion.

kate-goldenring:
@itowlson this is a good point. In aim of consistency, we could switch all references of `Plugin` to `Plugins` (the file names and directories for plugins is already plural) and switch the alias to `plugin`. We could do this in a subsequent PR.

PR #995, opened 2022-12-19
chore(*): v0.7.1 version bumps

vdice:
* bumps versions for the forthcoming v0.7.1 release now that the [milestone has completed](https://github.com/fermyon/spin/milestone/6)

### DIFF
--- a/docs/content/release-process.md
+++ b/docs/content/release-process.md
@@ -9,7 +9,7 @@ url = "https://github.com/fermyon/spin/blob/main/docs/content/release-process.md
 To cut a release of Spin, you will need to do the following:
 
 1. Create a pull request that changes the version number for your new version
-   (e.g. `0.7.0` becomes `0.7.1`)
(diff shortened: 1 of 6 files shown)


docs/content/release-process.md
```diff
@@ -9,16 +9,16 @@ url = "https://github.com/fermyon/spin/blob/main/docs/content/release-process.md
 To cut a release of Spin, you will need to do the following:
 
 1. Create a pull request that changes the version number for your new version
-   (e.g. `0.7.0` becomes `0.7.1`)
+   (e.g. `0.7.1` becomes `0.7.2`)
```
itowlson, line 12: We don't need to update the example numbers each time do we?!

itowlson (approved):
LGTM - thanks!

docs/content/release-process.md
vdice: No, I suppose not :).  I can remove the changes to this file if we'd like, so we don't get into the habit/expectation.

PR #998, opened 2022-12-20
perf: use wasmtime's pooling instance allocator

tschneidereit:
For environments like Spin that create many wasm instances with a short lifetime, the pooling allocator can substantially improve performance, in particular in high-concurrency setups.

This commit enables the pooling allocator, improving performance for the `spin-executor/sleep-1ms/concurrency-{8,32}` benchmarks by about 85%. E.g. the `concurrency-8` benchmark on my machine goes from about 2.7ms to about 350us per iteration. These numbers are all on macOS, and should be better on Linux, where the kernel provides more efficient syscalls for the pooling allocator to use.

The pooling allocator requires choosing limits on the resources an instance can use. This patch makes those configurable, and sets the following defaults:
- max number of memories: 1
- max memory size, in 64kb wasm pages: 2048 (128MB)
- max number of tables: 1
- max number of table entries: 100,000

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -64,7 +64,12 @@ impl Config {
         &mut self.inner
     }
 
-    /// Enable the Wasmtime compilation cache with the given path, if any, to load configuration from.
+    /// Enable the Wasmtime compilation cache. If `path` is given it will override
+    /// the system default path.
+    ///
+    /// For more information, see the [Wasmtime cache config documentation][docs].
+    ///
+    /// [docs]: https://docs.wasmtime.dev/cli-cache.html
     pub fn configure_cache(&mut self, config_path: &Option<PathBuf>) -> Result<()> {
         match config_path {
             Some(p) => self.inner.cache_config_load(p)?,
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -75,6 +75,7 @@ impl Config {
     }
 
     /// Enable or update parameters for the pooling instance allocator.
+    #[cfg(not(target_os = "windows"))]
     pub fn enable_pooling(
         &mut self,
         max_memories: u32,
@@ -99,6 +100,7 @@ impl Config {
     }
 
     /// Disable the pooling instance allocator.
+    #[cfg(not(target_os = "windows"))]
     pub fn disable_pooling(&mut self) -> &mut Self {
         self.inner
             .allocation_strategy(wasmtime::InstanceAllocationStrategy::OnDemand);
@@ -113,6 +115,7 @@ impl Default for Config {
         inner.epoch_interruption(true);
 
         let mut config = Self { inner };
+        #[cfg(not(target_os = "windows"))]
         config.enable_pooling(
             DEFAULT_INSTANCE_MEMORIES,
             DEFAULT_INSTANCE_MEMORY_PAGES,
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -13,8 +13,9 @@ mod limits;
 mod store;
 
 use std::{
+    path::PathBuf,
     sync::{Arc, Mutex},
-    time::Duration, path::PathBuf,
+    time::Duration,
 };
 
 use anyhow::Result;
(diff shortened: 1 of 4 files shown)


tschneidereit:
ah right, I had been meaning to look into that in more detail. One possible explanation could be that we're only measuring the time the instance is active, not sleep time?

lann:
I'd like to see how much of a performance improvement we get with a full load test before making this change. Given the pooling allocator's recent history of security vulnerabilities there is some real tradeoff to consider here.

crates/trigger/src/cli.rs
```diff
@@ -18,6 +18,10 @@ pub const FOLLOW_LOG_OPT: &str = "FOLLOW_ID";
 pub const WASMTIME_CACHE_FILE: &str = "WASMTIME_CACHE_FILE";
 pub const RUNTIME_CONFIG_FILE: &str = "RUNTIME_CONFIG_FILE";
 
+// pub const DISABLE_POOLING_ALLOCATOR: &str = "DISABLE_POOLING_ALLOCATOR";
+// pub const POOLING_SLOTS: &str = "POOLING_SLOTS";
+// pub const POOLING_: &str = "POOLING_";
```
radu-matei, line 23: Are the comments here needed?

crates/trigger/src/cli.rs
```diff
@@ -56,6 +60,47 @@ where
     )]
     pub cache: Option<PathBuf>,
 
+    /// Disable Wasmtime's pooling instance allocator.
+    #[cfg(not(target_os = "windows"))]
+    #[clap(long = "disable-pooling")]
+    pub disable_pooling: bool,
```
radu-matei, line 66: Really appreciate that this is configurable on the command line. Do you have any considerations re: this being opt-in vs. opt-out by default? 

Specifically, should one think about this as an extra optimization to enable when needed, or is it something we should encourage everyone to always use?

crates/trigger/src/cli.rs
tschneidereit: Nope, thanks! Those were left over from a previous approach to making this configurable.

crates/trigger/src/cli.rs
tschneidereit: That's a good question. Given the pretty massive wins this brings, I'd argue it should be enabled by default. However, this isn't an entirely clear-cut choice, because it introduces a new failure mode, where components use more resources than are allocated for each instance in the pooling allocator configuration.

Additionally, the pooling allocator really only makes sense for use cases with short component lifetimes. I think that should apply to Spin pretty generally though?

lann:
I'm convinced. Spin has always had a concurrency bottleneck that I hadn't quite figured out. Looks like this was probably (a large part of) it!

endocrimes:
> I'm convinced.

I think I am too. This seems super promising.

crates/core/src/lib.rs
```diff
@@ -47,13 +63,57 @@ impl Config {
...
+    }
+
+    /// Enable or update parameters for the pooling instance allocator.
+    #[cfg(not(target_os = "windows"))]
+    pub fn enable_pooling(
+        &mut self,
+        max_memories: u32,
+        max_memory_pages: u64,
+        max_tables: u32,
+        max_table_entries: u32,
```
lann, line 84: I wonder if we can do a better job of selecting reasonable defaults here. The implications of these settings aren't going to be obvious to most people.

crates/core/src/lib.rs
```diff
@@ -47,13 +63,57 @@ impl Config {
...
+        match config_path {
+            Some(p) => self.inner.cache_config_load(p)?,
+            None => self.inner.cache_config_load_default()?,
+        };
+
+        Ok(())
+    }
+
+    /// Enable or update parameters for the pooling instance allocator.
+    #[cfg(not(target_os = "windows"))]
```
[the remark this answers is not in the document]
tschneidereit: No reason, turns out: I misremembered and thought the pooling allocator wasn't usable on Windows, but appear to have been wrong, so I removed these cfgs.

itowlson:
What's the status of this PR?

alexcrichton:
I'm picking this back up in https://github.com/fermyon/spin/pull/1721 (which is based on this PR)

Issue #936, opened 2022-12-05
Support TLS for outgoing Postgres


mikkelhegn:
Connecting to a Postgres server using 'outbound-pg' currently doesn't support TLS. https://github.com/fermyon/spin/blob/c368fce72261339b6de908e22b81cf21217a8181/crates/outbound-pg/src/lib.rs#L242

This prevents a Spin application from using e.g. [Neo](https://neon.tech/)

etehtsea:
I'll open a PR soon

etehtsea:
Should be fixed by https://github.com/fermyon/spin/pull/1003

PR #1003, opened 2022-12-21
feat: Add TLS connection support for outbound PG

etehtsea:
Underneath it conditionally uses `NoTls` when `sslmode=disable` and
sets up tls connector otherwise.

NOTE:
By default, `tokio_postgres` is using `sslmode=prefer`, so to
avoid [SSL setup overhead](https://github.com/fermyon/spin/pull/1003/files#diff-f65e2bbae4f787ee6d317acab2e0e1b4d583fe1ba676febf33162986ee978200R263-R264) `sslmode=disable` should be provided
explicitly.

Refs: https://github.com/fermyon/spin/issues/936

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2468,7 +2468,7 @@ version = "0.29.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "522f2f30f72de409fc04f88df25a031f98cfc5c398a94e0b892cabb33a1464cb"
 dependencies = [
- "base64 0.13.0",
+ "base64",
  "bigdecimal",
  "bindgen",
  "bitflags",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2501,9 +2501,9 @@ dependencies = [
 
 [[package]]
 name = "native-tls"
-version = "0.2.10"
+version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


itowlson (approved):
Looks good to me - thanks!

itowlson:
+ @lann for 👀 on the TLS side of things - I know this has been a thorn in our side elsewhere

lann (commented):
Shouldn't be a problem on Linux where `native-tls` will use `openssl` like other things do already. Probably worth checking on MacOS where iirc we also ran into TLS library issues before.

michelleN (commented):
Thanks @etehtsea. Would it be possible to write a test for this PR?

michelleN:
Thanks @etehtsea. What scenario/example did you use to manually test this PR? Were you able to get it working with neodb?

lann:
I just tested locally; appears to work mostly as expected. The one hiccup is that `sslmode=prefer` seems to behave the same as `sslmode=require`

michelleN (approved):
Manually tested using the c#/.NET + ElephantSQL [example](https://developer.fermyon.com/cloud/data-postgres#persistent-data-postgresql) by setting `sslmode=require` in the connection string and it worked for me as expected as well. Same set up without these changes (on main) errored as expected. 🎉

etehtsea:
> Thanks @etehtsea. What scenario/example did you use to manually test this PR? Were you able to get it working with neodb?

Yep, I've tested it against free [neon](https://neon.tech/) instance

PR #1008, opened 2022-12-21
ci(build.yml): only upload to codecov if fermyon org

vdice:
For now, only attempt to upload codecov reports if this is running on this `fermyon` origin, where we can guarantee a codecov token is present.  Forks may not have such a token and we don't wish to fail their CI runs on this detail.

(We _do_ wish to preserve the `fail_ci_if_error` toggle so that we're alerted when there are issues uploading when running from the `fermyon` org, hence keeping this set to true.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -317,9 +317,12 @@ jobs:
         env:
           RUST_LOG: spin=trace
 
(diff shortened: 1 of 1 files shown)


itowlson (approved):
Yay, thanks!

PR #1014, opened 2023-01-04
Add prototype for distributing Spin applications using OCI

radu-matei:
See SIP — #1033 


docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+
+Changing the distribution mechanism for Spin applications is a breaking change
+for the project - to address this, the change should be implemented gradually,
+giving users the tool to migrate an application from Bindle to an OCI registry.
+
+This would be done in two steps: pulling the application from Bindle, then
+pushing it to the desired OCI registry:
+
+```bash
+$ spin bindle pull --bindle-server <server> --bindle-username <username> --bindle-password <password> [bindle-name]
```
itowlson, line 106: If we want to do this then we need a better sense of how this fits together - what is the bindle pulled into, what is push pushing from?  Currently `oci push` appears to push from a local manifest, similar to `bindle push` - is the idea to reconstitute a local app from a bindle?  This feels a bit odd given that we are going between two content-addressed formats.  But I could be wrong

An alternative approach might be to create a `bindle2oci` Spin plugin that pulls down the bindle, massages it into OCI layers, and pushes that directly.  This might also, longer term, help with migrating applications in the cloud, where we would want to do a migrate apps internally without a need for users to pull and push.

It might also be good to get a sense for how many users are affected by this - how many people are running apps in the Platform and would not be willing/able to redeploy from source?  That is, to what extent does this need to be a user tool vs a cloud internal one?

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+
+### Impact to `spin deploy`
+
+The first step in `spin deploy` is pushing the application to the registry. In
+line with the gradual approach to this change, `spin deploy` will also change to
+publish the application to an OCI registry instead, and by the time this change
+will be default, Fermyon Platform and Fermyon Cloud will support this change.
+
+The goal is to have at least one minor release of Spin with `spin oci` before
+updating the `spin deploy` behavior to expect an OCI registry.
```
itowlson, line 276: To be clear, we should offer a `spin deploy --oci` as soon as Platform and Cloud support it, not just `spin oci`.  This allows for testing of the Platform/Cloud OCI integration.  (`spin oci push` isn't enough because we need to make sure Hippo/Cloud app creation and update works correctly.)  [This is I think what your first para says, but the second one seems more ambiguous.]

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+
+### Implementation status
+
+The `spin oci push`, `spin oci pull`, and `spin oci run` commands are currently
+implemented in this prototype. The implementation uses the
+[`oci-distribution`](https://github.com/krustlet/oci-distribution) crate from
+Krustlet to interact with a container registry (and currently uses a fork that
+should be patched upstream).
+
+The internals of the loaders need additional work before being merged, and the
```
itowlson, line 286: Are you amenable to merging the SIP separately from the code?  Just thinking it will be easier to collaborate on the implementation (and reconcile with parallel changes) in PRs to main than to a branch in your fork.

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+files = ["my-file.json"]
+allowed_http_hosts = ["https://hooks.slack.com"]
+[component.trigger]
+route = "/..."
+```
+
+Distributing it to the GitHub Container Registry:
+
+```bash
+$ spin oci push ghcr.io/radu-matei/spin-example:v1
```
itowlson, line 159: I notice here you're passing an explicit version.  Is that how you see this working in production?

Versioning does seem like an area we should define carefully (or at least identify as requiring mindful discussion during implementation).  Immutability was both a benefit and a curse of Bindle, and we ended up using bonus versions (content or random) as a way to update the bindle store without revving the Spin app version.  Will we derive the OCI version from the Spin app version?  Will that result in mutation?  What consequences will that have?

(I realise the consequences may not be seen until we look at the Platform/Cloud side.  But by that point we will already have established how versions are set/used.  So I do reckon we need to think about it now.)

itowlson (commented):
I haven't looked at the code yet but the SIP is really good.  The main thing I'd like to understand better is how we see versioning working.  Otherwise looking good!

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+$ spin oci pull ghcr.io/<username>/my-spin-application:v1
+INFO spin_publish::oci::client: Pulled ghcr.io/<username>/my-spin-application:v1@sha256:9f4e7eebb27c0174fe6654ef5e0f908f1edc8f625324a9f49967ccde44a6516b
+
+$ spin oci run ghcr.io/radu-matei/my-spin-application:v1
+INFO spin_publish::oci::client: Pulled ghcr.io/<username>/my-spin-application:v1@sha256:9f4e7eebb27c0174fe6654ef5e0f908f1edc8f625324a9f49967ccde44a6516b
+Serving http://127.0.0.1:3000
+```
+
+The commands and arguments shown above are a suggestion - eventually, the
+`spin oci run` functionality should be integrated into `spin up`.
```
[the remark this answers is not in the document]
radu-matei: That makes sense to me, I will update the SIP accordingly.

docs/content/sips/008-using-oci-registries.md
radu-matei: We have used Bindle as the default mechanism for distributing Spin applications since the very first launch — because of that, I would be inclined to have this tool public.

The point re: this being a plugin rather than something built in is excellent — it makes so much more sense to pull that as it is needed. Great suggestion!

On pulling from Bindle, my assumption had been that it was easier to fetch a standalone bindle and push that, but I might be wrong. The two step approach here was based on that — happy to consider a single step approach that just translates the two formats.

TL; DR for me here — I feel strongly that, given that we will surely need to use this functionality, making it available to everyone makes a lot of sense. Building it as a plugin is a perfect use for a plugin, and I will update the SIP to reflect that.

In that case, the experience would be something like:

```
spin bindle2oci --bindle-server <server> --bindle-username <username> --bindle-password <password> ---bindle-name <name> --oci <new-reference>
```

How does that sound?

docs/content/sips/008-using-oci-registries.md
radu-matei: Good point, the second paragraph there is indeed a bit ambiguous.
Will try to reword it.

docs/content/sips/008-using-oci-registries.md
radu-matei: Absolutely.
With that in mind, I would propose we go through a full round of reviews and updates here, and when we are happy with the state of the SIP, break it out into its own PR.

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+{
+  "schemaVersion": 2,
+  "config": {
+    "mediaType": "application/vnd.fermyon.spin.application.v1+config",
+    "digest": "sha256:b36160facea3076ad136c09bd4975a429805945ad313b4674363841d5a7f66a0",
+    "size": 643
+  },
+  "layers": [
+    {
+      "mediaType": "application/vnd.wasm.content.layer.v1+wasm",
```
radu-matei, line 197: We might consider using the 
`application/vnd.module.wasm.content.layer.v1+wasm` content media type for Wasm modules; this might make more sense as we could add `wasm.component` in the future.

docs/content/sips/008-using-oci-registries.md
radu-matei: This is the first can of worms of using OCI for this.
I debated whether to go on a tangent in the SIP, but decided against it...

As you point out in the comment, tags in Bindle were immutable, whereas tags in OCI are famously mutable (see `latest`...).


The first prototype for this had the following behavior: 
- if `spin oci push` had no arguments, the full OCI reference would be derived from `<app-name>:v<version>`
- if `spin oci push` had 1 argument, the OCI reference, that would be used instead.

For `spin oci push`, which would be the method for distributing applications, decoupled from Platform/Cloud, users being able to control the tag independently is something I would like to have.

Platform/Cloud represents an opinionated way to distribute and deploy the application. The current behavior of using the build metadata to push a new tag makes sense as the starting point IMO.

What do you think?

docs/content/sips/008-using-oci-registries.md
itowlson: Sounds good!

docs/content/sips/008-using-oci-registries.md
itowlson: You should _totally_ go on that 'tangent' - it's really useful for a SIP to discuss alternatives and trade-offs, and explain why a particular design choice was preferred.  Otherwise we hit a situation where we are on the wrong side of the chosen trade-off and we have to rediscover why it's better than the alternatives, usually accompanied by much grumbling!  And I do think this is an important and subtle design point, not a tangent at all!

Maybe I'm focused too much on interactive scenarios, but I do like some ability to derive the name and version from the manifest, if for no other reason than to avoid typos.  People are just not going to get it right every time.

But mutability does make that dangerous.  You don't want to do a push and find you've blatted a stable tag with an experimental release because you forgot to bump the version.  We could resume the Bindle strategy of appending a bonus sub-version, and require users to opt out by passing the `--no-build-metadata` (really `--this-really-is-the-real-version-for-real`) flag.  But then maybe we're trying to recreate Bindleisms on top of a fundamentally different model.  Maybe we embrace mutability, and if consumers want stability they can reference by digest instead of version.

(There's also a reasonable argument that "well really you should only be publishing from CI, we intentionally made it so you have to be more deliberate when pushing random code from a random laptop," but again, if that's a factor then we should capture it.)

In short, I am conflicted.  Sorry for the ramble, I am very much thinking out loud!

Could you say more about why you decided against your first approach?  It seems like the 1-arg version would have allowed "users ... to control the tag independently" but not _required_ them to manage it on every single interaction.

(Oh, and re Platform/Cloud/deploy, yeah you're right, in a sense the whole versioning story is kind of an internal implementation detail, at least for now.  Though it will become significant again if we bring back the auto-rev channels from the original Hippo.)

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+
+This operation created the following local cache directory structure:
+
+```bash
+$ tree /Users/radu/Library/Application\ Support/fermyon/registry
+└── oci
+    ├── data
+    │   └── sha256:a4699e4f9ef3f4922f38f0d017aa26438908f38caf020a739e0ee27fe796eb02
+    ├── manifests
+    │   └── ghcr.io
```
kate-goldenring, line 174: why have all of these subdirectories rather than just one subfolder to the manifest and config?

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+
+Note: the media types used are not final and *can* change based on the community
+standardization for Wasm modules in OCI registries.
+
+The manifest contains layers for the Wasm module for the component and for the
+one static asset referenced by the component. For each additional component, the
+Wasm module and static assets would be individual layers in the manifest above.
+
+Let's explore the OCI configuration object referenced in the manifest —
+`config.json` - it is a locked application manifest that Spin can use to run the
```
kate-goldenring, line 218: for app deletion (and garbage collection), I wonder if we need an in text way of tying the config to the manifest, maybe adding the path to the manifest in the config? Or the way it is now, we can assume that the manifest is located in the same directory as the config

docs/content/sips/008-using-oci-registries.md
kate-goldenring: is there any standardization on wasm oci layer names  yet (from Docker for example)?

kate-goldenring (commented):
Just looked at the SIP to start -- thanks for walking through it with an example!

docs/content/sips/008-using-oci-registries.md
radu-matei: There are a few implementations out there that distribute Wasm _modules_ using OCI:

- [`wasm-to-oci`](https://github.com/engineerd/wasm-to-oci) and [`oci-distribution`](https://github.com/krustlet/oci-distribution) use `application/vnd.wasm.content.layer.v1+wasm`
- [`solo-io/wasm/spec`](https://github.com/solo-io/wasm/blob/master/spec/spec.md) uses `application/vnd.module.wasm.content.layer.v1+wasm`
- Docker's preview appears to distribute them as `application/vnd.docker.container.image.v1+json`

In short, there is no clear standardization at this point.

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,298 @@
...
+compatible registry, pull an application locally, and run an application:
+
+```bash
+$ spin oci push ghcr.io/<username>/my-spin-application:v1
+INFO spin_publish::oci::client: Pushed "https://ghcr.io/v2/<username>/my-spin-application/manifests/sha256:9f4e7eebb27c0174fe6654ef5e0f908f1edc8f625324a9f49967ccde44a6516b"
+
+$ spin oci pull ghcr.io/<username>/my-spin-application:v1
+INFO spin_publish::oci::client: Pulled ghcr.io/<username>/my-spin-application:v1@sha256:9f4e7eebb27c0174fe6654ef5e0f908f1edc8f625324a9f49967ccde44a6516b
+
+$ spin oci run ghcr.io/radu-matei/my-spin-application:v1
```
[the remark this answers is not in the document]
radu-matei: The `spin oci run` command is a placeholder for the changes we need to make to `spin up --reference <oci-reference>`.

docs/content/sips/008-using-oci-registries.md
radu-matei: One subfolder per what?
Per registry? Repository? Tag?

The directory structure makes it simple to directly access the desired file (manifest or config) in O(1).

If there are more manifests from multiple registries, multiple accounts under the same registry, or multiple repositories of the same account, they would share subdirectories.

The Wasm modules and data files on the other hand are all in flat directories, as they are always addressed by digest.

Which subdirectories would you suggest flattening exactly?

docs/content/sips/008-using-oci-registries.md
radu-matei: As far as I can tell, you only need one of the manifest or locked application file (config) in order to do garbage collection, as they both are referring to the same layers / files on disk.

You would need to traverse all of them before marking something for garbage collection though, as layers can be shared across manifests / applications.

docs/content/sips/008-using-oci-registries.md
radu-matei: Fundamentally, `spin oci push` and `spin deploy` are two entirely different operations, that target different use cases — which is the reason I thought expanding in this SIP for the changes to `spin deploy` would have been unnecessary (or a 'tangent'), as I think it would require its own SIP.
However, you are right, expanding in this SIP is necessary, and I'd like to use this conversation to inform the SIP before updating it.


Let me expand a bit on why I think they are different operations, even if they happen to (and really _should_) use the same distribution mechanism.

`spin oci push` is intended to give a user the option to distribute their application using widely available registry services — and to be paired with `spin up`. The idea behind this pair of operations is to give a user as much flexibility and freedom in choosing their strategy for _how_ to do versioning, tagging, and to make their own decision in relation to tag mutability.
Some people really _do_ want and _need_ to use `:latest`, despite what opinions we might or might not have on the subject.
Deriving the name and tag from `spin.toml` when no explicit name and tag are defined is something we can do — however, this would imply two things:

- the name of the application now has to reference the registry URL, account, and repository — otherwise, trying to push to the registry/user/repository `my-simple-app` makes no sense.
- if the tag is derived from the version, we are restricting tags to semantic versioning, something that OCI does not dictate.

To sum up this up, having _the option_ to derive the name and tag from `spin.toml` sounds good — even though I might consider having a flag for this instead of it being the default (I don't have too strong of an opinion here though).

itowlson: That sounds good to me, and we can iterate on whether "derive name and version from manifest" should be default or opt-in - it doesn't affect the design in significant ways.  I am sure folks like Mikkel will have a view on this.  If we can capture this discussion into the SIP then I'm happy with the conclusion we've reached.  Thanks!

itowlson: (even if you want to capture by reference by linking back to the PR here... just something so that when we look at the SIP we can easily find the discussion)

docs/content/sips/008-using-oci-registries.md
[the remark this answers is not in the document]
radu-matei: Done.

docs/content/sips/008-using-oci-registries.md
radu-matei: Thanks for the suggestion, updated the SIP.

docs/content/sips/008-using-oci-registries.md
radu-matei: Tried to capture the context for this in the SIP.
Let me know if you think it does capture it or you think that section can be further expaded.

Thanks, the conversation so far on this has been really helpful!

docs/content/sips/008-using-oci-registries.md
radu-matei: Added a note in the SIP referencing the above.

docs/content/sips/008-using-oci-registries.md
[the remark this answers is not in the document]
radu-matei: Replaced this entirely with `spin up`.
Thanks!

docs/content/sips/008-using-oci-registries.md
kate-goldenring: I was wondering if it could be fully flattened as so:
```
ghcr.io-radu-matei-spin-example-v1
    │                   ├── config.json
    │                   └── manifest.json
```
But makes sense to navigate directories rather than try to predictably create a single directory path. I am converted to your way :)

docs/content/sips/008-using-oci-registries.md
kate-goldenring: Thanks for laying that out. I wonder if there should be cross organization work to look at standardizing this

itowlson (approved):
SIP looks good to me.  Thanks!

fibonacci1729 (approved):

docs/content/sips/008-using-oci-registries.md
```diff
@@ -0,0 +1,363 @@
...
+for a Spin application is `application/vnd.fermyon.spin.application.v1+config`,
+and each Wasm module and static asset from the Spin components becomes an individual
+*layer* in the resulting registry entity. Because each file and Wasm module becomes
+a separate *layer*, we can efficiently de-duplicate and distribute applications.
+
+The remaining question is representing the Spin application definition — Spin
+introduced the internal representation of a *locked application* - an
+intermediate representation of a Spin application that *can* have a way to
+content-address Wasm modules and static assets. The implementation for this SIP
+uses the Spin locked application as the OCI configuration object.
```
lann, line 135: This effectively makes the locked app format part of the compatibility surface of spin. We should take a critical look at the format to see if there are any breaking changes we want to make before committing to that.

lann (commented):
SIP LGTM

kate-goldenring:
SIP LGTM!

radu-matei:
As discussed earlier, https://github.com/fermyon/spin/pull/1033 breaks out the SIP into a separate PR so we can iterate on this easier.

Thanks for the reviews, everyone!

crates/cache/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+oci-distribution = { git = "https://github.com/radu-matei/oci-distribution", branch = "update-dockerhub-default-registry" }
+regex = "1.5.4"
+reqwest = "0.11"
+semver = "1.0"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+tempfile = "3.3.0"
+thiserror = "1.0.37"
+tokio = { version = "1.20", features = ["full"] }
+tracing = "0.1"
```
fibonacci1729, line 24: `tracing` is also a workspace dependency; should/could we use that here? `tracing = { workspace  = true }`

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /// The data directory for the current cache.
+    pub fn data_dir(&self) -> PathBuf {
+        self.root.join(DATA_DIR)
+    }
+
+    /// Get the file path to a manifest given a reference.
+    /// If the directory for the manifest does not exist, this will create it.
+    pub async fn manifest_for_reference(&self, reference: &str) -> Result<PathBuf> {
```
fibonacci1729, line 54: `reference: impl AsRef<str>`?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+        let reference: Reference = reference.parse().context("cannot parse reference")?;
+        let p = self
+            .manifests_dir()
+            .join(reference.registry())
+            .join(reference.repository())
+            .join(reference.tag().unwrap_or("latest"));
```
fibonacci1729, line 60: nit: could "latest" be a `const` like the dir names above?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        if !p.is_dir() {
+            fs::create_dir_all(&p).await?;
```
fibonacci1729, line 63: Should we `.context(...)` here?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        Ok(p.join("manifest.json"))
+    }
+
+    /// Get the file path to a config object given a reference.
+    pub async fn config_for_reference(&self, reference: &str) -> Result<PathBuf> {
```
fibonacci1729, line 70: `reference: impl AsRef<str>`

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+        let reference: Reference = reference.parse().context("cannot parse reference")?;
+        let p = self
+            .manifests_dir()
+            .join(reference.registry())
+            .join(reference.repository())
+            .join(reference.tag().unwrap_or("latest"));
+
+        if !p.is_dir() {
+            fs::create_dir_all(&p).await?;
```
fibonacci1729, line 79: I am once again asking if we should `.context(...)` here 😄

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        Ok(p.join("config.json"))
+    }
+
+    /// Write the contents in the cache's wasm directory.
+    pub async fn write_wasm(&self, bytes: &Vec<u8>, digest: &str) -> Result<()> {
```
fibonacci1729, line 86: Personally, i feel as if for `pub` apis the params should take the most flexible type, i.e. `impl AsRef<[u8]>/AsRef<str>`. I won't nit this anymore, feel free to ignore.

crates/cache/src/oci.rs
lann: ...but why?

crates/cache/src/oci.rs
lann: It should at least be `&[u8]`

crates/cache/src/oci.rs
fibonacci1729: Why not? Alternatively, why choose to be more strict in a public api? `impl AsRef<str>` has better caller ergonomics.

crates/publish/src/lib.rs
```diff
@@ -1,13 +1,7 @@
-#![deny(missing_docs)]
+// #![deny(missing_docs)]
```
fibonacci1729, line 1: friendly todo reminder

vdice (commented):
Only real feedback at this prototype stage is to see where we can make it obvious to users that this functionality is experimental and subject to change.

Also, where do we want to track the prototype's status of implementation of the corresponding SIP?

Lastly, could use a rebase to remove the SIP from this PR (now that it is in main).

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+        let manifest_path = self
+            .cache
+            .manifest_for_reference(&reference.to_string())
+            .await?;
+        let config_path = self
+            .cache
+            .config_for_reference(&reference.to_string())
+            .await?;
+        if !manifest_path.exists() || !config_path.exists() {
+            self.pull(&reference.to_string()).await?;
```
itowlson, line 209: The function name here made me think that it computed a descriptor - given that pulling is a heavyweight operation, could the name imply that it will pull if necessary?

(I'm not yet sure what this function is used for, so I don't have any suggestions yet, sorry.)

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+            Err(CredentialRetrievalError::NoCredentialConfigured) => Ok(RegistryAuth::Anonymous),
+            Err(CredentialRetrievalError::ConfigReadError) => Ok(RegistryAuth::Anonymous),
+            Err(e) => bail!("Error handling docker configuration file: {:?}", e),
+
+            Ok(DockerCredential::UsernamePassword(username, password)) => {
+                tracing::trace!("Found docker credentials");
+                Ok(RegistryAuth::Basic(username, password))
+            }
+            Ok(DockerCredential::IdentityToken(_)) => {
+                tracing::warn!("Cannot use contents of docker config, identity token not supported. Using anonymous auth");
```
itowlson, line 260: Do warnings get printed by default?  If not, this is going to cause surprise.

But even if so, the console tracer is pretty spewy.  It would be nice to present this as a clean, clear message rather than wrapped up in all the tracing rigmarole.  (Unfortunately, I don't think we have a great mechanism for this at the moment.)

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    /// Build the OCI client configuration given the insecure option.
+    pub fn build_config(insecure: bool) -> ClientConfig {
+        let protocol = if insecure {
+            oci_distribution::client::ClientProtocol::Http
+        } else {
+            oci_distribution::client::ClientProtocol::Https
+        };
+
+        oci_distribution::client::ClientConfig {
```
itowlson, line 274: Nit, but not sure why this is qualified here but not in the function signature.

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+            protocol,
+            ..Default::default()
+        }
+    }
+
+    /// Read the contents of a file and return it as a byte array.
+    pub async fn data(file: &Path) -> Result<Vec<u8>> {
+        let mut buf = Vec::new();
```
itowlson, line 282: This stanza seems to do the same as `tokio::fs::read()`?

crates/trigger/src/cli.rs
```diff
@@ -114,21 +117,43 @@ where
         let working_dir = std::env::var(SPIN_WORKING_DIR).context(SPIN_WORKING_DIR)?;
         let locked_url = std::env::var(SPIN_LOCKED_URL).context(SPIN_LOCKED_URL)?;
 
-        let loader = TriggerLoader::new(working_dir, self.allow_transient_write);
+        // TODO: I assume there is a way to do this with a single let mut loader: Box<dyn Loader>
+        // variable instead of the entire executor.
+        let executor: Executor = match self.oci {
+            true => {
+                let loader = OciTriggerLoader::new(working_dir, self.allow_transient_write).await?;
```
itowlson, line 124: Do the triggers really need to know about OCI?  One of the lovely things about the lockfile design was that it got the application to a place where _triggers didn't need to care_ - `spin up` just handed over the lockfile URL and the trigger took it from there.

(Maybe I've been avoiding some horrible truth about lockfiles but that was the way I understood it anyway!)

crates/trigger/src/cli.rs
```diff
@@ -114,21 +117,43 @@ where
         let working_dir = std::env::var(SPIN_WORKING_DIR).context(SPIN_WORKING_DIR)?;
         let locked_url = std::env::var(SPIN_LOCKED_URL).context(SPIN_LOCKED_URL)?;
 
-        let loader = TriggerLoader::new(working_dir, self.allow_transient_write);
+        // TODO: I assume there is a way to do this with a single let mut loader: Box<dyn Loader>
+        // variable instead of the entire executor.
```
itowlson, line 121: Would be good to dedupe for sure... `fn xxx_impl(loader: impl Loader, ...) -> Executor` then call it with a TriggerLoader or OciTriggerLoader depending on the value of `self.oci`?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Root directory for the cache instance.
+    pub root: PathBuf,
+}
+
+impl Cache {
+    /// Create a new cache given an optional root directory.
+    pub async fn new(root: Option<PathBuf>) -> Result<Self> {
+        let root = match root {
+            Some(root) => root,
+            None => dirs::config_dir()
```
itowlson, line 27: `config_dir` rather than, say, `data_local_dir` or `cache_dir`?

(I doubt we are too concerned about Windows at this stage but I'll mention that `config_dir` apparently roams on Windows... so...)

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+                .join(CONFIG_DIR),
+        };
+        let root = root.join(REGISTRY_CACHE_DIR).join(OCI_CACHE_DIR);
+        Self::ensure_dirs(&root).await?;
+
+        Ok(Self { root })
+    }
+
+    /// The manifests directory for the current cache.
+    pub fn manifests_dir(&self) -> PathBuf {
```
itowlson, line 38: Is it desirable that these are `pub`, or should we aspire to make the structure the concern of the cache, and hand out specific file paths or content on demand, rather than expecting consumers to construct file paths into the cache?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.root.join(WASM_DIR)
+    }
+
+    /// The data directory for the current cache.
+    pub fn data_dir(&self) -> PathBuf {
+        self.root.join(DATA_DIR)
+    }
+
+    /// Get the file path to a manifest given a reference.
+    /// If the directory for the manifest does not exist, this will create it.
```
itowlson, line 53: OCI manifest?  Spin application manifest?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+            .join(reference.tag().unwrap_or("latest"));
+
+        if !p.is_dir() {
+            fs::create_dir_all(&p).await?;
+        }
+
+        Ok(p.join("manifest.json"))
+    }
+
+    /// Get the file path to a config object given a reference.
```
itowlson, line 69: What is a config?  Is this a lockfile?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /// Write the contents in the cache's data directory.
+    pub async fn write_data(&self, bytes: &Vec<u8>, digest: &str) -> Result<()> {
+        fs::write(self.data_dir().join(digest), bytes).await?;
+        Ok(())
+    }
+
+    /// Ensure the expected configuration directories are found in the root.
+    /// └── fermyon
```
itowlson, line 98: Should this be `spin` rather than `fermyon`?  We have used both I fear, but for non-Platform/Cloud features should we try to stick to `spin`?

crates/cache/src/oci.rs
```diff
@@ -0,0 +1,130 @@
...
+    ///     └── registry
+    ///         └── oci
+    ///             └──manifests
+    ///             └──wasm
+    ///             └─-data
+    pub async fn ensure_dirs(root: &Path) -> Result<()> {
```
itowlson, line 104: A distinct function rather than taking `&self`?  Why would a consumer ensure the structure under a _different_ directory?

Also, does this need to be `pub`?  Why is calling this a consumer's problem instead of the cache doing it as needed?

crates/cache/src/oci.rs
itowlson, line 98: Is this using `fermyon` to stand for the root, or is `fermyon` under the root?  (Thinking about Brian's comment above.)

itowlson (changes requested):

crates/cache/src/oci.rs
radu-matei: This is using the same logic for getting the token as `deploy` and `login`.

crates/cache/src/oci.rs
radu-matei: The idea behind having `fermyon` is to have a single directory where we could have such configuration data that we could potentially share across tooling.

I would find it _really_ confusing if Spin would store configuration data in both `fermyon` and `spin`.

crates/cache/src/oci.rs
radu-matei: This should not be public, and is only used in `Cache::new`, when you don't have a `&self` yet.

Updated it to be private.

crates/cache/src/oci.rs
radu-matei: Updated these to be private and added functions to directly return the file path based on digest.

crates/trigger/src/cli.rs
radu-matei: It needs to know whether it is running an application from a relative file path or from the global cache.

crates/publish/src/oci/client.rs
radu-matei: Good catch, removing this.

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+            // Add a layer for each file referenced in the mount directory.
+            // Note that this is in fact a directory, and not a single file, so we need to
+            // recursively traverse it and add layers for each file.
+            let mut files = Vec::new();
+            for f in c.files {
+                let source = f
+                    .content
+                    .source
+                    .expect("file mount loaded from disk should contain a file source");
+                let source = spin_trigger::parse_file_url(source.as_str())?;
```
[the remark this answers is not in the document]
radu-matei: This is less about convenience for `spin_trigger::parse_file_url` (which can be easily duplicated) and more for `spin_trigger::build_locked_app`.

To expand on this a bit, my understanding is that we _really_ want the functionality to build a locked application configuration from a `spin.toml` to live in the `spin_app` crate. This is a refactoring I am not willing to take as part of this PR, which is intended to make as few structural changes as possible.

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+            .cache
+            .config_for_reference(&reference.to_string())
+            .await?;
+        fs::write(&c, &cfg).await?;
+
+        // If a layer is a Wasm module, write it in the Wasm directory.
+        // Otherwise, write it in the data directory.
+        for layer in manifest.layers {
+            // Skip pulling if the digest already exists in the wasm or data directories.
+            if std::fs::metadata(&self.cache.wasm_dir().join(&layer.digest)).is_ok()
```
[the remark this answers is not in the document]
radu-matei: See above.

crates/cache/src/oci.rs
lann: I would agree this should be `cache_dir`.

crates/cache/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+//! Local cache for Wasm modules and static assets pulled from a registry.
```
lann, line 1: The pattern of `spin_loader` grouping `local` and `bindle` implementations hasn't worked well and I don't think we should expand that to a new crate, even if this PR isn't refactoring `spin_loader`. I would suggest either packing this into `spin-loader` (e.g. `spin_loader::oci::cache`) or - as a step toward where I would hope to refactor all these things - a `spin-oci` crate.

crates/publish/src/oci/client.rs
radu-matei: Easy fix for this one — the function was not called anymore, so removed it altogether.

crates/cache/src/oci.rs
radu-matei: Updated all magic strings to be constants. Good catch.

crates/cache/src/oci.rs
itowlson: Oh, is this for configuration settings?  I thought it was where the cache was actually stored?

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+        let server = reference
+            .resolve_registry()
+            .strip_suffix("/")
+            .unwrap_or_else(|| reference.resolve_registry());
+
+        match docker_credential::get_credential(server) {
+            Err(CredentialRetrievalError::ConfigNotFound) => Ok(RegistryAuth::Anonymous),
+            Err(CredentialRetrievalError::NoCredentialConfigured) => Ok(RegistryAuth::Anonymous),
+            Err(CredentialRetrievalError::ConfigReadError) => Ok(RegistryAuth::Anonymous),
+            Err(e) => bail!("Error handling docker configuration file: {:?}", e),
```
bacongobbler, line 253: CredentialRetrievalError implements fmt::Display. We should call that so we get more helpful errors. Right now this is what I see:

Error: Error handling docker configuration file: HelperFailure

If we called .display(), I think we'd get this:

Error: Error handling docker configuration file: Credential helper returned non-zero response code

bacongobbler: https://github.com/keirlawson/docker_credential/blob/622bef657e2a0bc83abf1ac3f4ab3ed18f9ebf27/src/lib.rs#L24

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+                tracing::warn!("Cannot use contents of docker config, identity token not supported. Using anonymous auth");
+                Ok(RegistryAuth::Anonymous)
+            }
+        }
+    }
+
+    /// Build the OCI client configuration given the insecure option.
+    pub fn build_config(insecure: bool) -> ClientConfig {
+        let protocol = if insecure {
+            oci_distribution::client::ClientProtocol::Http
```
bacongobbler, line 269: The help text for --insecure shows "Ignore server certificate errors". It does not specify that it switches the protocol to HTTP.

We should either put this behind a different flag, or update the help text.

crates/cache/src/oci.rs
radu-matei: The intention is to use the same root directory for the configuration files _and_ the `registry` directory and cache.

itowlson: It might be useful to capture _why_ that is the intention - apologies if I seem obtuse but it's not jumping out at me.

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,288 @@
...
+        self.oci
+            .pull_blob(&reference, &manifest.config.digest, &mut cfg_bytes)
+            .await?;
+        let cfg = std::str::from_utf8(&cfg_bytes)?;
+        tracing::debug!("Pulled config: {}", cfg);
+
+        // Write the config object in `<cache_root>/registry/oci/manifests/repository:<tag_or_latest>/config.json`
+        let c = self
+            .cache
+            .config_for_reference(&reference.to_string())
```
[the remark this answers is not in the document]
radu-matei: This is a specific object called an "OCI config" — and we are using a serialized Spin locked application manifest for it.

crates/cache/src/oci.rs
radu-matei: The main reason is having _one_ directory where Spin (and potentially future Fermyon projects) hold state by default.

My assumption so far has been that we would rather have _one_ default directory for this, compared to scattering config, state, and other things around on a user's system — such that if someone wants to remove Spin, they would just delete this directory.

I would be interested in hearing why we would consider splitting all this information in multiple directories _by default_.

lann: It's how it's usually done. For XDG for example, you expect anything a user might directly configure to be under `.config/` and you expect to be able to delete `.cache/` without any permanent problems.

radu-matei: Ok — then it's a matter of changing the defaults. Happy to do that.

So is the global cache supposed to be placed in `dirs::cache_dir()` then, while everything else is in `dirs::config_dir()`?

itowlson: I have bad news I'm afraid, templates and plugins are already under `data_local_dir()`... though we can change that of course if we feel cache or config or something is better.

lann: Those are "data", arguably. These things are a little fuzzy...

itowlson: Yes, I agree - I'm just trying to figure out if we are trying to prioritise the "single base directory that you can delete in one command" over trying to fit things into XDG.

radu-matei: If we are already splitting things into multiple directories, the registry work should follow that.

crates/cache/src/lib.rs
radu-matei: Moved the separate cache crate into `spin_loader::oci::cache`.

crates/publish/src/oci/client.rs
radu-matei: Qualified it in the signature as well — there are potentially too many things called "client" and wanted to make sure this is clear.

crates/cache/src/oci.rs
radu-matei: This is now moving to the cache directory, and it will be side by side with templates.

If we decide on _where_ to move things, they should move together.

crates/publish/src/oci/client.rs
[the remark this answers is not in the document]
radu-matei: Tried to rename variables here to make things a bit more obvious.

radu-matei:
Thanks for the review and feedback so far, everyone!
These are the main changes compared to the initial state of the PR:

- moved the cache functionality inside `spin_loader::oci::cache` instead of a new crate (this should help refactoring the loaders cc @lann)
- the registry cache now stores all data under `$XDG_CACHE/spin/registry` (TBD final location based on the `~/.spin` vs. `XDG` debate). Instead of a directory for `oci`, this is renamed to `registry`, so that it can be used as a global cache (including for the HTTP component sources, which currently downloads the Wasm module on every `spin up`)
- `spin oci run` now works with all trigger types (including potentially external triggers) (cc @itowlson) and can accept environment variables
- the trigger CLI functionality still needs to be cleaned up — triggers currently expect an `--oci` flag, and I suspect this can be automatically inferred from the lockfile (i.e. if it has content digests instead of file paths, it is run from the registry). However, this is not something I want to address in this PR.
- this still does not integrate with `spin up`. I would like to do this gradually for the next iteration.

Things to implement compared to the SIP:

- `spin up` needs to be integrated
- `spin oci login` needs to be implemented
- the `spin bindle2oci` plugin needs to be implemented


The main question I have for this PR is: do we think this is in a state where we could merge it, make it explicit it is experimental, and iterate on the implementation and continue to implement the SIP, or do we want to iterate in this PR without merging?


(Also a question that is probably better suited for the SIP: `spin registry` vs. `spin oci` for the top-level commands?)

See below an example of the current state of this PR:

```
❯ spin oci push ghcr.io/radu-matei/spin-redis-hello:v3
INFO spin_publish::oci::client: Pushed "https://ghcr.io/v2/radu-matei/spin-redis-hello/manifests/sha256:34e4f0133464a4271ba2189ad24ec07095c61c8f79dc4d753d32b1a0dc69c73a"

❯ spin oci run --env abc=def ghcr.io/radu-matei/spin-redis-hello:v3
INFO spin_publish::oci::client: Pulled ghcr.io/radu-matei/spin-redis-hello:v3@sha256:34e4f0133464a4271ba2189ad24ec07095c61c8f79dc4d753d32b1a0dc69c73a
INFO spin_redis_engine: Connecting to Redis server at redis://localhost:6379
INFO spin_redis_engine: Subscribing component "rust-test" to channel "messages"
INFO spin_redis_engine: Received message on channel "messages"
Env: {"abc": "def"},
Hello from Rust, message: hello
```

```
❯ tree /Users/radu/Library/Caches/spin/registry
/Users/radu/Library/Caches/spin/registry
├── data
├── manifests
│   └── ghcr.io
│       └── radu-matei
│           └── spin-redis-hello
│               └── v3
│                   ├── config.json                 // OCI configuration object that contains the Spin lockfile
│                   └── manifest.json               // OCI manifest that references all Wasm modules and static assets
└── wasm
    └── sha256:7f894f2523cf5f7b4596942a29c6f8ee0d24468e1f8951f4e4864794caf283de
```

crates/loader/src/oci/cache.rs
```diff
@@ -0,0 +1,166 @@
...
+            false => bail!(format!(
+                "cannot find data file for digest {}",
+                digest.as_ref()
+            )),
+        }
+    }
+
+    /// Get the file path to an OCI manifest given a reference.
+    /// If the directory for the manifest does not exist, this will create it.
+    pub async fn manifest_for_reference(&self, reference: impl AsRef<str>) -> Result<PathBuf> {
```
itowlson, line 81: Suggestion: `oci_manifest_file(...)` - this makes it clear that it returns a _path_ rather than the manifest itself, and that this is the _OCI manifest_ rather than the Spin application manifest.

crates/loader/src/oci/cache.rs
```diff
@@ -0,0 +1,166 @@
...
+            fs::create_dir_all(&p)
+                .await
+                .context("cannot find directory for manifest")?;
+        }
+
+        Ok(p.join(MANIFEST_FILE))
+    }
+
+    /// Get the file path to the OCI configuration object given a reference.
+    pub async fn config_for_reference(&self, reference: impl AsRef<str>) -> Result<PathBuf> {
```
itowlson, line 102: Similarly I'd suggest `oci_config_file(...)`.

I am not sure quite where to draw the boundary but given that the cache knows about Wasm vs data, I wonder whether we should consider calling this `lockfile(...)` - do the layers above this care that the lockfile is implemented using the OCI config object?  (This is a question rather than a suggestion.)

crates/loader/src/oci/cache.rs
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    /// Get the file path to an OCI manifest given a reference.
+    /// If the directory for the manifest does not exist, this will create it.
+    pub async fn manifest_for_reference(&self, reference: impl AsRef<str>) -> Result<PathBuf> {
+        let reference: Reference = reference
+            .as_ref()
+            .parse()
+            .context("cannot parse reference")?;
```
itowlson, line 85: For this and the "cannot find" context (and same for the config function), suggest qualifying it as "OCI reference."  It is obvious at this point in the code but I don't know if it will be once it bubbles up to the command line.

Also, should they say which reference they are having problems with?  I suppose there should be only one OCI reference in play...?

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,261 @@
...
+        tracing::log::trace!("Reading data file from {:?}", file);
+        Ok(ImageLayer::new(
+            fs::read(&file).await?,
+            DATA_MEDIATYPE.to_string(),
+            None,
+        ))
+    }
+
+    /// Construct the registry authentication based on the reference.
+    pub fn auth(reference: &Reference) -> Result<RegistryAuth> {
```
itowlson, line 219: Do this and `build_config` need to be `pub`?  It feels like that stuff should be encapsulated in the `Client`.

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,261 @@
...
+        };
+
+        oci_distribution::client::ClientConfig {
+            protocol,
+            ..Default::default()
+        }
+    }
+
+    /// Read the contents of a file and return it as a byte array.
+    pub async fn data(file: &Path) -> Result<Vec<u8>> {
```
itowlson, line 258: Not sure quite why this exists - is it anything more than a renaming of `fs::read`?  If it is needed, suggest `read` or `read_file` so the name isn't entirely novel, and maybe a comment to explain why the wrapper is needed?

itowlson (commented):
A few style/naming comments but nothing crucial - I feel like the priority at this point is to get something in so folks can start bashing on it, and try for a tidying pass later?  _eyes that long loop body hungrily_

crates/publish/src/oci/client.rs
radu-matei: I changed all call sites of the function to use `tokio` directly in a previous commit, but forgot to remove the function.
Good catch.

itowlson (approved):

fibonacci1729 (approved):

crates/publish/src/oci/client.rs
```diff
@@ -0,0 +1,250 @@
...
+use spin_loader::oci::cache::Cache;
+use spin_manifest::Application;
+use tokio::fs;
+use walkdir::WalkDir;
+
+use std::path::{Path, PathBuf};
+
+// TODO: the media types for application, wasm module, and data layer are not final.
+const SPIN_APPLICATION_MEDIA_TYPE: &str = "application/vnd.fermyon.spin.application.v1+config";
+const WASM_LAYER_MEDIA_TYPE: &str = "application/vnd.wasm.content.layer.v1+wasm";
```
flouthoc, line 21: Just a drive-by-comment: I think media type can be added to image-spec, see discussion here: https://github.com/opencontainers/image-spec/pull/964#issuecomment-1278087823

radu-matei: Note that this is the media type of a layer, not of the top-level manifest or of the config.

In this scenario, the _artifact_ is a Spin application, and it contains Wasm layers.
We could use the upcoming referrer API, but that's not something stable AFAIK.

flouthoc: Thanks for the clarification. SGTM :)

PR #1017, opened 2023-01-04
Fix Vault config example to point to correct wasm source path

kate-goldenring:
The [example for using the Vault Config Provider](https://developer.fermyon.com/spin/configuration#vault-config-provider) points to the wrong path for the Wasm source.


lann:
Already fixed in #984

PR #1027, opened 2023-01-06
chore(docs): remove static dir and npm

vdice:
Cleanup after having moved Spin documentation to https://github.com/fermyon/developer and updating this docs site to be comprised solely of redirects.

* Removes npm/node logic previously used to build static assets
* Removes all items from the `static/` dir except for the logo.png file used in this repo's README
* Removes the static fileserver wasm module
* Removes the `downloads/` dir (Install script now tracked at https://github.com/fermyon/developer/blob/main/downloads/install.sh)


radu-matei:
The main readme renders the project logo from the static directory if I recall correctly.
If it would be good if we moved the current logos before removing the directory.

radu-matei (approved):

PR #1035, opened 2023-01-18
add outbound-kv host component, based on SQLite

dicej:
This is an experimental work-in-progress.  I'll switch it out of "draft" mode when it's ready for serious consideration.

### DIFF
--- a/crates/key-value/src/host_component.rs
+++ b/crates/key-value/src/host_component.rs
@@ -1,0 +1,47 @@
+use crate::{Impl, KeyValueDispatch};
+use spin_app::DynamicHostComponent;
+use spin_core::HostComponent;
+use std::{collections::HashMap, sync::Arc};
+
+pub const KEY_VALUE_STORES_METADATA_KEY: &str = "key_value_stores";
+
+pub struct KeyValueComponent {
+    impls: Arc<HashMap<String, Box<dyn Impl>>>,
+}
+
+impl KeyValueComponent {
+    pub fn new(impls: Arc<HashMap<String, Box<dyn Impl>>>) -> Self {
+        Self { impls }
+    }
+}
+
+impl HostComponent for KeyValueComponent {
+    type Data = KeyValueDispatch;
+
+    fn add_to_linker<T: Send>(
+        linker: &mut spin_core::Linker<T>,
+        get: impl Fn(&mut spin_core::Data<T>) -> &mut Self::Data + Send + Sync + Copy + 'static,
+    ) -> anyhow::Result<()> {
+        super::key_value::add_to_linker(linker, get)
+    }
+
+    fn build_data(&self) -> Self::Data {
+        KeyValueDispatch::new(self.impls.clone())
+    }
+}
+
+impl DynamicHostComponent for KeyValueComponent {
+    fn update_data(
+        &self,
+        data: &mut Self::Data,
+        component: &spin_app::AppComponent,
+    ) -> anyhow::Result<()> {
+        data.allowed_stores = component
+            .get_metadata::<Vec<String>>(KEY_VALUE_STORES_METADATA_KEY)?
+            .unwrap_or_default()
+            .into_iter()
+            .collect();
+
+        Ok(())
+    }
+}
--- a/crates/key-value/src/lib.rs
+++ b/crates/key-value/src/lib.rs
@@ -1,0 +1,126 @@
+use anyhow::Result;
+use std::{
+    collections::{HashMap, HashSet},
+    sync::Arc,
+};
+use table::Table;
+use wit_bindgen_wasmtime::async_trait;
+
+mod host_component;
+mod table;
+
+pub use host_component::KeyValueComponent;
+
+wit_bindgen_wasmtime::export!({paths: ["../../wit/ephemeral/key-value.wit"], async: *});
+
+pub use key_value::{Error, KeyValue, Store};
+
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl std::error::Error for Error {}
+
+#[async_trait]
+pub trait Impl: Sync + Send {
+    async fn open(&self, name: &str) -> Result<Box<dyn ImplStore>, Error>;
+}
+
+#[async_trait]
+pub trait ImplStore: Sync + Send {
+    async fn get(&self, key: &str) -> Result<Vec<u8>, Error>;
+
+    async fn set(&self, key: &str, value: &[u8]) -> Result<(), Error>;
+
+    async fn delete(&self, key: &str) -> Result<(), Error>;
+
+    async fn exists(&self, key: &str) -> Result<bool, Error>;
+
+    async fn get_keys(&self) -> Result<Vec<String>, Error>;
+}
+
+pub struct KeyValueDispatch {
+    pub allowed_stores: HashSet<String>,
+    impls: Arc<HashMap<String, Box<dyn Impl>>>,
+    stores: Table<Box<dyn ImplStore>>,
+}
+
+impl KeyValueDispatch {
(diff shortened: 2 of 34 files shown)


examples/rust-key-value/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+http = "0.2"
+spin-sdk = { path = "../../sdk/rust"}
+# The wit-bindgen-rust dependency generates bindings for interfaces.
+wit-bindgen-rust = { git = "https://github.com/bytecodealliance/wit-bindgen", rev = "cb871cfa1ee460b51eb1d144b175b9aab9c50aba" }
+
+[workspace]
+
+# Metadata about this component.
+[package.metadata.component]
+name = "spinhelloworld"
```
kate-goldenring, line 24: Does this metadata need to be changed to match the component name in the `spin.toml`?
```suggestion
name = "hello"
```

crates/trigger/src/lib.rs
```diff
@@ -95,6 +95,27 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
...
+                };
+
+                // TODO: Use relevant portion of app config here.
+
+                builder.add_host_component(key_value::KeyValueComponent::new(
+                    key_value::Config {
+                        configs: [(
+                            "".to_owned(),
+                            key_value::ImplConfig::Sqlite(key_value::sqlite::Config::Path(
+                                parent_dir.join("sqlite_key_value.db"),
```
kate-goldenring, line 111: Maybe add a comment here that for now only the file based SQLite implementation can be used. There is no current way to use in memory right, since we are only using default? This could also be a note around the `sqlite::Config` object

kate-goldenring, line 111: Consider making `sqlite_key_value.db` a constant.

crates/key-value/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+    Error::Runtime(format!("{err:?}"))
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn all() -> Result<()> {
+        let mut kv = KeyValueDispatch::new(Config {
```
kate-goldenring, line 139: is it worth using `tmpdir` to also test file path based Sqlite db rather than just in memory?

crates/key-value/src/host_component.rs
```diff
@@ -0,0 +1,28 @@
+use spin_core::HostComponent;
+
+use crate::{Config, KeyValueDispatch};
+
+pub struct KeyValueComponent {
```
kate-goldenring, line 5: I wonder if we are overloading the term `Config` here or whether we want to move away from using it

kate-goldenring (commented):
Some thoughts and comments. Overall looks pretty complete to me. Will look again once we've finalized the SIP

examples/rust-key-value/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+use spin_sdk::{
+    http::{Request, Response},
+    http_component,
+    key_value::{Error, Store},
+};
+
+#[http_component]
+fn handle_request(req: Request) -> Result<Response> {
+    // Open the default key-value store
+    let store = Store::open("")?;
```
itowlson, line 12: Is there any way we can hide the special empty string from consumers?  `Store::default()` and `Store::named/Store::open` maybe, though it might be a bit profligate?

dicej: Yes, that was my suggestion when we discussed it.  IIRC, @lann wanted to use the empty string so that runtime config files could refer to it explicitly, but perhaps we could just have special runtime config file syntax for configuring the default store?

lann: I'm fine with `Store::default()` being equivalent to `Store::open("")`. I don't recall being against it but if I was then I'm disagreeing with previous-me.

lann: Oh to be clear I'm fine with it _in the SDK_. I think it should be explicit at the Wasm interface level.

crates/key-value/src/lib.rs
dicej: I'd say no, since it's really SQLite's responsibility to ensure that in-memory databases function the same as disk-backed ones (minus durability, of course).  We don't need to test that for them -- they've already done it.

crates/key-value/src/host_component.rs
dicej: Happy to change it if desired.  Do you have an alternative in mind?

examples/rust-key-value/Cargo.toml
dicej: I had just cut-and-pasted that from another example.  I've removed the section entirely since it doesn't seem to do anything.

crates/key-value/src/table.rs
```diff
@@ -0,0 +1,39 @@
+use std::collections::HashMap;
+
+pub struct Table<V> {
```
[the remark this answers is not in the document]
dicej: Yeah, it's meant to be reusable for any case where you need to assign unique `u32` identifiers to a dynamic set of resources.  I've added docs, plus a TODO indicating we can move it to a utility crate when we're ready to reuse it.

bacongobbler:
Demo works for me!

```
><> spin up --listen 0.0.0.0:3000
Serving http://0.0.0.0:3000
Available Routes:
  hello: http://0.0.0.0:3000/test (wildcard)
```

```
><> curl -XPOST localhost:3000/test/hello --data "world"
><> curl localhost:3000/test/hello
world
><> curl -XPOST localhost:3000/test/hello --data "bacongobbler"
><> curl localhost:3000/test/hello
bacongobbler
```

dicej:
Sorry about all the force pushes -- I rebased onto main, and for whatever reason rebase doesn't re-sign commits by default, and then CI wasn't happy, etc., etc.

dicej:
TODO: I've updated [the SIP](https://github.com/fermyon/spin/pull/1045) to indicate that the default database should _not_ be shared among apps, which is not yet reflected in this implementation.  We need to decide where on the filesystem each app's database should be stored, at which point I'll update this PR.

crates/key-value/src/host_component.rs
kate-goldenring: the [`Config` enum](https://github.com/fermyon/spin/blob/2cab1dab9ea0e0fbfc622e9e186039c1d2feaf14/crates/key-value/src/sqlite.rs#L12) could be `DatabaseLocation`

dicej: Good idea -- thanks!

crates/key-value/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+    Sqlite(sqlite::DatabaseLocation),
+}
+
+#[derive(Clone)]
+pub struct Config {
+    pub configs: HashMap<String, ImplConfig>,
+}
+
+#[async_trait]
+trait Impl: Sync + Send {
```
radu-matei, line 35: The trait names here, `Impl` and `ImplStore` are fairly unintuitive. Is this a naming pattern inherited from other libraries?

crates/build/src/lib.rs
```diff
@@ -50,7 +50,7 @@ fn build_component(raw: RawComponentManifest, app_dir: &Path) -> Result<()> {
             );
             let workdir = construct_workdir(app_dir, b.workdir.as_ref())?;
             if b.workdir.is_some() {
-                println!("Working directory: {:?}", workdir);
+                println!("Working directory: {workdir:?}");
```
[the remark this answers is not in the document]
radu-matei: Ah, that makes sense. Thank you!

dicej:
I've updated the code to locate the default KV database as specified in the SIP.  I think this will be merge-able once the SIP is merged.

radu-matei (approved):
🎉

PR #1036, opened 2023-01-19
Trigger plugins

itowlson:
This possibly needs a SIP (and is definitely not ready for use, not even really for review) but I am creating the PR now to make it easier for me to track what has changed in the course of all the two-steps-forward-one-step-back-ing.  I'm not happy with the loader kludge but have struggled to find a better way so definitely open to suggestions on that.

I haven't yet figured out proper testing infrastructure but am currently testing with https://github.com/itowlson/spin-external-trigger-very-very-example


itowlson:
Current experience: given a trigger type of `thingy`, it wants to run `spin trigger-thingy`, so it checks the plugin store:

* If `trigger-thingy` is installed, it runs it
* If `trigger-thingy` is in the catalogue and compatible, but not installed, it prints `Error: No built-in trigger named 'thingy', but plugin 'trigger-thingy' is available to install`
* If `trigger-thingy` is in the catalogue but **not** compatible, but not installed, it prints `Error: No built-in trigger named 'thingy', and plugin 'trigger-thingy' is not compatible`
* If `trigger-thingy` is **not** in the catalogue at all, it prints `Error: No built-in trigger named 'thingy', and no plugin named 'trigger-thingy' was found`

I considered going further in the 'available' case and offering to install it, but am in two minds:

**Pro:** Gosh it would be convenient
**Con:** It introduces a user interaction into `spin up` which would cause it to block in programmatic environments, and while we could add silent-mode switches, it would further clutter an already complicated command line

Thoughts welcome!

itowlson:
Well, I still don't have a good testing story, but I've added some tests to `spin_trigger::locked` to verify that values are brought into the lockfile correctly,  Everything after that is existing trigger infrastructure and is pretty proven!  So may as well get reviews under way...

(Note:: The CI failure is in code coverage, on a test that runs fine earlier in the CI action without code coverage.  Please ignore it for now.  Unless you have any idea what's going on of course in which case LET ME BRIBE YOU.)

crates/manifest/src/lib.rs
```diff
@@ -306,6 +396,8 @@ pub enum TriggerConfig {
     Http(HttpConfig),
     /// Redis trigger configuration
     Redis(RedisConfig),
+    /// External trigger configuration
+    External(HashMap<String, String>),
```
lann, line 400: Maybe use `toml::Value` for the value here to allow non-string values? I _think_ it should work fine with `ValuesMapBuilder::serializable` (instead of `::string`) below.

crates/trigger/tests/triggers/dummy.wasm
```diff
@@ -0,0 +1 @@
+Ha ha!  I'm not really a Wasm file at all
```
lann, line 1: :scream:

src/commands/up.rs
```diff
@@ -257,3 +256,33 @@ fn parse_env_var(s: &str) -> Result<(String, String)> {
     }
     Ok((parts[0].to_owned(), parts[1].to_owned()))
 }
+
+fn try_trigger_plugin(trigger_type: &str) -> Result<String> {
```
lann, line 260: Given the Rust `try_*` convention, I read this as "try to trigger the plugin". Maybe "resolve"?

```suggestion
fn resolve_trigger_plugin(trigger_type: &str) -> Result<String> {
```

src/commands/up.rs
```diff
@@ -257,3 +256,33 @@ fn parse_env_var(s: &str) -> Result<(String, String)> {
...
+    use crate::commands::plugins::PluginCompatibility;
+    use spin_plugins::manager::PluginManager;
+
+    let subcommand = format!("trigger-{trigger_type}");
+    let plugin_manager = PluginManager::try_default()?;
```
lann, line 265: `.context`?

src/commands/up.rs
```diff
@@ -257,3 +256,33 @@ fn parse_env_var(s: &str) -> Result<(String, String)> {
...
+    let plugin_store = plugin_manager.store();
+    let is_installed = plugin_store
+        .installed_manifests()?
```
lann, line 268: This doesn't seem to actually be fallible...but in case it ever becomes fallible: `.context`?

src/commands/up.rs
```diff
@@ -257,3 +256,33 @@ fn parse_env_var(s: &str) -> Result<(String, String)> {
...
+        .iter()
+        .any(|m| m.name() == subcommand);
+
+    if is_installed {
+        return Ok(subcommand);
+    }
+
+    if let Some(known) = plugin_store
+        .catalogue_manifests()?
```
lann, line 277: `.context`?

kate-goldenring:
Worked like a charm for me! Buzzing and fizzing with excitement!
```
kagold@Kates-MacBook-Pro guest % spin up --follow-all
Buzz
Fizz
Fizz
Buzz
Fizz
Fizz
Buzz
Fizz
Buzz
```

crates/loader/src/local/mod.rs
```diff
@@ -361,6 +368,70 @@ impl ComponentDigest {
...
+    })
+}
+
+fn resolve_trigger(
+    app_trigger: &ApplicationTrigger,
+    partial: toml::Value,
+) -> Result<TriggerConfig> {
+    use serde::Deserialize;
+    let tc = match app_trigger {
+        ApplicationTrigger::Http(_) => TriggerConfig::Http(HttpConfig::deserialize(partial)?),
```
kate-goldenring, line 428: While this was not added by this PR, I wonder if at some point, we can rename `TriggerConfig` to `ComponentTriggerConfig`. `ApplicationTrigger` seems aptly named as it defines the trigger type for an application. However `TriggerConfig` does not have that level of specificity. Also, the enum options for each can be confusing: 
```
pub enum ApplicationTrigger {
    /// HTTP trigger type.
    Http(HttpTriggerConfiguration),
    /// Redis trigger type.
    Redis(RedisTriggerConfiguration),
    /// A trigger type that is not built in.
    External(ExternalTriggerConfiguration),
}

pub enum TriggerConfig {
    /// HTTP trigger configuration
    Http(HttpConfig),
    /// Redis trigger configuration
    Redis(RedisConfig),
    /// External trigger configuration
    External(HashMap<String, String>),
}
```

crates/trigger/tests/triggers/dummy.wasm
kate-goldenring: Instead of checking in this file, should we just create a tmpfile in line during each test case?

itowlson: We could... it feels like extra complexity that a reader of the test would need to decode though...  I can make the message in the file more explanatory if that's the concern?  (The issue is that validation fails if the file doesn't exist, so I can make it say that.)

kate-goldenring: That sounds good to me. I get what you're saying. It makes sense to have a checked in test app with real files or to completely write the files to a tempdir as we do [in `test_app()`](https://github.com/fermyon/spin/blob/76ed50dc2b6f9853fe82db1d5827dbc49abc6679/crates/trigger/src/locked.rs#L236). A hybrid approach may be messy.

src/commands/up.rs
itowlson: I'm proposing to change both this and the catalogue one so that, rather than printing error traces, they fall through to the "no plugin found" message.

crates/manifest/src/lib.rs
itowlson: Ooh nice!  Should I do this for the application-level trigger info as well?  That's currently:

```rust
pub struct ExternalTriggerConfiguration {
    trigger_type: String,
    parameters: HashMap<String, String>,
}
```

itowlson: Tried changing it and it seemed to work so YOLO

src/commands/up.rs
```diff
@@ -257,3 +256,36 @@ fn parse_env_var(s: &str) -> Result<(String, String)> {
...
+    use crate::commands::plugins::PluginCompatibility;
+    use spin_plugins::manager::PluginManager;
+
+    let subcommand = format!("trigger-{trigger_type}");
+    let plugin_manager = PluginManager::try_default()
+        .with_context(|| format!("Failed to access plugins looking for '{subcommand}'"))?;
+    let plugin_store = plugin_manager.store();
+    let is_installed = plugin_store
+        .installed_manifests()
+        .unwrap_or_else(|_| vec![])
```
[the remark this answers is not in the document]
itowlson: Doh!  Thanks

lann (approved):

itowlson:
+ @fibonacci1729 (in connection with the sample in particular)

lann:
Doesn't need to be in this PR, but before we publicize trigger plugins we should clean up the public `spin-trigger` interface/docs and add it to https://fermyon.github.io/rust-docs/spin/main/

examples/spin-timer/spin-timer.wit
```diff
@@ -1 +1 @@
-handle-timer-request: func(msg: string) -> string
+handle-timer-request: func()
```
fibonacci1729, line 1: nit: should `handle-timer-request` take a `u64` or `string` encoding of the time that triggered the invocation? Understandably for an example this isn't too important so feel free to ignore.

fibonacci1729 (approved):
LGTM

kate-goldenring (approved):
Just a nit on cleaning up the trigger plugin json.
LGTM!

examples/spin-timer/spin-timer.wit
itowlson: Probably it should - I was sort of trying to avoid wading into the TIME WHAT EVEN IS IT complexities of what the "current time" is and how to represent it across the WIT boundary.  If you don't mind, I'd prefer to punt on that for this PR, though I agree it would be good to show how to pass inputs and process results from the guest.

examples/spin-timer/trigger-timer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "trigger-timer",
+    "description": "IT IS I THE GREAT OZ",
+    "homepage": "www.example.com",
+    "version": "0.1.0",
+    "spinCompatibility": ">=0.7",
+    "license": "Apache-2.0",
+    "packages": [
+        {
+            "os": "linux",
```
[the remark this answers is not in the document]
itowlson: Would it be possible to defer that to a later PR?  I'm concerned about how we manage and test cross-compilations.  Although completely recognise that a lot of folks are on Macs so the Linux setting is not useful to them.

What I would really like to do is create a `spin-pluginify` plugin that takes a (Rust) project, cross-compiles it, tars all the various outputs, generates hashes, and updates the manifest instead of muggins here having to do it by hand...!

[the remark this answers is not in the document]
kate-goldenring: sounds good! Also love that idea about `spin-pluginify`

PR #1045, opened 2023-01-24
add SIP 010 - Key-Value Storage

dicej:


docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,60 @@
...
+
+Owner(s): joel.dice@fermyon.com
+
+Created: Jan 23, 2023
+
+## Background
+
+Spin should have native support for using a variety of key-value (KV) datastores.
+
+Currently, users can use Redis for persistent key-value storage for Spin applications. This [documentation](https://developer.fermyon.com/cloud/data-redis.md) walks through how to use the existing Spin Redis SDK which leverages the `[outbound-redis.wit` interface](https://github.com/fermyon/spin/blob/main/wit/ephemeral/outbound-redis.wit). In contrast to the Redis interface, we are proposing a more general-purpose interface which can be used with a variety of implementations.
```
kate-goldenring, line 16:
```suggestion
Currently, users can use Redis for persistent key-value storage for Spin applications. This [documentation](https://developer.fermyon.com/cloud/data-redis.md) walks through how to use the existing Spin Redis SDK which leverages the [`outbound-redis.wit` interface](https://github.com/fermyon/spin/blob/main/wit/ephemeral/outbound-redis.wit). In contrast to the Redis interface, we are proposing a more general-purpose interface which can be used with a variety of implementations.
```

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,60 @@
+title = "SIP 010 - Key-Value Storage"
+template = "main"
+date = "2023-01-23T01:01:01Z"
+---
+
```
kate-goldenring, line 5: Should we add a "Future Support" section that discusses what other KV interfaces we could support (and not support in the future). For example, we may support `get-many` and `set-many` functions but not the atomic `increment` function?

kate-goldenring (approved):
LGTM! Just added a nit and optional suggestion

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,60 @@
...
+exists: func(store: store, key: string) -> expected<bool, error>
+
+close: func(store: store)
+```
+
+*Note: the pseudo-resource design was inspired by the interface of similar functions in [WASI preview 2](https://github.com/bytecodealliance/preview2-prototyping/blob/d56b8977a2b700432d1f7f84656d542f1d8854b0/wit/wasi.wit#L772-L794).*
+
+#### Built-in local key-value database
+
+Spin will have a built-in database based on SQLite for testing, development, and some production use cases.  We’ve selected SQLite for this purpose because it’s easily embeddable, lightweight, and reliable.  It increases the Spin binary size by about 3% and the release build time by about 1.5%.  We’ve also considered lighter-weight options such as an in-memory hash map which is synced to a flat file on updates, but making this reliably atomic and durable would require reinventing the features SQLite already provides.
```
itowlson, line 60: There was talk at one point of a `spin kv` command so users could inspect their local store.  It would be good to say more about this.

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,60 @@
...
+
+Currently, users can use Redis for persistent key-value storage for Spin applications. This [documentation](https://developer.fermyon.com/cloud/data-redis.md) walks through how to use the existing Spin Redis SDK which leverages the `[outbound-redis.wit` interface](https://github.com/fermyon/spin/blob/main/wit/ephemeral/outbound-redis.wit). In contrast to the Redis interface, we are proposing a more general-purpose interface which can be used with a variety of implementations.
+
+## Proposal
+
+In order to support key-value stores, the following need to be added to Spin:
+
+- A `WIT` file that defines the key-value interface
+- SDK implementations for various programming languages
+- A default local key-value store
```
itowlson, line 24: We should mention that, as a new import in the SDK, this has interactions with cloud deployment:

1. An application that uses KV will currently fail with unresolved imports in the cloud
2. An application that doesn't use KV but where the compiler is not guaranteed to elide the dead imports will also fail with unresolved imports

and say what we propose to do about this (documentation, stubs in the cloud, etc.).

(I know this is a SIP and the details of cloud may be considered out of scope, but I reckon we should say _something._)

docs/content/sips/010-key-value.md
itowlson, line 60: Is there any more design info on this (e.g. are WIT stores modelled by SQLite tables, does this impose any limits on data [specifically are cloud KV stores likely to have similar limits/lack of limits], no transactions, etc.) or do we expect all that to be worked out in code?

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,60 @@
...
+
+get: func(store: store, key: string) -> expected<list<u8>, error>
+
+set: func(store: store, key: string, value: list<u8>) -> expected<unit, error>
+
+delete: func(store: store, key: string) -> expected<unit, error>
+
+exists: func(store: store, key: string) -> expected<bool, error>
+
+close: func(store: store)
```
itowlson, line 53: What happens if the guest module forgets to call `close()`?

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,60 @@
...
+### Key-Value Interface (`.wit`)
+
+Spin should leverage the WebAssembly WASI subgroup’s work to define a universal key-value `WIT` interface. That work is taking place in [the `wasi-keyvalue` proposal](https://github.com/WebAssembly/wasi-keyvalue). However, the proposal was made assuming that `WIT` star imports are available, which is not yet the case. Therefore, we're modeling stores as pseudo-[resource handles](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-resource) which may be created using an `open` function and disposed using a `close` function.  Each operation on the store is a function which accepts a handle as its first parameter.
+
+```
+type store = u32
+
+variant error {
+  store-table-full,
+  no-such-store,
```
itowlson, line 37: Does this imply stores have to be created?  If so how does that happen?

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,60 @@
...
+  invalid-store,
```
itowlson, line 38: Does this have a specific meaning?  I'm afraid my first reaction was "invalid _in what way?_"

docs/content/sips/010-key-value.md
dicej: This is at least partially addressed by https://github.com/fermyon/spin/pull/1048 (see the last paragraph in that document, in particular).  Perhaps it would be useful to reference that SIP from this one?

docs/content/sips/010-key-value.md
dicej: Right now the implementation uses this as the schema: https://github.com/fermyon/spin/pull/1048.  The store name, key name, and value can each be up to 1GB.  [See here for other limits](https://www.sqlite.org/limits.html).  Transactions and atomic operations should be no problem.

docs/content/sips/010-key-value.md
itowlson: At the moment, I'm afraid I don't feel I know enough from that paragraph about how we propose to handle this in the KV case, which is why I brought it up here.

The comment at the moment kind of feels like a side suggestion to an explanatory note about the rationale for a particular detail in a proposal that was primarily about improving error messages.  If it's a definite proposal, we could break it out into a top-level section about managing how changes propagate from Spin to Cloud, and make it more concrete (e.g. the timing of WIT iteration -> Cloud stubs -> Spin release).  That would certainly address my uncertainty.

docs/content/sips/010-key-value.md
dicej: That would be equivalent to leaving a file handle open in a native process.  It will be cleaned up automatically when the trigger function returns.  If the the app opens a bunch of stores in a loop without closing them, it will probably hit some resource limit (e.g. actual file handles to the database in the current SQLite-based implementation).  We could also consider an artificially-imposed limit to the number of simultaneously open handles allowed.

FWIW, the Rust SDK wraps this in a type that implements `Drop`, so RAII takes care of things for us there.

docs/content/sips/010-key-value.md
itowlson: Think that might be the wrong link, sorry.

But... honestly... I don't want you to tell me the design details in a comment... I'd really like to see them captured in the SIP so they can be shared and referred back to... or do you feel that's too much detail?

docs/content/sips/010-key-value.md
dicej: The current thinking is that only the default store (named "") is available default.  Other stores would need to be specified via runtime configuration.  If an app tries to open a store that is not the default and has not been configured, it will get `no-such-store`.

dicej: I'll add doc comments to the WIT file to describe each function and the errors it can raise.

docs/content/sips/010-key-value.md
dicej: This is an artifact of using pseudo-resource handles (which are forgeable).  So a guest could call `get(42, "foo")` without ever calling `open` (i.e. just making up a store handle), in which case the host would raise `invalid-store`.  Once we have real resource handles, that error won't be necessary anymore.  I'll add docs to clarify this.

itowlson: Aha, that makes perfect sense - thanks!

docs/content/sips/010-key-value.md
dicej: Was anything written down about `spin kv`?  I don't recall hearing about it, but perhaps I did and forgot.  Was it intended just to dump all the tuples or do more precise queries?  Would it support modifying the store at all?

itowlson: No, no `spin kv` command was defined as far as I know - some people expressed verbally that there should be a command for users to at least inspect their local store and see what was in it, but nothing was decided or specified - that's why I was flagging it for the SIP.

If there isn't a command for that, then maybe the SIP should at least highlight that, at least in the first iteration, the store will be opaque - there's no way to query or modify it except by writing a Wasm module - for me that seems like something we should consciously agree on before we ship.

dicej: Part of the reason I chose SQLite was because you can inspect and modify the database on disk with standard tools (e.g. `sqlite3` or your favorite SQLite GUI).

dicej: But yes, I agree it would be nice to have something built in to Spin.

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,112 @@
...
+- SDK implementations for various programming languages
+- A default local key-value store
+
+Although not in scope for this proposal, we'll also want to expand the runtime configuration code added in [this PR](https://github.com/fermyon/spin/pull/798) to support configuring various key-value stores, including the default one.
+
+### Key-Value Interface (`.wit`)
+
+Spin should leverage the WebAssembly WASI subgroup’s work to define a universal key-value `WIT` interface. That work is taking place in [the `wasi-keyvalue` proposal](https://github.com/WebAssembly/wasi-keyvalue). However, the proposal was made assuming that `WIT` star imports are available, which is not yet the case. Therefore, we're modeling stores as pseudo-[resource handles](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#item-resource) which may be created using an `open` function and disposed using a `close` function.  Each operation on the store is a function which accepts a handle as its first parameter.
+
+```
```
radu-matei, line 32: nit: could you please add some syntax highlighting to the code block?
It makes it so much easier to read it; I've been using `fsharp` for WIT so far, and it seems to be largely ok.

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Future work
+
+In addition to the built-in, SQLite-based implementation described above, we expect to add a number of other implementations backed by e.g. Redis, other relational databases, eventually consistent distributed stores, etc.  Each of these implementations will have its own performance, consistency, and durability characteristics, and some applications may use a combination of them to handle different types of data.  As mentioned above, this will require defining and configuring named stores in a runtime configuration file.
+
+We also expect to add additional interfaces for e.g. atomic, bulk, and asynchronous operations, key expiration, etc., following [`wasi-keyvalue`](https://github.com/WebAssembly/wasi-keyvalue) as much as possible.
```
radu-matei, line 112: Would we have a new SIP that defines adding the new interfaces?

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,114 @@
...
+// `error::no-such-store` will be raised if the `name` is not recognized.
+open: func(name: string) -> expected<store, error>
+
+// Get the value associated with the specified `key` from the specified
+// `store`.
+//
+// `error::invalid-store` will be raised if `store` is not a valid handle
+// to an open store, and `error::no-such-key` will be raised if there is no
+// tuple for `key` in `store`.
+get: func(store: store, key: string) -> expected<list<u8>, error>
```
Mossaka, line 73: Minor thing: could you please clarify what version of `.wit` is this document based on. I got confused for a second for the `exepcted` syntax as the current wit switches to `result` type but remembered the old wit syntax was using `expected`. 

Another comment is about the payload type from the `get` function. In wasi-keyvalue proposal, we realized that people might want differetn ways to consume the payload of a keyvalue store. This includes consuming it asynchronously and synchronously. Other that these, people might just want to examine the size of the payload, or some metadata of the payload. This observation leads to us proposing a payload `resource` type that includes some host functions for interpreting payload in different ways.

```fsharp
resource payload {
    consume_async: func() -> result<stream<u8>, Error>
    consume_sync: func() -> result<list<u8>, Error>
    size: func() -> result<u64, Error>
    // possibly more methods, like consume_async_with_key that returns a key payload pair `(key, stream<u8>)`
  }
```
Since `resource` is not there yet and this SIP is proposing using a `u32`-based resource, I would imagine it would difficult / non-ergonomic for users to call a different function to interpret the payload. But this is something I'd like your feedback on.

Mossaka (commented):
lgtm! Just had one comment about the payload of the keyvalue `get` function. Looking forward for future collaboration on this!

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,114 @@
...
+close: func(store: store)
+```
+
+*Note: the pseudo-resource design was inspired by the interface of similar functions in [WASI preview 2](https://github.com/bytecodealliance/preview2-prototyping/blob/d56b8977a2b700432d1f7f84656d542f1d8854b0/wit/wasi.wit#L772-L794).*
+
+#### Built-in local key-value database
+
+Spin will have a built-in database based on SQLite for testing, development, and some production use cases.  We’ve selected SQLite for this purpose because it’s easily embeddable, lightweight, and reliable.  It increases the Spin binary size by about 3% and the release build time by about 1.5%.  We’ve also considered lighter-weight options such as an in-memory hash map which is synced to a flat file on updates, but making this reliably atomic and durable would require reinventing the features SQLite already provides.
+
+The default database will be unique to each Spin application; i.e. setting a key to a value in one app will not conflict with or be visible to other apps.  Explicit runtime configuration will be required to allow multiple apps to share the same database.
```
kate-goldenring, line 108: When you say "the default db will be unique to each app" do you mean each app will have its own db? So for the local experience, a unique `db.db` file? Instead, are we saying each app's store's are unique to them, regardless of whether they exist in the same db?

docs/content/sips/010-key-value.md
dicej: Perhaps, although if we're just adding features from the `wasi-keyvalue` proposal there might not be much to discuss, in which case a GitHub issue followed by a PR might suffice in lieu of a full-blown SIP.

docs/content/sips/010-key-value.md
dicej: Thanks for the feedback, @Mossaka.  I've added a note about the out-of-date WIT syntax (which we're unfortunately stuck on until we can update to the latest `wit-bindgen`) -- good catch.

I like the `payload` API you've proposed.  Perhaps we could have both:

```fsharp
// pseudo-handle for payloads:
type payload = u32
get: func(store: store, key: string) -> expected<list<u8>, error>
get-payload: func(store: store, key: string) -> expected<payload, error>
close-payload: func(payload: payload)
// pseudo-handle for streams:
type stream = u32
consume-async: func(payload: payload) -> expected<stream, error>
consume-sync: func(payload: payload) -> expected<list<u8>, error>
size: func(payload: payload) -> expected<u64, error>
// etc.
close-stream: func(stream: stream)
```
Then `get` would just be shorthand for calling `get-payload`, `consume-sync`, and finally `close-payload`.

docs/content/sips/010-key-value.md
dicej: Whether each app has its own db or each app has its own table in a shared db (or each app has its own `namespace` column in a shared table in a shared db) is an implementation detail, I think.  So yes, we are just saying each app's store is unique to that app.  I'll update the wording to clarify.

michelleN (approved):

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,116 @@
...
+
+By default, each app will have its own, persistent, default store which is independent of all other apps.  This could be implemented as a separate SQLite database for each app, a separate table for each app in a shared database, or even row-level separation via an `app` column in a shared table.  For the initial implementation, we'll use a separate database for each app.  For local apps, the database will be stored by default in a hidden `.spin` directory adjacent to the app's `spin.toml`.  For remote apps (e.g. served by `bindle`), an in-memory database will be used by default.  Both of these defaults may be overridden via a `--key-value-database` option for specifying which file to use.
```
lann, line 110: We should print a message whenever the default database is created in either of these cases.

docs/content/sips/010-key-value.md
dicej: I've added a paragraph about this to the `Future work` section.

dicej: Honestly, I was reluctant to mention SQLite at all in this document since it could easily be replaced with a number of other embedded databases (e.g. RocksDB, BerkelyDB, etc.) without significantly affecting the DX.  It was mainly to address the "Why not just use a `HashMap` and serialize it to a JSON file as needed?" question that came up early on.  So yes, I'd prefer to discuss these details in https://github.com/fermyon/spin/pull/1035 rather than in the SIP.

docs/content/sips/010-key-value.md
dicej: I've added print statements to the implementation.

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,118 @@
...
+  // The store handle provided is not recognized, i.e. it was either never
+  // opened or has been closed.
+  invalid-store,
+
+  // No key-value tuple exists for the specified key in the specified
+  // store.
+  no-such-key,
+
+  // Some implementation-specific error has occurred (e.g. I/O)
+  runtime(string)
```
lann, line 58: This label is a little weird since `no-such-key` (and arguably `store-table-full`) are also "runtime" errors. `spin-config.wit` uses `provider(string)`. I could also see `other(string)`.

Not a blocker. Naming is hard.

docs/content/sips/010-key-value.md
lann: Given that this specific proposal is intentionally constrained to the wit-bindgen v0.2 feature set, is the payload pseudo-resource going to be worthwhile? We will never be able to implement `consume-async` in this version of the interface, for example.

docs/content/sips/010-key-value.md
lann: It might be worth thinking about _minimum_ sizes that implementations of the interface should provide, e.g. "at least 256 utf-8 bytes for keys".

docs/content/sips/010-key-value.md
dicej: `implementation` or `io`, perhaps?

dicej:
@lann I think the idea with `consume-async` would be that it returns a pseudo-stream handle which you could read from in chunks and thereby avoid buffering the whole thing in memory.  Although `consume-chunked` would be a better name for that.  Or we could go a step further and mimic WASI Preview 2's pseudo-streams and pseudo-futures (now called `waitable`, I believe) for a more genuine async experience, all while still using the plain old modules and the antiquated WIT syntax.  I do think that's all out of scope for the initial implementation, though.

lann:
> I do think that's all out of scope for the initial implementation, though.

No concerns from me then.

lann (approved):

radu-matei:
I realize this is late in the review process for the SIP, so apologies.

Do we think that we could also add a `list` or `get-all` (as it is defined in [the upstream WASI KV proposal](https://github.com/WebAssembly/wasi-keyvalue)) operation to the initial API?

I could imagine this being extremely useful for various (even early) scenarios, and I would like to use it in Spin components soon.

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,118 @@
...
+
+By default, each app will have its own, persistent, default store which is independent of all other apps.  This could be implemented as a separate SQLite database for each app, a separate table for each app in a shared database, or even row-level separation via an `app` column in a shared table.  For the initial implementation, we'll use a separate database for each app.  For local apps, the database will be stored by default in a hidden `.spin` directory adjacent to the app's `spin.toml`.  For remote apps (e.g. served by `bindle`), an in-memory database will be used by default.  Both of these defaults may be overridden via a `--key-value-database` option for specifying which file to use.
```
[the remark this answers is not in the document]
dicej: Perhaps it would help to rename it to `--default-key-value-file` and document that it only affects the default (i.e. empty name) store, overriding whatever the runtime config says, using the SQLite backend.  I.e. even if the runtime config says use such-and-such Redis server for the default store, this option would force it to use a local SQLite file instead.

dicej:
> Do we think that we could also add a `list` or `get-all` (as it is defined in [the upstream WASI KV proposal](https://github.com/WebAssembly/wasi-keyvalue)) operation to the initial API?

I'm not seeing `list` or `get-all` in the upstream proposal.  I'm guessing `list` would get all the keys in a store, while `get-all` would get all the key-value pairs in a store?  If so, they seem reasonable to add.

lann:
`list-keys` is probably the most generally applicable feature. Getting all _values_ might be getting into "probably want some kind of streaming/pagination" territory.

radu-matei:
Sorry, `get-keys` is the function:

```fsharp
interface "wasi:kv/data/get-many" {
  // get_many returns a stream of key-value pairs.
  // Notice hat this interface is non-atomic, meaning that the key-value pairs
  // returned may not be consistent.

  use { Error, payload } from "wasi:kv/types"
  
  get-many: func(keys: keys) -> result<stream<payload>, Error>

  get-keys: func() -> result<keys, Error>
}

```

dicej:
Recent changes:

- change `error::runtime` to `error::io`
- add `get-keys` function to interface
- add `Implementation requirements` section
- remove `--key-value-database` CLI option in favor of a runtime config CLI

docs/content/sips/010-key-value.md
```diff
@@ -0,0 +1,132 @@
...
+
+*Note: the pseudo-resource design was inspired by the interface of similar functions in [WASI preview 2](https://github.com/bytecodealliance/preview2-prototyping/blob/d56b8977a2b700432d1f7f84656d542f1d8854b0/wit/wasi.wit#L772-L794).*
+
+#### Implementation requirements
+
+In addition to the above interface, we specify a few additional implementation requirements which guest components may rely on.  At minimum, an conforming implementation must support:
+
+- Keys as large as 256 bytes (UTF-8 encoded)
+- Values as large as 1 megabyte
+- Capacity for 1024 key-value tuples
```
lann, line 118: This might be slightly confusing when combined with the value size requirement, suggesting a cloud provider must provide 1GB storage. Maybe qualify the whole section "subject to overall database size limits" or something?

lann (approved):

docs/content/sips/010-key-value.md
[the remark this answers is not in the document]
dicej: For the record, we ended up removing the option and will revisit when we have runtime configuration.

docs/content/sips/010-key-value.md
dicej: For the record, I've changed the name to `io`.

kate-goldenring (approved):

Issue #820, opened 2022-10-14
Http-go template fails


mikkelhegn:
Creating a new Spin application using the http-go template fails without any modifications to the code:

```
Serving http://127.0.0.1:3000
Available Routes:
  go-temp2: http://127.0.0.1:3000 (wildcard)
2022-10-14T17:31:26.421024Z ERROR spin_http_engine: Error processing request: wasm trap: wasm `unreachable` instruction executed
wasm backtrace:
    0: 0x3a8f - <unknown>!runtime.runtimePanic
```

radu-matei:
Could you please share which versions for Spin and TinyGo this is?

Just validated this with Spin from canary, TinyGo 0.26, and the latest template version and it fails for me as well.

For reference, I also tested our C# template, which is built on top of the same C interop, and that one works well.


cc @adamreese

etehtsea:
As it passes on CI, it looks like there are some breaking changes between 0.22 and 0.26 - https://github.com/fermyon/spin/actions/runs/3245485308/jobs/5323079898

mikkelhegn:
I've been testing the 0.5.0 release and the canary release (from October 14th) and both fail, with both SDK, Spin runtime, and template being from the same version.

TinyGo: `tinygo version 0.26.0 darwin/amd64 (using go version go1.19.2 and LLVM version 14.0.0)`

etehtsea:
I did a little research about it - it works fine up to 0.25.0 (incl).
Also, I've figured out the commit that is causing the issue - https://github.com/tinygo-org/tinygo/commit/03d1c44265e1f628b2730b9c214db2390a745ee6
Reverting it from the tinygo's dev branch make tests (`make test-sdk-go`) to pass again (so as `http-tinygo` example).

radu-matei:
Validated that TinyGo v0.25 works.
Thanks for investigating this!

lann:
Removing `-no-debug` from the `tinygo` build command and running spin with `WASMTIME_BACKTRACE_DETAILS=1` gives more detail:

```
    2: 0x141d9 - realloc
                    at /usr/lib/tinygo/src/runtime/arch_tinygowasm.go:136:28
```

lann:
Finally getting back to this... The backtrace appears to point at the indexing operation here:
```go
ptr := unsafe.Pointer(&buf[0])
```

Suggesting that the panic could be caused by `buf` having size zero, which could possibly happen if `realloc` was called with size 0.

lann:
[...]

PR #1047, opened 2023-01-24
go: Fix SDK HTTP response malloc

lann:
Some subtle errors in memory management lead to invalid pointers which could fail at runtime in wit-bindgen's recursive response parameter memory cleanup (and possibly corrupted response data).


lann:
In combination with https://github.com/tinygo-org/tinygo/pull/3410 I believe this fixes https://github.com/fermyon/spin/issues/820

lann:
Confirmed `examples/http-tinygo` still works with Tinygo 0.25.

adamreese (approved):
Manually tested on TinyGo v1.25

PR #1048, opened 2023-01-24
add SIP 011 - Component versioning

dicej:


docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,41 @@
...
+
+Note that when we say "SDK" in this document, we mean either one of the official Spin SDKs _or_ raw bindings generated from the `WIT` files and the `wit-bindgen` version shipped with a given version of Spin, e.g. for languages which do not yet have an "official" SDK.
+
+## Proposal
+
+Although `WIT` does not yet have a way to express API versions, we can emulate it by introducing a new interface called `spin-version.wit` containing a single function to be exported by the guest component:
```
lann, line 20: Maybe `spin-sdk-version`? `spin-version` sounds like the reverse; giving the host version to the guest.

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,41 @@
...
+```
+record version-details {
+  commmit-hash: string,
+  sdk-language: string
+}
+
+spin-version-$MAJOR-$MINOR: func() -> version-details
+```
+
+where `$MAJOR` and `$MINOR` are the Spin major and minor version numbers the SDK targets. The intention behind embedding these numbers in the name of the function is that it can be checked statically without instantiating or running the component, and it is immune to ABI changes.  The `version-details` record provides additional details about the SDK used to build the component.  Getting that information requires actually running the component, which may not be possible if the host does not support the ABI for that version, so it won't be useful in all scenarios.  (TODO: should we just encode everything in the name?)
```
lann, line 31: > should we just encode everything in the name?

Yes, actually. Not needing to execute to inspect would be nice.

Maybe several exports, like:

```
spin-sdk-version-$MAJOR-$MINOR: func()
spin-sdk-language-$LANGUAGE: func()
spin-sdk-commit-$HASH: func()
```

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Background
+
+Currently, Spin has no way to determine which version of SDK the components of an application were built with, which leads to obscure, hard-to-understand error messages for ABI and/or API incompatibilities.  For example, a component built with a Spin 0.4 SDK won't run on Spin 0.7 due to `wit-bindgen` ABI changes, leading to a variety of confusing runtime errors like out-of-bounds memory traps, assertion errors about invalid values, etc.  In addition, new APIs were added between 0.4 and 0.7, and thus a 0.7 component might not run on 0.4 even if the ABI had not changed.  These errors are usually a bit easier to understand (e.g. missing imports), but still don't point clearly to a resolution.  Ideally, the message presented to the user would be something like, "This component was built with a Spin 0.7 SDK and cannot be run using this version of Spin (0.4) -- please use Spin 0.7 or later."
+
+Note that when we say "SDK" in this document, we mean either one of the official Spin SDKs _or_ raw bindings generated from the `WIT` files and the `wit-bindgen` version shipped with a given version of Spin, e.g. for languages which do not yet have an "official" SDK.
```
itowlson, line 16: In that case would a safer phrasing be e.g. "This component targets Spin 0.7... (etc)"/

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,41 @@
...
+
+When Spin deploys or runs an application, it will have an opportunity to check the versions of its component(s) for compatibility, resulting in one of the following scenarios:
+
+- *No version found*: accept and run the component, raising an error if an incompatibility is detected.  This matches the current behavior.
+
+- *Version found, and it's within the range that this build of Spin supports*: accept and run the component.  Presumably no ABI or API incompatibilities are possible in this scenario, unless the component is reporting an inaccurate version or a breaking SDK change was made without bumping the version.
```
itowlson, line 37: "Within the range that this build of Spin supports" sounds like a possibly elusive target.  Consider the case that Spin 0.11 scans the exports and find `spin-sdk-version-0.9`.  How does it know whether it supports 0.9?  Does it have a table?  How is that table maintained?

This gets simpler once we get into non-zero major versions because within a major version we know that if Spin's minor version is higher than the exported version then we're safe.  We would still need to maintain a table of compatible major versions but hopefully that would be less onerous given the release cadence.

Then again, at non-zero major versions I really hope we are introducing some sort of rolling upgrade policy for APIs, meaning this would be needed only when the ABI changed, and when deprecated APIs were finally put out to pasture, all of which would hopefully be infrequent and well trailed.

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,41 @@
...
+
+- *Version found, and it's _not_ within the range that this build of Spin supports*: by default, optimistically accept and run the component, only raising an error if and when an ABI or API incompatibility is detected.  The error message should contain a clear, actionable report based on the component's version.  Alternatively, if "strict" mode is requested (e.g. via a command line option), Spin should reject the application immediately without attempting to deploy or run it, again providing a clear error message.
```
itowlson, line 39: Oh if we are going to execute by default anyway then maybe my concern above goes away.

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,41 @@
...
+```
+record version-details {
+  commmit-hash: string,
+  sdk-language: string
+}
+
+spin-sdk-version-$MAJOR-$MINOR: func() -> version-details
+```
+
+where `$MAJOR` and `$MINOR` are the Spin major and minor version numbers the SDK targets. The intention behind embedding these numbers in the name of the function is that it can be checked statically without instantiating or running the component, and it is immune to ABI changes.  The `version-details` record provides additional details about the SDK used to build the component.  Getting that information requires actually running the component, which may not be possible if the host does not support the ABI for that version, so it won't be useful in all scenarios.  (TODO: should we just encode everything in the name?)
```
itowlson, line 31: Will this WIT file be auto-generated as part of the GitHub action?  What will happen in canary where there may have been a breaking change but we won't have revved the version yet?  (Or will we?)

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,41 @@
...
+
+- *Version found, and it's _not_ within the range that this build of Spin supports*: by default, optimistically accept and run the component, only raising an error if and when an ABI or API incompatibility is detected.  The error message should contain a clear, actionable report based on the component's version.  Alternatively, if "strict" mode is requested (e.g. via a command line option), Spin should reject the application immediately without attempting to deploy or run it, again providing a clear error message.
+
+The rationale for the default, "optimistic" behavior described in the last scenario above is that, although an app may be built against a newer SDK version than Spin supports, it might not use any of the new features of that SDK and thus should run fine on an older Spin (assuming no relevant ABI changes, either).  In particular, Spin may provide trapping stub implementations for unknown imports to maximize compatibility with applications which may import newer functions but never actually call them.  This is particularly common for dynamic languages where dead code elimination is infeasible.
```
itowlson, line 41: Is there more detail to come here?  If not maybe move this to a "future discussion" section so it's clear this isn't part of the current proposal?

itowlson (commented):
This is interesting though I do wonder if we should try to capture our longer-term versioning strategy alongside it.  But as an "ambulance at the bottom of the cliff" for our current situation it at least gives users a clear explanation and guidance - thanks!

docs/content/sips/011-component-versioning.md
lann: Yeah I imagine this being primarily about nice error messages and perhaps compatibility hacks in the near term.

docs/content/sips/011-component-versioning.md
dicej: I've been deliberately vague about how a given Spin version decides which SDK versions are supported, given that the primary goals of this SIP are:

1. Start recording versions in components so we have _something_ to work with.
2. Provide better error messages when imports are missing and/or the ABI is broken.

I'll add a "future discussion" section to discuss various ways future Spin versions could make compatibility decisions vs. the "exact match only" approach I imagine we would start with.

docs/content/sips/011-component-versioning.md
dicej: > Will this WIT file be auto-generated as part of the GitHub action? What will happen in canary where there may have been a breaking change but we won't have revved the version yet? (Or will we?)

I've added a couple of sections addressing these questions: `Targetting unreleased Spin versions` and `SDK implementation`.

docs/content/sips/011-component-versioning.md
dicej: I've added a `Future work` section to address these topics.  Let me know if it's still not clear.

docs/content/sips/011-component-versioning.md
dicej: Good call -- I've updated the phrasing.

docs/content/sips/011-component-versioning.md
dicej: I've moved this to the `Future work` section.

fibonacci1729 (approved):

itowlson (approved):
With apologies for leaving it so long.  Thanks for working through this - there's a lot of potential here to reduce some of our major pain points.

dicej:
> With apologies for leaving it so long. Thanks for working through this - there's a lot of potential here to reduce some of our major pain points.

No worries at all.  This wasn't a high priority versus everything else we've been working on lately, so I deliberately left it on the shelf for a while.

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,56 @@
...
+
+```
+spin-sdk-version-$MAJOR-$MINOR
+spin-sdk-language-$LANGUAGE
+spin-sdk-commit-$HASH
+```
+
+where `$MAJOR` and `$MINOR` are the Spin major and minor version numbers the SDK targets, `$LANGUAGE` is the programming language the SDK supports, and `$HASH` is the Git commit hash the SDK was built from. The intention behind embedding this information in the name of the function is that it can be checked statically without instantiating or running the component, and it is immune to component model ABI changes.
+
+For languages which don't yet have an version-enabled Spin SDK (or developers who wish to use `wit-bindgen` directly, or even implement bindings by hand), the component may itself export any or all of the above functions.
```
vdice, line 28:
```suggestion
For languages which don't yet have a version-enabled Spin SDK (or developers who wish to use `wit-bindgen` directly, or even implement bindings by hand), the component may itself export any or all of the above functions.
```

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Targetting unreleased Spin versions
```
vdice, line 30:
```suggestion
### Targeting unreleased Spin versions
```

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,56 @@
...
+
+The above works fine for Spin releases, but we might also want to build components that target pre-release builds of Spin, which may include an ABI and/or API that's different from what ends up in the next release, and yet also different from what was in the previous release.  To handle this case, we propose adding a step to the Spin release process such that the version stored in the `workspace.package.version` field of Spin's Cargo.toml file is changed to `$MAJOR.$MINOR.0-pre0`, where `$MAJOR.$MINOR.0` is the anticipated next Spin release.  Then, each time an ABI or API change is made, the `-pre$N` suffix is incremented.  For example, when Spin 0.9 is released, the version on the main branch is updated to `0.10.0-pre0`, and if an API change is made a few days later, it's updated again to `0.10.0-pre1`.
```
vdice, line 32: Curious as to why the patch version is hard-coded to `0`.  Does this mean if spin or and SDK is on a non-zero patch release, eg 0.9.1, does the `workspace.package.version` need to remain `0.9.0-pre$N`?  Or should this be reworded to allow for arbitrary patch values, eg `$MAJOR.$MINOR.$N-pre$N`?

docs/content/sips/011-component-versioning.md
```diff
@@ -0,0 +1,56 @@
...
+
+Using the above scheme, components can target a given pre-release by appending the appropriate `-pre$N` suffix to `spin-sdk-version-$MAJOR-MINOR`, e.g. `spin-sdk-version-0-10-pre1`.
```
vdice, line 34: Ah, I see how the patch version is left out here.  So maybe it should be left out as well re: the comment above?

vdice (approved):
Late to the game here; feel free to move forward and we can address any/all as follow-ups.

docs/content/sips/011-component-versioning.md
dicej: Good question.  We've had some discussion with Luke Wagner and others about what role patch version numbers should play in Component Model interface versioning (which this SIP is designed to anticipate), and the current thinking is that it's not relevant.  Versioning interfaces is different than versioning implementations, and while patch versions definitely make sense for the latter, they're not necessarily useful for the former.

Also, the _anticipated_ next Spin release should always have a zero patch version, since non-zero patch versions are for _unanticipated_ issues.

docs/content/sips/011-component-versioning.md
dicej: Cargo.toml's version field _must_ have three components to it (major, minor, and patch), so we have to include the patch version there even if it's not relevant to SDK interface versioning.

PR #1049, opened 2023-01-24
Update toml dependency to 0.6

cardoso:
0.6 brings full TOML 1.0 compliance and several fixes.

Only had to adapt for a small breaking change:
- de::from_slice and ser::to_vec were removed, instead use from_str and to_string and convert with bytes manually

https://github.com/toml-rs/toml/blob/main/crates/toml/CHANGELOG.md#060---2023-01-23

Also a few new features which could help with parsing and editing spin.toml

### DIFF
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -141,7 +141,7 @@ fn inner_check_supported_version(
             eprintln!("Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Check overridden ... continuing to install or execute plugin.");
         } else {
             return Err(anyhow!(
-            "Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Try running `spin plugin update && spin plugin upgrade --all` to install latest or override with `--override-compatibility-check`."
+            "Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Try running `spin plugins update && spin plugins upgrade --all` to install latest or override with `--override-compatibility-check`."
         ));
         }
     }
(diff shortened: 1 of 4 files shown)


src/commands/new.rs
```diff
@@ -178,17 +178,30 @@ impl FromStr for ParameterValue {
...
-        .read_to_end(&mut buf)
-        .await
+/// This function reads a file and parses it as TOML, then
+/// returns the resulting hashmap of key-value pairs.
+async fn values_from_file(path: impl AsRef<Path>) -> Result<HashMap<String, String>> {
+    // Get the absolute path of the file we're reading.
+    let path = absolutize(path)
+        .with_context(|| anyhow!("Cannot find path to file"))?;
+
+    // Open the file.
```
itowlson, line 189: We could replace this stanza (down to line 200) with

```
let text = tokio::fs::read_to_string(&path)
    .await
    .with_context(|| format!("Failed to read text from values file {}", path.display()))?;
```

I think?

src/commands/new.rs
```diff
@@ -178,17 +178,30 @@ impl FromStr for ParameterValue {
...
-    File::open(&file)
-        .await?
-        .read_to_end(&mut buf)
-        .await
+/// This function reads a file and parses it as TOML, then
+/// returns the resulting hashmap of key-value pairs.
+async fn values_from_file(path: impl AsRef<Path>) -> Result<HashMap<String, String>> {
+    // Get the absolute path of the file we're reading.
+    let path = absolutize(path)
+        .with_context(|| anyhow!("Cannot find path to file"))?;
```
itowlson, line 187:
```suggestion
        .with_context(|| anyhow!("Cannot find path to values file {}", path.display()))?;
```

src/commands/new.rs
cardoso: Nice catch! Works fine here.

PR #1052, opened 2023-01-25
Update toml dependency to 0.6

cardoso:
Opened a new PR to fix the commits ( cc @itowlson )

0.6 brings full TOML 1.0 compliance and several fixes.

Only had to adapt for a small breaking change:

    de::from_slice and ser::to_vec were removed, instead use from_str and to_string and convert with bytes manually

https://github.com/toml-rs/toml/blob/main/crates/toml/CHANGELOG.md#060---2023-01-23

Also a few new features which could help with parsing and editing spin.toml

Signed-off-by: Matheus Cardoso <matheus@cardo.so>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,7 +215,7 @@ dependencies = [
  "tokio-stream",
  "tokio-tar",
  "tokio-util 0.6.10",
- "toml",
+ "toml 0.5.9",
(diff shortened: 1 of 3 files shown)


itowlson (approved):
LGTM, thanks!

PR #1053, opened 2023-01-25
Update clap dependency from v3 to v4.1

cardoso:
The current clap version could make it difficult to improve the CLI's DevX with the newest CLI practices.
The update by itself already makes some improvements, but I'm looking forward to using this new version to add new capabilities.

Here's the full changelog: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#400---2022-09-28

Leaving this as draft for now while I test this further, but it seems to be working fine.
Also decided to depend on #1052 to test these crate updates together rather than after merge.

Signed-off-by: Matheus Cardoso <matheus@cardo.so>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,9 @@ async-trait = "0.1"
 bindle = { workspace = true }
 bytes = "1.1"
 chrono = "0.4"
-clap = { version = "3.1.15", features = ["derive", "env"] }
+clap = { version = "4.1", features = ["derive", "env"] }
+clap_complete = { version = "4.1", optional = true }
+clap_complete_fig = { version = "4.1", optional = true }
 cloud = { path = "crates/cloud" }
 cloud-openapi = { git = "https://github.com/fermyon/cloud-openapi" }
 comfy-table = "5.0"
@@ -80,12 +82,13 @@ vergen = { version = "7", default-features = false, features = [
 ] }
 
 [features]
-default = []
+default = ["generate-completions"]
 e2e-tests = []
 outbound-redis-tests = []
 config-provider-tests = []
 outbound-pg-tests = []
 outbound-mysql-tests = []
+generate-completions = ["dep:clap_complete", "dep:clap_complete_fig"]
 
 [workspace]
 members = [
--- a/crates/http/src/lib.rs
+++ b/crates/http/src/lib.rs
@@ -8,9 +8,9 @@ mod wagi;
 use std::{
     collections::HashMap,
     future::ready,
-    net::{Ipv4Addr, SocketAddr, ToSocketAddrs},
+    net::{Ipv4Addr, SocketAddr, SocketAddrV4, ToSocketAddrs},
     path::PathBuf,
-    sync::Arc,
+    sync::Arc, str::FromStr, fmt::Display,
 };
 
 use anyhow::{Context, Error, Result};
@@ -56,18 +56,71 @@ pub struct HttpTrigger {
     component_trigger_configs: HashMap<String, HttpTriggerConfig>,
 }
 
-#[derive(Args)]
+#[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord, Debug)]
+pub struct SocketAddress(SocketAddr);
+
+impl SocketAddress {
+    pub fn new(v4: Ipv4Addr, port: u16) -> Self {
+        Self(SocketAddr::V4(SocketAddrV4::new(v4, port)))
+    }
+}
+
+impl Into<SocketAddr> for SocketAddress {
+    fn into(self) -> SocketAddr {
+        return self.0
+    }
+}
+
+impl ToSocketAddrs for SocketAddress {
+    type Iter = <SocketAddr as ToSocketAddrs>::Iter;
+
+    fn to_socket_addrs(&self) -> std::io::Result<Self::Iter> {
+        self.0.to_socket_addrs()
+    }
+}
+
+
+
+impl Display for SocketAddress {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl Default for SocketAddress {
+    fn default() -> Self {
+        Self::new(Ipv4Addr::LOCALHOST, 3000)
+    }
+}
+
+impl FromStr for SocketAddress {
+    type Err = anyhow::Error;
+    fn from_str(addr: &str) -> Result<SocketAddress> {
+        let addrs: Vec<SocketAddr> = addr.to_socket_addrs()?.collect();
+        // Prefer 127.0.0.1 over e.g. [::1] because CHANGE IS HARD
+        if let Some(addr) = addrs
+            .iter()
+            .find(|addr| addr.is_ipv4() && addr.ip() == Ipv4Addr::LOCALHOST)
+        {
+            return Ok(Self(*addr));
+        }
+        // Otherwise, take the first addr (OS preference)
+        addrs.into_iter().next().context("couldn't resolve address").map(Self)
+    }
+}
+
+#[derive(Args, Clone)]
 pub struct CliArgs {
     /// IP address and port to listen on
-    #[clap(long = "listen", default_value = "127.0.0.1:3000", value_parser = parse_listen_addr)]
-    pub address: SocketAddr,
+    #[arg(long = "listen", default_value_t = SocketAddress::default())]
+    pub address: SocketAddress,
 
     /// The path to the certificate to use for https, if this is not set, normal http will be used. The cert should be in PEM format
-    #[clap(long, env = "SPIN_TLS_CERT", requires = "tls-key")]
+    #[arg(long, env = "SPIN_TLS_CERT", requires = "tls_key")]
     pub tls_cert: Option<PathBuf>,
 
     /// The path to the certificate key to use for https, if this is not set, normal http will be used. The key should be in PKCS#8 format
-    #[clap(long, env = "SPIN_TLS_KEY", requires = "tls-cert")]
+    #[arg(long, env = "SPIN_TLS_KEY", requires = "tls_cert")]
     pub tls_key: Option<PathBuf>,
 }
 
@@ -177,11 +230,10 @@ impl TriggerExecutor for HttpTrigger {
         }
 
         if let Some(tls) = tls {
-            self.serve_tls(listen_addr, tls).await?
+            self.serve_tls(listen_addr, tls).await
         } else {
-            self.serve(listen_addr).await?
-        };
-        Ok(())
+            self.serve(listen_addr).await
+        }
     }
 }
 
@@ -293,7 +345,7 @@ impl HttpTrigger {
             .body(Body::empty())?)
     }
 
-    async fn serve(self, listen_addr: SocketAddr) -> Result<()> {
+    async fn serve(self, listen_addr: SocketAddress) -> Result<()> {
         let self_ = Arc::new(self);
         let make_service = make_service_fn(|conn: &AddrStream| {
             let self_ = self_.clone();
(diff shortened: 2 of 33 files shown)


crates/http/src/lib.rs
```diff
@@ -63,11 +63,11 @@ pub struct CliArgs {
     pub address: SocketAddr,
 
     /// The path to the certificate to use for https, if this is not set, normal http will be used. The cert should be in PEM format
-    #[clap(long, env = "SPIN_TLS_CERT", requires = "tls-key")]
+    #[clap(long = "tls-cert", env = "SPIN_TLS_CERT", requires = "tls_key")]
```
cardoso, line 66: This is a tricky change: clap v4 doesn't make arguments kebab case from snake case ids anymore.

Tests are passing locally and I did some searching, but I'm not sure this isn't causing a bug or breaking change somewhere.

cardoso:
Tests are failing in CI but passing locally, so I'm opening this up for review while I figure out if it's related to any changes here.

vdice (commented):
Thanks for this update PR @cardoso!

Indeed, I think there may be a few spots to adjust per updated clap usage; both of the [test failures](https://github.com/fermyon/spin/actions/runs/4014224262/jobs/6894693899) seem to arise from either a flag now being unexpected (`--env`) or a value not received (bindle connection string):

```
test integration_tests::e2e_tests::test_headers_env_routes has been running for over 60 seconds
error: unexpected argument '--env' found
```
and

```
Error: Failed to prepare configuration

Caused by:
    Component hello requires a Bindle connection but none was specified
test integration_tests::e2e_tests::test_using_parcel_as_module_source ... FAILED
 ```

File: crates/trigger/src/cli.rs [hunk 2]
```diff
@@ -24,69 +24,67 @@ pub const SPIN_WORKING_DIR: &str = "SPIN_WORKING_DIR";
...
 
     /// Set the static assets of the components in the temporary directory as writable.
-    #[clap(long = "allow-transient-write")]
+    #[arg(long)]
     pub allow_transient_write: bool,
 
     /// Configuration file for config providers and wasmtime config.
-    #[clap(
-        name = RUNTIME_CONFIG_FILE,
+    #[arg(
```
vdice, line 77: Optional: Noticing that we should be able to reduce to `long` in the line below to inherit `"runtime-config-file"` by default, right?

crates/http/src/lib.rs
```diff
@@ -59,15 +59,15 @@ pub struct HttpTrigger {
 #[derive(Args)]
 pub struct CliArgs {
     /// IP address and port to listen on
-    #[clap(long = "listen", default_value = "127.0.0.1:3000", value_parser = parse_listen_addr)]
+    #[arg(long = "listen", default_value_t = std::net::SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 3000)), value_parser = parse_listen_addr)]
```
vdice, line 62: Q: Can you elaborate on this update (use of `default_value_t`)?  Is it utilizing a performance or efficiency gain w/ clap 4.x.x?

vdice (commented):
Had a couple Qs/comments, otherwise looking great to me.  Appreciative of the detail/attention required for this update!

crates/http/src/lib.rs
itowlson: I think this possibly relates to correctness more than efficiency - by writing the _actual constructor call_ rather than a string, the compiler can check that the default value is well formed and of the correct type.

src/commands/build.rs
```diff
@@ -9,21 +9,21 @@ use super::up::UpCommand;
...
+    #[arg(
+        short = 'f',
+        long = "file",
+        id = APP_CONFIG_FILE_OPT,
+    )]
     pub app: Option<PathBuf>,
 
     /// Run the application after building.
-    #[clap(name = BUILD_UP_OPT, short = 'u', long = "up")]
+    #[arg(id = BUILD_UP_OPT, short, long)]
```
vdice, line 23: It looks like clap has chosen `-U` and `--UP` here for some reason:

Screenshot 2023-01-30 at 4 24 15 PM

Can we figure out why and fix?  (Or perhaps go back to setting it explicitly to `-u` and `--up`?)

src/commands/build.rs
```diff
@@ -9,21 +9,21 @@ use super::up::UpCommand;
...
     pub up: bool,
 
-    #[clap(requires = BUILD_UP_OPT)]
+    #[arg(requires = BUILD_UP_OPT, hide = true)]
```
vdice, line 26: This is a good chance to check in on what we'd like to see re: these up args.

Previous to this PR, we have:

spin-build-old

With this PR, there is no mention of up args:

Screenshot 2023-01-30 at 4 26 12 PM

Admittedly, the old style of just showing 'UP_ARGS' is perhaps not ideal (also b/c `spin up` technically takes options/flags and not args).  But I do want to check with @itowlson to get input here.

As an aside, it would be nifty to inject all of the valid/applicable spin up flags/opts here -- I wonder if we can do so via clap?  Regardless, this would/should be a follow-up feature, to keep this PR focused purely on updating clap usage.

src/bin/spin.rs
```diff
@@ -38,26 +38,26 @@ fn version() -> &'static str {
...
 #[derive(Parser)]
-#[clap(
-    name = "spin",
+#[command(
+    id = "spin",
     version = version(),
 )]
 enum SpinApp {
-    #[clap(subcommand, alias = "template")]
+    #[command(subcommand, alias = "template")]
```
vdice, line 46: (Not a comment on this particular line.)

Using this opportunity to point out to other reviewers that clap 4.0.0+ has updated default display order.  Whereas in 3.x.x the commands are alphabetically sorted, 4+ appears to defer to the order we declare commands here in code.

### Previous to this PR (looking at `SUBCOMMANDS`):

Screenshot 2023-01-30 at 4 43 25 PM

### With this PR (looking at `Commands`):

Screenshot 2023-01-30 at 4 43 06 PM

@itowlson (and others) -- Do we have a preference on ordering?  If so, should we ensure commands are listed here in the order we'd like them displayed?  (Or, I imagine there is a clap 4+ setting we might utilize, regardless of order in code.)

vdice (commented):
@cardoso took one more close look at the updated cli output and raised a few more comments around notable changes, mostly to check in with others...

src/commands/build.rs
itowlson: It would be good to show _something_ to indicate that `up` options can be provided if `--up` is present, though if the worst comes to the worst we can mention it in `--up` doc string instead.

src/bin/spin.rs
itowlson: Eek, good catch.  I'd be inclined to go for alphabetical - some tools (such as Cargo do try to group similar commands together, but I'm not sure if we have a strong enough notion of that at the moment.  And if we do it, we should approach it intentionally rather than as part of a package rev.

src/commands/new.rs
```diff
@@ -21,48 +21,48 @@ pub struct TemplateNewCommandCore {
...
 
     /// The directory in which to create the new application or component.
     /// The default is the name argument.
-    #[clap(short = 'o', long = "output")]
+    #[arg(short, long = "output")]
     pub output_path: Option<PathBuf>,
 
     /// Parameter values to be passed to the template (in name=value format).
-    #[clap(short = 'v', long = "value", multiple_occurrences = true)]
+    #[arg(short, long)]
```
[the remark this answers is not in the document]
cardoso: Nice catch. I totally overlooked the plural.

crates/trigger/src/cli.rs
cardoso: The long will take the id and convert _ to - , but won't convert to lowercase. So in this case that won't work since it's all caps behind that const...

crates/http/src/lib.rs
cardoso: I think I was going somewhere with that. Ended up leaving it due to the reasons @itowlson mentioned. Also seems to be preferred when reading clap's docs.

src/commands/build.rs
cardoso: I started looking into a better way to display it and forgot to leave the change out for now.

I don't know if there's a right way to do this. I guess the flag is trying to do too much?

src/bin/spin.rs
cardoso: Should I do anything here? Change the order by hand, force the old way, leave it be for now?

Using a flag would be fighting clap quite a bit as it wants to be WYSIWYG going forward.

```
New behavior:

    Sorted order is derived by default
    Sorting is turned on by cmd.next_display_order(None)
        This is not recursive, it must be set on each level
    The display order incrementing is mixed with arguments
        This does make it slightly more difficult to predict
```

from: https://github.com/clap-rs/clap/pull/3975

src/commands/build.rs
cardoso: Oops. Forgot to reinclude when re-writing. Same as above. Clap keeps the uppercase from the const.

tests/integration.rs
```diff
@@ -885,16 +885,24 @@ mod integration_tests {
...
+
+            let args = vec![
+                vec!["up"],
+                spin_args.to_vec(),
+                bindle_args,
+                env_args,
+                trigger_args,
+            ]
+            .into_iter()
+            .flatten();
```
cardoso, line 905: Probably worth mentioning I had to change this test. Clap changed the way it parses, so that --env was being seen as part of spin trigger and not spin up.

I could try to force it, but it's not going to be pretty. The old behavior seemed less correct since --env appeared after other args which were going to spin trigger.

src/commands/build.rs
vdice: Thanks @cardoso; sticking with current behavior is great for now.  I'll create a separate issue for potentially improving things here.

src/commands/build.rs
```diff
@@ -1,51 +1,55 @@
...
 
-    /// Run the application after building.
-    #[clap(name = BUILD_UP_OPT, short = 'u', long = "up")]
+#[derive(Args, Debug)]
+pub struct BuildUpArgs {
+    #[arg(short, long, group = "UpArgs", required=true)]
     pub up: bool,
+    #[command(flatten, next_help_heading="Up")]
+    pub args: UpArgs
+}
```
cardoso, line 28: I think this worked and it's far better than I expected.

If we separate UpArgs into its own struct, clap implicitly creates a group with the same name as the struct.

Now all I had to do is create separate group for the up flag with the up args flattened into it and mark up as required and as part of the implicit "UpArgs" group.

Screenshot 2023-02-02 at 20 56 05

itowlson: Oh nice!

cardoso: Oops! It's not working right yet, but I think I'm getting close.

Cargo.toml
```diff
@@ -80,12 +82,13 @@ vergen = { version = "7", default-features = false, features = [
 ] }
 
 [features]
-default = []
+default = ["generate-completions"]
```
cardoso, line 85: I added completions here just to make sure the less conventional features like build's `--up` option aren't breaking it.

I'll make sure to remove once this is ready for review again.

src/bin/spin.rs
vdice: Maybe the best bet is to change the order by hand, assuming we'd like to retain the current alphabetical order.

Cargo.toml
itowlson: Ooh, if this provides us with a better and more consistent way to https://github.com/fermyon/spin/pull/1032 then I'd love to have your feedback on that PR (or your own PR).  Thanks!  (I agree on not trying to shoehorn it into this one, of course.)

cardoso:
Closing this since it's fallen behind. Will open a new one soon.

vdice:
Thanks @cardoso and apologies -- this has definitely been a busy period for the Spin project and particularly the CLI.  Happy to help in any way I can.

PR #1060, opened 2023-01-25
Fix reserved name causing new Rust app to not build

### LINKED ISSUE #1038: "Reserved" names for components
If I create a Spin component using the `http-rust` template and I name it `static`, the rust component cannot be built, as the function will take on that name.

I've made similar types of "errors" with the .NET SDK 🤔 https://github.com/fermyon/spin-dotnet-sdk/issues/33 and https://github.com/fermyon/spin-dotnet-sdk/issues/36

Having a way to make sure the naming of a component, does not impact the ability to compile it through "reserved" names or appending / prepending how that name is being used in the component would be good to avoid anyone having to spent time trying to figure out while a component will fail on build right after being crated through a template.

itowlson:
Fixes #1038.

### DIFF
--- a/templates/http-rust/content/src/lib.rs
+++ b/templates/http-rust/content/src/lib.rs
@@ -6,7 +6,7 @@ use spin_sdk::{
 
 /// A simple Spin HTTP component.
 #[http_component]
-fn {{project-name | snake_case}}(req: Request) -> Result<Response> {
+fn handle_{{project-name | snake_case}}(req: Request) -> Result<Response> {
     println!("{:?}", req.headers());
     Ok(http::Response::builder()
         .status(200)


kate-goldenring:
@itowlson like this approach of prefixing handle. Is there a way we can also add these changes to our other SDK's such as the dotnet one which is also referenced in the issue? I am not sure we can say that this PR "fixes" https://github.com/fermyon/spin/issues/1038 without resolving it for all sdks

itowlson:
I think the .NET issue is already fixed (and if not then we probably need a separate issue in that repo to track it there), and I went through the other templates in this repo and I don't think any of them were affected.  Not sure if there's a way to demonstrate that in a review though!

kate-goldenring (approved):

michelleN (approved):

fibonacci1729 (approved):

fibonacci1729:
I am wondering if this exposes a similar collision with "handle_http_request" if the compinent I'd template parameter is set to "handle-request". I'll verify this tomorrow unless you get to it first.

itowlson:
@fibonacci1729 Tried a handler called `handle_http_request` and it built fine.  I assume the WIT generated name is in a `spin_http` module so wouldn't clash?  (if that's what you're concerned about)

fibonacci1729:
Precisely! Thank you @ivan. I checked the SDK too and confirmed.

PR #1062, opened 2023-01-26
`spin templates list`: if empty, offer to install

itowlson:
Previously, if you ran `spin templates list` with no templates, it told you how to install but didn't offer to do it.  Now it offers:

```
$ spin templates list
You don't have any templates yet. Would you like to install the default set? no
You can install the default templates later with 'spin install --git https://github.com/fermyon/spin'

$ ./target/debug/spin templates list
You don't have any templates yet. Would you like to install the default set? yes
Copying remote template source
Installing template redis-go...
Installing template http-empty...
Installing template http-rust...
Installing template http-grain...
# rest elided
```

### DIFF
--- a/src/commands/templates.rs
+++ b/src/commands/templates.rs
@@ -13,6 +13,10 @@ use spin_templates::{
 const INSTALL_FROM_DIR_OPT: &str = "FROM_DIR";
 const INSTALL_FROM_GIT_OPT: &str = "FROM_GIT";
 
+const DEFAULT_TEMPLATES_INSTALL_PROMPT: &str =
+    "You don't have any templates yet. Would you like to install the default set?";
+const DEFAULT_TEMPLATE_REPO: &str = "https://github.com/fermyon/spin";
+
 /// Commands for working with WebAssembly component templates.
 #[derive(Subcommand, Debug)]
 pub enum TemplateCommands {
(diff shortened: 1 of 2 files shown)


src/commands/templates.rs
```diff
@@ -265,3 +269,36 @@ fn list_warn_reason_text(reason: &InstalledTemplateWarning) -> String {
...
+    let should_install = dialoguer::Confirm::new()
+        .with_prompt(DEFAULT_TEMPLATES_INSTALL_PROMPT)
+        .default(true)
+        .interact_opt()?;
+    if should_install == Some(true) {
+        install_default_templates().await?;
+        Ok(Some(template_manager.list().await?.templates))
+    } else {
+        println!(
+            "You can install the default templates later with 'spin install --git {}'",
```
kate-goldenring, line 285: missing `templates` subcommand... which was previously the case too. oops
```suggestion
            "You can install the default templates later with 'spin templates install --git {}'",
```

kate-goldenring (approved):
Love it! Just one nit

src/commands/templates.rs
itowlson: Yikes!  Thanks for the spot Kate, how embarrassing!

PR #1066, opened 2023-01-26
Run code coverage only on a schedule

itowlson:


vdice (approved):
Assume we'll fix lint errors in a separate PR; LGTM!

itowlson:
I think this verifies the syntax even though the flake 🎉 failed 🎉 and the lints have decided to pick today to go berserk:

https://github.com/itowlson/spin/actions/runs/4018501566

vdice:
> I think this verifies the syntax

Yup, the workflow wouldn't have run if there were syntax errors.

By the way, good call to add the workflow_dispatch support for manual triggers as well 👏

PR #1068, opened 2023-01-26
Pin Rust version in GitHub actions

itowlson:
It don't know how to centralise the value or to make magic PRs to update it but NO MORE SURPRISE LINTS PLEASE

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,10 +77,10 @@ jobs:
       - name: "Install Wasm Rust target"
         run: rustup target add wasm32-wasi
 
-      - name: Install latest Rust stable toolchain
+      - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.66
           default: true
           target: ${{ matrix.config.target }}
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,10 +19,10 @@ jobs:
     runs-on: "ubuntu-latest"
     steps:
       - uses: actions/checkout@v2
-      - name: Install latest Rust stable toolchain
+      - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.66
           default: true
           components: clippy, rustfmt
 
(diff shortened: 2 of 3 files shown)


vdice (approved):
As for centralizing the value, we can utilize a [GH Action variable](https://docs.github.com/en/actions/learn-github-actions/variables#defining-configuration-variables-for-multiple-workflows).  Or, perhaps we could parse the value set in [spin's Cargo.toml](https://github.com/fermyon/spin/blob/main/Cargo.toml#L7)  in the workflows?

Either way, in support of getting this change in to unblock CI for other PRs and then working on a follow-up.

itowlson:
If we used a config variable would we be able to do a "run with a different value to confirm that we are now ready to rev the variable" run?  I'm not very familiar with this stuff, sorry.

vdice:
> If we used a config variable would we be able to do a "run with a different value to confirm that we are now ready to rev the variable" run? I'm not very familiar with this stuff, sorry.

Perhaps, but maybe the need you've expressed is better served by the latter approach: having the `rust_version` set in the Cargo.toml be the canonical version.  We'd then get signal via CI for PRs that rev that version.  Plus, it's eligible for all to change (via PR) as opposed to a variable that only those with access to this repo's settings may modify.  WDYT?

itowlson:
Yeah the value from Cargo.toml is tempting.  It does mean we tie the _minimum_ version to the version we'd like to get lints for.  But it also means we _prove_ that Spin builds with the declared minimum, which seems like a good plan!

lann:
I'm a little wary of pinning our release toolchain without some dependabotesque upgrade automation in place. We don't want to release with a version that is missing a security patch...

.github/workflows/release.yml
```diff
@@ -77,10 +77,10 @@ jobs:
       - name: "Install Wasm Rust target"
         run: rustup target add wasm32-wasi
 
-      - name: Install latest Rust stable toolchain
+      - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.66.1
```
lann, line 83: I believe this should work, and would at least pick up security patches for a while

```suggestion
          toolchain: 1.66
```

itowlson:
Completely agree on wanting something like dependabot for this - Danielle suggested Renovate but I'm a bit lost in the docs for that right now.  We would definitely want to revisit this and how it aligns with the MSRV in Cargo.toml and how we maintain it and all that stuff - I'll be honest and admit that this is a bit of an IVAN IS OVER THE TOOLS FAILING ALL THE TIME rage-PR...

itowlson:
There's a Renovate app for GitHub: https://github.com/marketplace/renovate

lann (approved):

Issue #1059, opened 2023-01-25
sdk: Pick a `rust-version` for `spin-sdk`


lann:
We can make nicer errors for people on too-old Rust versions by adding a [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) to the SDK's `Cargo.toml`.

itowlson:
There was discussion a while back of documenting a minimum Rust version for building Spin itself, and we punted because it's not clear how we would maintain that.  This risks running into the same problem unless we accompany it with some CI that builds (a minimal consumer of?) the SDKs _using the nominated minimum version_.  So we should plan for that as part of this work too.

lann:
That is a good idea, but we would still get incremental benefit from just adding it. If it gets out of sync at worst you'll end up with the current less-helpful errors.

itowlson:
Core covered by #1071; moving remaining suggestion to new issue.

PR #1071, opened 2023-01-27
Provide MSRV in Rust SDK

itowlson:
Partial fix for #1059.

This uses 1.64 as that is the MSRV for building Spin itself.  I tested with Rust 1.64 and a basic app worked fine:

image

### DIFF
--- a/sdk/rust/Cargo.toml
+++ b/sdk/rust/Cargo.toml
@@ -3,6 +3,7 @@ name = "spin-sdk"
 version = { workspace = true }
 authors = { workspace = true }
 edition = { workspace = true }
+rust-version = "1.64"
 
 [lib]
 name = "spin_sdk"


lann (approved):

fibonacci1729 (approved):

sdk/rust/Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "spin-sdk"
 version = { workspace = true }
 authors = { workspace = true }
 edition = { workspace = true }
+rust-version = "1.64"
```
michelleN, line 6: Curious to understand why we went with 1.64. I remember us fixing some lint errors last Spin release because rust 1.66 was released the same day. Could we use that version or is another reason we picked 1.64 here?

itowlson: 1.64 is the MSRV for Spin itself, so this follows that.  And... if everything builds and works with 1.64, why force people to move to the bleeding edge?  This doesn't _stop_ people using 1.66 or 1.67 or whatever, as we do ourselves - it just provides a better error message if they are on a version of Rust that is unable to build the SDK.

michelleN (approved):

sdk/rust/Cargo.toml
michelleN: sgtm!

mikkelhegn:
@itowlson Does this need an update to docs?

itowlson:
I don't _think_ so - if anything, it should reduce the need for docs because people with old Rusts will see an informative message about versions rather than needing to pore through the docs for version requirements! 
There's something to be said for documenting the minimum tool versions for various languages, but I reckon that's independent of this.  Happy to discuss in more depth though!

PR #1079, opened 2023-01-27
Support SSL/TLS in `outbound-mysql`

calebschoepp:
This is a hack workaround to support SSL/TLS in the `outbound-mysql` host component. Our dependency `mysql_async` wasn't properly parsing the ssl parameters so we do it ourselves here and then hide that parameter from the `mysql_async`.

I've tested without SSL/TLS against a local database and against Planetscale with SSL/TLS.

### DIFF
--- a/crates/outbound-mysql/src/lib.rs
+++ b/crates/outbound-mysql/src/lib.rs
@@ -1,9 +1,10 @@
 use mysql_async::consts::ColumnType;
-use mysql_async::{from_value_opt, prelude::*};
+use mysql_async::{from_value_opt, prelude::*, Opts, OptsBuilder, SslOpts};
 pub use outbound_mysql::add_to_linker;
 use spin_core::HostComponent;
 use std::collections::HashMap;
 use std::sync::Arc;
+use url::Url;
 use wit_bindgen_wasmtime::async_trait;
 
 wit_bindgen_wasmtime::export!({paths: ["../../wit/ephemeral/outbound-mysql.wit"], async: *});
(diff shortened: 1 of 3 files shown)


itowlson:
@etehtsea I think you looked at SSL for Postgres, and also did the original MySQL support - would you be able to take a look at this please?  Thanks!

crates/outbound-mysql/src/lib.rs
```diff
@@ -244,11 +245,89 @@ impl OutboundMysql {
...
-    let connection_pool = mysql_async::Pool::new(address);
+    let opts = build_opts(address)?;
+
+    let connection_pool = mysql_async::Pool::new(opts);
 
     connection_pool.get_conn().await
 }
 
+/// The mysql_async crate does not currently support SSL opts properly. This function is a workaround to
+/// manually set SSL if opts if the use requests them. We only support ssl-mode in the query as per
```
itowlson, line 256: super nitty typo 
```suggestion
/// manually set SSL if opts if the user requests them. We only support ssl-mode in the query as per
```

crates/outbound-mysql/src/lib.rs
```diff
@@ -244,11 +245,89 @@ impl OutboundMysql {
...
+/// https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-connp-props-security.html#cj-conn-prop_sslMode.
+fn build_opts(address: &str) -> Result<Opts, mysql_async::Error> {
+    let url = Url::parse(address)?;
+
+    let is_ssl_param = |s: &str| ["ssl-mode", "sslmode"].contains(&s.to_lowercase().as_str());
```
itowlson, line 261: To me, this would jump out as a function more easily if it was declared `fn is_ssl_param(...) -> bool` rather than using let.  (You could still have it as local rather than module-scope.)  Subjective though!

crates/outbound-mysql/src/lib.rs
```diff
@@ -244,11 +245,89 @@ impl OutboundMysql {
...
+    let query_without_ssl: Vec<(_, _)> = url
+        .query_pairs()
+        .filter(|(k, _v)| !is_ssl_param(k))
+        .collect();
+    let mut cleaned_url = url.clone();
+    cleaned_url.set_query(None);
+    for pair in query_without_ssl {
+        cleaned_url
+            .query_pairs_mut()
+            .append_pair(&pair.0.to_string()[..], &pair.1.to_string()[..]);
```
itowlson, line 276: You might be able to use `extend_pairs` to do this all in one go but I don't know the API so I could be wrong!

crates/outbound-mysql/src/lib.rs
```diff
@@ -244,11 +245,89 @@ impl OutboundMysql {
...
+    }
+
+    Ok(OptsBuilder::from_opts(cleaned_url.as_str())
```
itowlson, line 279: This may be a stupid question but... do we need to do the `cleaned_url` thing?  Could we not Do the `from_opts` with the original URL, and then call `ssl_opts` to override whatever incorrect SSL option was inferred from that original URL?

crates/outbound-mysql/src/lib.rs
```diff
@@ -244,11 +245,89 @@ impl OutboundMysql {
...
+fn build_opts(address: &str) -> Result<Opts, mysql_async::Error> {
+    let url = Url::parse(address)?;
+
+    let is_ssl_param = |s: &str| ["ssl-mode", "sslmode"].contains(&s.to_lowercase().as_str());
+
+    let use_ssl = url
+        .query_pairs()
+        .any(|(k, v)| is_ssl_param(&k) && v.to_lowercase() != "disabled");
+
+    let query_without_ssl: Vec<(_, _)> = url
```
dicej, line 267: This makes me wish the `url` crate had an API to make removing query parameters easier -- it's quite a process.

dicej (approved):
LGTM.  Like we discussed, it would be great to move this responsibility into the `mysql_async` crate itself if you haven't already opened an issue there about it.

calebschoepp:
> LGTM. Like we discussed, it would be great to move this responsibility into the `mysql_async` crate itself if you haven't already opened an issue there about it.

I'm in the process of opening an issue with `mysql_async` to get that process started.

crates/outbound-mysql/src/lib.rs
calebschoepp: Valid question, this is how the fix originally worked when I was debugging with Joel. However, the `mysql_async` library blows up if you pass it a query parameter with `sslMode` in it. This is why I have to clean it out of the url. Hopefully they can fix this in upstream when I file the issue.

crates/outbound-mysql/src/lib.rs
calebschoepp: Good call

crates/outbound-mysql/src/lib.rs
calebschoepp: Yep, I like it. Done.

crates/outbound-mysql/src/lib.rs
itowlson: Ah, then maybe clarify the comment to say you _can't even pass_ a SSL parameter without it blowing up, more than "doesn't currently support properly" which I took to mean "didn't recognise the parameter / recognised it wrong."  A link to the upstream issue would be really good too, so that a maintainer can easily check "is this fixed yet."

itowlson (approved):
Suggestion for clarification on the comment but otherwise LGTM

crates/outbound-mysql/src/lib.rs
calebschoepp: Will do.

PR #1081, opened 2023-01-30
Show common trigger options on `spin up --help`

### LINKED ISSUE #1078: Hidden cli flags on `spin up`
Some cli flags are not displayed for `spin up --help`.  For example `--log-dir`, `--follow`, and `--follow-all`.  

I believe they're trigger specific but should there be a standard for logging or  a way to view available flags without `spin trigger http --help`?

itowlson:
Fixes #1078, #815.

When the user runs `spin up --help` and there is no application (that is, the default application file doesn't exist), it now shows the common trigger CLI flags.

To maximise consistency with the standard help flow (where there _is_ an application), this gets the trigger help options by handing off to a trigger command; however in the no-app case it is a hidden no-op trigger type that has only the base set of trigger CLI flags.

If an application manifest is present, its trigger type is invoked as usual (e.g. if there is a HTTP app present, the user still sees the `--listen` and TLS flags).

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -205,3 +205,25 @@ async fn warn_slow_wasm_build() {
     println!("Preparing Wasm modules is taking a few seconds...");
     println!();
 }
+
+pub mod help {
+    use super::*;
+
+    /// Null object to support --help-args-only in the absence of
+    /// a `spin.toml` file.
+    pub struct HelpArgsOnlyTrigger;
+
+    #[async_trait::async_trait]
+    impl TriggerExecutor for HelpArgsOnlyTrigger {
+        const TRIGGER_TYPE: &'static str = "help-args-only";
+        type RuntimeData = ();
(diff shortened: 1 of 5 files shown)


rajatjindal:
Hi @itowlson please pardon me if I understood something incorrectly, but shouldn't this wait for https://github.com/fermyon/spin/pull/974 to complete?

fibonacci1729 (approved):
LGTM

itowlson:
@rajatjindal No, this is completely orthogonal.  The issue that this fixes is that if you are in a directory without a `spin.toml` file then `spin up --help` doesn't show the flags that are defined on `spin_trigger::cli` instead of directly on `up`.  Even if we later change or reshuffle, those options, the principle that "`spin up --help` should show all the options" remains true.

As an example of non-logging optioins, the Wasm cache, transient write, and runtime config flags are defined on the trigger and currently disappear if there is no `spin.toml` in the current directory.

PR #1085, opened 2023-01-31
Record where templates were installed from

itowlson:
This adds a `spin templates list --verbose` flag which adds an "installed from" column to the table:

```
+-----------------------------------------------------------------------------------------------------------------+
| Name                Description                                          Installed from                         |
+=================================================================================================================+
| http-grain          HTTP request handler using Grain                     /home/ivan/github/spin                 |
| http-js             HTTP request handler using Javascript                https://github.com/fermyon/spin-js-sdk |
| http-php            HTTP request handler using PHP                       /home/ivan/github/spin                 |
```

If there is no record (as there won't be for templates installed before this patch) then the column is blank; it will get populated on next update.

This will also provide a foundation for a simpler update/sync experience in a future thrilling episode.

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -125,7 +125,7 @@ impl TemplateManager {
 
         for template_dir in template_dirs {
             let install_result = self
-                .install_one(&template_dir, options, reporter)
+                .install_one(&template_dir, options, source, reporter)
                 .await
                 .with_context(|| {
                     format!("Failed to install template from {}", template_dir.display())
@@ -146,6 +146,7 @@ impl TemplateManager {
         &self,
         source_dir: &Path,
         options: &InstallOptions,
+        source: &TemplateSource,
         reporter: &impl ProgressReporter,
     ) -> anyhow::Result<InstallationResult> {
         let layout = TemplateLayout::new(source_dir);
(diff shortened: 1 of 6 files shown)


crates/templates/src/manager.rs
```diff
@@ -340,9 +345,22 @@ async fn copy_template_into(
...
 }
 
+fn write_install_record(dest_dir: &Path, source: &TemplateSource) {
+    let layout = TemplateLayout::new(dest_dir);
+    let install_record_path = layout.install_record_file();
+
+    // A failure here shouldn't fail the install
+    let install_record = source.to_install_record();
+    if let Ok(record_text) = toml::to_string_pretty(&install_record) {
+        let _ = std::fs::write(install_record_path, record_text);
```
fibonacci1729, line 360: nit: `_ = std::fs::write(install_record_path, record_text);`

fibonacci1729, line 360:
```suggestion
       _ = std::fs::write(install_record_path, record_text);
```

itowlson: WHAT

I DID NOT KNOW YOU COULD DO THIS

THANK YOU BRIAN

itowlson: Can't believe Clippy gets in my face about double references but doesn't bother to tell me about things like this

PR #1092, opened 2023-01-31
chore(*): v0.8.0 version bumps

vdice:
Version bumps in anticipation of a v0.8.0 release

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -570,7 +570,7 @@ dependencies = [
 
 [[package]]
 name = "cloud"
-version = "0.7.1"
+version = "0.8.0"
 dependencies = [
  "anyhow",
  "cloud-openapi",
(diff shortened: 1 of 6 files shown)


itowlson (approved):

itowlson:
I see in previous version bump PRs we've done the downloader, which is no longer in this repo - is that on our checklist?  (I don't doubt you... I'm just double-checking!)

vdice:
Good reminder @itowlson; now that the install script is in [fermyon/developer](https://github.com/fermyon/developer/blob/main/downloads/install.sh) it is liable to drift (just checked; it still contains the `v0.5.0` string 🙈 ).

Shall we create a follow-up issue to decide on a good mechanism to fix this drift?

itowlson:
Ah, I think it might look for the latest (there are _moar_ greps and seds to set the version) - the help text mentions 0.5.0 but it's not crucial.

PR #1095, opened 2023-02-01
Upgrade templates without git incantation

### LINKED ISSUE #990: `spin templates update` should try to update all known templates
Just upgraded to 0.7.0 and `spin add` didn't work, which I figured would be because the templates needed updating too.

Updating from there should be `spin templates update`, however at present we need to run:

`spin templates install --git https://github.com/fermyon/spin --update`

itowlson:
Fixes #990.

Some details on which I would like feedback:

* I have proposed the name `spin templates upgrade` rather than `update`.  This aligns with the plugins UI, where 'update' refers to updating the _catalogue_ and `upgrade` is used to actually install a new version.  However, `update` is established in `spin templates install --update` (easy to alias), and is misleading if a user is moving back to an earlier version of Spin.
* ~I have defaulted it to upgrading templates from _all_ known repositories.  If the user wants to be selective, they can pass `--ask`.  An alternative approach would be to prompt by default, and require `--all` to upgrade all without prompting.~  After feedback, it now prompts by default, with a `--all` flag to upgrade all without prompting.
* I have allowed the user to upgrade a specific repo (optionally, to a specific tag) via the `--repo/--branch` options.  This is an exact synonym for `install --update` - is it worth bothering?  I'm leaning against it now.  (The user could pass `--ask` and just select the one repo, but could not then select a specific tag.)

~The report at the end is a work in progress.  It will not, actually, boo errors (deserve it though they might).~  This is now fixed: have at it.

The ~`--ask`~ prompt UI is via the `dialoguer` crate and looks like this:

image

I'm not much in love with it but I don't want to invest in re-implementing it.  If there are better Rust terminal libraries then I'm happy to change.

### DIFF
--- a/crates/templates/src/source.rs
+++ b/crates/templates/src/source.rs
@@ -74,6 +74,16 @@ impl TemplateSource {
             }
         }
     }
+
+    // Sorry I know this is a bit ugly
+    /// For a Git source, resolves the tag to use as the source.
+    /// For other sources, returns None.
+    pub async fn resolved_tag(&self) -> Option<String> {
+        match self {
+            Self::Git(g) => version_matched_tag(g.url.as_str(), &g.spin_version).await,
+            _ => None,
+        }
+    }
 }
 
 pub(crate) struct LocalTemplateSource {
--- a/crates/templates/src/template.rs
+++ b/crates/templates/src/template.rs
@@ -163,6 +163,17 @@ impl Template {
         }
     }
 
+    /// The Git repository from which the template was installed, if
+    /// it was installed from Git; otherwise None.
+    pub fn source_repo(&self) -> Option<&str> {
+        // TODO: this is kind of specialised - should we do the discarding of
+        // non-Git sources at the application layer?
+        match &self.installed_from {
+            InstalledFrom::Git(url) => Some(url),
+            _ => None,
+        }
+    }
+
     /// A human-readable description of where the template was installed
     /// from.
     pub fn installed_from_or_empty(&self) -> &str {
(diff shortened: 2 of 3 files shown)


rajatjindal:
> prompt by default, and require --all to upgrade all without prompting

I like this one. this is also inline with our `spin new --accept-defaults` UX.

> I have allowed the user to upgrade a specific repo (optionally, to a specific tag) via the --repo/--branch options

hm, i can see it can be useful when user is working on changing template from e.g. fork of a repo.

mikkelhegn (approved):
No approving code, just experience

itowlson:
@rajatjindal @mikkelhegn Thanks for the feedback - it now prompts by default.

kate-goldenring:
> However, update is established in spin templates install --update (easy to alias), and is misleading if a user is moving back to an earlier version of Spin.

So are we keeping `install --update` for backwards compat? My vote would be to change it to `install --upgrade` if we are choosing to distinguish update and upgrade. Better to eliminate the confusion now than it become bothersome post 1.0

itowlson:
Yes, if we agree that the command should be called `upgrade`, we should change the flag to `--upgrade` with an alias to `--update` for back compat.  I was waiting for agreement on the new command name but there seems to be general acceptance.  I've made that change.

(I have left the Rust field as being called `update` for now, because that term is widely used in the template installation code.  It would be good to change it to stay in sync with the UI, but I don't want to clutter this PR with renaming noise.)

fibonacci1729 (approved):

fibonacci1729:
LGTM code-wise. I am +1 for the experience discussed above.

rawkode:
Awesome upgrade! 💯

PR #1096, opened 2023-02-01
Use term application rather than project in `spin new` and `spin add`

### LINKED ISSUE #1074: Remove confusing "project" terminology from `spin new` and `spin add` components
From what I can tell the terminology we expose to our users is that a Spin "application" consists of multiple "components". However, in the `spin new` and `spin add` flows we refer to what they're making as a "project". I find this confusing because I wasn't sure if "project" was the same thing as an "application".

I haven't found anything in the documentation that indicates we want to use the "project" terminology. I propose that we modify the `spin new` and `spin add` flows as such:

## `spin new`

Old: 
```
$ spin new
Pick a template to start your project with: http-rust (HTTP request handler using Rust)
Enter a name for your new project: My project
Project description: Description of my project
HTTP base: /
HTTP path: /...
```
 
New:
```
$ spin new
Pick a template to start your application with: http-rust (HTTP request handler using Rust)
Enter a name for your new application: My application
Application description: Description of my application
HTTP base: /
HTTP path: /...
```

## `spin add`

Old:
```
$ spin add
Pick a template to start your project with: http-rust (HTTP request handler using Rust)
Enter a name for your new project: Second component
Project description: My new component
HTTP path: /...
```

The "project" terminology used in `spin add` is extra confusing because you're not even creating a new application, you're just adding to it. I thought this command was broken at first when it asked me for the project name. Ideally we use component terminology here.

New:
```
$ spin add
Pick a template to start your new component with: http-rust (HTTP request handler using Rust)
Enter a name for your new component: Second component
Component description: My new component
HTTP path: /...
```

## Potential issues

One potential issue with this change I wanted to highlight is that it appears that `project-description` is baked into our templates. If we wanted to rename that key too we might have to update all of the templates. I'm not fully up to speed on the implications of that.

calebschoepp:
## Changes
- Make prompts for name specific to `spin new` and `spin add` and user "application/component" rather than "project"
- Make prompts for description `spin new` and `spin add` agnostic.

## Example

```
$ spin target/release/spin new
Pick a template to start your application with: http-rust (HTTP request handler using Rust)
Enter a name for your new application: asdf
Description:
HTTP base: /
HTTP path: /...
$ ../target/release/spin add
Pick a template to start your component with: http-rust (HTTP request handler using Rust)
Enter a name for your new component: new
Description: asdf
HTTP path: /...
```

## TODO
- Update template in js sdk repo
- File issue for internal template names.

Closes #1074

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -116,7 +116,7 @@ impl TemplateNewCommandCore {
                     return Ok(());
                 }
             },
-            None => match prompt_template(&template_manager).await? {
+            None => match prompt_template(&template_manager, &variant).await? {
                 Some(template) => template,
                 None => return Ok(()),
             },
@@ -133,7 +133,7 @@ impl TemplateNewCommandCore {
 
         let name = match &self.name {
             Some(name) => name.to_owned(),
-            None => prompt_name().await?,
+            None => prompt_name(&variant).await?,
         };
 
         let output_path = self.output_path.clone().unwrap_or_else(|| path_safe(&name));
@@ -201,7 +201,10 @@ fn merge_values(from_file: &mut HashMap<String, String>, from_cli: &[ParameterVa
     }
 }
 
-async fn prompt_template(template_manager: &TemplateManager) -> anyhow::Result<Option<Template>> {
+async fn prompt_template(
+    template_manager: &TemplateManager,
+    variant: &TemplateVariantInfo,
+) -> anyhow::Result<Option<Template>> {
     let mut templates = match list_or_install_templates(template_manager).await? {
         Some(t) => t,
         None => return Ok(None),
@@ -210,8 +213,10 @@ async fn prompt_template(template_manager: &TemplateManager) -> anyhow::Result<O
         .iter()
         .map(|t| format!("{} ({})", t.id(), t.description_or_empty()))
         .collect::<Vec<_>>();
+    let noun = variant.prompt_noun();
+    let prompt = format!("Pick a template to start your {noun} with");
     let index = match dialoguer::Select::new()
-        .with_prompt("Pick a template to start your project with")
+        .with_prompt(prompt)
         .items(&opts)
         .default(0)
         .interact_opt()?
@@ -234,14 +239,15 @@ async fn list_or_install_templates(
     }
 }
 
-async fn prompt_name() -> anyhow::Result<String> {
-    let mut prompt = "Enter a name for your new project";
+async fn prompt_name(variant: &TemplateVariantInfo) -> anyhow::Result<String> {
+    let noun = variant.prompt_noun();
+    let mut prompt = format!("Enter a name for your new {noun}");
     loop {
         let result = dialoguer::Input::<String>::new()
             .with_prompt(prompt)
             .interact_text()?;
         if result.trim().is_empty() {
-            prompt = "Name is required. Try another project name (or Ctrl+C to exit)";
+            prompt = format!("Name is required. Try another {noun} name (or Crl+C to exit)");
             continue;
         } else {
             return Ok(result);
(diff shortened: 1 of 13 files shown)


itowlson:
Oh!  The prompts for the name are controlled by Spin code, not by the template.  So we could relatively easily have different prompts for new and add if we can't find a common term.  (And, as I say, I'm not too concerned about the inward-facing name.)  And we could have the prompt in the templates be just `Description` without it being ambiguous.

calebschoepp:
@itowlson Wow good catch. Totally spaced and forgot about the problem with `spin add` needing different terminology too.

Yes let's do something like "Enter a name for your new application" for `spin new` and "Enter a name for your new component" for `spin add`.

Then we can just do "Description" for both add and new like you said.

RE: Inward facing names, I can just remove those changes. My gut says it would be helpful to have ubiquitous language and keep the inward facing names the same. But, I trust your judgement and we can punt on this for now. I also like the `spin-*` prefix for magic variables as future work.

itowlson:
Yeah `TemplateNewCommandCore::run()` could pass the `variant` to `prompt_name` (https://github.com/fermyon/spin/blob/b3b1692c2e1816cd0f2f5fa8659c9c7e4df1aa47/src/commands/new.rs#L136) and it could use that to choose between "application:" and "component", that would solve it.

templates/static-fileserver/metadata/spin-template.toml
```diff
@@ -5,12 +5,12 @@ trigger_type = "http"
...
 [add_component]
 skip_files = ["spin.toml"]
-skip_parameters = ["http-base", "project-description"]
+skip_parameters = ["http-base", "application-description"]
 [add_component.snippets]
 component = "component.txt"
 
 [parameters]
-project-description = { type = "string",  prompt = "Project description", default = "" }
+application-description = { type = "string",  prompt = "Application description", default = "" }
```
michelleN, line 13: Now that I'm looking at this, I agree with Ivan's suggestion for letting the prompt be just "Description". What do you think @calebschoepp ?

calebschoepp: Yep think that's great. Changing it right now.

itowlson:
I totally hear what you're saying about ubiquitous language - unfortunately unless we can find a ubiquitous term that may be a doomed wish - `application` risks implying specifically the `new` case.  But we could change it to `spin-item-name` or `spin-template-instance-name` or something.

calebschoepp:
> I totally hear what you're saying about ubiquitous language - unfortunately unless we can find a ubiquitous term that may be a doomed wish - `application` risks implying specifically the `new` case. But we could change it to `spin-item-name` or `spin-template-instance-name` or something.

Yep agreed that it will be hard to find the term. Is this worth a new issue talking about `spin-*` prefix and a potentially ubiquitous name?

itowlson:
Hmm, yeah, as that would be setting a precedent for a convention (albeit one we are unlikely to ever extend!), it might be better to have a broader issue to gather consensus on that - good plan.

Apologies for the back-and-forth but hopefully we can at least nail the UI here which is definitely the main consideration!

calebschoepp:
> Hmm, yeah, as that would be setting a precedent for a convention (albeit one we are unlikely to ever extend!), it might be better to have a broader issue to gather consensus on that - good plan.
> 
> Apologies for the back-and-forth but hopefully we can at least nail the UI here which is definitely the main consideration!

Don't worry about it. This back-and-forth has been very helpful for me. I'll open an issue for the above.

src/commands/new.rs
```diff
@@ -201,7 +201,10 @@ fn merge_values(from_file: &mut HashMap<String, String>, from_cli: &[ParameterVa
     }
 }
 
-async fn prompt_template(template_manager: &TemplateManager) -> anyhow::Result<Option<Template>> {
+async fn prompt_template(
```
itowlson, line 204: Ooh good catch, I missed this one - thanks!

itowlson (approved):
LGTM, thanks for the patient discussion and for the attention to detail on all the prompts!

lann (approved):
IMO this is _just_ on the edge of "(not) worth it" for the amount of deduplication, so feel free to ignore if you don't like it:

```rust
impl TemplateVariantInfo {
    pub fn prompt_noun(&self) -> &'static str {
        match self {
            Self::NewApplication => "application",
            Self::AddComponent { .. } => "component",
        }
    }
}

...
let noun = variant.prompt_noun();
let prompt = format!("Pick a template to start your {noun} with");
```

:shrug:

itowlson:
Agree re marginal on deduplication, but it might be nice to have it next to `TemplateVariantInfo::description` so the two are more likely to stay in sync next time we change something _grin_

itowlson:
@calebschoepp to be clear, I also do not feel strongly either way - I am happy for you to merge without the change if you disagree / don't think it's worth it

calebschoepp:
Borderline, but who doesn't like a little deduplication. Change implemented and I'm going to merge.

michelleN (approved):

Issue #1098, opened 2023-02-01
Change HTTP routing algorithm to "longest matching prefix wins"


lann:
Currently, the matching of routes is dependent on the ordering of component configs in `spin.toml`. I would like to remove this ordering dependency and instead use the longest matching route, with "literal" routes taking precedence over "prefix" routes.

For these routes:
```
/
/...
/blog/...
/blog/new
/static/...
```
The following routings would apply:
```
/ -> /
/fun -> /...
/blog/my-post -> /blog/...
/blog/new -> /blog/new
/static/img.png -> /static/...
```

itowlson:
_abruptly stops work on unreachable route detection_

PR #1099, opened 2023-02-01
Warn if HTTP app contains duplicate routes

### LINKED ISSUE #341: Validate HTTP routes are unique
Spin currently permits non-unique HTTP routes, e.g. this loads without complaint:

```toml
[[component]]
id = "hello1"
[component.trigger]
route = "/hello"

[[component]]
id = "hello2"
[component.trigger]
route = "/hello"
```

This is overwhelmingly likely to be an error, and should be rejected, or at the very least should receive a trace message at an on-by-default level.

(~We should also document the behaviour of overlapping routes, e.g. `/...` and `/static/...` - I don't think we have that yet, though could be wrong!~  _[ETA: I was, indeed, wrong.]_  It would also be good to error or warn if a route can never be matched.)

itowlson:
Fixes #341.

This does not check for routes that are unreachable due to shadowing by routes appearing later in the manifest - that would be good to do if #1098 were not adopted, but depends on the outcome of that.

### DIFF
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -14,20 +14,37 @@ pub struct Router {
     pub(crate) routes: IndexMap<RoutePattern, String>,
 }
 
+pub(crate) struct DuplicateRoute {
+    pub route: RoutePattern,
+    pub replaced_id: String,
+    pub effective_id: String,
+}
+
 impl Router {
     /// Builds a router based on application configuration.
     pub(crate) fn build<'a>(
         base: &str,
         component_routes: impl IntoIterator<Item = (&'a str, &'a str)>,
-    ) -> Result<Self> {
-        let routes = component_routes
-            .into_iter()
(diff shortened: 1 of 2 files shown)


crates/http/src/routes.rs
```diff
@@ -14,20 +14,29 @@ pub struct Router {
     pub(crate) routes: IndexMap<RoutePattern, String>,
 }
 
+type DuplicateRoute = (RoutePattern, String, String);
```
fibonacci1729, line 17: Could you add a comment on how to interpret the two `String`s of this tuple? At glance it's not immediately obvious. Thanks!

itowlson: I knew I was shooting myself (and my readers) in the foot trying to get away with a tuple rather than a struct.  Thank you for keeping me honest!

fibonacci1729 (approved):
One minor nit, but LGTM!

PR #1104, opened 2023-02-03
Rename `spin oci` to `spin registry`

radu-matei:
As I have started using the OCI functionality a bit more, I realized that "OCI" is a fairly specialized three letter acronym, and expecting a user to know it and immediately understand "this has something to do with registries" might not be reasonable.

This commit changes the top level `spin oci` command to `spin registry`, while still keeping an alias for `spin oci`.

This should be more intuitive for a user.
Happy to close this PR if people think `spin oci` is more specific.

### DIFF
--- a/src/bin/spin.rs
+++ b/src/bin/spin.rs
@@ -9,8 +9,8 @@ use spin_cli::commands::{
     external::execute_external_subcommand,
     login::LoginCommand,
     new::{AddCommand, NewCommand},
-    oci::OciCommands,
     plugins::PluginCommands,
+    registry::RegistryCommands,
     templates::TemplateCommands,
     up::UpCommand,
 };
@@ -52,8 +52,8 @@ enum SpinApp {
     Up(UpCommand),
     #[clap(subcommand)]
     Bindle(BindleCommands),
-    #[clap(subcommand)]
-    Oci(OciCommands),
+    #[clap(subcommand, alias = "oci")]
+    Registry(RegistryCommands),
     Deploy(DeployCommand),
     Build(BuildCommand),
     Login(LoginCommand),
@@ -82,7 +82,7 @@ impl SpinApp {
             Self::New(cmd) => cmd.run().await,
             Self::Add(cmd) => cmd.run().await,
             Self::Bindle(cmd) => cmd.run().await,
-            Self::Oci(cmd) => cmd.run().await,
+            Self::Registry(cmd) => cmd.run().await,
             Self::Deploy(cmd) => cmd.run().await,
             Self::Build(cmd) => cmd.run().await,
             Self::Trigger(TriggerCommands::Http(cmd)) => cmd.run().await,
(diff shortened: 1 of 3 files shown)


fibonacci1729:
👍

fibonacci1729 (approved):

calebschoepp (approved):
LGTM 🚢

michelleN (approved):
YESSS

michelleN:
Do any docs need to be updated? What do we do about the SIP that talks about the OCI commands? We usually don't update SIPs but just wanted to bring it up.

itowlson:
Forgive me for being the voice of scepticism here... but... while I agree with the reservations about "OCI," 'registry' seems so open-ended that, as a reader, I don't really know what it means.  Is this the Warg registry?  My language's package registry? 
 The _twitch twitch_ Windows registry?  "OCI" has at least the benefit of telling me it's specifically an _OCI_ registry, giving me the context (and google term) to mentally link it to ACR, GHCR, Docker Hub, etc.

I'm not sure of a more helpful term though.  OCI registries tend to call themselves container registries, and `spin container-registry` is verbose and puts people in the wrong mental frame.  ("Does that put my application in a container?  I thought Wasm was meant to replace containers.")  "Artifact registry" is also verbose and is unlikely to be an immediate click to "oh they mean my ECR account."

Incidentally, if we find a term that we agree is better than 'oci' then I'm not sure it's worth bothering with the alias.  We don't have a significant backward compatibility obligation on `spin oci` at the moment - if we have a better idea then let's go for it.

radu-matei:
I definitely agree that making commands intuitive and explicit should be a primary goal; But I also think that you can't put all the context on the command name itself. You could make the same argument for `spin up`. Up where? Up what? You also can't name the command `spin prepare-the-wasm-modules-and-start-the-trigger-then-invoke-the-spin-entrypoint-when-an-event-happens-and-run-it-to-completion`.
(as a side note, the help text for `spin up` probably needs some attention)

While the command should be intuitive and representative of it does, I _think_ (and I'm making an assumption based on how I use tools) people look at the help text to see what commands are available, then at the specific subcommand help to dive into one — and `spin <insert-intuitive-command-for-distributing-spin-apps-with-container-registries>` should tell someone that it's used to distribute Spin applications using existing container registry services (and detail the fact it's using the OCI spec and specific registry services they can use).

As the person who initially named it, I think `spin oci` is the wrong name. It uses the name of a specification even people who have been using container registries for years probably don't know about (and most likely shouldn't know about).
(Also, you are not using a specification, you are using a _registry_ that is based on that specification, so "log in to OCI" doesn't really make a lot of sense.)

itowlson:
Yep, agree that `oci` is the wrong name.  I wondered about `remote`.  But if `registry` is the people's choice then I'm happy to go for it - the mental connection to the R in ACR/ECR/GHCR should click once people expand the acronyms.

PR #1106, opened 2023-02-03
chore(supply-chain): update audits/exemptions

vdice:
Update audits and exemptions per [failing audit check](https://github.com/fermyon/spin/actions/runs/4079876899/jobs/7031723390) (also seen in https://github.com/fermyon/spin/pull/1107).

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -6,6 +6,12 @@ who = "Radu Matei <radu.matei@fermyon.com>"
 criteria = "safe-to-deploy"
 version = "1.0.65"
 
+[[audits.anyhow]]
+who = "Vaughn Dice <vaughn.dice@fermyon.com>"
+criteria = "safe-to-deploy"
+delta = "1.0.65 -> 1.0.68"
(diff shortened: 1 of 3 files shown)


michelleN (commented):
@vdice Did you run a command to update `supply-chain/config.toml`. Just wondering what the process looked like for you? Looks like you did a manual check for anyhow 👍  and curious about the rest.

vdice:
@michelleN yes, I had to at least perform _one_ audit 😂 ; for this, as expected, I followed the steps mentioned in the [SIP](https://github.com/fermyon/spin/blob/main/docs/content/sips/009-auditing-third-party-dependencies.md#how-is-an-audit-performed).

For the exemptions, I started bumping by hand per the cargo vet failures but then the tedium of doing so motivated the creation of a bash script to parse the <dep>:<version> entries from the cargo vet output and run `cargo vet add-exemption <dep> <version>` on each.  Of course, simply bumping exemptions doesn't feed into the long-term goal of actually auditing dependencies, but while we shake out how we'd like our process to look like, it's a stop-gap.

lann (approved):

PR #1107, opened 2023-02-03
ci(workflows): add cargo-vet job to build.yml; rm cron-based audits.yml

vdice:
- Adds a cargo vet job to our build.yml
- Removes the cron-based audits.yml

Currently, the [cron-based approach](https://github.com/fermyon/spin/blob/main/.github/workflows/audits.yml) places the burden of tending to `cargo vet` failures to the person who last updated the workflow (happened to be me).  (When the job fails, a GH notification/email is sent to that user.)

While the burden was intended to be shared by the team, we'd need further tooling to notify the team at large in order for awareness of failures to be effective.  Then it's a case of depending on the kindness of a volunteer to address said failures.

Therefore, if we truly do want to get serious about maintaining good supply-chain hygiene, I'd think it would be more effective to add the check to the build.yml workflow which runs on PRs and main.

We still have a TBD on the iterative process of migrating exemptions out of config.toml and into audits.toml.  In fact, though we have the [SIP](https://github.com/fermyon/spin/blob/main/docs/content/sips/009-auditing-third-party-dependencies.md), we don't really have pragmatic steps that we'd expect the dev to perform when cargo vet fails on their changes.  Perhaps that's a TODO to accompany this change?  What would work best?  A doc to link to?

(Failing cargo vet check intended to be fixed in https://github.com/fermyon/spin/pull/1106)


radu-matei (changes requested):
I'm not sure this change is necessarily helpful.
Any pull request from a non maintainer would either be blocked on every dependency update, we would have to trust an audit from a non maintainer, or we would have to update their pull requests.

Neither of those scenarios ideal.
I'd rather move to run the audit weekly / bi-weekly for now, as we make our way though the audits.

vdice:
> Any pull request from a non maintainer would either be blocked on every dependency update, we would have to trust an audit from a non maintainer, or we would have to update their pull requests.

That's a fair point.  For non-maintainer contributions, we could allow additions/modifications to the exemptions list (as applicable to the PR), rather than expecting an audit.  Then, the burden would be on maintainers to follow-up with the dependencies in question.

> I'd rather move to run the audit weekly / bi-weekly for now, as we make our way though the audits.

The problem with spacing the runs out even further is the sheer mass of errors that accrue.  (See the [170 failures on ths PR](https://github.com/fermyon/spin/actions/runs/4088584396/jobs/7050415647).) With so many dependencies and the rapid cadence of releases for many of them, it's a daunting task for anyone to face -- even when the approach is to mostly just update/add exemptions (as done in https://github.com/fermyon/spin/pull/1106).

I can cede that perhaps we're not ready to run this on PRs (the idea was inspired by seeing how wasmtime does it) but I'd love to find an approach where the load is shared and cargo vet errors are addressed in maneagable chunks (say, ideally paired to a given changeset).

vdice:
Closing for now; can revisit when ready.

PR #1109, opened 2023-02-06
fix: update HTTP routing to longest matching prefix wins

fibonacci1729:
Today, HTTP routing in Spin is dependent upon the ordering of component configs in `spin.toml`. This PR evolves the HTTP routing algorithm to longest matching prefix wins where exact matches take highest precedence followed by longest matching wildcard patterns.

`test_router` is updated in place to reflect the new algorithm.

### DIFF
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -2,7 +2,7 @@
 
 #![deny(missing_docs)]
 
-use anyhow::{Context, Result};
+use anyhow::{anyhow, Result};
 use http::Uri;
 use indexmap::IndexMap;
 use std::{borrow::Cow, fmt};
@@ -38,11 +38,29 @@ impl Router {
     /// If there are multiple possible components registered for the same route or
     /// wildcard, this returns the last entry in the component map.
     pub(crate) fn route(&self, p: &str) -> Result<&str> {
-        self.routes
-            .iter()
-            .rfind(|(rp, _)| rp.matches(p))
-            .map(|(_, c)| c.as_ref())
-            .with_context(|| format!("Cannot match route for path {}", p))
+        let matches = self.routes.iter().filter(|(rp, _)| rp.matches(p));
+
+        let mut best_match: (Option<&str>, usize) = (None, 0); // matched id and pattern length
+
+        for (rp, id) in matches {
+            match rp {
+                RoutePattern::Exact(_m) => {
+                    // Exact matching routes take precedence over wildcard matches.
+                    return Ok(id);
+                }
+                RoutePattern::Wildcard(m) => {
+                    // Check and find longest matching prefix of wildcard pattern.
+                    let (_id_opt, len) = best_match;
+                    if m.len() >= len {
+                        best_match = (Some(id), m.len());
+                    }
+                }
+            }
+        }
+
+        best_match
+            .0
+            .ok_or_else(|| anyhow!("Cannot match route for path {p}"))
     }
 }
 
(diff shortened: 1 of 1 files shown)


fibonacci1729:
Closes [#1098](https://github.com/fermyon/spin/issues/1098)

lann (approved):

lann:
Docs should be updated to match

crates/http/src/routes.rs
```diff
@@ -38,11 +38,29 @@ impl Router {
     /// If there are multiple possible components registered for the same route or
     /// wildcard, this returns the last entry in the component map.
     pub(crate) fn route(&self, p: &str) -> Result<&str> {
-        self.routes
-            .iter()
-            .rfind(|(rp, _)| rp.matches(p))
-            .map(|(_, c)| c.as_ref())
-            .with_context(|| format!("Cannot match route for path {}", p))
+        let matches = self.routes.iter().filter(|(rp, _)| rp.matches(p));
```
itowlson, line 41: This is probably a micro-optimisation but YOLO

If I understand the usage pattern here, this reiterates the _full_ list of routes on every request, whereas the previous version iterated only until it found a match.

Would be be worth sorting the list once, as we build it?  Then this function could go back to being "return the first match."

crates/http/src/routes.rs
```diff
@@ -314,7 +332,10 @@ mod route_tests {
 
         let r = Router { routes };
 
-        assert_eq!(r.route("/one/two/three/four")?, "one_wildcard".to_string());
+        assert_eq!(
```
itowlson, line 335: This covers the "longer beats shorter" case but we also need one for "exact beats wildcard."

itowlson (changes requested):
The request changes is specifically for the testing - the other comment is a question/suggestion/think out loud and is definitely not a blocker!

crates/http/src/routes.rs
fibonacci1729: Ah yeah, good catch!

crates/http/src/routes.rs
fibonacci1729: cc @lann because we talked about the re-use of this logic for `lhc`. Is pre-sorting lexicographically to stateful for where `lhc` is going and how it might implement routing?

lann: I think this should still be lazy; `matches` is an iterator.

lann: That said, I'm not opposed to sorting while building either.

itowlson: Ah, yes, good point.  I amend my comment to "we iterate the full list unless we hit an exact match" (which will indeed often happen and get us out early).  Thanks!

crates/http/src/routes.rs
```diff
@@ -38,11 +38,29 @@ impl Router {
...
+            match rp {
+                RoutePattern::Exact(_m) => {
+                    // Exact matching routes take precedence over wildcard matches.
+                    return Ok(id);
+                }
+                RoutePattern::Wildcard(m) => {
+                    // Check and find longest matching prefix of wildcard pattern.
+                    let (_id_opt, len) = best_match;
+                    if m.len() >= len {
+                        best_match = (Some(id), m.len());
```
itowlson, line 55: I glanced over this first time round, but just wanted to say I thought this was an insightful shortcut - I had been imagining having to parse slashes and count bits, but the "only look at things that match" strategy indeed means you only need to consider string length (maximum prefix).  Nice one!

crates/http/src/routes.rs
fibonacci1729: @itowlson updated with new test!

itowlson (approved):
Thanks!

PR #1110, opened 2023-02-06
ci(perf): Improve job parallelization and disable incremental Rust builds

tschneidereit:
This set of changes should hopefully make our CI runs take less time. More importantly, the first one, disabling incremental cargo builds, should also reduce disk usage enough to make Windows tests pass more reliably, and give us a bit more time before we have to [move things to another partition](https://github.com/actions/runner-images/issues/1341#issuecomment-667478747)  ...

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -78,6 +78,8 @@ jobs:
 
       - name: Cargo Build
         run: cargo build --workspace --release --all-targets --all-features ${{ matrix.config.extraArgs }}
+        env:
+          CARGO_INCREMENTAL: 0
 
       - name: "Archive executable artifact"
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


rajatjindal (approved):
Thanks Till (probably should wait for review from Danielle also)

tschneidereit:
While this seems like an improvement, it unfortunately doesn't always decrease the wallclock time of CI runs too much: `build-rust --> e2e-tests` is often the critical path, and that still exists.

We could parallelize the step of building the tests container with `build-rust`, and/or we can look into making `e2e-tests` only depend on the Linux build. Unfortunately GH Actions' `needs` parameter for jobs doesn't allow specifying only some matrix configurations though, so the latter would get somewhat complicated pretty quickly :/

Anyway, this seems like strictly an improvement, so I think landing it for now makes sense :)

endocrimes (approved):

lann:
I believe the `Swatinem/rust-cache` step does this for you.

PR #1112, opened 2023-02-06
remove assets test

rajatjindal:
remove assets test as this testcase has been migrated to new e2e tests

https://github.com/fermyon/spin/blob/b1685404e933fc080825674890a42a42ee363241/tests/testcases/mod.rs#L174

Signed-off-by: Rajat Jindal <rajatjindal83@gmail.com>


itowlson:
I feel there was some value in having "integration lite" tests with no dependencies (or at most Bindle) that could be run locally relatively quickly and easily, though I'm not sure I ever really ended up catching or diagnosing an issue with them, so eh, maybe not.  I guess I don't have strong feelings either way. _grin_

rajatjindal:
Hi Ivan

It is a fair point. Thank you for bringing it up.

If you are able to run tests using docker compose file (checked-in as part of repo already), will that be a reasonable workaround?

I still need to optimize the docker compose process for local laptop and also for arm, but it generally worked for me. I can also add this to Makefile if that makes it any easier.

What do u think

itowlson:
Ah, yep, if I can run specific tests (a la `cargo test my_test`) in the Docker harness then that should be grand.  Thanks!

vdice (approved):
Assuming @itowlson is 👍 ; LGTM

itowlson (approved):
ah yep sorry :shipit:

michelleN (approved):

rajatjindal:
> Ah, yep, if I can run specific tests (a la `cargo test my_test`) in the Docker harness then that should be grand. Thanks!

I just pushed a change to add entrypoint to docker compose file to demonstrate how to use (or change) that for running specific tests using docker compose.

PR #1115, opened 2023-02-07
ref(*): make log output to stdout by default

### LINKED ISSUE #928: Make log output to stdout default

michelleN:
+ adds --silent flag to silence logs if needed
+ resolves #928

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -65,11 +65,18 @@ where
     pub follow_components: Vec<String>,
 
     /// Print all component output to stdout/stderr
-    #[clap(
-        long = "follow-all",
-        conflicts_with = FOLLOW_LOG_OPT,
-    )]
+    /// Deprecated
+    #[clap(long = "follow-all")]
     pub follow_all_components: bool,
+
+    /// Silence all component output to stdout/stderr
+    #[clap(
+        long = "quiet",
+        short = 'q',
+        aliases = &["sh", "shush"],
+        conflicts_with = FOLLOW_LOG_OPT,
+        )]
+    pub silence_component_logs: bool,
 
     /// Set the static assets of the components in the temporary directory as writable.
     #[clap(long = "allow-transient-write")]
@@ -179,9 +186,12 @@ where
 
     pub fn follow_components(&self) -> FollowComponents {
         if self.follow_all_components {
-            FollowComponents::All
-        } else if self.follow_components.is_empty() {
+            eprintln!("NOTE: --follow-all has been deprecated. All component ouput is now printed to stdout by default.")
+        }
+        if self.silence_component_logs {
             FollowComponents::None
+        } else if self.follow_components.is_empty() {
+            FollowComponents::All
         } else {
             let followed = self.follow_components.clone().into_iter().collect();
             FollowComponents::Named(followed)
--- a/examples/spin-timer/readme.md
+++ b/examples/spin-timer/readme.md
@@ -10,4 +10,4 @@ To test:
   * Update the URL too, to reflect the directory where the tar file is
 * `spin plugin install --file ./trigger-timer.json --yes`
 
-Then you should be able to `spin build --up --follow-all` the guest.
+Then you should be able to `spin build --up` the guest.
--- a/src/commands/registry.rs
+++ b/src/commands/registry.rs
@@ -121,7 +121,7 @@ pub struct Run {
     pub reference: String,
 
     /// All other args, to be passed through to the trigger
-    ///  TODO: The arguments have to be passed like `-- --follow-all` for now.
+    ///  TODO: The arguments have to be passed like `-- --quiet` for now.
     #[clap(hide = true)]
     pub trigger_args: Vec<OsString>,
 }
@@ -156,8 +156,8 @@ impl Run {
             .arg(trigger_type)
             // TODO: This should be inferred from the lockfile.
             .arg("--oci")
-            // TODO: Once we figure out how to handle the flags for triggers, i.e. `-- --follow-all`, remove this.
-            .arg("--follow-all")
+            // TODO: Once we figure out how to handle the flags for triggers, i.e. `-- --quiet`, remove this.
+            .arg("--quiet")
             .args(&self.trigger_args)
             .env(SPIN_WORKING_DIR, working_dir.path());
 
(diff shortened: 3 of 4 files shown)


examples/spin-timer/readme.md
```diff
@@ -2,12 +2,12 @@
...
 To test:
 
-* `cargo build --release`
-* Copy: `cp ./target/release/trigger-timer .`
-* `tar czvf trigger-timer.tar.gz trigger-timer`
-* Update the plugin manifest (`trigger-timer.json`):
-  * Get the SHA: `shasum -a 256 trigger-timer.tar.gz` and copy it into the `sha256` field
-  * Update the URL too, to reflect the directory where the tar file is
-* `spin plugin install --file ./trigger-timer.json --yes`
+- `cargo build --release`
```
itowlson, line 5: Is there a reason to change these?  (I don't have an opinion about the change - just surprised to see it in this PR.)

crates/trigger/src/cli.rs
```diff
@@ -64,12 +64,12 @@ where
...
 
-    /// Print all component output to stdout/stderr
+    /// Silence all component output to stdout/stderr
     #[clap(
-        long = "follow-all",
+        long = "silent",
         conflicts_with = FOLLOW_LOG_OPT,
         )]
-    pub follow_all_components: bool,
+    pub silence_component_logs: bool,
```
itowlson, line 72: I wonder if we are at the stage where `--follow-all` is muscle memory enough for people that we should continue to allow the flag as a "no op and deprecation warning."  Not a blocker, just raising it as a thought.

crates/trigger/src/cli.rs
```diff
@@ -64,12 +64,12 @@ where
         )]
     pub follow_components: Vec<String>,
 
-    /// Print all component output to stdout/stderr
+    /// Silence all component output to stdout/stderr
     #[clap(
-        long = "follow-all",
+        long = "silent",
```
itowlson, line 69: I'm a bit unsure about the name here - this promises _silence_, but users will still see Spin output such as "Available routes" links, and Spin logs per the Rust environment variable (errors by default).  I don't have any more constructive suggestions, though, sorry.

itowlson (approved):
I'd like to bikeshed the "suppress component logs" option name before we merge, but if we don't come up with better ideas then LGTM.  Thank you for cutting through this blasted Gordian knot.

examples/spin-timer/readme.md
michelleN: updated. good catch. think that was some linter changing things for me.

crates/trigger/src/cli.rs
michelleN: `--silence-component-logs` ?

michelleN: with a short flag `-sh` like 🤫  ??

itowlson: `-sh` got a giggle out of me!

The long form might be _too_ long, though, and is a bit disconnected from the `follow` terminology used for following single components.  I am wondering about `follow-none` but I can't make myself love it.

itowlson: (for the avoidance of doubt, "got a giggle out of me" was intended to convey that I really liked it)

michelleN: How do you feel about `--silence-components`?

michelleN: `no-app-logs`?
`no-component-logs`?

I'd really love to use `sh`

itowlson: In isolation I'd probably go for `silence-components`.  But `no-{app|component}-logs` has the advantage that we could rename `follow` to `{app|component}-log` so you could have the related names we had with the `follow` family.

Let's wake @mikkelhegn up and make him have an opinion...

crates/trigger/src/cli.rs
michelleN: Good thought. Wonder if this will clutter peoples logs.

crates/trigger/src/cli.rs
radu-matei: Ok, hear me out: proposal I'm not feeling strongly about, so feel free to ignore.

The default (`spin up` with no flag outputs Spin logs + component logs to stdout / stderr.

`--log-files` / `--log-dir` / `--component-logs-<file-or-dir?>` sends the component logs to files / directory, and the Spin output (available routes, Spin errors) would still go to Spin's stdout / stderr.

Would something like this solve our issue?

crates/trigger/src/cli.rs
michelleN: Added a deprecation notice. Please lmk what you think.

crates/trigger/src/cli.rs
michelleN: I think there is a world where someone may want to see the logs in the output AND save logs to a directory.

itowlson: If I've understood that right, the way to turn off component log output would be to specify `--log-dir`.  For me that is extremely non-obvious, sorry.  It also leaves us with the "follow only one" problem.

crates/trigger/src/cli.rs
itowlson: Looks good - thanks!

crates/trigger/src/cli.rs
```diff
@@ -179,9 +183,12 @@ where
 
     pub fn follow_components(&self) -> FollowComponents {
         if self.follow_all_components {
-            FollowComponents::All
-        } else if self.follow_components.is_empty() {
+            println!("NOTE: --follow-all has been deprecated. All component ouput is now printed to stdout by default.")
```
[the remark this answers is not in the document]
michelleN: updated

crates/trigger/src/cli.rs
michelleN: Should we remove the deprecation notice before cutting 1.0?

itowlson: Yes, that would be a 4-6 week timeframe, long enough for people who use it routinely enough to have muscle memory to adjust.

crates/trigger/src/cli.rs
lann: `--quiet`/`-q` (no component logs) and `--quieter`/`-qq` (no spin logs either)

itowlson (approved):

Issue #919, opened 2022-11-27
Improve template discoverability and categorisation


itowlson:
From @mikkelhegn (https://github.com/fermyon/spin/issues/629#issuecomment-1326572845):

> What I believe we should achieve with this is when a person, who is educating themselves by exploring Spin using the CLI, will be able to easily identify what a given Spin template gives them in terms of functionality and availability in programming languages. If I'm a developer who prefers writing code in Go, I can easily identify the templates (using `spin new`, `spin add` and `spin templates list`, that are available for Go, sorted by their key features (e.g., type of trigger). I am also able to find a reference to a document (a link), where I can find more information about the specific template (e.g., if it's a pre-built component, a link to the components GitHub repo would be a good example. If it's an http-trigger component, a link to the description of http-triggers would be a good option).

Starting a new issue for this so we can talk about this independently of the "application build dependencies" issue.

itowlson:
A detail to consider here (as with the parent issue) is "day one vs day two."

Day one is component creation time.  A user walks up to Spin and goes "okay I want to make a Web application in Go" or "okay my application needs to serve CSS files."  This is where templates come in, and where the template UI can help via filtering and categorisation.

Day two is when the template output has scrolled off the screen.  A user walks up to an _existing_ application and goes "how the heck do I get the full URL" or "how the heck do I do set operations."  This is where usage links should _remain_ readily discoverable, rather than relying on the user to work back to the template.  (This is not to say they shouldn't be presented along with the other template discoverability stuff - just that they should not be part of the template UI _only_.)

itowlson:
Regarding template discoverability, one idea that I've been mulling is tags.

A template could be annotated with any number of tags, e.g. trigger type, language, use case.  These need not be particularly structured, though trigger and language could be.  Then the user could type `spin templates list --tag http --tag go` and see only templates with those tags.

We could go even further and allow the user to write `spin templates list http` and it would filter to tag matches (and, I'd suggest, name matches).  However, this wouldn't work so well for `spin new` and `spin add`, because in `spin new http go` you couldn't tell whether the second argument was an additional filter or a project name.

itowlson:
Distinguishing a language (rather than having it as just another tag) would open the way for users to set a preferred language via an environment variable, so they only ever needed to see templates compatible with their preferred language.

(Though more generally one could imagine the environment variable as 'filter' rather than specifically 'language.')

mikkelhegn:
[...]

PR #1119, opened 2023-02-07
feat: implement basic template tags

fibonacci1729:
This PR implements "template tags" to [improve discoverability](https://github.com/fermyon/spin/issues/919).

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -84,6 +84,16 @@ pub struct ListResults {
     pub templates: Vec<Template>,
     /// Any warnings identified during the list operation.
     pub warnings: Vec<(String, InstalledTemplateWarning)>,
+    /// Any skipped templates (populated as a result of filtering by tags).
+    pub skipped: Vec<Template>,
+}
+
+impl ListResults {
+    /// Returns true if no templates were found or skipped indicating that
+    /// templates may not be installed.
+    pub fn needs_install(&self) -> bool {
+        self.templates.is_empty() && self.skipped.is_empty()
+    }
 }
 
 /// A recoverable problem while listing templates.
@@ -220,6 +230,26 @@ impl TemplateManager {
         Ok(ListResults {
             templates,
             warnings,
+            skipped: vec![],
+        })
+    }
+
+    /// Lists all installed templates that match all the provided tags.
+    pub async fn list_with_tags(&self, tags: &[String]) -> anyhow::Result<ListResults> {
+        let ListResults {
+            templates,
+            warnings,
+            ..
+        } = self.list().await?;
+
+        let (templates, skipped) = templates
+            .into_iter()
+            .partition(|tpl| tpl.matches_all_tags(tags));
+
+        Ok(ListResults {
+            templates,
+            warnings,
+            skipped,
         })
     }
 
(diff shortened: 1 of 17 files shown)


templates/http-c/metadata/spin-template.toml
```diff
@@ -1,6 +1,7 @@
 manifest_version = "1"
 id = "http-c"
 description = "HTTP request handler using C and the Zig toolchain"
+tags = ["http", "c"]
```
fibonacci1729, line 4: @itowlson happy to change the tags for all these templates if you are keen on a particular set.

itowlson: Nah, protocol and language seems like the perfect starter set.

src/commands/templates.rs
```diff
@@ -468,15 +475,23 @@ impl List {
...
+
         let json_text = serde_json::to_string_pretty(&json_vals)?;
         println!("{}", json_text);
         Ok(())
     }
+
+    fn filter_templates_with_tags<'i>(&'i self, templates: &'i [Template]) -> Vec<&'i Template> {
+        templates
+            .iter()
+            .filter(|tpl| self.tags.iter().all(|tag| tpl.tags().contains(tag)))
```
itowlson, line 492: I found the nested expression with 4 occurrences of the word `tag` in it a bit hard to follow.  I wonder if rather than returning the `tags` collection, `Template` should have a `has_all_tags` or `matches_all_tags` method so this could be `filter(|tpl| tpl.has_all_tags(&self.tags))`?

src/commands/templates.rs
```diff
@@ -390,6 +390,10 @@ impl Uninstall {
 /// List the installed templates.
 #[derive(Parser, Debug)]
 pub struct List {
+    /// Filter templates with associated tags.
```
itowlson, line 393: May be worth saying if this filters to templates with _all_ given tags or _any_ given tag.  (I think it's the former and I think that's the right decision... just suggesting docs.)

src/commands/templates.rs
itowlson, line 492: This looks like it's case-sensitive; might be good to be more forgiving.

itowlson (commented):
I realise this is a draft and I hope these comments are not premature - please feel free to ignore them if you are still changing things.

As you'll see, I am having second thoughts about what we chatted about yesterday, but only in the most milquetoast of ways - I'm still happy for you to go ahead doing it this way, just wanted to capture my thoughts in a comment.

src/commands/templates.rs
```diff
@@ -441,7 +448,7 @@ impl List {
             table.set_header(header);
             table.load_preset(comfy_table::presets::ASCII_BORDERS_ONLY_CONDENSED);
 
-            for template in templates {
+            for template in self.filter_templates_with_tags(templates) {
```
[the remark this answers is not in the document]
fibonacci1729: @itowlson 

w.r.t your first comment: Great catch. I think i can maneuver some of this code to avoid the empty table headers being printed as well to include a more informative "no templates matched with tags" message as you suggest. Thanks!

w.r.t your second comment: These are all great thoughts and absolutely appreciated. I had originally started to implement `TemplatesManager::list_with_tags` that does exactly as you suggest (with updated list results to reflect the behavior you mention). Happy to go that approach as its not too much more involved than what is already here.

src/commands/templates.rs
fibonacci1729: I was hoping this would you'd say that. Will factor out some helpers to prettify this combinator.

src/commands/templates.rs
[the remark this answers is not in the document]
fibonacci1729: Updated to address some of the UI concerns.

# With no templates installed 
```
❯ spin templates list --tag http
You don't have any templates yet. Would you like to install the default set? [Y/n]
```

# With default set installed but no matching tag
```
❯ spin templates list --tag foo
No templates matched tags (12 templates skipped)
```

# The Spin new experience with tags
```
❯ spin new --tag http
Pick a template to start your application with:
> http-c (HTTP request handler using C and the Zig toolchain)
  http-empty (HTTP application with no components)
  http-go (HTTP request handler using (Tiny)Go)
  http-grain (HTTP request handler using Grain)
  http-php (HTTP request handler using PHP)
  http-rust (HTTP request handler using Rust)
  http-swift (HTTP request handler using SwiftWasm)
  http-zig (HTTP request handler using Zig)
  static-fileserver (Serves static files from an asset directory)
```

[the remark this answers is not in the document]
itowlson: BRIAN YOU DID IT

I was not expecting that as part of this PR - very cool!

src/commands/templates.rs
```diff
@@ -409,27 +413,37 @@ impl List {
...
-            .list()
+            .list_with_tags(&self.tags)
             .await
             .context("Failed to list templates")?;
 
-        match (&self.format, list_results.templates.is_empty()) {
-            (ListFormat::Table, false) => self.print_templates_table(&list_results),
-            (ListFormat::Table, true) => {
+        match self.format {
+            ListFormat::Table if list_results.needs_install() => {
```
itowlson, line 423: I like this, much more intention-revealing.

itowlson (changes requested):
I'm delighted with this - thank you!

But... and I am really sorry about this... I think the new filtering stuff does need some unit-ish tests.  But the actual code is really nice!  I love that you got support into new/add, and it's all super clear to read.

crates/templates/src/manager.rs
```diff
@@ -487,6 +517,97 @@ mod tests {
...
+        let store = TemplateStore::new(temp_dir.path());
+        let manager = TemplateManager { store };
+        let source = TemplateSource::File(project_root());
+
+        manager
+            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .await
+            .unwrap();
+
+        let tags_to_match = vec!["c".to_string(), "".to_string()];
```
itowlson, line 549: I'd suggest something like `"unused_tag"` or `"nonexistent_tag_1234"` - to me that feels more explanatory than the empty string?

itowlson (approved):
Tests look good - thanks for adding them.  I left one stylistic niggle but it's not a big deal - good to go!  :shipit:

PR #1130, opened 2023-02-09
ref(build): updates build to allow interactive commands

### LINKED ISSUE #1026: `spin build` shouldn't capture stdio
Some build tools like `npm` require user interaction. `spin build` currently captures `stdout`, making it effectively impossible to use interactive tools.

michelleN:
Previously, spin build would hang for commands that were interactive (like a login command). This was a result of how we were setting up standard streams for the subprocess created to run the build command. This change updates how we set up standard streams such that the subprocess inherits standard streams from the parent which  enables interactive commands.

resolves #1026

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -4,11 +4,10 @@
 
 mod manifest;
 
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use spin_loader::local::parent_dir;
 use std::path::{Path, PathBuf};
 use subprocess::{Exec, Redirection};
-use tracing::log;
 
 use crate::manifest::{BuildAppInfoAnyVersion, RawComponentManifest};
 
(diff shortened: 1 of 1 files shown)


crates/build/src/lib.rs
```diff
@@ -53,27 +46,27 @@ fn build_component(raw: RawComponentManifest, app_dir: &Path) -> Result<()> {
...
+            let exit_status = Exec::shell(&b.command)
                 .cwd(workdir)
-                .stdout(Redirection::Pipe)
-                .capture()
-                .with_context(|| {
-                    format!(
-                        "Cannot spawn build process '{:?}' for component {}.",
-                        &b.command, raw.id
+                .stdout(Redirection::None)
+                .stderr(Redirection::Merge)
```
itowlson, line 52: Worth documenting why this is Merge rather than None?  (I'm secretly curious about it myself - I would have thought None would let consumers capture the streams separately, but I assume it causes some problem?)

michelleN: `Using Redirection::Merge for PopenConfig::stderr requests the child’s stderr to refer to the same underlying file as the child’s stdout (which may or may not itself be redirected),`

itowlson: Sorry, I wasn't clear - I know what it does, I was curious _why_ we wanted it to do that.

michelleN: yea you're right and I think I should change it to None.

michelleN: updated

itowlson (approved):

fibonacci1729 (approved):

PR #1134, opened 2023-02-13
add general-purpose `execute` function to `outbound-redis`

dicej:
By popular request, this function should allow executing any arbitrary Redis command and retrieving its result, if any.  Ideally, we'd support a statically typed interface for every command, but that would be a pretty huge effort, so we settle for dynamic typing for now. Once we have TCP socket support, users will be able to use their favorite Redis client library and enjoy static typing at that level, if applicable.

### DIFF
--- a/crates/abi-conformance/src/outbound_redis.rs
+++ b/crates/abi-conformance/src/outbound_redis.rs
@@ -1,6 +1,6 @@
 use super::Context;
 use anyhow::{ensure, Result};
-use outbound_redis::{Error, ValueParam, ValueResult};
+use outbound_redis::{Error, RedisParameter, RedisResult};
 use serde::Serialize;
 use std::collections::{HashMap, HashSet};
 use wasmtime::{InstancePre, Store};
@@ -100,7 +100,7 @@ pub(super) struct OutboundRedis {
     sadd_map: HashMap<(String, String, Vec<String>), i64>,
     srem_map: HashMap<(String, String, Vec<String>), i64>,
     smembers_map: HashMap<(String, String), Vec<String>>,
-    execute_map: HashMap<(String, String, Vec<String>), Vec<ValueResult>>,
+    execute_map: HashMap<(String, String, Vec<String>), Vec<RedisResult>>,
 }
 
 impl outbound_redis::OutboundRedis for OutboundRedis {
@@ -178,8 +178,8 @@ impl outbound_redis::OutboundRedis for OutboundRedis {
         &mut self,
         address: &str,
         command: &str,
-        arguments: Vec<ValueParam<'_>>,
-    ) -> Result<Vec<ValueResult>, Error> {
+        arguments: Vec<RedisParameter<'_>>,
+    ) -> Result<Vec<RedisResult>, Error> {
         self.execute_map
             .remove(&(
                 address.into(),
@@ -379,7 +379,7 @@ pub(super) fn test(store: &mut Store<Context>, pre: &InstancePre<Context>) -> Re
                     "append".to_owned(),
                     vec!["foo".to_owned(), "baz".to_owned()],
                 ),
-                vec![ValueResult::Int(3)],
+                vec![RedisResult::Int64(3)],
             );
 
             super::run_command(
--- a/crates/abi-conformance/src/outbound_redis.rs
+++ b/crates/abi-conformance/src/outbound_redis.rs
@@ -1,5 +1,6 @@
 use super::Context;
 use anyhow::{ensure, Result};
+use outbound_redis::{Error, ValueParam, ValueResult};
 use serde::Serialize;
 use std::collections::{HashMap, HashSet};
 use wasmtime::{InstancePre, Store};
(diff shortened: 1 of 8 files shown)


dicej:
FYI, the CI failure is due to the dreaded "no space left on device" issue running the E2E tests on Windows.

wit/ephemeral/redis-types.wit
```diff
@@ -6,3 +6,11 @@ enum error {
 
 // The message payload.
 type payload = list<u8>
+
+// An argument and return type for the general-purpose `execute` function.
+variant value {
+    nil,
+    %string(string),
```
itowlson, line 13: We're of course under no obligation to be consistent, and it may be more appropriate to use Redis type names for a product-specific instance like this, but the database WIT uses the names `str`, `int64` and `binary` for the variant cases.  Obviously your call as to whether alignment is appropriate or not.

tests/outbound-redis/http-rust-outbound-redis/src/lib.rs
```diff
@@ -28,10 +30,99 @@ fn test(_req: Request) -> Result<Response> {
...
+
+    redis::execute(
+        &address,
+        "append",
+        &[
+            ValueParam::String("spin-example"),
+            ValueParam::Data(b" I've got it!"),
+        ],
+    )
+    .map_err(|_| anyhow!("Error executing Redis set command"))?;
```
itowlson, line 56:
```suggestion
    .map_err(|_| anyhow!("Error executing Redis append command"))?;
```

tests/outbound-redis/http-rust-outbound-redis/src/lib.rs
```diff
@@ -28,10 +30,99 @@ fn test(_req: Request) -> Result<Response> {
...
+
+    let values = redis::execute(&address, "get", &[ValueParam::String("spin-example")])
+        .map_err(|_| anyhow!("Error executing Redis get command"))?;
```
itowlson, line 59: As there is already a named `get` command, would be be useful in this message to specify that the error was in executing it _via_ the `execute` command?

radu-matei:
Really love to see the addition of this, thanks for taking it on.

Do we need a checklist for keeping track of other changes outside of this repo? Specifically thinking about other SDKs.

crates/outbound-redis/src/lib.rs
```diff
@@ -3,13 +3,33 @@ mod host_component;
...
+
+impl FromRedisValue for Values {
+    fn from_redis_value(value: &Value) -> RedisResult<Self> {
+        fn append(values: &mut Vec<ValueResult>, value: &Value) {
+            match value {
+                Value::Nil | Value::Okay => (),
+                Value::Int(v) => values.push(ValueResult::Int(*v)),
+                Value::Data(bytes) => values.push(ValueResult::Data(bytes.to_owned())),
+                Value::Bulk(bulk) => bulk.iter().for_each(|value| append(values, value)),
+                Value::Status(message) => values.push(ValueResult::String(message.to_owned())),
```
radu-matei, line 24: If I understand this correctly, it will only return a `ValueResult::String` when a status is returned? I used list operations that return strings (`lpush` and `lrange`), and was surprised that the results were always `ValueResult::Data` that needed to be decoded. The only ask here is that we document what operation is expected to return a string.

crates/abi-conformance/src/outbound_redis.rs
```diff
@@ -76,6 +77,15 @@ pub struct RedisReport {
     /// "baz"))` as the result.  The host will assert that said function is called exactly once with the specified
     /// arguments.
     pub smembers: Result<(), String>,
+
```
lann, line 80: As a total aside, it would be interesting to run these conformance tests under tarpaulin.

crates/outbound-redis/src/lib.rs
```diff
@@ -65,6 +85,33 @@ impl outbound_redis::OutboundRedis for OutboundRedis {
...
+    async fn execute(
+        &mut self,
+        address: &str,
+        command: &str,
+        arguments: Vec<ValueParam<'_>>,
+    ) -> Result<Vec<ValueResult>, Error> {
+        let conn = self.get_conn(address).await.map_err(log_error)?;
+        let mut cmd = redis::cmd(command);
+        arguments.iter().for_each(|value| match value {
+            ValueParam::Nil => (),
```
lann, line 98: This seems very odd; should it return an error instead?

wit/ephemeral/redis-types.wit
lann, line 13: Should this be `%string(list<u8>)`? I don't think Redis "strings" are guaranteed to be unicode.

lann: Unless this is representing error messages?

radu-matei: Also see my comment above.
My assumption is that it's only returning status messages and not actual Redis strings.

crates/outbound-redis/src/lib.rs
dicej: Yeah, I initially assumed more things would return strings and integers, but apparently Redis really likes returning byte arrays, leaving it up to the client to parse the expected type.  I agree that renaming it to `status` makes sense.

wit/ephemeral/redis-types.wit
```diff
@@ -7,10 +7,17 @@ enum error {
 // The message payload.
 type payload = list<u8>
 
-// An argument and return type for the general-purpose `execute` function.
-variant value {
+// A parameter type for the general-purpose `execute` function.
+variant redis-parameter {
+    str(string),
```
lann, line 12: Is this useful down at this level? Any difference with encoding strings into `binary`?

dicej: I figured it would be more ergonomic for many/most languages.  So yes, it's unnecessary, but still seems useful.

lann: I guess I've been looking at WIT as _mostly_ a low-level interface for libraries to wrap, though we haven't always done that...

lann: If we keep it, this should document that the host must utf-8 encode.

lann: I would keep the int type since it enforces redis's actual int range and has the separate wire encoding. Also presumably generated JS bindings will have to solve the JS number precision problem for `int64` _somehow_, so redis bindings over there wouldn't have to invent their own solution.

lann (approved):

radu-matei (approved):

PR #1139, opened 2023-02-14
move `deploy` and `login` to `spin cloud`

### LINKED ISSUE #1117: move cloud commands under spin cloud
Going forward, it would be nice to move cloud related commands under the spin cloud command.

$ spin login -> spin cloud login
$ spin deploy -> spin cloud deploy

For convenience, we should leave aliases for `spin login` and `spin deploy` but all other cloud related commands should be added under `spin cloud`.

bacongobbler:
closes #1117

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -4,11 +4,13 @@
 pub mod bindle;
 /// Commands for building Spin applications.
 pub mod build;
-/// Command for deploying a Spin app to Hippo
+/// Commands for publishing applications to the Fermyon Platform.
+pub mod cloud;
+/// Command to package and upload an application to the Fermyon Platform.
 pub mod deploy;
 /// Commands for external subcommands (i.e. plugins)
 pub mod external;
-// Command for logging into the server
+/// Command for logging into the Fermyon Platform.
 pub mod login;
 /// Command for creating a new application.
 pub mod new;
--- a/src/commands/cloud.rs
+++ b/src/commands/cloud.rs
@@ -1,0 +1,24 @@
+use anyhow::Result;
+use clap::Subcommand;
+
+use super::deploy::DeployCommand;
+use super::login::LoginCommand;
+
+/// Commands for publishing applications to the Fermyon Platform.
+#[derive(Subcommand, Debug)]
+pub enum CloudCommands {
+    /// Package and upload an application to the Fermyon Platform.
+    Deploy(DeployCommand),
+
+    /// Log into the Fermyon Platform.
+    Login(LoginCommand),
+}
+
+impl CloudCommands {
+    pub async fn run(self) -> Result<()> {
+        match self {
+            Self::Deploy(cmd) => cmd.run().await,
+            Self::Login(cmd) => cmd.run().await,
+        }
+    }
+}
--- a/src/bin/spin.rs
+++ b/src/bin/spin.rs
@@ -5,6 +5,7 @@ use lazy_static::lazy_static;
 use spin_cli::commands::{
     bindle::BindleCommands,
     build::BuildCommand,
+    cloud::CloudCommands,
     deploy::DeployCommand,
     external::execute_external_subcommand,
     login::LoginCommand,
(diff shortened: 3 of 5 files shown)


fibonacci1729 (approved):

src/commands.rs
```diff
@@ -4,12 +4,14 @@
 pub mod bindle;
 /// Commands for building Spin applications.
 pub mod build;
-/// Command for deploying a Spin app to Hippo
+/// Commands for building Spin applications.
```
rajatjindal, line 7: nit, we should fix this comment.

src/commands/cloud.rs
```diff
@@ -0,0 +1,26 @@
...
+use anyhow::Result;
+
+use super::deploy::DeployCommand;
+use super::login::LoginCommand;
+
+/// Commands for publishing applications to Fermyon Cloud.
+#[derive(Subcommand, Debug)]
+pub enum CloudCommands {
+    /// Package and upload to Fermyon Cloud
+    #[clap(alias = "deploy")]
```
radu-matei, line 11: The alias doesn't seem to work for me, as it seems to expect a plugin:

```
$ spin deploy       
Unknown command: plugin 'deploy' not found at expected location /Users/radu/Library/Application Support/spin/plugins/manifests/deploy.json: No such file or directory (os error 2)

$ spin login       
Unknown command: plugin 'login' not found at expected location /Users/radu/Library/Application Support/spin/plugins/manifests/login.json: No such file or directory (os error 2
```

bacongobbler: Yeah this was part of an earlier experimentation that does not work.

See https://github.com/fermyon/spin/issues/1117#issuecomment-1428908182 for more details.

src/commands/cloud.rs
```diff
@@ -0,0 +1,24 @@
+use anyhow::Result;
+use clap::Subcommand;
+
+use super::deploy::DeployCommand;
+use super::login::LoginCommand;
+
+/// Commands for publishing applications to Fermyon Cloud.
+#[derive(Subcommand, Debug)]
+pub enum CloudCommands {
+    /// Package and upload an application to Fermyon Cloud.
```
itowlson, line 10: Nit: despite the new naming, this works with the Platform as well - should we say e.g. "to Fermyon" (could be confusing for Fermyon the product vs Fermyon the company) or "to Fermyon Cloud or the Fermyon Platform"?

itowlson (approved):

src/commands.rs
```diff
@@ -4,11 +4,13 @@
 pub mod bindle;
 /// Commands for building Spin applications.
 pub mod build;
-/// Command for deploying a Spin app to Hippo
+/// Commands for publishing applications to the Fermyon Platform.
```
calebschoepp, line 7: Super nit: Throughout you keep referring to the "Fermyon Platform" and I just wanted to make sure that this is our preferred terminology vs "Fermyon Cloud Platform".

itowlson: There are two distinct things, Fermyon Cloud and Fermyon Platform.  These commands work with both.  A new "Fermyon Cloud Platform" term would only add to the confusion I fear!

bacongobbler: I believe the current verbiage is the correct convention.

My understanding is that we use "Fermyon Platform" to refer to the self-hosted as well as the managed service (Fermyon Cloud). We refer to both the self-hosted and the hosted option on [fermyon.dev](https://www.fermyon.dev/).

calebschoepp: Cool, thanks for the thorough response you two 😄 . Looks good.

michelleN (commented):
The command help for `spin deploy` and `spin cloud deploy` look different. I think we should make these the same?

```
$ spin deploy --help
spin-deploy
Deploy a Spin application

USAGE:
    spin deploy [OPTIONS]
```

```
$ spin cloud deploy --help
spin-cloud-deploy
Package and upload an application to the Fermyon Platform

USAGE:
    spin cloud deploy [OPTIONS]
```

bacongobbler:
Nice catch. This seems to be due to the fact that `DeployCommand` uses the `clap(about = "")` macro, while `spin cloud deploy` uses the docstring attached to the subcommand.

Fixed in 7b08785.

bacongobbler:
Did the same in 4e62660 for `spin login`.

michelleN (approved):
Thanks for the changes. This looks great 👍

PR #1143, opened 2023-02-14
key-value-sqlite: Temporarily downgrade rusqlite

lann:
We're seeing incompatible versions of libsqlite3 in a dependent project. Downgrading temporarily until we solve the incompatibility.

https://github.com/rusqlite/rusqlite/releases/tag/v0.28.0

### DIFF
--- a/crates/key-value-sqlite/src/lib.rs
+++ b/crates/key-value-sqlite/src/lib.rs
@@ -54,7 +54,7 @@ impl Impl for KeyValueSqlite {
 
                            PRIMARY KEY (store, key)
                         )",
-                        (),
+                        [],
                     )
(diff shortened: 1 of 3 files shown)


dicej (approved):

crates/key-value-sqlite/src/lib.rs
```diff
@@ -54,7 +54,7 @@ impl Impl for KeyValueSqlite {
 
                            PRIMARY KEY (store, key)
                         )",
-                        (),
+                        [],
```
lann, line 57: @dicej FYI, one annoying change: they implemented Params for tuples in 0.28

PR #1144, opened 2023-02-14
refactor key-value host component for per-component configuration

dicej:
Previously, the set of available stores needed to be determined by the time `KeyValueComponent` was created, which turned out to be too early for some use cases.  In this commit, I've changed
`KeyValueComponent::new` to take a function which may be used to create a `StoreManager` (formerly known as `Impl`) for any number of applications.  That `StoreManager`, in turn, may be used to open any number of `Store`s (formerly known as `ImplStore`s).

Note that I've used `Arc<dyn _>` rather than `Box<dyn _>` to represent trait objects in order to facilitate connection reuse where appropriate, among both `Store`s and `StoreManager`s.

### DIFF
--- a/crates/key-value/src/host_component.rs
+++ b/crates/key-value/src/host_component.rs
@@ -1,17 +1,37 @@
-use crate::{Impl, KeyValueDispatch};
-use spin_app::DynamicHostComponent;
+use crate::{KeyValueDispatch, StoreManager};
+use spin_app::{App, AppComponent, DynamicHostComponent};
 use spin_core::HostComponent;
-use std::{collections::HashMap, sync::Arc};
+use std::sync::Arc;
 
 pub const KEY_VALUE_STORES_METADATA_KEY: &str = "key_value_stores";
 
+pub trait StoreManagerManager: Sync + Send {
+    fn get(&self, app: &App) -> Arc<dyn StoreManager>;
+}
+
+impl<F: (Fn(&App) -> Arc<dyn StoreManager>) + Sync + Send> StoreManagerManager for F {
+    fn get(&self, app: &App) -> Arc<dyn StoreManager> {
+        self(app)
+    }
+}
+
+/// Help the rustc type inference engine understand that the specified closure has a higher-order bound so it can
+/// be used as a [`StoreManagerManager`].
+///
+/// See https://stackoverflow.com/a/46198877 for details.
+pub fn manager<F: for<'a> Fn(&'a App) -> Arc<dyn StoreManager>>(f: F) -> F {
+    f
+}
+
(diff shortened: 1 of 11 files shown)


crates/key-value/src/host_component.rs
```diff
@@ -1,17 +1,37 @@
...
+use crate::{KeyValueDispatch, StoreManager};
+use spin_app::{App, AppComponent, DynamicHostComponent};
 use spin_core::HostComponent;
-use std::{collections::HashMap, sync::Arc};
+use std::sync::Arc;
 
 pub const KEY_VALUE_STORES_METADATA_KEY: &str = "key_value_stores";
 
+pub trait StoreManagerManager: Sync + Send {
+    fn get(&self, app: &App) -> Arc<dyn StoreManager>;
```
dicej, line 9: It just occurred to me that this could easily take an `&AppComponent` instead, in case it needs access to metadata at that level.

kate-goldenring: are you saying to reduce what is passed since `AppComponent` can be accessed from `App`?

kate-goldenring: I agree. I think we want component here because we want to prepare the store for a specific component of an app

dicej: I'll make the change.  Regarding your question: it was more the other direction -- you can easily get the `App` from an `AppComponent` if that's all you need.

crates/key-value/src/host_component.rs
```diff
@@ -1,17 +1,37 @@
-use crate::{Impl, KeyValueDispatch};
-use spin_app::DynamicHostComponent;
+use crate::{KeyValueDispatch, StoreManager};
+use spin_app::{AppComponent, DynamicHostComponent};
 use spin_core::HostComponent;
-use std::{collections::HashMap, sync::Arc};
+use std::sync::Arc;
 
 pub const KEY_VALUE_STORES_METADATA_KEY: &str = "key_value_stores";
```
kate-goldenring, line 6: I think it may make sense to make this const (`KEY_VALUE_STORES_METADATA_KEY`) public outside the crate

kate-goldenring (approved):
Love this change! Does a great job of generalizing `StoreManagers` to make it easy to add new ones. Thanks @dicej!

kate-goldenring:
@dicej do you know if the e2e tests were manually canceled by someone or can we rerun?

dicej:
> @dicej do you know if the e2e tests were manually canceled by someone or can we rerun?

I was just wondering about that.  I didn't cancel them.  I'll try to rerun.

PR #1145, opened 2023-02-14
Feature flag `rusqlite`

kate-goldenring:
I went down a rabbit hole and wanted to see if I could find a way to conditionally remove the `rusqlite` dependency.

Anything that brings in the publish crate can remove the `rusqlite` dependency like so:
```
spin-publish = { git = "https://github.com/kate-goldenring/spin", branch = "feat-flag-sqlite", default-features = false}
```

### DIFF
--- a/crates/trigger/Cargo.toml
+++ b/crates/trigger/Cargo.toml
@@ -18,7 +18,7 @@ outbound-redis = { path = "../outbound-redis" }
 outbound-pg = { path = "../outbound-pg" }
 outbound-mysql = { path = "../outbound-mysql" }
 spin-key-value = { path = "../key-value" }
-spin-key-value-sqlite = { path = "../key-value-sqlite", optional = true }
+spin-key-value-sqlite = { path = "../key-value-sqlite", default-features= false, optional = true }
 sanitize-filename = "0.4"
 serde = "1.0"
 serde_json = "1.0"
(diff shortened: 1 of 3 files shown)


crates/trigger/Cargo.toml
```diff
@@ -17,7 +17,7 @@ outbound-redis = { path = "../outbound-redis" }
 outbound-pg = { path = "../outbound-pg" }
 outbound-mysql = { path = "../outbound-mysql" }
 key-value = { path = "../key-value" }
-key-value-sqlite = { path = "../key-value-sqlite" }
+key-value-sqlite = { path = "../key-value-sqlite", default-features= false, optional = true }
```
itowlson, line 20: super mega nit 
```suggestion
key-value-sqlite = { path = "../key-value-sqlite", default-features = false, optional = true }
```

itowlson (approved):
I get the short-term motivation (and I'm fine with it), but I wonder if this would be more cohesive as, say, a "run" feature, which includes all the execution stuff, rather than a special "fully featured except for KV" option.  But that would be more invasive and higher risk for sure.  And if we plan to separate the "prep" part from the "run" part - so as to entirely break the odd dependency from publish to trigger - then it's not worth investing in.

kate-goldenring:
The issue of ignoring a dependency of `spin-trigger` brought in through `spin-publish` has been resolved by @lann's https://github.com/fermyon/spin/pull/1166. Do we anticipate someone using `spin-trigger` but not wanting key-value or the `rusqlite` dependency? I'm assuming we don't have an answer to this right now and suggest we close this and can reopen it if someone has an issue bringing in the dependency

kate-goldenring:
closing in favor of https://github.com/fermyon/spin/pull/1166

PR #1146, opened 2023-02-14
feat: spin up from oci registry

fibonacci1729:
~WIP-ish~

~NOTE: this PR hides bindle commands / related flags on up~


crates/trigger/src/cli.rs
```diff
@@ -119,15 +120,15 @@ where
...
+        let executor: Executor = match self.from_registry {
             true => {
                 let loader =
                     OciTriggerLoader::new(working_dir, self.allow_transient_write, None).await?;
 
                 let trigger_config =
                     TriggerExecutorBuilderConfig::load_from_file(self.runtime_config_file.clone())?;
 
-                let _sloth_warning = warn_if_wasm_build_slothful();
+                _ = warn_if_wasm_build_slothful();
```
itowlson, line 131: I don't think this is safe.  If I recall correctly, the reason the variable is named is because we found that discarding it caused the drop to happen immediately.

src/bin/spin.rs
```diff
@@ -50,7 +50,7 @@ enum SpinApp {
     New(NewCommand),
     Add(AddCommand),
     Up(UpCommand),
-    #[clap(subcommand)]
+    #[clap(subcommand, hide = true)]
```
itowlson, line 53: This seems orthogonal to adding registry support, and I thought we were going to allow a bit of time for deprecation.  For 0.9 we should maybe print a deprecation message; I wouldn't necessarily entwine that with this PR though.

crates/trigger/src/cli.rs
```diff
@@ -119,15 +120,15 @@ where
 
         // TODO: I assume there is a way to do this with a single let mut loader: Box<dyn Loader>
         // variable instead of the entire executor.
-        let executor: Executor = match self.oci {
+        let executor: Executor = match self.from_registry {
```
itowlson, line 123: Man I would really like to get rid of this switch and reunify the lockfile model.  I had guessed that was going to be done as part of this work but time is short!  (This is an observation, rather than a passive-aggressive way of asking you to change it now!)

src/commands/up.rs
```diff
@@ -41,6 +42,7 @@ pub struct UpCommand {
             long = "bindle",
             conflicts_with = APP_CONFIG_FILE_OPT,
             requires = BINDLE_SERVER_URL_OPT,
+            hide = true,
```
itowlson, line 45: Again, probably only a deprecation message at the point?

src/commands/up.rs
```diff
@@ -66,10 +70,21 @@ pub struct UpCommand {
         name = BINDLE_PASSWORD,
         long = "bindle-password",
         env = BINDLE_PASSWORD,
-        requires = BINDLE_USERNAME
+        requires = BINDLE_USERNAME,
+        hide = true,
     )]
     pub bindle_password: Option<String>,
 
+    /// Reference to run the application from a container registry.
```
itowlson, line 78: Given our messaging about Wasm and containers, saying we run things from a 'container' registry risks muddying the waters.  "Oh, does Spin run containers as well as Wasm, or is this running it in a container?"

src/commands/up.rs
```diff
@@ -66,10 +70,21 @@ pub struct UpCommand {
...
+    #[clap(
+        name = REFERENCE_OPT,
+        long = "reference",
```
itowlson, line 81: To me, this name is very vague now that it's decoupled from the `oci/registry` noun context.  I am not sure what is better: `registry-reference` is certainly too long.

src/commands/up.rs
```diff
@@ -66,10 +70,21 @@ pub struct UpCommand {
...
+        env = REFERENCE_OPT,
+        conflicts_with = BINDLE_ID_OPT,
+        conflicts_with = APP_CONFIG_FILE_OPT
+    )]
+    pub reference: Option<String>,
+
     /// Ignore server certificate errors from bindle server
```
itowlson, line 88: Looks like this is also used for registries.

src/commands/up.rs
```diff
@@ -261,6 +268,85 @@ impl UpCommand {
...
+        let locked_app_url = self.write_locked_app(&locked_app, working_dir).await?;
+        Ok((trigger_command, locked_app_url))
+    }
+
+    async fn prepare_app_from_oci(
+        &self,
+        reference: &str,
+        working_dir: &Path,
+    ) -> Result<(Vec<String>, String)> {
+        let mut client = spin_publish::oci::client::Client::new(self.insecure, None)
```
itowlson, line 296: Oh dear, I hadn't clocked that Client lived in `spin_publish`, which feels off when we're in a "pull" situation.  Not an issue for this PR though!

itowlson (changes requested):
The core `spin up` functionality looks fine - the request changes is for some peripheral stuff and UX bikeshedding.

src/bin/spin.rs
radu-matei: Yeah, splitting the PRs seems like a good idea to me.

src/commands/up.rs
radu-matei: Possibly terrible suggestion: `--from-registry ghcr.io/username/app:v1`

crates/trigger/src/cli.rs
lann: Even if it is safe, I think giving it a `_name` helps clarify its purpose in scoping.

src/commands/up.rs
lann: `--open-container-initiative-artifact-reference`

fibonacci1729: Is there resistance to simply `--oci` ?

fibonacci1729: Otherwise `--from-registry` is fine with me.

lann: We just renamed `spin oci` to `spin registry`. :shrug:

lann: Yeah `spin up --from-registry` reads fine to me.

fibonacci1729: Meh, but I could imagine a future where we have a `--warg` option.

Alternatively we could make `spin up` accept a positional argument and infer the source from the scheme....

src/commands/up.rs
fibonacci1729: Punting bindle deprecation out of this PR.

src/commands/up.rs
lann: As a developer tool I think it would be fine for `spin up` to detect the registry type. We can always add e.g. `--registry-type warg` (defaulting to `auto`) later too.

src/commands/up.rs
fibonacci1729: Changed to simply `... from a registry`. Thanks!

src/commands/up.rs
fibonacci1729: Changed to `--from-registry`.

src/commands/up.rs
fibonacci1729: @itowlson I'm not sure what this comment is suggesting.

radu-matei: This ignores certificate errors from the registry as well.

fibonacci1729: Ah. So just adding a `conflicts_with = INSECURE_OPT` is the ask?

fibonacci1729: Nevermind, I see. Just a comment update.

itowlson (approved):
Looks great - thanks!

itowlson:
Not sure why the integration test is failing - the help output looks okay to me... 🤔

PR #1147, opened 2023-02-14
Warn when some templates can't be upgraded

itowlson:
Originally, `spin templates upgrade` ignored templates with no origin info, because there's not much it can do about them.

However, this was a poor initial experience because existing installs would not have origin info.  So now, if we detect that no templates have origin info (and so cannot be upgraded), we print a warning that tells people how to upgrade the old way.

But this is still a suboptimal experience because a user could, say, old-skool upgrade the Spin repo templates and forget about the JS SDK repo.  Now the upgrade command would offer to upgrade the Spin repo, but mysteriously not show the JS repo, and there would be no clue as to why.

This PR adds YET MOAR verbiage to the flood.  In the situation above, the user will now see:

```
$ spin templates upgrade
Spin could not determine where the following templates were installed from:
- http-js
- http-ts
To upgrade them, run `spin templates install --upgrade` with the --git or --dir option

The following template repositories can be automatically upgraded:
> [ ] https://github.com/fermyon/spin (at spin/templates/v0.8)
```

(The additional "can be automatically upgraded" line is shown only if there was a warning; it is there to separate the warning from the selector, and make the context of the selector clear.)

The hope is that, after the initial experience, users will not see this verbiage again.  However, users with templates installed from local directories will see it each time.  Hopefully that is edge case enough not to be a significant issue - and if it is we can refine it over time.

### DIFF
--- a/src/commands/templates.rs
+++ b/src/commands/templates.rs
@@ -226,7 +226,10 @@ impl Upgrade {
         template_manager: &TemplateManager,
     ) -> anyhow::Result<Option<Vec<RepoSelection>>> {
         let existing_templates = template_manager.list().await?.templates;
-        let repos = existing_templates
+        let (origin, no_origin): (Vec<_>, Vec<_>) = existing_templates
+            .iter()
+            .partition(|t| t.source_repo().is_some());
+        let repos = origin
             .iter()
             .filter_map(|t| t.source_repo())
             .collect::<HashSet<_>>();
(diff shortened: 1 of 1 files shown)


src/commands/templates.rs
```diff
@@ -252,6 +255,20 @@ impl Upgrade {
...
+        if !no_origin.is_empty() {
+            eprintln!(
+                "Spin could not determine where the following templates were installed from:"
+            );
+            for template in no_origin {
+                eprintln!("- {}", template.id());
+            }
+            eprintln!("To upgrade them, run `spin templates install --upgrade` with the --git or --dir option");
+            eprintln!();
+            if !self.all {
```
kate-goldenring, line 267: is there a possibility of this message ("The following template repositories can be automatically upgraded:") printing and then there not being any with an origin to display after?

itowlson: I don't believe so.  This is ensured by the `sources.is_empty()` block starting on line 244 - it is guaranteed to early exit on 255.

I have no qualms about adding a belt and braces check here, though, if you think it's worth it?

(Well no qualms except the THIRD RUN through CI but okay that's not so much a qualm as a pet peeve.)

kate-goldenring: Thats perfect as it is. Didnt take a deep look, so thanks for pointing out the check.

kate-goldenring (approved):

PR #1148, opened 2023-02-15
fail fast on ci when lint job fails

rajatjindal:
We should fail fast on lint job failures as generally those fixes are quick ones to do but if we run other jobs before returning this failure, the feedback cycle could be upto an hour. 

Before this fix:

Screenshot 2023-02-15 at 7 01 05 AM

After this fix:

Screenshot 2023-02-15 at 7 03 42 AM

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,7 @@ jobs:
   build-rust:
     name: Build Spin
     runs-on: ${{ matrix.config.os }}
+    needs: [lint-rust]
     strategy:
       matrix:
         config:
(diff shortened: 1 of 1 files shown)


rajatjindal:
The lint job takes about 1m 30s to finish, so any other job will start after that only. 

cc'ing some folks who have made changes or proposed changes to ci/cd for improving the time performance.

@tschneidereit @lann @itowlson

itowlson (approved):
LGTM - thanks from someone who keeps forgetting `cargo fmt`

rajatjindal:
hmm, I merged it without realizing that the checks had not finished yet. we should probably restrict us from doing that.

itowlson:
Given our history of failures that come out of nowhere and won't go away, and how long it takes to run and re-run and re-re-run the checks, I am firmly in favour of being able to override the checks, though yes, better accident prevention measures would be good.

Issue #940, opened 2022-12-06
spin login should store the token in the platofrm keystore


gdamjan:
- `org.freedesktop.secrets` DBus API on Linux (supported by Gnome and KDE at least)
- [macOS Keychain](https://support.apple.com/en-au/guide/keychain-access/welcome/mac)
- [Windows Credential Manager](https://support.microsoft.com/en-au/help/4026814/windows-accessing-credential-manager)

gdamjan:
something like [keyring](https://crates.io/crates/keyring) can help I think.

ps.
I'm willing to work on this - time permitting, but I wont be mad if someone beats me to it first :)

gdamjan:
Some notes:
`service` (in the keyring api) should probably be `spin.fermyon.dev`
`username` should be some identifier of the cloud I guess (but we can go with `"default"` for now)?
`password` will be just the token

michelleN:
@gdamjan I think this would be a great addition. Would you mind writing a quick proposal in the form of a [SIP](https://github.com/fermyon/spin/blob/main/docs/content/sips/index.md) so we can all be on the same page around design and any key decisions? Any contribution here would be much appreciated.

PR #1150, opened 2023-02-15
Implement `spin registry login`

### LINKED ISSUE #1087: Implement `spin registry login`
As of #1014, the `spin registry` set of commands can interact with an OCI registry.
Currently, these commands will reuse the credentials in `~/.docker/config.json` to authenticate to registries, as most registries have instructions on how to authenticate the Docker CLI.

While Spin should continue to use the credentials in `~/.docker/config.json`, and it should continue to allow users to use registries they already authenticated for using their existing toolchains, it should not depend on the Docker CLI being installed on the target machine.

As a result, Spin should implement:

- `spin registry login --username <> --password <>`
and
- `<echo-password-or-get-it-using-cloud-specific-tool> | spin registry login --username <> --pasword-stdin`

This would mirror the default ways to authenticate to container registries today.

radu-matei:
close #1087 

This commit implements functionality for logging in to an OCI registry. Until this commit, a user would have had to use external tooling to log in to a target registry (such as `docker login` or other registry specific tools). This had the advantage that users who already had a setup for working with container regsitries could continue to use it, but it introduced a dependency on external tools for pushing a Spin application.

This commit implements the functionality to login to a registry natively into Spin.
Specifically, it adds this a `spin regsitry login` command:

```
$ spin registry login --help
Log in to a registry

USAGE:
    spin registry login [OPTIONS] <SERVER>

ARGS:
    <SERVER>

OPTIONS:
    -h, --help                   Print help information
    -p, --password <PASSWORD>    Password for the container registry
        --password-stdin         Take the password from stdin
    -u, --username <USERNAME>    Username for the container registry
```

This command mirrors existing tools that perform login operations to registries (flags for the username and passwords, flag to read the password from standard in that takes precedence).

The credentials are base64 encoded (standard for container tools that do not use Docker's credential helpers) and written to `$XDG_CONFIG_HOME/fermyon/registry-auth.json` (side-by-side with the Fermyon Cloud authentication file).

Functionality that needs to work with a registry (`spin registry push/pull`, `spin up`) will first search the Spin configuration for the credentials, then will fallback to searching the authenticated registries in the Docker configuration file.

Examples:

```
$ echo $GHCR_PAT | spin registry login --username radu-matei ghcr.io --password-stdin
Successfully authenticated as radu-matei to registry ghcr.io

$ spin registry login --username radumatei --password <password> https://index.docker.io
Successfully authenticated as radumatei to registry https://index.docker.io

$ spin registry login index.docker.io
Username:
radumatei
Password: (prompt, hidden input)

Successfully authenticated as radumatei to registry https://index.docker.io

$ spin registry login --username radumatei --password bad-password https://index.docker.io
Error: cannot login to the registry

Caused by:
    0: cannot check credentials
    1: cannot authenticate to the registry
    2: unexpected HTTP status 401 Unauthorized
```

This functionality should make `spin registry login` a drop-in replacement for most of the existing tools that login to registries, while continuing to reuse existing credentials as a fallback.

A future commit needs to make writing the configuration file more flexible, and add unit tests for decoding credentials.

### DIFF
--- a/crates/publish/Cargo.toml
+++ b/crates/publish/Cargo.toml
@@ -7,6 +7,8 @@ edition = { workspace = true }
 [dependencies]
 anyhow = "1.0"
 bindle = { workspace = true }
+base64 = "0.21.0"
+dkregistry = { git = "https://github.com/camallo/dkregistry-rs", rev = "37acecb4b8139dd1b1cc83795442f94f90e1ffc5" }
 docker_credential = "1.0"
 dirs = "4.0"
 dunce = "1.0"
--- a/crates/publish/src/oci/client.rs
+++ b/crates/publish/src/oci/client.rs
@@ -1,20 +1,25 @@
 //! Client for distributing Spin applications using OCI registries.
 
 use anyhow::{bail, Context, Result};
-use docker_credential::{CredentialRetrievalError, DockerCredential};
+use docker_credential::DockerCredential;
 use oci_distribution::{
     client::{Config, ImageLayer},
     manifest::OciImageManifest,
     secrets::RegistryAuth,
     Reference,
 };
+use reqwest::Url;
+use serde::{Deserialize, Serialize};
 use spin_app::locked::{ContentPath, ContentRef};
 use spin_loader::oci::cache::Cache;
 use spin_manifest::Application;
 use tokio::fs;
 use walkdir::WalkDir;
 
-use std::path::{Path, PathBuf};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
 
 // TODO: the media types for application, wasm module, and data layer are not final.
 const SPIN_APPLICATION_MEDIA_TYPE: &str = "application/vnd.fermyon.spin.application.v1+config";
@@ -43,7 +48,7 @@ impl Client {
             .as_ref()
             .parse()
             .with_context(|| format!("cannot parse reference {}", reference.as_ref()))?;
-        let auth = Self::auth(&reference)?;
+        let auth = Self::auth(&reference).await?;
         let working_dir = tempfile::tempdir()?;
 
         // Create a locked application from the application manifest.
@@ -139,7 +144,7 @@ impl Client {
     /// Pull a Spin application from an OCI registry.
     pub async fn pull(&mut self, reference: &str) -> Result<()> {
         let reference: Reference = reference.parse().context("cannot parse reference")?;
-        let auth = Self::auth(&reference)?;
+        let auth = Self::auth(&reference).await?;
 
         // Pull the manifest from the registry.
         let (manifest, digest) = self.oci.pull_image_manifest(&reference, &auth).await?;
@@ -209,28 +214,84 @@ impl Client {
         ))
     }
 
+    /// Save a credential set containing the registry username and password.
+    pub async fn login(
+        server: impl AsRef<str>,
+        username: impl AsRef<str>,
+        password: impl AsRef<str>,
+    ) -> Result<()> {
+        // We want to allow a user to login to both https://ghcr.io and ghcr.io.
+        let server = server.as_ref();
+        let server = match server.parse::<Url>() {
+            Ok(url) => url.host_str().unwrap_or(server).to_string(),
+            Err(_) => server.to_string(),
+        };
+
+        // First, validate the credentials. If a user accidentally enters a wrong credential set, this
+        // can catch the issue early rather than getting an error at the first operation that needs
+        // to use the credentials (first time they do a push/pull/up).
+        Self::validate_credentials(&server, &username, &password).await?;
+
+        // Save an encoded representation of the credential set in the local configuration file.
+        let mut auth = AuthConfig::load_default().await?;
+        auth.insert(server, username, password)?;
+        auth.save_default().await
+    }
+
+    /// Validate the credentials by attempting to send an authenticated request to the registry.
+    async fn validate_credentials(
+        server: impl AsRef<str>,
+        username: impl AsRef<str>,
+        password: impl AsRef<str>,
+    ) -> Result<()> {
+        let client = dkregistry::v2::Client::configure()
+            .registry(server.as_ref())
+            .insecure_registry(false)
+            .username(Some(username.as_ref().into()))
+            .password(Some(password.as_ref().into()))
+            .build()
+            .context("cannot create client to send authentication request to the registry")?;
+
+        match client
+            // We don't need to configure any scopes, we are only testing that the credentials are
+            // valid for the intended registry.
+            .authenticate(&[""])
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(e) => bail!(format!(
+                "cannot authenticate as {} to registry {}: {}",
+                username.as_ref(),
+                server.as_ref(),
+                e
+            )),
+        }
+    }
+
     /// Construct the registry authentication based on the reference.
-    fn auth(reference: &Reference) -> Result<RegistryAuth> {
+    async fn auth(reference: &Reference) -> Result<RegistryAuth> {
         let server = reference
             .resolve_registry()
             .strip_suffix('/')
             .unwrap_or_else(|| reference.resolve_registry());
 
-        let creds = docker_credential::get_credential(server);
-        match creds {
-            Err(CredentialRetrievalError::ConfigNotFound) => Ok(RegistryAuth::Anonymous),
-            Err(CredentialRetrievalError::NoCredentialConfigured) => Ok(RegistryAuth::Anonymous),
-            Err(CredentialRetrievalError::ConfigReadError) => Ok(RegistryAuth::Anonymous),
-            Err(e) => bail!("Error handling docker configuration file: {}", e),
+        match AuthConfig::get_auth_from_default(server).await {
+            Ok(c) => Ok(c),
+            Err(_) => match docker_credential::get_credential(server) {
+                Err(e) => {
+                    tracing::trace!("Cannot retrieve credentials from Docker, attempting to use anonymous auth: {}", e);
+                    Ok(RegistryAuth::Anonymous)
+                }
 
-            Ok(DockerCredential::UsernamePassword(username, password)) => {
-                tracing::trace!("Found docker credentials");
-                Ok(RegistryAuth::Basic(username, password))
-            }
-            Ok(DockerCredential::IdentityToken(_)) => {
-                println!("Cannot use contents of docker config, identity token not supported. Using anonymous auth");
-                Ok(RegistryAuth::Anonymous)
-            }
+                Ok(DockerCredential::UsernamePassword(username, password)) => {
+                    tracing::trace!("Found Docker credentials");
+                    Ok(RegistryAuth::Basic(username, password))
+                }
+                Ok(DockerCredential::IdentityToken(_)) => {
+                    tracing::trace!("Cannot use contents of Docker config, identity token not supported. Using anonymous auth");
+                    Ok(RegistryAuth::Anonymous)
+                }
+            },
         }
     }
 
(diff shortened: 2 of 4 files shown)


crates/publish/Cargo.toml
```diff
@@ -7,6 +7,8 @@ edition = { workspace = true }
 [dependencies]
 anyhow = "1.0"
 bindle = { workspace = true }
+base64 = "0.21.0"
+dkregistry = { git = "https://github.com/camallo/dkregistry-rs", rev = "37acecb4b8139dd1b1cc83795442f94f90e1ffc5" }
```
radu-matei, line 11: The latest published version of this crate uses Tokio 0.18, which makes it incompatible with our existing setup.

The functionality used from that crate is minimal (only making an authenticated request to the registry using the credentials, to validate them). If we consider the git source a blocker, we can:

a) remove the functionality altogether — I'm opposed to this, since it catches user errors early
b) fork the repo and publish a new version ourselves
c) update the existing OCI client library to do this instead.

I'm in favor of c).

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,86 @@ impl Client {
...
+
+        // First, test the credentials. If a user accidentally enters a wrong credential set, this
+        // can catch the issue early rather than the user having to figure it out on their own.
+        Self::test_credentials(&server, &username, &password)
+            .await
+            .context("cannot check credentials")?;
+
+        // Save an encoded representation of the credential set in the local configuration file.
+        let mut auth = AuthConfig::load_default().await?;
+        auth.insert(server, username, password)?;
```
lann, line 237: I know this is "how its done" but WHAT YEAR IS THIS?!?!

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,86 @@ impl Client {
...
+    /// Test the credentials by attempting to send an authenticated request to the registry.
+    async fn test_credentials(
+        server: impl AsRef<str>,
+        username: impl AsRef<str>,
+        password: impl AsRef<str>,
+    ) -> Result<()> {
+        let client = dkregistry::v2::Client::configure()
+            .registry(server.as_ref())
+            .insecure_registry(false)
+            .username(Some(username.as_ref().into()))
```
lann, line 250: This suggests that your args should be `impl Into<String>`.

crates/publish/src/oci/client.rs
```diff
@@ -248,3 +311,94 @@ impl Client {
...
+#[derive(Serialize, Deserialize)]
+struct AuthConfig {
+    /// Map between registry server and base64 encoded username:password credential set.
+    pub auths: HashMap<String, String>,
+}
+
+impl AuthConfig {
+    /// Load the authentication configuration from the default location
+    /// ($XDG_CONFIG_HOME/fermyon/registry-auth.json).
+    pub async fn load_default() -> Result<Self> {
```
radu-matei, line 324: Do we want this path to be configurable?

I would really like us to refactor the way we handle configuration throughout Spin, and have a central place where we handle reading and writing configuration...

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,86 @@ impl Client {
...
+        let client = dkregistry::v2::Client::configure()
+            .registry(server.as_ref())
+            .insecure_registry(false)
+            .username(Some(username.as_ref().into()))
+            .password(Some(password.as_ref().into()))
+            .build()
+            .context("cannot create client to send authentication request to the registry")?;
+
+        let dclient = client
+            .authenticate(&[&""])
```
lann, line 256: Comment would help here.

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,86 @@ impl Client {
...
+            .await
+            .context("cannot authenticate to the registry")?;
+        match dclient.is_auth().await {
+            Ok(res) => match res {
+                true => Ok(()),
+                false => bail!("Registry client is not authenticated."),
+            },
```
lann, line 263: Totally optional/preference:
```suggestion
            Ok(true) => Ok(()),
            Ok(false) => bail!("Registry client is not authenticated."),  
```

crates/publish/src/oci/client.rs
lann: Yeah I've been thinking about how to centralize this sort of thing. We could call it `spin-config` no wait... `spin-core`! no.... `spin-app`? shoot...

crates/publish/src/oci/client.rs
radu-matei: I wouldn't optimize the arguments for that call, since it is most likely going away in a future version.

crates/publish/src/oci/client.rs
radu-matei: ref #940

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,89 @@ impl Client {
...
+    pub async fn login(
+        server: impl AsRef<str>,
+        username: impl AsRef<str>,
+        password: impl AsRef<str>,
+    ) -> Result<()> {
+        // We want to allow a user to login to both https://ghcr.io and ghcr.io.
+        let server = match server.as_ref().parse::<Url>() {
+            Ok(url) => url
+                .host_str()
+                .unwrap_or_else(|| server.as_ref())
```
lann, line 227: nit: this is one of the few cases where aliasing variables is really nice: `let server = server.as_ref();`

crates/publish/src/oci/client.rs
lann: I'm mostly complaining about storing _passwords_ rather than tokens, but thats a good point too.

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,89 @@ impl Client {
...
+    /// Save a credential set containing the registry username and password.
+    pub async fn login(
+        server: impl AsRef<str>,
+        username: impl AsRef<str>,
+        password: impl AsRef<str>,
+    ) -> Result<()> {
+        // We want to allow a user to login to both https://ghcr.io and ghcr.io.
+        let server = match server.as_ref().parse::<Url>() {
+            Ok(url) => url
+                .host_str()
```
lann, line 226: This is a little odd as it will strip paths too. Spec aside, that might be surprising for some people: https://github.com/distribution/distribution/issues/29

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,89 @@ impl Client {
...
+                .unwrap_or_else(|| server.as_ref())
+                .to_string(),
+            Err(_) => server.as_ref().to_string(),
+        };
+
+        // First, test the credentials. If a user accidentally enters a wrong credential set, this
+        // can catch the issue early rather than the user having to figure it out on their own.
+        Self::test_credentials(&server, &username, &password)
+            .await
+            .context("cannot check credentials")?;
```
lann, line 236: This would be weird context for "Registry client is not authenticated."

crates/publish/src/oci/client.rs
```diff
@@ -248,3 +314,94 @@ impl Client {
...
+}
+
+impl AuthConfig {
+    /// Load the authentication configuration from the default location
+    /// ($XDG_CONFIG_HOME/fermyon/registry-auth.json).
+    pub async fn load_default() -> Result<Self> {
+        // TODO: add a way to override this path.
+        let default_path = dirs::config_dir()
+            .context("Cannot find configuration directory")?
+            .join("fermyon")
```
lann, line 331: Shouldn't this be "spin"?

crates/publish/src/oci/client.rs
radu-matei: Yeah, we need to address both of them..

crates/publish/src/oci/client.rs
```diff
@@ -248,3 +314,94 @@ impl Client {
...
+        }
+    }
+
+    /// Save the authentication configuration to the default location
+    /// ($XDG_CONFIG_HOME/fermyon/registry-auth.json).
+    pub async fn save_default(&self) -> Result<()> {
+        let default_path = dirs::config_dir()
+            .context("Cannot find configuration directory")?
+            .join("fermyon")
+            .join("registry-auth.json");
```
lann, line 348: Dedupe with above.

crates/publish/src/oci/client.rs
lann: Maybe I'm not following code flow closely enough, but this seems particularly odd given https://github.com/fermyon/spin/pull/1150/files#diff-d0a4fbcfd5b61ee9445cd726d5a0f80132f470ae85c56fdbdc4dfed7e3f48d15R372

radu-matei: Those were 2 different code paths, but the extra mapping for Docker Hub is not needed anymore.

That being said, removing the path is... not ideal. Most registries can redirect, and I can't find a registry service with a path. If we do, and this becomes an issue, we would update this.

crates/publish/src/oci/client.rs
radu-matei: This is where the Cloud auth file is.. really don't want to split to yet another directory before we consolidate..

lann: Well, I think that's where it ultimately belongs and we'll have to do some kind of migration if we want to move it later...

radu-matei: When we moved things to the `fermyon`, we talked about potentially having multiple Fermyon tools sharing that configuration.

I still think that is a good idea, so moving them under `spin` wouldn't really make sense. That being said, without a proposal for consolidating configuration..

crates/publish/src/oci/client.rs
```diff
@@ -248,3 +308,87 @@ impl Client {
...
+                .to_string(),
+            parts
+                .get(1)
+                .context("expected secret as second element of the decoded auth")?
+                .to_string(),
+        ))
+    }
+
+    async fn load(p: &Path) -> Result<Self> {
+        let contents = tokio::fs::read_to_string(&p).await?;
```
itowlson, line 387: Does this need context?  Currently the only place we consume it discards all errors, but if we ever expect errors to be surfaced (and if not, why not return an Option), then the ol' `file or directory does not exist os error 2` could really do with some context for which file it was.

src/commands/registry.rs
```diff
@@ -90,3 +94,65 @@ impl Pull {
...
+    #[clap()]
+    pub server: String,
+}
+
+impl Login {
+    pub async fn run(self) -> Result<()> {
+        let username = match self.username {
+            Some(u) => u,
+            None => {
+                println!("Username: ");
```
itowlson, line 125: Hmm, this will go to a new line and put the cursor at the start of that new line.  It would be nicer to do it on one line using `print!` and `flush`, or the `dialoguer` crate has a helper for this (`dialoguer::Input::new().with_prompt(...).interact_text()`).

(We use `dialoguer` in other commands including templates and plugins.)

src/commands/registry.rs
```diff
@@ -90,3 +94,65 @@ impl Pull {
...
+        // neither was passed, prompt the user to input the password.
+        let password = if self.password_stdin {
+            let mut buf = String::new();
+            let mut stdin = std::io::stdin().lock();
+            stdin.read_to_string(&mut buf)?;
+            buf
+        } else {
+            match self.password {
+                Some(p) => p,
+                None => rpassword::prompt_password("Password:")?,
```
itowlson, line 144: It may be a bit moot with hidden input but the `rpassword` example shows a space after the prompt (i.e. it does not supply the space for you) (if it matters)

src/commands/registry.rs
```diff
@@ -90,3 +94,65 @@ impl Pull {
...
+            }
+        };
+
+        Client::login(&self.server, &username, &password)
+            .await
+            .context("cannot login to the registry")?;
```
itowlson, line 150: for consistency with the help text and because pedantry

```suggestion
            .context("cannot log in to the registry")?;
```

src/commands/registry.rs
```diff
@@ -90,3 +94,65 @@ impl Pull {
...
+
+        println!(
+            "Successfully authenticated as {} to registry {}",
```
itowlson, line 153: Any reason not to use the `logged in` terminology here?

src/commands/registry.rs
```diff
@@ -90,3 +94,65 @@ impl Pull {
         Ok(())
     }
 }
+
+#[derive(Parser, Debug)]
+pub struct Login {
+    /// Username for the container registry
```
itowlson, line 100: Just `registry` rather than bringing up containers?

src/commands/registry.rs
```diff
@@ -90,3 +94,65 @@ impl Pull {
...
+    #[clap(long = "username", short = 'u')]
+    pub username: Option<String>,
+
+    /// Password for the container registry
```
itowlson, line 104: Same comment re 'container'

crates/publish/src/oci/client.rs
```diff
@@ -248,3 +308,87 @@ impl Client {
...
+            .join("fermyon")
+            .join("registry-auth.json"))
+    }
+
+    /// Get the registry authentication for a given registry from the default location.
+    async fn get_auth_from_default(server: impl AsRef<str>) -> Result<RegistryAuth> {
+        let auths = Self::load_default().await?;
+        let encoded = match auths.auths.get(&server.as_ref().to_string()) {
+            Some(e) => e,
+            None => bail!("Cannot find credentials"),
```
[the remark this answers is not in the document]
itowlson: Oh maybe it's moot because we currently always discard the error and fall back to Docker auth.

itowlson (changes requested):
I think it's mostly nits and suggestions - if you just want to land it, then ping me and I can approve instead.

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,83 @@ impl Client {
...
+                e
+            )),
+        }
+    }
     /// Construct the registry authentication based on the reference.
-    fn auth(reference: &Reference) -> Result<RegistryAuth> {
+    async fn auth(reference: &Reference) -> Result<RegistryAuth> {
         let server = reference
             .resolve_registry()
             .strip_suffix('/')
```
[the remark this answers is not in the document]
radu-matei: It's not just `AuthConfig` that needs that formatting.
It was there before, and it's what the Docker credential helper expects as well.

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,83 @@ impl Client {
...
+
+    /// Test the credentials by attempting to send an authenticated request to the registry.
+    async fn test_credentials(
+        server: impl AsRef<str>,
+        username: impl AsRef<str>,
+        password: impl AsRef<str>,
+    ) -> Result<()> {
+        let client = dkregistry::v2::Client::configure()
+            .registry(server.as_ref())
+            .insecure_registry(false)
```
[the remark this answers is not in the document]
radu-matei: Following the UX of the Docker CLI for login (https://docs.docker.com/engine/reference/commandline/login), there is no way to authenticate against an insecure registry. You can use it anonymously to run from, but you should not be able to send credentials to it.

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,83 @@ impl Client {
...
+
+        let dclient = client
+            // We don't need to configure any scopes, we are only testing that the credentials are
+            // valid for the intended registry.
+            .authenticate(&[""])
+            .await
+            .context("cannot authenticate to the registry")?;
+        match dclient.is_auth().await {
+            Ok(true) => Ok(()),
+            Ok(false) => bail!("Registry client is not authenticated."),
```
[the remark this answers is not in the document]
radu-matei: Refactored this a bit and simplified the error checking.

src/commands/registry.rs
radu-matei: Ah, great idea. Updated.

radu-matei:
I addressed most of your comments, @itowlson.
Thanks for the review!

crates/publish/src/oci/client.rs
```diff
@@ -209,28 +214,84 @@ impl Client {
...
+        // to use the credentials (first time they do a push/pull/up).
+        Self::validate_credentials(&server, &username, &password).await?;
+
+        // Save an encoded representation of the credential set in the local configuration file.
+        let mut auth = AuthConfig::load_default().await?;
+        auth.insert(server, username, password)?;
+        auth.save_default().await
+    }
+
+    /// Valiedate the credentials by attempting to send an authenticated request to the registry.
```
fibonacci1729, line 241:
```suggestion
    /// Validate the credentials by attempting to send an authenticated request to the registry.
```

fibonacci1729 (approved):

itowlson (approved):
LGTM!

PR #1152, opened 2023-02-15
Make template upgrade kinder to old installs of common repos

itowlson:
If a user has template installs from 0.8 or earlier, they will not have repo information.  So if the user tries to upgrade them, they will instead get prompted to install-update them.

This PR attempts to provide a better experience for the Spin and Spin JS SDK repos, which are the ones users are most likely to have existing installs of.  It tries, heuristically, to detect if the user has old installs of those repos, and if so inserts them into the list of repos to offer for upgrade.  This means that a user who has installed from one of those repos in the past will now get the nice auto upgrade rather than being asked to use the incantation.  E.g.:

```
# User installed from JS SDK repo with old Spin so no origin info
$ spin templates list --verbose
+------------------------------------------------------------------+
| Name      Description                             Installed from |
+==================================================================+
| http-js   HTTP request handler using Javascript                  |
| http-ts   HTTP request handler using Typescript                  |
+------------------------------------------------------------------+

# What sorcery is this? Upgrade still gives friendly selector instead of manual warning
 $ spin templates upgrade
> [ ] https://github.com/fermyon/spin-js-sdk (at spin/templates/v0.8)
```

There are a couple of ways this can go wrong:

1. The heuristic is fooled if the user has uninstalled the templates we use for sniffing.  In this case the user will be shown the remaining templates as "we couldn't tell where to upgrade from."  This is not delightful but it is benign and a user who is engaged enough to curate individual templates can probably deal with it.

2. If the user is sniffed as having the special repos, we hide *all* manual upgrade warnings.  This is to avoid embedding a full list of special templates into the code.  If the user has other templates that cannot be upgraded, from other repos or from local directories, _no warning will be displayed for these_.  Once the user has upgraded their blessed repos, so that they have origin info, the warnings for the skipped templates will reappear.  This behaviour is admittedly weird and hard to explain to users, but I think the number of people affected will be very small.

Otherwise it seems to work correctly though I'm a bit anxious about having logic this complicated.  Hopefully we can remove it after a few releases!

### DIFF
--- a/src/commands/templates.rs
+++ b/src/commands/templates.rs
@@ -289,13 +289,14 @@ impl Upgrade {
             eprintln!("To upgrade them, run `spin templates install --upgrade` with the --git or --dir option");
             eprintln!();
             if !self.all {
-                eprintln!("The following template repositories can be automatically upgraded:");
+                eprintln!("The following template repositories can be automatically upgraded.");
             }
         }
 
         let selected_sources = if self.all {
             sources
         } else {
+            eprintln!("Select repos to upgrade. Use Space to select/deselect and Enter to confirm selection.");
             let selected_indexes = match dialoguer::MultiSelect::new()
                 .items(&sources)
                 .interact_opt()?
--- a/src/commands/templates.rs
+++ b/src/commands/templates.rs
@@ -229,10 +229,22 @@ impl Upgrade {
         let (origin, no_origin): (Vec<_>, Vec<_>) = existing_templates
             .iter()
             .partition(|t| t.source_repo().is_some());
-        let repos = origin
+
+        let mut repos = origin
             .iter()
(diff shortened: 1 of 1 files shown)


vdice (approved):
From a UX perspective, all appears well to me.  (The selection instructions/prompt is very welcome!)

For testing, I didn't attempt to muck with any of the contents in the `$XDG_HOME/spin/templates` local store, more just removed them and tested scenarios of: 
- templates installed via `--dir`/relative paths (prompted to upgrade from fermyon/spin(-js-sdk) repo and latest templates tag)
- templates installed from my fork of the repos (prompted to upgraded from same forked repo)
- templates installed from fermyon/spin(-js-sdk) (prompted to upgrade from same repo and latest templates tag).

PR #1153, opened 2023-02-15
ref(Dockerfiles): go/tinygo bumps per latest 0.9.0 go templates/SDK

vdice:
- further updates bumping go/tinygo (follow-up from https://github.com/fermyon/spin/pull/1165)

~~Passing e2e is blocked on https://github.com/fermyon/spin/pull/1153#issuecomment-1433512964.  I don't think these bumps are urgent, so this is fine to wait.~~

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -12,12 +12,12 @@ ARG BINDLE_URL="https://github.com/deislabs/bindle/releases/download/v0.8.0/bind
 RUN curl -sL "$BINDLE_URL" | tar -xzf - -C /usr/local/bin bindle bindle-server
 
 # Go installation, see https://go.dev/doc/install
-ARG GO_URL="https://go.dev/dl/go1.17.9.linux-amd64.tar.gz"
+ARG GO_URL="https://go.dev/dl/go1.20.1.linux-amd64.tar.gz"
 RUN curl -sL "$GO_URL" | tar -xzf - -C /usr/local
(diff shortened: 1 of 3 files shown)


vdice:
I should note this updates tinygo binary and cli usage but doesn't attempt any SDK updates.

Although the http-go template and http-tinygo example each build and run fine with these changes and `make test-sdk-go` passes locally, there may be SDK updates necessary.  (Alas, the integration tests currently only test a running app in the form of the [http-go app](https://github.com/fermyon/spin/tree/main/sdk/go/http/testdata/http-tinygo).  We're seeing behavior where [other (tiny)go-based apps aren't functioning as expected](https://github.com/fermyon/spin/issues/820#issuecomment-1432118718), so we may want to add a few more to the mix.)

I'll move to Draft while we learn a bit more from the linked issue; also @dicej is working on feature-related updates to the go SDK as well, so there will be at least one associated PR.

vdice:
The e2e test failure is due to https://github.com/fermyon/spin/issues/1161; the http-go template under test is from the most recent spin template tag and thus doesn't have the changes added in this PR for tinygo 0.27.0 compatibility.

vdice:
e2e is happy, so moving to ready for review.  Although behavior described in https://github.com/fermyon/spin/issues/1161 remains, the tests have fetched the latest v0.9.0 templates which are compatible with the bumps in this PR.

itowlson (approved):

PR #1154, opened 2023-02-15
fix: handling up for local apps was breaking CI

fibonacci1729:
Slight oversight with how we prepare locked application for local vs. remote introduced by [#1146](https://github.com/fermyon/spin/pull/1146) being merged _eagerly_.

This commit rearranges some code in `up` to allow inspecting the trigger type from `spin_manifest::Application` for local applications (vs. `LockedApp` for remote).

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -99,7 +99,7 @@ where
     pub help_args_only: bool,
 
     /// Load the application from the registry.
-    #[clap(long = "from-registry")]
+    #[clap(long = "from-registry", hide = true)]
     pub from_registry: bool,
 }
 
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -9,7 +9,7 @@ use clap::{CommandFactory, Parser};
 use reqwest::Url;
 use spin_app::locked::LockedApp;
 use spin_loader::bindle::BindleConnectionInfo;
-use spin_manifest::ApplicationTrigger;
+use spin_manifest::{Application, ApplicationTrigger};
 use spin_trigger::cli::{SPIN_LOCKED_URL, SPIN_WORKING_DIR};
 use tempfile::TempDir;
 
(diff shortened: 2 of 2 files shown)


crates/trigger/src/cli.rs
```diff
@@ -99,7 +99,7 @@ where
     pub help_args_only: bool,
 
     /// Load the application from the registry.
-    #[clap(long = "from-registry")]
+    #[clap(long = "app-from-registry")]
```
radu-matei, line 102: Does the fix need to change the name of the flag as well?

fibonacci1729: Not necessary for the fix however it is confusing to see `--from-registry` appear twice in the unified help output.

itowlson: It should probably be hidden in the trigger options.  It's not something a user should be able to pass explicitly.  (Eventually we will want to remove it entirely of course.)

fibonacci1729: Good point!

radu-matei: Not sure the fix would be replacing _one_ of them though.
It would appear that there are now two different flags with different names?

fibonacci1729: Yeah, i changed my mind. Back to one flag but hid the trigger one.

src/commands/up.rs
```diff
@@ -199,17 +186,21 @@ impl UpCommand {
             TriggerExecOpts::NoApp => {
                 cmd.arg("--help-args-only");
             }
-            TriggerExecOpts::App {
-                locked_app_url,
+            TriggerExecOpts::Local { app, working_dir } => {
+                let locked_app = spin_trigger::locked::build_locked_app(app, &working_dir)?;
+                let locked_url = self.write_locked_app(&locked_app, &working_dir).await?;
```
itowlson, line 191: I'm not keen on the lockfile builder being moved inside `run_trigger`.  But if that's the quick way out of this issue then we can do it as a fix and revisit the large-scale structure later.

itowlson (approved):
As you noted, this is a bit of a quick-fix to address the ever thorny "trigger options" problem.  It would be nice to revisit the structure - as you note, unifying on the lockfile is an ideal solution, thwarted only by the ugly reality that we want the lockfile to be known valid - but for expediency I reckon it's better to ship this now and reassess later.

src/commands/up.rs
fibonacci1729: Yeah that's where it was before the merge. Just seems easiest until there is time for a proper refactor.

radu-matei (approved):

PR #1159, opened 2023-02-16
Flagless `spin up` application source

### LINKED ISSUE #1155: Flagless `spin up` application source
Instead of having to write `spin up -f ~/myapp/spin.toml` or `spin up --from-registry ghcr/me/myapp:v1`, the user should be allowed to write:

* `spin up ~/myapp/spin.toml` - Spin detects it's a file and runs it
* `spin up ~/myapp` - Spin detects that it's a directory and, if it contains a file named `spin.toml`, runs it
* `spin up ghcr.io/me/myapp:v1` - Spin detects that it's not a file and "looks like" a registry URI, tries to pull the artifact using the OCI protocol, and runs it if successful

The strategy will work like this:

* If no application source if specified:
  * If the current directory contains a file with the default manifest name, use it
  * Otherwise, display help
* If an application source string is specified, try to access the string as a file system reference.
  * If it exists and is a file, use it
  * If it exists and is a directory:
    * If it contains a file with the default manifest name, use that
    * Otherwise, fail with a "directory doesn't contain a manifest" error
  * If it doesn't exist, or isn't a legal filename, try to parse it as a registry reference _[to do: consider heuristic checks first, TBD]_
    * If the parse succeeds, try to pull from the reference (respecting cache etc.)
      * If the pull succeeds, run the artifact.  _(In future, if there are multiple registry protocols with "similar looking" references, this could be elaborated with e.g. check for well-known servers of different types, fail over to the other type, etc.)_
      * If pull fails, fail with a registry error
    * If the parse fails, fail with a "don't understand this source" error

The error messages alone will be glorious.

There should also be a flag that the user can set to force a specific interpretation of the source string e.g. `--source-type file`, `source-type oci`.  The default will be `--source-type auto` which tells Spin to use the strategy given above.

As we are aiming to move from Bindle to OCI, this will not support bindle IDs.  For the time being, the existing bindle-related flags will remain available to support this scenario.

itowlson:
Fixes #1155.

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -1,6 +1,6 @@
 use std::{ffi::OsString, path::PathBuf};
 
-use anyhow::Result;
+use anyhow::{bail, Result};
 use clap::Parser;
 
 use crate::opts::{APP_CONFIG_FILE_OPT, BUILD_UP_OPT, DEFAULT_MANIFEST_FILE};
@@ -11,13 +11,19 @@ use super::up::UpCommand;
 #[derive(Parser, Debug)]
 #[clap(about = "Build the Spin application", allow_hyphen_values = true)]
 pub struct BuildCommand {
+    /// Path to application manifest. The default is "spin.toml".
+    #[clap(name = "APPLICATION")]
+    pub app_source: Option<PathBuf>,
+
     /// Path to spin.toml.
     #[clap(
-            name = APP_CONFIG_FILE_OPT,
-            short = 'f',
-            long = "file",
-        )]
-    pub app: Option<PathBuf>,
+        hide = true,
+        name = APP_CONFIG_FILE_OPT,
+        short = 'f',
+        long = "file",
+        conflicts_with = "APPLICATION"
+    )]
+    pub file_source: Option<PathBuf>,
 
     /// Run the application after building.
     #[clap(name = BUILD_UP_OPT, short = 'u', long = "up")]
@@ -29,10 +35,12 @@ pub struct BuildCommand {
 
 impl BuildCommand {
     pub async fn run(self) -> Result<()> {
-        let manifest_file = self
-            .app
-            .as_deref()
-            .unwrap_or_else(|| DEFAULT_MANIFEST_FILE.as_ref());
+        let manifest_file = match (self.app_source.as_deref(), self.file_source.as_deref()) {
+            (None, None) => DEFAULT_MANIFEST_FILE.as_ref(),
+            (Some(p), None) => p,
+            (None, Some(p)) => p,
+            (Some(_), Some(_)) => bail!("Cannot specify more than one application"),
+        };
 
         spin_build::build(manifest_file).await?;
 
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -44,7 +44,7 @@ impl BuildCommand {
                 )))
                 .chain(self.up_args),
             );
-            cmd.app = Some(manifest_file.into());
+            cmd.file_source = Some(manifest_file.into());
             cmd.run().await
         } else {
             Ok(())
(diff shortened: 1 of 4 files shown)


src/commands/up.rs
```diff
@@ -26,15 +26,34 @@ pub struct UpCommand {
...
+    /// Specifies how to resolve the application argument. The default is 'auto',
+    /// meaning Spin will work it out; you need only specify this if Spin
+    /// is inferring incorrectly.
+    #[clap(
+        name = "APP_SOURCE_TYPE",
+        long = "app-source-type",
+        value_enum,
+        requires = "APPLICATION"
+    )]
+    app_source_type: Option<AppSourceType>,
```
radu-matei, line 44: You could avoid the option and go with `AppSourceType::Auto` as the default value.

itowlson:
I'd particularly appreciate folks bringing their best edge cases (except Lann, his edge cases are _too_ good).  Thanks!

itowlson:
Weird issue where `spin up --bindle spin-headers-env-routes-test/1.0.0 --bindle-server http://127.0.0.1:35339/v1/` works but `spin up --bindle spin-headers-env-routes-test/1.0.0 --bindle-server http://127.0.0.1:35339/v1/ --listen 127.0.0.1:39453` says you can't use `--bindle` with the positional arg.  Investigating.

itowlson:
Seems to happen when a trigger option is specified alongside a conflicty (source) flag.

If I remove the conflict I can see what's going on and CLAP IS INTERPRETING THE TRIGGER OPTION AS THE APP SOURCE AAAHH

itowlson:
So the issue is that we already have the positional `trigger_args` collection, which picks up all arguments not matched by `up` arguments.  But now we have a positional arg for the app source, which picks up the first positional arg.  But that means that if the app source is omitted, and there are trigger args being interpreted positionally, _the first trigger arg gets interpreted as the app source_ (because it's in the first positional, er, position).

I am... actually not sure how to get around this.

itowlson:
At the moment the closest thing to a solution I can think of to this is to retain `-f` but allow it to take directories and registry references as well as plain files.

itowlson:
Lann notes that we could pretend `-f` stood for `--from` if that helps sell anybody on the above proposition.

calebschoepp:
> Lann notes that we could pretend `-f` stood for `--from` if that helps sell anybody on the above proposition.

Commence massive disinformation campaign. In all seriousness though I like this solution.

src/commands/build.rs
```diff
@@ -29,10 +35,12 @@ pub struct BuildCommand {
...
-        let manifest_file = self
-            .app
-            .as_deref()
-            .unwrap_or_else(|| DEFAULT_MANIFEST_FILE.as_ref());
+        let manifest_file = match (self.app_source.as_deref(), self.file_source.as_deref()) {
+            (None, None) => DEFAULT_MANIFEST_FILE.as_ref(),
+            (Some(p), None) => p,
+            (None, Some(p)) => p,
+            (Some(_), Some(_)) => bail!("Cannot specify more than one application"),
+        };
```
calebschoepp, line 43: Is there a reason we don't need the app source type in the build command? Does `spin_build::build(manifest_file)` just out of the box support bindle/registry sources?

itowlson: @calebschoepp Remote sources are already built (and usually wouldn't carry around their source code).  It's not meaningful to pass a remote reference to `spin build`, only a manifest file.  Which is why it currently accepts only the `-f` flag, and in a flagless world could only ever accept the `File` source type, making the option pointless.

I'm not sure I've understood the question correctly - if this doesn't answer it, hound me.

itowlson:
#1179 presents an alternative proposal that is not as beautiful but might be more workable.

src/commands/build.rs
calebschoepp: Nope, totally on the mark. Thanks for answering my question. I didn't put two and two together that a registry source would already be built.

itowlson:
I think we are closing this in favour of #1179.

PR #1160, opened 2023-02-16
lazily create and reuse store manager for each app

dicej:
My previous refactor introduced a performance and logging verbosity regression such that we created a separate store manager for each component of each app on each request, which is a lot more expensive than doing it once per app and reusing it across requests.  It also resulted in printing the database location on every request, which was annoying.

Now we lazily create and reuse the store manager, printing the message only once when the database is first opened.


kate-goldenring (approved):

crates/trigger/src/lib.rs
```diff
@@ -101,34 +102,42 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
...
                                 ))
-                                    as Arc<dyn spin_key_value::StoreManager>,
-                            )]
-                            .into_iter()
-                            .collect(),
-                        ))
-                    })),
+                            };
+
+                            manager.get_or_init(init).clone()
```
lann, line 138: Can this be done eagerly outside of `KeyValueComponent::new`? Is it slow?

dicej: I needs `component.app`, which it doesn't have until the callback is called.  Technically, we could rearrange the code in the outer scope to load the app earlier, which is what I was doing earlier, but that was awkward for lifetime reasons (multiple mutable references to the loader, etc.).

PR #1162, opened 2023-02-16
add e2e tests for redis trigger

rajatjindal:
This PR restructures framework to add support for:

- Passing custom values when creating new app
- use `hyper` http client instead of `reqwest` 
- pass stdout/stderr stream to assertions function to verify logs if needed
- add `trigger_type` to `TestCase` to support `redis trigger`
- added test for `redis trigger`

### DIFF
--- a/crates/e2e-testing/src/testcase.rs
+++ b/crates/e2e-testing/src/testcase.rs
@@ -3,42 +3,69 @@ use crate::metadata_extractor::AppMetadata;
 use crate::spin;
 use crate::utils;
 use anyhow::{Context, Result};
+use core::pin::Pin;
+use derive_builder::Builder;
 use std::fs;
-use tokio::task;
+use std::future::Future;
+use tokio::io::BufReader;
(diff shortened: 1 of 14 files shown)


crates/e2e-testing/src/testcase.rs
```diff
@@ -3,42 +3,69 @@ use crate::metadata_extractor::AppMetadata;
...
+    #[builder(default)]
     pub appname: Option<String>,
 
     /// optional
     /// template to use to create new app
+    #[builder(default)]
     pub template: Option<String>,
 
+    /// optional
+    /// template to use to create new app
```
vdice, line 36:
```suggestion
    /// if provided, appended to `spin new` command
```

vdice (approved):
A lot of great additions here! :clap: LGTM.

crates/e2e-testing/src/testcase.rs
rajatjindal: done

Issue #1102, opened 2023-02-02
Add a `spin dev` command


calebschoepp:
I propose that we add a new top level command `spin dev`. This command seeks to address two issues we have with the dev inner loop: a desire for directory watching/hot reloading (#148 #324) and confusion around where logs go (#993).

When you run `spin dev` the following would happen:

1. `spin build` is run.
2. `spin up --follow-all` is run.
3. The application directory is watched (recursively) to see if there any file changes and if so then steps 1 and 2 are repeated.

`spin dev` would become the new command designed for the dev inner loop experience. It would make finding logs trivial because they'll always be dumped to `stdout`. It would also give us hot reloading. We can also turn on other things like [`wasm_backtrace_details`](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.wasm_backtrace_details) (#389) that we would only want turned on in a dev environment.

The new user flow would become something like:

1. Install spin
2. `spin dev`
3. `spin deploy`

`spin build` and `spin up` would then become more focused on the CI/CD and production use cases. This means we can keep the performant default of writing logs to disk for `spin up`. It also means we don't have to clutter up these commands with flags for hot reloading. For example it would be confusing for them to have to do two flags for the dev workflow `spin build --up --watch`.

Note: One potential issue with this approach is that hot reloading can be a slippery slope of work and weird bugs. Perhaps this isn't a feature we want to maintain. However, IMO hot reloading is critical to the DX.

Note: When I refer to hot reloading I'm talking about completely rerunning `spin build` when any file changes. I'm not talking about a spin server grabbing new web assembly artifacts without stopping.

lann:
I like the idea overall, but I would tweak the rebuild behavior a bit:

- By default, only watch the explicit inputs listed in `spin.toml`, i.e. `component.source`, `component.files`, and `spin.toml` itself.
- Explicitly list patterns to watch for `build` in the `[component.build]` section, e.g. `watch = ["component/src/**/*"]`. These would be relative to `build.workdir`.

Users can still opt-in to "watch everything" with `"**/*"`, but in my experience that can be very frustrating when combined with e.g. rust-analyzer rebuilding stuff in the background all the time.

lann:
Another option would be to delegate rebuilding "watch" behavior to a `build.watch_command` which would be started and stopped with `spin dev` and would be expected to update e.g. `component.source` itself.

calebschoepp:
[...]

Issue #666, opened 2022-07-28
`spin up`: hot reload when component files change


bacongobbler:
As a user of [bartholomew](https://github.com/fermyon/bartholomew), I'd like `spin up` to reload when I change any of the site's content.

For comparison, here's `hugo server`: https://gohugo.io/commands/hugo_server/

If this should be rejected in favour of `bartholomew` introducing its own `serve` command, that's okay too.

radu-matei:
ref #148, #324

As a workaround, we have been using a script similar to this for watching changes and restarting — let me know if this doesn't satisfy your current use case.


 https://github.com/fermyon/spin/blob/e5b4a8dc536d874498e1f3bed274519ff413b066/docs/package.json#L33-L36

bacongobbler:
Thanks. Closing as a duplicate of #148.

Issue #324, opened 2022-04-06
Hot reload of Wasm artifact(s)


mreferre:
A quick tests I have done seems to indicate that Spin does not support hot reloading of Wasm artifacts. The use case would be to have a fleet of Spin "servers" with a remote file system mounted where the Wasm artifacts leave and an update to the app would be as simple as re-provisioning the artifact(s) on the shared volume. This works today if you restart the Spin process (it apparently does not hot reload the new file/version). I think this would be a very useful feature but I don't have a good feeling re if this is a "yeah we will get there" or a "this is just technically impossible with the current state of the art". 

Thoughts? I personally think this would be a useful feature (in some circumstances, surely not if your updates require blue/green or canary etc).

itowlson:
There's some discussion of this in #148 - certainly something we're thinking about, but as your last paragraph notes it's something we need to approach carefully!

For now, we have a sample Node script that can restart if a file changes - see the `spin` script in https://github.com/fermyon/spin/blob/main/docs/package.json.

mreferre:
Thanks. Good to know. To be clear (and to check if I am understanding correctly), that issue is more for external files that are consumed by a Spin-ned application (not for a change of the Wasm module itself). However, the same technique can be applied to Wasm artifacts when they change. For now this is possible via an external application/process that runs and "watches" Wasm artifacts changes but possibly, in the future, it could/should be implemented as a feature flag in Spin that would be able to "auto-reload" when it detects a change. 

Am I getting it right? Thanks for the super fast turn around BTW.

radu-matei:
As you correctly point out, there are two types of files we could watch for changes:

- WebAssembly modules (from the `source` field in the Spin component manifest) — we currently create a module from the source and create pre-instances at startup, i.e. we only read the `.wasm` file once, then, for each new request, we use the pre-instance to create the final instance that handles the request. 
- external static assets (from the `files` field in the Spin component manifest) — as @itowlson points out, at startup we _copy_ all static assets in a temporary directory and _mount the temporary directory_. This means that, as observed in #148, changes in the current directory are not reflected in the Spin application until a restart.

Thinking about a solution that would hot-reload Wasm sources, this would somehow need to tie into the execution context instantiation workflow (which is currently executed when building the execution context):

https://github.com/fermyon/spin/blob/6d650ef75cddb93d6c46f876226b269d28ab769d/crates/engine/src/lib.rs#L119-L143\

[...]

Issue #148, opened 2022-03-09
Option to watch directories for changes and re-start Spin


radu-matei:
Say you are developing a website that runs in Spin (with Bartholomew, for example).

Because we are copying static assets into  a temporary directory, rather than mounting directly, making a change to a file requires restarting `spin up`.
A workaround is having a directory watcher that restarts Spin (see https://github.com/fermyon/spin/pull/145#discussion_r822241029), but it would be great if `spin up` had this option built-in.

joepio:
Not entirely sure if this is the same usecase but:

I'd like to run `spin build --up` on saving a source file. [Notify](https://docs.rs/notify/latest/notify/) might help with this.

I think `spin build --up` or `spin up --watch` or `spin watch` would be fitting candidates.

calebschoepp:
@joepio Just sharing that I've got a prototype `spin watch` PR out that you might want to check out. https://github.com/fermyon/spin/pull/1164

joepio:
> @joepio Just sharing that I've got a prototype `spin watch` PR out that you might want to check out. https://github.com/fermyon/spin/pull/1164

Awesome, thanks!

calebschoepp:
Closing now that #1237 has merged

PR #1164, opened 2023-02-16
Prototype: Add a `spin watch` command

calebschoepp:
# Summary 

There have been many requests for watch functionality in Spin to improve the local developer experience (#148 #666 #324 #1102). This PR is a prototype for a top level `spin watch` command that implements this functionality.

`spin watch` is very simple. When you start it builds and runs your spin app. Then when any relevant file (relevant as defined in the `spin.toml` `component.build.watch` field) changes the watch command rebuilds and runs your app. To keep the command simple all it does is manage children processes of `spin build` and `spin up`.

The goal of this PR is __NOT__ to get feedback on the code — it's super janky and will be reworked when we actually ship this feature. The goal is to come to consensus on what the interface of this feature should look like.

# Try it out

1. Pull this branch
2. `make build`
3. Reinstall templates `spin-dev templates install --dir .`
4. `spin new my-watch-app` (Choose the rust app as I've only updated that template)
5. `cd my-watch-app`
6. `spin watch`

Apologies if it breaks on your machine — it worked on my machine™.

# Interfaces considered

## `spin build --watch --up`

This interface was ruled out because the incantation is too convoluted. It also brings up questions around what happens if the user runs `spin build --watch`, should that still run the spin app or only build it?

## `spin up --watch`

There were two main reasons this interface was ruled out. First, `spin up` is already a very complicated command so adding another flag is undesirable. Second, `spin up` doesn't have any way of building the app. We would be breaking users expectations if passing this one flag suddenly allowed `spin up` to build the spin app.

The watch experience is fundamentally doing both the work of build and up. Both of the flag interfaces suffer from the problem that it is attached to one halve of what watch is doing and thus is ambiguous.

## `spin watch` via plugin

Having watch as a top level command offers a more clear DX with limited overhead — we're just calling out to other spin commands.

Using a plugin was ruled out because it complicates the configuration of what files to watch. If the watch command is in a plugin we can't reasonably store configuration in the `spin.toml`. We also then can't include this configuration in the templates. This forces users to know the correct configuration to pass to the `spin watch` command. 

## `spin watch` via top level command — My preferred interface

`spin watch` as a top level command is the best of all worlds: It has a clear DX and the configuration can be included in `spin.toml` via pre-populated templates.

### DIFF
--- a/src/commands/watch.rs
+++ b/src/commands/watch.rs
@@ -1,0 +1,191 @@
+use std::{
+    ffi::OsString,
+    path::{Path, PathBuf},
+    time::Duration,
+};
+
+use anyhow::{Context, Result};
+use clap::Parser;
+use glob::Pattern;
+use nix::sys::signal::{self, Signal};
+use notify::RecursiveMode;
+use notify_debouncer_mini::{new_debouncer, DebounceEventResult};
+use spin_build::manifest::BuildAppInfoAnyVersion;
+use spin_loader::local::parent_dir;
+use subprocess::{Exec, Redirection};
+use tokio::sync::mpsc::channel;
+
+use crate::opts::{APP_CONFIG_FILE_OPT, DEFAULT_MANIFEST_FILE, WATCH_DELAY_OPT};
+
+/// Rebuild and restart the Spin application when files changes.
+#[derive(Parser, Debug)]
+#[clap(
+    about = "Rebuild and restart the Spin application when files changes",
+    allow_hyphen_values = true
+)]
+pub struct WatchCommand {
+    /// Path to spin.toml.
+    #[clap(
+            name = APP_CONFIG_FILE_OPT,
+            short = 'f',
+            long = "file",
+        )]
+    pub app: Option<PathBuf>,
+
+    /// Milliseconds to delay before rebuilding and restarting the Spin application.
+    #[clap(
+            name = WATCH_DELAY_OPT,
+            short = 'd',
+            long = "delay",
+            default_value = "1000",
+    )]
+    pub delay: u64,
+
+    /// Arguments to be passed through to spin up
+    #[clap()]
+    pub up_args: Vec<OsString>,
+}
+
+impl WatchCommand {
+    pub async fn run(self) -> Result<()> {
+        // Prepare manifest file and find application directory
+        let manifest_file = self.app.unwrap_or_else(|| DEFAULT_MANIFEST_FILE.into());
+        let app_dir = parent_dir(&manifest_file)?;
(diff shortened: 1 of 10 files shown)


src/commands/watch.rs
```diff
@@ -0,0 +1,191 @@
...
+                            println!("====================================\n");
+                        }
+
+                        let _build_status = Exec::shell(format!(
+                            "{} build -f {}",
+                            &spin_cmd.display(),
+                            &manifest_file.display()
+                        ))
+                        .stdout(Redirection::None)
+                        .stderr(Redirection::None)
```
lann, line 142: See https://github.com/fermyon/spin/pull/1130

lann:
It would nice if this watched `component.source`s, maybe iff there is no `component.build.watch`. I don't necessarily want to rebuild my Rust project on every auto-save, especially since rust-analyzer is probably _also_ rebuilding at the same time, but I would like `spin up` to restart when I `cargo build` (or hit the equivalent hotkey).

itowlson:
I like the idea of inferring from `component.files` and `component.source` (when that is a file).  Forcing a build every time I hit Ctrl+S is a recipe for a lot of failed builds and, in Rust at least, those failed builds are pretty costly.  This would also avoid the need for me to list out my source files.

lann:
> This would also avoid the need for me to list out my source files.

Note its a "recursive" watch, so I believe you could just point at your `src/` (though that isn't _always_ adequate).

calebschoepp:
Thanks for the comments @lann and @itowlson. Differentiating between "build watching" (re-run `spin build` when anything in `component.build.watch` changes and "up watching" (restart `spin up` when `component.files` or `component.source` changes) is something we've chatted about before in #1102. For the prototype I ended up grouping them into a single concept to keep things simple. Moreover this is the kind of workflow I would expect from a watch tool. With the way I develop I want it constantly rebuilding in the background and if I think it is being overly sensitive to my auto-saves then I can just bump the delay `spin watch --delay 3000`.

However, you guys are bringing to my attention that other developers have different workflows and may want to kick off the build themselves. This totally make sense to me. While I don't think that this should be the default behaviour of `spin watch` I definitely think it should be supported. I would propose that the watching be internally split back into the "build watching" and "up watching". By default both are constantly running. However, by passing flags `--no-build` or `--no-up` (names pending) you could tailor it to your needs. For example for your guys use case you would run `spin watch --no-build`.

How do you guys feel about that?

lann:
Makes sense to me. In practice I probably just wouldn't set a `component.build.watch` if it was my own project, but a flag is nice to have too.

PR #1166, opened 2023-02-16
Split spin-publish into spin-bindle and spin-oci

lann:
This will help us manage downstream dependencies.


Cargo.toml
```diff
@@ -95,12 +96,14 @@ new-e2e-tests = []
 members = [
```
lann, line 96: Somehow `publish` wasn't in here. Not sure how that was working before... :shrug:

lann:
(not for 0.9)

itowlson (approved):
YES

radu-matei (approved):

PR #1167, opened 2023-02-16
ci(build.yml): rm e2e-tests concurrency config

vdice:
This configuration can block PRs as it is currently written.  However, I don't see a need to prevent concurrent runs for this job, so removing.


rajatjindal (approved):
Thanks Vaughn

PR #1169, opened 2023-02-16
Rename `spin_version` to `spin_manifest_version` in Spin.toml

### LINKED ISSUE #1069: Rename `spin_version` to `spin_manifest_version` in Spin application manifest
## Issue
Spin application manifests (`spin.toml`) must specify what application manifest version it is using. This is set by setting the [`spin_version` in a `spin.toml`](https://github.com/fermyon/spin/blob/7cad829d6c87d6956a049bb6fa7e8d3711b6d70c/examples/http-rust/spin.toml#L1). This field suggests that it is declaring the version of Spin that it runs on when it is really declaring the version of the manifest TOML that it is. We should rename this field to `spin_manifest_version` to make this clear.

## Proposed solution
1) non-breaking: Ideally, we could add an alias to the `spin_version` field of the manifest so that the name change would not be breaking. This may not be possible with `serde` defaults. Instead, we may need to write a custom deserializer that deserializes that field appropriately and calls to the default deserializer for the remaining fields. 

2) Alternatively, we could make it a breaking change with a major release.

kate-goldenring:
Ensures backwards compatibility.

Required bringing in `spin-app` crate to `build` and `loader` crates to access `LockedVersion` type.

fixes #1069 
Thank you @lann for suggested implementation and @fibonacci1729 for pairing!

### DIFF
--- a/crates/build/src/manifest.rs
+++ b/crates/build/src/manifest.rs
@@ -1,12 +1,30 @@
+use serde::{Deserialize, Serialize};
+use spin_loader::local::config::FixedStringVersion;
 use std::path::PathBuf;
 
-use serde::{Deserialize, Serialize};
-
-#[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(tag = "spin_version")]
+#[derive(Clone, Debug, Deserialize)]
+#[serde(untagged)]
 pub(crate) enum BuildAppInfoAnyVersion {
-    #[serde(rename = "1")]
-    V1(BuildAppInfoV1),
+    V1Old {
+        #[allow(dead_code)]
+        spin_version: FixedStringVersion<1>,
+        #[serde(flatten)]
+        manifest: BuildAppInfoV1,
+    },
+    V1New {
+        #[allow(dead_code)]
+        spin_manifest_version: FixedStringVersion<1>,
+        #[serde(flatten)]
+        manifest: BuildAppInfoV1,
+    },
+}
+impl BuildAppInfoAnyVersion {
+    pub fn into_v1(self) -> BuildAppInfoV1 {
+        match self {
+            BuildAppInfoAnyVersion::V1New { manifest, .. } => manifest,
+            BuildAppInfoAnyVersion::V1Old { manifest, .. } => manifest,
+        }
+    }
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
--- a/crates/loader/src/local/config.rs
+++ b/crates/loader/src/local/config.rs
@@ -22,12 +22,43 @@ pub(crate) type RawAppManifestAnyVersionPartial = RawAppManifestAnyVersionImpl<t
 pub(crate) type RawComponentManifestPartial = RawComponentManifestImpl<toml::Value>;
 
 /// Container for any version of the manifest.
-#[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(tag = "spin_version")]
+#[derive(Clone, Debug, Deserialize)]
+#[serde(untagged)]
 pub enum RawAppManifestAnyVersionImpl<C> {
-    /// A manifest with API version 1.
-    #[serde(rename = "1")]
-    V1(RawAppManifestImpl<C>),
+    /// Old Spin manifest versioned with `spin_version` key
+    V1Old {
+        /// Version key name
+        spin_version: FixedStringVersion<1>,
+        /// Manifest
+        #[serde(flatten)]
+        manifest: RawAppManifestImpl<C>,
+    },
+    /// New Spin manifest versioned with `spin_manifest_version` key
+    V1New {
+        /// Version key name
+        spin_manifest_version: FixedStringVersion<1>,
+        /// Manifest
+        #[serde(flatten)]
+        manifest: RawAppManifestImpl<C>,
+    },
+}
+
+impl<C> RawAppManifestAnyVersionImpl<C> {
+    /// Converts `RawAppManifestAnyVersionImpl` into underlying V1 manifest
+    pub fn into_v1(self) -> RawAppManifestImpl<C> {
+        match self {
+            RawAppManifestAnyVersionImpl::V1New { manifest, .. } => manifest,
+            RawAppManifestAnyVersionImpl::V1Old { manifest, .. } => manifest,
+        }
+    }
+
+    /// Returns a reference to the underlying V1 manifest
+    pub fn as_v1(&self) -> &RawAppManifestImpl<C> {
+        match self {
+            RawAppManifestAnyVersionImpl::V1New { manifest, .. } => manifest,
+            RawAppManifestAnyVersionImpl::V1Old { manifest, .. } => manifest,
+        }
+    }
 }
 
 /// Application configuration local file format.
@@ -175,4 +206,28 @@ pub struct FileComponentUrlSource {
     /// The digest of the Wasm binary, used for integrity checking. This must be a
     /// SHA256 digest, in the form `sha256:...`
     pub digest: String,
+}
+
+/// FixedStringVersion represents a schema version field with a const value.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(into = "String", try_from = "String")]
+pub struct FixedStringVersion<const V: usize>;
+
+impl<const V: usize> From<FixedStringVersion<V>> for String {
+    fn from(_: FixedStringVersion<V>) -> String {
+        V.to_string()
+    }
+}
+
+impl<const V: usize> TryFrom<String> for FixedStringVersion<V> {
+    type Error = String;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let value: usize = value
+            .parse()
+            .map_err(|err| format!("invalid version: {}", err))?;
+        if value != V {
+            return Err(format!("invalid version {} != {}", value, V));
+        }
+        Ok(Self)
+    }
 }
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -101,9 +101,8 @@ async fn prepare_any_version(
     base_dst: Option<impl AsRef<Path>>,
     bindle_connection: &Option<BindleConnectionInfo>,
 ) -> Result<Application> {
-    match raw {
-        RawAppManifestAnyVersion::V1(raw) => prepare(raw, src, base_dst, bindle_connection).await,
-    }
+    let manifest = raw.into_v1();
+    prepare(manifest, src, base_dst, bindle_connection).await
 }
 
 /// Iterates over a vector of RawComponentManifest structs and throws an error if any component ids are duplicated
@@ -122,13 +121,11 @@ fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
 
 /// Validate fields in raw app manifest
 pub fn validate_raw_app_manifest(raw: &RawAppManifestAnyVersion) -> Result<()> {
-    match raw {
-        RawAppManifestAnyVersion::V1(raw) => {
-            raw.components
-                .iter()
-                .try_for_each(|c| validate_allowed_http_hosts(&c.wasm.allowed_http_hosts))?;
-        }
-    }
+    let manifest = raw.as_v1();
+    manifest
+        .components
+        .iter()
+        .try_for_each(|c| validate_allowed_http_hosts(&c.wasm.allowed_http_hosts))?;
     Ok(())
 }
 
@@ -405,7 +402,7 @@ impl ComponentDigest {
 fn resolve_partials(
     partially_parsed: RawAppManifestAnyVersionPartial,
 ) -> Result<RawAppManifestAnyVersion> {
-    let RawAppManifestAnyVersionPartial::V1(manifest) = partially_parsed;
+    let manifest = partially_parsed.into_v1();
 
     let app_trigger = &manifest.info.trigger;
     let components = manifest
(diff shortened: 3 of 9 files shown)


crates/build/src/manifest.rs
```diff
@@ -1,12 +1,43 @@
...
+use std::path::PathBuf;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(tag = "spin_version")]
+#[serde(untagged)]
 pub(crate) enum BuildAppInfoAnyVersion {
-    #[serde(rename = "1")]
-    V1(BuildAppInfoV1),
+    V1Old {
+        #[serde(deserialize_with = "deserialize_fixed_version_1")]
```
itowlson, line 9: I notice the enum is marked as Serialize (though I'm not sure we ever serialise it!) but we only customise the Deserialize - does Serialize Just Work?

fibonacci1729 (approved):
LGTM

crates/loader/Cargo.toml
```diff
@@ -24,6 +24,8 @@ reqwest = "0.11.9"
 semver = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+# Brought in for FixedVersion
+spin-app = { path = "../app" }
```
itowlson, line 28: Should we duplicate the implementation rather than relying on the app crate?

SPIN UTILS NOW

crates/loader/src/local/mod.rs
```diff
@@ -405,20 +402,40 @@ impl ComponentDigest {
...
-
-    Ok(RawAppManifestAnyVersion::V1(RawAppManifest {
+    let full_manifest = RawAppManifest {
         info: manifest.info,
         components,
         variables: manifest.variables,
-    }))
+    };
+
+    Ok(match is_old {
```
itowlson, line 429: I don't believe you need to preserve oldness and newness here.  I think it's safe to always emit it as new - nothing cares about oldness or newness, just about the content, which remains identical.

crates/loader/src/local/config.rs
```diff
@@ -23,11 +24,57 @@ pub(crate) type RawComponentManifestPartial = RawComponentManifestImpl<toml::Val
...
+    /// Get V1 manifest
+    pub fn try_into_v1(self) -> Option<RawAppManifestImpl<C>> {
+        Some(match self {
+            RawAppManifestAnyVersionImpl::V1New { manifest, .. } => manifest,
+            RawAppManifestAnyVersionImpl::V1Old { manifest, .. } => manifest,
+        })
+    }
+
+    /// Get V1 manifest
+    pub fn into_v1(self) -> RawAppManifestImpl<C> {
```
itowlson, line 67: I am dubious about the implementation - it looks like the sort of thing that could accidentally get used on odd places and blow up at runtime if/when there is a v2.  If we have this function (and I appreciate we don't want to repeat both patterns in multiple places), I'd like it to do the pattern match so that when v2 comes along the compiler shouts at us and we have to grapple with a safe way of addressing it and hopefully are forced to look mindfully at the sites where this is used.

crates/loader/src/local/mod.rs
```diff
@@ -122,13 +121,11 @@ fn error_on_duplicate_ids(components: Vec<RawComponentManifest>) -> Result<()> {
...
 /// Validate fields in raw app manifest
 pub fn validate_raw_app_manifest(raw: &RawAppManifestAnyVersion) -> Result<()> {
-    match raw {
-        RawAppManifestAnyVersion::V1(raw) => {
-            raw.components
-                .iter()
-                .try_for_each(|c| validate_allowed_http_hosts(&c.wasm.allowed_http_hosts))?;
-        }
-    }
+    let manifest = raw.try_as_v1().expect("expected v1 manifest");
```
itowlson, line 124: Similar comment about compile time vs run time detection, though this one will of course be super obvious.

I wonder if an alternative approach would be to have _yet another_ enum a la `CanonicalisedRawAppManifestAnyVersion { V1(RawAppManifestV1) }` (Nobel Prize for Naming committee, take note), and the loader throws away the New and Old cases _right at the start_ and the rest of the program only ever sees the canonicalised one and can pattern match on it and _know_ at compile time that there is no `V2` case to consider.

itowlson (changes requested):

crates/build/src/manifest.rs
lann: It would "work" but serialize the version as an int, which isn't right. I'm wondering if `FixedVersion` shouldn't be reused from `spin_app` and instead copied & adapted to be a string value.

fibonacci1729: Ahh good catch. I'd vote for just dropping the Serialize.

crates/loader/src/local/config.rs
fibonacci1729: Ivan, I generally agree. But I'd argue that when we introduce V2 we would have revisited and consolidated these serde datastructures in a uniform way. 

FWIW this isn't an uncommon pattern used in rust code. You lose compile time type safety but the idea is that if you need to check if v1 or v2 you'd pattern match only called into_v1 in guaranteed infallible situations

lann: Ideally once we're moving to a V2 any differences between V1 and V2 would be hidden in some wrapper. This pattern is fine when needed but YAGNI here I think.

itowlson: The problem is situations which are infallible now but become fallible later, though, or am I missing your point?  How do you know where it is safe to call this?

fibonacci1729: I agree. But for 1.0 time frame this seems sufficient.

itowlson: Agree but this is the stage where we are dealing with the raw form before we wrap it.

Sounds like others are comfortable though so I'll recede into my hedge.

fibonacci1729: @ivan I agree completely with your concern. This is just a bandaid for technical debt we'll need to address for a more incisive refactoring to anticipate v2.

crates/build/src/manifest.rs
kate-goldenring: I think for backwards compat of `LockedApp`, we need `FixedVersion` to continue to contain a `usize` instead of string.

crates/loader/src/local/mod.rs
kate-goldenring: @fibonacci1729 and I were wondering about this. Sounds good! Thanks

crates/build/src/manifest.rs
kate-goldenring: Created a `FixedStringVersion` which deserializes from `String` into a `usize` constant. I duplicated this across `manifest` and `config` instead of bringing in `spin-app`. I also removed `Serialize` since it is not used.

crates/loader/Cargo.toml
kate-goldenring: Yeah i agree duplication is better. Went that route instead

crates/loader/src/local/mod.rs
kate-goldenring: Cleaned this up since we are only concerned with maintaining the `manifest`

crates/loader/src/local/config.rs
kate-goldenring: We added an `Option` wrapper around `try_into_v1` for flexibility when v2 comes, but it sounds like we will need a refactor when v2 comes anyways. I removed the Option so there isn't any unnecessary unwrapping. It seems pointless to anticipate what we will need for v2 since we will likely be making another wrapper.

crates/build/src/manifest.rs
```diff
@@ -29,3 +46,42 @@ pub(crate) struct RawBuildConfig {
     pub command: String,
     pub workdir: Option<PathBuf>,
 }
+
+/// FixedVersion represents a schema version field with a const value.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(into = "usize", try_from = "String")]
+pub struct FixedVersion<const V: usize>;
```
lann, line 53: `spin-build` already depends on `spin-loader`, so it should probably import `FixedStringVersion` from there.

crates/loader/src/local/config.rs
```diff
@@ -176,3 +207,42 @@ pub struct FileComponentUrlSource {
     /// SHA256 digest, in the form `sha256:...`
     pub digest: String,
 }
+
+/// FixedStringVersion represents a schema version field with a const value.
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(into = "usize", try_from = "String")]
```
lann, line 213:
```suggestion
#[serde(into = "String", try_from = "String")]
```

crates/loader/src/local/config.rs
```diff
@@ -176,3 +207,42 @@ pub struct FileComponentUrlSource {
...
+pub struct FixedStringVersion<const V: usize>;
+
+impl<const V: usize> From<FixedStringVersion<V>> for usize {
+    fn from(_: FixedStringVersion<V>) -> usize {
+        V
+    }
+}
+
```
lann, line 221: Probably just drop this so it isn't used accidentally.

```suggestion
```

crates/loader/src/local/config.rs
```diff
@@ -176,3 +207,42 @@ pub struct FileComponentUrlSource {
...
+impl<const V: usize> TryFrom<String> for FixedStringVersion<V> {
+    type Error = String;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let value: usize = value
+            .parse()
+            .map_err(|err| format!("invalid version: {}", err))?;
+        value.try_into()
+    }
+}
```
lann, line 248: Can collapse this:

```suggestion
impl<const V: usize> TryFrom<String> for FixedStringVersion<V> {
    type Error = String;

    fn try_from(value: String) -> Result<Self, Self::Error> {
        let value: usize = value
            .parse()
            .map_err(|err| format!("invalid version: {}", err))?;
        if value != V {
            return Err(format!("invalid version {} != {}", value, V));
        }
        Ok(Self)
    }
}
```

crates/loader/src/local/config.rs
kate-goldenring: Yikes nice catch!

crates/loader/src/local/config.rs
```diff
@@ -22,12 +22,43 @@ pub(crate) type RawAppManifestAnyVersionPartial = RawAppManifestAnyVersionImpl<t
...
+        /// Manifest
+        #[serde(flatten)]
+        manifest: RawAppManifestImpl<C>,
+    },
+}
+
+impl<C> RawAppManifestAnyVersionImpl<C> {
+    /// Get V1 manifest
+    pub fn into_v1(self) -> RawAppManifestImpl<C> {
+        match self {
```
itowlson, line 49: This is great - the convenience of an `into_v1` with the safety of an exhaustive match.  Thanks!

crates/loader/src/local/config.rs
```diff
@@ -22,12 +22,43 @@ pub(crate) type RawAppManifestAnyVersionPartial = RawAppManifestAnyVersionImpl<t
...
+            RawAppManifestAnyVersionImpl::V1New { manifest, .. } => manifest,
+            RawAppManifestAnyVersionImpl::V1Old { manifest, .. } => manifest,
+        }
+    }
+
+    /// Try to get V1 manifest
```
itowlson, line 55: Comment might be out of date?

examples/http-rust/spin.toml
```diff
@@ -1,4 +1,4 @@
-spin_version = "1"
+spin_manifest_version = "1"
```
itowlson, line 1: Will other examples be changed too, or will we defer that to a later PR?

crates/loader/src/local/config.rs
```diff
@@ -22,12 +22,43 @@ pub(crate) type RawAppManifestAnyVersionPartial = RawAppManifestAnyVersionImpl<t
...
+        spin_manifest_version: FixedStringVersion<1>,
+        /// Manifest
+        #[serde(flatten)]
+        manifest: RawAppManifestImpl<C>,
+    },
+}
+
+impl<C> RawAppManifestAnyVersionImpl<C> {
+    /// Get V1 manifest
+    pub fn into_v1(self) -> RawAppManifestImpl<C> {
```
itowlson, line 48: Is it worth having a unit test that loads manifests with each signature and verifies that they both `into_v1` correctly?

tests/integration.rs
```diff
@@ -994,7 +991,7 @@ mod integration_tests {
     /// in `spin.toml` inside `dir`.
     async fn do_test_build_command(dir: impl AsRef<Path>) -> Result<()> {
         let manifest_file = dir.as_ref().join("spin.toml");
-        let RawAppManifestAnyVersion::V1(manifest) = raw_manifest_from_file(&manifest_file).await?;
+        let manifest = raw_manifest_from_file(&manifest_file).await?.into_v1();
```
itowlson, line 994: Worth having an integration test that checks both the old and new syntaxes?

itowlson (approved):
Thanks Kate, this thoroughly sorts all my concerns!  A few comments/thoughts but nothing to block :shipit:

examples/http-rust/spin.toml
kate-goldenring: Yes, this should be reverted. Like you said, i think we want to do this all at once in a later PR

tests/integration.rs
kate-goldenring: `raw_manifest_from_file` will always map to V1New manifest since that is what `resolve_partials` always maps to. I think a the unit test i added (per your suggestion above) may be sufficient.

crates/loader/src/local/config.rs
kate-goldenring: I added a `test_manifest_v1_signatures` unit test. It essentially just confirms that `into_v1` returns and verifies the manifest name.

fibonacci1729 (approved):

PR #1174, opened 2023-02-17
allow running e2e tests with aarch64

### LINKED ISSUE #1105: unable to use docker compose file to run e2e tests on newer macs
While trying to run the docker compose file for the e2e tests, I ran into the following error:

```
$ docker build -t spin-e2e-tests -f e2e-tests-aarch64.Dockerfile
$ docker compose -f e2e-tests-docker-compose.yml run e2e-tests
[+] Running 0/1
 ⠧ mysql Pulling                                                                                                                      0.6s
no matching manifest for linux/arm64/v8 in the manifest list entries
```

The mysql image being referenced does not support arm64. We might be able to use the `arm64v8/mysql` image or the `mysql/mysql-server` image maintained by the Oracle team in the docker compose file to resolve this issue.

rajatjindal:
fixes #1105

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -170,4 +170,4 @@ jobs:
       - name: Run e2e tests
         run: |
           chmod +x `pwd`/target/release/spin
-          docker compose -f e2e-tests-docker-compose.yml run e2e-tests
+          docker compose -f e2e-tests-docker-compose.yml run -v `pwd`/target/release/spin:/usr/local/bin/spin e2e-tests 
--- a/e2e-tests-aarch64.Dockerfile
+++ b/e2e-tests-aarch64.Dockerfile
@@ -1,5 +1,8 @@
 FROM ubuntu:22.04
 
+ARG BUILD_SPIN=false
+ARG SPIN_VERSION=canary
+
 WORKDIR /root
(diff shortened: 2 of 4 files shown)


vdice (commented):
Mounting in the local `spin` binary didn't work for me: I see an 'exec format error', presumably b/c I built it on darwin arm64 but it isn't compatible w/ the linux aarch64 Dockerfile (despite it being arm64v8-based?  Is that intended to be compatible w/ both?)

Not sure if @michelleN had a different experience.

However, the MYSQL_IMAGE change in the docker-compose file did certainly help.  I was able to run the tests (using the spin binary fetched at Dockerfile build-time).

michelleN (commented):
I wasn't able to run the tests. I got the below error while running the docker compose command.
```
  = note: collect2: fatal error: ld terminated with signal 9 [Killed]
          compilation terminated.


error: could not compile `spin-cli` due to previous error
```

rajatjindal:
Hi @michelleN, are you using arm based laptop? if so, did you build docker image using `aarch` version of Dockerfile. hmm, may be we can pair sometime to check what is going on here

rajatjindal:
it turns out Docker needed more resources.

michelleN (approved):
worked after we increased docker resource limits. thanks for helping debug this @rajatjindal !

PR #1179, opened 2023-02-19
Mostly flagless app source

### LINKED ISSUE #1155: Flagless `spin up` application source

itowlson:
Fixes #1155.

This is similar to #1159 but compromises on total flaglessness to make it play nicely with trigger options.  With this, you still have to write `-f` for anything other than the default source (current directory + spin.toml), but you don't need to write `--from-file` or `--from-registry` unless the "inference engine" gets it wrong.  E.g.

```
# Infers: File: ./spin.toml
spin up

# Infers: File: ./app.toml
spin up -f app.toml

# Infers: File: ./guest/spin.toml
spin up -f guest

# Infers: Registry: ghcr.io/test/myapp:v1
spin up -f ghcr.io/test/myapp:v1
```

Error cases:

```
# If not a file/directory
$ spin up -f guest2
Error: File or directory 'guest2' not found. If you meant to load from a registry, use the `--from-registry` option.

# If not a file/directory, *but* something that looks like a tag is present
$ spin up -f guest2/bish/bosh:v1
Error: cannot pull Spin application from registry

Caused by:
    Not authorized: url https://index.docker.io/v2/guest2/bish/bosh/manifests/v1
```

The following tweaks are also made to the `spin up` UI:
* Bindle flags are hidden but continue to work (no deprecation warning is issued)
* The `from-bindle` alias is added despite this, because consistency even on the edge of the pit
* `--file` is moved to an alias but continues to be supported

There should be no breaking change for the `-f` flag, except that if it is pointed at a non-existent file, it _may_ now fail with a registry error instead of a file not found error.

**NOTE:** this currently has the full, tragic commit history of the flagless work; if we decide to adopt it, then this should be squashed before merging.

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -15,6 +15,8 @@ use tempfile::TempDir;
 
 use crate::opts::*;
 
+const APPLICATION_OPT: &str = "APPLICATION";
+
 /// Start the Fermyon runtime.
 #[derive(Parser, Debug, Default)]
 #[clap(
@@ -26,29 +28,47 @@ pub struct UpCommand {
     #[clap(short = 'h', long = "help")]
     pub help: bool,
 
-    /// Path to spin.toml.
+    /// The application to run. This may be a manifest (spin.toml) file, a
+    /// directory containing a spin.toml file, or a remote registry reference.
+    /// If omitted, it defaults to "spin.toml".
     #[clap(
-        name = APP_CONFIG_FILE_OPT,
+        name = APPLICATION_OPT,
         short = 'f',
-        long = "file",
+        long = "from",
+    )]
+    pub app_source: Option<String>,
+
+    /// The application to run. This is the same as `--from` but forces the
+    /// application to be interpreted as a file or directory path.
+    #[clap(
+        hide = true,
+        name = APP_CONFIG_FILE_OPT,
+        long = "from-file",
+        alias = "file",
         conflicts_with = BINDLE_ID_OPT,
         conflicts_with = FROM_REGISTRY_OPT,
+        conflicts_with = APPLICATION_OPT,
     )]
-    pub app: Option<PathBuf>,
+    pub file_source: Option<PathBuf>,
 
-    /// ID of application bindle.
+    /// The application to run. This interprets the applicaiton as a bindle ID.
+    /// This option is deprecated; use OCI registries and `--from` where possible.
     #[clap(
+        hide = true,
         name = BINDLE_ID_OPT,
         short = 'b',
         long = "bindle",
+        alias = "from-bindle",
         conflicts_with = APP_CONFIG_FILE_OPT,
         conflicts_with = FROM_REGISTRY_OPT,
+        conflicts_with = APPLICATION_OPT,
         requires = BINDLE_SERVER_URL_OPT,
     )]
-    pub bindle: Option<String>,
+    pub bindle_source: Option<String>,
 
     /// URL of bindle server.
     #[clap(
+        hide = true,
         name = BINDLE_SERVER_URL_OPT,
         long = "bindle-server",
         env = BINDLE_URL_ENV,
@@ -57,6 +77,7 @@ pub struct UpCommand {
 
     /// Basic http auth username for the bindle server
     #[clap(
+        hide = true,
         name = BINDLE_USERNAME,
         long = "bindle-username",
         env = BINDLE_USERNAME,
@@ -66,6 +87,7 @@ pub struct UpCommand {
 
     /// Basic http auth password for the bindle server
     #[clap(
+        hide = true,
         name = BINDLE_PASSWORD,
         long = "bindle-password",
         env = BINDLE_PASSWORD,
@@ -73,14 +95,17 @@ pub struct UpCommand {
     )]
     pub bindle_password: Option<String>,
 
-    /// Reference to run the application from a registry.
+    /// The application to run. This is the same as `--from` but forces the
+    /// application to be interpreted as an OCI registry reference.
     #[clap(
+        hide = true,
         name = FROM_REGISTRY_OPT,
         long = "from-registry",
         conflicts_with = BINDLE_ID_OPT,
         conflicts_with = APP_CONFIG_FILE_OPT,
+        conflicts_with = APPLICATION_OPT,
     )]
-    pub reference: Option<String>,
+    pub registry_source: Option<String>,
 
     /// Ignore server certificate errors from bindle server or registry
     #[clap(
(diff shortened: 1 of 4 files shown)


itowlson:
I would _really really_ appreciate it if someone with access to some more realistic registry references could test this - I have been testing with one of Brian's that I don't have access to and confirmed that it gives a suitable OCI error, but it would be good to get more variation especially since the reference parsing involves heuristics.

(Hark at me, first inference engine and now heuristics!  It's a couple of `if` expressions Ivan, don't gussy it up all fancy on them.)

itowlson:
Found a regression where `spin up <file>` (note without the `-f`) now gives trigger options help instead of `Cannot read manifest file from "/home/ivan/spinbin/spin.toml"`.

itowlson:
At least, a regression in the sense of "both are bad but it is a different bad."  Displaying help text is arguably less surprising than firing up the app in the current directory when you thought you'd asked for a different one!

But now we're back to the positional trigger args problem: there's no definite way to distinguish `spin up foo` (wrong, meant to pass `-f`) from `spin up foo` (right, passing `foo` to the trigger)...

itowlson:
To clarify, this only happens if `./spin.toml` doesn't exist.  If It does exist, the current behaviour is to print:

```
error: Found argument 'foo' which wasn't expected, or isn't valid in this context

USAGE:
    spin trigger http [OPTIONS]

```

and this PR does not change that WHEEEE

itowlson:
I piled on another heuristic to handle the `spin up foo` case:

```
$ spin up examples/http-rust
Error: Default file 'spin.toml' found. Did you mean `spin up -f examples/http-rust`?`

$ spin up --listen 127.0.0.1:4000
Serving http://127.0.0.1:4000
```

With this sort of hack we could go back to the fully flagless design, though at least in this one the hack is controlling only the failure mode rather than the interpretation of the application source.

radu-matei:
I'm passing the `-f` and `--from-registry` flags to the same registry app:

```
❯ spin up --listen 127.0.0.1:3003  --from-registry ghcr.io/radu-matei/hello-registries:v1
2023-02-22T10:11:09.247402Z TRACE spin_cli::commands::up: Running trigger executor: "/Users/radu/projects/src/github.com/fermyon/spin/target/release/spin" "trigger" "http" "--listen" "127.0.0.1:3003" "--from-registry"
2023-02-22T10:11:09.272781Z DEBUG spin_loader::oci::cache: using cache root directory /Users/radu/Library/Caches/spin/registry
2023-02-22T10:11:09.289584Z TRACE spin_http: Constructed router for application hello-registries: {Wildcard(""): "hello-registries"}
Serving http://127.0.0.1:3003
2023-02-22T10:11:09.289679Z  INFO spin_http: Serving http://127.0.0.1:3003
Available Routes:
  hello-registries: http://127.0.0.1:3003 (wildcard)


❯ spin up --listen 127.0.0.1:3003  -f ghcr.io/radu-matei/hello-registries:v1
2023-02-22T10:11:51.056584Z DEBUG spin_loader::oci::cache: using cache root directory /Users/radu/Library/Caches/spin/registry
2023-02-22T10:11:52.482909Z TRACE spin_cli::commands::up: Running trigger executor: "/Users/radu/projects/src/github.com/fermyon/spin/target/release/spin" "trigger" "http" "--listen" "127.0.0.1:3003"
Error: loader error: LockedComponentSource missing source field

Caused by:
    LockedComponentSource missing source field
Error: exit status: 1
```

src/commands/up.rs
```diff
@@ -280,7 +388,7 @@ impl UpCommand {
         let exec_opts = if self.help {
             TriggerExecOpts::NoApp
         } else {
-            let from_registry = self.reference.is_some();
+            let from_registry = self.registry_source.is_some();
```
fibonacci1729, line 391: @itowlson Because this isn't necessarily set for a registry reference (e.g. `-f ghcr.io/alice/foobar:v1`) the trigger flag `--from-registry` will not properly be set when inferring an OCI reference from `-f` as the `AppSource`.

I'd suggest changing `prepare_locked_app` to explicitly accept a `from_registry: bool` parameter that is `true` when calling from `prepare_app_from_oci` and `false` for `prepare_app_from_bindle`.

itowlson:
Yikes, thanks to both of you for the catch!

It would be _super duper useful_ if we had a publicly accessible demo that could be used for testing OCI bits.  Or is there one in the e2e tests somewhere?

itowlson:
I think this is now fixed:

```
# Inference
$ spin up -f localhost:5000/itowlson/http-rust:1 --insecure
Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000/hello
    A simple component that returns hello.

$ spin up -f ./examples/http-rust
Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000/hello
    A simple component that returns hello.

$ spin up -f ./examples/http-rust/spin.toml 
Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000/hello
    A simple component that returns hello.

# Forcing flags
$ spin up --from-registry localhost:5000/itowlson/http-rust:1 --insecure
Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000/hello
    A simple component that returns hello.

$ spin up --bindle spin-hello-world/1.0.0 --bindle-server http://localhost:8080/v1 --insecure
Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000/hello
    A simple component that returns hello.

$ spin up --from-file ./examples/http-rust
Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000/hello
    A simple component that returns hello.
```

(I clicked through all of them too to make sure they worked, but will save you the log spew.)

Thanks to @vdice for help getting a working registry and @fibonacci1729 for volunteering his app as tribute.

itowlson:
Marking this as ready for review since people are testing it enough to imply this is the direction we want to go...?

radu-matei:
`ghcr.io/radu-matei/cloud-start:v1` should be public, so you could try with this one: https://github.com/users/radu-matei/packages/container/package/cloud-start

Alternatively, you could push an app to GHCR as well.

fibonacci1729 (approved):
I gave this another round of testing this morning and the code looks reasonable to me! LGTM.

src/commands/up.rs
```diff
@@ -134,13 +159,102 @@ impl UpCommand {
...
+            &self.file_source,
+            &self.bindle_source,
+            &self.registry_source,
+        ) {
+            (None, None, None, None) => self.default_manifest_or_none(),
+            (Some(source), None, None, None) => Self::infer_source(source),
+            (None, Some(file), None, None) => Self::infer_file_source(file.to_owned()),
+            (None, None, Some(id), None) => AppSource::Bindle(id.to_owned()),
+            (None, None, None, Some(reference)) => AppSource::OciRegistry(reference.to_owned()),
+            (_, _, _, _) => AppSource::unresolvable("More than one application was specified"),
```
dicej, line 174: Entirely optional slight simplification:
```suggestion
            _ => AppSource::unresolvable("More than one application was specified"),
```

dicej (approved):
Thanks for doing this (and saving the rest of us from having to)!

src/commands/up.rs
itowlson: Obvious in retrospect but TIL - thank you.

kate-goldenring:
Haven't looked over the implementation yet but the ux was great. Worked with oci images, local files and folders, and with bindle (`./target/debug/spin up --from-bindle spin-hello-world/1.0.0 --bindle-server http://127.0.0.1:8081/v1`)!

PR #1180, opened 2023-02-20
fix dependencies for code coverage build

rajatjindal:
cc @itowlson , thanks for noticing and reporting Ivan.


itowlson (approved):
Looks good and I see the passing action so :shipit:

PR #1185, opened 2023-02-20
Remove deprecated --follow-all option

itowlson:
This completes the shift from "silent by default" to "follow by default" by removing `--follow-all`, which was deprecated in 0.9 because it became a no-op.

cc @tpmccallum to make sure this is removed from docs and examples as it will now cause an error

I also took the chance to slightly reword the `--follow` help text in a way that I feel meshes better with the new default.

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -58,18 +58,13 @@ where
     )]
     pub cache: Option<PathBuf>,
 
-    /// Print output for given component(s) to stdout/stderr
+    /// Print output to stdout/stderr only for given component(s)
     #[clap(
         name = FOLLOW_LOG_OPT,
         long = "follow",
         multiple_occurrences = true,
     )]
     pub follow_components: Vec<String>,
-
-    /// Print all component output to stdout/stderr
-    /// Deprecated
-    #[clap(long = "follow-all")]
-    pub follow_all_components: bool,
 
     /// Silence all component output to stdout/stderr
     #[clap(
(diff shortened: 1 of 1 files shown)


crates/trigger/src/cli.rs
```diff
@@ -58,19 +58,14 @@ where
     )]
     pub cache: Option<PathBuf>,
 
-    /// Print output for given component(s) to stdout/stderr
+    /// Print output to stdout/stderr only for given component(s)
     #[clap(
         name = FOLLOW_LOG_OPT,
         long = "follow",
```
kate-goldenring, line 64: If an app has components A through D and someone specifies `--follow "A" "B" "C"`, does that automatically silence "D"?

itowlson: Yes.  `follow` is effectively "follow only these."  It comes from the previous opt-in model - it opted the nominated components in (whereas `follow-all` opted them all in, which is now replaced by the opt-out model of `--quiet`).  We could change it to be more opt-out, but the main use case for it is "I only want to see this one component that is giving me grief and not be distracted by anything else," so it's still the most concise way of doing that.  I hoped the reword would communicate that, but struggled with the placement of "only" and would be very happy to iterate if it's still unclear.

kate-goldenring: I like the new pattern and think this is a great devx. I just wanted to clarify that it works as expected which is the case! Thanks for making this so clean

kate-goldenring (approved):

radu-matei (approved):

PR #1191, opened 2023-02-22
fix: strip newline and carriage return when password is read from stdin

nico1510:
After reading https://www.fermyon.com/blog/spin-v09#:~:text=Running%20Spin%20Applications%20From%20Registries I tried to login to docker hub but wasn't successful. I traced it down to a newline character which is apparently added when reading from stdin with rust (see https://blog.v-gar.de/2019/04/rust-remove-trailing-newline-after-input).

This command failed: 
`echo $DOCKER_PAT | spin registry login registry-1.docker.io --username $user --password-stdin`

I also had to manually create the `registry-auth.json` where `spin` tries to store the saved credentials but that's just a fyi and is not fixed in this PR.


src/commands/registry.rs
```diff
@@ -143,6 +143,9 @@ impl Login {
             let mut buf = String::new();
             let mut stdin = std::io::stdin().lock();
             stdin.read_to_string(&mut buf)?;
+            // trim trailing newlines and carriage returns (see https://blog.v-gar.de/2019/04/rust-remove-trailing-newline-after-input)
+            let len = buf.trim_end_matches(&['\r', '\n'][..]).len();
+            buf.truncate(len);
```
itowlson, line 148: Is there a reason to use this length to truncate the original `buf` rather than return `buf.trim_end_matches(...).to_owned()`?  That feels more readable to me and I'm not worried if it incurs an extra allocation.  But I'm not familiar with this code and don't know the nuances of the string functions - I'll defer to you and Radu on this!

itowlson (commented):
Thanks for catching this and for taking the time to fix it.  I've asked @radu-matei to review since he knows this code best, but apart from one (non-critical) question it looks fine to me.

src/commands/registry.rs
nico1510: I think the only justification for the length variant would be to "fix" the faulty `buf` by mutating it. But yeah I agree that yours is much more readable. Pushed a change 👍

nico1510:
> Would you mind sharing your environment, please?

Yep sure I'm on:
ProductName:		macOS
ProductVersion:		13.2.1
BuildVersion:		22D68

zsh 5.8.1 (x86_64-apple-darwin22.0)

itowlson:
@radu-matei How do you want to proceed with this one?  Is there any risk or harm in merging it even though it might be environment-specific and we can't reproduce the issue?  Or do you want to try and qualify the issue further?

And @nico1510 my apologies for the long silence!

itowlson:
@radu-matei *bump*

PR #1195, opened 2023-02-23
add `CachingStoreManager` to KV component and use it by default

dicej:
From the doc comments:

> Wrap each `Store` produced by the inner `StoreManager` in an asynchronous, write-behind cache.
>
> This serves two purposes:
>
> - Improve performance with slow and/or distant stores
>
> - Provide a relaxed consistency guarantee vs. what a fully synchronous store provides
>
> The latter is intended to prevent guests from coming to rely on the synchronous consistency model of an
> existing implementation which may later be replaced with one providing a more relaxed, asynchronous
> (i.e. "eventual") consistency model.  See also https://www.hyrumslaw.com/ and https://xkcd.com/1172/.
>
> This implementation provides a "read-your-writes", asynchronous consistency model such that values are
> immediately available for reading as soon as they are written as long as the read(s) hit the same cache as the
> write(s).  Reads and writes through separate caches (e.g. separate guest instances or separately-opened
> references to the same store within a single instance) are _not_ guaranteed to be consistent; not only is
> cross-cache consistency subject to scheduling and/or networking delays, a given tuple is never refreshed from
> the backing store once added to a cache since this implementation is intended for use only by short-lived guest
> instances.
>
> Note that, because writes are asynchronous and return immediately, durability is _not_ guaranteed.  I/O errors
> may occur asyncronously after the write operation has returned control to the guest, which may result in the
> write being lost without the guest knowing.  In the future, a separate `write-durable` function could be added
> to key-value.wit to provide either synchronous or asynchronous feedback on durability for guests which need it.

I've also updated `spin-trigger` to use this by default.

### DIFF
--- a/crates/key-value/src/util.rs
+++ b/crates/key-value/src/util.rs
@@ -1,5 +1,13 @@
 use crate::{Error, Store, StoreManager};
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::{hash_map::Entry, HashMap, HashSet},
+    future::Future,
+    sync::Arc,
+};
+use tokio::{
+    sync::Mutex as AsyncMutex,
+    task::{self, JoinHandle},
+};
 use wit_bindgen_wasmtime::async_trait;
 
 pub struct EmptyStoreManager;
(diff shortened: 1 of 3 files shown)


crates/key-value/src/util.rs
```diff
@@ -31,3 +39,168 @@ impl StoreManager for DelegatingStoreManager {
...
+                e.insert(value.clone());
+
+                value
+            }
+            Entry::Occupied(e) => e.get().clone(),
+        }
+        .ok_or(Error::NoSuchKey)
+    }
+
+    async fn set(&self, key: &str, value: &[u8]) -> Result<(), Error> {
```
lann, line 138: Somewhat an aside, but I wonder if the trait could be changed here to allow `key` and `value` to be passed in already-owned rather than needing to `to_owned` them in here, like a `Cow<str>`. On the other hand, the compiler might be able to optimize away the clone anyway, so perhaps not worth the type noise.

lann (approved):

crates/key-value/src/util.rs
dicej: Yeah, there's a lot of allocation and copying in the current implementation, due to needing to bundle things up into a `'static` `Future`.  I agree `Cow` could be a good option.

crates/key-value/src/util.rs
```diff
@@ -31,3 +39,167 @@ impl StoreManager for DelegatingStoreManager {
...
+
+        Ok(())
+    }
+
+    async fn delete(&self, key: &str) -> Result<(), Error> {
+        // Update the cache and spawn a task to update the backing store asynchronously.
+
+        let mut state = self.state.lock().await;
+
+        state.cache.insert(key.to_owned(), None);
```
kate-goldenring, line 157: Why is this being set to `None` instead of being removed from the cache? I see that we are checking for `None` values for `get` and `get_keys` so implementation is sound, but i curious

kate-goldenring (approved):

crates/key-value/src/util.rs
dicej: If we remove it from the cache entirely, and then the guest asks for that key later, it won't be in the cache, meaning we'll need to hit the backing store (e.g. SQLite) to find out if there's a value or not.  By storing `None`, we know the value's been deleted without needing to consult the backing store.  This is sometimes called a "tombstone" by the way.

PR #1196, opened 2023-02-23
ref(up): update default log dir

adamreese:
+ This change updates where the logs are stored for an app by default. Previously, they were being stored in a user's home directory under `.spin/<app-name>/logs`. This resulted in naming conflicts and a confusing user experience.
+ For a user running an app locally, logs are stored in the same directory as the spin.toml file: <app-dir>/.spin/logs
+ For a user running an app from a registry, logs are no longer stored by default. They are streamed to stdout.
+ The value of the log_dir flag takes precedence if set.

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -23,6 +23,7 @@ pub const RUNTIME_CONFIG_FILE: &str = "RUNTIME_CONFIG_FILE";
 // Set by `spin up`
 pub const SPIN_LOCKED_URL: &str = "SPIN_LOCKED_URL";
 pub const SPIN_WORKING_DIR: &str = "SPIN_WORKING_DIR";
+pub const SPIN_STATE_DIR: &str = "SPIN_STATE_DIR";
 
 /// A command that runs a TriggerExecutor.
 #[derive(Parser, Debug)]
@@ -164,11 +165,28 @@ where
         let mut builder = TriggerExecutorBuilder::new(loader);
         self.update_wasmtime_config(builder.wasmtime_config_mut())?;
 
-        let logging_hooks =
-            StdioLoggingTriggerHooks::new(self.follow_components(), self.log.clone());
+        let logging_hooks = StdioLoggingTriggerHooks::new(self.follow_components(), self.log_dir());
         builder.hooks(logging_hooks);
 
         builder.build(locked_url, trigger_config).await
+    }
+
+    pub fn log_dir(&self) -> Option<PathBuf> {
+        match &self.log {
+            Some(l) => Some(l),
+            None => {
+                if !self.from_registry {
+                    if let Ok(dir) = std::env::var(SPIN_STATE_DIR) {
+                        let s = PathBuf::from(dir).join("logs");
+                        return Some(s);
+                    } else {
+                        tracing::info!("Unable to retrieve SPIN_STATE_DIR environment variable to persist logs",);
+                    };
+                }
+                return None;
+            }
+        };
+        None
     }
 
     pub fn follow_components(&self) -> FollowComponents {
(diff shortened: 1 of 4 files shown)


crates/trigger/src/cli.rs
```diff
@@ -164,13 +165,28 @@ where
...
         builder.hooks(logging_hooks);
 
         builder.build(locked_url, trigger_config).await
     }
 
+    pub fn log_dir(&self) -> Option<PathBuf> {
+        match &self.log {
+            Some(l) => return Some(l.clone()),
+            None => {
+                if self.from_registry {
```
itowlson, line 178: I know this is a decreasing consideration, but `from_registry` will be _false_ for bindles.

Do we have access to the `origin` attribute here?  If so that would be safer I think.

crates/trigger/src/cli.rs
```diff
@@ -164,13 +165,28 @@ where
...
+                    return None;
+                } else {
+                    let state_env = std::env::var(SPIN_STATE_DIR).context(SPIN_STATE_DIR).ok()?;
```
itowlson, line 181: The context here is not very human readable _grin_

crates/trigger/src/cli.rs
```diff
@@ -164,13 +165,28 @@ where
...
+                    let mut app_dir = PathBuf::from(state_env);
+                    app_dir.push(".spin");
```
itowlson, line 183: Suggest `PathBuf::from(...).join(...).join(...)` rather than using mutation.

itowlson, line 183: This appears to be appending `.spin` to the state directory.  I thought our concept for the state directory was that it would be the direct base for the `logs` and KV directories.  Would it be better to append the `.spin` up front where `spin up` sets the environment variable?

src/commands/up.rs
```diff
@@ -187,10 +187,20 @@ impl UpCommand {
                 cmd.arg("--help-args-only");
             }
             TriggerExecOpts::Local { app, working_dir } => {
-                let locked_app = spin_trigger::locked::build_locked_app(app, &working_dir)?;
+                let locked_app = spin_trigger::locked::build_locked_app(app.clone(), &working_dir)?;
+
+                let state_dir = match app.info.origin {
+                    ApplicationOrigin::File(mut f) => {
+                        f.pop();
+                        f
```
itowlson, line 195: Could this not just be `ApplicationOrigin::File(f) => f.parent`?  The mutation, and resultant need for statement flow rather than an expression, are a bit jarring to me.  (Subjective of course.)

itowlson, line 195: Maybe join `.spin` at this point?  That way when we have `--app-state-dir` we can just switch it in here, and won't need to worry about the trigger appending `.spin` to it regardless.

itowlson (changes requested):
Mostly style suggestions ("F# in Rust" vs "Go in Rust" dance off, anybody?) but there is one issue if someone were to use this with bindles.

crates/trigger/src/cli.rs
michelleN: yea I think that's a good point. I will update that.

crates/trigger/src/cli.rs
michelleN: We could check if the SPIN_STATE_DIR is set and if it is not set, return None.

crates/trigger/src/cli.rs
michelleN: updated.

crates/trigger/src/cli.rs
michelleN: updated 👍

michelleN: updated.

src/commands/up.rs
michelleN: done

crates/trigger/src/cli.rs
michelleN: updated.

src/commands/up.rs
michelleN: updated.

dicej (approved):

PR #1198, opened 2023-02-24
chore: update clap dependency to 4.1

cardoso:
I kept changes to a minimum this time, so there's redundant parameters or with better alternatives in v4, but those should be easy to improve as needed. The display order in help was also kept the same by overriding the new default in each Subcommand.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -563,13 +563,28 @@ checksum = "71655c45cb9845d3270c9d6df84ebe72b4dad3c2ba3f7023ad47c144e4e473a5"
 dependencies = [
  "atty",
  "bitflags",
- "clap_derive",
- "clap_lex",
+ "clap_derive 3.2.18",
+ "clap_lex 0.2.4",
  "indexmap",
  "once_cell",
  "strsim",
  "termcolor",
  "textwrap 0.16.0",
+]
+
+[[package]]
+name = "clap"
+version = "4.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


vdice:
Thanks @cardoso, appreciate how this brings the `clap` update but in perhaps a more staged approach so that usage can be further updated (and more recent functionality/features utilized) in follow-ups.

Glancing through the changes, all looks well; however, the e2e tests are catching similar breaking CLI behavior as seen in https://github.com/fermyon/spin/pull/1053#pullrequestreview-1272757152.  Do you think it should be fairly straightforward to address these (eg retain current CLI behavior) and still land the dependency updates?

cardoso:
It's not possible with clap 4. It really wants unknown args to be either after -- or next to each outer,  so it assumes anything after an unknown arg is also unknown. I'm pushing a sample workaround to see what the tests say...

cardoso:
REF: https://github.com/clap-rs/clap/issues/1404#issuecomment-1342502698

cardoso:
I will close this one as blocked by https://github.com/clap-rs/clap/issues/1404 or alternatively, once the breaking change can be made to the to the more conventional `spin up [args] -- [trigger args]`.

I think I reached a solution at some point, but it involved refactoring a lot of code that's receiving constant changes. Suffice to say...

image

itowlson:
Thank you @cardoso for all your effort on this - I'm sorry it didn't work out.  This definitely seems to be a regression between clap 3 and clap 4, but looking at that issue it seems to be 'by design.'  Frustrating!

PR #1200, opened 2023-02-24
remove `Clone` derive for `Store` in the Rust SDK

dicej:
It was a mistake to derive `Clone` for this type since its `Drop` implementation closes the inner store handle.  If there are multiple clones, the first one that gets dropped will close the handle, invalidating the others before they're dropped, too. :facepalm:

Apps that need to `Clone` `Store`s will need to wrap them in `Rc` or `Arc`.


dicej:
This affects other SDKs built on top of the Rust SDK, so I'll open PRs for them also.

karthik2804 (approved):

Issue #1163, opened 2023-02-16
allow setting token via environment variable for `spin cloud deploy`


rajatjindal:
it will be useful to support setting an env variable which can be consumed by `spin cloud deploy` to do the deployment.

vdice:
Can you elaborate on the use case @rajatjindal?  Perhaps for scenarios where the `spin cloud login` action to Fermyon Cloud has been done separate from the current environment (i.e. the json config wouldn't be present locally at `spin cloud deploy` time?)

Question on precedence: If the env var is non-empty, I'm assuming its value should take precedence over the token stored in the json config, if exists, for the current deployment environment?

rajatjindal:
> Can you elaborate on the use case @rajatjindal? Perhaps for scenarios where the `spin cloud login` action to Fermyon Cloud has been done separate from the current environment (i.e. the json config wouldn't be present locally at `spin cloud deploy` time?)
> 

yes, precisely that. if we have a cloud token, we should be able to use that directly without having to write it to a file in a specific format. 

> Question on precedence: If the env var is non-empty, I'm assuming its value should take precedence over the token stored in the json config, if exists, for the current deployment environment?

yes, I would think so too. 

on the other hand, if we want to introduce a flag `--token` instead of environment variable, that works too.

lann:
You'd need an env var for server URL (and arguably "danger_accept_invalid_certs") too, wouldn't you? I wonder if we'd be better off just adding a `--token` opt to `spin cloud login` to update the credentials file.

endocrimes:
I'd be a big proponent of adding env vars over _requiring_ we write config to a fixed file - esp for CI/CD Workflows where you probably wanna pull those from "some" kind of secret store

PR #1201, opened 2023-02-26
use token to login

rajatjindal:
this is a poc for #1163 

cc @itowlson @lann

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -125,10 +125,16 @@ impl DeployCommand {
         };
 
         let mut login_connection: LoginConnection = serde_json::from_str(&data)?;
+        let expired = match has_expired(&login_connection) {
+            Ok(val) => val,
+            Err(err) => {
+                eprintln!("{}\n", err);
+                eprintln!("Run `spin login` to log in again");
+                std::process::exit(1);
+            }
+        };
 
-        let expiration_date = DateTime::parse_from_rfc3339(&login_connection.expiration)?;
-        let now: DateTime<Utc> = Utc::now();
-        if now > expiration_date {
+        if expired {
             // session has expired - log back in
             match self.deployment_env_id {
                 Some(name) => {
@@ -312,7 +318,7 @@ impl DeployCommand {
             insecure: login_connection.danger_accept_invalid_certs,
             token: TokenInfo {
                 token: Some(login_connection.token.clone()),
-                expiration: Some(login_connection.expiration.clone()),
+                expiration: login_connection.expiration,
             },
         };
 
@@ -763,5 +769,21 @@ fn print_available_routes(
                 println!("    {}", description);
             }
         }
+    }
+}
+
+// Check if the token has expired.
+// If the expiration is None, assume the token has not expired
+fn has_expired(login_connection: &LoginConnection) -> Result<bool> {
+    match &login_connection.expiration {
+        Some(expiration) => match DateTime::parse_from_rfc3339(expiration) {
+            Ok(time) => Ok(Utc::now() > time),
+            Err(err) => Err(anyhow!(
+                "Failed to parse token expiration time '{}'. Error: {}",
+                expiration,
+                err
+            )),
+        },
+        None => Ok(false),
     }
 }
--- a/src/commands/login.rs
+++ b/src/commands/login.rs
@@ -20,7 +20,7 @@ use uuid::Uuid;
 use crate::opts::{
     BINDLE_PASSWORD, BINDLE_SERVER_URL_OPT, BINDLE_URL_ENV, BINDLE_USERNAME,
     DEPLOYMENT_ENV_NAME_ENV, HIPPO_PASSWORD, HIPPO_SERVER_URL_OPT, HIPPO_URL_ENV, HIPPO_USERNAME,
-    INSECURE_OPT,
+    INSECURE_OPT, SPIN_AUTH_TOKEN, TOKEN,
 };
 
 // this is the client ID registered in the Cloud's backend
@@ -94,6 +94,14 @@ pub struct LoginCommand {
         requires = HIPPO_USERNAME,
     )]
     pub hippo_password: Option<String>,
+
+    /// Auth Token
+    #[clap(
+        name = TOKEN,
+        long = "token",
+        env = SPIN_AUTH_TOKEN,
+    )]
+    pub token: Option<String>,
 
     /// Display login status
     #[clap(
@@ -256,8 +264,28 @@ impl LoginCommand {
         let login_connection = match self.auth_method() {
             AuthMethod::Github => self.run_interactive_gh_login().await?,
             AuthMethod::UsernameAndPassword => self.run_interactive_basic_login().await?,
+            AuthMethod::Token => self.login_using_token().await?,
         };
         self.save_login_info(&login_connection)
+    }
+
+    async fn login_using_token(&self) -> Result<LoginConnection> {
+        // Validate the token by calling list_apps API until we have a user info API
+        HippoClient::new(ConnectionInfo {
+            url: self.hippo_server_url.to_string(),
+            danger_accept_invalid_certs: self.insecure,
+            api_key: self.token.clone(),
+        })
+        .list_apps()
+        .await
+        .context("Login using the provided token failed. Run `spin login` or create a new token using the Fermyon Cloud user interface.")?;
+
+        let token_info = TokenInfo {
+            token: self.token.clone(),
+            expiration: None,
+        };
+
+        Ok(self.login_connection_for_token(token_info))
     }
 
     async fn run_interactive_gh_login(&self) -> Result<LoginConnection> {
@@ -332,7 +360,7 @@ impl LoginCommand {
             url: self.hippo_server_url.clone(),
             danger_accept_invalid_certs: self.insecure,
             token: token.token.unwrap_or_default(),
-            expiration: token.expiration.unwrap_or_default(),
+            expiration: token.expiration,
             bindle_url: Some(bindle_url),
             bindle_username,
             bindle_password,
@@ -344,7 +372,7 @@ impl LoginCommand {
             url: self.hippo_server_url.clone(),
             danger_accept_invalid_certs: self.insecure,
             token: token_info.token.unwrap_or_default(),
-            expiration: token_info.expiration.unwrap_or_default(),
+            expiration: token_info.expiration,
             bindle_url: None,
             bindle_username: None,
             bindle_password: None,
@@ -386,6 +414,8 @@ impl LoginCommand {
             // prompt the user for the authentication method
             // TODO: implement a server "feature" check that tells us what authentication methods it supports
             prompt_for_auth_method()
+        } else if self.token.is_some() {
+            AuthMethod::Token
         } else {
             AuthMethod::Github
         }
@@ -490,7 +520,8 @@ pub struct LoginConnection {
     pub bindle_password: Option<String>,
     pub danger_accept_invalid_certs: bool,
     pub token: String,
-    pub expiration: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub expiration: Option<String>,
 }
 
 #[derive(Deserialize, Serialize)]
@@ -546,6 +577,8 @@ pub enum AuthMethod {
     Github,
     #[clap(name = "username")]
     UsernameAndPassword,
+    #[clap(name = "token")]
+    Token,
 }
 
 fn prompt_for_auth_method() -> AuthMethod {
--- a/src/opts.rs
+++ b/src/opts.rs
@@ -21,3 +21,5 @@ pub const PLUGIN_ALL_OPT: &str = "ALL";
 pub const PLUGIN_OVERRIDE_COMPATIBILITY_CHECK_FLAG: &str = "override-compatibility-check";
 pub const HELP_ARGS_ONLY_TRIGGER_TYPE: &str = "provide-help-args-no-app";
 pub const FROM_REGISTRY_OPT: &str = "REGISTRY_REFERENCE";
+pub const TOKEN: &str = "TOKEN";
+pub const SPIN_AUTH_TOKEN: &str = "SPIN_AUTH_TOKEN";


bacongobbler:
I’d recommend calling the environment variable FERMYON_TOKEN. Hippo has no concept of a personal access token at this time.

src/commands/login.rs
```diff
@@ -95,6 +95,14 @@ pub struct LoginCommand {
     )]
     pub hippo_password: Option<String>,
 
+    /// Fermyon token
+    #[clap(
+        name = FERMYON_TOKEN,
+        long = "fermyon-token",
```
lann, line 101: It happens to be Fermyon-specific today, but I don't think the flag needs such a specific name.

```suggestion
        long = "token",
```

src/opts.rs
```diff
@@ -21,3 +21,4 @@ pub const PLUGIN_ALL_OPT: &str = "ALL";
 pub const PLUGIN_OVERRIDE_COMPATIBILITY_CHECK_FLAG: &str = "override-compatibility-check";
 pub const HELP_ARGS_ONLY_TRIGGER_TYPE: &str = "provide-help-args-no-app";
 pub const FROM_REGISTRY_OPT: &str = "REGISTRY_REFERENCE";
+pub const FERMYON_TOKEN: &str = "FERMYON_TOKEN";
```
lann, line 24: We should really rename _all_ the `HIPPO_*` auth vars rather than introducing a single different one. Maybe `SPIN_LOGIN_*`?

src/commands/login.rs
```diff
@@ -179,14 +187,18 @@ impl LoginCommand {
...
+            (false, true, false, None, None) => self.run_status().await,
+            (false, false, true, None, None) => self.run_get_device_code().await,
+            (false, false, false, Some(device_code), None) => {
                 self.run_check_device_code(device_code).await
             }
-            (false, false, false, None) => self.run_interactive_login().await,
+            (false, false, false, None, Some(fermyon_token)) => {
+                self.login_using_fermyon_token(fermyon_token).await
+            }
+            (false, false, false, None, None) => self.run_interactive_login().await,
```
lann, line 201: This is in dire need of refactoring, but doesn't need to happen here.

src/commands/login.rs
```diff
@@ -260,6 +272,16 @@ impl LoginCommand {
         self.save_login_info(&login_connection)
     }
 
+    async fn login_using_fermyon_token(&self, token: &str) -> Result<()> {
```
lann, line 275: This should verify the token before saving. I don't see a dedicated "user info" API but it could just check the apps list.

src/commands/login.rs
rajatjindal: done

src/commands/login.rs
rajatjindal: done

src/commands/login.rs
rajatjindal: we moved login under `run_interactive_login`, so this change is stale now

src/opts.rs
rajatjindal: renamed the FERMYON_TOKEN to token. about other renames, I think it is a bigger change as all the api/client at this moment refer to hippo. would it be ok if we defer that for a future PR?

rajatjindal:
this is ready for review again @lann @itowlson @bacongobbler, thanks

src/commands/login.rs
```diff
@@ -256,10 +264,29 @@ impl LoginCommand {
...
         self.save_login_info(&login_connection)
     }
 
+    async fn login_using_token(&self) -> Result<LoginConnection> {
+        HippoClient::new(ConnectionInfo {
+            url: self.hippo_server_url.to_string(),
+            danger_accept_invalid_certs: self.insecure,
+            api_key: self.token.clone(),
+        })
+        .list_apps()
```
radu-matei, line 278: Why is listing apps part of the login flow?

src/commands/deploy.rs
```diff
@@ -125,10 +125,9 @@ impl DeployCommand {
         };
 
         let mut login_connection: LoginConnection = serde_json::from_str(&data)?;
-
-        let expiration_date = DateTime::parse_from_rfc3339(&login_connection.expiration)?;
-        let now: DateTime<Utc> = Utc::now();
-        if now > expiration_date {
+        let expiration = login_connection.expiration.clone();
+        if expiration.is_some() && Utc::now() > DateTime::parse_from_rfc3339(&expiration.unwrap())?
```
radu-matei, line 129: This should not unwrap.

src/commands/login.rs
lann: We don't have a dedicated API to verify auth info.

radu-matei: Oh, so we're verifying the operation succeeds. Got it. I think it's worth adding a comment explaining that.

src/opts.rs
lann: This name is used for the env var, so it should be given _some_ prefix.

src/commands/deploy.rs
itowlson, line 129: Suggestion: move this logic out to a `has_expired(&LoginConnection)` function - the nesting of brackets and connectives is otherwise fiddly for a reader to parse.

src/commands/login.rs
```diff
@@ -95,6 +95,14 @@ pub struct LoginCommand {
     )]
     pub hippo_password: Option<String>,
 
+    /// Auth Token
+    #[clap(
+        name = TOKEN,
+        long = "token",
+        env = TOKEN,
```
itowlson, line 102: We are generally lazy about this but I think this case merits decoupling the environment variable name from the argument name.  As @lann suggests, something like `SPIN_LOGIN_TOKEN` would be a good bet for the environment variable.  But in help (which uses the `name` by default) you probably want it to appear as `spin login --token <TOKEN>`.

src/commands/login.rs
```diff
@@ -256,10 +264,29 @@ impl LoginCommand {
         let login_connection = match self.auth_method() {
             AuthMethod::Github => self.run_interactive_gh_login().await?,
             AuthMethod::UsernameAndPassword => self.run_interactive_basic_login().await?,
+            AuthMethod::Token => self.login_using_token().await?,
```
itowlson, line 267: I have probably misread but currently `auth_method()` looks like it only returns `Token` if the user passes `--auth-method token` on the command line.  Do we not infer it from the `--token` option?  (I guess that could cause issues if the user wanted to override an ambient EV with a different login option... but...)  Again, I have probably misunderstood, so please correct me!

src/opts.rs
rajatjindal: right 🤦, added `SPIN_LOGIN_` prefix for `TOKEN`. thank you Lann

src/commands/login.rs
rajatjindal: added the comment

src/commands/deploy.rs
rajatjindal: sure, done

src/commands/login.rs
rajatjindal: done

src/commands/deploy.rs
rajatjindal: Because `expiration` is an `Option<String>` now, we need `unwrap` to get the underlying `String`? There is a good chance that there is a better way to achieve this. Please let me know.

Also, I have moved this logic to the `has_expired` function, and using a `match` instead of an `unwrap`.

itowlson: @rajatjindal the better way is indeed the `match`, which you've already adopted _grin_

rajatjindal: I was asking if there was a better way to do this within the if condition

src/commands/deploy.rs
```diff
@@ -750,3 +747,12 @@ fn print_available_routes(
         }
     }
 }
+
+// Check if the token has expired.
+// If the expiration is None, assume the token has not expired
+fn has_expired(login_connection: &LoginConnection) -> bool {
+    match &login_connection.expiration {
+        Some(expiration) => Utc::now() > DateTime::parse_from_rfc3339(&expiration).unwrap(),
```
itowlson, line 755: The `unwrap` here is "crash the program if the parse fails", which is not what you want - sure, it _shouldn't_ happen, but it's not a program-stopping disaster.  I'd suggest falling back to the safe option: if we can't parse the expiry string, assume it's expired - it's more fiddly nesting but totally worth it.  E.g.

```suggestion
        Some(expiration) => match DateTime::parse_from_rfc3339(&expiration) {
            Ok(time) => Utc::now() > time,
            Err(_) => true,
        }
```

src/commands/deploy.rs
itowlson: Well, you could do the match right there in the condition.  _grin_  But that tends to hide the structure of the program.

Honestly, this kind of lifted fallible operation gets fiddly enough to make putting it in the condition of an `if` a bit hard to read, which is why I often prefer to pull out a function or named variable.  I _think_ `expiration.map(|e| Utc::now() > DateTime.parse(e).unwrap_or(DateTime::MIN_UTC)).unwrap_or_default` might work, though I'd have to think _really really_ hard to be sure about the None and Err cases, and it's far far worse to have in an `if` condition than `is_some/unwrap`!

src/commands/deploy.rs
rajatjindal: done

itowlson (approved):

src/commands/deploy.rs
```diff
@@ -750,3 +747,15 @@ fn print_available_routes(
...
     }
 }
+
+// Check if the token has expired.
+// If the expiration is None, assume the token has not expired
+fn has_expired(login_connection: &LoginConnection) -> bool {
+    match &login_connection.expiration {
+        Some(expiration) => match DateTime::parse_from_rfc3339(expiration) {
+            Ok(time) => Utc::now() > time,
+            Err(_) => true,
```
bacongobbler, line 757: So if I'm reading this correctly, if the expiration date cannot be parsed from the token string, we consider it to be expired.

Should we log an error message about the expiration date failing to be parsed? Shouldn't we inform the user that the token contains an invalid expiration date?

src/opts.rs
```diff
@@ -21,3 +21,5 @@ pub const PLUGIN_ALL_OPT: &str = "ALL";
 pub const PLUGIN_OVERRIDE_COMPATIBILITY_CHECK_FLAG: &str = "override-compatibility-check";
 pub const HELP_ARGS_ONLY_TRIGGER_TYPE: &str = "provide-help-args-no-app";
 pub const FROM_REGISTRY_OPT: &str = "REGISTRY_REFERENCE";
+pub const TOKEN: &str = "TOKEN";
+pub const SPIN_LOGIN_TOKEN: &str = "SPIN_LOGIN_TOKEN";
```
bacongobbler, line 25: I'm curious why "login" was chosen as part of the name. In my mind, that seems to imply that this token is to be used for logging in, and other tokens are meant for other operations (e.g. `SPIN_DEPLOY_TOKEN`).

What if we used SPIN_AUTH_TOKEN instead?

itowlson: `SPIN_LOGIN_TOKEN` was me and/or Lann thinking out loud - for me at least there was no particular reason.  I agree `SPIN_AUTH_TOKEN` is better.  Sorry Rajat for the bad steer.

src/commands/deploy.rs
rajatjindal: currently if the token is expired (or it has failed during parsing), we initialize the login process automatically. after this log msg, this is how it would look like to user

```
$) spin deploy
Failed to parse token expiration time 'abc'
Error: premature end of input

Copy your one-time code:

2geg5Xf9

...and open the authorization page in your browser:

https://cloud.fermyon.com/device-authorization

Waiting for device authorization...

```

does this look ok?

rajatjindal: in a CI scenario, if this were to happen, the step will be stuck forever. should we just exit instead of restarting auth if expiration parsing has failed.

here is what we currently do when token json file is invalid:

```
$) spin deploy                    
Error: expected `,` or `}` at line 3 column 3
```

src/opts.rs
rajatjindal: no worries. this certainly sounds better. i've changed it to `SPIN_AUTH_TOKEN` now.

radu-matei (approved):
Manually tested, things look great!

LGTM

src/commands/login.rs
```diff
@@ -256,10 +264,30 @@ impl LoginCommand {
...
+    async fn login_using_token(&self) -> Result<LoginConnection> {
+        // Validate the token by calling list_apps API until we have a user info API
+        HippoClient::new(ConnectionInfo {
+            url: self.hippo_server_url.to_string(),
+            danger_accept_invalid_certs: self.insecure,
+            api_key: self.token.clone(),
+        })
+        .list_apps()
+        .await
+        .context("token validation failed")?;
```
radu-matei, line 281: Entered a wrong token and this was the result:

```
$ spin cloud login
Error: token validation failed

Caused by:
```

rajatjindal: @itowlson earlier suggested:

```
"Login using the provided token failed.  Run spin login or create a new token using the Fermyon Cloud user interface."
```

does this look ok? 

(this will change `token validation failed` with above message. `Caused by:` will remain as is because it seems to be coming from `anyhow` crate)

src/commands/deploy.rs
bacongobbler: I think we should exit when we've received invalid data.

rajatjindal: changed it to print following msg, and exit if parsing of expiration time fails.

```
spin deploy
Failed to parse token expiration time 'abc'. Error: premature end of input

Run `spin login` to log in again
```

src/commands/login.rs
rajatjindal: this is how it looks with latest changes:

```
➜  spin login --token abc
Error: Login using the provided token failed. Run `spin login` or create a new token using the Fermyon Cloud user interface.

Caused by:
```

michelleN (approved):
lgtm. had a question about expiration details but not a blocker. Thanks!

src/commands/login.rs
```diff
@@ -256,10 +264,30 @@ impl LoginCommand {
...
+            danger_accept_invalid_certs: self.insecure,
+            api_key: self.token.clone(),
+        })
+        .list_apps()
+        .await
+        .context("Login using the provided token failed. Run `spin login` or create a new token using the Fermyon Cloud user interface.")?;
+
+        let token_info = TokenInfo {
+            token: self.token.clone(),
+            expiration: None,
```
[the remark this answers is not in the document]
bacongobbler: Yes. Eventually personal access tokens can be created with a timed expiration, but it's not part of the MVP.

rajatjindal:
thank you for the review/approval Michelle (and everyone)

Issue #1188, opened 2023-02-22
Update URL component source cache to use configured root


itowlson:
The URL component source cache PR (#1088) included the comment:

"TODO: when `spin up` integrates running an app from OCI, pass the configured cache root to this function. For now, use the default cache directory." (https://github.com/fermyon/spin/pull/1088/files/4d99bcf842d7e3e1166db8a43a438613e33b0090#r1098922615)

Now that `spin up` work has been done, we should check back on this and, if necessary, update URL source caching to use the configured cache root if there is one.

cc @fibonacci1729

itowlson:
I'm pretty sure this is done.  #2377 is building on what's already there and it sure as heck looks like it's done.  Closing: reopen if I'm wrong.

PR #1205, opened 2023-02-27
fix: wire up cache

fibonacci1729:
~DEFINITELY WIP~

Implements #1188

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -23,6 +23,7 @@ pub const RUNTIME_CONFIG_FILE: &str = "RUNTIME_CONFIG_FILE";
 // Set by `spin up`
 pub const SPIN_LOCKED_URL: &str = "SPIN_LOCKED_URL";
 pub const SPIN_WORKING_DIR: &str = "SPIN_WORKING_DIR";
+pub const SPIN_REGISTRY_CACHE_DIR: &str = "SPIN_REGISTRY_CACHE_DIR";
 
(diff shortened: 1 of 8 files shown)


fibonacci1729:
@itowlson putting this up now; would appreciate early eyes. Thanks!

crates/trigger/src/cli.rs
```diff
@@ -96,6 +97,9 @@ where
     /// Load the application from the registry.
     #[clap(long = "from-registry", hide = true)]
     pub from_registry: bool,
+    // /// The root directory of the registry cache.
+    // #[clap(long = "registry-cache-root", hide = true)]
+    // pub registry_cache_root: Option<PathBuf>,
```
fibonacci1729, line 102: I was lazy here...It was easier to pass this from `spin up` using an environment variable....

itowlson:
Acknowledging that this is WIP but since you asked for eyes... the core logic looks sound (though Radu will have to comment on the registry loader code paths), but I was a bit puzzled by the UI.  I had thought the import of the request was to use the existing registry cache directory - which we must already know in order to do `spin up --from-registry` - for URL components as well.  So I'm not sure if we should be adding a new flag to `spin up` to set that cache directory; and if we are, it seems like it's no longer purely a registry cache, more a General Remote Stuff cache?

But I'm not very familiar with this stuff and may be the wrong pair of eyes.  My main role in all this was copying and pasting a TODO comment from an old PR into a new issue...

fibonacci1729:
cc @radu-matei

Issue #1207, opened 2023-02-27
add deprecation notice for bindle flags on spin up


michelleN:

itowlson:
closed via #1208

PR #1208, opened 2023-02-27
Print deprecation message on Bindle features

### LINKED ISSUE #1156: Prepare users for moving from Bindle to OCI
We have a goal of replacing Bindle with OCI for application distribution.  The phase out will be gradual, and we will give users enough time and information to prepare.

We should agree a timeframe in which to add deprecation messages to:

* `spin up --bindle`
* Bindle component references in Spin manifests
* `spin bindle *`
* Anything else?

itowlson:
Fixes #1156 and #1207.

The cases picked up in this are:

* `spin bindle push`
* `spin bindle run`
* `spin up --bindle`
* `spin up` from a local file with a parcel as a module source

**Reviewers please flag if you can think of any other paths that could result in us interacting with Bindle.**

An example of how it looks:

```
$ spin bindle push -f ./examples/http-rust/spin.toml --bindle-server http://localhost:8080/v1 --insecure
WARNING: Bindle support is deprecated and will be removed in a future version.
For remote applications, please use registry commands and features instead.
See https://developer.fermyon.com/spin/spin-oci for more details

pushed: spin-hello-world/1.2.3
```

The cases _should_ be exclusive, but the warning is guarded to print at most once _just in case_.  Because computers.

### DIFF
--- a/crates/loader/src/bindle/deprecation.rs
+++ b/crates/loader/src/bindle/deprecation.rs
@@ -1,0 +1,13 @@
+use std::sync::Once;
+
+/// Prints a notice that Bindle support is deprecated.
+pub fn print_bindle_deprecation() {
+    static PRINT: Once = Once::new();
+
+    PRINT.call_once(|| {
+        eprintln!("WARNING: Bindle support is deprecated and will be removed in a future version.");
+        eprintln!("For remote applications, please use registry commands and features instead.");
+        eprintln!("See https://developer.fermyon.com/spin/spin-oci for more details");
+        eprintln!();
+    });
+}
--- a/crates/loader/src/bindle/mod.rs
+++ b/crates/loader/src/bindle/mod.rs
@@ -7,6 +7,8 @@ mod assets;
 /// Configuration representation for a Spin application in Bindle.
 pub mod config;
 mod connection;
+/// Functions relating to Bindle deprecation.
+pub mod deprecation;
 /// Bindle helper functions.
 mod utils;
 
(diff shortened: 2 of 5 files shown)


crates/loader/src/bindle/deprecation.rs
```diff
@@ -0,0 +1,17 @@
...
+        return;
+    }
+
+    eprintln!("WARNING: Bindle support is deprecated and will be removed in a future version.");
+    eprintln!("For remote applications, please use registry commands and features instead.");
+    eprintln!("See https://developer.fermyon.com/spin/spin-oci for more details");
+    eprintln!();
+
+    _ = PRINTED.set(true);
+}
```
lann, line 17: Stdlib has `Once`:

```suggestion
use std::sync::Once;

/// Prints a notice that Bindle support is deprecated.
pub fn print_bindle_deprecation() {
    static PRINT: Once = Once::new();
    PRINT.call_once(|| {
        eprintln!("WARNING: Bindle support is deprecated and will be removed in a future version.");
        eprintln!("For remote applications, please use registry commands and features instead.");
        eprintln!("See https://developer.fermyon.com/spin/spin-oci for more details");
        eprintln!();
    });
}
```

itowlson:
Though... with deploy we retain bindle as an implementation detail.  So maybe we should not warn in that flow (this PR doesn't warn on deploy, even though it uses bindle under the hood).

crates/loader/src/bindle/deprecation.rs
itowlson: Thanks!  I thought there was something in stdlib but found only its `OnceCell` which was nightly only...

lann (approved):

kate-goldenring:
> Though... with deploy we retain bindle as an implementation detail. So maybe we should not warn in that flow (this PR doesn't warn on deploy, even though it uses bindle under the hood).

It makes sense to me to leave it around for now and naturally remove it when we rework deploy for OCI

kate-goldenring (approved):
Love the use of `Once`!

PR #1209, opened 2023-02-27
export SDK version and language functions per SIP 011

dicej:
Per https://github.com/fermyon/spin/blob/main/docs/content/sips/011-component-versioning.md, this commit adds `spin-sdk-version-0-10-pre0`, `spin-sdk-language-$LANGUAGE`, and `spin-sdk-commit-$COMMIT` exports to any module using the Rust or Go SDKs.

Regarding the version, we'll want to update it before and after every Spin release _and_ each time the ABI or API changes.  See the SIP for details.

Note that the `spin-sdk-language-$LANGUAGE` export is gated on a default-enabled feature for the Rust SDK, since the JS and Python SDKs will want to override it.

### DIFF
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -138,7 +138,7 @@ fn inner_check_supported_version(
     let version = Version::parse(spin_version)?;
     if !comparator.matches(&version) {
         if !version.pre.is_empty() {
-            eprintln!("You're using a pre-release version of Spin ({spin_version}). This plugin might not be compatible (supported: {supported_on}).  Continuing anyway.");
+            eprintln!("You're using a pre-release version of Spin ({spin_version}). This plugin might not be compatible (supported: {supported_on}). Continuing anyway.");
         } else if override_compatibility_check {
             eprintln!("Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Check overridden ... continuing to install or execute plugin.");
         } else {
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -137,7 +137,9 @@ fn inner_check_supported_version(
     })?;
     let version = Version::parse(spin_version)?;
     if !comparator.matches(&version) {
-        if override_compatibility_check {
+        if !version.pre.is_empty() {
+            eprintln!("You're using a pre-release version of Spin ({spin_version}). This plugin might not be compatible (supported: {supported_on}).  Continuing anyway.");
+        } else if override_compatibility_check {
             eprintln!("Plugin is not compatible with this version of Spin (supported: {supported_on}, actual: {spin_version}). Check overridden ... continuing to install or execute plugin.");
         } else {
             return Err(anyhow!(
--- a/docs/content/release-process.md
+++ b/docs/content/release-process.md
@@ -9,7 +9,8 @@ url = "https://github.com/fermyon/spin/blob/main/docs/content/release-process.md
 To cut a release of Spin, you will need to do the following:
 
 1. Create a pull request that changes the version number for your new version
-   (e.g. `0.9.0` becomes `0.9.1`)
+   (e.g. `0.10.0-pre0` could become either `0.9.1` for a patch release or
+   `0.10.0` for a "major" release)
    - Bump the version in Spin's `Cargo.toml`
    - Update SDK_VERSION in `templates/Makefile`
    - Check the docs for hard-coded version strings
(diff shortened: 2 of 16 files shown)


sdk/go/Makefile
```diff
@@ -41,10 +41,18 @@ GENERATED_OUTBOUND_REDIS = redis/outbound-redis.c redis/outbound-redis.h
 GENERATED_SPIN_REDIS     = redis/spin-redis.c redis/spin-redis.h
 GENERATED_KEY_VALUE      = key_value/key-value.c key_value/key-value.h
 
+SDK_VERSION_SOURCE_FILE  = sdk_version/sdk-version-go-original.c
+SDK_VERSION_DEST_FILES   = config/sdk-version-go.c http/sdk-version-go.c \
```
vdice, line 45: This seems like the only fragile bit, in terms of new areas of functionality being added to the go SDK (e.g. `postgres/`, etc.)  I'm assuming it's our best option currently.  If so, what might our strategy be to ensure they are added to the list here?

docs/content/release-process.md
```diff
@@ -13,11 +13,23 @@ To cut a release of Spin, you will need to do the following:
...
+     in sdk/rust/src/lib.rs and sdk/go/sdk-version-go.c
+      - e.g. `spin-sdk-version-0-10-pre0` becomes `spin-sdk-version-0-10`
+      - See [sips/011-component-versioning.md](sips/011-component-versioning.md)
+        for details
 1. Merge the PR created in #1 (Such PRs are still required to get approvals, so
    make sure you get signoff on the PR)
 1. Before proceeding, verify that the merge commit on `main` intended to be
    tagged is green, i.e. CI is successful
 1. Create a new tag with a `v` and then the version number (`v0.9.1`)
+1. For a major release (e.g. `0.9.0` to `0.10.0`), update the versions in
```
vdice, line 26: Appreciate adding these steps here until we have them automated.

We could potentially update the `spin-sdk-version-0-X(-pre0)` string in both SDKs via a few make targets, as we do for the [templates](https://github.com/fermyon/spin/blob/main/templates/Makefile).  (Corresponding automation [here](https://github.com/fermyon/spin/blob/main/.github/workflows/release.yml#L234-L271).)

Whichever approach we choose, perhaps we can track in the same ticket for adding `spin-sdk-commit-$HASH`...

itowlson: @vdice This is subtly different, though, I think, because it updates the strings to the _next_ anticipated release, rather than the _current_ release, so that canaries built _after_ the release provide informative failures.

itowlson: @vdice Sorry - I think I might have misread your comment.  Please ignore me!

To be clear, I would _love_ to see an automated roll forward after a release, and I'd like to see it applied to the app version in Cargo.toml as well.  This was always intended so that template tagging would work correctly with canaries, but never quite happened.

.gitignore
```diff
@@ -13,3 +13,4 @@ tests/testcases/*-generated
 spin-plugin-update.lock
 package-lock.json
 .spin
+sdk/go/*/sdk-version-go.c
```
lann, line 16: Should this move into `sdk/go/sdk_version/.gitignore`?

sdk/rust/src/lib.rs
```diff
@@ -11,6 +11,13 @@ pub mod key_value;
 /// Exports the procedural macros for writing handlers for Spin components.
 pub use spin_macro::*;
 
+#[export_name = "spin-sdk-version-0-10-pre0"]
```
lann, line 14: Would it be worth doing something like this?

```suggestion
#[export_name = concat!("spin-sdk-version-",
                        env!("CARGO_PKG_VERSION_MAJOR"),
                        "-",
                        env!("CARGO_PKG_VERSION_MINOR"),
                    )]
```

docs/content/release-process.md
vdice: @itowlson ah, the templates would also benefit from similar post-release roll forward automation?  If so, it'd seem natural to scope that in with the same automation work.

sdk/go/Makefile
dicej: Good thinking.  I can do a bit of Makefile function magic to make it more robust.

dicej: On second thought, I just added a (hopefully attention-grabbing) comment indicating what needs to be updated on the rare occasion we add a new directory to the Go SDK.

docs/content/release-process.md
dicej: I went ahead and automated this a bit, so now the only extra manual step is to update the version strings in `main` to `$MAJOR.$MINOR.$PATCH-pre0` after tagging the release.  Which we should definitely automate at some point.

sdk/rust/src/lib.rs
dicej: Yup, done.  Plus now we have `spin-sdk-commit-$COMMIT` as well.

dicej:
I _think_ I've addressed all the feedback so far.  Anything else?

itowlson (approved):
I tried it out in a new app and it works great, and no other concerns that I can think of.  But I'm probably not qualified to review it properly.  On the other hand, YOLO!

dicej:
Looks like CI is failing because the plugin compatibility check doesn't like the `-pre0` on the Spin version.  Investigating...

crates/plugins/src/manifest.rs
```diff
@@ -137,7 +137,9 @@ fn inner_check_supported_version(
     })?;
     let version = Version::parse(spin_version)?;
     if !comparator.matches(&version) {
-        if override_compatibility_check {
+        if !version.pre.is_empty() {
```
dicej, line 140: @kate-goldenring @karthik2804 The `semver` crate is very picky about pre-releases, which probably makes sense given how they are treated in the Semantic Versioning spec.  I'm working around it based on the idea that a pre-release has unknown compatibility constraints in general, so anything goes.  Let me know if you think there's a better approach.

lann (approved):

PR #1212, opened 2023-02-28
Reorganize spin-loader cache code

lann:
This moves OCI-specific code from spin-loader to spin-oci. With OCI-specific code removed from `Cache`, also move `spin_loader::oci::cache` to `spin_loader::cache`.


itowlson (approved):
HUZZAH

PR #1213, opened 2023-02-28
More OCI refactoring

lann:
More intermediate refactoring moving towards extracting OCI logic from CLI code.

The diff obscures whats going on here a bit:

- Split `spin-oci` into `client` and `auth` modules.
- Moved `is_oci_like` to `spin_oci::is_probably_oci_reference`


crates/oci/src/lib.rs
```diff
@@ -1,448 +1,39 @@
...
+    // go down this path if the file does not exist, and the chances of
+    // a user choosing a filename containing 'docker' THAT ALSO does not
+    // exist are A MILLION TO ONE...
+
+    // If it doesn't parse as a reference, it isn't a reference
+    let Ok(reference) = oci_distribution::Reference::try_from(maybe_oci) else {
+        return false;
+    };
+    // If it has an explicit tag, its probably a reference
+    if reference.tag() != Some("latest") || maybe_oci.ends_with(":latest") {
```
lann, line 29: @itowlson This whole function is mostly a verbose refactoring of your code, but it is tweaked slightly here to detect e.g. `actually/oci/reference:latest` as a reference.

crates/oci/src/lib.rs
```diff
@@ -1,448 +1,38 @@
...
+        return true;
+    }
+    // If it doesn't have an explicit registry it's hard to tell whether its a
```
itowlson, line 31: This is definitely clearer than my original - thanks.  I would clarify this comment, though, maybe:

If the original string doesn't have a registry, the parser inserts Docker's registry.  Try to detect that, and if it happened, it's probably not a reference.

crates/oci/src/lib.rs
```diff
@@ -1,448 +1,38 @@
...
+/// an OCI reference.
+pub fn is_probably_oci_reference(maybe_oci: &str) -> bool {
+    // A relative file path such as foo/spin.toml will successfully
+    // parse as an OCI reference, because the parser infers the Docker
+    // registry and the `latest` version.  So if the registry resolves
+    // to Docker, but the source *doesn't* contain the string 'docker',
+    // we can guess this is likely a false positive.
+    //
+    // This could be fooled by, e.g., dockerdemo/spin.toml.  But we only
+    // go down this path if the file does not exist, and the chances of
```
itowlson, line 19: Now this is moved out of private `up` code, there's no longer a sense of "this gets called if a string doesn't exist in the file system" - the knowledge that the string is _not_ a file is, I think, important to the heuristic.  So if we move it here (and I agree it's good to concentrate all Spin's understanding of OCI in one place), we maybe need to reword the documentation (and this comment) to clarify that that context...?

itowlson (commented):
HUZZAH but also NITPICK

radu-matei (approved):

Issue #1210, opened 2023-02-28
`http-php` template not working with latest spin


rajatjindal:
while trying to add tests for `http-php`, it seems it is not working. 

steps followed (Refer below for full logs of commands run and logs)

- installed latest spin canary `spin 0.9.0 (77985cc 2023-02-28)` on `apple silicon` (didn't tried on other one)
- installed latest templates
- create new app using template `http-php`
- cd to new app folder and run `spin up`
- hit the endpoint `http://127.0.0.1:3000/`, observe following error on spin process side

```bash
2023-02-28T07:39:34.061058Z ERROR spin_http: Error processing request: error while executing at wasm backtrace:
    0: 0x34ca86 - <unknown>!<wasm function 6022>
    1: 0x34e793 - <unknown>!<wasm function 6058>
    2: 0x361e09 - <unknown>!<wasm function 6226>
note: using the `WASMTIME_BACKTRACE_DETAILS=1` environment variable may show more debugging information

Caused by:
    exit with invalid exit status outside of [0..126) 
```
<details>
  <summary>Detailed logs</summary>

```

➜  testcases git:(more-tests-2) ✗ /usr/local/bin/spin-canary --version                 
spin 0.9.0 (77985cc 2023-02-28)

➜  testcases git:(more-tests-2) ✗ /usr/local/bin/spin-canary  template install --git https://github.com/fermyon/spin --branch main --upgrade
Copying remote template source
Installing template redis-rust...
Installing template static-fileserver...
Installing template http-grain...
Installing template http-swift...
Installing template http-php...
Installing template http-c...
Installing template redirect...
Installing template http-rust...
Installing template http-go...
Installing template http-zig...
Installing template http-empty...
Installing template redis-go...
Installed 12 template(s)

+------------------------------------------------------------------------+
| Name                Description                                        |
+========================================================================+
| http-c              HTTP request handler using C and the Zig toolchain |
| http-empty          HTTP application with no components                |
| http-go             HTTP request handler using (Tiny)Go                |
| http-grain          HTTP request handler using Grain                   |
| http-php            HTTP request handler using PHP                     |
| http-rust           HTTP request handler using Rust                    |
| http-swift          HTTP request handler using SwiftWasm               |
| http-zig            HTTP request handler using Zig                     |
| redirect            Redirects a HTTP route                             |
| redis-go            Redis message handler using (Tiny)Go               |
| redis-rust          Redis message handler using Rust                   |
[...]

PR #1215, opened 2023-03-01
add e2e tests for php, swift, oci, longevity

rajatjindal:
This PR covers following:

- added tests for `php`, `python` and `swift` templates. `php` test is disabled right now due to #1210 
- added test for `oci`
- added a `longevity` test. Idea is we check-in `wasm` files and they should continue to work with future version's of `spin` and `Fermyon Cloud`


tests/testcases/longevity-apps-test/README.md
```diff
@@ -0,0 +1,12 @@
+## longevity-apps-test
+
+longevity test is to ensure `wasm` file(s) compiled with a version of `Spin` continues to work with runtime of future version of `Spin`.
```
vdice, line 3: Wondering how we anticipate handling failures here while we're pre-1.0?  Perhaps on a case-by-case basis?  (For example, when a breaking change is necessary and/or the preferred option -- I suppose the PR contributor would re-generate the wasm file(s) in question?)

rajatjindal: from what I gathered from discussion I had with @itowlson, the `wasm` module compiled once should continue to work with future spin runtime. 

but in any case, I will add instructions on how to generate these wasm files.

tests/testcases/longevity-apps-test/README.md
```diff
@@ -0,0 +1,12 @@
...
+longevity test is to ensure `wasm` file(s) compiled with a version of `Spin` continues to work with runtime of future version of `Spin`. 
+
+The current wasm files are created using following templates with `Spin v0.9.0`
+
+- http-go
+- http-rust
+- http-js
+- http-ts
+
+The `wasm` files are built using `spin build` and copied over here for validation.
```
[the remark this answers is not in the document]
rajatjindal: I'll include the version number in wasm file. and will also add additional msg to inform details.

tests/testcases/longevity-apps-test/README.md
rajatjindal: added the instructions to readme.

tests/testcases/longevity-apps-test/README.md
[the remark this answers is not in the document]
rajatjindal: this is how it looks now:

```
thread 'spinup_tests::longevity_apps_works' panicked at ':

app using a wasm module built by 'spin version 0.9.0 (a99ed51 2023-02-16)' no longer works with "spin 0.10.0-pre0 (492e509 2023-03-01)\n"
Error: assertion failed: `(left == right)`
  left: `"Hello from JS-S"`,
 right: `"Hello from JS-SDK"`: body assertion failed', tests/testcases/mod.rs:839:13
test spinup_tests::longevity_apps_works ... FAILED
```

tests/testcases/mod.rs
```diff
@@ -635,4 +735,108 @@ pub mod all {
...
+            .unwrap();
+
+        let version = match spin::version() {
+            Ok(version) => version,
+            Err(err) => {
+                println!("{}", err);
+                "".to_string()
+            }
+        };
+        let msg = format!("app using a wasm module built by 'spin version 0.9.0 (a99ed51 2023-02-16)' no longer works with {}", version);
```
vdice, line 837: Can we avoid hard-coding the version or perhaps pull it from some canonical source (variable or const?) that is known/expected to be updated if/when the longevity wasm files are built with a different version?  Or, we can iterate/brainstorm a bit on this if preferred -- but minimally for this PR perhaps add a mention in the longevity-apps-test/README.md to update this string on such events.

tests/testcases/longevity-apps-test/README.md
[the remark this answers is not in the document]
vdice: Nice!  Very clear.

tests/testcases/longevity-apps-test/README.md
itowlson: I like the idea of keeping "old" Wasm files and applications around.  I agree that if we were on early 0.x it would be a pain handling failures from known breaking changes.  But now we are close enough to stabilisation that any breaking change from here on would be worth flagging up.  Once we have 1.0 we could biff the pre-1.0 builds and only retain the 1.x cycle; but for now apps built on 0.9 and 0.10 should likely be keep working going into the 1.0 release cycle.

tests/testcases/mod.rs
rajatjindal: Yes, this is a good call. I don't know if we can fetch that info dynamically, but I have added a const within the test case with that value for now.

tests/spinup_tests.rs
```diff
@@ -51,6 +51,22 @@ mod spinup_tests {
         testcases::all::http_js_works(CONTROLLER).await
     }
 
+    #[tokio::test]
+    async fn http_python_works() {
+        testcases::all::http_python_works(CONTROLLER).await
+    }
+
+    #[tokio::test]
+    #[ignore] // https://github.com/fermyon/spin/issues/1210
```
vdice, line 60: I _think_ it should function if the '/index.php' path is used (https://github.com/fermyon/spin/issues/1210#issuecomment-1451129360)?

Anyways, follow-up material when ready!

vdice (approved):
Looks great to me!  Learning some Rust macro approaches :)

tests/spinup_tests.rs
rajatjindal: yes, i want to update the template for php to make it work out of the box without having to have explicit `index.php`. in a follow up PR

rajatjindal:
> Looks great to me! Learning some Rust macro approaches :)

yeah this one was fun to investigate and implement. I actually pulled it in a [separate repo](https://github.com/rajatjindal/ensure-assertions) also. will publish a crate when I have some bandwidth.

PR #1216, opened 2023-03-01
Add a log statement to tell the user to look at the logs if readiness is not achieved

calebschoepp:
This is a hacky fix until we can expose more specific errors from LHC and structured logging from cloud.

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -295,6 +295,7 @@ impl DeployCommand {
                 &app_base_url,
                 &bindle_id.version_string(),
                 self.readiness_timeout_secs,
+                Destination::Platform,
             )
             .await;
             print_available_routes(&app_base_url, &http_config.base, &cfg);
@@ -418,6 +419,7 @@ impl DeployCommand {
                 &app_base_url,
                 &bindle_id.version_string(),
                 self.readiness_timeout_secs,
+                Destination::Cloud(connection_config.url),
             )
             .await;
             print_available_routes(&app_base_url, &http_config.base, &cfg);
@@ -656,7 +658,17 @@ async fn check_healthz(base_url: &Url) -> Result<()> {
 
 const READINESS_POLL_INTERVAL_SECS: u64 = 2;
 
-async fn wait_for_ready(app_base_url: &Url, bindle_version: &str, readiness_timeout_secs: u16) {
+enum Destination {
+    Cloud(String),
+    Platform,
+}
+
+async fn wait_for_ready(
+    app_base_url: &Url,
+    bindle_version: &str,
+    readiness_timeout_secs: u16,
+    destination: Destination,
+) {
     if readiness_timeout_secs == 0 {
         return;
     }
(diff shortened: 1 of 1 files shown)


src/commands/deploy.rs
```diff
@@ -693,6 +700,9 @@ async fn wait_for_ready(app_base_url: &Url, bindle_version: &str, readiness_time
         if start.elapsed() >= readiness_timeout {
             println!();
             println!("Application deployed, but Spin could not establish readiness");
+            if deployed_to_cloud {
+                println!("The Fermyon Cloud dashboard may have some logs that explain why readiness could not be established")
```
itowlson, line 704: Link maybe?

Also maybe phrase as more of a concrete action a la "Check the Fermyon Cloud (LINK) dashboard to see the application status."

src/commands/deploy.rs
```diff
@@ -295,6 +295,7 @@ impl DeployCommand {
                 &app_base_url,
                 &bindle_id.version_string(),
                 self.readiness_timeout_secs,
+                false,
```
itowlson, line 298: Bit of a nit I guess, but these constants are hard to understand at the call site.  If it were me, and if I were feeling both energetic and smugly righteous, I would make an enum a la `Destination::{Cloud|Platform}`.  But I completely sympathise if you go 'eh not worth it.'

itowlson (changes requested):
The request changes is for the message text and link; the other is a nit.

src/commands/deploy.rs
calebschoepp: How convenient. You've caught me in an energetic and smugly righteous mood. 😄

calebschoepp: Random booleans are the death of a beautiful codebase.

src/commands/deploy.rs
```diff
@@ -656,7 +658,17 @@ async fn check_healthz(base_url: &Url) -> Result<()> {
...
+enum Destination {
+    Cloud(String),
+    Platform,
+}
+
+async fn wait_for_ready(
+    app_base_url: &Url,
+    bindle_version: &str,
+    readiness_timeout_secs: u16,
+    destination: Destination,
```
lann, line 670: I probably would have gone with:

```suggestion
    dashboard_url: Option<&str>,
```

(but this is fine too)

src/commands/deploy.rs
itowlson: PREACH IT

lann (approved):

itowlson (approved):
Looks good - I like using the enum to pass the URL!

src/commands/deploy.rs
lann:
```rust
rand::random<bool>()
```

src/commands/deploy.rs
itowlson: I prefer the enum because when I read the call site I see `Destination::Platform` instead of the rather uninformative `None`.  But it's subjective for sure.

lann: For better or worse Rust has decided not to use labeled function arguments :shrug:. It would be really helpful if Github could inline rust-analyzer hints into PR diffs.

PR #1217, opened 2023-03-01
Add SIP for signing Spin releases

radu-matei:
[Rendered SIP](https://github.com/radu-matei/spin/blob/sip012-signing-spin-releases/docs/content/sips/012-signing-spin-releases.md)


docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+certificate used to sign, and the signature itself), then validate the signature
+using `cosign`:
+
+```bash
+$ tree .
+├── artifact
+├── artifact.sig
+└── crt.pem
+
+$ cosign verify-blob \
```
lann, line 168: Cosign appears to have built-in support for github oidc now via `--certificate-github-workflow-*` flags. (which I haven't actually tested...)

radu-matei: `--certificate-github-workflow-sha` appears to be part of the signature, not sure about others, haven't been able to validate them.

radu-matei:
cc @vdice since you are the most familiar with the release process.

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+signatures auditable as well.
+
+This process uses a few central pieces that are assumed trusted:
+
+- an OIDC provider (such as GitHub)
+- Fulcio — free root certificate authority that issues temporary certificates
+bound to an OIDC identity, published to Rekor
+- Rekor — transparency and timestamp service, provides a ledger that can be
+audited
+- `cosign sign` — the CLI that connects an OIDC identity to Fulcio to generate the
```
vdice, line 48: Trying to get an idea of the external services involved when invoking this command...  Fulcio?  Others?

Perhaps a rare event but what is our approach when we're unable to sign due to external service issues?

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+
+Created: March 2, 2023
+
+## Background
+
+Signing release artifacts for software offers a way for consumers of the software
+to verify the integrity of the package they downloaded. The signature should
+offer two guarantees: the author of the artifact is indeed the one expected by
+the consumer, and the content of the artifact has not been tampered with since
+its creation.
```
itowlson, line 20: I don't have any comment on the technology, which I realise is the focus of the SIP, but I'd find it really useful to understand what guarantees it provides, what attacks it mitigates, what malicious actions it allows a consumer to detect.  The proposal mentions "validate the integrity" but clearly means something more than file integrity (which is presumably guaranteed with the digest).  Is it about the attack of someone posting a false file somewhere and claiming it to be Spin?

docs/content/sips/012-signing-spin-releases.md
itowlson, line 168: For me, it would be super useful to have the user experience of this clearly called out - the "GitHub actions example" is not a natural place to look for it _grin_

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+    # write the output certificate to a file that will be later added to the release
+    --output-certificate out/spin-v1.0.0-linux-amd64-keyless.pem \
+    # write the signature to a file that will be later added to the release
+    --output-signature out/spin-v1.0.0-linux-amd64-keyless.sig spin- \
+    # skip interactive confirmation
+    --yes
+```
+
+- when uploading the assets to the new release, the certificate and signatures
+need to be added side-by-side with the actual artifact
```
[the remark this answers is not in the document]
radu-matei: It depends on what we want to sign — the archive containing everything, or just the binary. 

Since we already compute the content digest _of the archive_, it might make sense to sign that. If we sign the archive, then the signature and certificate would be distributed as separate release artifacts.

If we only sign the binary, then we could distribute the signature and certificate inside the archive.

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+      - name: Upload assets as GitHub artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: artifact
+          path: out/*
+      ...(continue with adding artifacts to the release)
+```
+
+After the action is executed, the new artifact is released and its signature
+published using Sigstore. To validate the signature, a user will then need to
```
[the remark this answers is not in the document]
radu-matei: Yes, a lot of this requires instructions for how users would interact and perform checks on the signatures.

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+        with:
+          cosign-release: v2.0.0
+
+      - name: Create an artifact
+        run: |
+          mkdir out
+          echo 'hello world' > out/artifact
+          
+      - name: Sign the artifact with GitHub OIDC token
+        run: cosign sign-blob --output-certificate out/crt.pem --output-signature out/artifact.sig out/artifact --yes
```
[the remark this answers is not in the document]
radu-matei: Yes, each `<os>-<arch>` build (whether that is directly the Spin binary or the archive containing the readme and license) would have a signature and certificate associated.

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+the verification step also validates that not only the artifact has not been
+tampered with, but the identity used when performing the signature is associated
+with the GitHub repository and project.
+
+Attempting to verify the signature of a different artifact will result in an error:
+
+```bash
+$ cosign verify-blob \
+    --signature artifact.sig --certificate crt.pem \
+    --certificate-identity https://github.com/fermyon/spin/.github/workflows/release-sign.yml@refs/heads/main \
```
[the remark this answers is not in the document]
radu-matei: The documentation and release notes would contain optional instructions on getting the certificate and signature, and the command required to perform the signature validation.

docs/content/sips/012-signing-spin-releases.md
radu-matei: Fulcio and Rekor are both public good services that this mechanism uses while performing a signature.

If either is down, our signature process will fail.
Given that fairly large projects such as Kubernetes have already adopted this process, and that this has been marked as stable, my assumption is that the services should be relatively stable.

While we should sign canary builds, it's only stable releases that we are asking users to trust. Given that, rare events when we would be unable to sign stable releases should be indeed _rare_.
(if our experience is different as we test this, happy to re-evaluate this)

docs/content/sips/012-signing-spin-releases.md
radu-matei: Besides the file integrity (which is now verifiable using a transparency log, rather than just using a file), a user can validate that the build was actually generated by the `fermyon/spin` repo, through a GitHub action workflow they can inspect at a given SHA or tag, and that us or someone else has not replaced the binary since the GitHub infrastructure built it at the time of the SHA or tag.


Besides this guarantee, this can be later expanded to include even more attestations (such as [SLSA](https://slsa.dev/provenance/v0.2), which could include even more provenance data).

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,206 @@
...
+    --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+    fake-artifact
+
+Error: verifying blob [fake-artifact]: searching log query: [POST /api/v1/log/entries/retrieve][400] searchLogQueryBadRequest  &{Code:400 Message:unmarshalling entry: verifying signature: invalid signature when validating ASN.1 encoded signature}
+```
+
+Attempting to verify a signature using a different identity also results in an error:
```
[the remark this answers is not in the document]
radu-matei: Not sure if that's necessarily useful. If you modify them, they will most likely be invalid when decoding.

docs/content/sips/012-signing-spin-releases.md
radu-matei: Users should have this in a documentation section on the developer website.
For the SIP, I renamed the section.

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,210 @@
...
+└── crt.pem
+
+$ cosign verify-blob \
+    --signature artifact.sig --certificate crt.pem \
+    # the identity for the certificate is tied to the official repository of the Spin project
+    --certificate-identity https://github.com/fermyon/spin/.github/workflows/release-sign.yml@refs/heads/main \
+    --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+    artifact
+
+Verified OK
```
michelleN, line 179: would be nice to have a spin self update command or plugin that handles this verification part for the user if possible in the future.

michelleN (approved):

docs/content/sips/012-signing-spin-releases.md
```diff
@@ -0,0 +1,210 @@
...
+
+Signing release artifacts for software offers a way for consumers of the software
+to verify the integrity of the package they downloaded. The signature should
+offer two guarantees: the author of the artifact is indeed the one expected by
+the consumer, and the content of the artifact has not been tampered with since
+its creation.
+
+The goal of this SIP is to define the tools used for generating signatures for
+the Spin project, and to provide a standard process for signing release artifacts
+for Spin and related projects.
```
kate-goldenring, line 24: Should we consider changing the way we expect plugins to be signed as well? Or are SHA checksums sufficient?

PR #1227, opened 2023-03-03
Add outbound Postgres bindings to Go SDK

jpflueger:
I noticed that there were no bindings for outbound Postgres in the go sdk. I was able to generate bindings using wit-bindgen 0.3.0 by wrapping `outbound-pg.wit` into a temporary wit file using an interface for the rdbms-types.wit and a world for the query and execute functions. Most of the binding code was generated but I did make some modifications for better error handling and split out the functions for readability. I wanted to submit these changes as a starting point. Happy to rework whatever you suggest as this is my first time generating bindings and I'm not super familiar with C or Go.


jpflueger:
Hmm. I must have done something funky to get those merge conflicts. I'll try to resolve and re-push my changes.

jpflueger:
> Hmm. I must have done something funky to get those merge conflicts. I'll try to resolve and re-push my changes.

Weirdness caused by me not signing commits so I had to rebase with a merge conflict. Fixed locally and force-pushed my branch to update, sorry for the churn.

sdk/go/Makefile
```diff
@@ -27,19 +27,21 @@ build-examples: $(EXAMPLES_DIR)/http-tinygo/main.wasm
...
-GENERATED_OUTBOUND_REDIS = redis/outbound-redis.c redis/outbound-redis.h
-GENERATED_SPIN_REDIS     = redis/spin-redis.c redis/spin-redis.h
-GENERATED_KEY_VALUE      = key_value/key-value.c key_value/key-value.h
+GENERATED_SPIN_CONFIG     = config/spin-config.c config/spin-config.h
+GENERATED_OUTBOUND_HTTP   = http/wasi-outbound-http.c http/wasi-outbound-http.h
+GENERATED_SPIN_HTTP       = http/spin-http.c http/spin-http.h
+GENERATED_OUTBOUND_REDIS  = redis/outbound-redis.c redis/outbound-redis.h
+GENERATED_SPIN_REDIS      = redis/spin-redis.c redis/spin-redis.h
+GENERATED_KEY_VALUE       = key_value/key-value.c key_value/key-value.h
+GENERATED_OUTBOUND_PG     = postgres/outbound-pg.c postgres/outbound-pg.h
```
dicej, line 44: Would you please add `postgres/sdk-version-go.c` to `SDK_VERSION_DEST_FILES` below?  That helps ensure the SDK version information is embedded in components built with the SDK.

examples/tinygo-outbound-pg/README.md
```diff
@@ -0,0 +1,18 @@
+# Spin component with outbound postgres in TinyGo
```
dicej, line 1: Would you mind adding instructions for using `wit-bindgen` 0.3.0 to re-generate the bindings if necessary (e.g. because we add variant cases, new functions, etc. to outbound-pg.wit)?

dicej (commented):
Looks great overall -- just a couple of requests.

sdk/go/postgres/db_value.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	return n.val.([]uint8)
+}
+
+func (n *DbValue) SetBinary(v []uint8) {
+	n.val = v
+	n.kind = DbValueKindBinary
+}
+
+func DbValueDbNull() DbValue {
```
itowlson, line 263: Should we have something like an `IsNull` function?  What sort of experience should we expect for loading a value from a NULLable column, especially if the developer wants graceful failure rather than panic?

(Again this is me being ignorant of How Things Are in Go.)

sdk/go/postgres/pg_types.go
```diff
@@ -0,0 +1,119 @@
...
+	return PgError{kind: PgErrorKindSuccess}
+}
+
+func PgErrorConnectionFailed(v string) PgError {
+	return PgError{kind: PgErrorKindConnectionFailed, val: v}
+}
+
+func (n PgError) GetConnectionFailed() string {
+	if g, w := n.Kind(), PgErrorKindConnectionFailed; g != w {
+		panic(fmt.Sprintf("Attr kind is %v, not %v", g.String(), w.String()))
```
itowlson, line 47: Again panics seem quite harsh, especially given that the developer can't expect errors the way they can data types... is the expectation that the app would call `Kind()` before calling the relevant Get function?

itowlson (commented):
I'm a Go ignoramus so please ignore me if I'm completely off beam

sdk/go/postgres/db_value.go
```diff
@@ -0,0 +1,269 @@
...
+	return n.kind
+}
+
+func DbValueBoolean(v bool) DbValue {
+	return DbValue{kind: DbValueKindBoolean, val: v}
+}
+
+func (n DbValue) GetBoolean() bool {
+	if g, w := n.Kind(), DbValueKindBoolean; g != w {
+		panic(fmt.Sprintf("Attr kind is %v, not %v", g.String(), w.String()))
```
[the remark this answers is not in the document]
dicej: I think this is just how the tinygo binding generator in wit-bindgen works.  The idea is that the correct way to get the value is ask what the type is first and then extract the value using the appropriate getter.  It's like poor-man's pattern matching.  So if the app developer skips the first step and just assumes the type, it's a programmer error, hence the panic.  By that reasoning, making the developer check for errors in the second step would be redundant.

Mossaka:
Hey I am glad that you were able to use the `wit-bindgen-go` generator to generate the bindings! I would greatly appreicate any feedback to improve the binding's ergonomics as I didn't pay too much attention to it as I was trying to get the it done as soon as possible.

radu-matei:
Hey, @Mossaka — I think this is still using the C bindings generator.

sdk/go/postgres/db_value.go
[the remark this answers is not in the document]
jpflueger: Agreed with what Joel said here. With the lack of generics in Go usually this method is accompanied by a check like this:

```
if dbValue.Kind() == DbValueKindBoolean {
  dbBool := dbValue.GetBoolean()
}
```

My understand of the go idioms is that a tuple is essentially used like a `Result<T>` in rust where a function like `func x() (bool, Error)` is equivalent to `fn x() -> Result<bool, Error>` in the Rust world.

All that being said, this is generated code so our request will need to go to wit-bindgen.

sdk/go/postgres/pg_types.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return n.val.(string)
+}
+
+func (n *PgError) SetConnectionFailed(v string) {
```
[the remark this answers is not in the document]
jpflueger: The set functions could be private yes but this is just how wit-bindgen works as it stands today. I'm unclear on how we could customize this behavior. We could wrap the generated code better but that would incur a maintenance cost.

sdk/go/postgres/db_value.go
itowlson: Let me clarify what I'm driving at.  Consider a People table with two (interesting) columns, Age (integer) and MiddleName (string).  Both are NULLable in the database (to reflect unknown age and people who don't have a middle name).

I want to query for the set of People and process that list, for example rendering to HTML (printing `unknown` for NULL ages and `-` for no middle name), or marshalling to JSON (where I guess I'd omit the NULL fields), but for generality going via some intermediate Go structure.

How (generally) do I do this?

If memory serves, for the intermediate structure to allow `nil`, its values would have to be pointers e.g. `{ Age *int, MiddleName *string }`.  Is that right?  If so, what is the outline for populating them, something like:

```
var age
if v.Kind() == DbNull {
  age = nil
} else {
  age = &v.GetInt()
}

person := Person { Age: age, ... }
```

Is that right (modulo Go syntax!), and if so, does it seem like what a Go programmer would expect to right?

To be clear, I am _not_ suggesting it's not!  I'm not familiar with Go idioms - for all I know this is _exactly_ what a Go programmer would find familiar.  I'm just confirming that we are happy with whatever the pattern for nullable columns is.

Hope this makes more sense now - sorry for the somewhat muddled question first time around!

examples/tinygo-outbound-pg/README.md
jpflueger: @dicej I added instructions to the readme in the `sdk/go` folder. Let me know if that is too hacky to include :)

sdk/go/Makefile
jpflueger: Will do, I need to rebase onto v1.0 branch first and then I'll add it to avoid any merge conflicts.

sdk/go/postgres/db_value.go
jpflueger: > If memory serves, for the intermediate structure to allow nil, its values would have to be pointers e.g. { Age *int, MiddleName *string }. Is that right?

I do believe that is right, yes. When I created a struct without pointers, the struct is initialized with default values (i.e. 0 for int, "" for string, etc.)

Issue #1175, opened 2023-02-17
Add `spin up --app-state-dir`


lann:
We're starting to move some per-app local state (logs, default KV store) into a `.spin` dir adjacent to the `spin.toml` file, which we'll start calling the "app state" directory. This should be configurable, with different defaults between "local" (`spin.toml`) and "remote" (Bindle, OCI) apps:

- Add a `--app-state-dir` flag to `spin up`
  - Local apps: defaults to `.spin/` relative to the local `spin.toml`
  - Remote apps: defaults to being unset

The value of the app state dir will impact a couple of existing features:
- If `--app-state-dir` is set (local default):
  - App logs (stdout/err) will be written to `<app-state-dir>/logs/...` (when #1111 is done)
  - Default KV store will be `<app-state-dir>/sqlite_key_value.db`
- If `--app-state-dir` is unset (remote default):
  - App logs will not be written to disk (still written to stdio by default)
  - Default KV store will be in-memory (not persisted)

itowlson:
For the local case, it seems like there would be no way to unset `--app-state-dir` - is that correct?  (I don't see a lot of value to doing so, but just want to be sure that we're making that decision intentionally.)

lann:
> way to unset `--app-state-dir`

We can probably safely interpret `--app-state-dir ''` as "unset" which would at least be useful for testing this code.

lann:
Rethinking my own name suggestion, I'm not sure "state" really captures what we want to use this for (logs aren't really "state"). Maybe `--app-data-dir`? Just `--data-dir`?

PR #1230, opened 2023-03-03
chore: Refactor `spin up` OCI support

lann:
The main change here is to prepare the file mounts before calling the trigger executor, like local and bindle apps do.

### DIFF
--- a/crates/oci/src/loader.rs
+++ b/crates/oci/src/loader.rs
@@ -1,0 +1,126 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::{anyhow, ensure, Context, Result};
+use oci_distribution::Reference;
+use reqwest::Url;
+use spin_app::locked::{ContentPath, ContentRef, LockedApp, LockedComponent};
+use spin_loader::cache::Cache;
+
+use crate::{Client, ORIGIN_URL_SCHEME};
+
+/// OciLoader loads an OCI app in preparation for running with Spin.
+pub struct OciLoader {
+    working_dir: PathBuf,
+}
+
+impl OciLoader {
+    /// Creates a new OciLoader which builds temporary mount directory(s) in
+    /// the given working_dir.
+    pub fn new(working_dir: impl Into<PathBuf>) -> Self {
+        let working_dir = working_dir.into();
+        Self { working_dir }
+    }
+
+    /// Loads a LockedApp with the given OCI client and reference.
+    pub async fn load_app(&self, client: &mut Client, reference: &str) -> Result<LockedApp> {
+        // Fetch app
+        client.pull(reference).await.with_context(|| {
+            format!("cannot pull Spin application from registry reference {reference:?}")
+        })?;
+
+        // Read locked app
+        let lockfile_path = client
+            .lockfile_path(&reference)
+            .await
+            .context("cannot get path to spin.lock")?;
+        let locked_content = tokio::fs::read(&lockfile_path)
+            .await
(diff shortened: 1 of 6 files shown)


crates/trigger/src/cli.rs
```diff
@@ -175,14 +166,10 @@ where
         match &self.log {
             Some(l) => Some(l),
             None => {
-                if !self.from_registry {
-                    if let Ok(dir) = std::env::var(SPIN_STATE_DIR) {
-                        let s = PathBuf::from(dir).join("logs");
-                        return Some(s);
-                    } else {
-                        tracing::info!("Unable to retrieve SPIN_STATE_DIR environment variable to persist logs",);
```
lann, line 183: @michelleN FYI I felt OK removing this because the logic will change significantly with https://github.com/fermyon/spin/issues/1175

crates/oci/src/loader.rs
```diff
@@ -0,0 +1,127 @@
...
+        Self { working_dir }
+    }
+
+    /// Loads a LockedApp with the given OCI client and reference.
+    pub async fn load_app(&self, client: &mut Client, reference: &str) -> Result<LockedApp> {
+        // Fetch app
+        client
+            .pull(reference)
+            .await
+            .context("cannot pull Spin application from registry")?;
```
itowlson, line 30: It may seem like it should be obvious but I would like to include the reference in the context.  I am thinking of cases where we have incorrectly inferred a file path to be an OCI reference; rather than just telling the user "cannot pull Spin application" it might help to say "cannot pull x/y/z from registry" to prompt them to look at the string and ask "why is it trying to pull that" rather than "why is my application in a registry".  (Of course higher levels should help in the diagnosis too!)

crates/oci/src/loader.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        // Read locked app
+        let app_path = client
```
itowlson, line 33: Suggestion: `lockfile_path` to be more specific?

crates/oci/src/loader.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Loads a LockedApp with the given OCI client and reference.
+    pub async fn load_app(&self, client: &mut Client, reference: &str) -> Result<LockedApp> {
+        // Fetch app
+        client
+            .pull(reference)
+            .await
+            .context("cannot pull Spin application from registry")?;
+
+        // Read locked app
```
itowlson, line 32: I don't think you use the path or content again after this section - suggest extracting a `read_locked_app` function

crates/trigger/src/cli.rs
```diff
@@ -165,7 +155,8 @@ where
         let mut builder = TriggerExecutorBuilder::new(loader);
         self.update_wasmtime_config(builder.wasmtime_config_mut())?;
 
-        let logging_hooks = StdioLoggingTriggerHooks::new(self.follow_components(), self.log_dir());
+        let logging_hooks =
+            StdioLoggingTriggerHooks::new(self.follow_components(), self.log.clone());
```
itowlson, line 159: Is the change from `self.log_dir()` to `self.log` correct here?  I would have thought we wanted to use the new log directory logic, but maybe that is now addressed somewhere else?

itowlson (commented):
Delighted to see this: farewell, `--from-registry`, you will not be missed.  All looks great except for one concern about the default log directory, but probably that's me missing something...?

radu-matei (approved):

PR #1231, opened 2023-03-05
Fix registry login error if directory does not exist

### LINKED ISSUE #1128: Add some document for installing spin on k8s
Add some document for installing spin on k8s

### LINKED ISSUE #1228: spin registry login: no such file or directory
When the directory used by `spin registry login` doesn't already exist on the user's machine, the command will fail.

(Below I emulate non-existence via moving the dir to another location; but consider ephemeral environments, eg GitHub Action runners, wherein this directory won't have been created and this command is the first to attempt to access it.)

```
$ spin --version
spin 0.10.0 (e84ac6c 2023-03-02)

$ mv /Users/vdice/Library/Application\ Support/fermyon /Users/vdice/Library/Application\ Support/fermyon.bak

$ echo ${GITHUB_TOKEN} | spin registry login ghcr.io -u vdice --password-stdin
Error: cannot log in to the registry

Caused by:
    0: cannot save authentication file "/Users/vdice/Library/Application Support/fermyon/registry-auth.json"
    1: No such file or directory (os error 2)
```

itowlson:
Fixes #1228

### DIFF
--- a/crates/oci/src/auth.rs
+++ b/crates/oci/src/auth.rs
@@ -88,6 +88,11 @@ impl AuthConfig {
     }
 
     async fn save(&self, p: &Path) -> Result<()> {
+        if let Some(parent_dir) = p.parent() {
+            tokio::fs::create_dir_all(parent_dir)
+                .await
+                .with_context(|| format!("Failed to create config dir {}", parent_dir.display()))?;
+        }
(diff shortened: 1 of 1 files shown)


radu-matei (approved):
LGTM, thanks for fixing this!

Issue #1233, opened 2023-03-06
Change spin-config EnvProvider prefix to SPIN_CONFIG_


lann:
The current prefix is `SPIN_APP_*`, e.g. `SPIN_APP_FOO` sets the `foo` config value.

I'm not sure what I was thinking here. It isn't obvious and conflicts with other obvious env var names we would want to use.

I'd like to change it to the much more obvious `SPIN_CONFIG_*` before 1.0, which is a simple change to spin + docs.

lann:
Might need to change a couple of sample apps here and there.

lann:
Doc update: https://github.com/fermyon/developer/pull/424

PR #1234, opened 2023-03-06
config: Change EnvProvider prefix from SPIN_APP_ to SPIN_CONFIG_

lann:
ref #1233


itowlson (approved):
We have no way of knowing to what extent the config system is used; I'm wondering if for belt and braces we should temporarily accept `SPIN_APP_` variables with a deprecation warning.  It would complicate things rather though...!

lann:
I have a pretty strong sense that it is very rarely used. Config isn't functional in cloud currently and some github searching showed no hint of it being used outside of Fermyon.

PR #1235, opened 2023-03-07
Validate allowed key-value stores

### LINKED ISSUE #1232: Catch invalid key/value store at build time
Currently, Spin's SQLite key value feature only supports a single `"default"` store per application. The implementation [hard codes only supporting this store](https://github.com/fermyon/spin/blob/665d3a023d2a192b12f32dedec9d89b8fd9f9e7e/crates/trigger/src/lib.rs#L393). Access to this store is granted by the user on the component level in a `Spin.toml`, so to grant my `hello` component access to the store I do the following:
```toml
[[component]]
id = "hello"
source = "target/wasm32-wasi/release/rust_key_value.wasm"
```

However, currently I can do the following and build and run an application without an error on `spin build --up`. Only when my app tries `Store::open("foo")` during runtime is a `NoSuchStore` error returned. Which as a user may not make sense since I added it to the component:
```
[[component]]
id = "hello"
source = "target/wasm32-wasi/release/rust_key_value.wasm"
key_value_stores = ["default", "foo"]
# OR the following
# key_value_stores = ["foo"]
```

Clear documentation on default store only may suffice, but catching during build time may be better.

itowlson:
Fixes #1232.

```
$ spin up
Error: Invalid key-value store 'hello'. This version of Spin supports only the 'default' store.

$ spin deploy
Error: Invalid key-value store 'hello'. This version of Spin supports only the 'default' store.

Learn more at https://developer.fermyon.com/cloud/faq
```

Unlike `allowed_http_hosts`, this implementation puts the validation logic in the loader, rather than putting it with the `key-value` component and adding a crate reference.  I did this solely to avoid making `loader` dependent on EVEN MOAR things.  It does risk a maintenance burden; if we ever need more sophisticated parsing and validation logic for these IDs, then it's probably worth doing it in the host component and taking the dependency.  Happy to do that now if people feel it's better.

### DIFF
--- a/crates/loader/src/validation.rs
+++ b/crates/loader/src/validation.rs
@@ -1,0 +1,48 @@
+use anyhow::{anyhow, Result};
+
+const ONLY_ALLOWED_KV_STORE: &str = "default";
+
+pub(crate) fn validate_key_value_stores(key_value_stores: &Option<Vec<String>>) -> Result<()> {
+    match key_value_stores
+        .iter()
+        .flatten()
+        .find(|id| *id != ONLY_ALLOWED_KV_STORE)
+    {
+        None => Ok(()),
+        Some(invalid) => {
+            let err = anyhow!("Invalid key-value store '{invalid}'. This version of Spin supports only the '{ONLY_ALLOWED_KV_STORE}' store.");
+            Err(err)
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn kv_empty_list_is_allowed() {
+        validate_key_value_stores(&None).expect("None should be valid");
+        validate_key_value_stores(&Some(vec![])).expect("Empty vector should be valid");
+    }
+
+    #[test]
+    fn default_store_is_allowed() {
+        validate_key_value_stores(&Some(vec!["default".to_owned()]))
+            .expect("Default store should be valid");
+        validate_key_value_stores(&Some(vec!["default".to_owned(), "default".to_owned()]))
+            .expect("Default store twice should be valid");
+    }
+
+    #[test]
+    fn non_default_store_is_not_allowed() {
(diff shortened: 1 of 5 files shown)


itowlson:
We have integration tests that want to verify the difference between a store that doesn't exist and a store that isn't included in the manifest.  CURSE OUR OWN FORESIGHT

itowlson:
For now, I have taken out the integration test check that opening a store that is _permitted_ but does not _exist_ produces a distinctive error within the guest (compared to opening a store that is nor permitted at all).  This feels like it might leave a gap in the test regime but I can't see a way around it.  Ideas welcome!

crates/loader/src/validation.rs
```diff
@@ -0,0 +1,47 @@
+use anyhow::{anyhow, Result};
+
+const ONLY_ALLOWED_KV_STORE: &str = "default";
+
+pub(crate) fn validate_key_value_stores(key_value_stores: &Option<Vec<String>>) -> Result<()> {
+    match key_value_stores {
+        None => Ok(()),
+        Some(ids) => match ids.iter().find(|id| *id != ONLY_ALLOWED_KV_STORE) {
```
lann, line 8: Another option here:

```suggestion
    match key_value_stores.iter().flatten().find(|id| *id != ONLY_ALLOWED_KV_STORE) {
```

lann (approved):

kate-goldenring:
> For now, I have taken out the integration test check that opening a store that is _permitted_ but does not _exist_ produces a distinctive error within the guest (compared to opening a store that is nor permitted at all).

@itowlson I think this is fine. It was verifying the behavior we previously had but didn't necessarily want. Once we add non-default store capability we can add it back. We could add a TODO as a reminder

kate-goldenring (approved):

PR #1236, opened 2023-03-07
Refactor UpCommand

lann:
The most complex refactoring is easier to read in just the second commit: https://github.com/fermyon/spin/pull/1236/commits/3073adbacf5074093c9730de4fa24cb4522122f0

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -845,6 +845,8 @@ mod integration_tests {
                 .collect::<Vec<String>>()
                 .join(" "),
         );
+
+        cmd.env("RUST_LOG", "spin_cli=warn");
         if let Some(envs) = envs {
             for (k, v) in envs {
                 cmd.env(k, v);
@@ -852,10 +854,11 @@ mod integration_tests {
         }
 
         let output = cmd.output()?;
+        println!("STDOUT:\n{}", String::from_utf8_lossy(&output.stdout));
+        println!("STDERR:\n{}", String::from_utf8_lossy(&output.stderr));
+
         let code = output.status.code().expect("should have status code");
         if code != 0 {
-            println!("{:#?}", std::str::from_utf8(&output.stderr)?);
-            println!("{:#?}", std::str::from_utf8(&output.stdout)?);
             panic!("command `{:?}` exited with code {}", cmd, code);
         }
 
--- a/crates/trigger/src/locked.rs
+++ b/crates/trigger/src/locked.rs
@@ -193,13 +193,8 @@ fn content_ref_path(path: &Path) -> Result<ContentRef> {
 }
 
 fn file_uri(path: &Path) -> Result<String> {
-    let path = path.canonicalize()?;
-    let url = if path.is_dir() {
-        url::Url::from_directory_path(&path)
-    } else {
-        url::Url::from_file_path(&path)
-    }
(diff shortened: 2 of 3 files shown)


tests/integration.rs
```diff
@@ -983,7 +986,7 @@ trigger = { type = "http", base = "/" }
 version = "0.1.0"
 [[component]]
 id = "unbuilt"
-source = "DOES-NOT-EXIST.wasm"
+source = "spin.toml" # dirty trick
```
itowlson, line 989: This appears to be changing the test - the test is meant to assert that help works even when the app has not been build (i.e. `source` failed to resolve _at all_).  I.e. that `spin new > spin up --help` works correctly.  This seems (if I'm reading it right) to be using a file that exists but isn't valid Wasm, which isn't what this test wants to prove   What am I missing?

itowlson (changes requested):
The refactoring looks great - love getting that knot of application origins untangled and getting a unified path for figuring out the trigger args.

The request changes is for an integration test change that seems unrelated to all this, but appears to be removing a desired check that we've had regressions on in the past.  But I may be misunderstanding the change!

tests/integration.rs
lann: Ah, I didn't realize this was testing that intentionally; indeed this is a regression for that situation. The basic problem is that deep in `spin_trigger::locked::build_locked_app` we call `canonicalize` on all our paths, which requires those paths to exist. Maybe I can resolve that problem...

itowlson: Yes, I think that was something Brian hit when trying to unify around the locked app, and is (at least part of) why the local-vs-remote code paths came to diverge.

itowlson: Certainly the local loader tries to absolutise paths as early as possible.

One of these days I want `AbsolutePath` and `RelativePath` types so we can document this behaviour at the type level _and actually enforce_ that it remains true...

itowlson (approved):

fibonacci1729 (approved):

tests/integration.rs
fibonacci1729: I was just testing this with a malformed `spin.toml` that is an `http` trigger, not sure if this  has been resolved just yet but i see the following notably missing the trigger options:

```❯  spin up -f examples/http-rust -h
spin-up
Start the Spin application

USAGE:
    spin up [OPTIONS]

OPTIONS:
        --direct-mounts         For local apps with directory mounts and no excluded files, mount
                                them directly instead of using a temporary directory
    -e, --env <ENV>             Pass an environment variable (key=value) to all components of the
                                application
    -f, --from <APPLICATION>    The application to run. This may be a manifest (spin.toml) file, a
                                directory containing a spin.toml file, or a remote registry
                                reference. If omitted, it defaults to "spin.toml"
    -h, --help
    -k, --insecure              Ignore server certificate errors from bindle server or registry
        --temp <TMP>            Temporary directory for the static assets of the components
```

itowlson: Ugh, good catch.  I bet an actually malformed `spin.toml` is a whole new failure mode.  Guessing it gets past the "no app so show default trigger options" stage, but then can't deserialise the file to work out the trigger and so gives up rather than calling to help-only trigger.  It's an existing problem though, not an issue with this PR; I'd do it as a separate issue.

PR #1237, opened 2023-03-07
Add `spin watch` command

calebschoepp:
Adds a command `spin watch`.

```
spin-watch
Rebuild and restart the Spin application when files changes

USAGE:
    spin watch [OPTIONS] [UP_ARGS]...

ARGS:
    <UP_ARGS>...    Arguments to be passed through to spin up

OPTIONS:
    -c, --clear                       Clear the screen before each run
    -d, --debounce <DEBOUNCE>         Set the timeout between detected change and re-execution
                                      [default: 100]
    -f, --file <APP_MANIFEST_FILE>    Path to spin.toml [default: spin.toml]
    -h, --help                        Print help information
        --skip-build                  Skip running spin build and only run spin up
        --watch-assets                Only watch component source and files ignoring watch
                                      configuration in spin.toml
```


src/bin/spin.rs
```diff
@@ -23,7 +24,10 @@ use spin_trigger::cli::TriggerExecutorCommand;
 async fn main() -> Result<(), Error> {
     tracing_subscriber::fmt()
         .with_writer(std::io::stderr)
-        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::from_default_env()
+                .add_directive("watchexec=off".parse()?),
```
calebschoepp, line 29: When I didn't do this `tracing::error!(..)`'s were making it through from watchexec.

src/watch_filter.rs
```diff
@@ -0,0 +1,313 @@
...
+        }
+
+        for (path, _) in event.paths() {
+            let metadata = std::fs::metadata(path)?;
+            let time_since_epoch = metadata
+                .modified()?
+                .duration_since(SystemTime::UNIX_EPOCH)?
+                .as_micros();
+            let mut modified_map = FILES_MODIFIED_AT.lock().unwrap();
+            let path_key = path.to_str().unwrap().to_owned();
```
lann, line 106: Probably shouldn't `unwrap` here. It isn't super common top have non-unicode file names, but its also pretty nasty to panic if it happens. Maybe just log and `continue` on error?

src/watch_filter.rs
```diff
@@ -0,0 +1,313 @@
...
+                        return Ok(true);
+                    }
+                }
+                None => {
+                    if time_since_epoch > self.process_start_since_epoch {
+                        modified_map.insert(path_key, time_since_epoch);
+                        return Ok(true);
+                    }
+                }
+            }
```
lann, line 122: A little tidier (assuming it actually works...):
```suggestion
            let base_time = modified_map
                .get(&path_key)
                .unwrap_or(&self.process_start_since_epoch);
            if time_since_epoch > *base_time {
                modified_map.insert(path_key, time_since_epoch);
                return Ok(true);
            }
```

src/commands/watch.rs
```diff
@@ -0,0 +1,304 @@
...
+    #[clap(
+        name = WATCH_DEBOUNCE_OPT,
+        short = 'd',
+        long = "debounce",
+        default_value = "100"
+    )]
+    pub debounce: u64,
+
+    /// Skip running spin build and only run spin up.
+    #[clap(name = WATCH_SKIP_BUILD_OPT, long = "skip-build")]
```
lann, line 67: nit: you can leave off the `long` name if it matches the field name (with conversion to kebab-case).
```suggestion
    #[clap(name = WATCH_SKIP_BUILD_OPT, long)]
```

src/commands/watch.rs
```diff
@@ -0,0 +1,304 @@
...
+
+        // Prepare RuntimeConfig for Watchexec
+        let mut runtime_config = RuntimeConfig::default();
+        let (spin_cmd, spin_args) = self.generate_command();
+        runtime_config.commands(vec![watchexec::command::Command::Exec {
+            prog: spin_cmd,
+            args: spin_args,
+        }]);
+        let app_dir = parent_dir(&self.app)?;
+        runtime_config.pathset([app_dir.to_str().unwrap().to_owned()]);
```
lann, line 107: Based on [the docs](https://docs.rs/watchexec/2.3.0/watchexec/config/struct.RuntimeConfig.html#method.pathset) it doesn't look like you need to convert to a string here. Maybe this would work:
```suggestion
        runtime_config.pathset([app_dir.clone()]);
```

src/commands/watch.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    /// Clear the screen before each run.
+    #[clap(
+            name = WATCH_CLEAR_OPT,
+            short = 'c',
+            long = "clear",
+    )]
+    pub clear: bool,
+
+    /// Set the timeout between detected change and re-execution.
```
lann, line 57:
```suggestion
    /// Set the timeout between detected change and re-execution, in milliseconds.
```

src/commands/watch.rs
```diff
@@ -0,0 +1,304 @@
...
+        );
+        (spin_cmd, spin_args)
+    }
+
+    async fn generate_path_patterns(&self) -> Result<Vec<String>> {
+        let app_manifest = spin_loader::local::raw_manifest_from_file(&self.app).await?;
+        let app_manifest = match app_manifest {
+            RawAppManifestAnyVersion::V1New { manifest, .. } => manifest,
+            RawAppManifestAnyVersionImpl::V1Old { manifest, .. } => manifest,
+        };
```
lann, line 191:
```suggestion
        let app_manifest = spin_loader::local::raw_manifest_from_file(&self.app).await?.into_v1();
```

lann (changes requested):
Only real blockers are the `debounce` docs and `unwrap`ing Patterns.

src/watch_filter.rs
```diff
@@ -0,0 +1,313 @@
...
+        path_patterns: Vec<String>,
+        ignore_patterns: Vec<String>,
+    ) -> Self {
+        // Record approximate start time of process for the purpose of making sure files have been modified since
+        let start_time = SystemTime::now()
+            .duration_since(std::time::UNIX_EPOCH)
+            .unwrap()
+            .as_micros();
+
+        tracing::info!("watching relative to app dir: {:?}", app_dir);
```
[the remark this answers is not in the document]
calebschoepp: I've fixed this throughout except where I need to do field access b/c that is not supported in format strings.

src/watch_filter.rs
```diff
@@ -0,0 +1,313 @@
...
+        filekind::{FileEventKind, ModifyKind},
+        Event, Priority, Tag,
+    },
+    filter::Filterer,
+    signal::source::MainSignal::Interrupt,
+};
+
+lazy_static! {
+    /// A default set of ignore patterns that can be used by the consumer of WatchFilter.
+    pub static ref DEFAULT_IGNORE_PATTERNS: Vec<String> = vec![String::from("*.swp")];
```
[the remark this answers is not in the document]
lann: Oh it's a `lazy_static` thing

src/watch_filter.rs
calebschoepp: Noice!

src/watch_filter.rs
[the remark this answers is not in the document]
calebschoepp: I think it is just required as special syntax for `lazy_static!`. Did some googling and I couldn't find another way to define it. I could also just make it something like below if we like:

```rust
fn default_ignore_patterns() -> Vec<String> {
    vec![String::from("*.swp")]
}
```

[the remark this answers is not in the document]
lann: I was mostly just confused by the non-standard syntax. Might be nice to drop the whole lazy_static thing if you remove the other one though.

src/watch_filter.rs
```diff
@@ -0,0 +1,313 @@
...
+    /// Filter doesn't have a mutable reference so we need to store this state outside of filter.
+    static ref FILES_MODIFIED_AT: Mutex<HashMap<String, u128>> = Mutex::new(HashMap::new());
+}
+
+#[derive(Debug)]
+pub(crate) struct WatchFilter {
```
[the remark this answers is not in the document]
calebschoepp: I see, that would only be necessary if we were re-exporting the module

src/watch_filter.rs
```diff
@@ -0,0 +1,313 @@
...
+            .as_micros();
+
+        tracing::info!("watching relative to app dir: {:?}", app_dir);
+        tracing::info!("watching path patterns: {:?}", path_patterns);
+        tracing::info!("ignoring path patterns: {:?}", ignore_patterns);
+
+        WatchFilter {
+            path_patterns: path_patterns
+                .iter()
+                .map(|path| Pattern::new(app_dir.join(path).to_str().unwrap()).unwrap())
```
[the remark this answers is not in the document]
calebschoepp: Really good idea

src/commands/watch.rs
calebschoepp: Cool, didn't know that. I tried it out and it made the flag names all caps which I didn't like so I'm going to keep this as is.

src/commands/watch.rs
calebschoepp: Worked like a charm.

calebschoepp:
Thanks for the feedback @lann . I've made the changes and left two unresolved where I had comments or questions.

src/commands/watch.rs
```diff
@@ -0,0 +1,302 @@
...
+    #[clap(name = WATCH_SKIP_BUILD_OPT, long = "skip-build")]
+    pub skip_build: bool,
+
+    /// Only watch component source and files ignoring watch configuration in spin.toml.
+    #[clap(name = WATCH_WATCH_ASSETS_OPT, long = "watch-assets", requires = WATCH_SKIP_BUILD_OPT)]
+    pub watch_assets: bool,
+
+    /// Arguments to be passed through to spin up.
+    #[clap()]
+    pub up_args: Vec<OsString>,
```
dicej, line 74: Should we make this a `Vec<String>` given that we're calling `into_string().unwrap()` on them below?  Might as well let Clap do the UTF-8 checking rather than panic later on.

src/watch_filter.rs
```diff
@@ -0,0 +1,314 @@
...
+            .to_str()
+            .with_context(|| format!("non-unicode pattern {path:?}"))?;
+        Pattern::new(pat_str).with_context(|| format!("invalid glob pattern {path:?}"))
+    }
+
+    fn has_valid_event_kind(&self, event: &Event) -> bool {
+        event.tags.iter().any(|tag| {
+            matches!(
+                tag,
+                Tag::FileEventKind(FileEventKind::Modify(ModifyKind::Data(_)))
```
dicej, line 72: Could make this more concise, e.g.: `Tag::FileEventKind(FileEventKind::Modify(ModifyKind::Data(_) | ModifyKind::Name(_)) | FileEventKind::Create(_) | FileEventKind::Remove(_)))`

src/watch_filter.rs
```diff
@@ -0,0 +1,314 @@
...
+    },
+    filter::Filterer,
+    signal::source::MainSignal::Interrupt,
+};
+
+lazy_static! {
+    /// A default set of ignore patterns that can be used by the consumer of WatchFilter.
+    pub static ref DEFAULT_IGNORE_PATTERNS: Vec<String> = vec![String::from("*.swp")];
+    /// Filter doesn't have a mutable reference so we need to store this state outside of filter.
+    static ref FILES_MODIFIED_AT: Mutex<HashMap<String, SystemTime>> = Mutex::new(HashMap::new());
```
lann, line 25: Remove?

src/watch_filter.rs
```diff
@@ -0,0 +1,314 @@
...
+            }
+        }
+        Ok(false)
+    }
+
+    fn one_is_after_another(one: &SystemTime, another: &SystemTime) -> Result<bool> {
+        let one = one.duration_since(SystemTime::UNIX_EPOCH)?.as_micros();
+        let another = another.duration_since(SystemTime::UNIX_EPOCH)?.as_micros();
+        Ok(one > another)
+    }
```
lann, line 134: It looks like `SystemTime`s can be directly compared with each other (they implement `PartialOrd`).

src/commands/watch.rs
calebschoepp: Sure. I like that.

src/watch_filter.rs
```diff
@@ -0,0 +1,314 @@
...
+                Some(s) => s.to_owned(),
+                None => {
+                    tracing::warn!("can't check non-unicode path: {path:?}");
+                    continue;
+                }
+            };
+            let base_time = modified_map
+                .get(&path_key)
+                .unwrap_or(&self.process_start_time);
+            if WatchFilter::one_is_after_another(&path_time, base_time)? {
```
lann, line 122: It looks like `SystemTime`s can be directly compared with each other (they [implement `PartialOrd`](https://doc.rust-lang.org/std/time/struct.SystemTime.html#impl-PartialOrd%3CSystemTime%3E-for-SystemTime)).

```suggestion
            if path_time > base_time {
```

src/watch_filter.rs
calebschoepp: Good catch.

lann:
Code LGTM.

calebschoepp:
Okay, looks like I've got some approvals. Once I can figure out why the unit tests are failing it should be good to go.

@itowlson lemme know if you want to take a look too.

PR #1240, opened 2023-03-08
Upgrade cargo-vet and add imports

bholley:
We recently released a new version of cargo-vet with improved ergonomics in various areas. In case this is of interest to Fermyon, I prepared a PR to upgrade the tool and get the instance back into a green state (via `cargo vet regenerate exemptions`).

Previously, the tool would conservatively import any remote audit that _might_ be relevant, whereas now `imports.lock` contains the minimal set of what's actually used. This is the reason for the changes in that file.

I also proposed a few additional imports to the list to match the imports Mozilla and wasmtime are now (or will soon be) using. This allows a number of exemptions to be pruned, but happy to drop any of those as desired.

### DIFF
--- a/.github/workflows/audits.yml
+++ b/.github/workflows/audits.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.repository_owner == 'fermyon' }}
     env:
-      CARGO_VET_VERSION: 0.3.0
+      CARGO_VET_VERSION: 0.4.0
     steps:
       - uses: actions/checkout@v3
       - name: Install Rust
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -17,4 +17,3 @@ who = "Radu Matei <radu.matei@fermyon.com>"
 criteria = "safe-to-deploy"
 version = "0.3.1"
 notes = "The Bytecode Alliance is the author of this crate."
-
--- a/supply-chain/config.toml
+++ b/supply-chain/config.toml
@@ -5,7 +5,7 @@
 url = "https://raw.githubusercontent.com/bytecodealliance/wasmtime/main/supply-chain/audits.toml"
 
 [imports.mozilla]
-url = "https://hg.mozilla.org/mozilla-central/raw-file/tip/supply-chain/audits.toml"
+url = "https://raw.githubusercontent.com/mozilla/supply-chain/main/audits.toml"
 
 [policy.bindle]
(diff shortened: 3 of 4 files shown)


radu-matei (approved):
Thank you so much for the update, @bholley, really appreciate the update!

The new imports addition makes sense to me, particularly as they are shared across a set of projects we trust.

LGTM!

radu-matei:
Could you please rebase and sign your commits, please?
(both sign-off and with a GPG key)

Thanks!

bholley:
@radu-matei Done.

I might also suggest reconfiguring CI to run cargo-vet as a blocking step rather than as a cron job. It appears to have gotten stale in the latter configuration, and the tool is a lot more useful when you keep it in a green state. You can always add exemptions in situations where you want to defer the audits on a PR for whatever reason.

radu-matei:
The main reasoning behind having a cron job rather than blocking on a PR was that we usually have non-maintainers contributing to the code base, and for the audits, we would like to keep them to just maintainers. Besides trusting audits from external people, we would also add an extra burden to someone submitting a PR.

Do you have thoughts around this?

bholley:
> The main reasoning behind having a cron job rather than blocking on a PR was that we usually have non-maintainers contributing to the code base, and for the audits, we would like to keep them to just maintainers.

Agree this is an important property to maintain!

> Besides trusting audits from external people, we would also add an extra burden to someone submitting a PR.
> 
> Do you have thoughts around this?

Indeed. There are basically two ways to handle the situation when an external contributor submits a patch which adds new third-party Rust code.

The first is for the reviewer (presumably a maintainer) to go and perform the relevant audits (or add the relevant exemptions) and ask the contributor to rebase on top of that. This is what wasmtime and Firefox do.

The second is to ask the contributor to add exemptions (either manually, or with the tool) and then file a followup for a maintainer to go remove those exemptions and replace them with actual audits. This can reduce the synchronous demand on the reviewer's time and remove contribution delay in some cases, but does result in new unaudited code appearing in the tree until the followup happens (Firefox and wasmtime have policies against this, hence the former approach).

It may vary by use-case, but our experience has been that casual contributors _mostly_ don't need to do major fiddling with Rust dependencies, so it hasn't been a super-frequent situation in practice.

radu-matei:
Rebasing makes sense in this scenario. Let's track moving to blocking on PRs separately and give other maintainers a chance to comment.

I think this change is good to go, thank you so much for your contribution, @bholley!

PR #1243, opened 2023-03-08
bump Wasmtime version to 5.0.1

dicej:


lann (approved):

lann:
I assume this doesn't require bumping `wit-bindgen-backport`? It makes sense that it wouldn't necessarily.

fibonacci1729 (approved):

dicej:
> I assume this doesn't require bumping `wit-bindgen-backport`? It makes sense that it wouldn't necessarily.

The same thing occurred to me, but I think we're good, both from a semver perspective and a security perspective, given that we're pulling in the right versions of everything in Cargo.lock.

PR #1245, opened 2023-03-08
App state dir

lann:


crates/app/src/app.rs
```diff
@@ -0,0 +1,225 @@
...
+        match self.get_metadata::<Url>("origin")? {
+            Some(url) if url.scheme() == "file" => {
+                let path = url.to_file_path().map_err(|_| {
+                    Error::MetadataError(format!("Invalid origin file URL: {url:?}"))
+                })?;
+                Ok(Some(path))
+            }
+            _ => Ok(None),
+        }
+    }
```
lann, line 55: This file is mostly just code moved from `lib.rs`, but I also added this method.

lann: ...which I just realized I don't actually use in this revision :facepalm:

crates/trigger/src/lib.rs
```diff
@@ -361,27 +304,9 @@ fn decode_preinstantiation_error(e: anyhow::Error) -> anyhow::Error {
     e
 }
 
-fn key_value_file_location(app: &App) -> Option<PathBuf> {
```
lann, line 364: Moved to `RuntimeConfig::sqlite_db_path`. Maybe should be `key_value_file_location` instead? :shrug:

crates/trigger/src/lib.rs
```diff
@@ -101,27 +85,21 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
                 builder.add_host_component(outbound_pg::OutboundPg::default())?;
                 builder.add_host_component(outbound_mysql::OutboundMysql::default())?;
 
-                let manager = OnceCell::new();
-
+                let key_value_file = runtime_config.sqlite_db_path()?;
+                let store_manager = make_store_manager(key_value_file);
```
lann, line 89: @dicej I _think_ its fine to build this eagerly?

crates/trigger/src/cli.rs
```diff
@@ -88,6 +88,14 @@ where
     )]
     pub runtime_config_file: Option<PathBuf>,
 
+    /// Set the application state directory path. This is used in the default
+    /// locations for logs, key value stores, etc.
+    ///
+    /// For local apps, this defaults to `./.spin/` relative to the `spin.toml` file.
```
michelleN, line 94: is the `./` prefix necessary in `./.spin/`?

crates/trigger/src/cli.rs
```diff
@@ -22,8 +22,8 @@ pub const RUNTIME_CONFIG_FILE: &str = "RUNTIME_CONFIG_FILE";
 
 // Set by `spin up`
 pub const SPIN_LOCKED_URL: &str = "SPIN_LOCKED_URL";
+pub const SPIN_LOCAL_APP_DIR: &str = "SPIN_LOCAL_APP_DIR";
```
michelleN, line 25: This refers now to the application directory not `<app-dir>/.spin/` anymore right? Just want to confirm.

michelleN (commented):
What is the reasoning for changing the behavior of `--log-dir` to not creating the directory specified?

lann:
> What is the reasoning for changing the behavior of `--log-dir` to not creating the directory specified?

OK, it turns out I actually forgot to remove the code that creates the log directory, so...nothing to revert!

crates/trigger/src/cli.rs
lann: Yep.

dicej (approved):

crates/trigger/src/cli.rs
```diff
@@ -147,29 +155,42 @@ where
...
         self.update_wasmtime_config(builder.wasmtime_config_mut())?;
 
-        let logging_hooks = StdioLoggingTriggerHooks::new(self.follow_components(), self.log_dir());
+        let log_dir = runtime_config.log_dir().unwrap_or_else(|err| {
+            tracing::warn!("Logging disabled: {err:?}");
+            None
+        });
+        if let Some(log_dir) = &log_dir {
+            println!("Logging component stdio to {:?}\n", log_dir.join(""))
+        }
```
lann, line 171: @michelleN 


```console
$ spin up

Serving http://127.0.0.1:3000
...
```

crates/trigger/src/key_value.rs
```diff
@@ -0,0 +1,41 @@
+use std::{path::Path, sync::Arc};
```
lann, line 1: @dicej Please take another look. In going back to `change_dir_all` here I ended up moving this into a new file and refactoring a bit.

dicej (approved):

crates/trigger/src/cli.rs
```diff
@@ -147,29 +156,44 @@ where
...
-            TriggerExecutorBuilderConfig::load_from_file(self.runtime_config_file.clone())?;
+        let runtime_config = self.build_runtime_config()?;
 
         let _sloth_warning = warn_if_wasm_build_slothful();
 
         let mut builder = TriggerExecutorBuilder::new(loader);
         self.update_wasmtime_config(builder.wasmtime_config_mut())?;
 
-        let logging_hooks = StdioLoggingTriggerHooks::new(self.follow_components(), self.log_dir());
+        if runtime_config.state_dir().is_none() {
```
itowlson, line 166: ~Is there a particular reason to perform this check (and the `log_dir` one) in the trigger rather than in the `up` command?  Apart from the 'niceness' factor of having the UI stuff as far as possible on the `up` side, `up` has context for local vs remote and could skip the warning in cases where the user explicitly unset the state directory.~  Except this is a trigger-only option and `up` doesn't know about it Ivan

crates/trigger/src/cli.rs
```diff
@@ -147,29 +156,44 @@ where
...
+            println!("No local state dir; logs and key value data may not be written to disk.");
```
itowlson, line 167: It would be good to correlate this with `log_dir`.  Currently if `state_dir` is unset but `log_dir` is set, it seems like Spin will print `Logs may not be written to disk, except here is where I'm writing my logs`, which feels contradictory.

crates/trigger/src/cli.rs
```diff
@@ -147,29 +156,44 @@ where
...
+            println!("You can specify a state dir with `spin-up --state-dir <dir>`.\n");
+        }
+
+        let log_dir = runtime_config.log_dir();
+        if let Some(log_dir) = &log_dir {
+            println!("Logging component stdio to {:?}\n", log_dir.join(""))
```
itowlson, line 173: Is this only component stdio?  Spin's own logs are... transient? or written somewhere else?

crates/trigger/src/key_value.rs
```diff
@@ -0,0 +1,41 @@
...
+    Ok(KeyValueComponent::new(spin_key_value::manager(move |_| {
+        manager.clone()
+    })))
+}
+
+fn create_parent_dir(path: &Path) -> Result<()> {
+    let dir = path
+        .parent()
+        .with_context(|| format!("{path:?} missing parent dir"))?;
+    std::fs::create_dir_all(dir).with_context(|| format!("Failed to create parent dir {dir:?}"))
```
itowlson, line 40: nit (as if the others weren't also nits): I _think_ that if this fails, it will print "Failed to create key-value store directory" and "Failed to create parent dir ..." - this makes it sound like the second is the parent directory of the first when they are actually the same directory.  Should we maybe drop 'parent' here?  Or tweak the message to something like 'Unable to create K/V store file" (then "parent dir" explains why)?

crates/trigger/src/runtime_config.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub fn new(local_app_dir: Option<PathBuf>) -> Self {
+        Self {
+            local_app_dir: local_app_dir.map(Into::into),
+            ..Default::default()
+        }
+    }
+
+    /// Load a runtime config file from the given path. Options specified in a
+    /// later-loaded file take precedence over any earlier-loaded files.
+    pub fn load_config_file(&mut self, path: impl AsRef<Path>) -> Result<()> {
```
itowlson, line 37: If there can be multiple files in play, then maybe `merge_config_from` rather than `load`???

crates/trigger/src/runtime_config.rs
```diff
@@ -0,0 +1,262 @@
...
+        // Default EnvProvider
+        let dotenv_path = self.local_app_dir.as_deref().map(|path| path.join(".env"));
+        let env_provider = EnvProvider::new(SPIN_CONFIG_ENV_PREFIX, dotenv_path);
+
+        let mut providers: Vec<BoxedConfigProvider> = vec![Box::new(env_provider)];
+        for provider_config in self
+            .opts_layers()
+            .flat_map(|opts| opts.config_providers.iter())
+        {
+            providers.push(match provider_config {
```
itowlson, line 58: It is a nit for now but this is a very beefy match expression to squeeze into a function argument!

crates/trigger/src/runtime_config.rs
```diff
@@ -0,0 +1,262 @@
...
+            // If the state dir is set, build the default path
+            Some(state_dir.join(DEFAULT_SQLITE_DB_FILENAME))
+        } else {
+            None
+        }
+    }
+
+    /// Returns an iterator of RuntimeConfigOpts in order of decreasing precedence
+    fn opts_layers(&self) -> impl Iterator<Item = &RuntimeConfigOpts> {
+        std::iter::once(&self.overrides).chain(self.files.iter().rev())
```
itowlson, line 121: Nice

crates/trigger/src/cli.rs
lann: We've never written Spin's own logs anywhere. I considered adding `<log-dir>/spin.log` but this PR is already an overgrown mess.

crates/trigger/src/key_value.rs
lann: Changed outer context to "Failed to create key value store".

crates/trigger/src/cli.rs
lann: You have shamed me in my laziness. As penance I further engorged this PR with new functionality, leading to this behavior:

```console
$ spin up --from examples/http-rust
Logging component stdio to "examples/http-rust/.spin/logs/"

Serving [...]
```
```console
$ spin up --from examples/rust-key-value
Logging component stdio to "examples/rust-key-value/.spin/logs/"
Storing key-value data in "examples/rust-key-value/.spin/sqlite_key_value.db"

Serving [...]
```
```console
$ spin up --from-registry ttl.sh/lann-rust-key-value:1m
Using in-memory key-value store; data will not be saved!

Serving [...]
```

lann:
@itowlson @dicej 
To give better warnings I expanded this PR in [these commits](https://github.com/fermyon/spin/pull/1245/files/1ced45bb9fcf9199d9129ca307bb72ddacefc279..f00199aa3a82f8a6f39f9d82e9df4439cd53f7d2), which (among other things) move the spin-key-value-sqlite stdout messages to spin-trigger and print them on startup instead of on first KV use.

crates/trigger/src/cli.rs
itowlson: Is that... is that only printing the K/V message if the app actually uses the default key-value store?  BEAUTIFUL.  Love this.

lann: Indeed! Well it just looks for the `key_value_stores` key in the manifest, but that should be approximately the same.

PR #1247, opened 2023-03-09
update version to v1.0.0-rc.1

michelleN:


kate-goldenring:
Do we also want to update the rust templates to point to this new rc: `templates/http-rust/content/Cargo.toml` and `templates/redis-rust/content/Cargo.toml`?

vdice (commented):
This looks good, though I _think_ we should restrict the bump in the `templates/` dir to only the `SDK_VERSION` update in the Makefile, if only because [automation handles updating the templates per that version](https://github.com/fermyon/spin/blob/main/.github/workflows/release.yml#L234-L271) as part of the release process.  (Most likely it _should_ be able to function fine if we've already updated the values in the template Cargo.tml/go.mod files... but might be safe to just stick to the usual process.)

fibonacci1729 (approved):

michelleN:
Thank you @vdice. I was curious about that. I have removed all of the version updates from the templates directory except for the Makefile.

vdice (approved):

PR #1249, opened 2023-03-09
return 405 instead of panicking on unsupported method

dicej:
Previously, Spin would panic and not return a response when handling a request with an unsupported HTTP method, e.g. `curl -X WHAT http://127.0.0.1:3000/`.  Now it will return a 405 instead.


itowlson (approved):

michelleN (approved):

fibonacci1729 (approved):
Nice catch!

PR #1256, opened 2023-03-10
docs: Add branch creation to release process

lann:


fibonacci1729 (approved):

docs/content/release-process.md
```diff
@@ -18,6 +18,10 @@ To cut a release of Spin, you will need to do the following:
    make sure you get signoff on the PR)
 1. Before proceeding, verify that the merge commit on `main` intended to be
    tagged is green, i.e. CI is successful
+1. If this is the first release for this major/minor version, create a release
+   branch e.g. `v2.0`. With our branch protection rules this is easiest from
+   the Github UI with the
+   [New Branch button here](https://github.com/fermyon/spin/branches).
```
vdice, line 24: Shall we update the next item to explictly mention "Check out the `v*` branch created above and create a new tag with the same name..."?

Optional: I wonder if we should also mention specific commands/steps for the tag creation bit.  We could opt for cli (`git tag v* && git push origin v*`) -- also do we want to start signing and/or annotating the tag? -- or maybe we opt for a UI-based approach?

lann: I'd say if we're going to add any more complexity to the process at this point we should script it.

Issue #1254, opened 2023-03-10
Funky looking help text


fibonacci1729:
Specifically the indentation around the `registry` command.

<img width="1570" alt="Screenshot 2023-03-10 at 4 29 35 PM" src="https://user-images.githubusercontent.com/7605165/224432111-9bd23d07-3dca-4dfa-af38-f680c3ae0428.png">

itowlson:
That seems to be how Clap likes to do long command descriptions - I added some text to `spin new` and it did the same thing.  I think the idea is that if several of your strings are long, and there wasn't an indent, then the entries would feel like a wall of text where the start of each entry was not visually obvious.

![image](https://user-images.githubusercontent.com/865538/224563779-dc7fdb2a-6677-4da7-92ba-24d6a993fc6e.png)

That said, the `registry` command has drifted out of proportion to the others in both length and detail, and needs things like the "experimental" warning removing.  So the fix is probably to prune back the registry text rather than to add delicious bacon to the other commands.

PR #1262, opened 2023-03-13
chore: clean up some command help text

fibonacci1729:


fibonacci1729:
closes #1254

itowlson (approved):

itowlson:
+ @bacongobbler for the login change

src/commands/login.rs
```diff
@@ -28,7 +28,7 @@ const SPIN_CLIENT_ID: &str = "583e63e9-461f-4fbe-a246-23e0fb1cad10";
 
 const DEFAULT_CLOUD_URL: &str = "https://cloud.fermyon.com/";
 
-/// Log into the Fermyon Platform.
+/// Log into the Fermyon Platform using credentials from ~/.docker/config.json.
```
bacongobbler, line 31: I don't think this makes sense at the present moment. `spin deploy` is still [using bindle API endpoints](https://github.com/fermyon/spin/blob/9cdf8aebf01f2dee6c2f1515d476f7d3642250c9/src/commands/deploy.rs#L350-L359), and when we implement OCI support, we won't be reading configuration from `~/.docker/config.json` IIRC.

Would you mind removing this until we make more progress on the OCI migration?

bacongobbler: @vdice do you have anything you want to add here?

bacongobbler (changes requested):
Thanks for the ping @itowlson! Dropping my 2 cents here.

src/commands/login.rs
fibonacci1729: Ahh good catch @bacongobbler . This was meant to be attached to `registry login` but i had fatally goofed.

itowlson: Oops, I somehow misread this as a remove.  Yeah we should definitely not have this text.  Thanks for the catch @bacongobbler!

@fibonacci1729 I think this used to be true of `spin registry login` (not `spin login` itself), but I'm not sure if that's still the case.

src/commands/registry.rs
```diff
@@ -6,16 +6,13 @@ use std::{io::Read, path::PathBuf};
...
-/// Currently, the OCI commands are reusing the credentials from ~/.docker/config.json to
-/// authenticate to registries.
 #[derive(Subcommand, Debug)]
 pub enum RegistryCommands {
     /// Push a Spin application to a registry.
     Push(Push),
     /// Pull a Spin application from a registry.
     Pull(Pull),
-    /// Log in to a registry.
+    /// Log in to a registry using credentials from ~/.docker/config.json.
```
itowlson, line 15: Looking at the `oci::auth` module I think even this no longer uses the Docker credentials, except maybe as a fallback?

```
    fn default_path() -> Result<PathBuf> {
        Ok(dirs::config_dir()
            .context("Cannot find configuration directory")?
            .join("fermyon")
            .join("registry-auth.json"))
    }
```

I'd suggest removing the auth file path altogether.

itowlson (changes requested):

src/commands/registry.rs
fibonacci1729: Yeah you beat me to the words. I agree. Thanks!

bacongobbler (approved):

itowlson (approved):
Thanks and sorry for the faff!

PR #1264, opened 2023-03-13
In `spin plugin list`, do not print prerelease warnings ahead of list

### LINKED ISSUE #1263: spin plugin list warnings are confusing
I think these warnings are a little confusing because they aren't pre-fixed with which plugin the warning is for.
```
~  $ spin plugin list
You're using a pre-release version of Spin (1.0.0-rc.1). This plugin might not be compatible (supported: >=0.7). Continuing anyway.
You're using a pre-release version of Spin (1.0.0-rc.1). This plugin might not be compatible (supported: >=0.7). Continuing anyway.
You're using a pre-release version of Spin (1.0.0-rc.1). This plugin might not be compatible (supported: >=0.9). Continuing anyway.
You're using a pre-release version of Spin (1.0.0-rc.1). This plugin might not be compatible (supported: >=0.6). Continuing anyway.
You're using a pre-release version of Spin (1.0.0-rc.1). This plugin might not be compatible (supported: >=0.9). Continuing anyway.
You're using a pre-release version of Spin (1.0.0-rc.1). This plugin might not be compatible (supported: >=0.9). Continuing anyway.
js2wasm 0.1.0
js2wasm 0.2.0
js2wasm 0.3.0
js2wasm 0.4.0
py2wasm 0.1.0
py2wasm 0.1.1
```

itowlson:
Fixes #1263.

The output with a prerelease is now:

```
# Catalogue
$ spin plugins list
arm-only 0.1.0 [incompatible]
js2wasm 0.1.0
js2wasm 0.2.0
js2wasm 0.3.0 [installed]
js2wasm 0.4.0
not-spin-ver 10.2.0
py2wasm 0.1.0 [installed]
trigger-armstuff 0.1.0 [incompatible]
trigger-madeup 0.1.0

# Installed
$ spin plugins list --installed
js2wasm 0.3.0 [installed]
pluginify 0.3 [installed]
py2wasm 0.1.0 [installed]
timer 0.1.0 [installed]
trigger-composite 0.3 [installed]
trigger-sqs 0.2 [installed]
trigger-timer 0.1.0 [installed]

# Making sure we still get a warning when actually executing a plugin
$ spin js2wasm
You're using a pre-release version of Spin (1.0.0-pre0). This plugin might not be compatible (supported: >=0.7). Continuing anyway.
error: The following required arguments were not provided:
    <input>

USAGE:
    js2wasm <input> -o <output>
```

Note that the compatibility status errs on the side of liberal. As the check function allows any plugin to run on a prerelease, the list treats it as "everything is compatible."  We could be strict - the downside then is that pretty much _every_ plugin would then say `[requires other Spin version]` which is a bit discouraging.  Happy to discuss or change though!

Also: I renamed some existing functions to make it clearer that they were about interactively _warning_ about version compatibility, not just about _checking_ it.

### DIFF
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -42,7 +42,7 @@ impl PluginManifest {
         self.packages.iter().any(|p| p.matches_current_os_arch())
     }
     pub fn is_compatible_spin_version(&self, spin_version: &str) -> bool {
-        check_supported_version(self, spin_version, false).is_ok()
+        is_version_compatible_enough(&self.spin_compatibility, spin_version).unwrap_or(false)
     }
     pub fn is_installed_in(&self, store: &PluginStore) -> bool {
         match store.read_plugin_manifest(&self.name) {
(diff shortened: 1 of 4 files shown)


crates/plugins/src/manifest.rs
```diff
@@ -117,26 +117,51 @@ impl Architecture {
...
     })?;
     let version = Version::parse(spin_version)?;
-    if !comparator.matches(&version) {
+    Ok(comparator.matches(&version))
+}
+
+/// This is more liberal than `is_version_fully_compatible`; it relaxes the semver requirement
+/// for Spin pre-releases, so that you don't get *every* plugin showing as incompatible when
+/// you run a pre-release.  This is intended for listing; when executing, we use the interactive
+/// `warn_unsupported version`, which provides the full nuanced feedback.
```
kate-goldenring, line 144:
```suggestion
/// `warn_unsupported_version`, which provides the full nuanced feedback.
```

kate-goldenring (approved):
LGTM thanks Ivan!

crates/plugins/src/manifest.rs
itowlson: I'm embarrassed at how long, _even with highlighting_, it took me to spot this.  Thank you!

PR #1266, opened 2023-03-14
feat(deploy): Add deploy flags to seed key/values

adamreese:
Allow seeding the key-value store on spin deploy.

```
spin deploy --key-value 'kate=kate is rad'
```

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -34,7 +34,7 @@ use super::login::LoginCommand;
 use super::login::LoginConnection;
 
 const SPIN_DEPLOY_CHANNEL_NAME: &str = "spin-deploy";
-
+const SPIN_DEFAULT_KV_STORE: &str = "default";
 const BINDLE_REGISTRY_URL_PATH: &str = "api/registry";
 
 /// Package and upload an application to the Fermyon Platform.
@@ -93,6 +93,10 @@ pub struct DeployCommand {
         env = DEPLOYMENT_ENV_NAME_ENV
     )]
     pub deployment_env_id: Option<String>,
+
+    /// Pass a key/value (key=value) to all components of the application.
+    #[clap(long, parse(try_from_str = parse_kv))]
+    pub key_values: Vec<(String, String)>,
 }
 
 impl DeployCommand {
@@ -385,6 +389,18 @@ impl DeployCommand {
                 .await
                 .context("Problem patching a channel")?;
 
+                for kv in self.key_values {
+                    CloudClient::add_key_value_pair(
+                        &client,
+                        app_id,
+                        SPIN_DEFAULT_KV_STORE.to_string(),
+                        kv.0,
+                        kv.1,
+                    )
+                    .await
+                    .context("Problem creating key/value")?;
+                }
+
                 existing_channel_id
             }
             Err(_) => {
@@ -399,7 +415,7 @@ impl DeployCommand {
                     .get_revision_id_cloud(&client, bindle_id.version_string().clone(), app_id)
                     .await?;
 
-                CloudClient::add_channel(
+                let channel_id = CloudClient::add_channel(
                     &client,
                     app_id,
                     String::from(SPIN_DEPLOY_CHANNEL_NAME),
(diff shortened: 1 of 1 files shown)


src/commands/deploy.rs
```diff
@@ -385,6 +389,18 @@ impl DeployCommand {
                 .await
                 .context("Problem patching a channel")?;
 
+                for kv in self.key_value {
+                    CloudClient::add_key_value_pair(
+                        &client,
+                        app_id,
+                        String::from("default"),
```
adamreese, line 396: I'm not sure what to use for store here

src/commands/deploy.rs
```diff
@@ -93,6 +93,10 @@ pub struct DeployCommand {
         env = DEPLOYMENT_ENV_NAME_ENV
     )]
     pub deployment_env_id: Option<String>,
+
+    /// Pass a key/value (key=value) to all components of the application.
+    #[clap(short, long, parse(try_from_str = parse_kv))]
```
kate-goldenring, line 98: IMO `kv` is a more descriptive short flag and less likely to conflict with later flags
```suggestion
    #[clap(short = "kv", long, parse(try_from_str = parse_kv))]
```

src/commands/deploy.rs
```diff
@@ -93,6 +93,10 @@ pub struct DeployCommand {
...
+    pub key_value: Vec<(String, String)>,
```
kate-goldenring, line 99: nit seems this should be plural
```suggestion
    pub key_values: Vec<(String, String)>,
```

src/commands/deploy.rs
```diff
@@ -783,3 +813,12 @@ fn has_expired(login_connection: &LoginConnection) -> Result<bool> {
         None => Ok(false),
     }
 }
+
+// Parse the environment variables passed in `key=value` pairs.
```
kate-goldenring, line 817: i think this comment may need an update

src/commands/deploy.rs
kate-goldenring: I think `"default"` is correct for now. However, that is a good point. We may need to provide the option to set the store name when we provide more stores. if we can assume all KVs can be set for a single store, the following may be good: `spin deploy --kv key=value --store "myStore"` where `--store` defaults to "default" if not set. The concern becomes if we want to presetting kv values for multiple stores. At that point it may be better for users to use a CLI or UI. 

Circling back, for the first version of this PR, we should probably just set "default" as a constant since we don't have support for other stores yet. @bacongobbler thoughts?

bacongobbler: Setting "default" as a constant would be a good first step.

src/commands/deploy.rs
adamreese: I believe short flags are a single character.  I agree with removing `-k` as the short flag though

src/commands/deploy.rs
adamreese: Added a constant for default store name

src/commands/deploy.rs
```diff
@@ -93,6 +93,10 @@ pub struct DeployCommand {
         env = DEPLOYMENT_ENV_NAME_ENV
     )]
     pub deployment_env_id: Option<String>,
+
+    /// Pass a key/value (key=value) to all components of the application.
+    #[clap(long, parse(try_from_str = parse_kv))]
+    pub key_values: Vec<(String, String)>,
```
bacongobbler, line 99: The code assumes we return a vector, but parse_kv only extracts a single KV.

Is there a risk by allowing any key and any value with an equal sign as the separator? If we eventually change parse_kv to parse KV pairs separated by commas, would that break compatibility?

This would be allowed in its current form: `”greeting=Hello,World”`

itowlson: I think the `Vec` type means that `clap` will allow multiple instances of the argument, each containing one pair.  This is how it works for `spin up --env` which looks pretty similar.

adamreese: It's a direct lift from `spin up --env`

bacongobbler: Ah. That makes much more sense. Thank you.

bacongobbler (approved):

PR #1270, opened 2023-03-14
fix: Absolutize direct mounts

lann:
Locked mount sources are expected to have absolute paths.


michelleN (commented):
Are there any unit tests we can add here?

itowlson (approved):

PR #1272, opened 2023-03-15
Print digest on registry push

### LINKED ISSUE #1267: Print digest after `spin registry push`
At the moment `spin registry push` succeeds silently (or, at trace level, logs the manifest); it should print the digest.

itowlson:
Fixes #1267.

```
$ spin registry push ghcr.io/itowlson/osf:v1.3
Pushed with digest sha256:05fe91ed63884099819ccd92e6d031999c62423fa7c2c8021b90d39d8adadfa1
```

@radu-matei Please confirm the logic is sound or if there is a better way! This was kind of me reverse engineering from the response...

image

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -42,8 +42,13 @@ impl Client {
         Ok(Self { oci: client, cache })
     }
 
-    /// Push a Spin application to an OCI registry.
-    pub async fn push(&mut self, app: &Application, reference: impl AsRef<str>) -> Result<()> {
+    /// Push a Spin application to an OCI registry and return the digest (or None
+    /// if the digest cannot be determined).
+    pub async fn push(
+        &mut self,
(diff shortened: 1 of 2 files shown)


lann:
If [the spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-manifests) is to be believed:

> The <location> is a pullable manifest URL. The Docker-Content-Digest header returns the canonical digest of the uploaded blob, and MUST be equal to the client provided digest. Clients MAY ignore the value but if it is used, the client SHOULD verify the value against the uploaded blob data.

It sounds like you'd be better off getting the digest from the `Docker-Content-Digest` header.

itowlson:
@michelleN @lann should this be cherry picked onto the 1.0 branch and if so what is the procedure for that (or do you folks have a magic wand for it)

PR #1274, opened 2023-03-15
fix wagi redirect status code

### LINKED ISSUE #1273: inconsistent status codes returned when using `wagi`
@karthik2804 reported this behavior where curl to same endpoint is returning different status codes intermittently:

image

<details><summary>Sample code</summary>
<p>

```
package main

import (
	"fmt"
	"os"
)

func main() {
	dest := os.Getenv("DESTINATION")
	status := os.Getenv("STATUS")
	if status != "" {
		fmt.Printf("status: %s\nlocation: %s\n\n", status, dest)
	} else {
		fmt.Printf("status: 302\nlocation: %s\n\n", dest)
	}
}
```

spin.toml
```
spin_version = "1"
authors = ["karthik Ganeshram <karthik.ganeshram@fermyon.com>"]
description = ""
name = "test"
trigger = { type = "http", base = "/" }
version = "0.1.0"

# Redirect /spin/contributing/ to spin/contributing-spin
[[component]]
id = "spin-contributing"
source = "main.wasm"
environment = { DESTINATION = "http://localhost:3000/wow", STATUS = "301"}
[component.trigger]
route = "/try"
executor = { type = "wagi" }
[component.build]
command = "tinygo build -target=wasi -gc=leaking -no-debug -o main.wasm main.go"

```
</p>
</details>

I believe the problem is:

1. This [fn](https://github.com/fermyon/spin/blob/88c7dbe987af72be7faf950df9af3a40fca02a72/crates/http/src/wagi/util.rs#L304) here returns a `HashMap`, so the order of headers is random here.
2. This [line](https://github.com/fermyon/spin/blob/88c7dbe987af72be7faf950df9af3a40fca02a72/crates/http/src/wagi/util.rs#L276) overrides the status code to `302` if a `location` header is specified. 

so depending on which header (`status` or `location`) comes last, the status code will be different.

rajatjindal:
attempt to fix #1273 

- it changes function `parse_cgi_headers` to return `Vec<(String, String)>` instead of `HashMap` (do we need to worry about duplicate headers)
- set status code in `location` block only if status code not set explicitly.

### DIFF
--- a/crates/http/src/wagi/util.rs
+++ b/crates/http/src/wagi/util.rs
@@ -242,6 +242,7 @@ pub fn compose_response(stdout: &[u8]) -> Result<Response<Body>, Error> {
     });
     let mut res = Response::new(Body::from(buffer));
     let mut sufficient_response = false;
+    let mut explicit_status_code = false;
     parse_cgi_headers(String::from_utf8(out_headers)?)
         .iter()
         .for_each(|h| {
@@ -258,6 +259,7 @@ pub fn compose_response(stdout: &[u8]) -> Result<Response<Body>, Error> {
                     // do not set content type correctly if a status is an error.
                     // See https://datatracker.ietf.org/doc/html/rfc3875#section-6.2
                     sufficient_response = true;
+                    explicit_status_code = true;
                     // Status can be `Status CODE [STRING]`, and we just want the CODE.
                     let status_code = h.1.split_once(' ').map(|(code, _)| code).unwrap_or(h.1);
                     tracing::debug!(status_code, "Raw status code");
@@ -273,7 +275,9 @@ pub fn compose_response(stdout: &[u8]) -> Result<Response<Body>, Error> {
                     sufficient_response = true;
                     res.headers_mut()
                         .insert(LOCATION, HeaderValue::from_str(h.1).unwrap());
-                    *res.status_mut() = StatusCode::from_u16(302).unwrap();
+                    if !explicit_status_code {
+                        *res.status_mut() = StatusCode::from_u16(302).unwrap();
+                    }
                 }
                 _ => {
                     // If the header can be parsed into a valid HTTP header, it is


crates/http/src/wagi/util.rs
```diff
@@ -301,17 +308,17 @@ pub fn compose_response(stdout: &[u8]) -> Result<Response<Body>, Error> {
     Ok(res)
 }
 
-fn parse_cgi_headers(headers: String) -> HashMap<String, String> {
-    let mut map = HashMap::new();
+fn parse_cgi_headers(headers: String) -> Vec<(String, String)> {
```
itowlson, line 311: I am not sure this change is needed.  The logic in the iterator seems (wisely) order-independent:

* If status comes out (of the Vec or HashMap) _before_ location, location sees that `explicit_status_code` is set, and respects it.
* If status comes out _after_ location, then status overrides the 302 that was set by location.
* If only one of them comes out then we don't have a problem.

Given that, I _think_ you can revert this and the `as_str` changes.  But I'm quite prepared to believe I'm missing something!

itowlson (changes requested):
This looks fine but I think some of the changes are not needed - but I am happy to be wrong!

crates/http/src/wagi/util.rs
rajatjindal: i am slightly confused with this statement in Spec:

```
The order in which header fields with differing field names are received is not significant. However, it is "good practice" to send general-header fields first, followed by request-header or response- header fields, and ending with the entity-header fields.

Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded.
```

Although I should also confess that I implemented this change as I thought headers order was required for the status code (because `golang` wanted me to call `WriteHeader(code)` before adding any other header). 

I can make this change tomorrow, but in the meantime, if anyone has the cycles to take this ticket over, I will be happy to close this PR and let other PR go through.

rajatjindal: Because we are working like a proxy between the response sent by the `wagi` module and the client, we should potentially preserve the order of headers when sending the response back?

rajatjindal: cc @lann @radu-matei what do you think?

lann: The CGI spec certainly implies that ordering matters, at least for the first header(s): https://datatracker.ietf.org/doc/html/rfc3875#section-6.2

```
CGI-Response = document-response | local-redir-response |
               client-redir-response | client-redirdoc-response

document-response = Content-Type [ Status ] *other-field NL
                    response-body

client-redir-response = client-Location *extension-field NL

client-redirdoc-response = client-Location Status Content-Type
                           *other-field NL response-body
```

For `client-redirdoc-response`:
> The Status header field MUST be supplied and MUST contain a status
> value of 302 'Found', or it MAY contain an extension-code, that is,
> another valid status code that means client redirection.

It seems ambiguous whether `document-response` is allowed to have a `Location` header, though certainly there are protocols out there that expect to send a `Location` with a `2xx` response.

itowlson (approved):
Awesome, thanks!

PR #1275, opened 2023-03-15
 ci: Add backport workflow 

lann:
This will allow merged PRs to be automatically backported to a release
branch via either a label (before the original PR is merged) or a
comment.

Uses:
https://github.com/marketplace/actions/backport-merged-pull-requests-to-selected-branches

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,32 @@
+name: Backport pull request
+on:
+  pull_request_target:
+    types: [closed]
+  issue_comment:
+    types: [created]
+permissions:
+  contents: write # so it can comment
+  pull-requests: write # so it can create pull requests
+jobs:
+  backport:
+    name: Backport pull request
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


itowlson:
Does the slash command need to say what branch to backport onto, or does it Just Know?

lann:
Hmm...for some reason I thought you would specify `/backport <branch>` but after a closer look I think you actually have to add the label and then trigger it with the comment (if you already merged).

itowlson:
I guess we need another action to add the label then. `</deadpan>`

michelleN (approved):
Looks awesome. Will wanna add some guidance to the release process around this once we've tested it out.

Also, we might want to consider that the `/backport` comment might be abused where as we can set permissions for who can add labels.

itowlson (approved):

lann:
> Also, we might want to consider that the /backport comment might be abused where as we can set permissions for who can add labels.

I don't think the action does anything that a user couldn't do manually, so we should be OK.

PR #1283, opened 2023-03-15
Unbindle the `up` command and app manifest

itowlson:
okay let's see what blows up in CI

### DIFF
--- a/crates/bindle/src/expander.rs
+++ b/crates/bindle/src/expander.rs
@@ -108,9 +108,6 @@ async fn bindle_component_manifest(
 
             sha256_digest(&full_path)?
         }
-        local_schema::RawModuleSource::Bindle(_) => {
-            return Err(PublishError::BindlePushingNotImplemented);
-        }
         local_schema::RawModuleSource::Url(us) => {
             let source = UrlSource::new(us)?;
             source.digest_str().to_owned()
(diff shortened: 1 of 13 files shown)


itowlson:
I have removed integration tests that depended on `spin up --bindle`.  These related to bindle round-tripping and are probably okay to get rid of.  However this does potentially undermine our test coverage for Cloud; if something goes wrong with the bindle push logic, these tests will no longer catch it.  Are we okay with that as the price for removing the bindle flags now?

There is an integration test `test_headers_env_routes` which is failing.  It currently relies on Bindle, but I am not sure what it is aiming to test or why Bindle is in the loop for it, so am reluctant to remove it.  Any ideas folks?

radu-matei:
cc @rajatjindal as to whether we could use the Cloud end to end test to make sure we don't undermine testing there.

radu-matei:
Re: the `test_headers_env_routes` test, it's probably the first end to end test that I wrote as part of the really early suite.
The rest of the tests should capture anything tested there, so we should be ok removing it.

rajatjindal:
> cc @rajatjindal as to whether we could use the Cloud end to end test to make sure we don't undermine testing there.

There was a plan to cover these tests as part of `Fermyon Platform` e2e tests, but with Bindle flags getting deprecated/removed, I don't see a lot of value in having these tests (assuming Fermyon Platform will also make use of OCI registry instead of Bindle in future).

`e2e-tests` currently deploy to Cloud, which uses Bindle behind the scenes. So we have some coverage but not specifically for the "spin bindle push".

rajatjindal:
> 


This test covers two scenarios:

1. environment variables are read from `spin.toml`.
2. env variables passed using `--env` during `spin-up`.

We have coverage for the first scenario for `spin up` and `Fermyon Cloud`. But for the second scenario, we have the test coverage on `spin up` only, as it is not supported on the Cloud yet.

itowlson:
Great stuff, thank you @radu-matei and @rajatjindal

lann (approved):

fibonacci1729 (approved):

PR #1290, opened 2023-03-16
[Backport v1.0] feat(deploy): Add deploy flags to seed key/values

vdice:
Backporting #1266 to v1.0

PR #1295, opened 2023-03-16
fix(cli): Rename --key-values to --key-value on deploy command

adamreese:
The flag `--key-values` only accepts a single key/value but can be used multiple times.  This changes the flag name and adds help text.


fibonacci1729 (approved):

michelleN:
needs backport after #1290

PR #1307, opened 2023-03-17
docs(release-process.md): updates

vdice:
- Updates from today's `v1.0.0-rc.2` pre-release with @melissaklein24 and @michelleN

Definitely would be neat to explore full (or more) automation, but first wanted to make sure the process was updated correctly.

Rendered file: https://github.com/fermyon/spin/blob/a8887380e6059d2a2a468a45dd95449035085c96/docs/content/release-process.md

### DIFF
--- a/docs/content/release-process.md
+++ b/docs/content/release-process.md
@@ -9,41 +9,77 @@ url = "https://github.com/fermyon/spin/blob/main/docs/content/release-process.md
 To cut a release of Spin, you will need to do the following:
 
 1. Create a pull request that changes the version number for your new version
-   (e.g. `0.10.0-pre0` could become either `0.9.1` for a patch release or
-   `0.10.0` for a "major" release)
+   (e.g. `1.1.0-pre0` could become either `1.0.1` for a patch release or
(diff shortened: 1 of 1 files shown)


docs/content/release-process.md
```diff
@@ -9,41 +9,77 @@ url = "https://github.com/fermyon/spin/blob/main/docs/content/release-process.md
 To cut a release of Spin, you will need to do the following:
 
 1. Create a pull request that changes the version number for your new version
-   (e.g. `0.10.0-pre0` could become either `0.9.1` for a patch release or
-   `0.10.0` for a "major" release)
+   (e.g. `1.1.0-pre0` could become either `1.0.1` for a patch release or
+   `1.1.0` for a "major" release)
```
michelleN, line 13: why major instead of minor?

vdice: Thanks for catching; updated to 'minor'.  Although I suppose we could also add the major release (eg 2.0.0) mention, I figured that will be a ways off and not a regular occurrence.

michelleN (approved):

PR #1313, opened 2023-03-21
Typed metadata

lann:
This adds a `MetadataKey<T>` type that gives stronger typing for metadata values.


crates/app/src/values.rs
```diff
@@ -71,30 +69,3 @@ impl ValuesMapBuilder {
         std::mem::take(&mut self.0)
     }
 }
-
-pub(crate) trait MetadataExt {
```
lann, line 75: Moved into `metadata.rs`

fibonacci1729 (approved):

PR #1320, opened 2023-03-22
ref(http): factor out common HTTP for reuse

fibonacci1729:
~WIP~

This factors out `spin_http::routes` into a distinct crate, `spin_http_common`. NOTE: I anticipate WAGI will be factored out for reuse and live in this "common" crate.


crates/http-trigger/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+[package]
+name = "spin-http-trigger"
```
lann, line 2: Bike-shedding, I prefer `spin-trigger-http` for two reasons:
- it matches the pattern of external trigger plugins, e.g. `spin-trigger-sqs`
- it shows the relationship with `spin-trigger` (and puts them next to each other in my editor file tree :sweat_smile:)

crates/http/src/routes.rs
```diff
@@ -47,37 +51,47 @@ impl Router {
...
 
-    /// This returns the component ID that should handle the given path, or an error
-    /// if no component matches.
-    ///
-    /// If multiple components could potentially handle the same request based on their
-    /// defined routes, components with matching exact routes take precedence followed
-    /// by matching wildcard patterns with the longest matching prefix.
-    pub(crate) fn route(&self, p: &str) -> Result<&str> {
+    /// Returns the constructed routes.
+    pub fn routes(&self) -> &IndexMap<RoutePattern, String> {
```
lann, line 55: It doesn't look like this method is used yet; whats the plan for it? If we can avoid exposing the inner map type it would be easier to change.

fibonacci1729: It's used by the `http-trigger` crate to print the available routes.

fibonacci1729: https://github.com/fermyon/spin/blob/12a038a3b29039dc31dc0a2d6752a8059a414d04/crates/http-trigger/src/lib.rs#L182

lann: Cool; could it be `-> impl Iterable<Item = (&RoutePattern, &String)>` or whatever `IndexMap::iter` returns?

lann: Ah I missed the "Large diffs are not rendered by default." among all the "This file was deleted." :sweat_smile:

fibonacci1729:
cc @lann I'm ready for my close up...._again_

crates/http/src/routes.rs
```diff
@@ -47,37 +51,47 @@ impl Router {
...
-    /// This returns the component ID that should handle the given path, or an error
-    /// if no component matches.
-    ///
-    /// If multiple components could potentially handle the same request based on their
-    /// defined routes, components with matching exact routes take precedence followed
-    /// by matching wildcard patterns with the longest matching prefix.
-    pub(crate) fn route(&self, p: &str) -> Result<&str> {
+    /// Returns the constructed routes.
+    pub fn routes(&self) -> impl Iterator<Item = (&RoutePattern, &String)> {
+        self.routes.iter()
```
lann, line 56: Probably needs some of this action:
```suggestion
        self.routes.iter().map(|(a, b)| (&a, &b))
```

fibonacci1729: Doesn't seem like that's necessary.

lann: Weird!

lann: I guess I'm thinking of `HashMap<String, String> -> Iterator<Item = (&str, &str)>`

lann (approved):

PR #1321, opened 2023-03-22
implement component support

dicej:
This updates Spin to use [the Component Model](https://github.com/WebAssembly/component-model/tree/main/design/high-level) natively instead of core Wasm modules (except for WAGI -- see below). Core modules are still supported transparently via [`spin-componentize`](https://github.com/fermyon/spin-componentize).

For now, the SDKs will continue to produce core modules (which the developer can convert to components ahead of time via `spin-componentize` if desired).  Eventually, we'll want the SDKs to target the Component Model directly as language toolchain support matures.  For the time being, though, we'll consider Component support experimental.

WAGI is stuck on Preview 1 and modules because there's no reliable way to wrap an arbitrary Preview 1 command in a component -- the Preview 1 -> 2 adapter only works with modules that either export `canonical_abi_realloc` (e.g. native Spin apps) or use a recent version of `wasi-sdk`, which contains patches to allow the adapter to safely allocate memory via `memory.grow`.

In theory, someone could build a WAGI app using a new-enough version of `wasi-sdk` and wrap it in a component using the adapter, but that wouldn't add any value beyond leaving it as a module, and any toolchain capable of natively producing components will be capable enough to produce native Spin apps, so we probably won't ever support WAGI components.

Note that I've moved the `spin-abi-conformance` project from this repo to `spin-componentize` since it seems more relevant (and more likely to be kept up-to-date) there.  We can move it back later if that changes.


dicej:
The `wit/preview2` files were copied from https://github.com/bytecodealliance/preview2-prototyping/tree/main/wit and augmented with Spin interfaces.

crates/core/src/store.rs
```diff
@@ -19,6 +20,28 @@ use crate::{
...
+/// support WAGI components.
+///
+// TODO: As of this writing, the plan is to merge the WASI Preview 1 and Preview 2 implementations together, at
+// which point we'll be able to avoid all the duplication here and below.
+pub struct Wasi {
+    /// Preview 1 `WasiCtx`
+    pub preview1: wasi_preview1::WasiCtx,
+    /// Preview 2 `WasiCtx`
+    pub preview2: wasi_preview2::WasiCtx,
+}
```
lann, line 43: It would be nice if we could make this an enum instead, switched by a flag to `StoreBuilder::new`. Mostly so that we could bail if the "wrong" version is used with features below.

Do we have any sense of the timeline for merging the implementations?

lann (commented):
Partial review

crates/core/src/store.rs
dicej: Yeah, I like the enum idea.  Not sure about the timeline for merging; I'll look into it.  It probably depends on when and how much we contribute to making it happen :)

dicej: I'm implementing the enum idea now.  It will also allow me to get rid of the `Clone` bounds I added in various places, so definitely an improvement.

Mossaka (commented):
Great work! Made some comments on the WIT interfaces

wit/preview2/inbound-http.wit
```diff
@@ -0,0 +1,5 @@
+default interface inbound-http {
```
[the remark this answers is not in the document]
dicej: There seems to be active work on `wasi-http` in multiple places, and I'll admit I haven't been keeping close track of it.  For the time being, I've done a simple translation of the Spin interfaces from old-style WIT to modern WIT.

We definitely will adopt `wasi-http` as it begins to mature, but I think that's best left for a later PR.

wit/preview2/key-value.wit
```diff
@@ -0,0 +1,80 @@
+default interface key-value {
```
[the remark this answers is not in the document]
dicej: Yes, agreed -- we plan to use WASI interfaces as much as possible going forward.  I don't think we're quite ready to adopt `wasi-keyvalue` at the moment (we don't yet support atomic or batch operations, for example), but it's definitely on the radar.

crates/core/src/store.rs
```diff
@@ -141,71 +177,102 @@ impl StoreBuilder {
...
-        self.with_wasi(|wasi| wasi.stdout(Box::new(file)))
+    /// Sets the WASI `stdout` descriptor to the given [`Write`]er.
+    pub fn stdout(&mut self, w: Box<dyn wasi_preview1::WasiFile>) -> Result<()> {
+        self.try_with_wasi(|wasi| match wasi {
+            Wasi::Preview1(ctx) => {
+                ctx.set_stdout(w);
+                Ok(())
+            }
+            Wasi::Preview2(_) => Err(anyhow!(
+                "`Store::stdout` only supported with WASI Preview 1"
```
lann, line 214: :+1:

crates/trigger-http/src/wagi.rs
```diff
@@ -117,17 +118,23 @@ impl HttpExecutor for WagiHttpExecutor {
...
+        store_builder.stdout(Box::new(stdout.clone()))?;
 
         let (instance, mut store) = engine
             .prepare_instance_with_store(component, store_builder)
             .await?;
+        let instance = if let EitherInstance::Module(instance) = instance {
+            instance
+        } else {
+            unreachable!()
+        };
```
lann, line 137: :shrug: 
```suggestion
        let EitherInstance::Module(instance) = instance else {
            unreachable!()
        };
```

lann (approved):

PR #1327, opened 2023-03-25
Temporarily remove spin watch from templates to prevent breaking older versions

calebschoepp:


fibonacci1729:
@calebschoepp are we OK to close this?

calebschoepp:
Yep

PR #1331, opened 2023-03-28
fix `spin-timer` example and add a test so it stays fixed

dicej:
This primarily required updating both the trigger and the `app-example` that exercises it to use the latest, component-oriented, binding generators.

Eventually, I'd like to see _all_ the examples tested (or at least built) via either integration or end-to-end tests.  Consider this the first step in that direction.

### DIFF
--- a/examples/spin-timer/src/main.rs
+++ b/examples/spin-timer/src/main.rs
@@ -1,14 +1,21 @@
 use std::collections::HashMap;
 
 use anyhow::Error;
-use clap::Parser;
+use clap::{Args, Parser};
 use serde::{Deserialize, Serialize};
 use spin_app::MetadataKey;
-use spin_trigger::{cli::TriggerExecutorCommand, TriggerAppEngine, TriggerExecutor};
+use spin_core::async_trait;
+use spin_trigger::{
+    cli::TriggerExecutorCommand, EitherInstance, TriggerAppEngine, TriggerExecutor,
+};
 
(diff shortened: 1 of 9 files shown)


fibonacci1729 (approved):

lann (approved):

examples/spin-timer/src/main.rs
```diff
@@ -103,9 +123,11 @@ impl TimerTrigger {
     async fn handle_timer_event(&self, component_id: &str) -> anyhow::Result<()> {
         // Load the guest...
         let (instance, mut store) = self.engine.prepare_instance(component_id).await?;
-        let engine = spin_timer::SpinTimer::new(&mut store, &instance, |data| data.as_mut())?;
+        let EitherInstance::Component(instance) = instance else {
```
itowlson, line 126: Does this mean the SQS trigger needs to be updated too?

dicej: Yes.  If it's supposed to work with the main branch of Spin, we should update it now.  If it's only supposed to work with the latest stable release, we have until Spin v1.1 is released.  Either way, I'm happy to do it.

itowlson: We can defer until 1.1 for sure.

@lee-orr a heads up for you that this will affect https://github.com/lee-orr/spin-message-trigger as well (not immediately but just so you know it's coming)

lee-orr: Thanks for the update!

PR #1333, opened 2023-03-28
New readme

melissaklein24:
Updated README with additional example, language support table, and other updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,85 +99,21 @@ When handling a request, Spin will create a new isolated Wasm instance correspon
 
 ## Language Support for Spin Features
 
-### Rust
-
-| Feature | SDK Supported? |
-|-----|-----|
+| Feature | Rust SDK Supported? | TypeScript SDK Supported? | Python SDK Supported? | Tiny Go SDK Supported? | C# SDK Supported? |
+|-----|-----|-----|-----|-----|-----|
 | **Triggers** |
-| [HTTP](https://developer.fermyon.com/spin/http-trigger) | Supported |
-| [Redis](https://developer.fermyon.com/spin/redis-trigger) | Supported |
+| [HTTP](https://developer.fermyon.com/spin/http-trigger) | Supported | Supported | Supported | Supported | Supported |
+| [Redis](https://developer.fermyon.com/spin/redis-trigger) | Supported | Not Supported | Not Supported | Supported | Not Supported |
 | **APIs** |
-| [Outbound HTTP](https://developer.fermyon.com/spin/rust-components.md#sending-outbound-http-requests) | Supported |
-| [Key Value Storage](https://developer.fermyon.com/spin/kv-store.md) | Supported |
-| [MySQL](https://developer.fermyon.com/spin/rdbms-storage#using-mysql-and-postgresql-from-applications) | Supported |
-| [PostgreSQL](https://developer.fermyon.com/spin/rdbms-storage#using-mysql-and-postgresql-from-applications) | Supported |
-| [Outbound Redis](https://developer.fermyon.com/spin/rust-components.md#storing-data-in-redis-from-rust-components) | Supported |
+| [Outbound HTTP](https://developer.fermyon.com/spin/rust-components.md#sending-outbound-http-requests) | Supported | Supported | Supported | Supported | Supported |
+| [Key Value Storage](https://developer.fermyon.com/spin/kv-store.md) | Supported | Supported | Supported | Supported | Not Supported |
+| [MySQL](https://developer.fermyon.com/spin/rdbms-storage#using-mysql-and-postgresql-from-applications) | Supported | Not Supported | Not Supported | Not Supported | Not Supported |
+| [PostgreSQL](https://developer.fermyon.com/spin/rdbms-storage#using-mysql-and-postgresql-from-applications) | Supported | Not Supported | Not Supported | Not Supported | Supported |
+| [Outbound Redis](https://developer.fermyon.com/spin/rust-components.md#storing-data-in-redis-from-rust-components) | Supported | Supported | Supported | Supported | Supported |
 | **Extensibility** |
-| [Authoring Custom Triggers](https://developer.fermyon.com/spin/extending-and-embedding) | Supported |
+| [Authoring Custom Triggers](https://developer.fermyon.com/spin/extending-and-embedding) | Supported | Not Supported | Not Supported | Not Supported | 
 
-### TypeScript
 
-| Feature | SDK Supported? |
-|-----|-----|
-| **Triggers** |
-| [HTTP](https://developer.fermyon.com/spin/javascript-components#http-components) | Supported |
-| Redis | Not Supported |
-| **APIs** |
-| [Outbound HTTP](https://developer.fermyon.com/spin/javascript-components#sending-outbound-http-requests) | Supported |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,21 +34,93 @@ $ ./spin --help
...
+Below is an example of using the `spin` CLI to create a new Spin Python application, then adding a JavaScript component:
+
+```bash
+# Create a new Spin application based on the Python language template.
+$ spin new http-py hello-python
+# Add a new JavaScript component based on the language template.
+$ spin add http-js goodbye-javascript
+```
+
+Running the `spin add` command will generate the proper configuration for our component and add it to the [`spin.toml` manifest file](https://developer.fermyon.com/spin/manifest-reference). For example, here is the `spin.toml` section for our Python component:
```
radu-matei, line 49: This looks great!
Only comment I have here is: do we want to include a sample source code ahead of the config?

radu-matei (commented):
The readme looks great, I have one non-blocking question about it.

You need to rebase the PR against the newest main to avoid the conflicts and old commits.

README.md
```diff
@@ -37,21 +34,93 @@ $ ./spin --help
...
+# Send a request to the JavaScript component.
+$ curl localhost:3000/goodbye
+Goodbye, JavaScript!
 ```
 
+When handling a request, Spin will create a new isolated Wasm instance corresponding to the Wasm module for the matching component, execute the handler function, then terminate the instance. Each new request will get a fresh Wasm instance.
+
+## Language Support for Spin Features
+
+The table below shows [feature support](https://developer.fermyon.com/spin/language-support-overview) in each of the language SDKs.
```
itowlson, line 99: I worry about how well maintained this will be over time.  Do we need it in the repo read-me?  Could we not link to the docs?

README.md
```diff
@@ -37,21 +34,93 @@ $ ./spin --help
 
 After you follow the [quickstart document](https://developer.fermyon.com/spin/quickstart/),
 you can follow the
-[Rust](https://developer.fermyon.com/spin/rust-components/) or [Go](https://developer.fermyon.com/spin/go-components/)
+[Rust](https://developer.fermyon.com/spin/rust-components/), [JavaScript](https://developer.fermyon.com/spin/javascript-components), [Python](https://developer.fermyon.com/spin/python-components), or [Go](https://developer.fermyon.com/spin/go-components/)
```
itowlson, line 37: I can't comment on unchanged lines but the install instructions we show on lines 28-30 are different from those in the docs.

README.md
```diff
@@ -37,21 +34,93 @@ $ ./spin --help
...
 language guides, and the [guide on configuring Spin applications](https://developer.fermyon.com/spin/configuration/).
 
-After you build your application, run it using Spin:
+Below is an example of using the `spin` CLI to create a new Spin Python application, then adding a JavaScript component:
```
itowlson, line 40: This sounds like creating a Python app and kind of linking in a JS module.  We don't really have good terminology for this - we loosely use "Python app" or "Rust app" for monolingual applications, but what you describe here is a multilingual app that has one component in Python and one in JS, and I don't know how to express that concisely.

My suggestion would be to show a single-component app, which gives more room to include the questions/responses from the `spin new` command, then link to the docs for more info.

README.md
itowlson, line 49: The manifest reference is a pretty heavyweight place to drop a reader into.  What do you think of linking to the "Writing Applications" page, which covers a lot of manifest authoring but is hopefully more readable?

README.md
itowlson, line 37: Also, on line 22, as this is the code repo rather than the docs, maybe we should more clearly tease apart the "for application developers" info (which should mostly be covered by the docs) and the "for people interested in the source code" info (which might point to the contributor guides and the SIPs).

README.md
```diff
@@ -37,21 +34,93 @@ $ ./spin --help
...
+# The route for this component.
+route = "/hello"
+[component.build]
+# The command to execute for this component with `spin build`.
+command = "spin py2wasm app -o app.wasm"
+# The working directory for the component.
+workdir = "hello-python"
 ```
+
+We can now build our application with `spin build`, then run it locally with `spin up`:
```
itowlson, line 67: This skips the plugin installation steps, so anyone using this as a quick start is going to get weird errors... again my suggestion would be to link to the docs rather than trying to write a separate quickstart here.  Could we just hit the highlights here, in a way that is visibly "here is the 30000 ft view" rather than looking like a tutorial but actually being incomplete?

README.md
```diff
@@ -37,21 +34,93 @@ $ ./spin --help
...
+# The Wasm module to instantiate and execute when receiving a request.
+source = "hello-python/app.wasm"
+[component.trigger]
+# The route for this component.
+route = "/hello"
+[component.build]
+# The command to execute for this component with `spin build`.
+command = "spin py2wasm app -o app.wasm"
+# The working directory for the component.
+workdir = "hello-python"
```
itowlson, line 64: This isn't what they'll see with the commands above.

itowlson (changes requested):

README.md
melissaklein24: I also have my concerns about maintaining this list.  The big benefit I see to keeping this in is the list of features.  We do not have a straight list of features anywhere else and if we list them, we should also set expectation on the language support.  Hopefully we will get to the point where all the features are supported by all SDKs.

README.md
melissaklein24: I revamped the example to make it simpler. It is now just a quick spin new, spin build, spin up.

README.md
melissaklein24: I updated the installation commands to match the installation guide and linked to it.

README.md
melissaklein24: I revamped the example to make it simpler. It is now just a quick spin new, spin build, spin up.

README.md
melissaklein24: I revamped the example to make it simpler. It is now just a quick spin new, spin build, spin up.

README.md
melissaklein24: I revamped the example to make it simpler. It is now just a quick spin new, spin build, spin up.

README.md
```diff
@@ -24,34 +24,92 @@ applications and how to get started.
...
-[Rust](https://developer.fermyon.com/spin/rust-components/) or [Go](https://developer.fermyon.com/spin/go-components/)
-language guides, and the [guide on configuring Spin applications](https://developer.fermyon.com/spin/configuration/).
+To get started writing apps, follow the [quickstart guide](https://developer.fermyon.com/spin/quickstart/),
+and then follow the
+[Rust](https://developer.fermyon.com/spin/rust-components/), [JavaScript](https://developer.fermyon.com/spin/javascript-components), [Python](https://developer.fermyon.com/spin/python-components), or [Go](https://developer.fermyon.com/spin/go-components/)
+language guides, and the [guide on writing Spin applications](https://developer.fermyon.com/spin/configuration/).
 
-After you build your application, run it using Spin:
+## Usage
+Below is an example of using the `spin` CLI to create a new Spin application.  To run the example you will need to install the `wasm-wasi` target for Rust.
```
itowlson, line 42:
```suggestion
Below is an example of using the `spin` CLI to create a new Spin application.  To run the example you will need to install the `wasm32-wasi` target for Rust.
```

itowlson (approved):

PR #1338, opened 2023-03-30
Add KV to list of interfaces to provide SDK version warning

itowlson:
There is a magic function deep in Spin that is meant to provide a more meaningful warning if someone uses certain mismatched SDKs (I _think_ it is new SDK on old Spin but have long since forgotten).  To do this it looks at the interface name in the import error message.  This PR adds the key-value interface to that list.

(We may want to revisit whether this function still has value - old SDK/new Spin seems much more likely now.  But this PR doesn't take a view on that.)

### DIFF
--- a/crates/trigger/src/lib.rs
+++ b/crates/trigger/src/lib.rs
@@ -323,13 +323,8 @@ pub fn decode_preinstantiation_error(e: anyhow::Error) -> anyhow::Error {
 
     if err_text.contains("unknown import") && err_text.contains("has not been defined") {
         // TODO: how to maintain this list?
-        let sdk_imported_interfaces = &[
-            "outbound-pg",
-            "outbound-redis",
-            "spin-config",
-            "wasi_experimental_http",
-            "wasi-outbound-http",
-        ];
+        let sdk_imported_interfaces =
+            &["config", "http", "key-value", "mysql", "postgres", "redis"];
 
         if sdk_imported_interfaces
             .iter()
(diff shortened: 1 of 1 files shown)


dicej:
As part of the `spin-componentize` effort, I took the liberty of renaming some of the interfaces when porting them to component-land, based on comments I had heard from others that the `outbound-` prefix was annoying and the `wasi-outbound-` prefix both misleading and annoying.  Here are the new names:

https://github.com/fermyon/spin/blob/3ee52a032b4785e503737db2fc28ca20786cd569/wit/preview2/reactor.wit#L22-L29

Those new names are the ones Spin will be looking for, so I believe those are the names we should use in this list as well.

itowlson:
Aha, thanks!  How will that affect modules that were built against the old import names?

dicej:
Old modules are wrapped in components by `spin-componentize` (i.e. they use the old names, but the adapter wraps them in an interface which uses the new names).  I guess if we want to get fancy, we could detect whether the guest was originally a module and translate the new name in the error message to the old name prior to printing it.

itowlson:
That seems a bit _too_ fancy!  Sounds like the "unresolved import of an old name" could never happen because the list of old names is now closed, and the adapter implements all of them.  So I have updated to the new names, with increasing scepticism that it will ever matter anyway...!

dicej (approved):

fibonacci1729 (approved):

PR #1339, opened 2023-03-31
feat(sdk): add HTTP router for Rust SDK

fibonacci1729:
NOTE: I alias the `http` crate to `http_types` because it's inconvenient to submodule our crate `http` mod due to the resulting name collisions.


sdk/rust/src/lib.rs
```diff
@@ -28,23 +28,26 @@ pub mod http {
...
-    pub type Request = http::Request<Option<bytes::Bytes>>;
+    pub type Request = http_types::Request<Option<bytes::Bytes>>;
 
     /// The Spin HTTP response.
-    pub type Response = http::Response<Option<bytes::Bytes>>;
+    pub type Response = http_types::Response<Option<bytes::Bytes>>;
 
     pub use crate::outbound_http::send_request as send;
 
+    pub use router::*;
```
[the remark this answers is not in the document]
radu-matei: I see the router macro is exposed as `spin_sdk::router`, and the router struct is `spin_sdk::http::Router` — any reason for the different structure here?

radu-matei (commented):
This is GREAT!

I have a small comment re: structure, but overall, I love the experience!

sdk/rust/src/lib.rs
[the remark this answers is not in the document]
fibonacci1729: This is a side-effect of how you export macros via `#[macro_export]`; it seems to export it from the crate root. Alternatively i could name it something like `http_router` if that is a bit more descriptive for a crate-level symbol. Wyt?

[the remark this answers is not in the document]
fibonacci1729: Renamed to `http_router` and added an example using the macro.

examples/http-rust-router/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+    router.all("/*", api::echo_wildcard);
+    router.handle(req)
+}
+
+mod api {
+    use super::*;
+
+    // /hello/:planet
+    pub fn hello_planet(_req: Request, params: Params) -> Result<Response> {
+        let planet = params.get("planet").expect("PLANET");
```
rylev, line 26: Many other routers try to autoextract params so that there's no need for this runtime checking which shouldn't fail. Was such an approach considered?

sdk/rust/src/http/router.rs
```diff
@@ -0,0 +1,339 @@
...
+
+type Handler = dyn Fn(Request, Params) -> Result<Response>;
+
+/// Route parameters extracted from a URI that match a route pattern.
+pub type Params = Captures<'static, 'static>;
+
+/// The Spin SDK HTTP router.
+pub struct Router {
+    methods_map: HashMap<http_types::Method, MethodRouter<Box<Handler>>>,
+    all_methods: MethodRouter<Box<Handler>>,
```
rylev, line 13: nit: `any_methods` might be clearer?

sdk/rust/src/http/router.rs
```diff
@@ -0,0 +1,339 @@
...
+                        handler: &not_found,
+                        params: Captures::default(),
+                    }
+                }
+            }
+        }
+    }
+
+    /// Register a handler at the path for all methods.
+    pub fn all<F>(&mut self, path: &str, handler: F)
```
rylev, line 86: Same as above: for some reason, `any` parses more easily than `all` for me.

rylev (commented):
LGTM! So other HTTP frameworks have fancier routing where params are auto-extracted and checked at compile time, but the simplicity of how this router works is appealing.

examples/http-rust-router/src/lib.rs
fibonacci1729: Definitely considering it! I have been imagining that such an implementation could be layered over what has been provided here in a follow up iteration.  Is this something you are interested in exploring?

rylev (approved):

PR #1342, opened 2023-04-01
Add simple Go router based on github.com/julienschmidt/httprouter

radu-matei:
This commit adds a simple Go router, based on https://github.com/julienschmidt/httprouter, which is a zero dependency, high performance router.

This PR also adds type alias for the most commonly used types from the package and a simple example for using the router.


examples/http-tinygo-router/README.md
```diff
@@ -0,0 +1,74 @@
...
+$ spin build
+Executing the build command for component tinygo-hello: tinygo build -target=wasi -gc=leaking -no-debug -o main.wasm main.go
+Successfully ran the build command for the Spin components.
+```
+
+Finally, we can create a Spin application configuration to execute this
+component:
+
+```toml
+apiVersion = "0.1.0"
```
fibonacci1729, line 42:
```suggestion
spin_manifest_version = "1"
```

fibonacci1729 (approved):

radu-matei:
Example of using the router from this PR: https://github.com/radu-matei/spin-kv-explorer/blob/feat/golang/golang-explorer/main.go

.gitignore
```diff
@@ -6,6 +6,7 @@ ignored-assets
 main.wasm
 .parcel-cache
 .vscode/*.log
+go.sum
```
radu-matei, line 9: cc @adamreese — is adding `go.sum` to `.gitignore` the right thing to do here?
I mainly want to make sure we're not polluting the repo with changes here.

examples/http-tinygo-router/README.md
radu-matei: That's the readme of the example, I'd like to update all of them in one go.

PR #1344, opened 2023-04-03
Add /.well-known/spin/locked endpoint

lann:
Returns the locked app JSON for the running app.


mikkelhegn:
Will this be exposed always? And on which path, if so? Or is the idea that we need a switch in the CLI to enable it?

lann:
Hmm...I was only thinking about it in terms of local development but its true that it shouldn't be always-on. Probably ought to look at a different way of doing this.

mikkelhegn:
I like the idea for debug. Opening an issue to capture this.

PR #1346, opened 2023-04-03
Add --dump-locked flag to `spin up`

### LINKED ISSUE #1345: Expose the spin.lock file on debug endpoint for local development
#1344 was a quick PoC for how to expose the spin.lock file for an http trigger, but there is probably more to this.

It would be great to have a way to enable a debug endpoint, maybe for all triggers, so you can easily see the content of the lockfile while debugging you Spin app.

lann:
The `--dump-locked` flag dumps the generated "locked app" file to the given path, e.g. `spin up --dump-locked spin.lock`.

Closes #1345

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -1,6 +1,7 @@
 use std::{
     ffi::OsString,
     fmt::Debug,
+    fs,
     path::{Path, PathBuf},
 };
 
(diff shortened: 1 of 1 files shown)


lann:
@mikkelhegn Dumping to "`spin.lock` next to `spin.toml`" was a little tricky because of `spin up --from-registry`

fibonacci1729 (approved):

itowlson (changes requested):
The code is fine but I'd like to spend a bit more time iterating on the UI before we go ahead with this.

lann:
I have been reminded that you can already get this (albeit slightly more akwardly) with `--tmp <dir>` (which will place it in `<dir>/spin.lock`). Given the internalsiness of this option I think thats sufficient.

PR #1351, opened 2023-04-04
clear all env vars except PATH when running cargo

dicej:
This is necessary to avoid confusing rustc when running cargo-tarpaulin on Mac and Windows using --engine llvm.


dicej:
This appears to cause DNS resolution failures on Windows, so I'm closing it.  We can revisit later if `cargo-tarpaulin` is a priority, but right now it's more trouble than it's worth.

MahdiAl-Qubli:
M

PR #1352, opened 2023-04-04
sip: Add runtime config section to Key-Value SIP

lann:


dicej (approved):

Mossaka:
Do you want to account for user-provided environment variables that specify conection to the keyvalue store? 

An example would be the azure blob storage config in spiderlightning https://github.com/deislabs/spiderlightning/blob/main/examples/keyvalue-demo/keyvalue_azblob_slightfile.toml#L14-L15

kate-goldenring (approved):

radu-matei:
Environment variables _or_ the Spin configuration system?

radu-matei (approved):

lann:
Injecting config variables into KV store config is definitely on my radar, but there are some operational subtleties to consider around access control, e.g. if an SRE wanted to use a Vault config provider to set my Redis KV password, should that config value be available in app configs as well? Seems like it shouldn't need to be, but we either need a way to deal with that scenario or decide its ok.

Mossaka:
> Environment variables _or_ the Spin configuration system (or both)?

I am not familiar with spin configuration system. What's that? 

> if an SRE wanted to use a Vault config provider to set my Redis KV password, should that config value be available in app configs as well?

Forgive my ignorance on this field, but couldn't the Vault config value be injected as environment variables to the host and thus be feeded into the app configs which specify what env vars are needed?

lann:
> I am not familiar with spin configuration system. What's that?

https://developer.fermyon.com/spin/dynamic-configuration

lann:
> couldn't the Vault config value be injected as environment variables to the host and thus be feeded into the app configs which specify what env vars are needed?

Yes, but I'm thinking of a scenario where you don't want a secret exposed to an app. Say if an organization has an operations group responsible for administering infrastructure that wants to provide a Redis-backed KV implementation to an applications group, but doesn't want to expose the Redis password to applications.

michelleN (approved):
Should we add an Updated date to this doc?

PR #1357, opened 2023-04-05
Remove unused 'namespace' field from ApplicationInformation

rylev:
It appears that this field is no longer used. We leave the `namespace` fields in the "raw" structs as technically removing those fields could break users since we're using `deny_unknown_fields`.


lann (approved):
FYI, I'm hoping to remove a lot of this code entirely after we drop bindle support.

radu-matei (approved):
Oh yes, "this looks like it could be useful in the future, let's add it" coming back to haunt me...


LGTM!

PR #1358, opened 2023-04-05
Show error source list when printing handler error

rylev:
This changes the Rust SDK to print out the causes of any error returned from the HTTP handler. This can make debugging much easier. 

For example, take the following error handling code in an HTTP handler:

```rust
let query: Query = serde_qs::from_str(query_string).context("failed to parse query parameters")?;
```

Previously this would print out: 

```
Handler returned an error: failed to parse query parameters
```

Now this prints out:

```
Handler returned an error: failed to parse query parameters
  caused by: missing field `email`
```

Now it is immediately clear that the cause is the missing email parameter. 

This can be further improved with some creative use of emoji and terminal colors, but that should be left to a future improvement.


itowlson (approved):
approved on the strict condition that you do _not_ "improve it with the creative use of emoji"

fibonacci1729 (approved):

radu-matei (approved):

PR #1360, opened 2023-04-05
Key value store runtime config WIP

lann:


fibonacci1729:
Was using this to test #1365 and it works great! Given that PR is in draft now, does it make sense to include the following configuration here in this PR or a follow up?

```
#[derive(Debug, Default, Deserialize)]
pub struct RedisKeyValueStore {
    pub url: String,
}
```

lann:
> does it make sense to include the following configuration here in this PR or a follow up?

Follow-up

crates/trigger/src/key_value.rs
lann, line 1: Moved under `runtime_config/`

crates/trigger/src/runtime_config.rs
```diff
@@ -27,47 +33,63 @@ impl RuntimeConfig {
...
         providers.extend(self.opts_layers().flat_map(|opts| {
             opts.config_providers
                 .iter()
-                .map(Self::build_config_provider)
+                .map(|opts| opts.build_provider())
         }));
         providers
     }
 
-    fn build_config_provider(provider_config: &ConfigProvider) -> BoxedConfigProvider {
```
lann, line 62: Moved to `config_provider::ConfigProviderOpts::build_provider`

crates/loader/src/validation.rs
```diff
@@ -1,19 +1,37 @@
...
-        .iter()
-        .flatten()
-        .find(|id| *id != ONLY_ALLOWED_KV_STORE)
-    {
-        None => Ok(()),
-        Some(invalid) => {
-            let err = anyhow!("Invalid key-value store '{invalid}'. This version of Spin supports only the '{ONLY_ALLOWED_KV_STORE}' store.");
-            Err(err)
-        }
+    for store in key_value_stores.iter().flatten() {
```
itowlson, line 4: Non-default stores can't currently be deployed to cloud, I think, so either cloud needs to reject them with a meaningful message or we need to keep the old validation on the `deploy` path.

crates/loader/src/validation.rs
```diff
@@ -1,19 +1,37 @@
...
+        validate_component_like_label(store)
+            .with_context(|| format!("Invalid store label {store:?}"))?;
+    }
+    Ok(())
+}
+
+// Validates that the given string is like a component model label, except (currently)
```
itowlson, line 11: These rules are very specific (I was especially struck by "it must all all uppercase or all lowercase, no mixing") - is there a particular reason that drives them?  Might be good to comment it if so.  (Or is that what the allusion to the component model is aiming to say?)

lann: This attempts to match the `label` rule [from the component model design](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#instance-definitions) to give us forward-compatibility with value import names (once those are implemented).

crates/loader/src/validation.rs
```diff
@@ -1,19 +1,37 @@
...
-        .flatten()
-        .find(|id| *id != ONLY_ALLOWED_KV_STORE)
-    {
-        None => Ok(()),
-        Some(invalid) => {
-            let err = anyhow!("Invalid key-value store '{invalid}'. This version of Spin supports only the '{ONLY_ALLOWED_KV_STORE}' store.");
-            Err(err)
-        }
+    for store in key_value_stores.iter().flatten() {
+        validate_component_like_label(store)
```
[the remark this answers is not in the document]
lann: Yeah that would be quite awkward to manage here. It could make sense to add a new e.g. `DynamicHostComponent::validate_app(&self, app: &App) -> Result<()>` method that could be plumbed through to validate things like that, called from `spin-trigger`.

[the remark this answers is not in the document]
lann: https://github.com/fermyon/spin/issues/1371

crates/loader/src/validation.rs
lann: Expanded comment

crates/loader/src/validation.rs
lann: This really ought to be checked in cloud.

crates/loader/src/validation.rs
```diff
@@ -1,19 +1,38 @@
...
+            .with_context(|| format!("Invalid store label {store:?}"))?;
```
rylev, line 6: Nit:
```suggestion
            .with_context(|| format!("invalid store label {store:?}"))?;
```

crates/trigger/src/runtime_config.rs
```diff
@@ -132,34 +144,38 @@ impl RuntimeConfig {
...
+        return Ok(path.to_owned());
+    }
+    let base_path = match &config_opts.file_path {
+        Some(file_path) => file_path
+            .parent()
+            .with_context(|| {
+                format!("failed to get parent of runtime config file path {file_path:?}")
+            })?
+            .to_owned(),
+        None => std::env::current_dir().context("failed to get current directory")?,
```
rylev, line 175: Nit: this error message just repeats the underlying error. We should provide context about runtime config like we do in the lines above.

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -0,0 +1,107 @@
...
+// Holds deserialized options from a `[key_value_store.<name>]` runtime config section.
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "snake_case", tag = "type")]
+pub enum KeyValueStoreOpts {
+    Spin(SpinKeyValueStoreOpts),
+
+    // TODO: Remove when adding next store variant
+    #[serde(skip)]
+    #[allow(dead_code)]
+    Unreachable,
```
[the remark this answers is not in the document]
rylev: Is this trying to do the same thing as [non_exhaustive attribute](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute)?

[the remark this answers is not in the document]
lann: Apparently WAI: https://github.com/rust-lang/rust/issues/72469

I guess it makes sense; similar to dead_code. :shrug:

crates/trigger/src/runtime_config.rs
lann: > this error message just repeats the underlying error

Not sure what you mean here; afaict `current_dir` just returns an `io::Error` without any calling context?

crates/trigger/src/runtime_config/key_value.rs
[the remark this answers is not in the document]
itowlson: In that case I'd suggest a comment in the code explaining this reasoning, and renaming the variant to `DeleteMeWhenThereIsASecondCase` or `DeleteMeWhenRedisLands` or `OnlyHereToSuppressIrrefutableMatchLints`.  The current comment gives an instruction but doesn't explain, which for me at least always makes me anxious "have I correctly understood the words of mighty Apollo? or should I just leave it for next time? eh I'll leave it, somebody who understands will be along in a minute".

radu-matei (approved):
Tested the local runtime configuration with alternative default database path and everything LGTM!

crates/loader/src/validation.rs
macolso: @bacongobbler - do we have logic to check store type in Nimbus?

lann: This issue is just with the store _name_: https://linear.app/fermyon/issue/NIM-594/validate-app-manifest-key-value-stores-on-revision-creation

lann: https://github.com/fermyon/spin/pull/1389

PR #1361, opened 2023-04-06
add spin sdk based http client

rajatjindal:
This PR helps setting up a `http.Client` backed by spin-sdk. We can then pass this to any 3rd party lib, which will make use of the `RoundTrip` provided by the transport in this client.

e.g. to make use of existing Twitter sdk, I could do following:

```go
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/dghubble/go-twitter/twitter"
	"github.com/dghubble/oauth1"

	spinhttp "github.com/fermyon/spin/sdk/go/http"
)

func init() {
	spinhttp.Handle(func(w http.ResponseWriter, r *http.Request) {
		config := oauth1.NewConfig("consumerKey", "consumerToken")
		token := oauth1.NewToken("token", "tokenSecret")
		
		spinhttpclient := spinhttp.NewClient()
		ctx := context.WithValue(oauth1.NoContext, oauth1.HTTPClient, spinhttpclient)
		httpclient := config.Client(ctx, token)
		client := twitter.NewClient(httpclient)

		_, _, err := client.Statuses.Update("test tweet, pls ignore !", nil)
		if err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		}

		w.Header().Set("Content-Type", "text/plain")
		fmt.Fprintln(w, "Hello Fermyon!")
	})
}

```

### DIFF
--- a/sdk/go/http/http.go
+++ b/sdk/go/http/http.go
@@ -34,6 +34,26 @@ func NewRouter() *Router {
 	return httprouter.New()
 }
 
+// NewTransport returns http.RoundTripper backed by Spin SDK
+func NewTransport() http.RoundTripper {
+	return &Transport{}
+}
+
+// Transport implements http.RoundTripper
+type Transport struct{}
+
+// RoundTrip makes roundtrip using Spin SDK
+func (r *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	return Send(req)
+}
+
+// NewClient returns a new HTTP client compatible with the Spin SDK
+func NewClient() *http.Client {
+	return &http.Client{
+		Transport: &Transport{},
+	}
+}
+
 // handler is the function that will be called by the http trigger in Spin.
 var handler = defaultHandler
 
--- a/sdk/go/http/testdata/spin-roundtrip/Makefile
+++ b/sdk/go/http/testdata/spin-roundtrip/Makefile
@@ -1,0 +1,3 @@
+.PHONY: build
+build:
+	tinygo build -target=wasi -gc=leaking -o main.wasm main.go
--- a/sdk/go/http/testdata/spin-roundtrip/main.go
+++ b/sdk/go/http/testdata/spin-roundtrip/main.go
@@ -1,0 +1,32 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+
+	spinhttp "github.com/fermyon/spin/sdk/go/http"
+)
+
+func init() {
+	spinhttp.Handle(func(w http.ResponseWriter, r *http.Request) {
+		transport := spinhttp.NewTransport()
+
+		// or you can also do client := spinhttp.NewClient()
+		client := &http.Client{
+			Transport: transport,
+		}
+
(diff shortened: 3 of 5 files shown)


sdk/go/http/http.go
```diff
@@ -9,6 +9,23 @@ import (
 	"os"
 )
 
+func NewTransport() http.RoundTripper {
```
rajatjindal, line 12: happy to rename struct/functions if there are suggestions here

sdk/go/http/http.go
```diff
@@ -9,6 +9,23 @@ import (
...
+}
+
+// Transport implements http.RoundTripper
+type Transport struct{}
+
+func (r *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	return Send(req)
+}
+
+func NewClient() *http.Client {
```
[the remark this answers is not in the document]
rajatjindal: sure, done

rajatjindal:
> As far as I can tell, this depends on the current `dev` branch of TinyGo.
> 
> Once there is a new release, this looks good to me, tested with a few client libraries that had configurable clients and everything LGTM!

on second thought, that might not be the case. as in the testdata app `spin-roundtrip` we are exercising this functionality already in our ci (which runs `v0.27.0` of `tinygo`)

radu-matei:
OH, you're right!

radu-matei (approved):

sdk/go/http/http.go
```diff
@@ -34,6 +34,26 @@ func NewRouter() *Router {
...
+
+// Transport implements http.RoundTripper
+type Transport struct{}
+
+// RoundTrip makes roundtrip using spin-go sdk
+func (r *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
+	return Send(req)
+}
+
+// NewClient returns spin-go sdk backed *http.Client
```
radu-matei, line 50: nit: we never refer to the Go SDK as `spin-go`.
suggestion: `NewClient returns a new HTTP client compatible with the Spin SDK` maybe?

rajatjindal: done

rajatjindal:
> OH, you're right!
> 
> The one question I have is whether we want to add this to be the default client.
> 
> Specifically, we could add an `init` function (or more?) _somewhere_ in the SDK, where we set the default client:
> 
> ```go
> func init() {
> 	http.DefaultClient = NewClient()
> }
> ```
> 
> This will allow the handler to directly call `http.Get` for example, without setting up your own client / transport.
> 
> In any case, this change looks great, and this would be a follow-up addition — not blocking this PR.

yeah, I am not too sure about that. This almost feels magical and sometimes people don't like magic :). what do you think @adamreese

radu-matei:
Well, HTTP operations will not work without it, so it's not as much magic as it is making something work.
But if we decide that would be a good idea, this would be a follow-up, I don't want to hold this PR.

rajatjindal:
ok, that sounds good to me. I'll go ahead and merge this PR

PR #1363, opened 2023-04-06
Don't assume that the error can be dereferenced

rylev:
#1358 assumed that the error type returned from the HTTP handler could be dereferenced which is true for `anyhow::Error` but a bit restrictive. This refactors so that the only assumption is that the error implements `std::error::Error`.

### DIFF
--- a/sdk/rust/macro/src/lib.rs
+++ b/sdk/rust/macro/src/lib.rs
@@ -25,10 +25,10 @@ pub fn http_component(_attr: TokenStream, item: TokenStream) -> TokenStream {
                     Err(error) => {
                         let body = error.to_string();
                         eprintln!("Handler returned an error: {}", body);
-                        let mut error: &(dyn std::error::Error + 'static) = &*error;
-                        while let Some(source) = error.source() {
-                            eprintln!("  caused by: {}", source);
-                            error = source;
+                        let mut source = error.source();
+                        while let Some(s) = source {
+                            eprintln!("  caused by: {}", s);
+                            source = s.source();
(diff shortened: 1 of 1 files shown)


sdk/rust/macro/src/lib.rs
```diff
@@ -25,10 +25,10 @@ pub fn http_component(_attr: TokenStream, item: TokenStream) -> TokenStream {
                     Err(error) => {
                         let body = error.to_string();
                         eprintln!("Handler returned an error: {}", body);
-                        let mut error: &(dyn std::error::Error + 'static) = &*error;
-                        while let Some(source) = error.source() {
-                            eprintln!("  caused by: {}", source);
-                            error = source;
+                        let mut source = error.source();
+                        while let Some(s) = source {
+                            eprintln!("  caused by: {}", s);
```
fibonacci1729, line 30: nit: ```suggestion
                            eprintln!("  caused by: {s}");
```

Comment (fibonacci1729) [inline sdk/rust/macro/src/lib.rs:30]:
```suggestion
                            eprintln!("  caused by: {s}");
```

fibonacci1729 (approved):
Optional nit, LGTM!

sdk/rust/macro/src/lib.rs
rylev: Apparently this is not possible in a macro expansion (which I was not aware of before this).

fibonacci1729: Oh interesting, TIL also!

PR #1365, opened 2023-04-07
feat(kv): add redis backed KV host component

fibonacci1729:
This PR implements a Redis backed KV host component for Spin. Once #1360 is merged and supports `KeyValueStore::Redis` users will be able to seamlessly swap out the the default backend by overriding the `default` store via runtime config, e.g.:

```toml
# spin-config.toml

[key_value_store.default]
type = "redis"
url = "redis://127.0.0.1/"
```

```
$ spin up -f examples/rust-key-value --runtime-config-file spin-config.toml
```


crates/key-value-redis/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+            .map_err(log_error)?;
+
+        Ok(Arc::new(RedisStore {
+            connection: connection.clone(),
+        }))
+    }
+}
+
+struct RedisStore {
+    connection: Arc<RwLock<Connection>>,
```
rylev, line 44: Since we're only using the write lock is there a reason we'd prefer a RwLock over a plain Mutex?

kate-goldenring (approved):
LGTM

crates/key-value-redis/src/lib.rs
fibonacci1729: Ah good catch, thanks!

radu-matei:
Tested this together with the latest commit from #1360 and everything works great!


ref  https://github.com/fermyon/spin/commit/8e8f15e705683a227bc80ee56152de22e2c8caff
(in case anyone else wants to play around with this (the diff below is just a test to get myself familiar with how everything fits together)

PR #1369, opened 2023-04-10
Add default HTTP client compatible with Spin SDK

radu-matei:
This commit adds an `init` function to the SDK's HTTP package that
overrides the default HTTP client to use the internals for sending an
outbound HTTP request compatible with Spin.

This has two implications:

* calls to `http.Get` or `http.Post` (i.e using the standard way of
  making HTTP requests in Go) should now work by default in Spin
  applications
* future updates to WASI and Go networking should mean fewer changes to
  applications that take this approach.

Note that this does not prevent passing a custom HTTP client to
libraries that support this.

### DIFF
--- a/sdk/go/http/http.go
+++ b/sdk/go/http/http.go
@@ -11,6 +11,11 @@ import (
 	"github.com/julienschmidt/httprouter"
 )
 
+// Override the default HTTP client to be compatible with the Spin SDK.
+func init() {
+	http.DefaultClient = NewClient()
+}
+
 // Router is a http.Handler which can be used to dispatch requests to different
 // handler functions via configurable routes
 type Router = httprouter.Router


michelleN (approved):
Paired with @radu-matei to manually test this. We did a simple http get to ensure that using the sdk still works. The real exciting part of this change is that you can now use import and use libraries that make standard http requests under the hood so yay 🎉 ! We discussed testing and decided to hold off since this is more in an experimental state for using with deps. We also discussed how this impacts non http spin apps that use libraries that make http calls.  @radu-matei has some ideas around that (potentially adding the same init func to the redis package with an underscore import `_ github.com/fermyon/spin/sdk/go/http` and will submit a follow up PR to address that.

PR #1375, opened 2023-04-13
Add key-value redis runtime config

fibonacci1729:


radu-matei (approved):
🎉

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -95,14 +108,20 @@ impl TriggerHooks for KeyValuePersistenceMessageHook {
...
-        // Only print if the 'default' store has type 'spin'.
-        if let KeyValueStoreOpts::Spin(store_opts) = runtime_config.default_key_value_opts() {
-            if let Some(path) = &store_opts.path {
-                println!("Storing default key-value data to {path:?}");
-            } else {
-                println!("Using in-memory default key-value store; data will not be saved!");
+
+        match runtime_config.default_key_value_opts() {
+            KeyValueStoreOpts::Redis(RedisKeyValueStoreOpts { url }) => {
+                println!("Storing default key-value data to Redis at {url:?}");
```
michelleN, line 114: Will the url here contain any passwords/secrets?

lann: I don't think we necessarily need to print here at all. I wanted to for the defaults because it may not be obvious what's happening but for this you have to explicitly configure it anyway.

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -81,6 +78,22 @@ impl SpinKeyValueStoreOpts {
     }
 }
 
+#[derive(Clone, Debug, Deserialize)]
+pub struct RedisKeyValueStoreOpts {
+    pub url: String,
+}
+
+impl RedisKeyValueStoreOpts {
+    fn build_store(&self, _config_opts: &RuntimeConfigOpts) -> Result<KeyValueStore> {
```
lann, line 87: You can just remove the arg if you want. Its not a trait

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -81,6 +78,22 @@ impl SpinKeyValueStoreOpts {
...
+        _ = spin_key_value_redis::parse_redis_url(&self.url)
+            .ok_or_else(|| anyhow!("Invalid Redis URL in runtime config"))?;
```
lann, line 89: Can/should this happen in `KeyValueRedis::new`?

crates/key-value-redis/src/lib.rs
```diff
@@ -1,21 +1,24 @@
...
-    database_url: String,
+    database_url: Url,
     connection: OnceCell<Arc<Mutex<Connection>>>,
 }
 
 impl KeyValueRedis {
-    pub fn new(database_url: String) -> Self {
-        Self {
+    pub fn new(address: String) -> Result<Self> {
+        let database_url = parse_redis_url(&address).ok_or_else(|| anyhow!("Invalid Redis URL"))?;
```
lann, line 16: `anyhow::Context` is impled for `Option`, so you can:
```suggestion
        let database_url = parse_redis_url(&address).context("Invalid Redis URL")?;
```

lann (approved):

PR #1392, opened 2023-04-16
tests(sdk/go): kill spin subprocesses after test runs

adamreese:
Spin creates a new process for executing the trigger.  This change uses the group process id to kill spin, it's subprocesses, and repels zombies.


itowlson (approved):
#badunixsyscallattitude

PR #1394, opened 2023-04-16
[sdk]: Custom Display implementation for Router

ThorstenHans:
This PR contains a custom implementation for the `Display` trait. This allows users of the HTTP router to print all registered routes at once.

### Usage
```rust
let mut router = Router::default();
router.get("/:x", echo_param);
router.post("/", echo_param);

print!("{}", router);
```

This would print the following to stdout:

```
Registered routes:
- GET: /:x
- POST: /
```

### DIFF
--- a/sdk/rust/src/http/router.rs
+++ b/sdk/rust/src/http/router.rs
@@ -327,7 +327,7 @@ mod tests {
     fn test_wildcard_last_segment() {
         let mut router = Router::default();
         router.get("/:x/*", echo_param);
-        
+
         let req = make_request(http_types::Method::GET, "/foo/bar");
         let res = router.handle(req).unwrap();
         assert_eq!(res.into_body().unwrap(), "foo".to_string());
@@ -337,13 +337,13 @@ mod tests {
     fn test_router_display() {
         let mut router = Router::default();
         router.get("/:x", echo_param);
-        
+
         let expected = "Registered routes:\n- GET: /:x\n";
         let actual = format!("{}", router);
-        
+
         assert_eq!(actual.as_str(), expected);
     }
-    
+
     #[test]
     fn test_ambiguous_wildcard_vs_star() {
         fn h1(_req: Request, _params: Params) -> Result<Response> {
--- a/sdk/rust/src/http/router.rs
+++ b/sdk/rust/src/http/router.rs
@@ -21,11 +21,11 @@ impl Default for Router {
 
 impl Display for Router {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_str("Registered routes:\n")?;
+        writeln!(f, "Registered routes:")?;
         for (method, router) in &self.methods_map {
-            router.iter().for_each(|route| {
-                f.write_fmt(format_args!("- {}: {}\n", method, route.0)).unwrap();
-            });
+            for route in router.iter() {
+                writeln!(f, "- {}: {}", method, route.0)?;
+            }
         }
         Ok(())
     }
@@ -327,7 +327,7 @@ mod tests {
     fn test_wildcard_last_segment() {
         let mut router = Router::default();
         router.get("/:x/*", echo_param);
-
+        
         let req = make_request(http_types::Method::GET, "/foo/bar");
         let res = router.handle(req).unwrap();
         assert_eq!(res.into_body().unwrap(), "foo".to_string());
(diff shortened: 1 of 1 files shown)


radu-matei (approved):
This change is great, thanks!


LGTM (once you run `cargo fmt`)

sdk/rust/src/http/router.rs
```diff
@@ -19,6 +19,18 @@ impl Default for Router {
     }
 }
 
+impl Display for Router {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("Registered routes:\n")?;
+        for (method, router) in &self.methods_map {
+            router.iter().for_each(|route| {
+                f.write_fmt(format_args!("- {}: {}\n", method, route.0)).unwrap();
```
fibonacci1729, line 27: To avoid the `unwrap`, could we use `try_for_each` in the loop?

rylev, line 27: Could we use the `write!` macro here?

rylev: Or instead of using an iterator combinator, you can just use another for loop and use the `?` operator.

sdk/rust/src/http/router.rs
```diff
@@ -321,6 +333,15 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_router_display() {
+        let mut router = Router::default();
+        router.get("/:x", echo_param);
+        
+        let expected = "Registered routes:\n- GET: /:x\n";
+        let actual = format!("{}", router);
+        assert_eq!(actual, expected.to_string());
```
rylev, line 343: Nit:
```suggestion
        let expected = "Registered routes:\n- GET: /:x\n";
        let actual = format!("{}", router);
        assert_eq!(actual.as_str(), expected);
```

rylev (commented):
I just have some small code quality nits. Looks great overall!

sdk/rust/src/http/router.rs
fibonacci1729: Sorry meant to hit reply! 

Similarly,  there is a `writeln!` if you prefer not having the `\n` in the format string.

ThorstenHans: Thanks for the feedback. Just having two for loops is way easier to read. Makes definitely sense. Thanks for the feedback guys!

ThorstenHans: so obvious 🤦🏼‍♂️

sdk/rust/src/http/router.rs
```diff
@@ -315,12 +327,23 @@ mod tests {
     fn test_wildcard_last_segment() {
         let mut router = Router::default();
         router.get("/:x/*", echo_param);
-
+
```
rylev, line 330: I believe you'll need to run `cargo fmt` again.

sdk/rust/src/http/router.rs
ThorstenHans: @rylev you comment got my attention 😄 

I guess your mentioning this because (IMO) `expected.to_string()` is more expensive. I assume `.to_string()` doing additional heap allocation (due to calling `String::from("")` internally)

The implementation of `as_str()` on `String` simply returns `self`, which I interpret as a borrowed `&str` with "potentially" shorter lifetime.  

So the change would prevent an additional heap allocation. Is my interpretation here correct, or did I get something wrong?

rylev: Everything you said is more or less correct. However I made the suggestion just because it's shorter 😀, and I find it easier to read. After all, one heap allocation in a single test won't really make any difference at all in the performance of the test suite.

ThorstenHans: Thanks. Just wanted to understand it. shorter is better! agree 💯 on that 😄 Was just curious.

sdk/rust/src/http/router.rs
```diff
@@ -19,6 +19,18 @@ impl Default for Router {
     }
 }
 
+impl Display for Router {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "Registered routes:")?;
+        for (method, router) in &self.methods_map {
+            for route in router.iter() {
```
fibonacci1729, line 26: One minor preference (feel free to ignore):

```rust
for (path, _handler) in router.iter() {
     writeln!(f, "- {method}: {path}")?;
}
```

fibonacci1729 (approved):

PR #1395, opened 2023-04-16
Improve error when the component wasm does not exist

rylev:
We should probably be validating the component source path when we parse the spin.toml file, but this at least makes sure that the error from spin trigger is more understandable when the wasm component does not exist.

**Before**

```
Error: Failed to instantiate component 'todo'

Caused by:
    0: loader error: No such file or directory (os error 2)
    1: No such file or directory (os error 2)
```

**After**

```
Error: Failed to instantiate component 'todo'

Caused by:
    0: failed to read component source from disk at path '/Users/rylev/Code/fermyon/todo/target/wam32-wasi/release/todo.wasm'
    1: No such file or directory (os error 2)
```

### DIFF
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -366,8 +366,8 @@ pub enum Error {
     #[error("host component error: {0:#}")]
     HostComponentError(#[source] anyhow::Error),
     /// An error from a [`Loader`] implementation.
-    #[error("loader error: {0:#}")]
-    LoaderError(#[source] anyhow::Error),
+    #[error(transparent)]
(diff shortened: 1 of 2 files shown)


itowlson (approved):
Ugh, we have fixed this SO MANY TIMES, and it feels like it keeps going backwards.  Thank you for the catch and the fix.

Longer term, we clearly need tests around this and other missing-file scenarios, otherwise this will go back again in six months' time.  Not a blocker for this PR though.

fibonacci1729 (approved):
LGTM!

PR #1378, opened 2023-04-13
Improve error message on manifest schema mismatch

itowlson:
Fixes #1322.

But I'm really unhappy with it.  Putting it out as a draft in the hope that someone has an insight on how to simplify it (and make it more test-friendly!).

PR #1396, opened 2023-04-16
Simplify version handling in manifest

rylev:
I might be missing something, but I think this way of handling manifest parsing is a bit more robust. The code is a bit shorter, and the error messages are nicer.

**Before**

```
Caused by:
    data did not match any variant of untagged enum RawAppManifestAnyVersionImpl
```

**After**

```
Caused by:
    unknown field `FOOescription` at line 14 column 1
```


itowlson:
Cf. #1378, which attempts to solve the same problem while keeping the door open to future versions, but is fugly.

rylev:
I think if we want graceful handling of multiple versions we'll need to write our own custom deserializer. It shouldn't be too bad, but I think we can leave that until we have a need for a second version.

itowlson (approved):

PR #1398, opened 2023-04-16
feat(sdk/go): add constants for Spin headers

adamreese:
This commit adds constants for header names added by Spin.


itowlson (approved):

michelleN (approved):
Will be so nice to have these handy! We should add some examples of how folks could use these at some point.

Issue #1337, opened 2023-03-30
Compile warning from `watchexec` dependency


itowlson:
Building with Cargo 1.68 reports `warning: the following packages contain code that will be rejected by a future version of Rust: nom v5.1.2` (the problematic feature is uninteresting but you can get it via `cargo report` if you are keen).

nom 5.1.2 appears to come in as an indirect dependency of `watchexec` via `clearscreen` and `terminfo`.  nom 7.1.3 is also imported, by various other crates (but doesn't seem to cause the warning).

calebschoepp:
Some googling indicates that this is fixed in versions >5 of `nom`. I'll file an issue with `watchexec` to try and upgrade their dependency on nom.

itowlson:
Fixed via #1401.

PR #1401, opened 2023-04-17
Fix warning when building spin

calebschoepp:
I've been tracking the upstream fixes for #1337 and it is now fixed. `watchexec` hasn't cut a release yet so I've pinned us to the latest commit. If we're not comfortable with that then we can wait for the to cut a release (not clear how long that will take).


itowlson (approved):
Nice one, thanks Caleb!

PR #1405, opened 2023-04-17
ref(sdk/go): remove unused custom error type

adamreese:
- Remove unused `Error` type
- Remove exported error consts.  These are set for the internal error code from spin.  The error codes aren't exposed to the user.
- Fix shadowing of the `string` type

The types removed are exported and part of the api but are no use. I think it's safe to remove them. _last words_


sdk/go/key_value/key-value.go
```diff
@@ -111,24 +104,29 @@ func fromCStrList(list *C.key_value_list_string_t) []string {
...
+	case errorKindStoreTableFull:
+		return errors.New("store table full")
+	case errorKindNoSuchStore:
+		return errors.New("no such store")
+	case errorKindAccessDenied:
+		return errors.New("access denied")
+	case errorKindInvalidStore:
+		return errors.New("invalid store")
+	case errorKindNoSuchKey:
+		return errors.New("no such key")
```
itowlson, line 125: I realise this is not a change in behaviour but it does prompt me to ask how consumers of the Go API are meant to identify the "no such key" case (e.g. in order to default an initial value) - do they have to compare on the message?

adamreese: Great point.  They would need to check the error string.  I'm going to flip the approach and implement `Error` as an actual `error` and see how it feels.

adamreese: Creating the custom error is bloating this PR.  I'll create a new PR when it's completed

itowlson (approved):

itowlson:
@adamreese okay to merge this?

PR #1407, opened 2023-04-18
fix multiple header values not preserved by Rust SDK

### LINKED ISSUE #1406: Multiple header values not preserved by Rust SDK
In the course of investigating other SDKs' behaviour for https://github.com/fermyon/spin-python-sdk/pull/28, @jasonwashburn identified that the Rust SDK did not appear to preserve multiple header values.

Looking at the Spin source code, it seems like Spin faithfully maps headers into the WIT format; but when the Rust SDK converts that into a `http::Request` (https://github.com/fermyon/spin/blob/3d20ffdab31bac660bb0e3b2387f27641e9f6221/sdk/rust/macro/src/lib.rs#L82), it uses `http::header::HeaderMap::insert()` (https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.insert), which is documented as "If the map did have this key present, the new value is associated with the key _and all previous values are removed._" (emphasis added).

It seems like we might want to instead use `HeaderMap::append()` (https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.append), which is documented as "If the map did have this key present, the new value is pushed to the end of the list of values currently associated with the key."

**Compatibility considerations**

This would cause a few changes in observable behaviour in the presence of multiple header values:

* Iterators and `get_all` would provide more values than they did before.
* `get()` would currently get the _last_ value for the key; after this change, it would get the _first_ value for the key.

cc @fibonacci1729

Patrick0308:
close #1406

### DIFF
--- a/sdk/rust/macro/src/lib.rs
+++ b/sdk/rust/macro/src/lib.rs
@@ -81,7 +81,7 @@ pub fn http_component(_attr: TokenStream, item: TokenStream) -> TokenStream {
         ) -> anyhow::Result<()> {
             let headers = http_req.headers_mut().unwrap();
             for (k, v) in &spin_req.headers {
-                headers.insert(
+                headers.append(
                     <http::header::HeaderName as std::str::FromStr>::from_str(k)?,
                     http::header::HeaderValue::from_str(v)?,
                 );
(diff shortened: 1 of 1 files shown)


lann (approved):

fibonacci1729 (approved):

lann:
@Patrick0308 Thanks for the contribution. The Spin project requires all commits to be signed; could you please [sign your commit](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#gpg-commit-signature-verification) and update the PR?

Patrick0308:
@lann @fibonacci1729 PTAL, I had signed the commits.

lann (approved):

lann:
Thanks!

PR #1408, opened 2023-04-18
[SIP] sqlite host component

rylev:
This introduces [a SIP](https://github.com/fermyon/spin/blob/sqlite-host-component/docs/content/sips/000-sqlite.md) for adding a sqlite interface to Spin as well as an initial MVP implementation.

Much like the key-value interface, this would be a "batteries included" implementation where users gain access to a SQL database with no configuration required. 

I have also created [an example TODO application](https://github.com/rylev/spin-todo) that uses this MVP implementation.


docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
...
+// If `database` is "default", the default instance is opened.
+//
+// `error::no-such-database` will be raised if the `name` is not recognized.
+open: func(name: string) -> expected<connection, error>
+
+// TODO: Do we want a separate statment type? This is currently what `wasi-sql` does.
+// allows parameterized queries
+type statement = u32
+drop-statement: func(statement: statement)
+prepare-statement: func(query: string, params: list<data-type>) -> expected<statement, error>
```
itowlson, line 66: My impression of a prepared statement was that it generally didn't include parameter values (or, at least, that you could reuse it with different parameter values).  This interface seems to bake the parameter values into the prepared statement.  Is that intentional?

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
...
+// Close the specified `connection`.
+close: func(conn: connection)
+
+// A database row
+record row {
+  values: list<data-type>,
+}
+
+// Common data types
+variant data-type {
```
itowlson, line 83: nit: the "type" would be e.g. integer or string; these are values

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
+title = "SIP 000 - sqlite"
```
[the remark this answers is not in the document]
rylev: I definitely think we want to be very clear that the semantics of this interface are sqlite. I think we're asking for trouble if we brand this as generic SQL. This is definitely a point that is causing a lot of discussion about the feasibility of `wasi-sql` (as opposed to `wasi-mysql`, `wasi-postgresql`, etc.)

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
...
+// Open a connection to a named database instance.
+//
+// If `database` is "default", the default instance is opened.
+//
+// `error::no-such-database` will be raised if the `name` is not recognized.
+open: func(name: string) -> expected<connection, error>
+
+// TODO: Do we want a separate statment type? This is currently what `wasi-sql` does.
+// allows parameterized queries
+type statement = u32
```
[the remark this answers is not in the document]
rylev: I agree. I've gone ahead and removed prepared statements because from the host side there is no difference so the added complexity for the guest gets us nothing.

docs/content/sips/000-sqlite.md
rylev: Changed this to just `Value` but happy to keep bikeshedding on it.

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
...
+drop-statement: func(statement: statement)
+prepare-statement: func(query: string, params: list<data-type>) -> expected<statement, error>
+
+// Execute a statement
+execute: func(conn: connection, statement: statement) -> expected<unit, error>
+
+// Query data
+query: func(conn: connection, q: statement) -> expected<list<row>, error>
```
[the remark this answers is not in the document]
rylev: Done!

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
...
+* Schema tracking schemes (e.g., a "migrations" table) themselves require some sort of bootstrap step.
+* This goes against the design principle of keeping components handler functions simple and single purpose.
+
+There are several possible ways to address this issue such as:
+* Some mechanism for running spin components before others where the component receives the current schema version and decides whether or not to perform migrations. 
+* The spin component could expose a current schema version as an exported value type so that an exported function would not need to called. If the exported schema version does not match the current schema version, an exported migrate function then gets called.
+* A spin component that gets called just after pre-initialization finishes. Similarly, this component would expose a schema version and have an exported migration function called when the exported schema version does not match the current schema version.
```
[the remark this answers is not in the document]
rylev: Add this and made a note that we may want to select multiple of these options.

rylev:
@itowlson I think this is ready for another look. The implementation pretty much matches the SIP now. That means the open questions still remain, but we have a working prototype. I guess the question now is: should we answer the open questions or merge and *then* figure out the open questions?

(FYI: the CI failure is because `spin-componentize` needs to be updated)

itowlson (commented):
Commented only on the SIP (haven't read through the implementation but can if you want)

crates/sqlite/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+            Ok(rows
+                .into_iter()
+                .map(|r| r.map_err(|e| sqlite::Error::Io(e.to_string())))
+                .collect::<Result<_, sqlite::Error>>()?)
+        }
+        .await)
+    }
+
+    async fn close(&mut self, connection: sqlite::Connection) -> anyhow::Result<()> {
+        Ok(async {
```
dicej, line 140: Can replace the `async { ... }.await` with a normal block here.

dicej: And likewise for the other methods in this `impl` block.

dicej: Also, can we wrap these in `tokio::task::block_in_place` like we do in https://github.com/fermyon/spin/blob/b1ac4d8307948bbcd833fc689be3471f8fa174a0/crates/key-value-sqlite/src/lib.rs#L71-L138?

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
...
+
+### What about `wasi-sql`?
+
+[`wasi-sql`](https://github.com/WebAssembly/wasi-sql) is a work-in-progress spec for a generic SQL interface that aims to support "the features commonly used by 80% of user application". It is likely that when `wasi-sql` is more mature users will be able to successfully use functionality based on the `wasi-sql` interface to interact with a sqlite databases. However, there are still reasons that a dedicated sqlite interface would still be useful:
+
+* For the 20% of use cases where `wasi-sql` is too generic, a dedicated `sqlite` interface can provide that functionality. 
+* The `wasi-sql` spec is under active investigation, and there are large questions about how to best support such a wide breadth of sql flavors. This implementation can help clarify those questions and push upstream work further along.
```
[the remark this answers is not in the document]
dicej: For me the question is: do we want to use the opportunity to help _make_ `wasi-sql` ready?

Specifically, I'm thinking it would be nice to follow the precedent we set with key-value such that you can run the same code locally and in Cloud with only a runtime config change.  And since that's kind of the point of `wasi-sql`, we could implement it and making any changes required as necessary.

I'm wary about adding interfaces which can only practically be implemented locally, so I'd either like to see something more like `wasi-sql` or a plan for providing a "cloud-scale" SQLite service.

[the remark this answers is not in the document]
rylev: To add on top of what Radu says: `wasi-sql` is a long way from stabilization. We can use this work to help inform the evolution of `wasi-sql`, but there are a lot of open questions about the viability of `wasi-sql` in general since sql flavors differ very substantially. Focusing on one sql flavor, sqlite, will make our lives a lot easier and unblock things much more quickly than trying to conform to a future `wasi-sql` standard.

Cargo.toml
```diff
@@ -128,10 +129,10 @@ tracing = { version = "0.1", features = ["log"] }
 wasmtime-wasi = { version = "7.0.0", features = ["tokio"] }
 wasi-common-preview1 = { package = "wasi-common", version = "7.0.0" }
 wasmtime = { version = "7.0.0", features = ["component-model"] }
-spin-componentize = { git = "https://github.com/fermyon/spin-componentize" }
-wasi-host = { package = "host", git = "https://github.com/fermyon/spin-componentize" }
-wasi-common = { git = "https://github.com/fermyon/spin-componentize" }
-wasi-cap-std-sync = { git = "https://github.com/fermyon/spin-componentize" }
+spin-componentize = { path = "../spin-componentize" }
```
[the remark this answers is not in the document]
rylev: Yes, though now that [I have this up on GitHub](https://github.com/rylev/preview1.wasm/tree/sqlite) I can at least refer to the a git dependency and not on a local dependency. But before we merge this we'll want to merge the support into spin-componentize.

crates/sqlite/src/lib.rs
rylev: Yes though we might move to a different sqlite library which has an async interface.

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,139 @@
...
+
+Note that the syntax of the following `WIT` file matches the `wit-bindgen` version currently used by Spin, which is out-of-date with respect to the latest [`WIT` specification](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md) and implementation. Once we're able to update `wit-bindgen`, we'll update the syntax of all the Spin `WIT` files, including this one.
+
+```fsharp
+// A handle to an open sqlite instance
+type connection = u32
+
+// The set of errors which may be raised by functions in this interface
+variant error {
+  // The host does not recognize the database name requested.
```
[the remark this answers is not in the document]
rylev: Well it's not that the name is defined, it's that there is no database by that name. This differs from `access-denied` in that in the `access-denied` case there is explicitly no access granted for a database with that name while in the `no-such-database` case it's not a question of permission, there simply is no database that exists for that identifier. Any suggestions on how to make that clearer?

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,146 @@
...
+
+// Query data
+query: func(conn: connection, query: string, parameters: list<value>) -> expected<list<row>, error>
+
+// Close the specified `connection`.
+close: func(conn: connection)
+
+// A database row
+record row {
+  values: list<column-value>,
```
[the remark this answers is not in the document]
rylev: Changed the wit definitions to not encode the column name for every row. This would be needlessly expensive for large result sets.

docs/content/sips/000-sqlite.md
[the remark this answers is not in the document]
itowlson: "No database named {...} is defined." or "There is no database named {...}." maybe?  (Of these, I somewhat prefer the first, but I am sure there are better options.)

Your comment also made me wonder about which of "not defined" and "not granted access" takes precedence.  If I write a component that tries to use a database called `no-ivans`, but the manifest does not grant my component access to `no-ivans`, will my component get `access-denied` rather than "leaking" information about whether the database exists or doesn't exist?

(Longer term this should be a non-issue because `spin up` should fail manifest validation if it refers to a name that isn't defined, but I don't think we can do that today.)

docs/content/sips/000-sqlite.md
```diff
@@ -0,0 +1,164 @@
...
+
+// The set of errors which may be raised by functions in this interface
+variant error {
+  // A database with the supplied name does not exist
+  no-such-database,
+  // The requesting component does not have access to the specified database (which may or may not exist).
+  access-denied,
+  // The provided connection is not valid
+  invalid-connection,
+  // The database no longer has reached its capacity
```
vdice, line 55: Just checking on the wording here -- Should be 'The database has reached its capacity'?

vdice (commented):
Quick proofreading review for the SIP

crates/sqlite/src/lib.rs
rylev: So the reason I wrapped these in `async` blocks is the same reason it's done in `key-value` crate. We want the `?` to propagate the inner error of the method's return type. `async` blocks have different scoping rules than regular blocks and allow for a "try block" like control flow. I agree it's a bit ugly, but the alternatives are also not that great...

rylev: We could do something like this: https://github.com/fermyon/spin/compare/sqlite-host-component...block-in-place

rylev:
I am marking this PR as ready to review. While this feature is not complete, I do believe it is useful enough for folks to start experimenting with. Below I've tried to capture the current state and the open questions we need to answer before we can officially make this a stable feature:

This PR provides the following features:
* Support in spin for a sqlite host implementation that provides a default sqlite database when running spin locally.
* Support for runtime configuration for adding additional local sqlite databases.
* Rust SDK support for targeting these databases from a Rust based spin component
* The ability to run arbitrary sql statements against the default database on start up with `spin up --sqlite-migration="YOUR SQL HERE"`

* **I do not believe this feature is ready to marked stable**. As such I have put sqlite support behind a feature flag in the only SDK that currently supports it: Rust. This will still allow folks to target sqlite if they create spin components manually (i.e., using wit-bindgen directly), but I believe we already consider such usage experimental and not officially supported by our backwards compatibility guarantees. If we believe that by merging this PR that we are committing to backwards compatibility of this feature in its current form, then I think we should wait and consider how we want to allow for experimentation before committing to backwards compatibility.
* There are still [some open questions](https://github.com/fermyon/spin/blob/7d97005de534e8516894925dc6c8c177169eeaaf/docs/content/sips/013-sqlite.md?plain=1#L128) in the SIP with regards to sqlite versioning and capacity limits. I don't believe those need to be solved before experimentation can begin.
* There is no support for this feature in Fermyon Cloud. Since this is an experimental feature only, I believe that this is fine for now - though we will quickly want to change that.

crates/loader/src/bindle/config.rs
```diff
@@ -49,6 +49,8 @@ pub struct RawWasmConfig {
     pub allowed_http_hosts: Option<Vec<String>>,
     /// Optional list of key-value stores the component is allowed to use.
     pub key_value_stores: Option<Vec<String>>,
+    /// Optional list of sqlite databases the component is allowed to use.
```
itowlson, line 52: I don't know if this counts as human readable text, but if it does, then SUPER DUPER NIT we should follow the project by using "SQLite" rather than "sqlite."

docs/content/sips/013-sqlite.md
```diff
@@ -0,0 +1,165 @@
...
+
+#### Interface open questions
+
+* `row-result` can be very large. Should we provide some paging mechanism or a different API that allows for reading subsets of the returned data?
+  * Crossing the wit boundary could potentially be expensive if the results are large enough. Giving the user control of how they read that data could be helpful.
+* Is there really a need for query *and* execute functions since at the end of the day, they are basically equivalent?
+
+#### Database migrations
+
+Database tables typically require some sort of configuration in the form of database migrations to get table schemas into the correct state. To begin with a command line option supplied to `spin up` will be available for running SQL statements on start up (i.e., `--sqlite-migration "CREATE TABLE users..."`). It will be up to the user to provide itempotent statements such that running them multiple times does not produce unexpected results.
```
itowlson, line 110: Couple of ill-formed thoughts on this:

* As a point of detail, I suggest `--sqlite` rather than specifically talking about migrations, unless the flag is specifically bound to some sort of migrations format or infrastructure.  It's a pretty general purpose command facility.
* Because SQL statements can be quite long and fiddly, it would be good for users not to have to type them interactively on the CLI.  So it might be worth, even on day one, having a syntax for running a statement (or statements?) from a file, e.g. `--sqlite @usertable.sql`.
* Longer term, could consider a way to connect the user's favourite migration to hook up to the private database so it could generate the required scripts (or execute them directly?).  Not sure what this would look like though.
* `--key-value` currently operates only on the default store.  Are we okay with `--sqlite` operating only on the default database?  I guess if they have non-default ones then those are BYO and they can easily use their favourite SQLite tools on them.

docs/content/sips/013-sqlite.md
```diff
@@ -0,0 +1,165 @@
...
+
+However, such an approach (while useful) is likely to not be sufficient for more advanced use cases. We could require the user to ensure that the database is in the correct state each time the trigger handler function is run. However, there are a few issues with this:
+* Schema tracking schemes (e.g., a "migrations" table) themselves require some sort of bootstrap step.
+* This goes against the design principle of keeping components handler functions simple and single purpose.
```
itowlson, line 114: These two lines seem to be talking about something other than the CLI option... am I misunderstanding?

docs/content/sips/013-sqlite.md
```diff
@@ -0,0 +1,165 @@
...
+It should be noted that many of these options are not mutually exclusive and we could introduce more than one (perhaps starting with one option that will mostly be replaced later with a more generalized approach).
+
+For now, we punt on this question and only provide a mechanism for running SQL statements on start up.
+
+#### Implementation requirements
+
+**TODO**: Open questions:
+* Assumed sqlite version?
+  * Semantics may change slightly depending on the sqlite version. It's unlikely that we'll be able to match the exact versions between whatever sqlite implementation spin users, Fermyon Cloud, and the user (if they decide to create their own databases manually). Having some guidance on which versions are expected to work might make it easier to guide the user down the right path.
+* Capacity limits? The following are different capacities we might want to control:
```
itowlson, line 131: This feels like something we probably don't need to be concerned about in local Spin.  Fermyon Cloud or other hosts may want to place limits, but I don't think we need to plan for that here (except maybe assigning an error code for them).

The main thing I can think of locally is if there could be problems with the size of a query result set, which brings us back round to the list vs cursor vs hybrid issue all over again...

docs/content/sips/013-sqlite.md
itowlson, line 110: tiny typo 
```suggestion
Database tables typically require some sort of configuration in the form of database migrations to get table schemas into the correct state. To begin with a command line option supplied to `spin up` will be available for running SQL statements on start up (i.e., `--sqlite-migration "CREATE TABLE users..."`). It will be up to the user to provide idempotent statements such that running them multiple times does not produce unexpected results.
```

sdk/rust/src/lib.rs
```diff
@@ -8,6 +8,10 @@ pub mod outbound_http;
 /// Key/Value storage.
 pub mod key_value;
 
+/// Sqlite
+#[cfg(feature = "experimental-sqlite")]
```
itowlson, line 12: Is this for your testing, or do you intent that SDK users would need to turn the feature on in Cargo.toml until it stabilises?

itowlson (approved):
Some comments / questions / nits but nothing that can't be iterated on as we build it out.  Nice one - thank you!

sdk/rust/src/lib.rs
itowlson: Oh sorry I see discussion of this in your ready-for-review comment.

itowlson:
By the way, what is the order of landing things for the first iteration?  I saw the PR took a dependency on a branch of `spin-componentize`.  Is the idea that we would land `spin-componentize` first?  Is that going to be the general workflow for as long as we need to continue to support modules?  (cc @dicej)

sdk/rust/src/lib.rs
itowlson: I had a look at the docs (https://developer.fermyon.com/spin/api-guides-overview) and we currently have several APIs that are described as "experimental" but to which don't require a specific opt-in by the developer.  (I don't present this as advocacy for either approach, just as s data point.)

sdk/rust/src/sqlite.rs
```diff
@@ -0,0 +1,112 @@
...
+/// A single column's result from a database query
+pub type DataTypeResult = sqlite::ValueResult;
+
+/// Represents a store in which key value tuples may be placed
+#[derive(Debug)]
+pub struct Connection(RawConnection);
+
+impl Connection {
+    /// Open a connection
+    pub fn open(database: &str) -> Result<Self, Error> {
```
itowlson, line 19: Could we add an `open_default()` please, to save folks having to use the magic string everywhere, and for consistency with the Rust KV SDK?  Thanks!

sdk/rust/src/sqlite.rs
```diff
@@ -0,0 +1,112 @@
+wit_bindgen_rust::import!("../../wit/ephemeral/sqlite.wit");
+
+use sqlite::Connection as RawConnection;
+
+/// Errors which may be raised by the methods of `Store`
+pub type Error = sqlite::Error;
+
+/// A parameter used when executing a sqlite statement
+pub type DataTypeParam<'a> = sqlite::ValueParam<'a>;
```
itowlson, line 9: suggestion: `ParameterValue` would be consistent with our other Rust RDBMS SDKs, and is clear that it's a _value_ rather than a data type.  And similar comment for `DataTypeResult` - the other SDKs use `DbValue` for this.

docs/content/sips/013-sqlite.md
rylev: Yes, this is about an alternative approach that we will *not* be doing. I'll try to make the wording more clear.

sdk/rust/src/lib.rs
rylev: The current way of describing things as experimental but not putting any opt-in in front of the user is pretty dangerous, and I think we should move away from doing so. I plan on writing something up on how we can do experimental features in the future.

docs/content/sips/013-sqlite.md
rylev: Perhaps we should give some sort of *minimum* requirement that we expect every host environment running spin applications to conform to. The [key-value SIP does this](https://github.com/fermyon/spin/blob/608b5f4b11914109e84f9c950cba6db5b50ce512/docs/content/sips/010-key-value.md?plain=1#L123-L125) for that feature.

docs/content/sips/013-sqlite.md
rylev: 1.) I've renamed the option to just `sqlite` 👍
2.) I think this would be a good idea. Is `@` the right syntax for indicating a file?
3 and 4.) I think 3 is related to number 4. If we give the user a way to run sqlite statements (perhaps through a file) on any of their databases, we won't need a different mechanism for the non-default databases.

radu-matei: Big +1 on something like `--sqlite @usertable.sql`.

crates/trigger/src/runtime_config.rs
```diff
@@ -92,6 +96,29 @@ impl RuntimeConfig {
...
+        for opts in self.opts_layers() {
+            for (name, database) in &opts.sqlite_databases {
+                if !databases.contains_key(name) {
+                    let store = database.build(name, opts)?;
+                    databases.insert(name.to_owned(), store);
+                }
+            }
+        }
+        // Upsert default store
+        if !databases.contains_key("default") {
```
dicej, line 114: (Optional): could use `HashMap::entry` and https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with here and above instead of the `contains_key`/`insert` combo.

crates/sqlite/src/lib.rs
dicej: I definitely like the `block-in-place` option until we're ready to move to an async sqlite library (which will likely do something similar under the hood).

dicej (approved):
Looks great overall.

The `async { $BODY }.await` stuff is aggravating my OCD, mainly just because `$BODY` does synchronous I/O and no async, so I'm definitely in favor of using `block_in_place` instead.  No need to split each method, either -- just make it `task::block_in_place(|| { $BODY })`.

docs/content/sips/013-sqlite.md
itowlson: I like that plan!

radu-matei (approved):
LGTM, and the todo example works great for me!

This is really exciting, thanks for working on it, @rylev!

crates/trigger/src/runtime_config.rs
rylev: Entry requires an owned key every time so we would do an extra allocation. I find in this particular case the non-entry based code is easier to read too. I think I'd prefer to keep it as it is, but let me know if you disagree!

docs/content/sips/013-sqlite.md
rylev: Done!

docs/content/sips/013-sqlite.md
rylev: Done!! You can now run sql from files:

```
spin up --sqlite="@migration.sql"
```

rylev:
I think this is ready to merge. We'll wait until the next version of spin is released so that we have a full release cycle to let this bake.

docs/content/sips/013-sqlite.md
```diff
@@ -0,0 +1,173 @@
...
+  null
+}
+```
+
+*Note: the pseudo-resource design was inspired by the interface of similar functions in [WASI preview 2](https://github.com/bytecodealliance/preview2-prototyping/blob/d56b8977a2b700432d1f7f84656d542f1d8854b0/wit/wasi.wit#L772-L794).*
+
+#### Interface open questions
+
+**TODO**: answer these questions
+* `row-result` can be very large. Should we provide some paging mechanism or a different API that allows for reading subsets of the returned data?
```
lann, line 105: We could hedge on this by changing the function name to e.g. `query-all`, reserving `query` for a future cursor API.

PR #1412, opened 2023-04-19
docs(sdk/go): Add top-level package doc string for godocs

adamreese:


sdk/go/redis/redis.go
```diff
@@ -1,3 +1,5 @@
+// Package redis provides the handler function for the redis trigger, as well
```
itowlson, line 1:
```suggestion
// Package redis provides the handler function for the Redis trigger, as well
```

sdk/go/redis/redis.go
```diff
@@ -1,3 +1,5 @@
...
+// as access to redis within Spin components.
```
itowlson, line 2:
```suggestion
// as access to Redis within Spin components.
```

itowlson (approved):
Gotta get my brand values on but not blockers!

PR #1418, opened 2023-04-21
Large refactor of `spin watch` to only run what is necessary when files change

### LINKED ISSUE #1403: `spin watch` loops infinitely if the `component.build` command modifies anything in `component.files`
By default `spin watch` watches for changes to `component.files` ([docs](https://developer.fermyon.com/common/cli-reference#watch)). If the `component.build.command` modifies these files then `spin watch` falls into an infinite loop. It is very typical for JS applications to have a build command that modifies the static files.

I see two potential solutions here:

1. Stop watching `component.files` when you run `spin watch` and only watch it when you run `spin watch --skip-build`. Passing the `--direct-mounts` flag can mitigate a lot of the downside here.
2. Differentiate between changes to files vs source code and dynamically change the underlying command that `spin watch` is running. When source code changes run `spin build --up`. When files change run `spin up`. (If we can get this to work this is my preferred approach.)

CC @itowlson I'd love to hear your thoughts here.

calebschoepp:
## Changes
- Large refactoring of code to clean things up and make it more clear
- `spin watch` now dynamically generates the command it will run. In practice what this means is that `spin build` is only run when actually necessary and we no longer get infinite loops.
- Closes #1403 .

## Notes
- The mental model you should have is that:
    - Watchexec produces `events` when things happen
    - Via debouncing these `events` are grouped into `actions`
    - These `events` from an `action` are reduced into a single `effect`
    - Based on the current `state` an `effect` produces an `outcome` and potentially changes the `state`
    - The `outcome` is passed back through the `action` to affect the world
- I'm planning to dump this into a README when I migrate this stuff to a crate

## Future work
- I feel like this command is really growing in scope -- as such I'm going to move it to its own crate in a future PR.
- I'm going to get live reconfiguration from the manifest working in a future PR.

### DIFF
--- a/src/commands/watch.rs
+++ b/src/commands/watch.rs
@@ -1,16 +1,22 @@
-use std::{convert::Infallible, path::PathBuf, sync::Arc, time::Duration};
+use std::{
+    convert::Infallible,
+    path::{Path, PathBuf},
+    sync::Arc,
+    time::Duration,
+};
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use clap::Parser;
 use spin_loader::local::{
-    config::{RawFileMount, RawModuleSource},
+    config::{RawComponentManifestImpl, RawFileMount, RawModuleSource},
     parent_dir,
 };
+use spin_manifest::TriggerConfig;
 use watchexec::{
-    action::{Action, Outcome},
+    action::{Action, PreSpawn},
     config::{InitConfig, RuntimeConfig},
     error::RuntimeError,
-    event::{Event, Priority},
+    event::{Event, Priority, ProcessEnd, Tag},
     handler::SyncFnHandler,
     signal::source::MainSignal::Interrupt,
     ErrorHook, Watchexec,
@@ -21,7 +27,8 @@ use crate::{
         APP_MANIFEST_FILE_OPT, DEFAULT_MANIFEST_FILE, WATCH_CLEAR_OPT, WATCH_DEBOUNCE_OPT,
         WATCH_SKIP_BUILD_OPT,
     },
-    watch_filter::WatchFilter,
+    watch_filter::{Filter, WatchPattern},
+    watch_state::{Effect, Effects, State, WatchState},
 };
 
 /// Build and run the Spin application, rebuilding and restarting it when files change.
@@ -87,54 +94,80 @@ impl WatchCommand {
         ));
 
         // Prepare RuntimeConfig for Watchexec
-        let mut runtime_config = RuntimeConfig::default();
-        let (spin_cmd, spin_args) = self.generate_command();
-        runtime_config.commands(vec![watchexec::command::Command::Exec {
-            prog: spin_cmd,
-            args: spin_args,
-        }]);
         let app_dir = parent_dir(&self.app)?;
-        runtime_config.pathset([app_dir.clone()]);
-        let filter = WatchFilter::new(
-            app_dir.clone(),
-            self.generate_path_patterns().await?,
-            WatchFilter::default_ignore_patterns(),
-        )?;
+        let filter = Arc::new(Filter::new(self.generate_filter_config().await?)?);
+        let watch_state = WatchState::new(self.skip_build, self.clear);
+        let watch_state_clone = watch_state.clone();
+        let mut runtime_config = RuntimeConfig::default();
+        runtime_config.pathset([app_dir]);
         runtime_config.command_grouped(true);
-        runtime_config.filterer(Arc::new(filter));
+        runtime_config.filterer(filter.clone());
         runtime_config.action_throttle(Duration::from_millis(self.debounce));
-        runtime_config.on_action(move |action: Action| {
-            let app_manifest_path = self.app.clone();
+        runtime_config.commands(vec![watchexec::command::Command::Exec {
+            prog: self.generate_command(),
+            args: vec![],
+        }]);
+        runtime_config.on_pre_spawn(move |prespawn: PreSpawn| {
+            let up_args = self.up_args.clone();
+            let manifest_path = self.app.clone().to_str().unwrap().to_owned();
+            let watch_state = watch_state.clone();
             async move {
+                // Dynamically modify the command we're running based on the watch state
+                let state = watch_state.get_state();
+                let spin_args = WatchCommand::generate_arguments(state, up_args, manifest_path);
+                let mut cmd = prespawn.command().await.unwrap();
+                cmd.args(spin_args);
+                tracing::debug!("modifying command to: {cmd:?}");
+                Ok::<(), Infallible>(())
+            }
+        });
+        runtime_config.on_action(move |action: Action| {
+            tracing::debug!("handling action: {action:?}");
+            let filter = filter.clone();
+            let watch_state = watch_state_clone.clone();
+            async move {
+                // Map all the events of this action to an effect
+                let mut effects = Effects::new();
                 for event in action.events.iter() {
-                    // Exit if interrupt signal sent
                     if event.signals().any(|s| s.eq(&Interrupt)) {
-                        action.outcome(Outcome::both(Outcome::Stop, Outcome::Exit));
-                        return Ok::<(), Infallible>(());
+                        effects.add(Effect::Exit);
                     }
-
-                    if event.paths().any(|(p, _)| p.ends_with(app_manifest_path.clone())) {
+                    if event
+                        .tags
+                        .iter()
+                        .any(|t| matches!(t, Tag::ProcessCompletion(Some(ProcessEnd::Success))))
+                    {
+                        effects.add(Effect::ChildProcessCompleted);
+                    }
+                    if event.tags.iter().any(|t| {
+                        matches!(
+                            t,
+                            Tag::ProcessCompletion(None)
+                                | Tag::ProcessCompletion(Some(
+                                    ProcessEnd::Exception(_)
+                                        | ProcessEnd::ExitError(_)
+                                        | ProcessEnd::ExitStop(_)
+                                ))
+                        )
+                    }) {
+                        effects.add(Effect::ChildProcessFailed);
+                    }
+                    if filter.matches_manifest_pattern(event) {
                         // TODO: Reconfigure the watcher
                         eprintln!("Application manifest has changed. If this included changes to the watch configuration, please restart Spin.");
+                        effects.add(Effect::ManifestChange);
+                    }
+                    if filter.matches_source_pattern(event) {
+                        effects.add(Effect::SourceChange);
+                    }
+                    if filter.matches_artifact_pattern(event) {
+                        effects.add(Effect::ArtifactChange);
                     }
                 }
-
-                action.outcome(Outcome::if_running(
-                    Outcome::both(
-                        Outcome::both(
-                            Outcome::Stop,
-                            match self.clear {
-                                true => Outcome::Clear,
-                                false => Outcome::DoNothing,
-                            },
-                        ),
-                        Outcome::Start,
-                    ),
-                    Outcome::Start,
-                ));
-
-            Ok::<(), Infallible>(())
-        }});
+                action.outcome(watch_state.handle(effects.reduce()));
+                Ok::<(), Infallible>(())
+            }
+        });
 
         // Start watching
         let runtime = Watchexec::new(init_config, runtime_config.clone())?;
@@ -145,91 +178,77 @@ impl WatchCommand {
         Ok(())
     }
 
-    fn generate_command(&self) -> (String, Vec<String>) {
+    fn generate_command(&self) -> String {
         // The docs for `current_exe` warn that this may be insecure because it could be executed
         // via hard-link. I think it should be fine as long as we aren't `setuid`ing this binary.
-        let spin_cmd = String::from(
+        String::from(
             std::env::current_exe()
                 .unwrap()
                 .to_str()
                 .expect("to find exe path"),
-        );
-        let mut spin_args = match self.skip_build {
-            false => vec![String::from("build"), String::from("--up")],
-            true => vec![String::from("up")],
-        };
-        spin_args.append(&mut vec![
-            String::from("-f"),
-            self.app.clone().to_str().unwrap().to_owned(),
-        ]);
-        spin_args.append(
-            self.up_args
-                .clone()
-                .into_iter()
-                .collect::<Vec<String>>()
-                .as_mut(),
-        );
-        tracing::info!(
-            "proceeding with command: {} {}",
-            spin_cmd,
-            spin_args.join(" ")
-        );
-        (spin_cmd, spin_args)
+        )
     }
 
-    async fn generate_path_patterns(&self) -> Result<Vec<String>> {
-        // TODO: Spend some time cleaning up this function
+    fn generate_arguments(
+        state: State,
+        up_args: Vec<String>,
+        manifest_path: String,
+    ) -> Vec<String> {
+        let mut spin_args = match state {
+            State::Building => vec![String::from("build")],
+            State::Running => vec![String::from("up")],
+        };
+        spin_args.append(&mut vec![String::from("-f"), manifest_path]);
+        if matches!(state, State::Running) {
+            spin_args.extend(up_args);
+        }
+        spin_args
+    }
+
+    async fn generate_filter_config(&self) -> Result<crate::watch_filter::Config> {
+        let app_dir = parent_dir(&self.app)?;
         let app_manifest = spin_loader::local::raw_manifest_from_file(&self.app)
             .await?
             .into_v1();
 
-        let watch_patterns = app_manifest
-            .components
-            .iter()
-            .filter_map(|c| {
-                if let Some(build) = c.build.as_ref() {
-                    match build.watch.clone() {
-                        Some(watch) => {
-                            let Some(workdir) = build.workdir.clone() else {
-                                return Some(watch);
-                            };
-                            Some(
-                                watch
-                                    .iter()
-                                    .filter_map(|ws| workdir.join(ws).to_str().map(String::from))
-                                    .collect::<Vec<String>>(),
-                            )
-                        }
-                        None => {
-                            eprintln!(
-                                "You haven't configured what to watch for the component: '{}'. Learn how to configure Spin watch at https://developer.fermyon.com/common/cli-reference#watch",
-                                c.id
-                            );
-                            None
-                        }
-                    }
-                } else {
-                    // No build config for this component so lets watch the source instead
-                    if let RawModuleSource::FileReference(path) = &c.source {
-                        if let Some(string_path) = path.to_str() {
-                            return Some(vec![String::from(string_path)]);
-                        }
-                    }
-                    None
-                }
-            })
-            .flatten();
+        // We always want to watch the application manifest
+        let manifest_pattern = WatchPattern::new(
+            self.app
+                .to_str()
+                .with_context(|| format!("non-unicode manifest path {:?}", self.app))?
+                .to_owned(),
+            app_dir.as_path(),
+        )?;
+
+        // We want to watch the source code if we aren't skipping the build step
+        let source_patterns = match !self.skip_build {
+            false => vec![],
+            true => app_manifest
+                .components
+                .iter()
+                .filter_map(|c| WatchCommand::create_source_pattern(c, &app_dir))
+                .flatten()
+                .collect::<Result<Vec<WatchPattern>>>()?,
+        };
+
+        // We want to watch a component's source if it has no build step. If we're skipping
+        // building then we'll watch all component sources.
         let component_source_patterns = app_manifest
             .components
             .iter()
             .filter_map(|c| {
-                if let RawModuleSource::FileReference(path) = &c.source {
-                    return path.to_str();
-                }
-                None
+                let RawModuleSource::FileReference(path) = &c.source else {
+                    return None;
+                };
+                let path_str = path.to_str()?;
+                let watch_this_source = self.skip_build || c.build.is_none();
+                watch_this_source
+                    .then_some(WatchPattern::new(path_str.to_owned(), app_dir.as_path()))
             })
-            .map(String::from);
-        let component_file_patterns = app_manifest
+            .collect::<Result<Vec<WatchPattern>>>()?;
+
+        // We always want to watch component files
+        let files_patterns = app_manifest
             .components
             .iter()
             .filter_map(|c| c.wasm.files.as_ref())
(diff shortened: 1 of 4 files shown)


src/watch_state.rs
```diff
@@ -0,0 +1,246 @@
...
+                    if !self.skip_build {
+                        *state = State::Building;
+                        return self.restart_outcome(false);
+                    }
+                    Outcome::DoNothing
+                }
+                Effect::ArtifactChange => self.restart_outcome(false),
+                Effect::DoNothing => Outcome::if_running(Outcome::DoNothing, Outcome::Start),
+            },
+        };
```
lann, line 73: This might be a little more readable written as e.g.:

```suggestion
        let outcome = match (*state, effect) {
            (_, Effect::Exit) => {
                Outcome::if_running(Outcome::both(Outcome::Stop, Outcome::Exit), Outcome::Exit)
            },
            (_, Effect::ProcessFailed) => Outcome::if_running(Outcome::Stop, Outcome::DoNothing),
            (State::Building, Effect::ProcessCompleted) => {
                *state = State::Running;
                self.restart_outcome(true)
            },
            (State::Building, Effect::ProcessCompleted) => Outcome::DoNothing,
            ...
        };
```

src/watch_state.rs
```diff
@@ -0,0 +1,246 @@
...
+                Outcome::Start,
+            ),
+            Outcome::both(
+                match should_clear {
+                    true => Outcome::Clear,
+                    false => Outcome::DoNothing,
+                },
+                Outcome::Start,
+            ),
+        )
```
lann, line 98: Took me a minute to understand what was going on here. Would this work?

```suggestion
        Outcome::sequence([
            Outcome::if_running(Outcome::Stop, Outcome::DoNothing),
            match should_clear {
                true => Outcome::Clear,
                false => Outcome::DoNothing,
            },
            Outcome::Start,
        ].into_iter())
```

src/commands/watch.rs
```diff
@@ -87,54 +88,81 @@ impl WatchCommand {
...
+            args: vec![],
+        }]);
+        runtime_config.on_pre_spawn(move |prespawn: PreSpawn| {
+            let up_args = self.up_args.clone();
+            let manifest_path = self.app.clone().to_str().unwrap().to_owned();
+            let state_machine = state_machine.clone();
+            async move {
+                // Dynamically modify the command we're running based on the state machine
+                let state = state_machine.get_state();
+                let spin_args = WatchCommand::generate_arguments(state, up_args, manifest_path);
```
lann, line 111: This is pretty subjective, but you might want to consider making the internal `State` private and exposing just what you need here as e.g. `state_machine.is_building() -> bool`

src/watch_filter.rs
```diff
@@ -173,11 +200,68 @@ impl Filterer for WatchFilter {
...
+        Self { app_dir }
+    }
+
+    pub fn watched(&self, glob: String) -> Result<WatchPattern> {
+        self.build(glob, true)
+    }
+
+    pub fn not_watched(&self, glob: String) -> Result<WatchPattern> {
+        self.build(glob, false)
+    }
```
lann, line 249: I get the motivation for doing this, but given that these are only called inside boolean conditional expressions I'm pretty skeptical that this is a net positive. A single e.g. `WatchPattern::new(app_dir, glob, active)` method would cut quite a bit of code and probably be easier to read overall.

src/commands/watch.rs
```diff
@@ -87,54 +88,81 @@ impl WatchCommand {
...
-                            },
-                        ),
-                        Outcome::Start,
-                    ),
-                    Outcome::Start,
-                ));
-
-            Ok::<(), Infallible>(())
-        }});
+                action.outcome(state_machine.handle(effects.reduce()));
```
lann, line 162: I like the use of `Effect` ordering to choose the highest priority effect, but I wonder if you could replace `Effects` with a `Vec<Effect>` and do something like this here:

```suggestion
                // Pick the highest priority effect to apply
                let effect = effects.into_iter().min();
                action.outcome(state_machine.handle(effect));
```

The `Effects` abstraction is nice, but its only used in this one place so maybe not worth the extra code and indirection.

src/watch_filter.rs
```diff
@@ -173,11 +200,68 @@ impl Filterer for WatchFilter {
...
+    pub pattern: Pattern,
+}
+
+impl fmt::Debug for WatchPattern {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "WatchPattern {{ glob: {:?}, active: {:?} }}",
+            self.glob, self.active
+        )
```
lann, line 228: This is what `#[derive(Debug)]` does internally (minus the one field):

```suggestion
        f.debug_struct("WatchPattern")
            .field("active", &self.active)
            .field("glob", &self.glob)
            .finish()
```

lann (approved):
Bunch of style/complexity suggestions, all optional.

src/commands/watch.rs
```diff
@@ -145,116 +173,129 @@ impl WatchCommand {
...
+                    );
+                    return None;
+                };
+                let sources = match build.workdir.clone() {
+                    Some(workdir) => watch
+                        .iter()
+                        .filter_map(|w| workdir.join(w).to_str().map(String::from))
+                        .collect(),
+                    None => watch,
+                };
```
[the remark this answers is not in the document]
calebschoepp: After lots of handholding from Clippy I landed on:

```rust
let sources = build.workdir.clone().map(|workdir| watch.iter()
                        .filter_map(|w| workdir.join(w).to_str().map(String::from))
                        .collect()).unwrap_or(watch);
```

src/commands/watch.rs
calebschoepp: Leaving this one as is if only b/c I make heavy use of `get_state` in the tests.

lann: Its not just about removing code. If you add another internal state like `BuildingAndDownloading` or something, this code (possibly) doesn't actually care about that new state; it just cares about some boolean property of that state.

lann: (all of this with the caveat that it definitely isn't very important in the current codebase; this is just design navel-gazing)

src/commands/watch.rs
```diff
@@ -145,116 +173,129 @@ impl WatchCommand {
...
-                    None
-                }
+                let Some(build) = c.build.as_ref() else {
+                    return None;
+                };
+                let Some(watch) = build.watch.clone() else {
+                    eprintln!(
+                        "You haven't configured what to watch for the component: '{}'. Learn how to configure Spin watch at https://developer.fermyon.com/common/cli-reference#watch",
+                        c.id
+                    );
```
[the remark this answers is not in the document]
calebschoepp: If a component doesn't have a build section then it will never get to this warning. AFAICT it is reasonable to warn with this on any component that has a build section.

src/commands/watch.rs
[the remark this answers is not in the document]
lann: Yeah sorry; I should use an editor when writing some of these suggestions... :grimacing:

src/commands/watch.rs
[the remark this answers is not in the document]
lann: I guess its a little annoying for my pet use-case of watching just the `source`. If I want to suppress this warning I have to copy `source` into `watch`.

src/commands/watch.rs
calebschoepp: Think I'm going to keep this one. I like making the dependency on the ordering very explicit and built into `Effects`. It also means that we can make sure that this ordering isn't subtly broken by unit testing `Effects`.

src/watch_filter.rs
calebschoepp: Thanks. Keep the hits coming 😄 . I'm learning a lot about Rust in these code reviews.

src/watch_state.rs
```diff
@@ -0,0 +1,246 @@
...
+        Effects(vec![Effect::DoNothing])
+    }
+
+    pub fn add(&mut self, effect: Effect) {
+        self.0.push(effect);
+    }
+
+    pub fn reduce(&mut self) -> Effect {
+        self.0.sort();
+        let effect = *self.0.first().unwrap();
```
lann, line 147: ...in that case:

```suggestion
    pub fn reduce(&self) -> Effect {
        let effect = *self.0.min();
```

lann: Hmm...`min` isn't actually a method...what did I forget here?

src/watch_state.rs
calebschoepp: Cool. I never noticed `Outcome::sequence`

src/watch_state.rs
```diff
@@ -0,0 +1,246 @@
+use std::sync::{Arc, Mutex};
+
+use watchexec::action::Outcome;
+
+/// Tracks the current state of spin watch. Handles transitioning states and converting effects
+/// into outcomes based on the current state.
+#[derive(Debug, Clone)]
+pub struct StateMachine {
```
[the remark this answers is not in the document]
calebschoepp: I'm open to changing the name here but I don't think `WatchState` is better and I can't think of something. I'm trying to convey that it is a container tracking the current state and managing transitions.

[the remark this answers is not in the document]
lann: > I'm trying to convey that it is a container tracking the current state and managing transitions.

`WatchState` sounds right to me :shrug:

src/commands/watch.rs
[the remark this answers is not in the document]
calebschoepp: Now I understand. I get why that could be annoying for that use case. I'll file an issue.

src/commands/watch.rs
```diff
@@ -145,116 +173,124 @@ impl WatchCommand {
...
+                    return None;
+                };
+                let Some(path_str) = path.to_str() else {
+                    return None;
+                };
+                let watch_this_source = self.skip_build || c.build.is_none();
+                Some(match watch_this_source {
+                    false => WatchPattern::new(path_str.to_owned(), false, app_dir.as_path()),
+                    true => WatchPattern::new(path_str.to_owned(), true, app_dir.as_path()),
+                })
```
lann, line 265: I suspect you meant to do:

```suggestion
                Some(WatchPattern::new(path_str.to_owned(), watch_this_source, app_dir.as_path()))
```

src/commands/watch.rs
```diff
@@ -145,116 +173,124 @@ impl WatchCommand {
...
-                if let RawModuleSource::FileReference(path) = &c.source {
-                    return path.to_str();
-                }
-                None
+                let RawModuleSource::FileReference(path) = &c.source else {
+                    return None;
+                };
+                let Some(path_str) = path.to_str() else {
+                    return None;
+                };
```
lann, line 260:
```suggestion
                let path_str = path.to_str()?;
```

lann (approved):

src/watch_filter.rs
```diff
@@ -173,15 +200,54 @@ impl Filterer for WatchFilter {
...
 
         // By process of elimination the event is valid
-        tracing::trace!("passing event: {event:?}");
+        tracing::debug!("passing event: {event:?}");
         Ok(true)
     }
 }
 
+#[derive(Debug)]
+pub struct Config {
```
itowlson, line 209: 1000000% banned from calling this `Config`.  We already have 800 things called `Config` and I can no longer keep track of them.  I'd suggest something like `WatchPatterns` or `WhatToWatch` maybe?

itowlson, line 209: I am a bit surprised that the docs thing is not nagging about things being public without doc comments.  I'm not too fussed but it will certainly be needed when you crate it.

src/watch_state.rs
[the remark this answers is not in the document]
itowlson: Unresolving this to agree that `StateMachine` is insufficiently descriptive.  Even just `WatchStateMachine` would help.

src/watch_state.rs
```diff
@@ -0,0 +1,231 @@
...
+    state: Arc<Mutex<State>>,
+    clear: bool,
+    skip_build: bool,
+}
+
+impl StateMachine {
+    pub fn new(skip_build: bool, clear: bool) -> Self {
+        Self {
+            state: Arc::new(Mutex::new(match skip_build {
```
itowlson, line 17: This would be more readable (to me) if you pulled the match out into a `let initial_state = match...`.

src/watch_state.rs
```diff
@@ -0,0 +1,231 @@
...
+/// action has multiple events (this occurs when events are debounced) they will all produce
+/// effects. The highest precedence effect will be chosen to produce the outcome.
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq)]
+pub enum Effect {
+    /// Exit spin watch
+    Exit,
+    /// Either `spin build` or `spin up` failed to run
+    ProcessFailed,
+    /// `spin build` has completed running, `spin up` never completes
+    ProcessCompleted,
```
itowlson, line 104: Reading where this was consumed I wasn't sure what "process" it referred to, whether some internal procedure or an actual system process.  Maybe rename to `BuildCompleted` or `ChildProcessCompleted`?

src/watch_state.rs
```diff
@@ -0,0 +1,231 @@
...
+///
+/// The variants are ordered by highest to lowest precedence so that they can be sorted. When an
+/// action has multiple events (this occurs when events are debounced) they will all produce
+/// effects. The highest precedence effect will be chosen to produce the outcome.
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq)]
+pub enum Effect {
+    /// Exit spin watch
+    Exit,
+    /// Either `spin build` or `spin up` failed to run
+    ProcessFailed,
```
itowlson, line 102: Consider `ChildProcessFailed` for a bit more clarify at the call site?  I realise it is more verbose though.

src/watch_state.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub fn get_state(&self) -> State {
+        *self.state.lock().unwrap()
+    }
+
+    /// Based on the given effect return the correct outcome. Transition the internal state of the
+    /// machine if necessary.
+    pub fn handle(&self, effect: Effect) -> Outcome {
+        let mut state = self.state.lock().unwrap();
+        tracing::debug!("handling effect {:?} in current state {:?}", effect, *state);
+        let outcome = match (effect, *state) {
```
itowlson, line 36: I really like the clarity (and testability) of the state machine approach.  Very nice.

src/watch_state.rs
```diff
@@ -0,0 +1,231 @@
...
+            (Effect::Exit, _) => {
+                Outcome::if_running(Outcome::both(Outcome::Stop, Outcome::Exit), Outcome::Exit)
+            }
+            (Effect::ProcessFailed, _) => Outcome::if_running(Outcome::Stop, Outcome::DoNothing),
```
itowlson, line 40: I am struggling here a bit with `Outcome::if_running` and `State::Running`.  Does `if_running` refer to the `State`, or to something that is happening within the state machine itself?

src/watch_state.rs
```diff
@@ -0,0 +1,231 @@
...
+            }
+            (Effect::ArtifactChange, State::Building) => Outcome::DoNothing,
+            (Effect::ArtifactChange, State::Running) => self.restart_outcome(false),
+            (Effect::DoNothing, _) => Outcome::if_running(Outcome::DoNothing, Outcome::Start),
+        };
+        tracing::debug!("now in {:?} state with outcome {:?}", *state, outcome);
+        outcome
+    }
+
+    fn restart_outcome(&self, prevent_clear: bool) -> Outcome {
```
itowlson, line 69: The boolean here is not clear at the call site (I was reading the state machine going "what are all these false values"). It may seem a bit laborious but what about an enum a la `Clear::Allow` / `Clear::Prevent`?

src/watch_filter.rs
```diff
@@ -17,49 +12,64 @@ use watchexec::{
...
 };
 
+/// Filters Watchexec events.
+///
+/// Also enables checking whether an application manifest, source, or artifact was modified.
 #[derive(Debug)]
-pub struct WatchFilter {
-    path_patterns: Vec<Pattern>,
-    ignore_patterns: Vec<Pattern>,
+pub struct Filter {
```
itowlson, line 19: Consider keeping a mention of "watch" in the name.

src/watch_filter.rs
```diff
@@ -76,19 +86,24 @@ impl WatchFilter {
...
-            self.ignore_patterns
-                .iter()
-                .any(|pattern| pattern.matches_path(path))
+            self.config.manifest_pattern.pattern.matches_path(path)
+                || self
+                    .config
+                    .source_patterns
+                    .iter()
+                    .chain(self.config.artifact_patterns.iter())
+                    .any(|wp| wp.active && wp.pattern.matches_path(path))
```
itowlson, line 97: This checks `wp.active` whereas the earlier `matches_*_pattern` functions didn't.  Is this intentional?  If so, consider a comment on why; if not, maybe refactor so this calls the `matches_*_pattern` functions so they are guaranteed to stay in sync.

src/watch_filter.rs
```diff
@@ -76,19 +86,24 @@ impl WatchFilter {
...
 
-    fn matches_one_of_path_patterns(&self, event: &Event) -> bool {
+    fn matches_one_of_ignore_patterns(&self, event: &Event) -> bool {
         event.paths().any(|(path, _)| {
-            self.path_patterns
+            self.config
+                .ignore_patterns
                 .iter()
-                .any(|pattern| pattern.matches_path(path))
+                .any(|wp| wp.active && wp.pattern.matches_path(path))
```
itowlson, line 106: I like how you subtly changed the name of the lambda argument to avoid `cargo fmt` messing up the code structure.  _shakes puny fist at `cargo fmt`_

src/watch_filter.rs
```diff
@@ -127,31 +142,43 @@ impl WatchFilter {
...
+            .any(|t| matches!(t, Tag::ProcessCompletion(_)))
+        {
+            tracing::debug!("passing event (process completion): {event:?}");
+            return Ok(true);
+        }
+
         // Fail if the event kind wasn't creation, modification, or deletion
         if !self.has_valid_event_kind(event) {
-            tracing::trace!("failing event b/c of event kind: {event:?}");
+            tracing::trace!("failing event (wrong event kind): {event:?}");
```
itowlson, line 167: Is this "wrong" or just "we don't care about it"?

src/watch_filter.rs
```diff
@@ -173,15 +200,54 @@ impl Filterer for WatchFilter {
...
+#[derive(Debug)]
+pub struct Config {
+    pub manifest_pattern: WatchPattern,
+    pub source_patterns: Vec<WatchPattern>,
+    pub artifact_patterns: Vec<WatchPattern>,
+    pub ignore_patterns: Vec<WatchPattern>,
+}
+
+pub struct WatchPattern {
+    pub active: bool,
```
itowlson, line 217: When reading both the consuming code and the generation code, I found it a bit hard to follow what `active` was doing.  I got the impression that `active = false` meant "don't watch this," and that it was a permanent state, i.e. not temporarily moved inactive then reactivated again.  But if that's the case, why have the pattern at all?  Why not just leave it out?  So I'm clearly missing something (perhaps relating to how `active` is sometimes checked and sometimes not).  It would be good to comment the `active` field and/or rename it to make the intended use clearer.

src/commands/watch.rs
```diff
@@ -145,116 +173,126 @@ impl WatchCommand {
...
-                } else {
-                    // No build config for this component so lets watch the source instead
-                    if let RawModuleSource::FileReference(path) = &c.source {
-                        if let Some(string_path) = path.to_str() {
-                            return Some(vec![String::from(string_path)]);
-                        }
-                    }
-                    None
-                }
+                let build = c.build.as_ref()?;
```
itowlson, line 224: I'd strongly encourage you to move this lambda out to a named function ~so I can understand it~ to more visibly express intent.

src/commands/watch.rs
```diff
@@ -145,116 +173,126 @@ impl WatchCommand {
...
-        };
-        // Always watch spin.toml
-        path_patterns.push(String::from(
-            self.app.to_str().expect("conversion to str not to fail"),
-        ));
+            .collect::<Result<Vec<WatchPattern>>>()?;
 
-        Ok(path_patterns)
+        // We always want to watch component files
+        artifact_patterns.append(
```
itowlson, line 272: Consider giving the files iterator a name so you can write something like `let artifact_patterns = concat(source_patterns, files_patterns)`.

src/commands/watch.rs
```diff
@@ -145,116 +173,126 @@ impl WatchCommand {
...
+            &mut app_manifest
```
itowlson, line 273: Surely we can't need to mutate the application manifest here?!

itowlson (changes requested):
Looks good (despite the comments - mostly stylistic nits I think) - thanks!

src/watch_filter.rs
lann: Err...this is a very common pattern in rust packages...

lann: The module isn't public, so the struct isn't crate-public.

lann: To be more explicit: I think this is an appropriate name and certainly shouldn't be banned. Naming a type `Config` - like `Result`, `Error`, `Iter`, `Entry` - is a well-established convention. They can be differentiated by consumers by renaming (`use std::io::Error as IoError`) or module path (`watch_filter::Config`).

As a blatant appeal to authority, `wasmtime` appears to have 6 `struct Config`s: https://github.com/search?q=repo%3Abytecodealliance%2Fwasmtime+%22struct+Config%22&type=code

src/watch_state.rs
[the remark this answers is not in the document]
calebschoepp: Two's a pattern. Going to go with `WatchState`

src/watch_state.rs
calebschoepp: Thanks. It was really fun to design and code.

src/watch_state.rs
calebschoepp: I was debating if this deserved an enum and I decided it didn't. On second thought I agree that it does.

src/watch_filter.rs
calebschoepp: Since I'm planning to move this to it's own crate I'm going to keep the name as just `Filter` since it is implied that it is for watching -- given the crate.

src/commands/watch.rs
calebschoepp: `append` expects a `&mut`. We're taking a mutable reference to the collected vector `collect::<Result<Vec<WatchPattern>>>()`.

src/commands/watch.rs
calebschoepp: Done

src/watch_state.rs
calebschoepp: I actually opted for consts. https://stackoverflow.com/questions/52545734/is-there-a-way-to-use-a-boolean-as-an-underlying-type-for-an-enum.

I felt mapping between bool and an enum was a little clunky.

src/watch_state.rs
calebschoepp: I like it.

src/watch_state.rs
calebschoepp: I like this too.

src/watch_state.rs
calebschoepp: `Outcome::if_running` refers to [this](https://docs.rs/watchexec/latest/watchexec/action/enum.Outcome.html). Notably it says:

> Logic against the state of the command should be expressed using these variants, rather than inside the action handler, as it ensures the state of the command is always the latest available when the outcome is executed.

`State::Running` means we believe that `spin up` is currently running or at least we want it to be.

Basically wrapping all the outcomes in `Outcome::if_running` protects us from a scenario in which somehow our internal state gets out of sync with the actual state of watchexec. Likely not to matter but good protection regardless.

I added a comment to try and add some clarity.

src/watch_filter.rs
calebschoepp: Good call out. `active` is in fact permanent. A better name is `should_watch`. The reason we need this field is because we want to have all possible patterns in the filter so that we can match them in the `matches_*_pattern` methods. However, we don't always want to pay attention to changes in these patterns hence `should_watch`.

calebschoepp: Renamed and commented.

src/watch_filter.rs
calebschoepp: This is on purpose. Cleared up in code by renaming `active` and adding more comments.

calebschoepp:
@itowlson I've addressed your changes. Looking for a re-review.

src/commands/watch.rs
itowlson: Doh, that makes sense - thanks!  But I appreciate you refactoring it away anyway so I don't come across it in six months and ask all over again!

itowlson (approved):
I know we have an open question about whether `should_watch` is needed, but the rest of it looks great.  Approving so you can merge during my weekend (if you want to) regardless of the outcome of your `should_watch` investigation - thanks, this is good stuff!

calebschoepp:
> I know we have an open question about whether `should_watch` is needed, but the rest of it looks great. Approving so you can merge during my weekend (if you want to) regardless of the outcome of your `should_watch` investigation - thanks, this is good stuff!

I ended up pulling it out

Issue #1400, opened 2023-04-17
Tests do not respect `CARGO_TARGET_DIR`


rylev:
This is similar to #1385 but different enough that I think it warrants its own issue.

I use `CARGO_TARGET_DIR` to set my target directory to somewhere in my home directory. The tests, however, assume that spin is located in `./target/{debug,release}/`. Ideally we would use the `CARGO_TARGET_DIR` to detect the location of the target directory and form the path correctly.

rajatjindal:
Hi @rylev thank you for reporting this ticket. 

do you think if we add logic in Makefile to use `CARGO_TARGET_DIR` (with default set to `target` in current dir), and then append `debug or release` to that path will be ok?

rylev:
I think that will probably work. We should give it a try!

vdice:
This one appears to have been resolved by https://github.com/fermyon/spin/pull/1422.

PR #1422, opened 2023-04-25
add CARGO_TARGET_DIR/{debug,release} to path

### LINKED ISSUE #1385: Dev Loop Test Failure
When running `make test` spin is not present in dev container/system path which results in the following error:

---
running 1 test
**"/bin/bash: line 1: spin: command not found\n"**
""
thread 'spinup_tests::key_value_works' panicked at 'command `cd "/workspaces/spin/crates/e2e-testing/../../tests/testcases/key-value" && "/bin/bash" "-c" "spin build"` exited with code 127', crates/e2e-testing/src/utils.rs:50:9
test spinup_tests::key_value_works ... FAILED

failures:

failures:
    spinup_tests::key_value_works

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 20 filtered out; finished in 0.00s

error: test failed, to rerun pass `--test spinup_tests`
error: 1 target failed:
    `--test spinup_tests`
make: *** [Makefile:64: test-kv] Error 101

rajatjindal:
fixes #1385 #1400


rylev (approved):

vdice (approved):

itowlson:
@rajatjindal Is it okay to merge this?

rajatjindal:
Hi @itowlson, I need to verify that it works fine with local tests/ci and docker based e2e tests before merging. Also need to re-visit my conversation with Ryan who brought up a scenario where this may cause more pain. 

I will try to get this done this week.

rajatjindal:
I looked a bit more into this. If we set the CARGO_TARGET_DIR env, the tests on local setup would fail as the `wasm` file will be generated in a different folder then the one we are expecting in spin.toml file. 

I think to be able to use this env variable, we need to teach the build command inside `spin toml` to place the file at the right location. 

e.g.

```
[[component]]
id = "hello"
source = "target/wasm32-wasi/release/key_value.wasm"
[component.trigger]
route = "/..."
[component.build]
command = "cargo build --release --target wasm32-wasi"
```

in the above example, we expect the `source` wasm file to be in specific location. if we change the CARGO_TARGET_DIR, then the test fails as it is unable to find `target/wasm32-wasi/release/key_value.wasm` file.

rajatjindal:
I have updated the PR to include following dir in PATH (in that order)

`target/debug`
`target/release`
`$HOME/.cargo/bin`
`$PATH`

I have also modified the `testcase::run` function to print following info about which `spin` binary and version is being used for running the test.

e.g.

```
[testcase::run] Running testcase "config-variables" using spin ("/usr/local/bin/spin") with version "spin 1.4.0-pre0 (2e34ef5 2023-06-14)"
```

lann:
This doesn't appear to affect the version of spin used in e2e tests. Is that expected?

rajatjindal:
I am trying to ensure this cover the common usecases:

- Running on CI
- Running on local laptop (outside container and inside container) with Mac/Windows/Linux and AMD64/ARM64

rajatjindal:
following is the summary of changes now:

- for each test case print path to the spin binary is used and the version of that binary
- when running tests on host system (e.g. make test-kv), it uses the binary available in target/{debug,release} or ~/.cargo/bin or $PATH (in that order)
- when running tests inside the container (e.g. make test-spin-up), if host is a MAC-OS system, it compiles spin inside the container by default and use that. It makes use of docker volumes to cache stuff for faster iterations.
- Mounts target/{debug,release} dir inside container if host is a linux system

Makefile
```diff
@@ -1,25 +1,42 @@
 LOG_LEVEL ?= spin=trace
 CERT_NAME ?= local
 SPIN_DOC_NAME ?= new-doc.md
+export PATH := target/debug:target/release:$(HOME)/.cargo/bin:$(PATH)
```
rylev, line 4: Shouldn't this use `CARGO_TARGET_DIR` or the `target` directory if `CARGO_TARGET_DIR` is not set?

crates/e2e-testing/src/spin.rs
```diff
@@ -117,5 +117,17 @@ pub fn registry_login(registry_url: &str, username: &str, password: &str) -> Res
 pub fn version() -> Result<String> {
     let output = utils::run(&["spin", "--version"], None, None)?;
     utils::assert_success(&output);
-    Ok(format!("{:#?}", std::str::from_utf8(&output.stdout)?))
+    Ok(format!(
+        "{:#?}",
+        std::str::from_utf8(&output.stdout)?.trim().to_string()
```
rylev, line 122: I think you want this?
```suggestion
        "{}",
        std::str::from_utf8(&output.stdout)?.trim()
```

crates/e2e-testing/src/spin.rs
```diff
@@ -117,5 +117,17 @@ pub fn registry_login(registry_url: &str, username: &str, password: &str) -> Res
...
+    ))
+}
+
+pub fn which_spin() -> Result<String> {
+    let output = utils::run(&["which", "spin"], None, None)?;
+    utils::assert_success(&output);
+    Ok(format!(
+        "{:#?}",
+        std::str::from_utf8(&output.stdout)?.trim().to_string()
```
rylev, line 131: Same as above.

crates/e2e-testing/src/testcase.rs
```diff
@@ -216,3 +218,14 @@ impl TestCase {
...
 }
+
+fn print_version_info(testcase_name: String) {
+    let version = spin::version();
+    let which_spin = spin::which_spin();
+    println!(
+        r###"[testcase::run] Running testcase "{}" using spin ({}) with version {}"###,
+        testcase_name,
+        which_spin.unwrap_or("unknown".to_string()),
+        version.unwrap_or("unknown".to_string())
```
rylev, line 229:
```suggestion
        which_spin.as_deref().unwrap_or("unknown"),
        version.as_deref().unwrap_or("unknown")
```

Makefile
```diff
@@ -1,25 +1,42 @@
...
+
+ifneq (,$(wildcard $(shell pwd)/target/release/spin))
+	E2E_SPIN_RELEASE_VOLUME_MOUNT = -v $(shell pwd)/target/release/spin:/from-host/target/release/spin
+endif
+
+ifneq (,$(wildcard $(shell pwd)/target/debug/spin))
+	E2E_SPIN_DEBUG_VOLUME_MOUNT = -v $(shell pwd)/target/debug/spin:/from-host/target/debug/spin
+endif
+
+## Reset volume mounts for e2e-tests if Darwin
```
rylev, line 35: Can we note that the reason we do this is because the spin binaries built on macOS won't run in the Docker container?

Makefile
```diff
@@ -1,25 +1,42 @@
...
 ## dependencies for e2e-tests
-E2E_VOLUME_MOUNT     ?=
-E2E_BUILD_SPIN       ?= false
-E2E_TESTS_DOCKERFILE ?= e2e-tests.Dockerfile
-MYSQL_IMAGE          ?= mysql:8.0.22
-REDIS_IMAGE          ?= redis:7.0.8-alpine3.17
-POSTGRES_IMAGE       ?= postgres:14.7-alpine
-REGISTRY_IMAGE       ?= registry:2
+E2E_SPIN_RELEASE_VOLUME_MOUNT   ?=
+E2E_SPIN_DEBUG_VOLUME_MOUNT     ?=
```
rylev, line 10: What's the reason for keeping this separate? Wouldn't a better interface be for the user to specify which spin binary they want to run, and then we copy that binary into the docker container instead of mounting the entire directory?

Makefile
rajatjindal: when using cargo_target_dir, we had issues with where `.wasm` files would be saved when running `spin build`.

crates/e2e-testing/src/spin.rs
rajatjindal: done

crates/e2e-testing/src/spin.rs
rajatjindal: done

crates/e2e-testing/src/testcase.rs
rajatjindal: TIL, thank you. done.

Makefile
rajatjindal: 👍

Makefile
rajatjindal: but what would be the fallback if spin binary path is not provided by user

crates/e2e-testing/src/testcase.rs
```diff
@@ -216,3 +218,14 @@ impl TestCase {
         }
     }
 }
+
+fn print_version_info(testcase_name: String) {
```
rylev, line 222:
```suggestion
fn print_version_info(testcase_name: &str) {
```

crates/e2e-testing/src/testcase.rs
```diff
@@ -109,6 +109,8 @@ impl TestCase {
     ///
     /// The output from running `spin build` is returned.
     async fn do_run(&self, controller: &dyn Controller, bail_on_run_failure: bool) -> Result<()> {
+        print_version_info(self.name.clone());
```
rylev, line 112:
```suggestion
        print_version_info(&self.name);
```

Makefile
rylev: We could still prob for a binary using the same order that you currently do (debug, release, .cargo/bin, $PATH) but instead of mounting the various locations of the various spin binaries directly into the volume, we could instead copy the binary to a new location and mount just that directory that just contains the spin binary.

This is just an idea to make the set up a bit easier to follow since we have several mounting options.

Makefile
rylev: I guess I'm confused what this line is for?

rajatjindal: this is for local laptop scenario. e.g. if you run `make test-kv`, it runs the test directly on the laptop (instead of a container), so idea is to use spin from either of the directory (in that order)

crates/e2e-testing/src/testcase.rs
rajatjindal: done. thank you

crates/e2e-testing/src/testcase.rs
rajatjindal: done, thank you

Makefile
rylev: Ah you're right, sorry. I think in general, I'm really bad at Make, and I'd like to simplify this wherever possible.

Makefile
rajatjindal: Hi @rylev, I want to make sure I have answered all your questions before I mark this comment as resolved. pls let me know if you have more questions.

Makefile
rajatjindal: hi @rylev, do u have an alternate approach in mind or the current change looks ok to you?

Makefile
rylev: Ok this is fine for now. I just really wish all the tests could rely on a spin binary being in one canonical location and then we could have logic to decide how to copy a binary from a number of different possible locations to that canonical location. Right now, it feels a bit of like a house of cards as to which spin binary will actually be used.

That doesn't need to delay merging this anymore.

rylev (approved):

PR #1424, opened 2023-04-26
Update wit bindgen

rylev:
This PR updates our dependency on wit-bindgen to 0.5 (or more accurately [this commit](f4f5bb8169d62c677df5f8a03c0f7563cd55cb5e)). Unmodified components built using the old version of spin and the old SDK should still run correctly. 

There are a few considerations that need to be discussed before this can be merged:

* This will require changes to Fermyon Cloud before components built with the new version of the SDK can be run. 
* This requires a change to `spin-componentize` which will have to handle three types of input: a wasm component, a core module using wit-bindgen 0.2, and a new core module using wit-bindgen 0.5. (**TODO**: I'm having issues pushing to `spin-componentize` and thus can't link to a PR just yet) 
  * The logic for deciding whether a module uses wit-bindgen 0.5 or the older 0.2 is based on the embedded spin sdk version. 
  * We'll likely want to bump the "pre" version of spin so that when this lands it is the start of a new pre-release.
* There are a few subtle things that change that we need to decide how they interact with backwards compatibility:
  * The wit files have changed a bit and thus if users are using wit files directly (instead of using the SDKs) their code may break.
  * Spin can not handle modules using wit-bindgen 0.5 that *don't* embed an SDK version (this is what spin-componentize uses to different the two types). Therefore components *not* using the SDK won't work unless they also embed an SDK version.
  * `spin_sdk::key_value::Error` used to have its `Error` and `Display` impls managed by the SDK but it seems wit-bindgen is now implementing these. I need to investigate this.
* Templates still refer to the old SDK and old version of wit-bindgen. We'll need to coordinate upgrading those templates correctly.
* I've only updated the Rust SDK. We still need to update other language SDKs.


rylev:
The `` test is failing with the following error:

```
Error: Failed to instantiate component 'hello'

Caused by:
    0: import `filesystem` has the wrong type
    1: instance export `open-at` has the wrong type
    2: type mismatch with parameters
    3: expected flags type with 7 names, found 6 names

```

This might be due to a mismatch in wasi versions. Perhaps the wasi that we are using in `spin-componentize` and the version that `wasmtime` is expecting might not be lining up. It [seems others are facing such an issue](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime/topic/compatibility.20matrix.20update/near/354993897). I need to investigate this more though.

PR #1426, opened 2023-04-27
fix: Upgrade clap and suppress clippy error

lann:
Clap generated code was broken by a new clippy warning in Rust 1.69. This was partially fixed in clap 3.2.24 along with a clippy exception.

Ref: https://github.com/clap-rs/clap/issues/4857

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -1,3 +1,6 @@
+// Needed for clap derive: https://github.com/clap-rs/clap/issues/4857
+#![allow(clippy::almost_swapped)]
+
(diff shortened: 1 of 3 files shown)


src/commands/plugins.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::almost_swapped)] // needed for clap derive
```
rylev, line 1: Should we link to the [issue](https://github.com/clap-rs/clap/issues/4857) where this is being fixed? That makes it easy to tell when this should be safe to remove.

rylev (approved):

PR #1433, opened 2023-04-28
Upgrade to wasmtime 8.0.1

rylev:
This brings spin much further forward with regards to its usage of upstream wasm libraries which enables a lot more experimentation with new standards around the component model. For users this should effectively be a noop.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -845,18 +845,18 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.94.1"
+version = "0.95.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0853f4732d9557cc1f3b4a97112bd5f00a7c619c9828edb45d0a2389ce2913f9"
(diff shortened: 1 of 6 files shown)


dicej (approved):

Mossaka:
This PR might resolve issue #1437

rylev:
The failing test is due to the fact that we're defining our spin world in two different ways (due to the fact that our SDKs use old versions of wit-bindgen - fixed in #1424 - while our host implementation uses newer style wit documents). The config interface is called "spin-config" in the SDK while the host expects just "config". `spin-componentize` is smoothing this out for older modules which is why this wasn't an issue before, but we updated the failing test to use `wit-bindgen` 0.6 which thus skips over the `spin-componentize` world massaging. 

For now, we can merge https://github.com/fermyon/spin-componentize/pull/7 and revert the broken test to use `wit-bindgen` 0.4, but we'll eventually want to fix the issue of our world definitions not aligning. When #1424 is merged we can move to using the same wit definitions for both guest and host.

rylev:
Superseded by #1439.

Issue #965, opened 2022-12-12
Define a strategy for shared utility code


itowlson:
We are accumulating bits of code that are common across multiple crates, such as sloth warning and git error handling.  At the moment we are tending to duplicate these because there is no obvious shared place to put them.  We should figure out how to approach this in the long term:

1. What are the criteria at which something merits becoming a crate in itself as opposed to a hunk of duplicate code (or a module in the util crate)?  Considerations:
  a. At the moment our crates are generally aligned with functional components of Spin.
  b. Ergonomic impact of having even more top-level things in the folder tree.
2. Are we ready to accept the inevitable and have a `spin-util` crate, or do we still have a few of the stages of grief to pass through first?
3. Are we better off having fewer crates, and using modules to organise functionality and features to enable third parties to be selective about what they pull in?  (Would that even help, or would it just push the "shall we have `utils` or shall we have lots of little modules" problem down a level in the folder tree?)

PR #1435, opened 2023-04-28
`spin doctor` prototype

lann:
The subcommand is currently hidden from e.g. `spin --help`.

```console
$ spin doctor -f examples/http-rust/spin.toml
📟 The Spin Doctor is in.
🩺 Checking examples/http-rust/spin.toml...

❗ Diagnosis: Manifest 'spin_manifest_version' must be "1", not 1
🩹 The Spin Doctor can help! Would you like to:
  Set manifest version to "1"
  Do nothing
> See more details about the recommended changes

📋 Apply the following diff to "examples/http-rust/spin.toml":
@@ -1,2 +1,2 @@
-spin_manifest_version = 1
+spin_manifest_version = "1"
 authors = ["Fermyon Engineering <engineering@fermyon.com>"]

Would you like to apply this fix? [Y/n] 
❤️ Treatment applied!
```

### DIFF
--- a/crates/doctor/Cargo.toml
+++ b/crates/doctor/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "spin-doctor"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+anyhow = "1"
+async-trait = "0.1"
+serde = { version = "1", features = ["derive"] }
+similar = "2"
+spin-loader = { path = "../loader" }
+tokio = "1"
+toml = "0.7"
+toml_edit = "0.19"
+tracing = { workspace = true }
+
+[dev-dependencies]
+tempfile = "3"
--- a/crates/doctor/src/lib.rs
+++ b/crates/doctor/src/lib.rs
@@ -1,7 +1,7 @@
 //! Spin doctor: check and automatically fix problems with Spin apps.
 #![deny(missing_docs)]
 
-use std::{fmt::Debug, fs, future::Future, path::PathBuf, pin::Pin, process::Command, sync::Arc};
+use std::{fmt::Debug, fs, future::Future, path::PathBuf, pin::Pin, sync::Arc};
 
 use anyhow::{ensure, Context, Result};
 use async_trait::async_trait;
@@ -135,29 +135,36 @@ pub trait Diagnosis: Debug + Send + Sync + 'static {
 /// The Treatment trait represents a (potential) fix for a detected problem.
 #[async_trait]
 pub trait Treatment: Sync {
-    /// Return a human-readable description of what this treatment will do to
-    /// fix the problem, such as a file diff.
-    async fn description(&self, patient: &PatientApp) -> Result<String>;
+    /// Return a short (single line) description of what this fix will do, as
+    /// an imperative, e.g. "Upgrade the library".
+    fn summary(&self) -> String;
+
+    /// Return a detailed description of what this fix will do, such as a file
+    /// diff or list of commands to be executed.
+    ///
+    /// May return `Err(DryRunNotSupported.into())` if no such description is
+    /// available, which is the default implementation.
+    async fn dry_run(&self, patient: &PatientApp) -> Result<String> {
+        let _ = patient;
+        Err(DryRunNotSupported.into())
+    }
 
     /// Attempt to fix this problem. Return Ok only if the problem is
     /// successfully fixed.
     async fn treat(&self, patient: &mut PatientApp) -> Result<()>;
 }
 
-const SPIN_BIN_PATH: &str = "SPIN_BIN_PATH";
+/// Error returned by [`Treatment::dry_run`] if dry run isn't supported.
+#[derive(Debug)]
+pub struct DryRunNotSupported;
 
-/// Return a [`Command`] targeting the `spin` binary. The `spin` path is
-/// resolved to the first of these that is available:
-/// - the `SPIN_BIN_PATH` environment variable
-/// - the current executable ([`std::env::current_exe`])
-/// - the constant `"spin"` (resolved by e.g. `$PATH`)
-pub fn spin_command() -> Command {
-    let spin_path = std::env::var_os(SPIN_BIN_PATH)
-        .map(PathBuf::from)
-        .or_else(|| std::env::current_exe().ok())
-        .unwrap_or("spin".into());
-    Command::new(spin_path)
+impl std::fmt::Display for DryRunNotSupported {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "dry run not implemented for this treatment")
+    }
 }
+
+impl std::error::Error for DryRunNotSupported {}
 
 #[async_trait]
 trait BoxingDiagnostic {
--- a/crates/doctor/src/lib.rs
+++ b/crates/doctor/src/lib.rs
@@ -1,0 +1,161 @@
+//! Spin doctor: check and automatically fix problems with Spin apps.
+#![deny(missing_docs)]
+
+use std::{fmt::Debug, fs, future::Future, path::PathBuf, pin::Pin, process::Command, sync::Arc};
+
+use anyhow::{ensure, Context, Result};
+use async_trait::async_trait;
+use tokio::sync::Mutex;
+use toml_edit::Document;
+
+/// Diagnoses for app manifest format problems.
+pub mod manifest;
+/// Test helpers.
+pub mod test;
+/// Diagnoses for Wasm source problems.
+pub mod wasm;
+
+/// Configuration for an app to be checked for problems.
+pub struct Checkup {
+    manifest_path: PathBuf,
+    diagnose_fns: Vec<DiagnoseFn>,
+}
+
+type DiagnoseFut<'a> =
+    Pin<Box<dyn Future<Output = Result<Vec<Box<dyn Diagnosis + 'static>>>> + 'a>>;
+type DiagnoseFn = for<'a> fn(&'a PatientApp) -> DiagnoseFut<'a>;
+
+impl Checkup {
+    /// Return a new checkup for the app manifest at the given path.
+    pub fn new(manifest_path: impl Into<PathBuf>) -> Self {
+        let mut checkup = Self {
+            manifest_path: manifest_path.into(),
+            diagnose_fns: vec![],
+        };
+        checkup.add_diagnose::<manifest::version::VersionDiagnosis>();
+        checkup.add_diagnose::<manifest::trigger::TriggerDiagnosis>();
+        checkup.add_diagnose::<wasm::missing::WasmMissing>();
+        checkup
+    }
+
+    /// Add a detectable problem to this checkup.
+    pub fn add_diagnose<D: Diagnose + 'static>(&mut self) -> &mut Self {
+        self.diagnose_fns.push(|patient| {
+            Box::pin(async {
+                let diags = D::diagnose(patient).await?;
+                Ok(diags.into_iter().map(|diag| Box::new(diag) as _).collect())
+            })
+        });
+        self
+    }
+
+    fn patient(&self) -> Result<PatientApp> {
+        let path = &self.manifest_path;
+        ensure!(
+            path.is_file(),
+            "No Spin app manifest file found at {path:?}"
+        );
+
+        let contents = fs::read_to_string(path)
+            .with_context(|| format!("Couldn't read Spin app manifest file at {path:?}"))?;
+
+        let manifest_doc: Document = contents
+            .parse()
+            .with_context(|| format!("Couldn't parse manifest file at {path:?} as valid TOML"))?;
+
+        Ok(PatientApp {
+            manifest_path: path.into(),
+            manifest_doc,
+        })
+    }
+
+    /// Find problems with the configured app, calling the given closure with
+    /// each problem found.
+    pub async fn for_each_diagnosis<F>(&self, mut f: F) -> Result<usize>
+    where
+        F: for<'a> FnMut(
+            Box<dyn Diagnosis + 'static>,
+            &'a mut PatientApp,
+        ) -> Pin<Box<dyn Future<Output = Result<()>> + 'a>>,
+    {
+        let patient = Arc::new(Mutex::new(self.patient()?));
+        let mut count = 0;
+        for diagnose in &self.diagnose_fns {
+            let patient = patient.clone();
+            let diags = diagnose(&*patient.lock().await)
+                .await
+                .unwrap_or_else(|err| {
+                    tracing::debug!("Diagnose failed: {err:?}");
+                    vec![]
+                });
+            count += diags.len();
+            for diag in diags {
+                let mut patient = patient.lock().await;
+                f(diag, &mut patient).await?;
+            }
+        }
+        Ok(count)
+    }
+}
+
+/// An app "patient" to be checked for problems.
+#[derive(Clone)]
+pub struct PatientApp {
+    /// Path to an app manifest file.
+    pub manifest_path: PathBuf,
+    /// Parsed app manifest TOML document.
+    pub manifest_doc: Document,
+}
+
+/// The Diagnose trait implements the detection of a particular Spin app problem.
+#[async_trait]
+pub trait Diagnose: Diagnosis + Send + Sized + 'static {
+    /// Check the given [`Patient`], returning any problem(s) found.
+    async fn diagnose(patient: &PatientApp) -> Result<Vec<Self>>;
+}
+
+/// The Diagnosis trait represents a detected problem with a Spin app.
+pub trait Diagnosis: Debug + Send + Sync {
+    /// Return a human-friendly description of this problem.
+    fn description(&self) -> String;
+
+    /// Return true if this problem is "critical", i.e. if the app's
+    /// configuration or environment is invalid. Return false for
+    /// "non-critical" problems like deprecations.
+    fn is_critical(&self) -> bool {
+        true
+    }
+
+    /// Return a [`Treatment`] that can (potentially) fix this problem, or
+    /// None if there is no automatic fix.
+    fn treatment(&self) -> Option<&dyn Treatment> {
+        None
+    }
+}
+
+/// The Treatment trait represents a (potential) fix for a detected problem.
+#[async_trait]
+pub trait Treatment: Sync {
+    /// Return a human-readable description of what this treatment will do to
+    /// fix the problem, such as a file diff.
+    async fn description(&self, patient: &PatientApp) -> Result<String>;
+
+    /// Attempt to fix this problem. Return Ok only if the problem is
+    /// successfully fixed.
+    async fn treat(&self, patient: &mut PatientApp) -> Result<()>;
+}
+
+const SPIN_BIN_PATH: &str = "SPIN_BIN_PATH";
+
+/// Return a [`Command`] targeting the `spin` binary. The `spin` path is
+/// resolved to the first of these that is available:
+/// - the `SPIN_BIN_PATH` environment variable
+/// - the current executable ([`std::env::current_exe`])
+/// - the constant `"spin"` (resolved by e.g. `$PATH`)
+pub fn spin_command() -> Command {
+    let spin_path = std::env::var_os(SPIN_BIN_PATH)
+        .map(PathBuf::from)
+        .or_else(|| std::env::current_exe().ok())
+        .unwrap_or("spin".into());
+    Command::new(spin_path)
+}
--- a/crates/doctor/src/lib.rs
+++ b/crates/doctor/src/lib.rs
@@ -18,34 +18,25 @@ pub mod wasm;
 /// Configuration for an app to be checked for problems.
 pub struct Checkup {
     manifest_path: PathBuf,
-    diagnose_fns: Vec<DiagnoseFn>,
+    diagnostics: Vec<Box<dyn BoxingDiagnostic>>,
 }
-
-type DiagnoseFut<'a> =
-    Pin<Box<dyn Future<Output = Result<Vec<Box<dyn Diagnosis + 'static>>>> + 'a>>;
-type DiagnoseFn = for<'a> fn(&'a PatientApp) -> DiagnoseFut<'a>;
 
 impl Checkup {
     /// Return a new checkup for the app manifest at the given path.
     pub fn new(manifest_path: impl Into<PathBuf>) -> Self {
         let mut checkup = Self {
             manifest_path: manifest_path.into(),
-            diagnose_fns: vec![],
+            diagnostics: vec![],
         };
-        checkup.add_diagnose::<manifest::version::VersionDiagnosis>();
-        checkup.add_diagnose::<manifest::trigger::TriggerDiagnosis>();
-        checkup.add_diagnose::<wasm::missing::WasmMissing>();
+        checkup.add_diagnostic::<manifest::version::VersionDiagnostic>();
+        checkup.add_diagnostic::<manifest::trigger::TriggerDiagnostic>();
+        checkup.add_diagnostic::<wasm::missing::WasmMissingDiagnostic>();
         checkup
     }
 
     /// Add a detectable problem to this checkup.
-    pub fn add_diagnose<D: Diagnose + 'static>(&mut self) -> &mut Self {
-        self.diagnose_fns.push(|patient| {
-            Box::pin(async {
-                let diags = D::diagnose(patient).await?;
-                Ok(diags.into_iter().map(|diag| Box::new(diag) as _).collect())
-            })
-        });
+    pub fn add_diagnostic<D: Diagnostic + Default + 'static>(&mut self) -> &mut Self {
+        self.diagnostics.push(Box::<D>::default());
         self
     }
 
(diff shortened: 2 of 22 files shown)


crates/doctor/src/manifest/trigger.rs
```diff
@@ -0,0 +1,236 @@
...
+        let Some(trigger_type) = config.get("type") else {
+            return Some(Self::InvalidAppTrigger("trigger table missing type"));
+        };
+        let Some(trigger_type) = trigger_type.as_str() else {
+            return Some(Self::InvalidAppTrigger("type must be a string"));
+        };
+        if trigger_type == "http" && config.get("base").is_none() {
+            return Some(Self::HttpAppTriggerMissingBase);
+        }
+        None
```
calebschoepp, line 42: If a manifest has multiple of these issues won't this implementation require them to keep running `spin doctor` until they have it fixed correctly?

calebschoepp (commented):
Couple questions but otherwise very excited about this, LGTM

src/commands/doctor.rs
```diff
@@ -0,0 +1,96 @@
...
+        eprintln!("{icon}The Spin Doctor is in.", icon = Emoji("📟 ", ""));
+        eprintln!("{icon}Checking {:?}...", self.file, icon = Emoji("🩺 ", ""));
+
+        let count = spin_doctor::Checkup::new(self.file)
+            .for_each_diagnosis(move |diagnosis, patient| {
+                async move {
+                    show_diagnosis(&*diagnosis);
+
+                    if let Some(treatment) = diagnosis.treatment() {
+                        let desc = match treatment.description(patient).await {
```
itowlson, line 27: Under what circumstances would it be reasonable to error creating the _description_?  It seems like the previous line has already determined the treatment, so the error "couldn't prepare treatment" seems odd... I would kind of expect if we know the treatment then describing it should be infallible...?

itowlson (commented):
A number of nits and suggestions, most if not all of which can be deferred to future PRs; really good stuff and I really appreciated the clean structure of it.

src/commands/build.rs
```diff
@@ -20,8 +20,12 @@ pub struct BuildCommand {
     )]
     pub app: Option<PathBuf>,
 
+    /// Component ID to build. The default is all components.
+    #[clap(short = 'c', long)]
+    pub component_id: Option<String>,
+
     /// Run the application after building.
-    #[clap(name = BUILD_UP_OPT, short = 'u', long = "up")]
+    #[clap(name = BUILD_UP_OPT, short = 'u', long)]
```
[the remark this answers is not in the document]
lann: Oops; incomplete feature. The new flag is meant to scope the "missing wasm source" treatment to just the component in question.

crates/doctor/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// Return a new checkup for the app manifest at the given path.
+    pub fn new(manifest_path: impl Into<PathBuf>) -> Self {
+        let mut config = Self {
+            manifest_path: manifest_path.into(),
+            diagnose_fns: vec![],
+        };
+        config.add_diagnose::<manifest::version::VersionDiagnosis>();
+        config.add_diagnose::<manifest::trigger::TriggerDiagnosis>();
+        config.add_diagnose::<wasm::missing::WasmMissing>();
+        config
```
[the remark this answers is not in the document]
lann: Yes. We may want to move these into `Checkup::add_builtin_diagnoses` or something to allow running a subset but there will be a list somewhere in here.

crates/doctor/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+/// - the `SPIN_BIN_PATH` environment variable
+/// - the current executable ([`std::env::current_exe`])
+/// - the constant `"spin"` (resolved by e.g. `$PATH`)
+pub fn spin_command() -> Command {
+    let spin_path = std::env::var_os(SPIN_BIN_PATH)
+        .map(PathBuf::from)
+        .or_else(|| std::env::current_exe().ok())
+        .unwrap_or("spin".into());
+    Command::new(spin_path)
+}
```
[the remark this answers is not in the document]
lann: #965...

crates/doctor/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+
+/// The Diagnose trait implements the detection of a particular Spin app problem.
+#[async_trait]
+pub trait Diagnose: Diagnosis + Send + Sized + 'static {
+    /// Check the given [`Patient`], returning any problem(s) found.
+    async fn diagnose(patient: &PatientApp) -> Result<Vec<Self>>;
+}
+
+fn diagnose_boxed<D: Diagnose>(patient: &PatientApp) -> DiagnoseFut {
+    Box::pin(async {
```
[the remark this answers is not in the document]
lann: I started off with something very similar, but it adds a bit of boilerplate to every check:

```rust
#[derive(Defaiult)]
pub struct DiagnoseTheProblem;

#[async_trait]
impl Diagnose for DiagnoseTheProblem {
  type Diagnosis = TheProblem;

  async fn diagnose(&self, ... {
    ...
  }
}

pub struct TheProblem { ... }

impl Diagnosis for TheProblem {
  ...
}
```

src/commands/doctor.rs
```diff
@@ -0,0 +1,96 @@
...
+#[clap(hide = true, about = "Run the Spin Doctor")]
+pub struct DoctorCommand {
+    #[clap(short = 'f', long, default_value = "spin.toml")]
+    file: PathBuf,
+}
+
+impl DoctorCommand {
+    pub async fn run(self) -> Result<()> {
+        eprintln!("{icon}The Spin Doctor is in.", icon = Emoji("📟 ", ""));
+        eprintln!("{icon}Checking {:?}...", self.file, icon = Emoji("🩺 ", ""));
```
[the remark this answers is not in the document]
lann: I default to `Debug` for string-like things in general because it can reveal surprising characters to aid debugging. I agree that for this particular case `display` probably makes more sense.

src/commands/doctor.rs
```diff
@@ -0,0 +1,96 @@
...
+        Emoji("❗ ", "")
+    } else {
+        Emoji("⚠️ ", "")
+    };
+    eprintln!("\n{icon}Diagnosis: {}", diagnosis.description());
+}
+
+fn prompt_treatment(desc: String) -> Result<bool> {
+    let prompt = format!(
+        "{icon}Treatment available! Would you like to apply it?",
```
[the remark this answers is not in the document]
lann: :+1: I like this idea. It does complicate things if we wanted a `spin doctor --just-fix-everything` flag, but we could document that e.g. "the first Treatment will be the default".

crates/doctor/src/manifest/version.rs
```diff
@@ -0,0 +1,122 @@
...
+    spin_manifest_version: Option<Value>,
+    spin_version: Option<Value>,
+}
+
+#[async_trait]
+impl Diagnose for VersionDiagnosis {
+    async fn diagnose(patient: &PatientApp) -> Result<Vec<Self>> {
+        let doc = &patient.manifest_doc;
+        let test: VersionProbe =
+            from_document(doc.clone()).context("failed to decode VersionTest")?;
```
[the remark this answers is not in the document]
lann: Renamed `VersionTest` -> `VersionProbe` at some point. This is really meant to be an assertion, not a user-facing error; I would have made it an an `expect` if I was a little more confident in its infallibility.

crates/doctor/src/lib.rs
```diff
@@ -0,0 +1,163 @@
+//! Spin doctor: check and automatically fix problems with Spin apps.
+#![deny(missing_docs)]
+
+use std::{fmt::Debug, fs, future::Future, path::PathBuf, pin::Pin, process::Command, sync::Arc};
+
+use anyhow::{ensure, Context, Result};
```
rylev, line 6: nit: I know we do this everywhere, but I *really* prefer not using `Result` unqualified. Unqualified `Result` should always be `std::result::Result`. Everywhere else we should use it qualified (e.g., `anyhow::Result`, `io::Result`, etc.). You don't need to address that here though since this is far from the only place where this occurs.

crates/doctor/src/lib.rs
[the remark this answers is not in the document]
rylev: I agree with @itowlson. It seems to me that the boiler plate might be worth it.

crates/doctor/src/manifest/trigger.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// Invalid HTTP component trigger config
+    InvalidHttpComponentTrigger(String, &'static str),
+}
+
+impl TriggerDiagnosis {
+    fn diagnose_app_trigger(config: Option<&Value>) -> Option<Self> {
+        let Some(config) = config else {
+            return Some(Self::MissingAppTrigger);
+        };
+        let Some(config) = config.as_table() else {
```
rylev, line 30: I'm a bit worried that we're going to have to duplicate a lot of logic here and with the manifest parser...

src/commands/doctor.rs
```diff
@@ -0,0 +1,96 @@
...
+#[derive(Parser, Debug)]
+#[clap(hide = true, about = "Run the Spin Doctor")]
+pub struct DoctorCommand {
+    #[clap(short = 'f', long, default_value = "spin.toml")]
+    file: PathBuf,
+}
+
+impl DoctorCommand {
+    pub async fn run(self) -> Result<()> {
+        eprintln!("{icon}The Spin Doctor is in.", icon = Emoji("📟 ", ""));
```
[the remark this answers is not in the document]
rylev: To throw my opinion in the ring: I quite like emoji because I do find that it helps guide my eye, but more than anything I would prefer *color* output. I find spin output generally very hard to read because it all comes out looking like a wall of text.

src/commands/doctor.rs
```diff
@@ -0,0 +1,96 @@
...
+                    if let Some(treatment) = diagnosis.treatment() {
+                        let desc = match treatment.description(patient).await {
+                            Ok(desc) => desc,
+                            Err(err) => {
+                                show_error("Couldn't prepare treatment: ", err);
+                                return Ok(());
+                            }
+                        };
+
+                        let should_treat = prompt_treatment(desc).unwrap_or_else(|err| {
```
rylev, line 35: Many tools like `rustfmt` never diffs when the working tree is dirty. Perhaps we should also be cautious on behalf of the user and let them know that they might want to commit their changes first.

crates/doctor/src/manifest/version.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    fn treatment(&self) -> Option<&dyn Treatment> {
+        Some(self)
+    }
+}
+
+#[async_trait]
+impl ManifestTreatment for VersionDiagnosis {
+    async fn treat_manifest(&self, doc: &mut Document) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
lann: Hmm... I was thinking the issues should be separate so that a user can opt-in to the separate treatments, but now that I think of it the treatment fixes both problems unconditionally anyway... I'll change it one way or another.

crates/doctor/src/wasm/missing.rs
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+        Ok(vec![])
+    }
+}
+
+impl Diagnosis for WasmMissing {
+    fn description(&self) -> String {
+        let id = self.0.component_id();
+        let  WasmSource::Local(path) = self.0.source() else {
```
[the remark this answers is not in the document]
lann: Seems to actually be a `rustfmt` bug/oversight. If I add the same space to the line above it gets removed...

crates/doctor/src/lib.rs
lann: While I understand the rationale, I don't think it's necessary for `anyhow::Result` because 1) it is ubiquitous across the codebase (for better or worse) and 2) its definition makes it compatible with `std::result::Result`:

```rust
pub type Result<T, E = Error> = core::result::Result<T, E>;
```

crates/doctor/src/manifest/trigger.rs
lann: I think its inevitable for this kind of check unless we want to break the manifest parser up into pretty fine-grained stages with different levels of validation.

src/commands/doctor.rs
lann: I'm not sure I follow what you're saying here. Could you give an example?

rylev: If the git working tree is dirty (i.e., `git status` shows changes), then `cargo clippy --fix` and `rustfmt` will refuse to modify code as those modifications might corrupt local changes. Instead those tools actively encourage users to commit their changes first before running that tool (with a way for the user to opt into running the tool anyway with something like `--allow-dirty`.

crates/doctor/src/lib.rs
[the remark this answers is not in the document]
lann: Need to double-check a few things, but this seems to be working: https://github.com/fermyon/spin/commit/31a6b079e46d64deb02b93429eab21e518b14975

I'm ambivalent about how much of an improvement it is, but I can't argue that async generics are painful.

crates/doctor/src/manifest/trigger.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// HTTP trigger missing base field
+    HttpAppTriggerMissingBase,
+    /// HTTP component trigger missing route field
+    HttpComponentTriggerMissingRoute(String, bool),
+    /// Invalid HTTP component trigger config
+    InvalidHttpComponentTrigger(String, &'static str),
+}
+
+impl TriggerDiagnosis {
+    fn diagnose_app_trigger(config: Option<&Value>) -> Option<Self> {
```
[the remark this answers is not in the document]
itowlson: I prefer the term Config Curmudgeon.  Known to many as the Config Crank.

src/commands/doctor.rs
itowlson: At the risk of derailing, both `rustfmt` and `cargo fmt` modify working trees with uncommitted changes for me @rylev - am I misunderstanding the condition you're talking about here?

src/commands/doctor.rs
itowlson: But wouldn't (shouldn't?) that fail at the `.treatment()` stage?  It's the fact that we _have_ a treatment but barf when trying to _describe_ it that I find puzzling.  Like I'd expect the treatment to be kind of determined at this point, and describing it to be entirely deterministic.

Probably I need to go back and read how treatments work again...

lann: You would have to build the `Treatment::description` all the way back at `diagnose`, _or_ keep a copy of enough patient information in the `Diagnosis` to build it during `treatment` (surely a health record privacy violation). Either way your description could be out of date by the time you're ready to apply the treatment because - since `diagnose` can return multiple issues - other treatments may have already run in the interim.

lann: Thinking "out loud" (mechanical keyboard), it might make sense to parse the `manifest_doc` lazily for each diagnosis anyway to avoid issues with the file being modified in the middle of an interactive checkup...

itowlson: > You would have to build the Treatment::description all the way back at diagnose

Could you expand on why?  Is the idea that `diagnosis.treatment()` is a simple getter (the `Diagnosis` includes the treatment)?  This seems to run counter to your observation that "other treatments may have already run in the interim."  I assumed `treatment()` calculated the treatment lazily, and so could bundle the description into the Treatment object.  But I have not yet got round to that re-read so...

Sorry, I realise I am kind of obsessing over one not very crucial line; but I think it is hinting at a deeper misunderstanding (on my part) of how this stuff hangs together.  Feel free to drop this conversation if it's not useful to you.

crates/doctor/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+serde = { version = "1", features = ["derive"] }
+similar = "2"
+spin-loader = { path = "../loader" }
+tokio = "1"
+toml = "0.7"
+toml_edit = "0.19"
+tracing = { workspace = true }
+
+[dev-dependencies]
+tempfile = "3"
```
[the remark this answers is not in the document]
calebschoepp: Agreed

crates/doctor/src/lib.rs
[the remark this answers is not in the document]
calebschoepp: I like the idea of `Checkup::add_builtin_diagnoses`. Doesn't have to happen in this PR though.

crates/doctor/src/manifest/trigger.rs
calebschoepp: Got it, thanks.

src/commands/doctor.rs
```diff
@@ -0,0 +1,96 @@
...
+use clap::Parser;
+use dialoguer::{console::Emoji, Select};
+use futures::FutureExt;
+use spin_doctor::Diagnosis;
+
+#[derive(Parser, Debug)]
+#[clap(hide = true, about = "Run the Spin Doctor")]
+pub struct DoctorCommand {
+    #[clap(short = 'f', long, default_value = "spin.toml")]
+    file: PathBuf,
```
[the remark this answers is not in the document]
calebschoepp: Ooh I really like this

crates/doctor/src/lib.rs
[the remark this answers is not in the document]
calebschoepp: FINE I'LL RESOLVE THIS ISSUE JEEZ 😄

src/commands/doctor.rs
lann: I think the issue is that the `manfiest` checks effectively do a dry-run of the treatment and use the resulting diff as the result of this function, and `description` just isn't a good name for that. I'll take another crack at this, probably splitting this into two methods where one does the dry-run "diff" and the other can be used for the treatment choice menu as you suggested in the other comment.

src/commands/doctor.rs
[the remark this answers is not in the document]
lann:
```console
❗ Diagnosis: http trigger config missing base
🩹 The Spin Doctor can help! Would you like to:
  Set default HTTP trigger base '/'
  Do nothing
> See more details about the recommended changes
```
```console
🩹 The Spin Doctor can help! Would you like to: See more details about the recommended changes

📋 Apply the following diff to "examples/http-rust/spin.toml":
@@ -4,3 +4,3 @@
 name = "spin-hello-world"
-trigger = { type = "http" }
+trigger = { type = "http", base = "/" }
 version = "1.0.0"

Would you like to apply this fix? [Y/n] 
```

crates/common/src/spin_env.rs
```diff
@@ -0,0 +1,19 @@
+//! Spin environment
+
+use std::path::PathBuf;
+
+/// Return a path to the `spin` binary; the first of these that is available:
+/// - the `SPIN_BIN_PATH` environment variable
+/// - the current executable ([`std::env::current_exe`])
+/// - the constant `"spin"` (resolved by e.g. `$PATH`)
+pub fn spin_bin_path() -> PathBuf {
+    if let Some(path) = std::env::var_os("SPIN_BIN_PATH") {
```
itowlson, line 10: This environment variable will be relevant to plugins to "get back to" the Spin binary that launched them, but should probably be ignored within Spin itself.  There's bountiful potential for confusion and hard-to-debug inconsistencies if `spin doctor` or `spin up` were to run a _different version_ of Spin without the user realising it.

src/commands/doctor.rs
```diff
@@ -0,0 +1,115 @@
...
+                        let should_treat = prompt_treatment(treatment.summary(), dry_run)
+                            .unwrap_or_else(|err| {
+                                show_error("Prompt error: ", err);
+                                false
+                            });
+
+                        if should_treat {
+                            match treatment.treat(patient).await {
+                                Ok(()) => {
+                                    println!("{icon}Treatment applied!", icon = Emoji("❤️ ", ""));
```
itowlson, line 50: Something goes slightly awry in the alignment of this line.  It's as if the space is being ignored, even though it's definitely there!

image

src/commands/doctor.rs
```diff
@@ -0,0 +1,115 @@
...
+use clap::Parser;
+use dialoguer::{console::Emoji, Confirm, Select};
+use futures::FutureExt;
+use spin_doctor::{Diagnosis, DryRunNotSupported};
+
+#[derive(Parser, Debug)]
+#[clap(hide = true, about = "Detect and fix problems with Spin applications")]
+pub struct DoctorCommand {
+    #[clap(short = 'f', long, default_value = "spin.toml")]
+    file: PathBuf,
```
itowlson, line 13: Petition to allow a directory here as we do in `spin up`.  (I know this is not consistent through `build` and `deploy` yet, so no biggie if you prefer to punt on it - I can do a "-f dir everywhere" pass later.  ~Or I could do it now and make YOU have the merge conflicts.~  uh not that I would ever do a thing like that)

crates/common/src/spin_env.rs
lann: Hmm...I was specifically thinking of the case where a plugin might want to shell out to `spin doctor` here. Someone accidentally setting `SPIN_BIN_PATH` to something surprising seems like a pretty unlikely scenario to me.

src/commands/doctor.rs
lann: Hmm yes I can reproduce this problem with certain glyphs in some terminals (fonts?). It seems like certain double-width emoji are treated as single-width...

lann: Seems to be an issue with certain unicode emoji that are encoded as a "legacy" single-codepoint graphical character plus the "emoji presentation selector". Oh unicode :roll_eyes:

lann: I think I have this fixed now, though we should keep this in mind when we decide on specific emoji / whether to have them at all.

src/commands/doctor.rs
lann: :+1: Yeah lets tackle this in a separate PR. This is already turning into a giga-PR.

crates/common/src/spin_env.rs
calebschoepp: Possibly silly idea but we could also just print out a warning if `SPIN_BIN_PATH` is ever set when just directly running `spin`. That would prevent this edge case.

src/commands/doctor.rs
calebschoepp: For future reference I don't think `spin watch` supports directories for `-f`. I think this flag/functionality is perfect for the common crate.

lann: > perfect for the common crate

Yes, except that I believe all the users would be in the root `spin-cli` crate, in which case it should just go there :slightly_smiling_face:

calebschoepp: Not if/when we move `spin watch` to its own crate.

itowlson: Yeah, this is one of those things that should be resolved at the UI level, and everything below there should just not have to think about it.  I'll pick this up (and no worries @calebschoepp `spin watch` is ~in my sights~ uh ~on my radar~ uh appended to my to-do list).

crates/common/src/spin_env.rs
itowlson: But in the case where a plugin is shelling out to `spin doctor`, we are already inside a Spin process by the time this function gets called, so `current_exe` will report the right thing, won't it?

(Also, at the risk of derailing the serious question above, here's an "Edgiest of the Edge" case: Consider a plugin that intentionally shells out to `spin doctor` or `spin up` using a copy of Spin that was _not_ the one that called it for uh reasons.  Now suppose that `spin doctor` calls `spin build`. or `spin up` calls `spin trigger`.  With this override in place, those double-shells are going to call to the _original_ copy of Spin, not the copy that the plugin asked for.)

lann: > But in the case where a plugin is shelling out to spin doctor, we are already inside a Spin process by the time this function gets called, so current_exe will report the right thing, won't it?

Sorry, I wrote that wrong; I'm thinking of a plugin using the `spin-doctor` crate directly, then a spin doctor treatment shelling out to e.g. `spin build`.

itowlson: In that case maybe have the `spin-doctor` crate export a function explicitly for plugins, which could simply fail if (somehow) `SPIN_BIN_PATH` were not set rather than falling back to `current_exe` - then the plugin would need to link only to `spin-doctor` rather than also bringing in `spin-common` (which is not really what we want to present as The Face Of Spin).

I think what I'm trying to get at (admittedly not very clearly) is that plugins and Spin itself have entirely different strategies for locating the Spin binary, and there is no need or benefit for one function to try to serve both those purposes.

itowlson: Thinking about this and manifest resolution and no doubt other things, seems like a "plugin utils" library (to help plugins behave consistently with Spin) might be in our futures, though given that plugins are intentionally not bound to Rust, it could end up becoming a labour to ship across multiple languages.

lann: Removed.

src/commands/doctor.rs
lann: Given that this only has an interactive UI now I think this can wait until/if we add a non-interactive mode.

PR #1442, opened 2023-05-02
Add `spin up --key-value` option

### LINKED ISSUE #1404: Add support for `spin up --key-value KEY,VALUE`
It would be nice to have parity between `spin up` and `spin cloud deploy` for the pre-seed key value feature. Today only `spin cloud deploy --key-value KEY,VALUE` works. This issue will track the work to add support for `spin up --key-value KEY,VALUE`. That way, Spin & Fermyon Cloud users who have a dependency on this feature can move their applications between environments easily.

itowlson:
Fixes #1404.

Note this currently assumes the answer to my questions there is "default store suffices."  The core logic is easy enough to extend to support non-default stores; the difficulty is only in the CLI syntax.

This PR contains (in `key-value/src/host_component.rs`) the greatest Rust crime I have ever committed.  @dicej it would be great to get your thoughts on the right way to do this.  The trouble I kept running into was that `StoreManagerManager` would only dispense a Store if you gave it an AppComponent.  My guess is I need to add a `get_for_app` method to `StoreManagerManager` but then we will not longer be able to implement it via a function...

Also @lann does the proposed entry point for carrying out before-run host component initialisation seem reasonable?  I was wondering about using `builder.hooks` to set an `app_loaded` hook.  The place where hooks are currently set didn't seem to have access to the host component itself, but would it be reasonable to set a hook where we instantiate and add the key/value DynamicHostComponent (`TriggerExecutorBuilder::build`)?

Or are these questions two sides of the same coin, and if I set the right hook in the right place then I wouldn't need to go through `StoreManagerManager` at all?

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -9,6 +9,7 @@ use tokio::{
 };
 
 use spin_app::Loader;
+use spin_common::arg_parser::parse_kv;
 
 use crate::stdio::StdioLoggingTriggerHooks;
 use crate::{
@@ -104,6 +105,12 @@ where
     #[clap(flatten)]
     pub run_config: Executor::RunConfig,
 
+    /// Set a key/value pair (key=value) in the application's
+    /// default store. Any existing value will be overwritten.
+    /// Can be used multiple times.
+    #[clap(long = "key-value", parse(try_from_str = parse_kv))]
+    key_values: Vec<(String, String)>,
+
     #[clap(long = "help-args-only", hide = true)]
     pub help_args_only: bool,
 }
@@ -132,8 +139,12 @@ where
         let working_dir = std::env::var(SPIN_WORKING_DIR).context(SPIN_WORKING_DIR)?;
         let locked_url = std::env::var(SPIN_LOCKED_URL).context(SPIN_LOCKED_URL)?;
 
+        let init_data = crate::HostComponentInitData {
+            kv: self.key_values.clone(),
+        };
+
         let loader = TriggerLoader::new(working_dir, self.allow_transient_write);
-        let executor = self.build_executor(loader, locked_url).await?;
+        let executor = self.build_executor(loader, locked_url, init_data).await?;
 
         let run_fut = executor.run(self.run_config);
 
@@ -159,6 +170,7 @@ where
         &self,
         loader: impl Loader + Send + Sync + 'static,
         locked_url: String,
+        init_data: crate::HostComponentInitData,
     ) -> Result<Executor> {
         let runtime_config = self.build_runtime_config()?;
 
@@ -170,7 +182,7 @@ where
         builder.hooks(StdioLoggingTriggerHooks::new(self.follow_components()));
         builder.hooks(KeyValuePersistenceMessageHook);
 
-        builder.build(locked_url, runtime_config).await
+        builder.build(locked_url, runtime_config, init_data).await
     }
 
     fn build_runtime_config(&self) -> Result<RuntimeConfig> {
--- a/crates/trigger/src/lib.rs
+++ b/crates/trigger/src/lib.rs
@@ -102,6 +102,7 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
         mut self,
         app_uri: String,
         runtime_config: runtime_config::RuntimeConfig,
+        init_data: HostComponentInitData,
     ) -> Result<Executor>
     where
         Executor::TriggerConfig: DeserializeOwned,
@@ -115,7 +116,11 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
                 builder.add_host_component(outbound_mysql::OutboundMysql::default())?;
                 self.loader.add_dynamic_host_component(
                     &mut builder,
-                    runtime_config::key_value::build_key_value_component(&runtime_config)?,
+                    runtime_config::key_value::build_key_value_component(
+                        &runtime_config,
+                        &init_data.kv,
+                    )
+                    .await?,
                 )?;
                 self.loader.add_dynamic_host_component(
                     &mut builder,
@@ -142,6 +147,12 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
         // Run trigger executor
         Executor::new(TriggerAppEngine::new(engine, app_name, app, self.hooks).await?).await
     }
+}
+
+/// Initialisation data for host components.
+#[derive(Default)] // TODO: this is only for tests, would like to get rid of
+pub struct HostComponentInitData {
+    kv: Vec<(String, String)>,
 }
 
 /// Execution context for a TriggerExecutor executing a particular App.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4976,6 +4976,7 @@ dependencies = [
  "spin-app",
  "spin-bindle",
  "spin-build",
+ "spin-common",
  "spin-config",
  "spin-key-value",
  "spin-key-value-sqlite",
@@ -5005,6 +5006,7 @@ dependencies = [
 name = "spin-common"
 version = "1.2.0-pre0"
 dependencies = [
+ "anyhow",
  "sha2 0.10.6",
  "tempfile",
 ]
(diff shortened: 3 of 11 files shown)


itowlson:
Before I added the `is_empty` check, this made a couple of tests (spin-trigger-http `tests::test_spin_httptests::test_{spin|wagi}_http` and spin-redis-engine `tests::test_pubsub`) fail.  The problem was that the SQLite provider's `StoreManager::get(name)` calls `tokio::task::block_in_place`, which is forbidden in the Tokio test runtime.  This occurs because they instantiate and invoke trigger objects directly in the unit tests, not hosted in `main` (which uses the multi-threaded runtime and is perfectly happy).  This makes me worry that there is some hidden brittleness here, but I am not sure if it is in the way the initialisation code uses the StoreManager, or in the way the tests load the trigger.

dicej:
FYI, you can annotate Tokio tests using `#[tokio::test(flavor = "multi_thread")]`, which should allow `task::block_in_place` to work.

itowlson:
This has been manually tested, but doesn't have automated tests - I'm not sure how best to go about those.

dicej (approved):
LGTM.  Regarding testing: perhaps we could add a variation on the `key_value_works` test in `tests/spinup_tests.rs`, `tests/testcases/mod.rs`, and `tests/testcases/key-value`.

itowlson:
#1448 is thwarting my ability to implement the suggested test.  I am thinking I will merge this as-is and open a new issue to get it under e2e test coverage.

crates/key-value/src/host_component.rs
```diff
@@ -1,16 +1,42 @@
...
+use spin_app::{async_trait, AppComponent, DynamicHostComponent};
 use spin_core::HostComponent;
 use std::sync::Arc;
 
 pub trait StoreManagerManager: Sync + Send {
     fn get(&self, component: &AppComponent) -> Arc<dyn StoreManager>;
+    fn get_for_init(&self) -> Option<Arc<dyn StoreManager>>; // Some stores might not support init
+}
+
+pub struct SingletonStoreManagerManager {
```
lann, line 12: :scream:

crates/trigger/src/lib.rs
```diff
@@ -113,18 +114,27 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
...
-                )?;
-                self.loader.add_dynamic_host_component(
-                    &mut builder,
-                    spin_config::ConfigHostComponent::new(runtime_config.config_providers()),
-                )?;
+                self.loader
+                    .add_dynamic_host_component(
+                        &mut builder,
+                        runtime_config::key_value::build_key_value_component(&runtime_config)?,
+                        init_data.kv,
```
lann, line 121: Unless I'm missing something you should be able to pass this init data into the function that creates the component here:

```suggestion
                        runtime_config::key_value::build_key_value_component(&runtime_config, init_data.kv)?,
```

and then in `build_key_value_component` do something like:

```rust
pub fn build_key_value_component(runtime_config: &RuntimeConfig, initial_values: Vec<(String, String)>) -> Result<KeyValueComponent> {
    let mut stores = runtime_config
        .key_value_stores()
        .context("Failed to build key-value component")?
        .collect::<HashMap<_, _>>();
    if !initial_values.is_empty() {
        if let Some(manager) = stores.get("default") {
            let store = manager.get("default").unwrap();
            for (key, value) in initial_values {
                store.set(&key, value.as_bytes())?;
            }
        } else {
            // bail? log warning?
        }
    }
    ...
```

lann (changes requested):
I don't think the `DynamicHostComponent::initialize` method is necessary.

crates/trigger/src/lib.rs
itowlson: I was kind of thinking it would be nice to have the initialisation login in the component itself rather than as a side effect in a function in the `runtime_config` module.  But maybe I should get over myself!

lann, line 121: > I was kind of thinking it would be nice to have the initialisation login in the component itself

Sure!

```suggestion
                let component = runtime_config::key_value::build_key_value_component(&runtime_config)?;
                component.set_initial_values(init_data.kv)?;
                self.loader
                    .add_dynamic_host_component(
                        &mut builder,
                        component
```

```rust
impl KeyValueComponent {
  pub fn set_intitial_values ...
}
```

lann: https://github.com/fermyon/spin/pull/1442/files/09bedc013db837176e3b63631b5e21f6a1688524#r1185541401

lann: ...or some variation thereof... maybe the `set_initial_values` call would be better off inside `build_key_value_component` :shrug:

crates/app/src/lib.rs
```diff
@@ -79,7 +79,7 @@ impl AppLoader {
     ///
     /// This calls [`EngineBuilder::add_host_component`] for you; it should not
     /// be called separately.
-    pub fn add_dynamic_host_component<T: Send + Sync, DHC: DynamicHostComponent>(
+    pub async fn add_dynamic_host_component<T: Send + Sync, DHC: DynamicHostComponent>(
```
lann, line 82: I think this can be removed now?

crates/trigger/src/cli.rs
```diff
@@ -243,6 +254,16 @@ async fn warn_slow_wasm_build() {
     println!();
 }
 
+// Parse the key/values passed in as `key=value` pairs.
+// TODO: `spin-utils` WHEN
```
lann, line 258: :eyes:

crates/trigger/src/cli.rs
```diff
@@ -243,6 +254,16 @@ async fn warn_slow_wasm_build() {
...
+fn parse_kv(s: &str) -> Result<(String, String)> {
+    let parts: Vec<_> = s.splitn(2, '=').collect();
+    if parts.len() != 2 {
+        bail!("Key/Values must be of the form `key=value`");
+    }
+    Ok((parts[0].to_owned(), parts[1].to_owned()))
```
lann, line 264: Slightly nicer:
```suggestion
    let Some((key, value)) = s.split_once('=') else {
        bail!("Key/Values must be of the form `key=value`");
    }
    Ok((key.to_owned(), value.to_owned()))
```

crates/app/src/host_component.rs
```diff
@@ -8,6 +8,7 @@ use crate::AppComponent;
 ///
 /// This extends [`HostComponent`] to support per-[`AppComponent`] dynamic
 /// runtime configuration.
+#[async_trait::async_trait]
```
lann, line 11: Remove?

crates/key-value/src/host_component.rs
```diff
@@ -54,6 +54,7 @@ impl HostComponent for KeyValueComponent {
     }
 }
 
+#[async_trait]
```
lann, line 57: Remove?

crates/app/src/lib.rs
itowlson: Aargh, good spot, thanks!

crates/trigger/src/cli.rs
itowlson: Not only nicer, but more correct (if a value contains an `=` sign e.g. base64).  But I wanted to keep it exactly the same as the `deploy` one (and didn't want to change `deploy`) so that when spin-utils came along we wouldn't need to eyeball them for subtle differences.

But yeah now I can (rebase and) move to common.

lann: I think `splitn(..., 2)` makes it equivalent

itowlson: And you are, of course, correct.  Can I start today again please, this time with more tea...

lann: > move to common

Ooo yeah we're probably duplicating a few of these arg parsers.

itowlson: How similar this is to the environment variable parser will shock you.

lann (approved):

itowlson:
@Mossaka `HostComponentInitData` is for real, but the `#[derive(Default)]` macro is only there to support the testing infrastructure.  There's no meaningful default for `HostComponentInitData`, it _has_ to be determined by the CLI; but there's a bunch of testing infra that relies on being able to default things.  I'll clarify the comment - thanks for flagging this.

PR #1444, opened 2023-05-02
Add `spin-common` crate

### LINKED ISSUE #965: Define a strategy for shared utility code

lann:
Initially with just the spin_common::sha256 module, which replaces several implementations of "hash something to hex".

Closes #965


crates/common/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+//! Spin common modules
```
itowlson, line 1: Does this need a mandatory docs thingy, or are we not too concerned because these things are "logically internal", or does it inherit one from somewhere else?

itowlson (approved):

PR #1450, opened 2023-05-04
add (hidden) `spin crater` subcommand

dicej:
This command is useful for testing Spin changes (e.g. the recent componentization work) against a large set of apps stored in a Bindle repository.  The "test" in this case is simply pre-instantiating each component of each app, which serves to verify that `spin-componentize` can generate a valid Component Model component from each model and that linking succeeds.  We don't actually instantiate and call into those components since they may do arbitrary things at runtime, and there's no way to verify an arbitrary app "behaves correctly" given that each app will have its own notion of "correctness".


src/commands/up.rs
```diff
@@ -434,6 +416,23 @@ impl AppSource {
...
+    working_dir: &Path,
+) -> Result<String, anyhow::Error> {
+    let locked_path = working_dir.join("spin.lock");
+    let locked_app_contents =
+        serde_json::to_vec_pretty(&locked_app).context("failed to serialize locked app")?;
+    tokio::fs::write(&locked_path, locked_app_contents)
+        .await
+        .with_context(|| format!("failed to write {:?}", locked_path))?;
+    let locked_url = Url::from_file_path(&locked_path)
+        .map_err(|_| anyhow!("cannot convert to file URL: {locked_path:?}"))?
```
rylev, line 430:
```suggestion
        .map_err(|_| anyhow!("cannot convert locked app path to file URL: {locked_path:?}"))?
```

src/commands/crater.rs
```diff
@@ -0,0 +1,187 @@
...
+                let path = bindle_path
+                    .join("parcels")
+                    .join(&raw.source)
+                    .join("parcel.dat");
+
+                Ok::<_, Error>(CoreComponent {
+                    source: ModuleSource::Buffer(
+                        tokio::fs::read(&path)
+                            .await
+                            .with_context(|| path.display().to_string())?,
```
rylev, line 160:
```suggestion
                            .with_context(|| format!("failed to read parcel data at '{}", 'path.display()))?,
```

src/commands/crater.rs
```diff
@@ -0,0 +1,187 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+async fn pre_instantiate(bindle_path: &Path, invoice: &Invoice) -> Result<()> {
+    let working_dir = tempfile::tempdir()?;
+    let working_dir = working_dir.path().canonicalize()?;
```
rylev, line 112: We should probably add some context to these errors.

src/commands/crater.rs
```diff
@@ -0,0 +1,187 @@
...
+            })
+            .buffer_unordered(MAX_CONCURRENCY);
+
+        // Note that these will print somewhat out-of-order with respect to `index` due to the `buffer_unordered`
+        // call above.  If that becomes annoying, we can either stop using `buffer_unordered` or accumulate the
+        // results and sort them before printing.
+        while let Some((index, result, id)) = stream.next().await {
+            print!("({index:>5} of {total}) app {id}: ");
+            if let Err(e) = result {
+                println!("failed: {e:?}");
```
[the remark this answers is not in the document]
dicej: The reason I'm using `stdout` here is that it's reporting a failure in one of the apps being tested rather than a failure of `spin-crater` itself.  The idea is that we can run `spin crater some_directory > results.txt` and only write to `stderr` if `spin-crater` itself has an issue which prevents it from running normally (e.g. if the specified directory doesn't exist).

itowlson:
Is the intention that Spin application developers use this?  If so, maybe give it a more meaningful name?  If not, could it be a plugin?

rylev:
It would be nice if we could merge this as is so we can continue to improve it in parallel. I've had to create a PR on @dicej's fork in order to push my latest improvements: https://github.com/dicej/spin/pull/1

crates/trigger/src/lib.rs
```diff
@@ -60,6 +60,11 @@ pub trait TriggerExecutor: Sized + Send + Sync {
                 .with_context(|| format!("Failed to instantiate component '{}'", component.id()))?,
         ))
     }
+
+    /// Instantiate all components and verify that they export the expected trigger function(s).
+    async fn check(&self) -> Result<()> {
+        todo!()
```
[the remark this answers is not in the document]
dicej: Done.

dicej:
> Is the intention that Spin application developers use this? If so, maybe give it a more meaningful name? If not, could it be a plugin?

It's not intended for app developers, no.  The command itself could be packaged as a plugin, yes, as long as we make the required changes to `trigger`, `trigger-http`, and `loader`.

melissaklein24:
@dicej can you close this one?

PR #1451, opened 2023-05-04
Add 'spin_common::sloth' module

lann:
Replaces two similar patterns for printing a warning if an operation takes too long.


src/commands/deploy.rs
```diff
@@ -203,11 +204,13 @@ impl DeployCommand {
...
-        check_healthz(&login_connection.url).await?;
-        // Hippo has responded - we don't want to keep the sloth timer running.
-        drop(sloth_warning);
+        {
+            let _sloth_guard = sloth::warn_if_slothful(
+                2500,
+                format!("Checking status ({})", login_connection.url),
+            );
+            check_healthz(&login_connection.url).await?;
+        }
```
itowlson, line 213: My personal preference here is for the explicit drop, because it signals to the reader that it's intentional that the value be dropped _at this particular point._  I think I originally wrote it braced like this and switched because I was sure someone would go "these braces aren't doing anything, let's tidy them up."

itowlson (approved):
LGTM except for one suggested style tweak

src/commands/deploy.rs
lann: There seems to be a pretty strong convention to use scopes over `drop` in general, though I don't have a strong personal preference in this case.

fibonacci1729 (approved):

lann:
@itowlson I see how it is! You slowed my `sloth` (even more than its natural state!) so that your `arg_parser` conflicts could go through first!

PR #1452, opened 2023-05-04
Move Cloud commands out of Spin CLI and into a plugin

kate-goldenring:
WIP draft to move Fermyon Cloud command code (`spin cloud deploy` and `spin cloud login`) out of Spin CLI and into a plugin. 

[SIP](https://github.com/kate-goldenring/spin/blob/cloud-plugin-porting/docs/content/sips/014-cloud-plugin.md).

Cloud plugin: https://github.com/fermyon/cloud-plugin

~Do not merge until the plugin is added to the [Spin plugins registry](https://github.com/fermyon/spin-plugins).~ Added


src/bin/spin.rs
```diff
@@ -54,12 +52,7 @@ enum SpinApp {
     New(NewCommand),
     Add(AddCommand),
     Up(UpCommand),
-    #[clap(subcommand)]
-    Cloud(CloudCommands),
-    // acts as a cross-level subcommand shortcut -> `spin cloud deploy`
-    Deploy(DeployCommand),
-    // acts as a cross-level subcommand shortcut -> `spin cloud login`
-    Login(LoginCommand),
+    Cloud(CloudCommand),
```
itowlson, line 55: In the original `spin cloud` work, I recall people strongly wishing to keep the `spin deploy` and `spin login` aliases.  (Which had to be implemented as commands because clap doesn't allow "cross level" aliases.)

kate-goldenring: @itowlson I updated the implementation to be an "enlightened" plugin experience, wherein the `external` crate is now aware of the `cloud`, `login`, and `deploy` commands and looks to use the `cloud` plugin for any of them.

kate-goldenring:
@bacongobbler has been working on the SIP that will be added to this PR

src/bin/spin.rs
itowlson: Will that suffice for `login` and `deploy` to appear in help text?  I seem to recall that was asked for when the `cloud` subcommand was originally created, though my memory may be faulty of course!

kate-goldenring: No i dont believe it will. Do we want `cloud` to also appear in the help text?

kate-goldenring: This is a good call out for a section of the SIP (cc @bacongobbler)

itowlson: I believe that was the original plan (and one of the open problems from before).  The sense of the discussion was that this should all feel seamless from a user point of view, as if it was built in.  I don't know, though, whether that remains the feeling, or how strongly it's held!

kate-goldenring: I think it is a good goal. I'll try to iterate on it

bacongobbler:
> @bacongobbler has been working on the SIP that will be added to this PR

Added in 84fd741

src/bin/spin.rs
kate-goldenring: @itowlson I updated the implementation so that `cloud` `login` and `deploy` are still displayed as CLI subcommands but are executed by the plugin

itowlson: 🎉 Perfect - thanks!

docs/content/sips/013-cloud-plugin.md
```diff
@@ -0,0 +1,155 @@
...
+The proposal herein suggests that we re-release the core logic of `spin login`
+and `spin deploy` under a separate `spin cloud` plugin. Spin will alias the
+existing `spin login` and `spin deploy` commands to their respective `spin
+cloud` counterparts to retain backwards compatibility.
+
+When a user executes `spin login` or `spin deploy` and the cloud plugin is not
+installed on their system, Spin will inform the user that the `spin cloud`
+plugin has not been installed.
+
+```
```
kate-goldenring, line 79: @bacongobbler the current implementation takes this further. If the cloud plugin is not installed, it starts installing it. This is the current output:
```
The `cloud` plugin is required. Installing now.
Are you sure you want to install plugin 'cloud' with license Apache-2.0 from file:///Users/kagold/Programs/cloud-plugin/plugin/cloud.tar.gz? yes
Plugin 'cloud' was installed successfully!
```

We could even add the `--yes` flag to avoid prompting the user about the license since it is a Fermyon plugin. Thoughts?

docs/content/sips/013-cloud-plugin.md
```diff
@@ -0,0 +1,155 @@
...
+$ spin login
+Error: plugin `cloud` is not installed. Try running `spin plugin install cloud`.
+```
+
```
kate-goldenring, line 83: @bacongobbler should we add a "Distribution" section that lays out the following strategies for distributing the plugin with Spin? We could also merge the update FAQ into this too. Here is a quick take at what it could include
```suggestion
## Distribution
Rather than prompting users to install the plugin on first use of `cloud` commands, ideally the cloud plugin would be delivered with the Spin CLI package. Users should be able to get Spin through one of the following avenues:
1. The install script
1. Building from source
1. Package manager (coming soon)

The install script will be updated to automatically install the cloud plugin and packagings of Spin for package managers will be encouraged to do the same. However, in the case of local builds, upon use of the cloud plugin, users will be prompted of the need to install the cloud plugin and it will be subsequently automatically installed.

The plugin will be reinstalled to the latest compatible version with each new use of the install script or update of the package through a package manager. The plugin will also be able to be updated outside of a Spin update through `spin plugin update && spin plugin upgrade`
```

src/commands/external.rs
```diff
@@ -49,10 +50,19 @@ pub async fn execute_external_subcommand(
...
         }
         Err(Error::NotFound(e)) => {
-            tracing::debug!("Tried to resolve {plugin_name} to plugin, got {e}");
-            eprintln!("Error: '{plugin_name}' is not a known Spin command. See spin --help.\n");
-            print_similar_commands(app, &plugin_name);
-            process::exit(2);
+            if plugin_name == "cloud" {
+                println!("The `cloud` plugin is required. Installing now.");
+                let plugin_installer = Install {
+                    name: Some("cloud".to_string()),
```
kate-goldenring, line 56: To test this out now (before the cloud plugin is added to the plugins repository), replace the `name` field with  `local_manifest_src` that points to the local manifest after cloning the [`cloud-plugin` repo ](https://github.com/fermyon/cloud-plugin) 

```
local_manifest_src: Some("/Path/to/repo/cloud-plugin/plugin/cloud.json".into()),
```

docs/content/sips/013-cloud-plugin.md
radu-matei: Definitely out of scope for this PR (and perhaps something to consider at the plugin level): do we want to explore an update check mechanism that would prompt a user to consider installing a new version (and help the user install it)?

radu-matei: (missed the section below on updates)

kate-goldenring:
The `test_spin_deploy` will continue to fail till we add the plugin to the `spin-plugins` repo

docs/content/sips/013-cloud-plugin.md
bacongobbler: Ii will update the SIP to match. Thanks for catching this.

the “yes” flag is an interesting one that spans beyond the scope of this PR. Each “default” plugin could follow the same behaviour. Also by having “yes” as a flag, I don’t know how that would interact with the plugin when it is installed. Is it passed to the plug-in? Is it stripped from the CLI options before invoking the plug-in?

kate-goldenring: @bacongobbler thanks! by the `--yes` flag I mean the `spin plugins install cloud --yes` which will install the plugin without prompting about the license. I updated the implementation to use it by setting `yes_to_all: true`

kate-goldenring:
@tpmccallum added an issue to documentation on documenting this https://github.com/fermyon/developer/issues/621

docs/content/sips/013-cloud-plugin.md
bacongobbler: Ah yes that would be fine!

src/commands/external.rs
```diff
@@ -49,10 +50,20 @@ pub async fn execute_external_subcommand(
...
+                    yes_to_all: true,
+                    ..Default::default()
+                };
+                plugin_installer.run().await?;
```
bacongobbler, line 60: Do we want to attempt re-calling `spin cloud deploy` after installing the plugin? I can't tell if the newly installed plugin gets invoked after answering "yes" to install the plugin.

kate-goldenring: yes instead of returning, currently it will proceed with the command execution after install, so the flow looks like the following:

```sh
$ spin deploy
The `cloud` plugin is required. Installing now.
Plugin 'cloud' was installed successfully!
Uploading cloud_start version 0.1.0+XXXXXXXX...
Deploying...
Waiting for application to become ready... ready
Available Routes:
  cloud-start: https://cloud-start-xxxxxxxx.fermyon.app/ (wildcard)
```

kate-goldenring:
for the `cloud` plugin, @itowlson brought up the question of whether we should suppress prerelease compatibility warnings?. At the moment, using any plugin with canary or HEAD gives a "prerelease version of Spin, plugin may not be compatible." However, `cloud login` and `cloud deploy` are common paths so that is a lot of warnings and may become irksome. The main folks irked are contributors as users should be using a release.

bacongobbler:
> However, `cloud login` and `cloud deploy` are common paths so that is a lot of warnings and may become irksome.

How often is this warning displayed? If you're using a canary release, does it happen on every invocation to `spin deploy`?

Perhaps we could display a prerelease compatibility warning once per day. I don't think Spin has a date check system to warn the user, though.

IMO this is starting to feel like it falls out of scope of this PR. The system was designed to warn the user when you're using a pre-release of Spin that invokes a plugin. IMO I think this should be tackled separately from this PR, either as a follow-up discussion or a new SIP to modify the current plugin loading behavior. I can see why we may want to tackle this problem now especially because `cloud-plugin` is part of the default on-ramp experience to the Fermyon Cloud. Personally I'd prefer that we take that discussion to another thread as the decisions we make here will impact other initiatives to bring more plugins into the core CLI.

itowlson:
It's displayed every time any plugin is invoked.  I'm happy to decouple it from the `cloud` PR though - it occurred to me just because this was a blessed plugin that aimed for a relatively seamless experience, but I will open a separate issue for it.

bacongobbler:
It's good that you point out that this is just a canary/HEAD problem. Most users won't see this. I'd be fine with kicking this decision down the road.

docs/content/sips/013-cloud-plugin.md
```diff
@@ -0,0 +1,161 @@
...
+`spin login` and `spin deploy` experience can be addressed in future updates to
+the plugin.
+
+In fact, the movement to a separate `spin cloud` plugin grants us the
+flexibility to make changes to the core `spin deploy` experience without forcing
+us to wait until Spin 2.0. If anything, this proposal enables us to make these
+changes to the `spin cloud` plugin without waiting for a new release of Spin.
+
+### A generic `spin cloud` plugin supporting multiple cloud providers
+
```
kate-goldenring, line 127: @bacongobbler I believe this is not a future we are sure about. Can we remove this section on the generic plugin?

docs/content/sips/013-cloud-plugin.md
kate-goldenring: issue to track for this https://github.com/fermyon/spin/issues/1543

src/commands/external.rs
kate-goldenring: now added to the plugins repo so no need to modify

kate-goldenring:
@itowlson @fibonacci1729 @bacongobbler @radu-matei this should be ready for another review. I also added a test that ensures that the cloud plugin is installed during a `spin login` and removed all `fermyon-platform` tests.

kate-goldenring:
I am weirdly having an issue rebasing the commits to just 1. It will only go back `HEAD~2` worth of commits with `fatal: invalid upstream 'HEAD~5'` and will only show a few commits even if i point to the first commits. Maybe we can squash merge this so it isn't a bunch of commits?

src/commands/cloud.rs
```diff
@@ -1,24 +1,63 @@
...
+)]
+pub struct LoginCommand {
+    /// All args to be passed through to the plugin
+    #[clap(hide = true)]
+    args: Vec<String>,
+}
+
+#[derive(Debug, Args, PartialEq)]
+#[clap(
+    about = "Commands for publishing applications to the Fermyon Platform.",
```
itowlson, line 31:
```suggestion
    about = "Commands for publishing applications to the Fermyon Cloud.",
```

src/commands/external.rs
```diff
@@ -49,10 +50,20 @@ pub async fn execute_external_subcommand(
             }
         }
         Err(Error::NotFound(e)) => {
-            tracing::debug!("Tried to resolve {plugin_name} to plugin, got {e}");
-            terminal::error!("'{plugin_name}' is not a known Spin command. See spin --help.\n");
-            print_similar_commands(app, &plugin_name);
-            process::exit(2);
+            if plugin_name == "cloud" {
+                println!("The `cloud` plugin is required. Installing now.");
```
itowlson, line 54: Suggest `eprintln!` so that if the user is capturing plugin output for _waves hands_ reasons then the message doesn't get mixed up with it.  (E.g. imagine a hypothetical `cloud variables list --output json >vars.json`.)

Or we could define a `terminal::info!` and use that (I did one for the badger though might be too dark).

src/commands/plugins.rs
```diff
@@ -52,7 +52,7 @@ impl PluginCommands {
 }
 
 /// Install plugins from remote source
-#[derive(Parser, Debug)]
+#[derive(Parser, Debug, Default)]
```
itowlson, line 55: The default would be invalid.  If we want to avoid noise at the construction site, maybe add a `from_name` constructor rather than implying it can be defaulted?

itowlson (changes requested):
mostly nits - looks good!

itowlson:
@kate-goldenring I agree we don't need to keep the commit history but I think GH squashing is disabled because of GPG - not sure we can bypass it.

src/commands/external.rs
kate-goldenring: Good point. The "plugin installed" message still goes to stdout, though:
```
Plugin 'cloud' was installed successfully!

Description:
	The Fermyon Cloud Plugin

Homepage:
	https://github.com/fermyon/cloud-plugin
host
beep
```

src/commands/plugins.rs
kate-goldenring: Are you saying the default is invalid because there is not name? I wanted a way to construct the `Install` struct:
```
                let plugin_installer = Install {
                    name: Some("cloud".to_string()),
                    yes_to_all: true,
                    ..Default::default()
                };
```
https://github.com/fermyon/spin/blob/7f63517f2361521cc02a6273d6c1c8815393b516/src/commands/external.rs#L55

src/commands/external.rs
itowlson: Ah, no worries then; we can leave it for now and revisit both messages in a later pass.  It's not a big deal - realistically the command that triggers the install will be `spin cloud login` so who cares about what stream the output is on.

(We could also look at suppressing the description/homepage announcement for 'magic' plugins - the user has not asked for an install and doesn't really need to be told about the plugin - "I just want to log in, stop _talking_ at me" kind of feeling - but again not a blocker.)

src/commands/plugins.rs
kate-goldenring: im being picky, ill just set the rest directly to none

src/commands/external.rs
kate-goldenring: I like that idea of conditional suppression of magic plugin info

kate-goldenring:
> 

I was on a shallow clone, unshallowing it did the trick

kate-goldenring:
@itowlson I believe it should be cleaned up

itowlson (approved):
LGTM!

bacongobbler:
One other thing unrelated to this PR: I noticed `spin plugin list` seems to show the cloud plugin twice:

```
 ><> spin plugin list
cloud 0.1.0 [installed]
cloud 0.1.0 [installed]
js2wasm 0.1.0
js2wasm 0.2.0
js2wasm 0.3.0
js2wasm 0.4.0 [installed]
py2wasm 0.1.0
py2wasm 0.1.1
py2wasm 0.2.0
```

bacongobbler (approved):

kate-goldenring:
@bacongobbler I added a step to do `spin plugin update` for the user if it doesn't see the cloud plugin.  We can probably remove this later once we have the badgering and it is highly unlikely to not have updated. We may be able to remove this step in the future once we think everyone will have it. Duplication of the plugin should now be fixed via https://github.com/fermyon/spin-plugins/pull/25

itowlson:
@bacongobbler Thanks for flagging the "double cloud" quirk in the list.  It looks like at one point there were both a `cloud.json` and a `cloud@0.1.0.json` in the registry, both reporting v0.1.0.  There's now just the one, so `spin plugins update` made the duplication go away (for me).  @kate-goldenring Is this a situation I should guard against in the client, or is it a "shouldn't happen" in the registry?

itowlson:
Doh sorry, I completely missed all the bits where @kate-goldenring covered the duplication stuff already, please ignore.

PR #1453, opened 2023-05-05
Validate components are not granted access to undefined  k/v stores

itowlson:
@lann Once again I'd appreciate your insight into how you envisaged this working.  Your suggestion in #1371 was to add a "validate app" function to DynamicHostComponent, which fits nicely with a general sense of "components do the things they need at the right points in the lifecycle," but it seemed tricky getting the host component and the app around at the same time (the loader consumes the host components, then the app consumes the loader, so by the time the app exists, the host components are gone).

So this kind of special-cases it using the TriggerHooks mechanism - which has a convenient "called when app is loaded" entry point - and, as the previous PR ended up doing, making `build_key_value_component` aware of the component's needs in this regard.  It would be great to get your feedback if you had a more elegant way of doing it in mind.

### DIFF
--- a/crates/key-value/src/util.rs
+++ b/crates/key-value/src/util.rs
@@ -34,6 +34,12 @@ impl DelegatingStoreManager {
     }
 }
 
+impl DelegatingStoreManager {
+    pub fn store_names(&self) -> Vec<String> {
+        self.delegates.keys().cloned().collect()
+    }
+}
+
 #[async_trait]
 impl StoreManager for DelegatingStoreManager {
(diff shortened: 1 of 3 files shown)


lann:
> it seemed tricky getting the host component and the app around at the same time

It is a little tricky, but I believe I know how to do it. I can prioritize this if its going to be immediately useful.

itowlson:
Nah, no urgenc-- _notices @lann already did it_

itowlson:
Superseded by #1486.

PR #1455, opened 2023-05-05
Add `DynamicHostComponent::validate_app`

### LINKED ISSUE #1371: Add `DynamicHostComponent::validate_app`
e.g.:

```rust
trait DynamicHostComponent {
  ...
  fn validate_app(&self, app: &App) -> Result<()> {
    Ok(()
  }

### PR DESCRIPTION
**Pull request by lann**
Closes #1371 
Ref #1453

### DIFF
--- a/crates/app/src/host_component.rs
+++ b/crates/app/src/host_component.rs
@@ -1,9 +1,9 @@
 use std::{any::Any, sync::Arc};
 
-use anyhow::{Context, Result};
+use anyhow::Context;
 use spin_core::{AnyHostComponentDataHandle, EngineBuilder, HostComponent, HostComponentsData};
 
-use crate::AppComponent;
+use crate::{App, AppComponent};
 
 /// A trait for "dynamic" Spin host components.
 ///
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by itowlson — COMMENTED** [review]
I'm afraid a lot of the Dyns and Handles went over my head, but that's not really an issue from the user point of view, we mortals just have a function to implement and life is good.  So that side of things looks great - thanks!

I wanted, though, to flag components may need to change to take advantage of this design.  Specifically, the K/V component currently contains basically just a Rust closure that you give it a component and it returns a store manager, which in turn you give it a store name and it returns a store.  That is, it doesn't contain a list of of valid stores that `validate_app` could consult.  Although we could still validate a K/V app from this (by trying out every store name in a component's `key_value_stores` and seeing if any of them error), that has side-effects (every valid non-BYO store gets created even if never used), and presumably takes more time than a string compare (though probably not a noticeable startup delay).

So the question is, are we comfortable having host components include info that is needed only for validation, and specifically having to change the design of an existing host component to accommodate that?

(As ever, apologies if I am missing something that is obvious or that you have already anticipated.)

**Comment by lann** [comment]
> So the question is, are we comfortable having host components include info that is needed only for validation, and specifically having to change the design of an existing host component to accommodate that?

If they need to, sure. This is an optional trait method that just is just intended to improve DX by moving some runtime errors to "deploy" time.

In the case of the K/V component, I think there are a couple of reasonable options:

- `validate_app` could call `StoreManagerManager::get` (but _not_ `StoreManager::get`) for each configured store, and the `StoreManagerManager` closure returned in `build_key_value_component` could perform the validation.

- Add a method to `StoreManager`, e.g.:
```rust
trait StoreManager {
  ...
  /// Validate that the given store _might_ be accessible.
  /// Should perform only "cheap" validation, i.e. no network or disk I/O.
  fn validate_store(&self, name: &str) -> Result<(), Error> {
    Ok(())
  }
}
```

itowlson (approved):

PR #1459, opened 2023-05-08
Exercise `spin up --key-value` in tests

itowlson:
Main question is whether this should be covered in the existing KV test (as is done here) or broken out into a separate test.

(It would also be nice to generate a random init pair but at the moment it is a bit painful to tell the guest module what the random values are.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6022,6 +6022,7 @@ version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1674845326ee10d37ca60470760d4288a6f80f304007d92e5c53bab78c9cfd79"
 dependencies = [
+ "getrandom 0.2.8",
  "serde",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -65,7 +65,7 @@ toml = "0.6"
 tracing = { workspace = true }
 tracing-subscriber = { version = "0.3.7", features = ["env-filter"] }
 url = "2.2.2"
-uuid = "^1.0"
+uuid = { version = "^1.0", features = ["v4"] }
 wasmtime = { workspace = true }
 watchexec = { git = "https://github.com/watchexec/watchexec.git", rev = "8e91d26ef6400c1e60b32a8314cbb144fa33f288" }
 subprocess = "0.2.9"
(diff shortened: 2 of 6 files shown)


rajatjindal:
having it in same test is fine, but we just need to ensure that the key/value pair is random to avoid false negatives

one of the thing we do currently in this test is to do the kv operations with in `lib.rs`. what we could do instead is to have those operations triggered by calling the http endpoint. 

so we can do things like

```
key=`random key`
value=`random value`

spin up --key-value $key=$value
actual_value=curl http://localhost:3000/get?key=$key
assert $value == $actual_value
```

I do this in some synthetic monitors that we have and it seems to work just fine.

itowlson:
@rajatjindal That sounds good but I can't figure how to do it without significant surgery.  The issue is getting the dynamically generated random strings into the URL, because the URL is hardwired into the `checks` function.  If I try to pass the values in via  the call the `checks`, it fails because `assertions` requires an actual `fn` rather than a closure (i.e. cannot capture the variables from the outer scope).  And I can't pass the values as arguments to the `assertions` lambda, because that has to conform to `ChecksFunc` which has no scope for custom data.

I could change the definition of ChecksFunc, but that's used widely.  Or have an alternative `assertions_extra` of a function type that accepts extra data, but not sure how that would pan out as you need to have exactly one.

I'll continue to noodle but if you have any thoughts then I'd be glad of the advice!

rajatjindal:
could we do something like:

```
 let key = "some-random-key";
 env::set_var("KV_TEST_INIT_KEY", key.to_string());
```

and then

```
async fn checks(
            metadata: AppMetadata,
            // TODO: investigate why omitting these two next parameters does not
            // cause a compile time error but causes a runtime one
            _: Option<BufReader<ChildStdout>>,
            _: Option<BufReader<ChildStderr>>,
        ) -> Result<()> {
            println!(
                "random key is {}",
                env::var("KV_TEST_INIT_KEY").unwrap_or("".to_string())
            );
            assert_http_response(
```

but it won't work if we try to reuse the `testcase` (may still work if we reuse `test app`).

Rajat 🏃‍♂️ and  hide.

itowlson:
@rajatjindal Oh, sorry, I forgot to update!  I found a way to let `assertions` accept a closure and _with a bit of bonus cloning_ it seems to work now.  So it is ready for re-review.

kate-goldenring (approved):
Love the way you are passing the expected key and value through a url query! This is slick!

itowlson:
@kate-goldenring That was entirely @rajatjindal's idea so all the credit is to him!  I agree it is nifty!

rajatjindal:
thank you for fixing this @itowlson

PR #1461, opened 2023-05-08
cleanup statedir before test

### LINKED ISSUE #1458: Run local integration tests with clean or predictable starting state
Per #1448, I got a failure in `make-kv` which neither @rajatjindal nor @lann could reproduce.  It turned out this was because I had previously tried to adapt the test to include the new `spin up --key-value` option; apparently that worked, because  it left turds in the store.  So when I re-ran `make-kv`, even on main, an assertion about the set of keys in the store misfired.

Our local integration tests should either:

* Clear each test's local state before running it (e.g. delete `.spin` directory);
* Run each test with its own randomly generated state directoryl; or
* Reset local state to a known state before running.

I'm not sure of a good approach to the third option; I'd say consider it only if we have tests we want to run with a non-empty initial state.

NOTE: We should tackle this at the per-test level, not the per-session level, because otherwise we could get nasty heisentests where a failure only happens if tests run in a specific order (if two tests drive the same application but do not clean up between themselves).  If we want tests to run in parallel then the second option is probably safest.

cc @rajatjindal

rajatjindal:
fixes #1458. this takes the easy path of cleaning up the state before starting to run the test (actually just before `spin up`).


crates/e2e-testing/src/spin_controller.rs
```diff
@@ -46,6 +49,22 @@ impl Controller for SpinUp {
     ) -> Result<AppInstance> {
         let appdir = spin::appdir(app_name);
 
+        //cleanup state dir before spin up
+        let state_dir = match env::var("STATE_DIR") {
```
lann, line 53: This could accidentally delete a "real" state dir if a user happens to have a full path in their env. Maybe:

```suggestion
        let state_dir = match env::var("TEST_STATE_DIR") {
```

crates/e2e-testing/src/spin_controller.rs
```diff
@@ -46,6 +49,22 @@ impl Controller for SpinUp {
...
+                .unwrap(),
+        };
+        if fs::remove_dir_all(&state_dir).is_err() {
+            println!(
+                "warning: failed to cleanup existing statedir {}",
+                &state_dir
+            )
+        };
+
         let mut cmd = vec!["spin", "up"];
```
lann, line 68: Do you want to add `"--state-dir", state_dir` here?

crates/e2e-testing/src/spin_controller.rs
itowlson, line 53: Arguably, if someone is using a custom state directory, then it is on them to manage it; I agree with @lann that it could be dangerous to delete anything the tests don't "own"? Maybe simply (unconditionally) delete the `.spin` directory if it exists?

crates/e2e-testing/src/spin_controller.rs
```diff
@@ -46,6 +49,22 @@ impl Controller for SpinUp {
...
         let appdir = spin::appdir(app_name);
 
+        //cleanup state dir before spin up
+        let state_dir = match env::var("STATE_DIR") {
+            Ok(val) => val,
+            Err(_) => Path::new(appdir.as_str())
+                .join(".spin")
+                .into_os_string()
+                .into_string()
+                .unwrap(),
```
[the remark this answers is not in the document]
itowlson: As mentioned in the original issue, an advantage of using a tempdir per app run is that it is more robust to running tests in parallel (if they use the same underlying app).  I am not sure what the cost and impact are though.

crates/e2e-testing/src/spin_controller.rs
rajatjindal: I tried adding a tempdir here, but it cleans up before we had opportunity to run the actual tests (as tempdir goes out of scope)

itowlson: The way you're doing it makes the TempDir a temporary value, which allows it to drop immediately, which in turn cleans up the temporary directory.  Could you try something like:

```
let temp_dir = TempDir::new().unwrap();
let state_dir = format!("{}", temp_dir.path().display());
// party on state_dir
tc.run(controller).await.unwrap();
drop(temp_dir);
```

and see if that helps?

(The explicit `drop` is optional: `temp_dir` will get dropped at the end of the scope anyway.  I quite like it because it makes it clear when we expect the drop to happen (on the happy path), and provides a compiler error if someone later "improves" the code in ways that would make `temp_dir` go out of state too early.  Others disagree though!)

tests/testcases/mod.rs
```diff
@@ -20,14 +21,16 @@ pub mod all {
     pub async fn key_value_works(controller: &dyn Controller) {
         async fn checks(
             metadata: AppMetadata,
-            test_init_key: String,
```
rajatjindal, line 23: Hi @itowlson, here I have tried to implement an alternative solution to verify random key/value pair. This is to keep the signature of `checks` function same across all the tests. but I am happy to revert my changes if you insist.

crates/e2e-testing/src/spin_controller.rs
```diff
@@ -51,6 +51,14 @@ impl Controller for SpinUp {
             cmd.append(&mut xargs);
         }
 
+        // use the appdir as state dir unless explicitly provided,
+        // in which case it is user's responsibility to clean it up.
+        // this also helps ignore STATE_DIR env variable
+        if !xargs.contains(&"--state_dir") {
```
rajatjindal, line 57: I am doing this here because right now we reuse all our test cases for  `spin up` as well as `spin cloud deploy`, and if I understand correctly, we don't support `state dir` with `spin cloud deploy`.

crates/e2e-testing/src/spin_controller.rs
rajatjindal: this is a good point. (and a one that you had mentioned to me earlier about something similar, so my bad for not remembering it). thanks Lann.

i have removed the use of STATE_DIR

crates/e2e-testing/src/spin_controller.rs
rajatjindal: so, I am not adding a temp dir, but adding the state-dir to be test app dir explicitly (unless the flag is provided by user). this will also mean our testcase will ignore STATE_DIR env variable

tests/testcases/mod.rs
itowlson: This "random" generation is deterministic within a process, so could disguise a problem if the same code was run twice within the same process.  That's not an issue here, but could be if someone later extracted this to a library function, or if we started using "theory" style tests that were run with multiple inputs.

I am not sure what problem the change is aiming to solve.  Yes, the `checks` function itself ends up with a different signature.  But the closure being passed to `TestCaseBuilder::assertions` still has the right signature.  The `checks` function itself is an internal implementation detail of this test case, and should be free to accept whatever information it needs to perform the test rigorously.

To be clear, I've very happy to see improvements and simplifications to a rather messy piece of code.  And I'm happy that this works here and now.  But I'm wary of establishing this as a pattern and of "checks functions must all look the same" as a design constraint.  Happy to keep iterating on this though of course!

itowlson: Also, if we do choose to do it this way, I'd suggest doing it as a separate PR.  This PR now seems to contain more lines relating to "consistent `checks` signature" than to the stated "clean up state directory before test"...  *grin*

crates/e2e-testing/src/spin.rs
```diff
@@ -65,6 +66,16 @@ pub fn build_app(appname: &str) -> Result<Output> {
     utils::run(vec!["spin", "build"], Some(&appdir), None)
 }
 
+pub fn state_dir(appname: &str) -> String {
```
itowlson, line 69: It seems like you've moved away from the approach of using a temporary directory.  Could you say something about why that didn't work out please?  It would be useful to understand.  Thanks!

crates/e2e-testing/src/spin_controller.rs
```diff
@@ -112,9 +120,16 @@ impl Controller for SpinUp {
 
     async fn stop_app(
         &self,
-        _: Option<&str>,
+        appname: Option<&str>,
         process: Option<tokio::process::Child>,
     ) -> Result<()> {
+        if spin::cleanup_state_dir(appname.unwrap()).is_err() {
```
itowlson, line 126: This is very pernickety, but...  my personal taste would be to delete the state directory only once the app had stopped running.  In practice, though, it's not going to be an issue given our usage patterns.

rajatjindal: this is such a silly mistake, I am feeling ashamed. fixed :)

crates/e2e-testing/src/spin.rs
rajatjindal: We cannot initialize `tempdir` in `spin_controller` as it will go out of scope (and hence will get cleaned up) soon after calling `spin up`. 

to avoid that, we will have to add it to the function where we are initializing the test case itself (in `mod.rs`). But because we are using same set of test cases for `spin deploy`, now those will fail as `spin deploy` does not support `--state-dir`.

one way could be to update `cloud controller` (the implementation of `trait Controller` to run tests against cloud) to ignore the state-dir flags (by iterating over the `deploy_args`), but that will add more complexity to it. 

I hope it makes sense. I am still very open to hear if there are other ideas that can help here.

tests/testcases/mod.rs
rajatjindal: it finally clicked !! 

you changed the arguments of checks function only. and modified the closure to reflect that. the `assertions` function can remain unchanged across other test cases.

Thank you @itowlson for explaining. I've now reverted my changes. this looks awesome.

crates/e2e-testing/src/spin.rs
itowlson: Would it be possible to make the `TempDir` object local to `TestCase::run`, rather than a member of the controller?  Then rather than passing it as a flag, you could pass it as an argument into `Controller::run_app`: the local controller would compose a `--state-dir` argument, and the cloud controller would ignore it.  And drop the `TempDir` only once `spin up` has returned and `TestCase::run` is exiting.

Disclaimer: insert the "I don't know what I'm doing" dog here.  You know this part of the code _waaay_ better than I do.

rajatjindal: yeah that sounds like a fair compromise to me. I have made the change for this. thank you for working with me on this @itowlson, it is really appreciated.

crates/e2e-testing/src/testcase.rs
```diff
@@ -146,11 +147,24 @@ impl TestCase {
             spin::registry_push(&appname, registry_app_url.as_str())?;
         }
 
+        // create temp dir as state dir
+        let tempdir = TempDir::new().unwrap().into_path();
```
itowlson, line 151: `into_path` means it will no longer clean up the tempdir - it "throws away" the temporary nature and gives you a plain path somewhere in temp space; and dropping that plain path no longer does anything.  (https://docs.rs/tempdir/latest/tempdir/struct.TempDir.html#method.into_path)

You need `tempdir` (the variable you hold on to and drop at the end) to be an _actual `TempDir` instance_.  And you _must_ use `path()` rather than `into_path()` to get the OS path.  (https://docs.rs/tempdir/latest/tempdir/struct.TempDir.html#method.path)  `path()` does not consume the `TempDir` so the `TempDir` continues to "own" the directory and is able to delete it on drop.

I'm sorry, I should have flagged this nuance in my comment suggesting this pattern.  It's a trap we've fallen into before!

itowlson (changes requested):
I am really sorry but I did not explain a tricky nuance of how `TempDir` works and unfortunately this will not do quite what you want.  (It will work!  But it will leave temporary directories lying around.)  Many apologies, and happy to chat if my explanation is still confusing!

crates/e2e-testing/src/testcase.rs
rajatjindal: no you are fine. it's me. I should have tested it more and read documentation about it. fix coming up.

itowlson (approved):
Thanks for slogging through all the back and forth on this one!  Great to see it land.

rajatjindal:
thank you for your guidance on the changes. appreciate it.

PR #1463, opened 2023-05-08
Add Azure CosmosDB implementation for KV

radu-matei:
This commit adds an implementation of Azure CosmosDB as a backend for Spin's key/value store.
Specifically, this uses a CosmosDB account created with the NoSQL API, and requires the following contents for the runtime configuration file:

```toml
[key_value_store.default]
type = "azure_cosmos"
key = "<key>"
account = "<cosmos-account>"
database = "<cosmos-database>"
container = "<cosmos-container>"
```

The CosmosDB container *must* be created with the default partition key, `/id`:

Screenshot 2023-05-07 at 15 47 05

To test the changes in this commit, the runtime configuration file above
can be passed (and can be further tested with the KV explorer template):

```bash
$ spin up --runtime-config-file runtime-config.toml --key-value keyfromcli=valuefromcli 
Storing default key-value data to Azure CosmosDB
```

Thanks to @fibonacci1729 for pairing on this one!

cc @Mossaka @devigned


crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub client: OnceCell<Arc<Mutex<CollectionClient>>>,
+}
+
+impl KeyValueAzureCosmos {
+    pub fn new(key: String, account: String, database: String, collection: String) -> Result<Self> {
+        Ok(Self {
+            key,
+            account,
+            database,
+            collection,
```
radu-matei, line 31: I do have a question about the terminology used by Azure CosmosDB with the NoSQL API this implementation is using: while the client SDK uses `collection`, the UI seems to show `container` as the name for the entity grouping the documents, and I'm wondering which one should we expose to the user.

Any thoughts, @Mossaka @devigned?

cc @itowlson

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+                }
+            }
+            None => Err(Error::NoSuchKey),
+        }
+    }
+
+    async fn get_keys(&self) -> Result<Vec<String>, Error> {
+        let client = self.client.lock().await;
+        let query = client
+            .query_documents(Query::new("SELECT * FROM c".to_string()))
```
radu-matei, line 135: I _think_ we can tweak the query to only return `c.id` instead of the entire document?

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    sync::{Mutex, OnceCell},
+    task,
+};
+
+pub struct KeyValueAzureCosmos {
+    key: String,
+    account: String,
+    database: String,
+    collection: String,
+    pub client: OnceCell<Arc<Mutex<CollectionClient>>>,
```
fibonacci1729, line 22:
```suggestion
    client: OnceCell<Arc<Mutex<CollectionClient>>>,
```

crates/key-value-azure/src/lib.rs
devigned: I think container would be consistent with [the object model documentation](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/quickstart-dotnet?tabs=azure-portal%2Cwindows%2Cpasswordless%2Csign-in-azure-cli#object-model). Maybe collection was from old specs / docs to feel more MongoDB'ish.

On a more meta note, would specifying the containers as part of the configuration be limiting? Would a user likely use multiple containers or dynamically named containers?

Mossaka: In wasi-keyvalue, the name used is `bucket`. https://github.com/WebAssembly/wasi-keyvalue/blob/main/wit/types.wit

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+            collection,
+            client: OnceCell::new(),
+        })
+    }
+}
+
+#[async_trait]
+impl StoreManager for KeyValueAzureCosmos {
+    async fn get(&self, name: &str) -> Result<Arc<dyn Store>, Error> {
+        let client = task::block_in_place(|| {
```
Mossaka, line 40: Sorry if this is a stupid question, but why is it using a blocking call in an async func?

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+            .lock()
+            .await
+            .document_client(&pair.id, &pair.id)
+            .map_err(log_error)?;
+        document_client.delete_document().await.map_err(log_error)?;
+        Ok(())
+    }
+
+    async fn exists(&self, key: &str) -> Result<bool, Error> {
+        Ok(self.get_keys().await?.contains(&key.to_string()))
```
Mossaka, line 101: I wonder if calling `get_pair` here is cheaper.

Mossaka (commented):
Looking good!

crates/key-value-azure/src/lib.rs
itowlson: In this particular case I'd align with Cosmos nomenclature rather than wasi-keyvalue, because this is talking _specifically_ about the Cosmos back end store, not the WASI front end API.

radu-matei: Updated this to be `container`.

crates/key-value-azure/src/lib.rs
radu-matei: Ooh, good catch. That was a left-over from an early implementation of this.
Fixed.

crates/key-value-azure/src/lib.rs
radu-matei: Updated.

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+                let token =
+                    AuthorizationToken::primary_from_base64(&self.key).map_err(log_error)?;
+
+                let cosmos_client = CosmosClient::new(self.account.clone(), token);
+                let database_client = cosmos_client.database_client(self.database.clone());
+
+                let collection_client = database_client.collection_client(self.container.clone());
+                Ok(Arc::new(Mutex::new(collection_client)))
+            })
+            .await?;
```
rylev, line 50: Is there a reason why we're initializing here instead of in `KeyValueAzureCosmos::new`?

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+        Ok(pair.value)
+    }
+
+    async fn set(&self, key: &str, value: &[u8]) -> Result<(), Error> {
+        let pair = Pair {
+            id: key.to_string(),
+            value: value.to_vec(),
+        };
+        self.client
+            .lock()
```
rylev, line 77: I'm unsure why we need to lock the client? We are just using methods that take `&self`, no?

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        // There can be no duplicated keys, so we create the stream and only take the first result.
+        let mut stream = query.into_stream::<Pair>();
+        let res = stream.next().await;
+        match res {
+            Some(r) => {
+                let r = r.map_err(log_error)?;
+                match r.results.first() {
+                    Some(p) => Ok(p.0.clone()),
+                    None => Err(Error::NoSuchKey),
```
rylev, line 127: It might be easier for us to have a method that returns a `Result<Option<Pair>, Error>` so that existence checks are a bit easier. I realize though that the public interface requires the current return type.

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+            .max_item_count(1);
+
+        // There can be no duplicated keys, so we create the stream and only take the first result.
+        let mut stream = query.into_stream::<Pair>();
+        let res = stream.next().await;
+        match res {
+            Some(r) => {
+                let r = r.map_err(log_error)?;
+                match r.results.first() {
+                    Some(p) => Ok(p.0.clone()),
```
rylev, line 126: Is this clone actually necessary?

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+        let query = client
+            .query_documents(Query::new("SELECT * FROM c".to_string()))
+            .query_cross_partition(true);
+        let mut res = Vec::new();
+
+        let mut stream = query.into_stream::<Pair>();
+        while let Some(resp) = stream.next().await {
+            let resp = resp.map_err(log_error)?;
+            for (pair, _) in resp.results {
+                res.push(pair.id.clone());
```
rylev, line 145: Is this clone necessary?

crates/trigger/Cargo.toml
```diff
@@ -27,7 +28,7 @@ spin-config = { path = "../config" }
 spin-core = { path = "../core" }
 spin-loader = { path = "../loader" }
 spin-manifest = { path = "../manifest" }
-tokio = { version = "1.23", features = ["fs"] }
+tokio = { version = "1.23", features = ["full"] }
```
rylev, line 31: What does this change allow?

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -141,6 +164,9 @@ impl TriggerHooks for KeyValuePersistenceMessageHook {
                     println!("Using in-memory default key-value store; data will not be saved!");
                 }
             }
+            KeyValueStoreOpts::AzureCosmos(_store_opts) => {
+                println!("Storing default key-value data to Azure CosmosDB");
```
rylev, line 168: Would it make sense to print the subscription?

radu-matei:
@rylev and I paired on this earlier today and addressed most of the feedback. This should be ready for a re-review.

Thanks!

crates/key-value-azure/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+        self.client
+            .create_document(pair)
+            .is_upsert(true)
+            .await
+            .map_err(log_error)?;
+        Ok(())
+    }
+
+    async fn delete(&self, key: &str) -> Result<(), Error> {
+        let pair = self.get_pair(key).await?;
```
Mossaka, line 64: It would be better to use `self.exists(key).await?`

Mossaka (approved):
LGTM! Just have one comment there.

crates/key-value-azure/src/lib.rs
radu-matei: Reading the interface, we should not return the error if the pair does not exist: https://github.com/fermyon/spin/blob/main/wit/preview2/key-value.wit#L56-L61

radu-matei: Updated this to follow the spec and to use `self.exists`.

crates/key-value-azure/src/lib.rs
```diff
@@ -61,13 +61,17 @@ impl Store for AzureCosmosStore {
...
+            Ok(true) => {
+                let document_client = self.client.document_client(key, &key).map_err(log_error)?;
+                document_client.delete_document().await.map_err(log_error)?;
+                Ok(())
+            }
+            // The Spin key-value interface mandates that no error should be raised if the pair did
+            // not exist.
+            Ok(false) => Ok(()),
+            Err(e) => Err(e),
+        }
```
rylev, line 74:
```rust
if self.exists(key)? {
  let document_client = self.client.document_client(key, &key).map_err(log_error)?;
  document_client.delete_document().await.map_err(log_error)?;
}
Ok(())
```

rylev (approved):
:shipit:

rylev (approved):
:shipit:

fibonacci1729 (approved):

tpmccallum:
Documentation for this is being worked on at https://github.com/fermyon/developer/issues/611

PR #1465, opened 2023-05-08
feat(sdk/go): key-value SDK v2

adamreese:
This change is a refactor of the key/value Go SDK

- The consumer of the key value sdk is responsible for opening and closing the connection.  This change will handle these operations in each method call
- Use a struct for `Store` to allow consumers to create an interface
- Use a custom error type enabling error type checking without doing a string match on error message
- Remove exposed C types
- Follow golang library conventions


sdk/go/kv/kv.go
```diff
@@ -0,0 +1,183 @@
...
+func (s *Store) Close() {
+	if s.active {
+		C.key_value_close(C.uint32_t(s.ptr))
+	}
+	s.active = false
+}
+
+// Get retrieves a value from Store.
+func (s *Store) Get(key string) ([]byte, error) {
+	if err := s.open(); err != nil {
```
endocrimes, line 39: I think we probably want to fail when we're not open rather an doing an implicit open - it's easier to add an implicit open later than it is to remove it 😅

adamreese: Agreed.  Thanks!

adamreese:
Closing until spin 2.0 or a new release process.

PR #1467, opened 2023-05-09
Rationalise `-f` and accept directory everywhere

itowlson:
At the moment, `spin up` allows you to pass a directory to `-f` e.g. `spin up -f myapp/`, but `build`, `deploy`, `registry push` and `watch` do not.  This rationalises the implementation of the `-f` option across those commands.  Please let me know if there's anywhere I've missed!

Note that `up` is still slightly different because it accepts registry references as well.  So I'd welcome feedback on whether changing the long form to be `--from` across all commands is reasonable or confusing.  (I have kept the `--file` alias so backward compatibility is maintained.)

cc @lann for `spin doctor`

### DIFF
--- a/src/manifest.rs
+++ b/src/manifest.rs
@@ -1,0 +1,32 @@
+use anyhow::anyhow;
+use std::path::{Path, PathBuf};
+
+use crate::opts::DEFAULT_MANIFEST_FILE;
+
+/// Resolves a manifest path provided by a user, which may be a file or
+/// directory, to a path to a manifest file.
+pub(crate) fn resolve_file_path(provided_path: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
+    let path = provided_path.as_ref();
+
+    if path.is_file() {
+        Ok(path.to_owned())
+    } else if path.is_dir() {
+        let file_path = path.join(DEFAULT_MANIFEST_FILE);
+        if file_path.is_file() {
+            Ok(file_path)
+        } else {
+            Err(anyhow!(
+                "Directory {} does not contain a file named 'spin.toml'",
+                path.display()
+            ))
+        }
+    } else {
+        let pd = path.display();
+        let err = match path.try_exists() {
+            Err(e) => anyhow!("Error accessing path {pd}: {e:#}"),
+            Ok(false) => anyhow!("No such file or directory '{pd}'"),
+            Ok(true) => anyhow!("Path {pd} is neither a file nor a directory"),
+        };
+        Err(err)
+    }
+}
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -11,14 +11,17 @@ use super::up::UpCommand;
 #[derive(Parser, Debug)]
 #[clap(about = "Build the Spin application", allow_hyphen_values = true)]
 pub struct BuildCommand {
-    /// Path to application manifest. The default is "spin.toml".
+    /// The application to build. This may be a manifest (spin.toml) file, or a
+    /// directory containing a spin.toml file.
+    /// If omitted, it defaults to "spin.toml".
     #[clap(
         name = APP_MANIFEST_FILE_OPT,
         short = 'f',
         long = "from",
         alias = "file",
+        default_value = DEFAULT_MANIFEST_FILE
     )]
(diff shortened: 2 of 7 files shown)


itowlson:
@bacongobbler noted that deploy is likely to move to a plugin, and we will need some way to ensure it (and other plugins that read manifests) behave consistently with the way Spin does.  This could be solved for Rust by exposing these functions to plugins (either by documenting the stable parts of the Spin API surface, or perhaps better by surfacing them from a "plugin utilities" crate).  This would suffice for the cloud plugin case; we can think about other languages longer term.

src/manifest.rs
```diff
@@ -0,0 +1,30 @@
...
+use crate::opts::DEFAULT_MANIFEST_FILE;
+
+pub(crate) fn manifest_path(path: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
+    let path = path.as_ref();
+    if path.exists() {
+        if path.is_file() {
+            Ok(path.to_owned())
+        } else if path.is_dir() {
+            let file_path = path.join(DEFAULT_MANIFEST_FILE);
+            if file_path.exists() && file_path.is_file() {
```
lann, line 13:
```suggestion
            if file_path.is_file() {
```

> Returns true if the path exists on disk and is pointing at a regular file.

src/manifest.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::anyhow;
+use std::path::{Path, PathBuf};
+
+use crate::opts::DEFAULT_MANIFEST_FILE;
+
+pub(crate) fn manifest_path(path: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
+    let path = path.as_ref();
+    if path.exists() {
```
lann, line 8: Consider [`try_exists`](https://doc.rust-lang.org/std/fs/fn.try_exists.html) which returns `Result(bool)`, erroring if existence cannot be determined due to e.g. permissions issues. Could flatten that with the nested conditionals since `is_file` and `is_dir` both imply `exists`.

src/manifest.rs
itowlson: Well, duh.  Thanks.

lann (approved):

src/manifest.rs
```diff
@@ -0,0 +1,30 @@
...
+                "Directory {} does not contain a file named 'spin.toml'",
+                path.display()
+            ))
+        }
+    } else {
+        let pd = path.display();
+        let err = match path.try_exists() {
+            Err(e) => anyhow!("Error accessing path {pd}: {e:#}"),
+            Ok(false) => anyhow!("No such file or directory '{pd}'"),
+            Ok(true) => anyhow!("Path {pd} is neither a file nor a directory"),
```
rylev, line 26: Seems like this should only happen on broken symbolic links. Maybe referencing that a link is broken might be more clear?

src/manifest.rs
```diff
@@ -0,0 +1,30 @@
+use anyhow::anyhow;
+use std::path::{Path, PathBuf};
+
+use crate::opts::DEFAULT_MANIFEST_FILE;
+
+pub(crate) fn manifest_path(path: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
```
rylev, line 6: Nit: can we add a doc comment here? It's not immediately clear why this function needs to take a path argument, and I think a document would easily clarify this.

rylev (commented):
I don't have particularly strong feelings on this, but it seems reasonable to me.

src/manifest.rs
lann: Or missing permissions on a parent dir, or (presumably) disk/fs error.

src/manifest.rs
itowlson: Added doc comment and renamed to `resolve_file_path`

src/manifest.rs
itowlson: This error message, which already existed, seems in danger of turning into a copy of _Troubleshooting the Unix (R) Operating System_ (Addison Wesley, 1991)...!  I'm happy to adopt concrete phrasing suggestions if we can nut one out quickly; otherwise I suggest we leave it as is and tackle improving it as a separate piece of work.

PR #1468, opened 2023-05-09
add loading spinner to registry push/pull

karthik2804:
Adds a dotted loading animation while the registry pushes/pulls happen to make it clearer about the progress being made.

### DIFF
--- a/src/commands/registry.rs
+++ b/src/commands/registry.rs
@@ -1,9 +1,9 @@
+use crate::opts::*;
 use anyhow::{Context, Result};
 use clap::{Parser, Subcommand};
+use indicatif::{ProgressBar, ProgressStyle};
 use spin_oci::Client;
-use std::{io::Read, path::PathBuf};
-
-use crate::opts::*;
+use std::{io::Read, path::PathBuf, time::Duration};
 
 /// Commands for working with OCI registries to distribute applications.
 #[derive(Subcommand, Debug)]
@@ -61,8 +61,10 @@ impl Push {
         let app = spin_loader::local::from_file(&app_file, Some(dir.path())).await?;
 
         let mut client = spin_oci::Client::new(self.insecure, None).await?;
-        let digest = client.push(&app, &self.reference).await?;
 
+        let _spinner = create_dotted_spinner(2000, "Pushing app to the Registry".to_owned());
+
+        let digest = client.push(&app, &self.reference).await?;
         match digest {
             Some(digest) => println!("Pushed with digest {digest}"),
             None => println!("Pushed; the registry did not return the digest"),
@@ -92,8 +94,11 @@ impl Pull {
     /// Pull a Spin application from an OCI registry
     pub async fn run(self) -> Result<()> {
         let mut client = spin_oci::Client::new(self.insecure, None).await?;
-        client.pull(&self.reference).await?;
 
+        let _spinner = create_dotted_spinner(2000, "Pulling app from the Registry".to_owned());
+
+        client.pull(&self.reference).await?;
+        println!("Successfully pulled the app from the registry");
         Ok(())
     }
 }
@@ -164,4 +169,16 @@ impl Login {
         );
         Ok(())
     }
+}
+
+fn create_dotted_spinner(interval: u64, message: String) -> ProgressBar {
+    let spinner = ProgressBar::new_spinner();
+    spinner.enable_steady_tick(Duration::from_millis(interval));
+    spinner.set_style(
+        ProgressStyle::with_template("{msg}{spinner}\n")
+            .unwrap()
+            .tick_strings(&[".", "..", "...", "....", "....."]),
+    );
+    spinner.set_message(message);
+    spinner
 }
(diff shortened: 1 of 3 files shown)


src/commands/registry.rs
```diff
@@ -92,8 +112,27 @@ impl Pull {
...
+        );
+        pb.set_message("Pulling app from registry");
+
+        let result = client.pull(&self.reference).await;
+        match result {
+            Ok(_) => {
+                pb.set_style(ProgressStyle::with_template("{msg:.green}")?);
+                pb.finish_with_message("Successfully pulled app from registry");
+            }
+            _ => {
```
itowlson, line 131: This looks like we are now discarding the error rather than reporting it back to the user?

src/commands/registry.rs
```diff
@@ -92,8 +112,27 @@ impl Pull {
     /// Pull a Spin application from an OCI registry
     pub async fn run(self) -> Result<()> {
         let mut client = spin_oci::Client::new(self.insecure, None).await?;
-        client.pull(&self.reference).await?;
 
+        let pb = ProgressBar::new_spinner();
```
itowlson, line 116: This pattern, which I think is used in both places, feels like it could be good to encapsulate e.g.

```
let spinner = SpinnySpinner::start("Pulling app from registry");
// Do the long-running work
match result {
  Case1 => spinner.success("Successfully pulled app from registry"),
  Case2 => spinner.error("Failed to pull app from registry"),
}
```

(The names are not great because they don't communicate stopping/finishing but hopefully gives a sense of what I mean... capturing the standard dots and tempo and styles into a simplified API.)

src/commands/registry.rs
```diff
@@ -61,11 +61,31 @@ impl Push {
...
+        pb.enable_steady_tick(Duration::from_millis(200));
+        pb.set_style(
+            ProgressStyle::with_template("{msg}{spinner}")
+                .unwrap()
+                .tick_strings(&[".", "..", "...", "....", "....."]),
+        );
+        pb.set_message("Pushing app to registry");
+
+        let digest = client.push(&app, &self.reference).await?;
```
itowlson, line 74: Do you need to do anything to the progress bar if this returns Err, or will it tidy itself up correctly as it goes out of scope?  (Asking because you explicitly handle the Err case in pull.)

itowlson (changes requested):
A few suggestions, some of which may be longer term, but also a couple of things that I think may be mistakes or might lose information?  Please correct me if I'm wrong of course!

src/commands/registry.rs
```diff
@@ -61,11 +61,31 @@ impl Push {
         let app = spin_loader::local::from_file(&app_file, Some(dir.path())).await?;
 
         let mut client = spin_oci::Client::new(self.insecure, None).await?;
-        let digest = client.push(&app, &self.reference).await?;
 
+        let pb = ProgressBar::new_spinner();
```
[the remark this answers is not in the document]
karthik2804: I am noticing it too. Will look into it.

src/commands/registry.rs
```diff
@@ -61,11 +61,31 @@ impl Push {
...
+        pb.enable_steady_tick(Duration::from_millis(200));
```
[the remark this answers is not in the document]
karthik2804: Makes sense. I will slow it down to every 2 seconds.

src/commands/registry.rs
```diff
@@ -61,11 +61,31 @@ impl Push {
...
+            Some(digest) => {
+                pb.set_style(ProgressStyle::with_template("{msg:.green}")?);
+                let success_string = format!(
+                    "Successfully pushed app to registry with digest: {}",
+                    digest
+                );
+                pb.finish_with_message(success_string);
+            }
+            None => {
+                pb.set_style(ProgressStyle::with_template("msg:.red}")?);
```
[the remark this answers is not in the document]
karthik2804: I am not sure if we use colored out elsewhere. We could add ✔️ and ❌  to help make it clear about the case.

src/commands/registry.rs
karthik2804: Good point. It tidies itself up when it goes out of scope. I think we can remove handling the Err case in pull.

src/commands/registry.rs
karthik2804: Will revert this.

src/commands/registry.rs
```diff
@@ -61,10 +61,14 @@ impl Push {
...
 
         let mut client = spin_oci::Client::new(self.insecure, None).await?;
-        let digest = client.push(&app, &self.reference).await?;
 
+        let _spinner = create_dotted_spinner(2000, "Pushing app to the Registry".to_owned());
+
+        let digest = client.push(&app, &self.reference).await?;
         match digest {
-            Some(digest) => println!("Pushed with digest {digest}"),
+            Some(digest) => {
```
itowlson, line 69: nit: are the braces needed/desired?

itowlson (approved):
LGTM - thanks!

src/commands/registry.rs
karthik2804: fixed

PR #1470, opened 2023-05-10
ref(sdk/go): use `spin build` for building test fixtures

adamreese:
This commit changes test fixtures to be built using `spin build` rather than using TinyGo directly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@ ignored-assets
 main.wasm
 .parcel-cache
 .vscode/*.log
-go.sum
 tests/**/Cargo.lock
 crates/**/Cargo.lock
 examples/**/Cargo.lock
--- a/sdk/go/http/testdata/http-tinygo/spin.toml
+++ b/sdk/go/http/testdata/http-tinygo/spin.toml
@@ -10,3 +10,5 @@ id = "http-tinygo-test"
 source = "main.wasm"
 [component.trigger]
 route = "/hello/..."
+[component.build]
+command = "tinygo build -target=wasi -gc=leaking -no-debug -o main.wasm main.go"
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -6,7 +6,6 @@ ignored-assets
 main.wasm
 .parcel-cache
 .vscode/*.log
-go.sum
```
adamreese, line 9: A few go.sum files were checked in before ignored. Some TinyGo examples have a go.sum file while others don't.  I figured it's best to leave them in.

sdk/go/integration_test.go
```diff
@@ -56,18 +56,12 @@ func startSpin(t *testing.T, spinfile string) *testSpin {
 	}
 }
 
-func buildTinyGo(t *testing.T, dir string) {
+func build(t *testing.T, dir string) {
 	t.Helper()
 
 	t.Log("building example: ", dir)
 
-	getCmd := exec.Command("go", "mod", "tidy")
```
adamreese, line 64: This is no longer required because go.sum files are now checked in.

sdk/go/http/testdata/http-tinygo/spin.toml
```diff
@@ -10,3 +10,5 @@ id = "http-tinygo-test"
 source = "main.wasm"
 [component.trigger]
 route = "/hello/..."
+[component.build]
+command = "tinygo build -target=wasi -gc=leaking -no-debug -o main.wasm main.go"
```
radu-matei, line 14: I am somewhat starting to reconsider `-no-debug` as the default — I think the context we get out of having debug symbols might be worth the extra binary size.


That being said, this does not block this PR. If we decide on it, a follow-up PR should change these across the examples and eventually docs.

PR #1471, opened 2023-05-10
Remove autobindle from recommended VS Code extensions

itowlson:
It no longer seems like good value for money.


fibonacci1729 (approved):

fibonacci1729:
👋

PR #1472, opened 2023-05-10
ensure deterministic component iteration order

### LINKED ISSUE #1469: Print HTTP routes in stable order
When Spin prints:

```
Available Routes:
  this: http://127.0.0.1:3000/this
  that: http://127.0.0.1:3000 (wildcard)
  other: http://127.0.0.1:3000/other
```

the order is currently random.  This defeats muscle memory, and requires users to scan the whole list to find the one they're interested in.  (You scoff?  Try it on https://github.com/fermyon/developer/.)

The order should:

* At minimum, be stable: not vary from run to run
* Ideally, lend itself to searching, e.g. alphabetical, or most general prefix first and alphabetically by URL within that

I have a feeling it used to be stable and the randomness is a regression - I only started noticing this on the docs site recently but I could be wrong!

dicej:
This regressed due to changes I made to `TriggerAppEngine::new` to support components.  I used a `HashMap` without realizing the iteration order of triggers (and their configs) matters.

Fixes #1469

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5352,6 +5352,7 @@ dependencies = [
  "ctrlc",
  "dirs 4.0.0",
  "futures",
+ "indexmap",
  "outbound-http",
  "outbound-mysql",
  "outbound-pg",
--- a/crates/trigger/Cargo.toml
+++ b/crates/trigger/Cargo.toml
@@ -11,6 +11,7 @@ clap = { version = "3.1.15", features = ["derive", "env"] }
 ctrlc = { version = "3.2", features = ["termination"] }
 dirs = "4"
 futures = "0.3"
+indexmap = "1"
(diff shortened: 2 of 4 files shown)


dicej:
Note that I considered writing a test for this, but the non-determinism makes it tricky.  I'd prefer to avoid tests which sometimes pass and sometimes fail, and I'm not sure how to avoid that here.

itowlson (approved):
Thanks for tracking this down and for the super quick fix!

itowlson:
Re testing, it could be done probabilistically - use a manifest with (say) 5 components in it, check the print order matches the manifest order, repeat 20 times, and that should get it to a "GUID collision/cosmic ray bitflip" risk of being wrong.  But given the very small impact of ordering issues, I'm not sure it's worth it!

Issue #1473, opened 2023-05-11
Add a short argument for `--accept-defaults`


mikkelhegn:
It would be nice with a short argument for `--accept-defaults` when using `spin new`.

It's particularly handy when you want to default to a flow like described here #1391 - particularly when doing a lot of demos, this is really convenient.

dicej:
Looks like this has been addressed; closing.

PR #1474, opened 2023-05-11
Add 'a' as short argumet for --accept-defaults

mikkelhegn:
Implements #1473 

Not sure if `-a` is the better option for this? Maybe `-d`?

Marking this as a draft, since the issue was create simultaneously.

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -49,7 +49,7 @@ pub struct TemplateNewCommandCore {
 
     /// An optional argument that allows to skip prompts for the manifest file
     /// by accepting the defaults if available on the template
-    #[clap(long = "accept-defaults", takes_value = false)]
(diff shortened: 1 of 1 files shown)


radu-matei (approved):
This seems like a reasonable change to me.
(non-blocking: Do we want to consider an environment variable so this is always enabled?)

rylev:
Bikeshed: `-y` (short for yes) could also be another option.

itowlson:
@rylev I am not sure that the `--accept-defaults` contract is "always say yes to whatever the computer suggests" - it is more narrowly "always say yes to the default parameter value suggestions."  We can of course add a `-y/--yes` with the broader meaning.

(That said, right now I'm not sure there are any other prompts, but I have a couple of WIPs that would introduce them on specific unhappy paths.)

itowlson (approved):

michelleN (approved):

PR #1475, opened 2023-05-11
Track memory and expose from Store Data

rylev:
This will allow spin runtime implementers to track memory usage.


lann:
Update `record_resource_usage` to use this?

lann (approved):

PR #1480, opened 2023-05-12
app: Add "inert" App feature

lann:
Adds an App::inert function which can return an App without an associated Loader. This is useful for accessing LockedApp metadata in contexts where a Loader is unavailable.

### DIFF
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -136,20 +136,26 @@ impl OwnedApp {
     }
 }
 
+// Implementation detail of [`App::inert`]; "sealed" to prevent external impls.
+mod private {
+    pub trait MaybeLoader {}
+}
+use private::MaybeLoader;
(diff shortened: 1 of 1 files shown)


crates/app/src/lib.rs
```diff
@@ -353,6 +368,14 @@ impl<'a> AppTrigger<'a> {
     }
 }
 
+// "Seal" MaybeLoader so external crates cannot implement it
+mod private {
+    pub trait MaybeLoader {}
+}
+use private::MaybeLoader;
+impl MaybeLoader for AppLoader {}
+impl MaybeLoader for () {}
```
rylev, line 377: Nit: if we do go this route, it might be better to make a `struct InertLoader;` instead of doing this on unit which is not very self-describing

fibonacci1729 (approved):

PR #1483, opened 2023-05-13
feat(sdk/rust): add optional feature for serde

radu-matei:
This commit updates the Rust SDK to include two new functions that integrates key/value with serde. Specifically:

- `set_json`: this function takes a key and a struct that can be serialized by serde into JSON, and that representation is added to the store
- `get_json`: this function takes a key, reads the key/value pair from the store, then deserializes it into the specified type

Considering that storing serialized state in Spin's key/value store is one of the popular use cases for the API, these convenience functions shoudl simplify that operation.

This functionality is added as an optional feature in the SDK, so it doesn't add to the binary size of the resulting component unless needed.

### DIFF
--- a/sdk/rust/src/key_value.rs
+++ b/sdk/rust/src/key_value.rs
@@ -8,6 +8,9 @@ wit_bindgen_rust::import!("../../wit/ephemeral/key-value.wit");
 
 use key_value::Store as RawStore;
 
+#[cfg(feature = "json")]
+use serde::{de::DeserializeOwned, Serialize};
+
 /// Errors which may be raised by the methods of `Store`
 pub type Error = key_value::Error;
 
@@ -60,6 +63,29 @@ impl Store {
     /// Get the set of keys in this store.
     pub fn get_keys(&self) -> Result<Vec<String>, Error> {
         key_value::get_keys(self.0)
+    }
+
+    #[cfg(feature = "json")]
+    /// Serialize the given data to JSON, then set it as the value for the specified `key`.
+    pub fn set_json<T: Serialize>(
+        &self,
+        key: impl AsRef<str>,
+        value: &T,
+    ) -> Result<(), anyhow::Error> {
+        Ok(key_value::set(
+            self.0,
+            key.as_ref(),
+            &serde_json::to_vec(value)?,
+        )?)
+    }
+
+    #[cfg(feature = "json")]
+    /// Deserialize an instance of type `T` from the value of `key`.
+    pub fn get_json<T: DeserializeOwned>(&self, key: impl AsRef<str>) -> Result<T, anyhow::Error> {
+        Ok(serde_json::from_slice(&key_value::get(
+            self.0,
+            key.as_ref(),
+        )?)?)
     }
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4595,9 +4595,9 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.154"
+version = "1.0.163"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


sdk/rust/src/key_value.rs
```diff
@@ -61,6 +62,30 @@ impl Store {
     pub fn get_keys(&self) -> Result<Vec<String>, Error> {
         key_value::get_keys(self.0)
     }
+
+    #[cfg(feature = "serde")]
+    /// Set the value for the specified key.
+    pub fn set_json<'a, T>(&self, key: impl AsRef<str>, value: &'a T) -> Result<(), anyhow::Error>
+    where
+        T: ?Sized + Serialize + Deserialize<'a>,
```
fibonacci1729, line 70:
```suggestion
        T: ?Sized + Serialize,
```

sdk/rust/src/key_value.rs
```diff
@@ -61,6 +62,30 @@ impl Store {
...
+            key.as_ref(),
+            &serde_json::to_vec(value)?,
+        )?)
+    }
+
+    #[cfg(feature = "serde")]
+    /// Get the value for the specified key.
+    pub fn get_json<T>(&self, key: impl AsRef<str>) -> Result<T, anyhow::Error>
+    where
+        T: ?Sized + Serialize + for<'de> Deserialize<'de>,
```
fibonacci1729, line 83:
```suggestion
        T: ?Sized + for<'de> Deserialize<'de>,
```

fibonacci1729 (approved):
Sounds like a good idea to me!

radu-matei:
I'm not entirely convinced that the names for these functions are the best.

`set_to_json` and `get_from_json` instead?

Naming...

sdk/rust/src/key_value.rs
```diff
@@ -61,6 +64,31 @@ impl Store {
     pub fn get_keys(&self) -> Result<Vec<String>, Error> {
         key_value::get_keys(self.0)
     }
+
+    #[cfg(feature = "serde")]
```
rylev, line 68: I think it's more appropriate to have this feature be called "json".

sdk/rust/src/key_value.rs
```diff
@@ -61,6 +64,31 @@ impl Store {
...
+        Ok(key_value::set(
+            self.0,
+            key.as_ref(),
+            &serde_json::to_vec(value)?,
+        )?)
+    }
+
+    #[cfg(feature = "serde")]
+    /// Deserialize an instance of type `T` from the value of `key`.
+    pub fn get_json<T>(&self, key: impl AsRef<str>) -> Result<T, anyhow::Error>
```
rylev, line 83: For naming I think `{get,set}_json` is fine. I could imagine `{get,set}_as`, but that hides the fact that we assume json...

sdk/rust/src/key_value.rs
```diff
@@ -61,6 +64,31 @@ impl Store {
...
+    where
+        T: ?Sized + for<'de> Deserialize<'de>,
```
rylev, line 85: Do we want these bounds? I would have expected this to be `T: DeserializedOwned` as it doesn't really make sense to return something that has a lifetime tied to the result from `key_value::get` as we drop that at the end of this method. I'm also struggling to think of a use cased to have an unsized type returned from this method.

rylev (changes requested):

sdk/rust/src/key_value.rs
radu-matei: Updated to `T: DeserializeOwned`.

rylev (approved):
:tada:

itowlson:
Probably a bit late but I am now wondering if calling the new functions `serialize` and `deserialize` (or variants such as `serialize_to`/`'deserialize_from`)  might have communicated it better than "JSON."

Issue #1481, opened 2023-05-12
Color output


rylev:
Currently spin's CLI output is almost exclusively in the default text color of the terminal it is in. This can lead to large walls of mono-colored output which is hard to parse. Most modern CLI tools output in color (when the functionality is available). I believe spin should as well. 

## Considerations 

There are certain considerations we should keep in mind:

* We should use color sparingly as if we use it too much, we end up right back in the situation we currently find ourselves in: hard to parse output.
* We should use color consistently. For example, if `Error:` is colored red in some output, it should be colored red in all output.
* We should not produce color escape codes when there is not terminal available, when the user has certain environment variables set (e.g., `NO_COLOR`), etc.

## Next steps

I believe it would be very difficult to design a grand vision for color and just go execute on it. We should think about how we want to apply color incrementally while still following the considerations laid out above. 

To that end I think there are some ideas we can start with. After implementing these ideas, we can evaluate how the user experience has improved (or not) and determine if we want to continue, revert, or stop making changes to the color output. 

Possible ideas include:
* Coloring `Error` in error messages red
* Coloring the first word of steps in `spin up` and `spin build` green
* Coloring warnings yellow (or at least adding a yellow "Warning" to them).

itowlson:
Accessibility should also be a consideration.  Naively, I'd guess that colouring at least won't make things worse that our current monochrome output.  But 1. we should validate that it doesn't and 2. if we can find options that play nicely with screen readers / colour blindness / etc. then that should be a consideration.

cc @flynnduism for graphic design and information layout sensibilities

rylev:
Going to close this as complete since we have color for various things now. We can open smaller issues for any further improvements.

PR #1484, opened 2023-05-13
feat(plugins): show description and homepage of installed plugin

cardoso:
It would be nice to give plugins a chance to display something once they're installed such as instructions or notices.

Showing the description would be a good start.

I also included the homepage (if it's HTTPS) as a good place to link to some documentation.

### DIFF
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -45,11 +45,7 @@ impl PluginManifest {
     }
 
     pub fn homepage_url(&self) -> Option<Url> {
-        self.homepage
-            .as_deref()
-            .into_iter()
-            .flat_map(Url::parse)
-            .next()
+        Url::parse(&self.homepage?).ok()
     }
 
     pub fn has_compatible_package(&self) -> bool {
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -45,7 +45,7 @@ impl PluginManifest {
     }
 
     pub fn homepage_url(&self) -> Option<Url> {
-        Url::parse(&self.homepage).ok()
+        Url::parse(self.homepage.as_deref()?).ok()
     }
 
     pub fn has_compatible_package(&self) -> bool {
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, Context, Result};
 use semver::{Version, VersionReq};
 use serde::{Deserialize, Serialize};
+use url::Url;
 
 use crate::PluginStore;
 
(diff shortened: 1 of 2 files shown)


itowlson (approved):
Looks good, thanks!

crates/plugins/src/manifest.rs
```diff
@@ -38,6 +39,19 @@ impl PluginManifest {
...
+    pub fn description(&self) -> Option<&str> {
+        self.description.as_deref()
+    }
+
+    pub fn homepage_url(&self) -> Option<Url> {
+        self.homepage
+            .as_deref()
+            .into_iter()
+            .flat_map(Url::parse)
+            .next()
```
rylev, line 52: This might be a bit easier to read:

```suggestion
        Url::parse(&self.homepage?).ok()
```

cardoso: Thanks!

crates/plugins/src/manifest.rs
```diff
@@ -38,6 +39,15 @@ impl PluginManifest {
     pub fn license(&self) -> &str {
         self.license.as_ref()
     }
+
+    pub fn description(&self) -> Option<&str> {
+        self.description.as_deref()
+    }
+
+    pub fn homepage_url(&self) -> Option<Url> {
+        Url::parse(&self.homepage?).ok()
```
cardoso, line 48:
```suggestion
        Url::parse(&self.homepage).ok()
```

crates/plugins/src/manifest.rs
```diff
@@ -38,6 +39,15 @@ impl PluginManifest {
     pub fn license(&self) -> &str {
         self.license.as_ref()
     }
+
+    pub fn description(&self) -> Option<&str> {
+        self.description.as_deref()
+    }
+
+    pub fn homepage_url(&self) -> Option<Url> {
+        Url::parse(&self.homepage).ok()
```
rylev, line 48: Looks like the `?` is missing:

```suggestion
        Url::parse(&self.homepage?).ok()
```

cardoso: Will need to solve this locally. With `?` it was also failing.

```
error[E0507]: cannot move out of `self.homepage` which is behind a shared reference
  --> crates/plugins/src/manifest.rs:48:21
   |
48 |         Url::parse(&self.homepage?).ok()
   |                     ^^^^^^^^^^^^^-
   |                     |
   |                     `self.homepage` moved due to this method call
   |                     move occurs because `self.homepage` has type `std::option::Option<std::string::String>`, which does not implement the `Copy` trait
   |
note: `std::ops::Try::branch` takes ownership of the receiver `self`, which moves `self.homepage`
  --> /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/ops/try_trait.rs:2[18](https://github.com/fermyon/spin/actions/runs/4979090393/jobs/8910255584#step:5:19):15
help: you can `clone` the value and consume it, but this might not be your desired behavior
   |
48 |         Url::parse(&clone().self.homepage?).ok()
   |                     ++++++++
```

cardoso: Just need good old `as_deref`.

src/commands/plugins.rs
```diff
@@ -467,6 +467,17 @@ async fn try_install(
     if continue_to_install(manifest, package, yes_to_all)? {
         let installed = manager.install(manifest, package).await?;
         println!("Plugin '{installed}' was installed successfully!");
+
+        if let Some(description) = manifest.description() {
+            println!("\nDescription:");
```
kate-goldenring, line 472: Is there a reason we want this additional newline added to a `println` statement? I'm assuming it is a stylistic choice

cardoso: Without some additional line breaks it disappears with the rest of the output. A mild improvement though. I believe #1481 will address this without needing the line breaks.

kate-goldenring (approved):

PR #1486, opened 2023-05-15
Validate components are not granted access to undefined k/v stores

itowlson:
```
ivan@hecate:~/testing/kvmadness$ spin up
Error: host component error: Component kvmadness is granted access to key-value store 'bibblybobbly', which is not defined

Caused by:
    Component kvmadness is granted access to key-value store 'bibblybobbly', which is not defined
Error: exit status: 1
ivan@hecate:~/testing/kvmadness$ spin up --runtime-config-file runtime-config.toml
Storing default key-value data to ".spin/my_super_duper_store.db"

Serving http://127.0.0.1:3000
Available Routes:
  kvmadness: http://127.0.0.1:3000 (wildcard)
```

(I'm not mad keen on the "host component error" string or the duplicated text, but this is a "should rarely happen" error so I'm wary of changing things that might break existing error paths.)

### DIFF
--- a/crates/key-value/src/host_component.rs
+++ b/crates/key-value/src/host_component.rs
@@ -1,4 +1,5 @@
-use crate::{KeyValueDispatch, StoreManager};
+use crate::{KeyValueDispatch, StoreManager, KEY_VALUE_STORES_KEY};
+use anyhow::anyhow;
 use spin_app::{AppComponent, DynamicHostComponent};
 use spin_core::HostComponent;
 use std::sync::Arc;
@@ -65,5 +66,39 @@ impl DynamicHostComponent for KeyValueComponent {
         );
 
         Ok(())
+    }
+
+    fn validate_app(&self, app: &spin_app::App) -> anyhow::Result<()> {
+        let mut errors = vec![];
+
+        for component in app.components() {
+            let store_manager = self.manager.get(&component);
+            for allowed in component
+                .get_metadata(KEY_VALUE_STORES_KEY)?
+                .unwrap_or_default()
+            {
+                if !store_manager.is_defined(&allowed) {
+                    let err = format!("- Component {} uses store '{allowed}'", component.id());
+                    errors.push(err);
+                }
+            }
+        }
+
+        if errors.is_empty() {
+            Ok(())
+        } else {
+            let prologue = vec![
+                "One or more components use key-value stores which are not defined.",
+                "Check the spelling, or pass a runtime configuration file that defines these stores.",
+                "See https://developer.fermyon.com/spin/dynamic-configuration#key-value-store-runtime-configuration",
+                "Details:",
+            ];
+            let lines: Vec<_> = prologue
+                .into_iter()
+                .map(|s| s.to_owned())
+                .chain(errors)
+                .collect();
+            Err(anyhow!(lines.join("\n")))
+        }
     }
 }
--- a/tests/testcases/key-value-undefined-store/spin.toml
+++ b/tests/testcases/key-value-undefined-store/spin.toml
@@ -1,0 +1,16 @@
+spin_version = "1"
+authors = ["Fermyon Engineering <engineering@fermyon.com>"]
+description = "A simple application that requests access to an undefined key/value store."
+name = "key-value"
+trigger = {type = "http", base = "/test"}
+version = "1.0.0"
+
+[[component]]
+id = "hello"
+# intentionally use undefined ~words~ stores
+key_value_stores = ["default", "anaspeptic", "pericombobulations"]
+source = "target/wasm32-wasi/release/key_value_undefined_store.wasm"
+[component.trigger]
+route = "/..."
+[component.build]
+command = "cargo build --release --target wasm32-wasi"
--- a/tests/testcases/key-value-undefined-store/src/lib.rs
+++ b/tests/testcases/key-value-undefined-store/src/lib.rs
@@ -1,0 +1,61 @@
+use anyhow::{ensure, Result};
+use itertools::sorted;
+use spin_sdk::{
+    http::{Request, Response},
+    http_component,
+    key_value::{Error, Store},
+};
+
+#[http_component]
+fn handle_request(req: Request) -> Result<Response> {
+    // TODO: once we allow users to pass non-default stores, test that opening
+    // an allowed-but-non-existent one returns Error::NoSuchStore
+    ensure!(matches!(Store::open("forbidden"), Err(Error::AccessDenied)));
+
+    let query = req.uri().query().expect("Should have a testkey query string");
+    let query: std::collections::HashMap::<String, String> = serde_qs::from_str(query)?;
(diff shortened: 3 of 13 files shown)


tests/testcases/key-value-undefined-store/spin.toml
```diff
@@ -0,0 +1,16 @@
+spin_version = "1"
+authors = ["Fermyon Engineering <engineering@fermyon.com>"]
+description = "A simple application that requests access to an undefined key/value store."
+name = "key-value"
+trigger = {type = "http", base = "/test"}
+version = "1.0.0"
+
+[[component]]
+id = "hello"
+# intntionally use undefined ~words~ stores
```
rylev, line 10:
```suggestion
# intentionally use undefined ~words~ stores
```

rylev (commented):
I think it might be somewhat confusing to users what it means to "define" a key value store. Perhaps we should more directly point to that fact that they are not creating a runtime config that configures a non-default store?

crates/key-value/src/host_component.rs
```diff
@@ -66,4 +67,27 @@ impl DynamicHostComponent for KeyValueComponent {
...
+        let mut errors = vec![];
+
+        for component in app.components() {
+            let store_manager = self.manager.get(&component);
+            for allowed in component
+                .get_metadata(KEY_VALUE_STORES_KEY)?
+                .unwrap_or_default()
+            {
+                if !store_manager.is_defined(&allowed) {
+                    let err = format!("Component {} is granted access to key-value store '{allowed}', which is not defined", component.id());
```
lann, line 81: I'm not sure that we need to frame `key_value_stores` in terms of access control. Maybe something like:

```suggestion
                    let err = format!("Component {} uses the key-value store '{allowed}', which is not defined", component.id());
```

tests/testcases/mod.rs
```diff
@@ -68,6 +68,45 @@ pub mod all {
...
+                },
+            )
+            .build()
+            .unwrap();
+
+        let e = tc.run(controller).await.unwrap_err();
+        let err_text = format!("{e:#}");
+
+        let expected1 = "Component hello is granted access to key-value store 'anaspeptic', which is not defined";
+        let expected2 = "Component hello is granted access to key-value store 'pericombobulations', which is not defined";
```
lann, line 98: If we want to test for specific UI text we should consider "golden file" testing which semi-automates updating this kind of assertion: https://crates.io/crates/goldenfile

itowlson: Agree but I propose to declare this out of scope for this PR.

itowlson:
@rylev I tried adding more pointers and I like it _but_ it really magnifies the effect of the doubling printing:

```
ivan@hecate:~/testing/kvmadness$ spin up
Error: host component error: One or more components use key-value stores which are not defined.
See https://developer.fermyon.com/spin/dynamic-configuration#key-value-store-runtime-configuration
Details:
- Component kvmadness uses store 'bibblybobbly'

Caused by:
    One or more components use key-value stores which are not defined.
    See https://developer.fermyon.com/spin/dynamic-configuration#key-value-store-runtime-configuration
    Details:
    - Component kvmadness uses store 'bibblybobbly'
Error: exit status: 1
```

I'll see if I can find what's doing this; if it's not an easy and obviously-safe fix, then maybe we just accept it and record a separate issue - again, not something people should run into too often!

itowlson:
Okay I am not sure how to fix the double printing.  `validate_app` returns an `anyhow::Error` which the caller wraps in a `Error::HostComponentError` to conform to the return type of `load_app`.  `HostComponentError` is declared as:

```
    #[error("host component error: {0:#}")]
    HostComponentError(#[source] anyhow::Error),
```

If I remove the `#[source]` from the attribute then I get "Error: host component error: One or more components use..." which is fine but I suspect it loses context for other, nastier, things that get turned into HCEs...  If I remove the `{0:#}` from the attribute, it probably preserves context better (?), but the printout is now "Error: host component error [blank line] Caused by: One or more components use..."  @lann is either of these two changes acceptable and if so do you have a recommendation?  Or should I wrap this error differently so as to avoid the problem altogether?

itowlson:
Or we could create a new `ValidationError` case and declare it as `#[error(transparent)]` so it gets passed through without the "host component error" bit.

tests/testcases/key-value-undefined-store/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    let init_key = query.get("testkey").expect("Should have a testkey query string");
+    let init_val = query.get("testval").expect("Should have a testval query string");
+
+    let store = Store::open_default()?;
+
+    store.delete("bar")?;
+
+    ensure!(!store.exists("bar")?);
+
+    ensure!(matches!(store.get("bar"), Err(Error::NoSuchKey)));
```
lann, line 26: I personally find it easier to read long tests like this if assertions are grouped with the side-effect-inducing call they are testing, e.g.:

```suggestion
    store.delete("bar")?;
    ensure!(!store.exists("bar")?);
    ensure!(matches!(store.get("bar"), Err(Error::NoSuchKey)));
```

tests/testcases/key-value-undefined-store/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    store.set("bar", b"baz")?;
+
+    ensure!(store.exists("bar")?);
+
+    ensure!(b"baz" as &[_] == &store.get("bar")?);
```
lann, line 32: `assert_eq!` gives more useful errors, e.g.

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `3`', src/main.rs:5:1
```

I _think_ it would also remove the need to cast `as &[_]` in this case, but my confidence in understanding Rust's coercion rules is NOT HIGH.

crates/key-value/src/host_component.rs
```diff
@@ -66,4 +67,38 @@ impl DynamicHostComponent for KeyValueComponent {
...
+                    errors.push(err);
+                }
+            }
+        }
+
+        if errors.is_empty() {
+            Ok(())
+        } else {
+            let prologue = vec![
+                "One or more components use key-value stores which are not defined.",
```
kate-goldenring, line 91: I wonder if we need to explicitly call out that only the "default" store is implicitly defined. However, this is explained in the docs that are linked so i think this works and if there is confusion we can always iterate on user feedback

kate-goldenring (approved):

tests/testcases/key-value-undefined-store/src/lib.rs
itowlson: actually none of this test body is needed because `spin up` will barf during validation and never actually run it so I just deleted it all.  Wake the heck up, Ivan

crates/key-value/src/host_component.rs
itowlson: Yeah, users should never see an error about the `default` store so hopefully they won't think they need to define it!.

Issue #1356, opened 2023-04-05
File mounting treats relative path as "file pattern"


rylev:
In my spin.toml, my component is mounting a single file:

```toml
files = ["foo/foo.txt"]
```

However spin treats this as a so called "file pattern" and does not allow `--direct-mount` to be passed to `spin up`. Instead I get the error:

```bash
this component cannot be run with `--direct-mount` because it uses file patterns
```

I would expect instead that users would be allowed to mount individual files.

dicej:
I don't think this is technically possible.  WASI preopened file descriptors need to be directories, AFAIK -- they can't be files.  So there's no way to mount a single host file directly without mounting the directory it contains (and thus everything in that directory).

dicej:
Per, https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-preopentype-variant, I don't believe this is possible, so closing.

PR #1488, opened 2023-05-15
Allow direct mounting directory 'pattern' (plus improve errors)

### LINKED ISSUE #1487: Confusing error when manifest `files` option includes path and `--direct-mount` is used
Issue #1356 was closed as #wont-fix. However, the error message given is still very confusing as the file path is not really a "file pattern". We should make the error more clear by stating that `--direct-mount` only supports mounting directories and not file paths *or* file patterns.

rylev:
Fixes #1487 and #1356. This PR does the following two things

## Direct mounting directory "patterns"
 
When the `files` option is a directory that exists and `--direct-mounts` is passed, we now allow mapping to happen (previously this was an error). Essentially this makes `files = ["some/directory"]` equivalent to `files = { source = "some/directory", destination = "some/directory" }`.

## Better error messages

The error message when using a file or glob pattern together with `--direct-mounts` was the following:

```
Error: Failed to prepare configuration

Caused by:
    this component cannot be run with `--direct-mount` because it uses file patterns
```

This was very confusing especially when using a file path and not glob pattern.

The error message now changes to the following in these various situations:

### Glob pattern

```
Error: Failed to prepare configuration

Caused by:
    0: `--direct-mount` flag cannot be used for this spin application
    1: only directories can be directly mounted but attempted to mount the glob pattern 'my-pattern*'
```

### Existing file path

```
Error: Failed to prepare configuration

Caused by:
    0: `--direct-mount` flag cannot be used for this spin application
    1: only directories can be directly mounted but attempted to mount the file path 'some/file.txt'
```

### Non-existing directory or file path

```
Error: Failed to prepare configuration

Caused by:
    0: `--direct-mount` flag cannot be used for this spin application
    1: only directories can be directly mounted but attempted to mount the non-existing path 'does/not/exist'
```

### DIFF
--- a/crates/loader/src/local/assets.rs
+++ b/crates/loader/src/local/assets.rs
@@ -75,9 +75,29 @@ fn prepare_component_with_direct_mounts(
                     .to_owned(),
                 host: placement.source.absolutize()?.into(),
             }),
-            RawFileMount::Pattern(_) => Err(anyhow!(
-                "this component cannot be run with `--direct-mount` because it uses file patterns"
-            )),
+            RawFileMount::Pattern(p) => {
+                let path = Path::new(p);
+                if path.is_dir() {
+                    Ok(DirectoryMount {
+                        guest: p.to_owned(),
+                        host: path.absolutize()?.into(),
+                    })
+                } else {
+                    let typ = if p.contains('*') {
+                        "glob pattern"
+                    } else if path.exists() {
+                        "file path"
+                    } else {
+                        "non-existing path"
+                    };
+
+                    Err(anyhow!(
+                        "only directories can be directly mounted but \
+                        attempted to mount the {typ} '{p}'"
+                    )
+                    .context("`--direct-mount` flag cannot be used for this spin application"))
+                }
+            }
         })
         .collect()
 }
--- a/crates/loader/src/local/assets.rs
+++ b/crates/loader/src/local/assets.rs
@@ -64,42 +64,45 @@ fn prepare_component_with_direct_mounts(
 ) -> Result<Vec<DirectoryMount>> {
     tracing::info!("directly mounting local asset directories into guest");
 
-    raw_mounts
-        .iter()
-        .map(|mount| match mount {
-            RawFileMount::Placement(placement) => Ok(DirectoryMount {
-                guest: placement
-                    .destination
-                    .to_str()
-                    .context("unable to parse mount destination as UTF-8")?
-                    .to_owned(),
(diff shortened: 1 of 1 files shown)


crates/loader/src/local/assets.rs
```diff
@@ -75,9 +75,29 @@ fn prepare_component_with_direct_mounts(
...
-            )),
+            RawFileMount::Pattern(p) => {
+                let path = Path::new(p);
+                if path.is_dir() {
+                    Ok(DirectoryMount {
+                        guest: p.to_owned(),
+                        host: path.absolutize()?.into(),
+                    })
+                } else {
+                    let typ = if p.contains('*') {
```
itowlson, line 86: SUPER NIT: should we check for `?` as well?

crates/loader/src/local/assets.rs
```diff
@@ -75,9 +75,29 @@ fn prepare_component_with_direct_mounts(
...
+                        "file path"
+                    } else {
+                        "non-existing path"
+                    };
+
+                    Err(anyhow!(
+                        "only directories can be directly mounted but \
+                        attempted to mount the {typ} '{p}'"
+                    )
+                    .context("`--direct-mount` flag cannot be used for this spin application"))
```
itowlson, line 98: Probably don't need to mention Spin here?

```suggestion
                    .context("`--direct-mount` flag cannot be used for this application"))
```

but if you do then please capitalise it.

crates/loader/src/local/assets.rs
itowlson, line 86: This is an extremely complicated clause, especially at the "an else inside a match inside a map" level of indentation - by the time I get to the end of it I have long since forgotten what I'm doing.  Consider pulling out a function e.g. `direct_mount_pattern_error(...)`, or even (maybe better) pulling out the whole match arm (starting on line 79) into `direct_mount_for_pattern()`.

Or even pulling out the whole map body...

itowlson (changes requested):
Thanks for this - love the improvement, just some nits on the implementation.

crates/loader/src/local/assets.rs
rylev: I pulled out the map body into its own function. I hope that that helps!

itowlson (approved):

crates/loader/src/local/assets.rs
itowlson: IT HELPS SO MUCH

PR #1492, opened 2023-05-15
feat(plugins): provide additional build info to plugins

cardoso:
I'm building a plugin that manages spin versions. It would be useful to understand where the build comes from, so it can be upgraded/downgraded/recovered safely.

I had to update vergen to version 8 and did some cleaning up on the way.

The additional info is sent to the plugin child process envs:

```
SPIN_BIN_PATH=/Users/cardoso/.cargo/bin/spin
SPIN_BRANCH=feat/more_build_info
SPIN_BUILD_DATE=2023-05-15
SPIN_COMMIT_DATE=2023-05-15
SPIN_COMMIT_SHA=49fb11b
SPIN_DEBUG=false
SPIN_TARGET_TRIPLE=aarch64-apple-darwin
SPIN_VERSION=1.2.0-pre0
SPIN_VERSION_MAJOR=1
SPIN_VERSION_MINOR=2
SPIN_VERSION_PRE=pre0
```

### DIFF
--- a/src/bin/spin.rs
+++ b/src/bin/spin.rs
@@ -2,6 +2,7 @@ use anyhow::Error;
 use clap::{CommandFactory, Parser, Subcommand};
 use is_terminal::IsTerminal;
 use lazy_static::lazy_static;
+use spin_cli::build_info::*;
 use spin_cli::commands::{
     build::BuildCommand,
     cloud::CloudCommands,
@@ -46,7 +47,7 @@ fn version() -> &'static str {
 #[derive(Parser)]
 #[clap(
     name = "spin",
-    version = version(),
+    version = version()
 )]
 enum SpinApp {
     #[clap(subcommand, alias = "template")]
@@ -105,10 +106,5 @@ impl SpinApp {
 
 /// Returns build information, similar to: 0.1.0 (2be4034 2022-03-31).
 fn build_info() -> String {
-    format!(
-        "{} ({} {})",
-        env!("VERGEN_BUILD_SEMVER"),
-        env!("VERGEN_GIT_SHA_SHORT"),
-        env!("VERGEN_GIT_COMMIT_DATE")
-    )
+    format!("{SPIN_VERSION} ({SPIN_COMMIT_SHA} {SPIN_COMMIT_DATE})")
 }
--- a/src/build_info.rs
+++ b/src/build_info.rs
@@ -1,0 +1,24 @@
+//! Build information for the Spin CLI.
+
+/// The version of the Spin CLI.
+pub const SPIN_VERSION: &str = env!("CARGO_PKG_VERSION");
+/// The major version of the Spin CLI.
+pub const SPIN_VERSION_MAJOR: &str = env!("CARGO_PKG_VERSION_MAJOR");
+/// The minor version of the Spin CLI.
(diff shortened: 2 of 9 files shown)


src/bin/spin.rs
```diff
@@ -35,18 +36,24 @@ async fn main() -> Result<(), Error> {
...
+fn version_verbose() -> &'static str {
+    &VERSION_VERBOSE
+}
+
 /// The Spin CLI
 #[derive(Parser)]
 #[clap(
     name = "spin",
     version = version(),
+    long_version = version_verbose(),
```
itowlson, line 56: We've been leading people to use `spin --version` in the docs, and this will now make that print a significant amount of extra info that most people shouldn't care about (and it had quite a bit of that in it already!).  Is there another way we could surface the long form?

(I am also wondering if, longer term, we should default it to _just_ showing the version, and having the commit SHA and date appear only in the long form.  But that is definitely out of scope for this PR!)

itowlson (commented):
This looks great (and I'm excited to see your plugin - sounds super useful).  My only reservation is the change to the existing behaviour of `spin --version`.  Our docs and learning materials do currently guide people to `--version` and I'm wary of confronting new users with this much detail.

My vote on this would be to (as you mention) follow Cargo and have `-V/--version` show the short form and `-V -v` show the long form.  But other folks may be less change-averse - happy to wait for others to weigh in.

And thanks again - I apologise for getting hung up on what may seem like a niggle - the content of the PR looks great.

itowlson:
+ @michelleN for thoughts on the UI question

src/build_info.rs
```diff
@@ -0,0 +1,24 @@
+//! Build information for the Spin CLI.
+
+/// The version of the Spin CLI.
+pub const SPIN_VERSION: &str = env!("CARGO_PKG_VERSION");
```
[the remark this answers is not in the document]
cardoso: Yes, as I understand VERGEN_BUILD_SEMVER went away. Ironic since it was the primary purpose of the library :)

src/bin/spin.rs
cardoso: I would prefer -V -v as well, but I guess introducing a verbose global variable could create further discussions and fighting with clap v3. If changing long_version output is too disruptive, I can explore -V -v or leave it out for now.

cardoso:
@itowlson I opted to reduce the scope here to just sending the env vars to the plugin and make a separate PR for the CLI eventually

itowlson (approved):
This looks great!  100% agree on reducing scope: as you say, this unblocks the plugin scenario, and we can always iterate on the UI.  Thanks!

kate-goldenring (approved):
@cardoso this is great providing plugins with more information about Spin. I created an issue to also update developer docs to call out what env vars are passed to plugins https://github.com/fermyon/developer/issues/632.

karthik2804 (approved):
This is great! Thanks!

itowlson:
@cardoso This just missed the cutoff for Spin 1.2, so for now you will need to use canary or HEAD for your plugin development.  I can't commit on when the next release will be; we are tracking a _roughly_ monthly schedule at the moment, but absolutely no promises!

cardoso:
No problem @itowlson . I still need to submit more PRs to have enough features for the plugin to be solid. Hopefully 1.3 will suffice.

Issue #641, opened 2022-07-14
installable via apt


bacongobbler:
As a user, I'd love to install the spin CLI similar to how I'd install other packages.

Something similar to

```
$ curl -fsSL https://apt.releases.fermyon.dev/gpg | sudo apt-key add -
$ sudo apt-add-repository "deb [arch=amd64] https://apt.releases.fermyon.dev $(lsb_release -cs) main"
$ sudo apt-get update && sudo apt-get install spin
```

(note: this is a very similar experience for HashiCorp's installation instructions for `nomad`, `consul`, `vault`, etc.)

radu-matei:
I'd expand the scope of this and suggest we look at supporting package managers for macOS and Window as well.

neodon:
I look for any kind of managed way to install something before sighing and running an installer script. Having only this option could be a deterrent for some types of devs to try out Fermyon. I realize it's a lot more work to define and generate other installation methods, however, so I understand if it's not time for this just yet.

Also, I'll just mention [Homebrew on Linux](https://docs.brew.sh/Homebrew-on-Linux) here since it doesn't seem well-known. Or people automatically treat it as heretical without learning more. I've been using it for a few years and it's been pretty awesome, especially when I want to match up versions of software on my Mac and Linux PC. It uses the same package sources for both, so it would provide an option to release often on both--if you can get people to try it.

radu-matei:
Hi, @neodon, thanks for the comment!
I agree, which is why I think this issue is something we want to prioritize really soon.
Particularly for the default package managers on Debian, macOS and Windows.

kate-goldenring:
The [`spin` software validation tool](https://spinroot.com/spin/whatispin.html) has claimed the `spin` package name in many OS installers (`brew` , `apt` , `chocolatey` at minimum). 

It seems like we may want to prioritize `brew`, `apt`, and `chocolatey`, for all of which `spin` is taken. 

Wanted to kick off some brainstorming on alternative package names:
- `fermyon-spin`
- `spin-up`

Also, for brew, I am not familiar with how difficult it is to create a `bottle` that can be directly installed with `brew install` vs creating a `fermyon` tap that needs to be installed first. I am assuming creating a `fermyon` tap is easier, creating an installation experience similar to [Nomad's](https://developer.hashicorp.com/nomad/docs/install).

bacongobbler:
[...]

PR #1494, opened 2023-05-15
Conditionally use Homebrew location for Spin plugins and templates

kate-goldenring:
The current experience for installing spin is to either use the install script or build from source. Users should be able to install Spin using their favorite package manager, such as Homebrew. https://github.com/fermyon/spin/issues/641

I am working on creating a `fermyon-spin` Homebrew formula. Homebrew is sandboxed such that you can only install into a subdirectory of `$HOMEBREW_PREFIX` ([`/opt/homebrew` on ARM macOs and `/usr/local` on Intel macOs](https://github.com/Homebrew/install/blob/fc8acb0828f89f8aa83162000db1b49de71fa5d8/install.sh#L132) ). The formula installs default plugins and templates to the `$HOMEBREW_PREFIX/etc/fermyon-spin` directory. 

The plugin and templates systems look at `$XDG_DATA_HOME` or more specifically `$HOME/Library/Application Support` for macOS. This updates them to look for the existence of a brew installation of Spin and if exists, use the `$HOMEBREW_PREFIX/etc/fermyon-spin` directory for plugins and templates.

If a user has a brew installation of spin and another installation or local execution, there could be duplication of plugins and templates, with each executable using their expected directory

### DIFF
--- a/crates/plugins/src/store.rs
+++ b/crates/plugins/src/store.rs
@@ -25,10 +25,12 @@ impl PluginStore {
     }
 
     pub fn try_default() -> Result<Self> {
-        if let Ok(test_dir) = std::env::var("TEST_PLUGINS_DIRECTORY") {
-            return Ok(Self::new(PathBuf::from(test_dir)));
-        }
-        Ok(Self::new(default_data_dir()?.join("plugins")))
+        let data_dir = if let Ok(test_dir) = std::env::var("TEST_PLUGINS_DIRECTORY") {
+            PathBuf::from(test_dir).join("spin")
+        } else {
+            default_data_dir()?
+        };
+        Ok(Self::new(data_dir.join("plugins")))
     }
 
     /// Gets the path to where Spin plugin are installed.
--- a/crates/plugins/src/store.rs
+++ b/crates/plugins/src/store.rs
@@ -24,6 +24,17 @@ impl PluginStore {
     }
 
     pub fn try_default() -> Result<Self> {
+        if let Ok(brew_prefix) = std::env::var("HOMEBREW_PREFIX") {
+            let plugins_dir = Path::new(&brew_prefix)
+                .join("var")
+                .join("spin")
+                .join("plugins");
+
+            if plugins_dir.is_dir() {
+                return Ok(Self::new(plugins_dir));
+                // TODO: check if they also have plugins in non-brew default dir and warn
+            }
+        }
         let data_dir = match std::env::var("TEST_PLUGINS_DIRECTORY") {
             Ok(test_dir) => PathBuf::from(test_dir),
             Err(_) => dirs::data_local_dir()
--- a/crates/plugins/src/store.rs
+++ b/crates/plugins/src/store.rs
@@ -1,5 +1,6 @@
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use flate2::read::GzDecoder;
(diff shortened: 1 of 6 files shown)


lann:
I don't know much about homebrew but this doesn't seem right to me. If I install e.g. iTerm from homebrew and spin from source, will spin see `HOMEBREW_PREFIX`?

crates/plugins/src/store.rs
```diff
@@ -24,6 +24,15 @@ impl PluginStore {
     }
 
     pub fn try_default() -> Result<Self> {
+        if let Ok(brew_prefix) = std::env::var("HOMEBREW_PREFIX") {
+            let plugins_dir = Path::new(&brew_prefix).join("var").join("spin").join("plugins");
+
+            if plugins_dir.exists() && plugins_dir.is_dir() {
```
lann, line 30: `is_dir` implies `exists`
```suggestion
            if plugins_dir.is_dir() {
```

kate-goldenring:
> Ohhh so the formula creates those dirs? I have no opinion then 😅

Yes exactly. In the formula `install` step, those directories are created:
```
  def spin_datadir
      var/"spin"
  end
  
  def install
    system "#{Formula["rustup-init"].bin}/rustup-init", "-qy", "--no-modify-path"
    ENV.prepend_path "PATH", HOMEBREW_CACHE/"cargo_cache/bin"
    system "rustup", "target", "add", "wasm32-wasi", "wasm32-unknown-unknown"
    system "cargo", "install", *std_cargo_args

    (spin_datadir/"templates").mkpath
    (spin_datadir/"plugins").mkpath

    system "#{bin}/spin", "templates", "install", "--git", "https://github.com/fermyon/spin", "--upgrade"
    system "#{bin}/spin", "templates", "install", "--git", "https://github.com/fermyon/spin-python-sdk", "--upgrade"
    system "#{bin}/spin", "templates", "install", "--git", "https://github.com/fermyon/spin-js-sdk", "--upgrade"
    system "#{bin}/spin", "plugins", "update"
    system "#{bin}/spin", "plugins", "install", "js2wasm", "--yes"
    system "#{bin}/spin", "plugins", "install", "py2wasm", "--yes"
  end
```

crates/plugins/src/store.rs
```diff
@@ -24,6 +24,17 @@ impl PluginStore {
     }
 
     pub fn try_default() -> Result<Self> {
+        if let Ok(brew_prefix) = std::env::var("HOMEBREW_PREFIX") {
```
itowlson, line 27: The check for `TEST_PLUGINS_DIRECTORY` should come first, so that tests running on a Mac with Homebrew don't party on the user's real plugins directory.

itowlson, line 27: This may be premature, but is it worth pulling the "package manager stanza" out into a separate function?  It almost overwhelms the main flow here, and doesn't really state intent.  Maybe even pull out the whole "figure out the right directory" logic to make early returns easier e.g.

```
fn try_default() {
  Self::new(default_dir()?))
}

fn default_dir() {
  if let Ok(test_dir) = std::env::var(...) {
    return test_dir;
  }
  if let Some(override) = package_manager_plugins_dir() {
    return override;
  }
  // XDG logic
}
```

I'm thinking out loud here - feel free to tell me this is a bad idea!

kate-goldenring: I like this. It's nice and clean

PR #1496, opened 2023-05-16
Remove unhelpful error when preparing spin configuration

rylev:
The error "Failed to prepare configuration" does not tell the user anything as they generally have no idea what "preparing" configuration actually means. The errors returned from the `core` function already provide all the context the user needs to address their issue, and so the "Failed to prepare configuration" error is just noise.

(As a bonus to this PR, I removed some unnecessary allocation)


itowlson (approved):
turbo-approved for getting rid of a reference to 'configuration'

PR #1497, opened 2023-05-16
Colorify errors

rylev:
This is the first step for #1481. Errors are now proceeded by "Error" in bold red text. This should happen for all errors and the user's browser settings such as `NO_COLOR` are respected.

image


crates/build/src/lib.rs
```diff
@@ -29,7 +29,7 @@ pub async fn build(manifest_file: &Path) -> Result<()> {
         .map(|c| build_component(c, &app_dir))
         .collect::<Result<Vec<_>, _>>()?;
 
-    println!("Successfully ran the build command for the Spin components.");
+    println!("Finished the build command for the Spin components.");
```
itowlson, line 32: Maybe move wording tweaks into another PR?

crates/colors/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+        let mut out = $crate::ColorText::stderr($color);
+        let _ = write!(out, $($arg)*);
+        drop(out); // Reset colors
+    };
+}
+
+pub mod colors {
+    use termcolor::{Color, ColorSpec};
+
+    pub fn red() -> ColorSpec {
```
itowlson, line 116: These names are physical rather than semantic descriptions; should we therefore mention that they also include bold?

src/commands/external.rs
```diff
@@ -50,7 +51,7 @@ pub async fn execute_external_subcommand(
         }
         Err(Error::NotFound(e)) => {
             tracing::debug!("Tried to resolve {plugin_name} to plugin, got {e}");
-            eprintln!("Error: '{plugin_name}' is not a known Spin command. See spin --help.\n");
+            error!("'{plugin_name}' is not a known Spin command. See spin --help.\n");
```
itowlson, line 54: Wondering if `print_error!` might make it clearer that this is a print operation.  This looks too much like `bail!` for me to readily infer that this is a benign piece of user interaction.

rylev: This is the same as `tracing::error` and `log::error`. How about having the convention of always using this in a qualified way `colors::error!` is hopefully less prone to confusion?

itowlson: That would be better though in that case I'd suggest a namespace like `console` or `terminal` - `tracing` and `log` directly imply output in a way that `colors` (to me) doesn't.

Which might be better because the formatting conventions could come to involve icon-based highlighting or layout functions as well, although I realise this is speculative!

itowlson (approved):
Happy with this (whether or not you decide to change the namespace, and we can always iterate on that).  Thanks!

src/commands/external.rs
rylev: I've renamed the dependency to `terminal`.

Issue #1498, opened 2023-05-16
sdk: Keep `#[http_component]` functions in their apparent scope


lann:
Currently the macro moves these into the generated `SpinHttp` impl, which means they aren't accessible for e.g. testing.

Could we keep them in place and reference from the impl with `self::...`?

PR #1499, opened 2023-05-16
Keep SDK annotated functions in scope

fibonacci1729:


fibonacci1729:
Closes #1498

lann (approved):

sdk/rust/macro/src/lib.rs
```diff
@@ -15,12 +15,12 @@ pub fn http_component(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
 
+        #func
+
         impl spin_http::SpinHttp for SpinHttp {
             // Implement the `handler` entrypoint for Spin HTTP components.
             fn handle_http_request(req: spin_http::Request) -> spin_http::Response {
-                #func
-
-                match #func_name(req.try_into().expect("cannot convert from Spin HTTP request")) {
+                match self::#func_name(req.try_into().expect("cannot convert from Spin HTTP request")) {
```
rylev, line 23: Is the `self` annotation necessary? I'm not sure why it would be

lann: It lets you name your function `req`.

rylev: Ah I didn't think about local variables... :+1:

lann: Really we should probably be even _more_ careful with our item paths in here with e.g. `::spin_http`

fibonacci1729: `::spin_http` won't work; as of rust 2018 paths prefixed with `::` must be followed by the name of a crate. We could however say `crate::spin_http` to prevent a collision with a dependency named `spin_http`.

ref: https://doc.rust-lang.org/reference/paths.html#path-qualifiers

lann: Uh right; I saw `spin_` and thought "crate" but obviously that's just the local module.

lann: The way the SDK works currently I think we actually depend on `spin_http` being in scope for handler functions.

lann: Oh you're right; I forgot that we always convert to/from `http` types for the SDK.

fibonacci1729: @rylev great point; the dunder possibility slipped my brain and i think that is a better choice in this case.

fibonacci1729: Unfortunately, it doesn't seem like the `standalone` option is surfaced in the version of the `wit-bindgen` macro we are using however alternatively we could do something like:

```rust
mod __spin_http {
        wit_bindgen_rust::export!({src["spin_http"]: #HTTP_COMPONENT_WIT});
        pub use spin_http::*;
}

struct SpinHttp;

impl __spin_http::SpinHttp for SpinHttp {
    ...
}
```

Does this make sense?

fibonacci1729: Errr actually i am remembering why this won't work.

The code `wit-bindgen` generates expects a type `SpinHttp` in the namespace outside of the generated module, i.e. (from the generated module):

```
...
let result = <super::SpinHttp as SpinHttp>::handle_http_request
```

The SDK is currently structured the way it is because the version of `wit-bindgen` we are using dictates a certain module structure. However, i still think `crate::...`, albeit less idiomatic, mitigates any collision.

rylev: `crate::` assumes that the generated code is at the top of the crate. This would prevent using the macro anywhere but at the top of the crate namespace.

lann (approved):

PR #1500, opened 2023-05-17
Tolerate component routes not starting with '/'

### LINKED ISSUE #1493: Forgotten `/` before route causes things to not work with no error
In your spin.toml, you can have a route for component like `route = "foo"`  that will silently just not work. The correct syntax for a route includes a proceeding "/" but not providing one does not produce an error - things just unexpectedly don't work.

itowlson:
Fixes #1493.

An alternative solution would be to fail validation if any component route didn't start with '/'.  That would be safer in the sense of not committing us to supporting "wrong" manifests, but would risk a user experience of "stupid computer, you know what's wrong so why do you need _me_ to fix it."  Happy to switch to that if people prefer.

### DIFF
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -165,11 +165,18 @@ impl RoutePattern {
 
     /// Sanitizes the base and path and return a formed path.
     pub fn sanitize_with_base<S: Into<String>>(base: S, path: S) -> String {
-        format!(
-            "{}{}",
-            Self::sanitize(base.into()),
-            Self::sanitize(path.into())
-        )
+        let path = Self::absolutize(path);
+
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):
LGTM. I think I prefer the magic correction here as opposed to a validation error but I could lean the other way if others feel strongly.

rylev (approved):
I think I'm ok with us not requiring the user explicitly put a "/" before the route, but there are a few more cases that I think we should think about:

* "..."
* ""

rylev (approved):
I think we should merge this as is. We can always improve in the future if need be.

itowlson:
@rylev awesome, thanks!

PR #1502, opened 2023-05-17
E2E Error Output Tests

rylev:
This adds the start of a testing of the error output from spin. The end vision is an easy way for developers to add tests for some sort of broken Spin experience and test the actual stdout/stderr output against some expected output. 

We're not quite there, but this PR makes significant changes to the e2e testing framework to allow for test Spin apps to fail with (necessarily) failing the test. I've added one test as well to give an idea of how these tests might work in the future. However, I expect the experience of adding new tests to improve with time as we bake in more convention on how to write such tests.

Note: I did quite a bit of clean up along the way which might make this somewhat difficult to review. I apologize for this, but I hope the intention is still clear. Please let me know if you need more explanation of the changes, and I can write up a more thorough summary.

### DIFF
--- a/src/bin/spin.rs
+++ b/src/bin/spin.rs
@@ -49,7 +49,7 @@ fn print_error_chain(err: anyhow::Error) {
         eprintln!("\nCaused by:");
         for (i, err) in err.chain().skip(1).enumerate() {
             if is_multiple {
-                eprintln!("{i:>5}: {}", err)
+                eprintln!("{i:>4}: {}", err)
             } else {
                 eprintln!("      {}", err)
             }
--- a/tests/testcases/error/error.txt
+++ b/tests/testcases/error/error.txt
@@ -1,4 +1,4 @@
 Error: Cannot read spin.toml manifest from "$APPDIR/spin.toml"
 
 Caused by:
-    missing field `component` at line 1 column 1
+      missing field `component` at line 1 column 1
--- a/tests/testcases/mod.rs
+++ b/tests/testcases/mod.rs
@@ -69,8 +69,20 @@ pub async fn key_value_validation_works(controller: &dyn Controller) {
     async fn checks(
         _: AppMetadata,
         _: Option<Pin<Box<dyn AsyncBufRead>>>,
-        _: Option<Pin<Box<dyn AsyncBufRead>>>,
+        stderr: Option<Pin<Box<dyn AsyncBufRead>>>,
     ) -> Result<()> {
+        let err_text = utils::get_output(stderr).await.unwrap();
+        let expected1 = "Component hello uses store 'anaspeptic'";
+        let expected2 = "Component hello uses store 'pericombobulations'";
+
+        assert!(
+            err_text.contains(expected1),
+            "Expected error containing '{expected1}' but got '{err_text}'"
+        );
+        assert!(
+            err_text.contains(expected2),
+            "Expected error containing '{expected2}' but got '{err_text}'"
+        );
         Ok(())
     }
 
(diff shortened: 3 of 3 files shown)


crates/e2e-testing/src/spin.rs
```diff
@@ -37,37 +35,33 @@ pub fn new_app<'a>(
...
-            vec!["spin", "plugin", "install", plugin, "--yes"],
-            None,
-            None,
-        )?;
+        output = utils::run(&["spin", "plugin", "install", plugin, "--yes"], None, None)?;
+        utils::assert_success(&output);
     }
 
-    //this ensure we have mutex lock until here
-    drop(x);
```
rylev, line 58: As far as I could tell this drop call is not necessary (the mutex lock will be dropped right after this at the end of the function.

crates/e2e-testing/src/spin.rs
```diff
@@ -113,12 +109,13 @@ pub fn registry_login(registry_url: &str, username: &str, password: &str) -> Res
...
+    Ok(output)
 }
 
 pub fn version() -> Result<String> {
-    match utils::run(vec!["spin", "--version"], None, None) {
-        Ok(output) => Ok(format!("{:#?}", std::str::from_utf8(&output.stdout)?)),
-        Err(err) => Err(err),
-    }
+    let output = utils::run(&["spin", "--version"], None, None)?;
+    utils::assert_success(&output);
```
rylev, line 119: We may eventually remove these assertions in favor of the top-level testing code asserting success. For now though we'll leave this here as every test assumes that these calls succeed.

e2e-tests-docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
   registry:
     image: ${REGISTRY_IMAGE:-registry:2}
     ports:
-      - "5000:5000"
+      - "5001:5000"
```
rylev, line 38: I made this change because macOS uses port 5000 for ControlCenter now...

crates/e2e-testing/src/spin.rs
rajatjindal: @itowlson suggested to use explicit drops for better readability.

crates/e2e-testing/src/spin.rs
rajatjindal: 👍

e2e-tests-docker-compose.yml
rajatjindal: we need to change this port in tests/testcases/mod.rs as we use it there

rajatjindal: or may be not. because tests run in docker context as well

crates/e2e-testing/src/utils.rs
```diff
@@ -1,63 +1,56 @@
...
+    cmd.arg(args.iter().map(AsRef::as_ref).collect::<Vec<_>>().join(" "));
     if let Some(envs) = envs {
         for (k, v) in envs {
             cmd.env(k, v);
         }
     }
 
-    let output = cmd.output()?;
-    let code = output.status.code().expect("should have status code");
+    println!("Running command: {cmd:?}");
```
rajatjindal, line 39: do we want to move this to debug instead of printing always?

crates/e2e-testing/src/spin.rs
rylev: Hmm... I understand that when the drop happens early, and you want to show that the lock has been given up before doing more work, but the drop will happen right after this... This strikes me as fairly unidiomatic.

crates/e2e-testing/src/utils.rs
rylev: I originally did this so that it was clear if the tests panicked due to the command failing what the command was (since the command itself is no longer printed in the panic message). I'm not against moving to `debug` but this might make it less obvious what went wrong when something did. Perhaps we leave it for now, and see how annoying it is before changing it?

rajatjindal (approved):
Lets 🚢 it.

rylev:
Annoyingly the broken test consistently works locally. Looks like for some reason there's some extra whitespace produced by `spin` when running on the CI server.

PR #1503, opened 2023-05-17
Canonicalize source path

rylev:
This allows users to have "~/some/path/../to/binary" as their source in their `spin.toml`.

### DIFF
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -235,11 +235,7 @@ fn canonicalize_and_absolutize(mut path: PathBuf, src: &Path) -> anyhow::Result<
             }
         }
     }
-    Ok(if path.is_absolute() {
-        path
-    } else {
-        src.join(path)
-    })
+    Ok(src.join(path))
 }
 
 /// A parsed URL source for a component module.
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -178,12 +178,7 @@ async fn core(
     let src = parent_dir(src)?;
     let source = match raw.source {
         config::RawModuleSource::FileReference(p) => {
-            let p = match p.is_absolute() {
-                true => p,
-                false => src.join(p),
-            };
-
-            ModuleSource::FileReference(p)
+            ModuleSource::FileReference(canonicalize_and_absolutize(p, &src))
         }
         config::RawModuleSource::Url(us) => {
             let source = UrlSource::new(&us)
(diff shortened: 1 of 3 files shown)


lann:
You'll have to rebase on main...sorry!

itowlson (approved):

radu-matei (approved):

lann:
Not sure how I feel about this feature; '~'-expansion is usually just a shell feature and while there are some applications that support it, it isn't really all that common ime. I would rather spell out `/home/lann/...` so that I'm more likely to catch it when reviewing a diff...

rylev:
Perhaps I should explain why I'd like to have this feature. My cargo target directory is always set to `.cargo_target` directory in my home directory. When I switch back and forth between Linux and macOS I have to constantly adjust the source path to fit the machine I'm on. With this feature, I wouldn't have to adjust since it's always in the same place relative to my home directory. 

Perhaps you could explain more about what you're worried about? Do we want to discourage source paths that aren't relative to spin application? We already support absolute paths, so I see this as just an extension of that.

lann:
It's partly about consistency. Do we support `~` just in source paths? Should we support `~<user>` too? Why just `~`-expansion and not e.g. `$HOME`?

lann:
But! Your use case seems reasonable to me and I'm definitely not vetoing this or anything.

rylev:
@lann I expanded this PR to also support ENV variables using the `shellexpand` crate. Unfortunately that crate does not support `~$USER` expansion, but I don't think that feature is used super often so perhaps that's fine (at least for now)?

lann:
:+1: The only other thing I would consider is whether we want to support `~` in runtime config for consistency here. I can't think of anywhere else it would apply.

rylev:
I opened https://github.com/fermyon/spin/issues/1526 to track that we do this consistently.

crates/loader/src/local/mod.rs
```diff
@@ -225,6 +220,28 @@ async fn core(
...
+                    e.var_name
+                ))
+            }
+        }
+    }
+    Ok(if path.is_absolute() {
+        path
+    } else {
+        src.join(path)
+    })
```
lann, line 242: I believe `Path::join` already does this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=09ffc9d54f48c56c5476c67d549c6ca8

```suggestion
    Ok(src.join(path))
```

lann (commented):
Sorry, had this review comment queued up but never submitted it...

lann (approved):

PR #1505, opened 2023-05-17
windows: Statically compile MSVC runtime

lann:
This should allow `spin` to be run on Windows without installing a MSVC runtime.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,2 +1,6 @@
 [net]
     git-fetch-with-cli = true
+
+# Statically compile MSVC runtime
+[target.'cfg(target_env = "msvc")']
+    rustflags = ["-Ctarget-feature=+crt-static"]


.cargo/config
```diff
@@ -1,2 +1,6 @@
 [net]
     git-fetch-with-cli = true
+
+# Statically compile MSVC runtime
+[target.'cfg(target_env = "msvc")']
+    rustflags = ["-Ctarget-feature=+crt-static"]
```
rylev, line 6: How much does this bloat the size of the binary? Are there any license restrictions that this imposes?

lann: Either "pretty much not at all" or it didn't actually work...

lann: Fwiw wasmtime does the same: https://github.com/bytecodealliance/wasmtime/blob/08b7c877931a3da4cb73bd3c0961fd475c0e8a60/.github/actions/binary-compatible-builds/main.js#L20

itowlson (approved):

rylev (approved):

itowlson:
@lann this has been sitting on approved for a while - is it okay to merge?

lann:
I was hoping the discord user who originally brought up this issue would test the fix but they seemingly never got around to it. A also spent a little time trying to wrangle a windows VM for this but ultimately gave up for boring linuxy reasons.

PR #1506, opened 2023-05-17
build: Fix --up long flag name

lann:
This was accidentally broken in an unrelated commit.


rylev (approved):

lann:
Bypassing CI because the fix is small, I tested manually, and this is blocking all other CI from passing.

PR #1507, opened 2023-05-17
go: Improve integration test failure

lann:
The integration test times out after 10 minutes if spin fails to start. This checks for spin exiting before the test is over.

### DIFF
--- a/sdk/go/integration_test.go
+++ b/sdk/go/integration_test.go
@@ -17,14 +17,18 @@ const spinBinary = "../../target/debug/spin"
 func retryGet(t *testing.T, url string) *http.Response {
 	t.Helper()
 
-	const maxTries = 120 // (10min/5sec)
+	const maxTries = 600 // (10min)
 	for i := 1; i < maxTries; i++ {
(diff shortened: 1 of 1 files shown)


sdk/go/integration_test.go
```diff
@@ -17,14 +17,18 @@ const spinBinary = "../../target/debug/spin"
 func retryGet(t *testing.T, url string) *http.Response {
 	t.Helper()
 
-	const maxTries = 120 // (10min/5sec)
+	const maxTries = 600 // (10min)
```
rylev, line 20: Why do we try for so long?

lann: :shrug:

https://github.com/fermyon/spin/blob/3f7c2fee0c22fe24ea4fc74fd516142b02936cce/sdk/go/integration_test.go#L44-L45

adamreese: The timeout is long because we need to wait for the build also.  TinyGo can take a bit of time in CI

adamreese (approved):
We might need to increase timeout again but LGTM

PR #1513, opened 2023-05-18
Rename doctor --file to --from for consistency

itowlson:
We've been using `--from` as the primary long form for `-f`.  It doesn't read as nicely for `doctor` as it does for `up` but I thought it was worth it to keep the flag name and text consistent throughout Spin.  Also renamed the field to be the same as other commands (again, this is purely cosmetic consistency, and we don't need to do it if we prefer the existing name).


lann (approved):
Yeah I picked `file` intentionally since the command doesn't make much sense for other sources, but I don't feel strongly about it.

PR #1514, opened 2023-05-18
Fix `spin doctor` spurious "missing Wasm file" if app is in another directory

itowlson:
I ran into this because I was playing with `spin doctor` using `spin doctor -f ~/testing/kvmadness` and it reported that my Wasm file was missing.  It looks like it currently probes for the file relative to the current directory rather than the app directory.

This fixes that, but it feels a bit verbose, like I am passing more stuff around more places than should be necessary.  @lann I'd be grateful for any insight on how you envisaged things like this being structured.  I saw that the `WasmDiagnostic` trait passes the `PatientApp` that includes the manifest file path, and originally used that to construct the absolute Wasm path in `diagnose_wasm`. 
 But it felt like that would end up getting repeated in multiple diagnostics so I moved it into the `PatientWasm` type.  Anyway I don't want to end up spoiling the design on my first outing _grin_

### DIFF
--- a/crates/doctor/src/wasm.rs
+++ b/crates/doctor/src/wasm.rs
@@ -1,44 +1,60 @@
 /// Diagnose missing Wasm sources.
 pub mod missing;
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use anyhow::Result;
 use async_trait::async_trait;
-use spin_loader::{local::config::RawComponentManifest, local::config::RawModuleSource};
+use spin_loader::{
+    local::config::RawComponentManifest,
+    local::{canonicalize_and_absolutize, config::RawModuleSource},
+};
 
 use crate::{Diagnosis, Diagnostic, PatientApp};
 
 /// PatientWasm represents a Wasm source to be checked for problems.
 #[derive(Debug)]
-pub struct PatientWasm(RawComponentManifest);
+pub struct PatientWasm {
+    app_dir: PathBuf,
+    component: RawComponentManifest,
+}
 
 #[allow(missing_docs)] // WIP
 impl PatientWasm {
-    pub fn component_id(&self) -> &str {
-        &self.0.id
+    fn new(app_dir: impl AsRef<Path>, component: RawComponentManifest) -> Self {
+        Self {
+            app_dir: app_dir.as_ref().to_owned(),
+            component,
+        }
     }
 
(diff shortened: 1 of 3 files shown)


itowlson:
Thanks @lann - I adopted that and it made for a cleaner diff I think.  But it did make some of the "when do I use what" a bit fuzzier, so I ended up doing some renames anyway...  Anyway ready for review now.

crates/doctor/src/wasm.rs
```diff
@@ -11,23 +11,33 @@ use crate::{Diagnosis, Diagnostic, PatientApp};
...
     }
 
-    pub fn source(&self) -> WasmSource {
-        match &self.0.source {
+    pub fn rel_source(&self) -> WasmSource {
+        match &self.component.source {
             RawModuleSource::FileReference(path) => WasmSource::Local(path),
             _ => WasmSource::Other,
         }
     }
```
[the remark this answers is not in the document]
lann: or maybe `local_source_path` and `local_source_abs_path` or something

crates/doctor/src/wasm.rs
```diff
@@ -11,34 +11,41 @@ use crate::{Diagnosis, Diagnostic, PatientApp};
...
 #[derive(Debug)]
-pub struct PatientWasm(RawComponentManifest);
+pub struct PatientWasm {
+    app_dir: PathBuf,
+    component: RawComponentManifest,
+}
 
 #[allow(missing_docs)] // WIP
 impl PatientWasm {
+    pub fn new(app_dir: impl AsRef<Path>, component: RawComponentManifest) -> Self {
```
lann, line 21: Making this `pub` makes it more difficult to extend `PatientWasm` in the future.

itowlson:
@rylev Ugh, good catch.  Probably not.  It looks like those paths are resolved at the apping stage rather than the raw load stage, and the Wasm doctor wisely relies only on the raw manifest (because the apping stage requires a high degree of validity).

I guess I can bodge around it or rework the loader pipeline to more finely reflect load stages.  The latter is my long term goal but I might try the former for now.  _sigh_

crates/doctor/src/wasm.rs
itowlson: Privatised

itowlson:
@rylev updated for #1503

lann (approved):

PR #1515, opened 2023-05-18
Support building only specific components

itowlson:
`spin doctor` is set up to build only the component that the user asks it to fix, but the `spin build` side of that was not ready in time.  Now it is.

```
# Happy path
ivan@hecate:~/testing/kvmadness$ spin build -c kvmadness
Executing the build command for component kvmadness: cargo build --target wasm32-wasi --release
    Finished release [optimized] target(s) in 0.03s
Successfully ran the build command for component 'kvmadness'.

# No-op path
ivan@hecate:~/testing/kvmadness$ spin build -c moarmadness
Component 'moarmadness' does not have a build command.
For information on specifying a build command, see https://developer.fermyon.com/spin/build#setting-up-for-spin-build.

# Sad path
$ spin build -c moarmadness2
Error: There is no component 'moarmadness2' in the manifest.
```

Closes #903 [lann: should cover the use case]

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -6,24 +6,49 @@ mod manifest;
 
 use anyhow::{anyhow, bail, Context, Result};
 use spin_loader::local::parent_dir;
-use std::path::{Path, PathBuf};
+use std::{
+    collections::HashSet,
+    path::{Path, PathBuf},
+};
 use subprocess::{Exec, Redirection};
 
 use crate::manifest::{BuildAppInfoAnyVersion, RawComponentManifest};
 
 /// If present, run the build command of each component.
-pub async fn build(manifest_file: &Path) -> Result<()> {
+pub async fn build(manifest_file: &Path, component_ids: &[String]) -> Result<()> {
     let manifest_text = tokio::fs::read_to_string(manifest_file)
         .await
         .with_context(|| format!("Cannot read manifest file from {}", manifest_file.display()))?;
     let app = toml::from_str(&manifest_text).map(BuildAppInfoAnyVersion::into_v1)?;
     let app_dir = parent_dir(manifest_file)?;
 
-    if app.components.iter().all(|c| c.build.is_none()) {
-        println!("No build command found!");
+    let components_to_build = if component_ids.is_empty() {
+        app.components
+    } else {
+        let all_ids: HashSet<_> = app.components.iter().map(|c| &c.id).collect();
+        let unknown_component_ids: Vec<_> = component_ids
+            .iter()
+            .filter(|id| !all_ids.contains(id))
+            .map(|s| s.as_str())
+            .collect();
+
+        if !unknown_component_ids.is_empty() {
+            bail!("Unknown component(s) {}", unknown_component_ids.join(", "));
+        }
+
+        app.components
+            .into_iter()
+            .filter(|c| component_ids.contains(&c.id))
+            .collect()
+    };
+
+    if components_to_build.iter().all(|c| c.build.is_none()) {
+        println!("None of the components have a build command.");
+        println!("For information on specifying a build command, see https://developer.fermyon.com/spin/build#setting-up-for-spin-build.");
         return Ok(());
     }
-    app.components
+
+    components_to_build
         .into_iter()
         .map(|c| build_component(c, &app_dir))
         .collect::<Result<Vec<_>, _>>()?;
@@ -101,6 +126,6 @@ mod tests {
     #[tokio::test]
     async fn can_load_even_if_trigger_invalid() {
         let bad_trigger_file = test_data_root().join("bad_trigger.toml");
-        build(&bad_trigger_file).await.unwrap();
+        build(&bad_trigger_file, &[]).await.unwrap();
     }
 }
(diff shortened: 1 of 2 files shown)


fibonacci1729 (approved):

crates/build/src/lib.rs
```diff
@@ -12,24 +12,42 @@ use subprocess::{Exec, Redirection};
...
+            println!("Successfully ran the build command for the Spin components.");
+        }
+        Some(id) => {
+            if let Some(component) = app.components.into_iter().find(|c| &c.id == id) {
+                if component.build.is_some() {
+                    build_component(component, &app_dir)?;
+                    println!("Successfully ran the build command for component '{id}'.");
+                } else {
+                    eprintln!("Component '{id}' does not have a build command.");
+                    eprintln!("For information on specifying a build command, see https://developer.fermyon.com/spin/build#setting-up-for-spin-build.");
```
lann, line 43: Should this return an error result? That would signal to doctor that the fix didn't work, and might be helpful for other kinds of automation.

lann:
Ah yes, I _purposefully_ left this unimplemented to....drive developer engagement?

src/commands/build.rs
```diff
@@ -38,7 +38,7 @@ pub struct BuildCommand {
 impl BuildCommand {
     pub async fn run(self) -> Result<()> {
         let manifest_file = crate::manifest::resolve_file_path(&self.app_source)?;
-        spin_build::build(&manifest_file).await?;
+        spin_build::build(&manifest_file, &self.component_id).await?;
```
lann, line 41: Would it make sense for `component_id` to be a `Vec<String>` (allowing the flag to be specified multiple times)? I'm thinking that would give feature parity with the zombie PR https://github.com/fermyon/spin/pull/903

crates/build/src/lib.rs
itowlson: The doctor should never call this for components without a build command - it checks before recommending the "build" treatment.

Yeah I dithered about the error result.  I think I kind of reached the conclusion that it hadn't _failed_, there was just _nothing to do._  And so I was a bit wary about Rust printing `Error: ...` which we can't currently stop it doing.  But if other folks would prefer an error than I'm happy to go along with that.

src/commands/build.rs
itowlson: It couldn't possibly do any harm, he said, cheerily placing his Jenga block.

crates/build/src/lib.rs
itowlson: Looking at the existing flow, "no components have a build command" is currently not an error (and prints to stdout rather than stderr so I am being inconsistent there at least).

itowlson: I am reworking it as you suggested to take a list of components so this will unify it to the existing flow.  I will use that existing flow for now; we can change it to be an error if we want.

itowlson:
Updated to allow multiple use, resulting in a slight change to some cases:

```
ivan@hecate:~/testing/kvmadness$ spin build -c moarmadness
None of the components have a build command.
For information on specifying a build command, see https://developer.fermyon.com/spin/build#setting-up-for-spin-build.
ivan@hecate:~/testing/kvmadness$ spin build -c moarmadness -c fie
Error: Unknown component(s) fie
ivan@hecate:~/testing/kvmadness$ spin build -c moarmadness -c fie -c zounds
Error: Unknown component(s) fie, zounds
```

crates/build/src/lib.rs
```diff
@@ -6,25 +6,49 @@ mod manifest;
...
+        let all_ids: HashSet<_> = app.components.iter().map(|c| &c.id).collect();
+        let unknown_component_ids: Vec<_> = component_ids
+            .iter()
+            .filter(|id| !all_ids.contains(id))
+            .map(|s| s.as_str())
+            .collect();
+
+        if !unknown_component_ids.is_empty() {
+            bail!("Unknown component(s) {}", unknown_component_ids.join(", "));
+        }
```
rylev, line 37: nit: Might as well avoid the unnecessary allocation in the happy path?

```suggestion
        let unknown_component_ids = component_ids
            .iter()
            .filter(|id| !all_ids.contains(id))
            .peekable();

        if unknown_component_ids.peek().is_some() {
            bail!("Unknown component(s) {}", unknown_component_ids.collect::<Vec<_>>().join(", "));
        }
```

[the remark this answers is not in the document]
lann: Sub-option: `itertools` (which other `spin` crates already use) has [`itertools::join`](https://docs.rs/itertools/latest/itertools/fn.join.html)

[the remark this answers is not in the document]
itowlson: I thought about this but felt that "this set is empty" was more explanatory than "this string we are composing for UI purposes is empty."

itowlson: TIL: `peekable()` - thank you!

"Avoid allocating an empty vector" doesn't feel like a huge priority when the happy path goes on to run an external process (a compiler, no less).  And I have to admit I find `is_empty` more obvious than `peek().is_some()`, though maybe the latter is idiomatic to Rust devs?

But I can change it (to this or @lann's suggestion) if other folks prefer.  We could also use @lann's mention of `itertools::join` to simplify the sad path in this one.

rylev: No need to block on my nits. I'll let you decide how to proceed from here. 😀

PR #1519, opened 2023-05-18
bump versions for 1.2.1

fibonacci1729:


lann:
This should target `v1.2` (after updating `v1.2` to include what you need for the release)

itowlson (approved):

PR #1521, opened 2023-05-18
Updates for v1.2.1

fibonacci1729:


Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "Apache-2.0 WITH LLVM-exception"
 rust-version = "1.67"
 
 [workspace.package]
-version = "1.2.0"
+version = "1.3.0-pre0"
```
itowlson, line 10: Should this be 1.2.1 or does it get overridden in CI anyway?

fibonacci1729: It will get overwritten by https://github.com/fermyon/spin/pull/1519 (once i change the base to v1.2 after merging this)

itowlson (approved):

PR #1524, opened 2023-05-19
Fix merge woe with new K/V trait function

itowlson:
K/V validation added a new `StoreManager` trait function.  Unfortunately, in the meantime, we had added a new `StoreManager` implementation (Azure CosmosDB).  And I didn't think about this when I merged (because there was no conflict).  And now `main` is broken.

So, uh, this unbreaks `main`.

### DIFF
--- a/crates/key-value-azure/src/lib.rs
+++ b/crates/key-value-azure/src/lib.rs
@@ -33,6 +33,10 @@ impl StoreManager for KeyValueAzureCosmos {
             client: self.client.clone(),
         }))
     }
+
+    fn is_defined(&self, _store_name: &str) -> bool {
+        true
+    }
 }
 
 struct AzureCosmosStore {


rajatjindal (approved):

rajatjindal:
hmm wondering how come our CI didn't catch this issue during the PR?

itowlson:
@rajatjindal I think because:

1. I based the validation branch off main a few days ago.
2. After the last commit to the validation branch, and after all checks were complete, Cosmos merged.
3. Cosmos did not touch the same files as the validation branch, so GitHub saw no conflict, and left the validation PR checks as green.
4. I did not rebase the validation branch before merging, so GH had no prompt to re-run the PR checks.

I am not sure if there is a reasonable way we can force checks to re-run on changes to what GH sees as unrelated parts of the code... even if we can force checks, it could get very noisy and inefficient without some smarts to it.  I'm not very knowledgeable about what we can do in this area though.

rajatjindal:
that makes sense. thank you for explaining.

PR #1527, opened 2023-05-19
Add color to each step of spin build and spin up

rylev:
This adds green to the first word of each logical step in `spin build` and `spin up`.

## Current 

image
image

## New 

image
image

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -23,13 +23,12 @@ pub async fn build(manifest_file: &Path) -> Result<()> {
         println!("No build command found!");
         return Ok(());
     }
-
     app.components
         .into_iter()
         .map(|c| build_component(c, &app_dir))
         .collect::<Result<Vec<_>, _>>()?;
 
-    println!("Successfully ran the build command for the Spin components.");
+    terminal::step!("Finished", "building all Spin components");
     Ok(())
 }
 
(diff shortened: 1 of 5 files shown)


crates/trigger-http/src/lib.rs
```diff
@@ -147,7 +147,7 @@ impl TriggerExecutor for HttpTrigger {
         // Print startup messages
         let scheme = if tls.is_some() { "https" } else { "http" };
         let base_url = format!("{}://{:?}", scheme, listen_addr);
-        println!("\nServing {}", base_url);
+        terminal::step!("Serving", "{}", base_url);
```
itowlson, line 150: This has lost its preceding blank line.

crates/build/src/lib.rs
```diff
@@ -19,28 +19,20 @@ pub async fn build(manifest_file: &Path) -> Result<()> {
     let app = toml::from_str(&manifest_text).map(BuildAppInfoAnyVersion::into_v1)?;
     let app_dir = parent_dir(manifest_file)?;
 
-    if app.components.iter().all(|c| c.build.is_none()) {
```
itowlson, line 22: I don't think we should remove this.  I'd rather have the specific "there is nothing to do" feedback than "huh it didn't do anything, what happened."

itowlson (approved):
This looks fine - I have to admit I am a bit unsure about the "first word highlighted" look, but I expect that is unfamiliarity more than anything else.  The best thing is to try it and see how people feel.

There are a couple of things that got deleted that need to be restored - the "no build command" output and the blank line before "Serving" - so technically this is "request changes," but they're trivial fixes and I wanted to save you waiting through another 24-hour review cycle.  So please don't merge without them, but once done please don't feel the need to wait for another go-around.

PR #1529, opened 2023-05-21
`spin doctor`: detect Rust not installed/Wasm target not installed

itowlson:
This works (as best I can tell without going through a full build-uninstall Rust-doctor-install Rust cycle every time), but was at least partly a learning exercise, and raised a few questions/considerations:

* This was implemented in Spin itself mostly to ease the proof of concept development.  I generally feel that Spin shouldn't know about this stuff and it should go out to a plugin, but wanted to get my head around this before working on a spin doctor plugin system.

* I initially tried to follow the Wasm model of having an `impl<T: RustDiagnostic> Diagnostic for T`, but this hits ye olde "conflicting implementations of Diagnostic" problem (because what if T implemented WasmDiagnostic and RustDiagnostic what then Ivan).  It seems like, bar shenanigans, only one family of diagnostics can have this blanket implementation (there's a similar issue with treatments).  I ducked the issue for this PR, but we might need to tackle this a different way in future.

* It wasn't quite clear to me how to manage multiple overlapping diagnoses and treatments.  The specific case here was "Rust is out of date".  I could have made this a third arm of the `TargetDiagnosis` enum (it's misnamed, I know) but then we need to get into "Rust is out of date and _does_ have the Wasm target" (so only needs updating) vs "Rust is out of date and _doesn't_ have the Wasm target" (so needs update and `target add`).  I didn't want to use a vector of diagnostics because that would result in multiple doctor prompts.  (To be clear, four cases of the enum would be fine!  I just felt like I should check in with other folks before embarking on a combinatoric explosion.)

### DIFF
--- a/crates/doctor/src/rustlang.rs
+++ b/crates/doctor/src/rustlang.rs
@@ -1,0 +1,2 @@
+/// Diagnose Wasm target problems.
+pub mod target;
--- a/crates/doctor/src/rustlang/target.rs
+++ b/crates/doctor/src/rustlang/target.rs
@@ -1,0 +1,232 @@
+use anyhow::Result;
+use async_trait::async_trait;
+
+use crate::{Diagnosis, Diagnostic, PatientApp, StopDiagnosing, Treatment};
+
+/// VersionDiagnostic detects problems with the app manifest version field.
+#[derive(Default)]
+pub struct TargetDiagnostic;
+
+#[async_trait]
+impl Diagnostic for TargetDiagnostic {
+    type Diagnosis = TargetDiagnosis;
+
+    async fn diagnose(&self, patient: &PatientApp) -> Result<Vec<Self::Diagnosis>> {
+        // TODO: this, down to the "does the app use Rust" check, probably ought to move up to the Rust level
+        // but we can defer this until we have more Rust diagnoses
+        let path = &patient.manifest_path;
+        let manifest = spin_loader::local::raw_manifest_from_file(&path)
+            .await?
+            .into_v1();
+        let uses_rust = manifest.components.iter().any(|c| {
+            c.build
+                .as_ref()
+                .map(|b| b.command.starts_with("cargo"))
+                .unwrap_or_default()
+        });
+
+        if uses_rust {
+            diagnose_rust_wasi_target().await
+        } else {
+            Ok(vec![])
+        }
+    }
+}
+
+async fn diagnose_rust_wasi_target() -> Result<Vec<TargetDiagnosis>> {
+    // does any component contain a build command with `cargo` as the program?
+    // if so, run rustup target list --installed and:
+    // - if rustup is not present, check if cargo is present
+    //   - if not, return RustNotInstalled
+    //   - if so, warn but return empty list (Rust is installed but not via rustup, so we can't perform a diagnosis - bit of an edge case this one, and the user probably knows what they're doing...?)
+    // - if rustup is present but the list does not contain wasm32-wasi, return WasmTargetNotInstalled
+    // - if the list does contain wasm32-wasi, return an empty list
+    // NOTE: this does not currently check against the Rust SDK MSRV - that could
+    // be a future enhancement or separate diagnosis, but at least the Rust compiler
+    // should give a clear error for that!
+
+    let diagnosis = match get_rustup_target_status().await? {
+        RustupStatus::AllInstalled => vec![],
+        RustupStatus::WasiNotInstalled => vec![TargetDiagnosis::WasmTargetNotInstalled],
+        RustupStatus::RustupNotInstalled => match get_cargo_status().await? {
+            CargoStatus::Installed => {
+                terminal::warn!(
+                    "Spin Doctor can't determine if the Rust wasm32-wasi target is installed."
+                );
+                vec![]
+            }
+            CargoStatus::NotInstalled => vec![TargetDiagnosis::RustNotInstalled],
+        },
+    };
+    Ok(diagnosis)
+}
+
+#[allow(clippy::enum_variant_names)]
+enum RustupStatus {
+    RustupNotInstalled,
+    WasiNotInstalled,
+    AllInstalled,
+}
+
+async fn get_rustup_target_status() -> Result<RustupStatus> {
+    let target_list_output = tokio::process::Command::new("rustup")
+        .args(["target", "list", "--installed"])
+        .output()
+        .await;
+    let status = match target_list_output {
+        Err(e) => {
+            if e.kind() == std::io::ErrorKind::NotFound {
+                RustupStatus::RustupNotInstalled
+            } else {
+                anyhow::bail!("Failed to run `rustup target list --installed`: {e:#}")
+            }
+        }
+        Ok(output) => {
+            let stdout = String::from_utf8_lossy(&output.stdout);
+            if stdout.lines().any(|line| line == "wasm32-wasi") {
+                RustupStatus::AllInstalled
+            } else {
+                RustupStatus::WasiNotInstalled
+            }
+        }
+    };
+    Ok(status)
+}
+
+enum CargoStatus {
+    Installed,
+    NotInstalled,
+}
+
+async fn get_cargo_status() -> Result<CargoStatus> {
+    let cmd_output = tokio::process::Command::new("cargo")
+        .arg("--version")
+        .output()
+        .await;
+    let status = match cmd_output {
+        Err(e) => {
+            if e.kind() == std::io::ErrorKind::NotFound {
+                CargoStatus::NotInstalled
+            } else {
+                anyhow::bail!("Failed to run `cargo --version`: {e:#}")
+            }
+        }
+        Ok(_) => CargoStatus::Installed,
+    };
+    Ok(status)
+}
+
+/// TargetDiagnosis represents a problem with the Rust target.
+#[derive(Debug)]
+pub enum TargetDiagnosis {
+    /// Rust is not installed: neither cargo nor rustup is present
+    RustNotInstalled,
+    /// The Rust wasm32-wasi target is not installed: rustup is present but the target isn't
+    WasmTargetNotInstalled,
+}
+
+impl Diagnosis for TargetDiagnosis {
+    fn description(&self) -> String {
+        match self {
+            Self::RustNotInstalled => "The Rust compiler isn't installed".into(),
+            Self::WasmTargetNotInstalled => {
+                "The required Rust target 'wasm32-wasi' isn't installed".into()
+            }
+        }
+    }
+
+    fn treatment(&self) -> Option<&dyn Treatment> {
+        Some(self)
+    }
+}
+
+#[async_trait]
+impl Treatment for TargetDiagnosis {
+    fn summary(&self) -> String {
+        match self {
+            Self::RustNotInstalled => "Install the Rust compiler and the wasm32-wasi target",
+            Self::WasmTargetNotInstalled => "Install the Rust wasm32-wasi target",
+        }
+        .into()
+    }
+
+    async fn dry_run(&self, _patient: &PatientApp) -> Result<String> {
+        let message = match self {
+            Self::RustNotInstalled => "Download and run the Rust installer from https://rustup.rs, with the `--target wasm32-wasi` option",
+            Self::WasmTargetNotInstalled => "Run the following command:\n    `rustup target add wasm32-wasi`",
+        };
+        Ok(message.into())
+    }
+
+    async fn treat(&self, _patient: &mut PatientApp) -> Result<()> {
+        match self {
+            Self::RustNotInstalled => {
+                install_rust_with_wasi_target().await?;
+            }
+            Self::WasmTargetNotInstalled => {
+                install_wasi_target()?;
+            }
+        }
+        Ok(())
+    }
+}
+
+async fn install_rust_with_wasi_target() -> Result<()> {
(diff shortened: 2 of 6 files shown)


lann:
> I initially tried to follow the Wasm model of having an impl<T: RustDiagnostic> Diagnostic for T, but this hits ye olde "conflicting implementations of Diagnostic" problem (because what if T implemented WasmDiagnostic and RustDiagnostic what then Ivan). It seems like, bar shenanigans, only one family of diagnostics can have this blanket implementation (there's a similar issue with treatments). I ducked the issue for this PR, but we might need to tackle this a different way in future.

The `*Diagnostic` pattern is just a convenience; it it isn't...err...convenient, then `impl Diagnostic` is fine.

lann:
> It wasn't quite clear to me how to manage multiple overlapping diagnoses and treatments.

Yeah, this is tricky and led to a lot of churn as I was writing `Diagnosis`/`Treatment`. I tried some much more complex solutions involving re-running diagnostics after treatments and tracking what had already been fixed, but they made it much more difficult to understand how to write them.

Something to note in this case: nothing constrains a `Diagnosis` to be an `enum`; it could just as easily be:

```rust
struct TargetDiagnosis {
    out_of_date: bool,
    missing_wasm_target: bool,
}
```

lann:
> I didn't want to use a vector of diagnostics because that would result in multiple doctor prompts.

I would think that "rust is out of date" maybe _should_ be a separate prompt, especially if it doesn't qualify as `Diagnosis::is_critical`. Maybe I don't _want_ to update Rust; stop pressuring me, `spin doctor`!

itowlson:
Sorry, @lann, "doctor knows best."  _prepares the instruments_

Seriously... for the "Rust is out of date" one, I was thinking "Rust doesn't meet the MSRV of the Spin SDK," not just "there is an upgrade ~threat~ opportunity."  So it would be a "you won't be able to compile your app" level of criticality, and could be folded into the "installed" diagnosis.  And you are of course right that a struct would represent this better than an enum.

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,192 @@
...
+    let installer_bin = resp.bytes().await?;
+
+    std::fs::write(&installer_path, &installer_bin)?;
+
+    let mut cmd = std::process::Command::new(installer_path);
+    let status = cmd.status()?;
+    Ok(status)
+}
+
+fn install_wasi_target() -> anyhow::Result<()> {
```
fibonacci1729, line 183:
```suggestion
fn install_wasi_target() -> Result<()> {
```

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,192 @@
...
+    std::thread::spawn(move || {
+        stdin.write_all(&script).unwrap();
+    });
+
+    let output = shell.wait_with_output()?;
+    Ok(output.status)
+}
+
+#[cfg(windows)]
+async fn run_rust_installer() -> anyhow::Result<std::process::ExitStatus> {
```
fibonacci1729, line 163:
```suggestion
async fn run_rust_installer() -> Result<std::process::ExitStatus> {
```

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+async fn install_rust() -> anyhow::Result<()> {
+    let status = run_rust_installer().await?;
+    anyhow::ensure!(status.success(), "Rust installation failed: {status:?}");
+    Ok(())
+}
+
+#[cfg(not(windows))]
+async fn run_rust_installer() -> anyhow::Result<std::process::ExitStatus> {
```
fibonacci1729, line 144:
```suggestion
async fn run_rust_installer() -> Result<std::process::ExitStatus> {
```

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,192 @@
...
+        };
+        Ok(message.into())
+    }
+
+    async fn treat(&self, _patient: &mut PatientApp) -> Result<()> {
+        self.treat_rust().await
+    }
+}
+
+async fn install_rust() -> anyhow::Result<()> {
```
fibonacci1729, line 137:
```suggestion
async fn install_rust() -> Result<()> {
```

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,192 @@
...
+impl RustTreatment for TargetDiagnosis {
+    fn summary(&self) -> String {
+        match self {
+            Self::RustNotInstalled => "Install the Rust compiler and the wasm32-wasi target",
+            Self::WasmTargetNotInstalled => "Install the Rust wasm32-wasi target",
+        }
+        .into()
+    }
+
+    async fn treat_rust(&self) -> anyhow::Result<()> {
```
fibonacci1729, line 104:
```suggestion
    async fn treat_rust(&self) -> Result<()> {
```

itowlson:
Status:

* I decided to skip the check for MSRV in this iteration
* We agreed to build this into the Spin CLI rather than as a plugin
* So this is now ready for review

Thanks all!

crates/doctor/src/rustlang.rs
```diff
@@ -0,0 +1,16 @@
+use anyhow::Result;
+use async_trait::async_trait;
+
+/// Diagnose Wasm target problems.
+pub mod target;
+
+/// RustTreatment helps implement [`Treatment`]s for Rust setup problems.
+#[async_trait]
+pub trait RustTreatment {
```
lann, line 9: I would probably remove this trait until its actually useful (YAGNI).

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,193 @@
...
+        match self {
+            Self::RustNotInstalled => "The Rust compiler isn't installed".into(),
+            Self::WasmTargetNotInstalled => {
+                "The required Rust target 'wasm32-wasi' isn't installed".into()
+            }
+        }
+    }
+
+    fn is_critical(&self) -> bool {
+        true
```
lann, line 87: This is the default impl. :shrug:

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,193 @@
...
+            Self::WasmTargetNotInstalled => "Install the Rust wasm32-wasi target",
+        }
+        .into()
+    }
+
+    async fn treat_rust(&self) -> Result<()> {
+        match self {
+            Self::RustNotInstalled => {
+                install_rust().await?;
+                install_wasi_target()?;
```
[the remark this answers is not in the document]
lann: On that note, I wonder if we should try to detect that rustup has been installed but the user still needs to restart their terminal. Just checking for the existence of `~/.rustup/` (or the Windows equivalent) may be a good-enough heuristic and could help reduce annoyance if people don't see the output of the rust install step (which tells you to restart your terminal).

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,193 @@
...
+    Ok(output.status)
+}
+
+#[cfg(windows)]
+async fn run_rust_installer() -> Result<std::process::ExitStatus> {
+    // We currently distribute Windows builds only for x64, so hopefully
+    // this won't be an issue.
+    if std::env::consts::ARCH != "x86_64" {
+        anyhow::bail!("Spin Doctor can only install Rust for Windows on x64 processors");
+    }
```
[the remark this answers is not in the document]
itowlson: That makes `Diagnosis::treatment` more complicated, because it gets an OS/arch dependency in one case but not others.  It would be worth it if we had Windows builds other than x64, but we don't, so this is a "should never happen" scenario; and as such I think it's okay to put it in the place where the OS and treatment are already selected for.  Not that the extra complexity is enormous, so happy to do it if you prefer.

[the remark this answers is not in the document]
lann: But what if someone compiles spin for arm windows and then...erm....deletes rust... 😐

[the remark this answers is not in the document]
itowlson: Pardon me while I allow that scenario to grip me with terror.

crates/doctor/src/rustlang/target.rs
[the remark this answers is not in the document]
itowlson: Thanks, good catch and thanks for the tip re passing arguments.

Windows runs `rustup-init` directly (we manage the download) so I should be able to simply pass the `--target` option.  WHAT COULD GO WRONG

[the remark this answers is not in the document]
itowlson: Okay, that seems to work.  You are right, though, about the risk of the PATH not being updated.  This is particularly vexing because `spin doctor` wants to charge on and have you build your application immediately, which of course fails with "Rust not installed."  Huh?  You told me you installed it.  Call yourself a doctor?

We could introduce a notion of a stopping treatment, after which further diagnoses should be suppressed (or are listed but indicated as "re-run to treat").  It might be a bit of a blunt instrument but I'm not sure how else to manage that transition.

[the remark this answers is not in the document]
lann: I can think of a few "magical" solutions like updating `PATH` for the current process but those generally seem too surprising.

Thinking about the implementation of "stopping treatment", given that this is (hopefully) an exceptional kind of treatment I think it would be fine to return an error like `UserInterventionRequired { message: String, continue_treatments: bool }`.

crates/doctor/src/rustlang/target.rs
rylev: I wish we actually kept it as `anyhow::Result` across the entire code base... 🥲

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,177 @@
...
+            let target_list_output = tokio::process::Command::new("rustup")
+                .args(["target", "list", "--installed"])
+                .output()
+                .await;
+            let diagnosis = match target_list_output {
+                Err(e) => {
+                    if e.kind() == std::io::ErrorKind::NotFound {
+                        vec![TargetDiagnosis::RustNotInstalled]
+                    } else {
+                        anyhow::bail!(e)
```
rylev, line 46: Can we add context of the command we're running? The error message may only make sense if the user knows we're running `rustup target list --installed`.

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,177 @@
...
+            // be a future enhancement or separate diagnosis, but at least the Rust compiler
+            // should give a clear error for that!
+            let target_list_output = tokio::process::Command::new("rustup")
+                .args(["target", "list", "--installed"])
+                .output()
+                .await;
+            let diagnosis = match target_list_output {
+                Err(e) => {
+                    if e.kind() == std::io::ErrorKind::NotFound {
+                        vec![TargetDiagnosis::RustNotInstalled]
```
rylev, line 44: Nit: `rustup` not being installed usually means Rust (i.e., `rustc` and `cargo`) isn't installed but not necessarily. We might want to be more specific with our error message.

crates/doctor/src/rustlang/target.rs
```diff
@@ -0,0 +1,177 @@
...
+            Self::RustNotInstalled => "Download and run the Rust installer from https://rustup.rs, with the `--target wasm32-wasi` option",
+            Self::WasmTargetNotInstalled => "Run the following command:\n    `rustup target add wasm32-wasi`",
+        };
+        Ok(message.into())
+    }
+
+    async fn treat(&self, _patient: &mut PatientApp) -> Result<()> {
+        match self {
+            Self::RustNotInstalled => {
+                install_rust().await?;
```
rylev, line 110: Do we also need to run `install_wasi_target` after installing Rust?

lann: It runs `rustup --target wasm32-wasi`. Maybe the function should be `install_rust_and_wasi`?

lann: Context: https://github.com/fermyon/spin/pull/1529#discussion_r1229633743

rylev: Ah yes, I totally missed that.

crates/doctor/src/rustlang/target.rs
fibonacci1729: Personally, I agree. As long as we are consistent.

crates/doctor/src/rustlang/target.rs
itowlson: That's a good point.  Should we try running something like `cargo --version` instead/as well?

If `rustup` isn't installed then what is the recommended way to install the WASI target?

lann: It would be more complex, but maybe less confusing in these less-common cases:

- if cargo isn't available, offer to install rust+wasi via rustup
- else if rustup is available and wasi isn't installed, offer to install wasi
- else if rustup isn't available, print a warning that wasi may not be installed (can we detect wasi without rustup?)

crates/doctor/src/rustlang/target.rs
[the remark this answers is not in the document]
itowlson: Done

crates/doctor/src/rustlang/target.rs
itowlson: Done

itowlson:
What more work is required to get this signed off?

lann (approved):

PR #1532, opened 2023-05-22
`spin doctor`: allow selecting which diagnostics to run

itowlson:
This is WIP for discussion.

The idea is to give each diagnostic an identifier (like a lint ID), so that `spin doctor` can be told to check for only specific problems.  Why would you not want to know about ALL the problems?  The goal I have in mind is the "Why doesn't Spin install the wasm32-wasi target so it _just works_" scenario, and the solution I am exploring is roughly like this:

* User creates a new app from a Rust template
* The Rust template has some configuration saying "run the `rust` checks after instantiation"
* We want to run _only_ the `rust` checks because we know that out of the box there will be issues like "missing Wasm"
  * Arguably we could run all the checks, but we wouldn't want to present "missing Wasm" as a problem when the user has just created the app - it makes it seem like app creation hasn't done what it should have done
  * Arguably we could run `spin build` after instantiation and save them the effort (and us the diagnostic)?  Yeah, but then we need to run the Rust check _first_ to make sure we can build at all...

This is, for sure, a few steps down the line.  And there might be better ways to do it.  But basically the idea is to allow `spin doctor` to act as a setup assistant, not only something the user can reach for when they have identified a problem.

Happy to discuss, or to hear if people already have planned solutions the same problems and specific ways they'd like to see them tackled.

NOTE: the PR currently has an issue where if you limit it to, say, `wasm` checks, but the manifest doesn't parse, then the doctor says "no problems found" rather than "your application is so busted that you need to run the full checks to unbust it before I can help with the thing you asked about," so I need to look at this if we are okay with this option existing.

### DIFF
--- a/crates/doctor/src/lib.rs
+++ b/crates/doctor/src/lib.rs
@@ -23,21 +23,12 @@ pub struct Checkup {
 
 impl Checkup {
     /// Return a new checkup for the app manifest at the given path.
-    pub fn new(manifest_path: impl Into<PathBuf>) -> Self {
-        let mut checkup = Self {
-            manifest_path: manifest_path.into(),
-            diagnostics: vec![],
-        };
-        checkup.add_diagnostic::<manifest::version::VersionDiagnostic>();
-        checkup.add_diagnostic::<manifest::trigger::TriggerDiagnostic>();
-        checkup.add_diagnostic::<wasm::missing::WasmMissingDiagnostic>();
-        checkup
-    }
-
-    /// Add a detectable problem to this checkup.
-    pub fn add_diagnostic<D: Diagnostic + Default + 'static>(&mut self) -> &mut Self {
-        self.diagnostics.push(Box::<D>::default());
-        self
+    pub fn new(manifest_path: impl Into<PathBuf>, checks: &[String]) -> Self {
+        let mut builder = CheckupBuilder::new(manifest_path, checks);
+        builder.add_diagnostic::<manifest::version::VersionDiagnostic>();
+        builder.add_diagnostic::<manifest::trigger::TriggerDiagnostic>();
+        builder.add_diagnostic::<wasm::missing::WasmMissingDiagnostic>();
+        builder.build()
     }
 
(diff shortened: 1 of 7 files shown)


lann:
> wasm checks, but the manifest doesn't parse, then the doctor says "no problems found"

These kinds of errors should be returned by `diagnose`, and are currently logged at `DEBUG`. I didn't want to display them by default exactly because if "your application is so busted that we can't even run checks" then you'd get a wall of unhelpful errors.

We could perhaps do something more sophisticated with logging like "if someone explicitly requests a check, show its errors".

lann:
Something I'd like to consider with this `ID` approach is how it would interact with plugins. Maybe we could "namespace" checks by some plugin ID? Should all the builtin checks be `spin-*:`?

itowlson:
I was initially thinking that plugins could namespace their checks how they liked, but I like your idea better because it means plugins don't have to get it right themselves.

My vague sense (which I should have included in the outline) was that we could have a convention that plugins named `doctor-*` would be run as part of `spin doctor`.  Given that, we could identify checks within a plugin by the `*` bit e.g. `rust:installed` would refer to the `installed` check within `doctor-rust`.  In which case identifying built-in ones as `spin` would work.  So you could write `spin doctor --check spin` to run only the built-in checks.

But there is clearly more pondering to be done on this (and that pondering likely needs to happen before this feature can be released into the wild).  Thanks for the discussion!

lann:
> convention that plugins named `doctor-*` would be run as part of spin doctor

Separately - while we're on the topic - I've been wondering if we might want to curtail this approach in favor of explicitly listing plugin types in the manifest. The main advantage I see of relying on binary name patterns is that it allows you to catalog your plugins lazily by traversing `$PATH`, but since we are already explicitly managing installed plugins I don't think that applies. One advantage of explicitly listing plugins types would be allowing one plugin to implement multiple types; the `js2wasm` plugin could grow a `doctor` feature for example.

itowlson:
For sure, the plugin types thing is definitely worth exploring!

PR #1536, opened 2023-05-23
Orchestrator Plugin Example

suneetnangia:
An approach to enable intra Spin (within same the Spin instance) guest modules orchestration using WIT contracts. It makes use of the plugin model extensibility Spin provides to build a domain specific orchestrator (a custom plugin). Domain specific plugin defines the WIT contracts as per the business domain for multiple guest (Wasm) modules, it will also host and move messages/data between these guest modules as per the domain logic. 
Drawing some parallels here from code level and service level design patterns, intra Spin orchestration of modules can be considered as nano services which do not need to cross the boundary of network but still need to be composed together to form a business logic. By not involving network, we avoid the complex compensation logic (idempotency, circuit-breaking) in absence of transactions and serialisation-deserialization of messages. WIT contracts defined in the custom Plugin allows a contract driven development of guest modules in a polyglot manner. 
These composed services may eventually then interact with other services (could be Spin based or otherwise) on the network using standard Spin SDKs e.g. Http.

[still in progress, will clean up a bit but I wanted to gather some initial thoughts on it]


suneetnangia:
Building on this, now considering adding export functions from WIT which can be used by components to make a service composition, this should reduce tight coupling between components in the plugin.

PR #1537, opened 2023-05-23
Run e2e tests against local build by default

rylev:
This switches local e2e testing to test against local changes. If users want the old default of running against canary they can do so with the `E2E_USE_CANARY` env variable set to `true`. 

I believe the default of testing against local changes is the more intuitive default rather than testing against an already release artifact (which already has e2e tests run against it in CI).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ ARCH = $(shell uname -p)
 
 ## dependencies for e2e-tests
 E2E_VOLUME_MOUNT     ?=
-E2E_BUILD_SPIN       ?= false
+E2E_USE_CANARY       ?= false ## Whether to use the canary build instead of the local build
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,7 +6,8 @@ ARCH = $(shell uname -p)
 
 ## dependencies for e2e-tests
 E2E_VOLUME_MOUNT     ?=
-E2E_BUILD_SPIN       ?= false
+E2E_USE_CANARY       ?= false ## Whether to use the canary build instead of the local build
+E2E_BUILD_SPIN 	     =  $(if $E2E_USE_CANARY == "false",true,false)
```
rajatjindal, line 10: wondering if we should use built binary in CI instead of building again. Downside is that on Github actions, the build jobs are really slow.

rylev:
Closing this as e2e tests have changed dramatically since this PR was opened.

PR #1539, opened 2023-05-23
Add initial implementation of turso integration

rylev:
Marking this as WIP since it needs a bit of clean up, but it works in its current state. I was able to run the [example todo app](https://github.com/rylev/spin-todo) against a Turso backend.

### DIFF
--- a/crates/sqlite-turso/src/lib.rs
+++ b/crates/sqlite-turso/src/lib.rs
@@ -1,0 +1,79 @@
+use std::sync::Arc;
+
+use libsql_client::DatabaseClient;
+use spin_world::sqlite::{self, RowResult};
+
+#[derive(Clone)]
+pub struct TursoClient {
+    client: libsql_client::reqwest::Client,
+}
+
+impl TursoClient {
+    pub fn new(url: String, token: String) -> Self {
+        Self {
+            client: libsql_client::reqwest::Client::new(url, token),
+        }
+    }
+}
+
+impl spin_sqlite::ConnectionManager for TursoClient {
+    fn get_connection(&self) -> Result<Arc<dyn spin_sqlite::Connection>, sqlite::Error> {
+        Ok(Arc::new(self.clone()))
+    }
+}
+
+impl spin_sqlite::Connection for TursoClient {
+    fn query(
+        &self,
+        query: &str,
+        parameters: Vec<spin_world::sqlite::Value>,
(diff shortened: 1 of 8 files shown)


crates/trigger/src/runtime_config/sqlite.rs
```diff
@@ -53,6 +52,7 @@ fn execute_statements(
 #[serde(rename_all = "snake_case", tag = "type")]
 pub enum SqliteDatabaseOpts {
     Spin(SpinSqliteDatabaseOpts),
+    Turso(TursoOpts),
```
radu-matei, line 55: Thoughts vs. naming this a `libsql` backend?

crates/sqlite-turso/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+fn convert_parameters(parameters: &[spin_world::sqlite::Value]) -> Vec<libsql_client::Value> {
+    parameters
+        .iter()
+        .map(|v| match v {
+            sqlite::Value::Integer(value) => libsql_client::Value::Integer { value: *value },
+            sqlite::Value::Real(value) => libsql_client::Value::Float { value: *value },
+            sqlite::Value::Text(t) => libsql_client::Value::Text { value: t.clone() },
+            sqlite::Value::Blob(_) => todo!(),
```
radu-matei, line 75: Does LibSQL support blob values?

radu-matei, line 75:
```suggestion
            sqlite::Value::Blob(b) => libsql_client::Value::Blob { value: b.clone() },
```

crates/trigger/src/runtime_config/sqlite.rs
rylev: I don't think it's really about libsql as that can be run locally and should work fine with the default spin configuration. Perhaps `sqld` or `remote_libsql`?

radu-matei: The library is `libsql-client`. LibSQL is intended to support both local scenarios and remote scenarios, depending on the URI, so I feel like calling it `libsql` wouldn't be inappropriate.

rylev: Should we also allow running libsql locally as well?

itowlson:
Incorporated into #1545.

PR #1549, opened 2023-06-01
SQLite manifest validation and tests

itowlson:
```
$ spin up
Error: One or more components use SQLite databases which are not defined.
See https://developer.fermyon.com/spin/dynamic-configuration#sqlite-storage-runtime-configuration
Details:
- Component sqlmadness uses database 'honkytonk'
```

The docs page doesn't exist yet so I could take that line out and put it back in when we do the formal release of SQLite, but I'm not immensely motivated!

### DIFF
--- a/tests/testcases/mod.rs
+++ b/tests/testcases/mod.rs
@@ -103,6 +103,95 @@ pub async fn key_value_validation_works(controller: &dyn Controller) {
     tc.try_run(controller).await.unwrap();
 }
 
+pub async fn sqlite_works(controller: &dyn Controller) {
+    async fn checks(
+        metadata: AppMetadata,
+        test_init_key: String,
+        test_init_value: String,
+        _: Option<Pin<Box<dyn AsyncBufRead>>>,
+        _: Option<Pin<Box<dyn AsyncBufRead>>>,
+    ) -> Result<()> {
+        assert_http_response(
+            get_url(
+                metadata.base.as_str(),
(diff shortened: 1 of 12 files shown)


itowlson:
It looks like #1550 is going to land first so I have made this dependent on that (which means that it temporarily includes a couple of extra commits from over there).

itowlson:
Rebased and ready for review.

tests/testcases/mod.rs
```diff
@@ -103,6 +103,95 @@ pub async fn key_value_validation_works(controller: &dyn Controller) {
...
 
+pub async fn sqlite_works(controller: &dyn Controller) {
+    async fn checks(
+        metadata: AppMetadata,
+        test_init_key: String,
+        test_init_value: String,
+        _: Option<Pin<Box<dyn AsyncBufRead>>>,
+        _: Option<Pin<Box<dyn AsyncBufRead>>>,
+    ) -> Result<()> {
+        assert_http_response(
```
rajatjindal, line 114: may be use `assert_status`?

rajatjindal (approved):

fibonacci1729 (approved):

tests/testcases/mod.rs
itowlson: This was copied from `key_value_works`... I will merge as stands (since you have approved and the phrasing on the comment sounds non-blocking), but am happy to update both places in a follow-up PR if you feel it's worth it.  (`assert_status` does seem terser I agree.)

PR #1553, opened 2023-06-02
experimental `wasi-http` support

dicej:
This adds experimental support for the [wasi-http](https://github.com/WebAssembly/wasi-http) proposal. Note that the WIT files have been copied from that repo and modified to avoid name conflicts. Eventually, we'll want to use `wit-deps` to manage them.

The good:
 - Asynchronous, concurrent streaming supported for both inbound and outbound requests and responses
 - Includes an example that uses a custom async executor which could evolve into an async SDK API

The bad:
 - Outbound request/response trailers not yet supported since `reqwest` does not yet support them
 - Some less-used stream operations not yet implemented (e.g. `write_zeros`), but should be easy to add

The ugly:
 - Had to rename a bunch of interfaces (e.g. poll -> poll2) to avoid conflicts with our snapshot of `preview2-protyping`
   - This should be resolved once we update `wasmtime`, `wit-bindgen`, and `wasm-tools` to incorporate https://github.com/WebAssembly/component-model/issues/193
 - We've commented-out the clock, random, and stdio parts of the world in favor of ambient WASI Preview 1 support, again due to the aforementioned conflicts
 - Implementing `wasi-cloud` as a host component and passing a handle to it to the wasi-http trigger is awkward
 - The host component currently traps on certain runtime errors because the WIT interface is not fallible but probably should be
 - The WIT spec is not clear on what should happen in certain corner cases (e.g. if you mutate headers after you've passed them to a constructor, or if you send the same request more than once)

Long term, we'd like to contribute the Spin-agnostic parts of this implementation upstream to the Bytecode Alliance.  For now, we'll let it incubate here in Spin.


crates/http/src/config.rs
```diff
@@ -26,6 +26,8 @@ pub enum HttpExecutorType {
     Spin,
     /// The component implements the Wagi CGI interface.
     Wagi(WagiTriggerConfig),
+    /// The component implements [`wasi-http`](https://github.com/WebAssembly/wasi-http) interface (experimental, subject to change)
+    Wasi,
```
fibonacci1729, line 30: Could you update the doc comment on the `HttpExecutorType`? Maybe remove the second sentence or amend it to mention the new variant.

fibonacci1729 (approved):

crates/trigger/src/lib.rs
```diff
@@ -135,10 +138,11 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
                     &mut builder,
                     spin_config::ConfigHostComponent::new(runtime_config.config_providers()),
                 )?;
-            }
+                Some(builder.add_host_component(wasi_cloud::WasiCloudComponent)?)
+            };
```
lann, line 142: I think #1561 + `TriggerExecutor::configure_engine` should allow some of this to be simplified.

.github/workflows/build.yml
```diff
@@ -173,3 +172,19 @@ jobs:
           chmod +x `pwd`/target/release/spin
           export E2E_VOLUME_MOUNT="-v `pwd`/target/release/spin:/usr/local/bin/spin"
           make test-spin-up
+
+  lint-examples:
+    name: Lint Examples
```
rylev, line 177: Should we try to break these changes out to a separate PR? Seems useful and it would be a shame to block these changes on landing `wasi-http` support (which isn't really related).

crates/common/Cargo.toml
```diff
@@ -7,7 +7,8 @@ edition = { workspace = true }
 [dependencies]
 anyhow = "1.0"
 sha2 = "0.10"
+thiserror = "1.0.40"
```
rylev, line 10: nit: we don't seem to have a strong policy on versioning but is there any reason not to do `thiserror = "1.0"`?

crates/common/src/table.rs
```diff
@@ -52,14 +59,29 @@ impl<V> Table<V> {
     }
 
     /// Get a reference to the resource identified by the specified `key`, if it exists.
-    pub fn get(&self, key: u32) -> Option<&V> {
-        self.tuples.get(&key)
+    pub fn get(&self, key: u32) -> Result<&V, UnknownKeyError> {
```
rylev, line 62: Nit: this change (and the one to `get_mut`) seems strange to me. It seems much more fluent to return `Option`.

crates/common/Cargo.toml
dicej: I found https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277 pretty convincing, plus `cargo add` and `cargo update` do this by default, so I've been following that practice by default.

lann: IMO it doesn't really matter for binary crates, but I'd agree that precise is better for libs.

Mossaka (commented):
Awesome work! @dicej 

This is one of these dense PRs that have tons of information and impl details. I enjoyed the async wasi-http-rust examlple a lot! 

Made some clarifications to the wasi-http specific questions and one comment to add non-rust examples. I know it's might be difficult to do for other languages like TinyGo, but I can help if needed!

crates/wasi-cloud/src/http.rs
```diff
@@ -0,0 +1,723 @@
...
+            })
+            .transpose()
+    }
+
+    async fn finish_outgoing_stream(
+        &mut self,
+        this: types::OutgoingStream,
+        trailers: Option<types::Trailers>,
+    ) -> Result<()> {
+        // TODO: We should change the WIT file so this can return an I/O error instead of trapping
```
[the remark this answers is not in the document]
dicej: I've opened an issue here: https://github.com/WebAssembly/wasi-http/issues/37

crates/wasi-cloud/src/http.rs
```diff
@@ -0,0 +1,723 @@
...
+            Err(())
+        })
+    }
+
+    async fn new_outgoing_response(
+        &mut self,
+        status: types::StatusCode,
+        headers: types::Headers,
+    ) -> Result<types::OutgoingResponse> {
+        // TODO: What is supposed to happen if you create a new outgoing response with some headers and then edit
```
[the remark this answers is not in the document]
dicej: I've opened an issue requesting clarification: https://github.com/WebAssembly/wasi-http/issues/38

crates/wasi-cloud/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    wasmtime::component::bindgen!({
+        path: "../../wit/preview2",
+        world: "proxy",
+        async: true
+    });
+}
+
+pub struct WasiCloudComponent;
+
+impl HostComponent for WasiCloudComponent {
```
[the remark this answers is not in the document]
dicej: Yes, I was thinking we'd add all the `wasi-cloud-core` state to this component (including `wasi-keyvalue`), partly because I expect they'll need to share the same `poll` and `streams` tables.  I'm open to other ways of organizing this, though.

examples/wasi-http-rust-async/src/lib.rs
```diff
@@ -0,0 +1,290 @@
+wit_bindgen::generate!("proxy" in "../../wit/preview2");
```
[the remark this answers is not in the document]
dicej: I think we can definitely add async/await support to JS and Python.  I'm not sure how it would work with TinyGo, though, since I have limited experience with Go.  My understanding is that the most natural way to do concurrent I/O in Go is via goroutines, but I don't know how that's supported in TinyGo when targeting WASI or how we would hook into it with a `poll_oneoff`-based solution.  Do you know anything about that?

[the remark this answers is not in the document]
Mossaka: I don't. I'd like to find it out myself.

crates/wasi-cloud/src/lib.rs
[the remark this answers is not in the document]
lann: I think we'll want to bake some of these common preview2 types into `spin-core`. I'm expecting to see them in `wasmtime-wasi` 10 anyway.

crates/common/Cargo.toml
rylev: I don't have strong opinions here. The resource you posted is helpful (I've not seen it before), but it more strongly argues for avoiding just having the major version number which I agree with. If the crate is following semver, then not constraining the patch number should be fine.

crates/common/src/table.rs
rylev: I would prefer to revert it if it's not too big of a deal. I don't feel very strongly about it, but it definitely distracts from the rest of the PR.

dicej: I've reverted it.

crates/common/Cargo.toml
dicej: This is no longer relevant since I've reverted the commit that added the dependency.

dicej:
Quick update on this: we'll probably never merge this PR since @pchickey and @elliottt are porting it over to the latest `wasmtime-wasi` streams implementation, and that port will live in Wasmtime, not here.

However, I went ahead and rebased this anyway since @Mossaka and I are preparing [a presentation](https://wasmcon2023.sched.com/event/1PCMG/taking-wasi-cloud-core-for-a-spin-jiaxiao-zhou-microsoft-joel-dice-fermyon-technologies?iframe=no&w=100%&sidebar=yes&bg=no) that will include a demo of `wasi-http`, `wasi-key-value`, and `wasi-messaging`, and I'm not sure if the Wasmtime `wasi-http` implementation will be stable by then.

dicej:
Closing this per my comment above.

PR #1560, opened 2023-06-08
List plugins in help

itowlson:
This is an experiment to see if people like this experience or not.  Here's what it looks like on the proverbial my machine:

```
$ spin
spin 1.3.0-pre0 (b6a752f 2023-06-08)
The Spin CLI

USAGE:
    spin <SUBCOMMAND>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

SUBCOMMANDS:
    add          Scaffold a new component into an existing application
    build        Build the Spin application
    cloud        Commands for publishing applications to the Fermyon Cloud.
    deploy       Package and upload an application to the Fermyon Platform
    help         Print this message or the help of the given subcommand(s)
    js2wasm      A plugin to convert js files to Spin compatible modules
    login        Log into to the Fermyon Cloud.
    new          Scaffold a new application based on a template
    pluginify    
    plugins      Install/uninstall Spin plugins
    py2wasm      A plugin to convert Python applications to Spin compatible modules
    registry     Commands for working with OCI registries to distribute applications
    templates    Commands for working with WebAssembly component templates
    timer        IT IS I THE GREAT OZ
    up           Start the Spin application
    watch        Build and run the Spin application, rebuilding and restarting it when files
                     change
```

At the moment, I expect there to be a bug when the `cloud` plugin is installed, since that's a legit command as well.  I have a feeling I can solve this by _steeples fingers_ abolishing the `cloud` command and just always injecting the help entry - because Kate has put all the magic into the `external` side rather than the command side, I think it will behave correctly even if no actual `CloudCommand` exists.  But I will investigate that if we decide to go forward with this; one way or another it will be fixable.

### DIFF
--- a/src/bin/spin.rs
+++ b/src/bin/spin.rs
@@ -1,11 +1,12 @@
 use anyhow::Error;
-use clap::{CommandFactory, Parser, Subcommand};
+use clap::{CommandFactory, FromArgMatches, Parser, Subcommand};
 use is_terminal::IsTerminal;
 use lazy_static::lazy_static;
 use spin_cli::build_info::*;
+use spin_cli::commands::external::predefined_externals;
 use spin_cli::commands::{
     build::BuildCommand,
-    cloud::{CloudCommand, DeployCommand, LoginCommand},
+    cloud::{DeployCommand, LoginCommand},
     doctor::DoctorCommand,
     external::execute_external_subcommand,
     new::{AddCommand, NewCommand},
@@ -38,7 +39,37 @@ async fn _main() -> anyhow::Result<()> {
         )
         .with_ansi(std::io::stderr().is_terminal())
         .init();
-    SpinApp::parse().run().await
+
+    let plugin_help_entries = plugin_help_entries();
+
+    let mut cmd = SpinApp::command();
+    for plugin in &plugin_help_entries {
+        let subcmd = clap::Command::new(plugin.display_text())
+            .about(plugin.about.as_str())
+            .allow_hyphen_values(true)
+            .disable_help_flag(true)
+            .arg(
+                clap::Arg::new("command")
+                    .allow_hyphen_values(true)
+                    .multiple_values(true),
+            );
+        cmd = cmd.subcommand(subcmd);
+    }
+
+    if !plugin_help_entries.is_empty() {
+        cmd = cmd.after_help("* implemented via plugin");
+    }
+
+    let matches = cmd.clone().get_matches();
+
+    if let Some((subcmd, _)) = matches.subcommand() {
+        if plugin_help_entries.iter().any(|e| e.name == subcmd) {
+            let command = std::env::args().skip(1).collect();
+            return execute_external_subcommand(command, cmd).await;
+        }
+    }
+
+    SpinApp::from_arg_matches(&matches)?.run(cmd).await
 }
 
 fn print_error_chain(err: anyhow::Error) {
@@ -76,7 +107,6 @@ enum SpinApp {
     New(NewCommand),
     Add(AddCommand),
     Up(UpCommand),
-    Cloud(CloudCommand),
     // acts as a cross-level subcommand shortcut -> `spin cloud deploy`
     Deploy(DeployCommand),
     // acts as a cross-level subcommand shortcut -> `spin cloud login`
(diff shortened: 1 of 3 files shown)


mikkelhegn:
I think this is great.

One suggestion may be to call out more explicit which commands are part of a plugin. E.g.
```
    add        Scaffold a new component into an existing application
    build      Build the Spin application
    cloud      [plugin] Commands for publishing applications to the Fermyon Cloud.
    deploy     [plugin] Package and upload an application to the Fermyon Platform
    help       Print this message or the help of the given subcommand(s)
    js2wasm    [plugin] A plugin to convert js files to Spin compatible modules
```

That way if you are comparing screenshots in tutorials or somewhere else, it's clear if a comand is missing in your list, that it came from a plugin.

kate-goldenring:
`IT IS I THE GREAT OZ` 🤣

itowlson:
@kate-goldenring I am _really_ hoping I didn't leave that in the actual timer trigger sample.

mikkelhegn:
Although not as satisfying as using an emoji, I think either of theses works. If we can add a line of text at the bottom. something like `[p] subcommand from plugin` or `[*] subcommand from plugin`.

Using `[p]` is probably less "scarry"?

image

image

itowlson:
Where this is now at:

The UI looks like this (@mikkelhegn let me know if you have a preference for the bracketed form):

image

The various combinations of plugins / flags / help options / etc. that I have tested all seem to work, though it would be great to have someone else trying to break it (and it might be good to merge it only _after_ we cut 1.3, because if I have fouled up the entire command line then it would be good to have a bit of time to discover it).

The `cloud` command has been removed: `spin cloud` now goes directly to `execute_external_subcommand`.  The external subcommands module provides a declarative way to define these 'magic' plugins via the `PREDEFINED_EXTERNALS` array.

Plugins are now considered for the "similar commands" prompt:

```
ivan@hecate:~/testing/pyfie$ spin clod
Error: 'clod' is not a known Spin command. See spin --help.

The most similar command is:
    cloud

ivan@hecate:~/testing/pyfie$ spin ps2wasm
Error: 'ps2wasm' is not a known Spin command. See spin --help.

The most similar commands are:
    js2wasm
    py2wasm
```

The commits still need squashing.

Feedback and complaints very much welcome.

mikkelhegn:
I think this looks great, and gets the job done 👍🏻

mikkelhegn:
Not sure what's going on here?

The cloud command is still highlighted, even though I uninstalled the plugin.

image

kate-goldenring:
@itowlson should `deploy` and `login` always be decorated with `*`?

itowlson:
@mikkelhegn `cloud` is always shown, even if not installed; I believe this was the plan, and is why the `CloudCommand` currently exists.  It is asterisked because it is delivered via a plugin.  That the plugin isn't currently installed doesn't change the fact that it's delivered via a plugin!

That said, if the goal of the asterisks is more to highlight "places where your list may be different from other people's" rather than "is implemented via a plugin" then agree, we should take the asterisk off `cloud`, _and we should not show it even when the plugin is installed_ (which probably means we need to rephrase the explanation of the asterisk).

@kate-goldenring That's a good question... this is probably an artifact of the way I implemented the asterisking system (which I guess is true of Mikkel's `cloud` as well).  There is also a sense in which `deploy` and `login` are Spin commands with their own logic which _make use of_ a plugin.

But yeah... I guess it comes back to "is the idea to show which commands are implemented via plugins or to show which commands may not appear in docs or on other machines."  Let's define that, and maybe that will guide us on how to handle these cross-over cases.

kate-goldenring:
I agree that the indicator should be "is implemented via a plugin". i am now torn on deploy and login. It would be nice to be able to list your plugins and see by name how they map to the cli and that is not the case for deploy and login

itowlson:
Given Mikkel's comment and Kate being torn, can I propose we merge this as stands, with the option to refine it before (or after) 1.4?  That would give people the opportunity to shake out any bugs in parallel with UI refinement; otherwise we risk going back and forth for ages (given that every round of conversation takes 24 hours because _shakes puny fist at timezones_), and finally landing it without time to bake.

If that is okay, it would be super helpful to get a review so we can start moving it forward.

fibonacci1729 (approved):
LGTM! I agree with @itowlson that it's probably best to merge now and tease out the desired enhancements before 1.4

src/bin/spin.rs
```diff
@@ -76,7 +107,7 @@ enum SpinApp {
     New(NewCommand),
     Add(AddCommand),
     Up(UpCommand),
-    Cloud(CloudCommand),
+    // Cloud(CloudCommand),
```
kate-goldenring, line 110: Should we remove this comment entirely

kate-goldenring (approved):
Very nifty use of adding commands to clap! LGTM

src/bin/spin.rs
itowlson: oops, thanks, fixed

PR #1566, opened 2023-06-08
Add e2e tests for config variables

kate-goldenring:
Tests to configuration variables. One tests that the default variable value is used if none is set and the other (which cannot be run on `spin up` due to the way `deploy_args` works) asserts that the `--variables` deploy flag works as expected.

### DIFF
--- a/tests/testcases/mod.rs
+++ b/tests/testcases/mod.rs
@@ -14,6 +14,49 @@ fn get_url(base: &str, path: &str) -> String {
     format!("{}{}", base, path)
 }
 
+pub async fn config_variables_default_works(controller: &dyn Controller) {
+    async fn checks(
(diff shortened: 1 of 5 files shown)


tests/testcases/mod.rs
```diff
@@ -14,6 +14,98 @@ fn get_url(base: &str, path: &str) -> String {
     format!("{}{}", base, path)
 }
 
+// Cannot run on Spin controller as `spin up` does not take a variables flag
+#[allow(dead_code)]
+pub async fn config_variables_works(controller: &dyn Controller) {
```
rajatjindal, line 19: Hi Kate, I think this is the same test we talked about. we should probably move this test to cloud tests repo

fibonacci1729 (approved):

PR #1567, opened 2023-06-08
fix(tests): Distinguish between deploy and up args in integration tests

kate-goldenring:
`spin deploy` does not support the same set of args as `spin up`. This helps distinguish that in the tests, right now most apparently for `--sqlite`

### DIFF
--- a/tests/testcases/mod.rs
+++ b/tests/testcases/mod.rs
@@ -46,6 +46,10 @@ pub async fn key_value_works(controller: &dyn Controller) {
         .name("key-value".to_string())
         .appname(Some("key-value".to_string()))
         .template(None)
+        .up_args(vec![
+            "--key-value".to_string(),
+            format!("{init_key}={init_value}"),
+        ])
         .deploy_args(vec![
             "--key-value".to_string(),
             format!("{init_key}={init_value}"),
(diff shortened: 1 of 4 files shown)


tests/testcases/mod.rs
```diff
@@ -46,6 +46,10 @@ pub async fn key_value_works(controller: &dyn Controller) {
         .name("key-value".to_string())
         .appname(Some("key-value".to_string()))
         .template(None)
+        .up_args(vec![
```
itowlson, line 49: I am anxious about this duplication for the cases where we are able to keep the syntax the same.

I wonder if it would be practical to have something like a `shared_args` (bad name though) which is applied to both `up` _and_ `deploy`, and use `up_args` and `deploy_args` only for target-specific cases like `--variables`?

kate-goldenring: the reason i like the duplication is that i think we have discussed removing many of the deploy args in the future. This way, if that happens, it is clear what was up and what was deploy.

rylev (approved):

rajatjindal:
Hi Kate,

I am a little reluctant for this change as well. with this now user will need to be aware of what should go into "up_args", vs what should go into "deploy_args". 

In past, if there was a flag that is supported on `spin up` but not with `spin deploy`, we would add that as a separate test scenario (e.g. `header_dynamic_env_works` which made use of `--env`) and disable it with `spin deploy`. 

Similarly if there is a test which is going to be available only on `spin deploy`, it should be added to cloud tests repo only.

I hope this explaination makes sense.

kate-goldenring:
It sounds like we want to keep this as is. I think we can close this and resurface it if other contributors find a separation of arguments useful

PR #1569, opened 2023-06-09
Prevent concurrent `spin plugin update` commands

itowlson:
This is in support of the plugin badger; it provides protection in case a badger tries to kick off an update while an update triggered by a previous badger is still running.  (It is built into `spin plugins update` rather than the badger, because it is probably wise for users not to accidentally do it either.)

User experience:

```
$ spin plugins update
Error: Another plugin update operation is already in progress
```

### DIFF
--- a/crates/plugins/src/manager.rs
+++ b/crates/plugins/src/manager.rs
@@ -194,6 +194,65 @@ impl PluginManager {
         };
         Ok(plugin_manifest)
     }
+
+    pub async fn update_lock(&self) -> PluginManagerUpdateLock {
+        let lock = self.update_lock_impl().await;
+        PluginManagerUpdateLock::from(lock)
+    }
+
+    async fn update_lock_impl(&self) -> anyhow::Result<fd_lock::RwLock<tokio::fs::File>> {
+        let plugins_dir = self.store().get_plugins_directory();
+        tokio::fs::create_dir_all(plugins_dir).await?;
+        let file = tokio::fs::File::create(plugins_dir.join(".updatelock")).await?;
+        let locker = fd_lock::RwLock::new(file);
+        Ok(locker)
+    }
+}
+
+// We permit the "locking failed" state rather than erroring so that we don't prevent the user
+// from doing updates just because something is amiss in the lock system. (This is basically
+// falling back to the previous, never-lock, behaviour.) Put another way, we prevent updates
+// only if we can _positively confirm_ that another update is in progress.
+pub enum PluginManagerUpdateLock {
+    Lock(fd_lock::RwLock<tokio::fs::File>),
+    Failed,
+}
+
+impl From<anyhow::Result<fd_lock::RwLock<tokio::fs::File>>> for PluginManagerUpdateLock {
+    fn from(value: anyhow::Result<fd_lock::RwLock<tokio::fs::File>>) -> Self {
+        match value {
+            Ok(lock) => Self::Lock(lock),
+            Err(_) => Self::Failed,
+        }
+    }
+}
+
+impl PluginManagerUpdateLock {
+    pub fn lock_updates(&mut self) -> PluginManagerUpdateGuard<'_> {
+        match self {
+            Self::Lock(lock) => match lock.try_write() {
+                Ok(guard) => PluginManagerUpdateGuard::Acquired(guard),
+                Err(e) if e.kind() == std::io::ErrorKind::WouldBlock => {
+                    PluginManagerUpdateGuard::Denied
+                }
+                _ => PluginManagerUpdateGuard::Failed,
+            },
+            Self::Failed => PluginManagerUpdateGuard::Failed,
+        }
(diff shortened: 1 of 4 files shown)


crates/plugins/src/manager.rs
```diff
@@ -194,6 +194,48 @@ impl PluginManager {
...
+        let plugins_dir = self.store().get_plugins_directory();
+        tokio::fs::create_dir_all(plugins_dir).await?;
+        let file = tokio::fs::File::create(plugins_dir.join(".updatelock")).await?;
+        let locker = fd_lock::RwLock::new(file);
+        Ok(locker)
+    }
+}
+
+pub struct PluginManagerUpdateLock {
+    lock: anyhow::Result<fd_lock::RwLock<tokio::fs::File>>,
```
rylev, line 213: Why do we keep a `Result` here?

crates/plugins/src/manager.rs
```diff
@@ -194,6 +194,48 @@ impl PluginManager {
...
+            let guard = Some(lock.try_write());
+            PluginManagerUpdateGuard { guard }
+        } else {
+            PluginManagerUpdateGuard { guard: None }
+        }
+    }
+}
+
+pub struct PluginManagerUpdateGuard<'lock> {
+    guard: Option<std::io::Result<fd_lock::RwLockWriteGuard<'lock, tokio::fs::File>>>,
```
rylev, line 228: Same as above. Why do we keep the `io::Result`?

kate-goldenring: `try_write` returns `io::Result<RwLockWriteGuard<'_, T>>` and the result appears to be what we want to query to check for the `WouldBlock` error

crates/plugins/src/manager.rs
kate-goldenring: I am also not sure about this one. Looks like any errors in `update_lock_impl` would be unexpected and should be surfaced.

crates/plugins/src/manager.rs
```diff
@@ -194,6 +194,48 @@ impl PluginManager {
...
+        if let Ok(lock) = self.lock.as_mut() {
+            let guard = Some(lock.try_write());
+            PluginManagerUpdateGuard { guard }
+        } else {
+            PluginManagerUpdateGuard { guard: None }
+        }
+    }
+}
+
+pub struct PluginManagerUpdateGuard<'lock> {
```
lann, line 227: "Guards" can get some help from the compiler to make sure they aren't accidentally forgotten:

```suggestion
#[must_use]
pub struct PluginManagerUpdateGuard<'lock> {
```

crates/plugins/src/manager.rs
lann: It seems this is juggling 3 scenarios:

1. The lock is successfully created and acquired :tada:
2. The lock is successfully created but already held; assume that another update is already happening and bail
3. The lock cannot be created; continue without the lock

itowlson: What @lann said.  The idea is to forbid the update only if we _positively_ detect that another update is in progress - error cases are "proceed because at least it's no worse than what we have today."  We could of course discard the Results and Options and use our own enums to track the case, and that would certainly make it more self-documenting...!

crates/plugins/src/manager.rs
itowlson: Indeed.  As above, we could resolve "does it mean denied" at creation time, and create a custom enum a la `Denied / DidNotLock / Locked(guard)`.  I was worried I was already being profligate with new types but maybe that would be worth it.

crates/plugins/src/manager.rs
itowlson: Good call, thanks!

itowlson:
Changed the Result wrappers to custom enums to hopefully address feedback.

crates/plugins/src/manager.rs
```diff
@@ -194,6 +194,61 @@ impl PluginManager {
         };
         Ok(plugin_manifest)
     }
+
+    pub async fn update_lock(&self) -> PluginManagerUpdateLock {
```
rylev, line 198: I would expect this to return a `Result<PluginManagerUpdateLock>`. This would mean we wouldn't have to thread through the failure case to the lock guard which would only have two states: `Acquired` and `Denied`.

Do we have the three states because we explicitly want to ignore failed locks? That doesn't seem right...

itowlson: Yes.  The intention is to ignore failed locks, and fall back to the previous behaviour (as Lann described).  I'm happy to discuss whether this is the desired behaviour, but I wanted to avoid the situation of "if something goes wrong in the locking system then you can never update."

itowlson:
Can I get either a LGTM or guidance on changes still required please?  Thanks!

kate-goldenring (approved):
LGTM!

PR #1573, opened 2023-06-12
Allow re-initializing the connection store

rylev:
This PR changes the SQLite implementation to mirror the Key-Value implementation more closely by allowing re-initialization of a store of connections inside of `DynamicHostComponent::update_data`. This is strictly needed for Spin itself, but other implementations may wish to take advantage of this capability.

### DIFF
--- a/crates/trigger/src/runtime_config/sqlite.rs
+++ b/crates/trigger/src/runtime_config/sqlite.rs
@@ -27,8 +27,15 @@ pub(crate) fn build_component(
 struct SimpleConnectionsStore(HashMap<String, Arc<dyn Connection>>);
 
 impl ConnectionsStore for SimpleConnectionsStore {
-    fn get_connection(&self, database: &str) -> Option<Arc<(dyn Connection + 'static)>> {
-        self.0.get(database).cloned()
+    fn get_connection(
+        &self,
+        database: &str,
+    ) -> Result<Option<Arc<(dyn Connection + 'static)>>, spin_world::sqlite::Error> {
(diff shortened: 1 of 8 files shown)


lann:
This is mostly @dicej's code; I just refactored parts of it more recently.

crates/trigger/src/runtime_config/sqlite.rs
```diff
@@ -20,30 +18,41 @@ pub(crate) fn build_component(
...
-                        .with_context(|| format!("failed to execute sql from file '{file}'"))?;
-                } else {
-                    c.query(m, Vec::new())
-                        .with_context(|| format!("failed to execute statement: '{m}'"))?;
-                }
-            }
+    if statements.is_empty() {
+        return Ok(());
+    }
+    let Some(default) = databases.get("default") else { return Ok(()) };
```
dicej, line 42: Why do we return `Ok(())` if there is no default database here?

dicej (commented):
Looks good overall; just one question.

crates/trigger/src/runtime_config/sqlite.rs
rylev: I guess we should probably be emitting a warning or an error. This PR doesn't change this behaviour though so I think we should tackle this in a follow up.

rylev:
@dicej I pushed one more commit that allows getting a connection to fail since there will be implementations that could potentially fail. Can you take one more look?

dicej (approved):

PR #1577, opened 2023-06-12
Plugin update badger

### LINKED ISSUE #1543: Plugin upgrades
This is a micro-SIP for informing Spin users when there are new versions of their favourite plugins.  It is for discussion.

**Principles:**

* Do not break the user's existing workflow (habits, scripts, etc.).
* Do not block the user (make them wait on a prompt, make progress conditional on a download).
* Do not surprise the user.
* Do not badger the user.
* Respect CI environments.
  * A specific consideration is that a GH action that installs an older plugin should continue to work.

**Classification:**

* If the user is on a _stable_ line of a plugin (>=1.x), and there is a higher version available within the same major version, that upgrade is eligible.
* Upgrades within unstable lines (0.x) are questionable.
* Upgrades across major versions (1.x -> 2.x) are questionable.
* Upgrades with pre-release entries in their version (1.2.3-alpha.1) are ineligible.
  * We could refine this rule to make pre-releases eligible if the user is already on an earlier pre-release of that patch version, but pre-release text doesn't sort well and I'm not sure we have a sense of "earlier/later" other than the version.

**When to notify:**

* When a user runs a plugin, we check for the availability of an upgrade.
* If there is an _eligible_ upgrade:.
  * Has the user already been notified about this version?  If so, do nothing.
  * Has the user been notified about _any_ upgrade _from their current version of the plugin_ within the last B days?  If so, do nothing.
    * The nuance here is for the following reason.  Suppose the user is on 1.3.  On Monday, the user runs the plugin, and we detect and offer an upgrade to 1.4.  They do the upgrade.  On Tuesday, 1.4.1 comes out.  On Wednesday, the user runs the plugin again.  We should probably mention 1.4.1 even though we prompted on Monday.  But if the user chose _not_ to do the upgrade to 1.4, then we should _not_ keep badgering them.  So we need a sense of "resetting" the badger clock when they do an upgrade.
    * There is still a risk of badger if a plugin has high churn.  The user keeps running to keep up but there's always a new upgrade to be pestered about.  We can probably cross this bridge if we come to it.
  * Otherwise, print a message informing the user that the upgrade is available and telling them how to get it.
* If there is a _questionable_ upgrade:
  * TBD.  We could notify them using the same rules as eligible upgrades but with the message warning about breaking changes, and possibly with a longer badger clock.  Or we could not notify them.  Or something else.
* If there is an _ineligible_ upgrade, do nothing.
* This proposal doesn't define the badger timeout B, but that is easy to tweak during implementation and rollout.  My guess is we would start it at 7 or 14 days.

**How to notify:**

* Spin could notify _before_ handing off to the plugin.  This gets it out of the way.  But if the plugin produces non-trivial amounts of output, the message will disappear.  It also means we have to either make the user wait while we check for availability, or do a background check and notify the _next_ time they run the plugin.
* Spin could notify _after_ the plugin exits.  This would allow Spin to do checks in the background while the plugin is running, avoiding any startup delay, ensures that the message is visible even if the plugin has produced a lot of output, and prints it at the moment the user/terminal comes free, so they can action the upgrade offer immediately.  I am not sure if this increases complexity though.
* My preference is to notify after the plugin exits.
* In either case, notifications should be "print to terminal" only.  Spin should not display a prompt that requires a response in order to proceed.
* In either case, the upgrade check must be _silent_ and errors must not be reported to the user.
  * Spin should skip the upgrade check If the environment is not a terminal, because there is no user to notify so it's a waste of bandwidth.

**Privacy;**

* The user should be able to opt out of upgrade checks, via an environment variable or such like.
  * Perhaps the first time the user runs a plugin (any plugin), we should inform the user how to opt out.  That doesn't give them an opportunity to back out, though, unless we have the more complex rule of "the first time, we print the opt out notification; only the second time do we actually perform an upgrade check."  I am not sure how much of a concern this is.
* The _most_ information sent from the user's machine during the upgrade check should be the plugin name and version.  (I suspect we won't need to send that because we will look at the git repo, but then we get into repeated pulls, rate limiting, and similar "the whole reason why the `update` command exists" issues.  So I want to leave this much wiggle room.)

**Future refinements:**

* It could be worth allowing authors to tag releases as urgent, e.g. security patches.  Urgent releases would change the rules a bit:
  * Urgent releases within the 0.x line would be eligible.  (Not sure about across stable major versions.  In principle a security patch would be backported.  In practice, that's not always the case.)
  * Urgent releases would be notified even if user had recently been badgered.  (Not sure if we would repeatedly notify for the _same_ urgent release.  But if the user had recently declined an upgrade, and then there was an urgent release, that should be notified.)

itowlson:
Fixes #1543.

### DIFF
--- a/crates/plugins/src/badger/mod.rs
+++ b/crates/plugins/src/badger/mod.rs
@@ -6,6 +6,33 @@ use is_terminal::IsTerminal;
 
 const BADGER_TIMEOUT_DAYS: i64 = 14;
 
+// How the checker works:
+//
+// * The consumer calls BadgerChecker::start().  This immediately returns a task handle to
+//   the checker.  It's important that this be immediate, because it's called on _every_
+//   plugin invocation and we don't want to slow that down.
+// * In the background task, the checker determines if it needs to update the local copy
+//   of the plugins registry.  If so, it kicks that off as a background process.
+//   * The checker may determine while running the task that the user should not be prompted,
+//     or hit an error trying to kick things off the check.  In this case, it returns
+//     BadgerChecker::Precomputed from the task, ready to be picked up.
+//   * Otherwise, the checker wants to wait as long as possible before determining whether
+//     an upgrade is possible.  In this case it returns BadgerChecker::Deferred from the task.
+//     This captures the information needed for the upgrade check.
+// * When the consumer is ready to find out if it needs to notify the user, it awaits
+//   the task handle.  This should still be quick.
+// * The consumer then calls BadgerChecker::check().
+//   * If the task returned Precomputed (i.e. the task reached a decision before exiting),
+//     check() returns that precomputed value.
+//   * If the task returned Deferred (i.e. the task was holding off to let the background registry
+//     update do its work), it now loads the local copy of the registry, and compares the
+//     available versions to the current version.
+//
+// The reason for the Precomputed/Deferred dance is to handle the two cases of:
+// 1. There's no point waiting and doing the calculations because we _know_ we have a decision (or an error).
+// 2. There's a point to waiting because there _might_ be an upgrade, so we want to give the background
+//    process as much time as possible to complete, so we can offer the latest upgrade.
+
 pub enum BadgerChecker {
     Precomputed(anyhow::Result<BadgerUI>),
     Deferred(BadgerEvaluator),
@@ -41,13 +68,23 @@ impl BadgerChecker {
             match BadgerEvaluator::new(&name, &current_version, spin_version).await {
                 Ok(b) => {
                     if b.should_check() {
+                        // We want to offer the user an upgrade if one is available. Kick off a
+                        // background process to update the local copy of the registry, and
+                        // return the case that causes Self::check() to consult the registry.
                         BadgerEvaluator::fire_and_forget_update();
                         Self::Deferred(b)
                     } else {
+                        // We do not want to offer the user an upgrade, e.g. because we have
+                        // badgered them quite recently. Stash this decision for Self::check()
+                        // to return.
                         Self::Precomputed(Ok(BadgerUI::None))
                     }
                 }
-                Err(e) => Self::Precomputed(Err(e)),
+                Err(e) => {
+                    // We hit a problem determining if we wanted to offer an upgrade or not.
+                    // Stash the error for Self::check() to return.
+                    Self::Precomputed(Err(e))
+                }
             }
         })
     }
@@ -259,14 +296,21 @@ impl PluginVersion {
 
 impl std::fmt::Display for PluginVersion {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.write_fmt(format_args!("{}", self.version))
+        write!(f, "{}", self.version)
     }
 }
 
 pub enum BadgerUI {
+    // Do not badger the user. There is no available upgrade, or we have already badgered
+    // them recently about this plugin.
     None,
+    // There is an available upgrade which is compatible (same non-zero major version).
     Eligible(PluginVersion),
+    // There is an available upgrade but it may not be compatible (different major version
+    // or major version is zero).
     Questionable(PluginVersion),
+    // There is an available upgrade which is compatible, but there is also an even more
+    // recent upgrade which may not be compatible.
     Both {
         eligible: PluginVersion,
         questionable: PluginVersion,
--- a/crates/plugins/src/badger/mod.rs
+++ b/crates/plugins/src/badger/mod.rs
@@ -1,0 +1,274 @@
+mod store;
+
+use self::store::{BadgerRecordManager, PreviousBadger};
+use crate::manifest::PluginManifest;
+use is_terminal::IsTerminal;
+
+const BADGER_TIMEOUT_DAYS: i64 = 14;
+
+pub enum BadgerChecker {
+    Precomputed(anyhow::Result<BadgerUI>),
+    Deferred(BadgerEvaluator),
+}
+
+pub struct BadgerEvaluator {
+    plugin_name: String,
+    current_version: semver::Version,
+    spin_version: &'static str,
+    plugin_manager: crate::manager::PluginManager,
+    record_manager: BadgerRecordManager,
+    previous_badger: PreviousBadger,
+}
+
+impl BadgerChecker {
+    pub fn start(
+        name: &str,
+        current_version: Option<String>,
+        spin_version: &'static str,
+    ) -> tokio::task::JoinHandle<Self> {
+        let name = name.to_owned();
+
+        tokio::task::spawn(async move {
+            let current_version = match current_version {
+                Some(v) => v.to_owned(),
+                None => return Self::Precomputed(Ok(BadgerUI::None)),
+            };
+
+            if !std::io::stderr().is_terminal() {
+                return Self::Precomputed(Ok(BadgerUI::None));
+            }
+
+            match BadgerEvaluator::new(&name, &current_version, spin_version).await {
+                Ok(b) => {
+                    if b.should_check() {
+                        BadgerEvaluator::fire_and_forget_update();
+                        Self::Deferred(b)
+                    } else {
+                        Self::Precomputed(Ok(BadgerUI::None))
+                    }
+                }
+                Err(e) => Self::Precomputed(Err(e)),
+            }
+        })
+    }
+
+    pub async fn check(self) -> anyhow::Result<BadgerUI> {
+        match self {
+            Self::Precomputed(r) => r,
+            Self::Deferred(b) => b.check().await,
+        }
+    }
+}
+
+impl BadgerEvaluator {
+    async fn new(
+        name: &str,
+        current_version: &str,
+        spin_version: &'static str,
+    ) -> anyhow::Result<Self> {
+        let current_version = semver::Version::parse(current_version)?;
+        let plugin_manager = crate::manager::PluginManager::try_default()?;
+        let record_manager = BadgerRecordManager::default()?;
+        let previous_badger = record_manager.previous_badger(name, &current_version).await;
+
+        Ok(Self {
+            plugin_name: name.to_owned(),
+            current_version,
+            spin_version,
+            plugin_manager,
+            record_manager,
+            previous_badger,
+        })
+    }
+
+    fn should_check(&self) -> bool {
+        match self.previous_badger {
+            PreviousBadger::Fresh => true,
+            PreviousBadger::FromCurrent { when, .. } => has_timeout_expired(when),
+        }
+    }
+
+    fn fire_and_forget_update() {
+        if let Err(e) = Self::fire_and_forget_update_impl() {
+            tracing::info!("Failed to launch plugins update process; checking using latest local repo anyway. Error: {e:#}");
+        }
+    }
+
+    fn fire_and_forget_update_impl() -> anyhow::Result<()> {
+        let mut update_cmd = tokio::process::Command::new(std::env::current_exe()?);
+        update_cmd.args(["plugins", "update"]);
+        update_cmd.stdout(std::process::Stdio::null());
+        update_cmd.stderr(std::process::Stdio::null());
+        update_cmd.spawn()?;
+        Ok(())
+    }
+
+    async fn check(&self) -> anyhow::Result<BadgerUI> {
+        let available_upgrades = self.available_upgrades().await?;
+
+        // TO CONSIDER: skipping this check and badgering for the same upgrade in case they missed it
+        if self
+            .previous_badger
+            .includes_any(&available_upgrades.list())
+        {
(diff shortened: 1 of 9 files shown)


src/commands/external.rs
```diff
@@ -94,6 +108,58 @@ pub async fn execute_external_subcommand(
     Ok(())
 }
 
+async fn report_badger_result(badger: tokio::task::JoinHandle<BadgerChecker>) {
+    // This seems very unlikely to happen but just in case
```
lann, line 112: Could you expound on what "this" is? :slightly_smiling_face:

crates/plugins/src/badger/store.rs
```diff
@@ -0,0 +1,103 @@
...
+            .or_else(|| dirs::home_dir().map(|p| p.join(".spin")))
+            .ok_or_else(|| anyhow!("Unable to get local data directory or home directory"))?;
+        let db_path = base_dir.join(DEFAULT_STORE_DIR).join(DEFAULT_STORE_FILE);
+        Ok(Self { db_path })
+    }
+
+    fn load(&self) -> Vec<BadgerRecord> {
+        match std::fs::read(&self.db_path) {
+            Ok(v) => serde_json::from_slice(&v).unwrap_or_default(),
+            Err(_) => vec![],
```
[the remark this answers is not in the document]
itowlson: I did ponder that, but couldn't think of anything useful to do at this point.  If the shape changes incompatibly, our choices are:

1. Adapt it. We can't do that until the change happens, that would be a matter for this function anyway.
2. Clear the decks and go "we can't understand this ancient file, let's assume we need to re-check everything," which is what this would do if left unchanged.

Could you expand on what additional error handling you're thinking of?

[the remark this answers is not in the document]
lann: I suppose if it is expected to be immediately overwritten there isn't much to do.

itowlson:
Thanks for the feedback - I added a bunch of comments that hopefully clarify things, but of course please let me know if it's still cryptic.

crates/plugins/src/badger/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+        });
+        let compatible_plugin_versions =
+            compatible_manifests.filter_map(|m| PluginVersion::try_from(m, &latest_version));
+        let considerable_manifests = compatible_plugin_versions
+            .filter(|pv| !pv.is_prerelease() && pv.is_higher_than(&self.current_version))
+            .collect::<Vec<_>>();
+
+        let (eligible_manifests, questionable_manifests) = if self.current_version.major == 0 {
+            (vec![], considerable_manifests)
+        } else {
```
kate-goldenring, line 194: Given that none of our plugins are v1.0, this badger will always throw "but may not be backward compatible with your current plugin", even if updating a patch version. For example, if i run `spin cloud --help` with `cloud 0.1.0` installed, I get:
```sh
This plugin can be upgraded. Version 0.1.1 is available,
but may not be backward compatible with your current plugin.
```
According to semver, it should be backwards compat

crates/plugins/src/badger/mod.rs
```diff
@@ -0,0 +1,318 @@
...
+                .partition(|pv| pv.version.major == self.current_version.major)
+        };
+
+        let highest_eligible_manifest = eligible_manifests
+            .into_iter()
+            .max_by_key(|pv| pv.version.clone());
+        let highest_questionable_manifest = questionable_manifests
+            .into_iter()
+            .max_by_key(|pv| pv.version.clone());
+
```
kate-goldenring, line 206: According to the definition of `Both` being that the questionable is higher than the elgible, do we need a check here of whether `highest_questionable_manifest > highest_eligible_manifest` and if false set `highest_questionable_manifest` to false?

crates/plugins/src/badger/store.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub fn includes_any(&self, version: &[&semver::Version]) -> bool {
+        version.iter().any(|version| self.includes(version))
+    }
+}
+
+impl BadgerRecordManager {
+    pub fn default() -> anyhow::Result<Self> {
+        let base_dir = dirs::cache_dir()
+            .or_else(|| dirs::home_dir().map(|p| p.join(".spin")))
+            .ok_or_else(|| anyhow!("Unable to get local data directory or home directory"))?;
```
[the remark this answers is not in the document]
kate-goldenring: I think this should reuse `default_data_dir` from `data_dir.rs`. Otherwise it will also mess up any colocation of data for package managers

crates/plugins/src/badger/store.rs
```diff
@@ -0,0 +1,103 @@
...
+        let new = BadgerRecord {
+            name: name.to_owned(),
+            badgered_from: from.clone(),
+            badgered_to: to.iter().map(|v| <semver::Version>::clone(v)).collect(),
+            when: chrono::Utc::now(),
+        };
+
+        // There is a potential race condition here if someone runs two plugins at
+        // the same time. As this is unlikely, and the worst outcome is that a user
+        // misses a badger or gets a double badger, let's not worry about it for now.
```
kate-goldenring, line 95: won't the PR to prevent simultaneous updates help with this too?

crates/plugins/src/badger/store.rs
```diff
@@ -0,0 +1,103 @@
...
+        let mut existing = self.load();
+        match existing.iter().position(|r| r.name == name) {
+            Some(index) => existing[index] = new,
+            None => existing.push(new),
+        };
+        _ = self.save(existing);
```
kate-goldenring, line 101: Why do we need anything more than latest badger in the file? I may have missed where we are reading previous badgers

kate-goldenring (commented):
The implementation is looking great. Just a few clarifying questions

src/commands/external.rs
```diff
@@ -94,6 +108,58 @@ pub async fn execute_external_subcommand(
...
+                "This plugin can be upgraded.",
+                "Version {to} is available and compatible."
+            );
+            eprintln!("To upgrade, run `{}`.", to.upgrade_command());
+        }
+        Ok(spin_plugins::badger::BadgerUI::Questionable(to)) => {
+            eprintln!();
+            terminal::info!("This plugin can be upgraded.", "Version {to} is available,");
+            eprintln!("but may not be backward compatible with your current plugin.");
+            eprintln!("To upgrade, run `{}`.", to.upgrade_command());
```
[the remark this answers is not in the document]
itowlson: Logging is used for errors, and `eprintln/terminal` for user output.  We don't want to display errors by default (this is meant to be a silent, unintrusive process), so we log those at `info`.  `terminal` is used when we want colour highlighting, `eprintln` for continuation lines that don't want additional colouring.

crates/plugins/src/badger/mod.rs
itowlson: I thought semver had an exception for major version 0?  If it doesn't, we've been doing it wrong... _grin_

crates/plugins/src/badger/mod.rs
itowlson: `considerable_manifests` contains only manifests whose version is >= current version.  If HEV exists then its major version is == current major version.  If HQV exists, its major version is != current major version.  But since HQV major must be >= current major version, this means HQV major > current major.  And since current major == HEV major, if both exist then HQV major > HEV major.

image

I can add a comment explaining this if it would be useful.

itowlson:
@kate-goldenring for some reason I can't reply to the comment on lines 44-46

> I think this should reuse default_data_dir from data_dir.rs. Otherwise it will also mess up any colocation of data for package managers

I chose `cache_dir` because this is "does not matter if the user blows it away" material - it's purely advisory.  But yeah it did feel a bit odd having it in a different place from the plugins.  Open to opinion.

crates/plugins/src/badger/store.rs
itowlson: Sadly, no - the potential race here is the load-save cycle for the notification records themselves.

crates/plugins/src/badger/store.rs
itowlson: I don't quite understand the question, sorry.  We store the latest badger per plugin, because the badger clock runs per plugin, not globally (because we only badger people about the plugin they are running, not all plugins).

kate-goldenring: Sorry i confused myself and thought we had a file per plugin for a second.

kate-goldenring:
> @kate-goldenring for some reason I can't reply to the comment on lines 44-46
> 
> > I think this should reuse default_data_dir from data_dir.rs. Otherwise it will also mess up any colocation of data for package managers
> 
> I chose `cache_dir` because this is "does not matter if the user blows it away" material - it's purely advisory. But yeah it did feel a bit odd having it in a different place from the plugins. Open to opinion.

@itowlson that seems like a reasonable distinction to me. I am just thinking of when people uninstall Spin we have multiple directories that need cleanup but registry uses it anyways so SGTM

crates/plugins/src/badger/mod.rs
kate-goldenring: just cackled 🤣  thank you for laying all this out! I don't think we need to comment it

crates/plugins/src/badger/mod.rs
kate-goldenring: Indeed you are right: "
Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable."

I guess my point is, are we concerned with every badger saying "may not be compatible". I guess this is a given, but i do think for our plugins it would be worth visiting what it would take to see them to `1.0`

kate-goldenring (approved):

crates/plugins/src/badger/mod.rs
itowlson: Yes, and this perhaps comes back to the discussion we had about stability vs maturity - it would be great to try to get our plugins to 1.x _even if_ they are not fully mature, so that we can at least send meaningful signals about interface stability and compatibility.  But the risk then is we have to do too many major revs in the early months...

crates/plugins/src/badger/store.rs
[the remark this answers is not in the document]
lann: @kate-goldenring note this is under `dirs::cache_dir`, not `data_dir`.

@itowlson I had this in the back of my mind when I wrote https://github.com/fermyon/spin/pull/1584#issuecomment-1589945266 :shrug: Either way I'm fine with punting for now.

crates/plugins/src/badger/store.rs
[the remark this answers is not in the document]
lann: Possibly worth a comment? Silently ignoring errors usually activates my bug-dar.

crates/terminal/src/lib.rs
```diff
@@ -91,6 +91,15 @@ macro_rules! error {
     }};
 }
 
+#[macro_export]
+macro_rules! info {
+    ($highlight:expr, $($arg:tt)*) => {{
+        $crate::ceprint!($crate::colors::bold_cyan(), $highlight);
```
lann, line 97: Not sure how I feel about `step!` going to stdout while `error!` and `info!` go to stderr. For text processing tools the distinction is clear to me: stdout is for "data" that you might be piping to another command, while stderr is for stuff that is only for human consumption. It isn't obvious how/if that distinction applies here.

crates/plugins/src/badger/mod.rs
```diff
@@ -0,0 +1,318 @@
+mod store;
+
+use self::store::{BadgerRecordManager, PreviousBadger};
+use crate::manifest::PluginManifest;
+use is_terminal::IsTerminal;
+
+const BADGER_TIMEOUT_DAYS: i64 = 14;
+
+// How the checker works:
```
lann, line 9: Thanks, very helpful!

src/commands/external.rs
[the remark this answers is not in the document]
lann: https://github.com/fermyon/spin/pull/1577#discussion_r1229691554

crates/terminal/src/lib.rs
itowlson: I sent the update badger to stderr because I explicitly didn't want it mixed up with plugin output that might be getting consumed by a program (e.g. a plugin that outputs JSON).  I can be more explicit and call this `einfo!` or something if you prefer - that might also make it clear the follow-up lines should use `eprintln` not `println`.  Let me know if you want me to make that or a similar change.

lann: Ah yes that makes perfect sense for :badger:. `einfo` would make sense to me, especially to help differentiate from `step`.

itowlson:
@lann added a comment on the error swallowing, and changed to `terminal::einfo!`.  I think that covers all your remaining comments - do you want to re-review, or are you happy for me to merge on that basis?

lann:
Merge away!

PR #1582, opened 2023-06-13
Small simplifications for spin-app

rylev:
Small tweaks to `spin-app` after some spelunking. The following changes were made:
* Add some docs. Even though these are internal types, it still helps to have some docs about *why* they're there.
* Unrolled some type aliases. I find type aliases that are not reused generally make things harder to understand. 
* Made `DynSafeDynamicHostComponent::update_data_any` take just a `&mut Data` instead of a `&mut Box<dyn Any + Send>`. There doesn't seem to be any need for that more complicated type
* Made `DynamicHostComponents` `pub(crate)`

### DIFF
--- a/crates/app/src/host_component.rs
+++ b/crates/app/src/host_component.rs
@@ -12,6 +12,11 @@ use crate::{App, AppComponent};
 ///
 /// This extends [`HostComponent`] to support per-[`AppComponent`] dynamic
 /// runtime configuration.
+///
+/// Dynamic host components differ from regular host components in that they can be
+/// reference and thus (re-)configured dynamically (on a per trigger instance basis)
+/// where as regular host components are static and do not change over the
+/// runtime of a Spin instance.
 pub trait DynamicHostComponent: HostComponent {
     /// Called on [`AppComponent`] instance initialization.
     ///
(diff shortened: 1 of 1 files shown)


lann:
> There doesn't seem to be any need for that more complicated type

Probably not. There is a lot of "tweak it till it works then never touch it again" going on in there :grimacing:

crates/app/src/host_component.rs
```diff
@@ -12,6 +12,11 @@ use crate::{App, AppComponent};
 ///
 /// This extends [`HostComponent`] to support per-[`AppComponent`] dynamic
 /// runtime configuration.
+///
+/// Dynamic host components differ from regular host components in that they can be
+/// reference and thus (re-)configured dynamically (on a per trigger instance basis)
+/// where as regular host components are static and do not change over the
+/// runtime of a Spin instance.
```
lann, line 19:
```suggestion
/// configured on a per-component basis.
```

lann:
> Unrolled some type aliases. I find type aliases that are not reused generally make things harder to understand.
:+1: I probably had these just because I was changing them so frequently trying to understand how to fit things together.

lann (approved):
Thanks!

PR #1583, opened 2023-06-13
Upgrade to wasmtime 9.0

lann:


crates/core/src/limits.rs
```diff
@@ -17,17 +18,27 @@ impl ResourceLimiterAsync for StoreLimitsAsync {
...
         if can_grow {
             self.memory_consumed =
                 (self.memory_consumed as i64 + (desired as i64 - current as i64)) as u64;
         }
-        can_grow
+        Ok(can_grow)
     }
 
-    async fn table_growing(&mut self, _current: u32, desired: u32, _maximum: Option<u32>) -> bool {
-        !matches!(self.max_table_elements, Some(limit) if desired > limit)
```
lann, line 30: "Who wrote this nonsense?!" <...checks blame...> "I did, of course."

lann: Also I screwed up this logic just now. Fixing + adding tests...

crates/core/src/preview1.rs
```diff
@@ -0,0 +1,193 @@
...
+use wasi_common_preview1::{
+    dir::{OpenResult, ReaddirCursor, ReaddirEntity},
+    file::{Advice, FdFlags, FileType, Filestat, OFlags},
+    Error, ErrorExt, SystemTimeSpec, WasiDir, WasiFile,
+};
+
+pub struct ReadOnlyDir(pub Box<dyn WasiDir>);
+
+#[async_trait::async_trait]
+impl WasiDir for ReadOnlyDir {
```
lann, line 12: @dicej

crates/core/src/limits.rs
```diff
@@ -45,3 +60,29 @@ impl StoreLimitsAsync {
...
+
+    #[tokio::test]
+    async fn test_store_limits_memory() {
+        let mut limits = StoreLimitsAsync {
+            max_memory_size: Some(65536),
+            ..Default::default()
+        };
+        assert!(limits.memory_growing(0, 65536, None).await.unwrap());
+        assert_eq!(limits.memory_consumed, 65536);
+        assert!(!limits.memory_growing(65536, 131072, None).await.unwrap());
```
rylev, line 76:
```suggestion
        assert!(!limits.memory_growing(65536, 131072, None).await.unwrap());
        assert_eq!(limits.memory_consumed, 65536);
```

crates/core/src/store.rs
```diff
@@ -336,29 +288,42 @@ impl StoreBuilder {
...
+                    }
+                    ctx.push_preopened_dir(dir, path)?;
+                }
+                Wasi::Preview2(ctx) => {
+                    let mut dir =
+                        Box::new(wasmtime_wasi_preview2::dir::Dir::from_cap_std(cap_std_dir)) as _;
+                    if !writable {
+                        dir = Box::new(wasi_preview2::dir::ReadOnlyDir(dir));
+                    }
+                    ctx.push_preopened_dir(dir, path)?;
```
rylev, line 322: nit: can this be moved out of the match statement to make it clear that that's what's happening regardless of the which wasi preview.

crates/core/src/preview1.rs
```diff
@@ -0,0 +1,193 @@
+use std::{any::Any, path::PathBuf};
```
rylev, line 1: Would it make sense to write a module doc explaining what these implementations are for?

crates/core/src/store.rs
rylev: Ah sorry! I read them as both trait objects of the same type. Whoops!

lann:
This last patch fixes it for me locally. tldr I had deleted all the new preview2 WasiFile methods but I didn't realize there were a bunch of preview1 methods that were removed with preview2, and that `WasiFile` provides default `Error::badf` impls for most of them...

crates/core/src/preview1.rs
lann: ...and adding some wrapper methods that were removed in preview2, apparently.

dicej (approved):

PR #1584, opened 2023-06-13
Move resolve_file_path to spin_common

fibonacci1729:
Moves `manifest::resolve_file_path` out of the CLI and into `spin-common`.  Currently this code is duplicated between `Spin` and the cloud plugin.

Will update `cloud-plugin` if we decide this fits the "common" criteria.

### DIFF
--- a/crates/common/src/lib.rs
+++ b/crates/common/src/lib.rs
@@ -10,5 +10,6 @@
 
 pub mod arg_parser;
 pub mod data_dir;
+pub mod paths;
 pub mod sha256;
 pub mod sloth;
--- a/crates/common/src/paths.rs
+++ b/crates/common/src/paths.rs
@@ -1,11 +1,14 @@
-use anyhow::anyhow;
+//! Resolves a file path to a manifest file
+
+use anyhow::{anyhow, Result};
 use std::path::{Path, PathBuf};
 
-use crate::opts::DEFAULT_MANIFEST_FILE;
+/// The name given to the default manifest file.
+pub const DEFAULT_MANIFEST_FILE: &str = "spin.toml";
 
 /// Resolves a manifest path provided by a user, which may be a file or
 /// directory, to a path to a manifest file.
-pub(crate) fn resolve_file_path(provided_path: impl AsRef<Path>) -> anyhow::Result<PathBuf> {
+pub fn resolve_manifest_file_path(provided_path: impl AsRef<Path>) -> Result<PathBuf> {
     let path = provided_path.as_ref();
 
     if path.is_file() {
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -37,7 +37,7 @@ pub struct BuildCommand {
 
 impl BuildCommand {
     pub async fn run(self) -> Result<()> {
-        let manifest_file = crate::manifest::resolve_file_path(&self.app_source)?;
+        let manifest_file = spin_common::paths::resolve_manifest_file_path(&self.app_source)?;
         spin_build::build(&manifest_file, &self.component_id).await?;
 
         if self.up {
--- a/src/commands/doctor.rs
+++ b/src/commands/doctor.rs
@@ -26,7 +26,7 @@ pub struct DoctorCommand {
 
 impl DoctorCommand {
     pub async fn run(self) -> Result<()> {
-        let manifest_file = crate::manifest::resolve_file_path(&self.app_source)?;
+        let manifest_file = spin_common::paths::resolve_manifest_file_path(&self.app_source)?;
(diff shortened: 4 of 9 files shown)


kate-goldenring (approved):
Thanks for reducing the duplication. Only nit suggestion would be to tag onto this a rename of `manifest.rs` to `manifest_path.rs` since when using the `common` crate it seems like this would give information about the Spin manifest object even though that is a part of a different crate

kate-goldenring:
@rylev thats a good point that there could be clarity around the use of the `common` crate. I don't see it as unrelated to spin, for example the `data_dir` module. But maybe we need a new crate, say `utils` for non-spin related utilities, while common could still be tied to spin.

fibonacci1729:
@rylev that's a good point but (as @kate-goldenring says above) `common` isn't entirely disjoint from `Spin` ATM. Alternatively, `spin-manifest` might be a better home?

kate-goldenring:
@fibonacci1729 whether it fits in `manifest` is an interesting point. I think it is helpful to think of whether it is central to the Spin runtime or just the spin CLI. It seems like the latter to me. This feels like an important distinction because i wonder whether a user of the runtime would need it, for example, the containerd-wasm-shims use the `manifest` crate but not the CLI and wouldn't need to use this function. On the other hand, `common` seems very tied to the CLI of spin, (arg parsing, local data paths, sloth warnings) so i think it should live in common. And we may consider renaming or rebranding common as CLI focused.

fibonacci1729:
@kate-goldenring I agree, that's how i was/am thinking about this too. Personally `spin-common` makes the most sense to me.

lann:
I would maybe merge this with the `data_dir` module and rename to `spin_paths` or something.

fibonacci1729:
@lann done

lann (approved):

crates/common/src/spin_paths.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use anyhow::{anyhow, Result};
+use std::path::{Path, PathBuf};
+
+/// The name given to the default manifest file.
+pub const DEFAULT_MANIFEST_FILE: &str = "spin.toml";
+
+/// Resolves a manifest path provided by a user, which may be a file or
+/// directory, to a path to a manifest file.
+pub fn resolve_file_path(provided_path: impl AsRef<Path>) -> Result<PathBuf> {
```
itowlson, line 11: This is specific to application manifests.  If we move it, the name needs to change to be manifest-specific.

But... why move it at all?  It fits nicely in the manifest crate.  If you want it for plugins, can't you import the manifest crate and use that?

itowlson (changes requested):

crates/common/src/spin_paths.rs
fibonacci1729: It doesn't live in the manifest crate, it lives in spin-cli.

Moving to spin-manifest is definitely an option, I don't have a strong preference either way.

fibonacci1729: Renamed to `resolve_manifest_file_path`

itowlson: Sorry, should have read other comments first - and apologies for the mistake about the current location.  I take the point about the `manifest` crate being the abstract description of the lockfile, and agree that it would not, in fact, fit there.  Given that, I think you're right that there's no clearer home than `common` (possibly `loader` but that's really concerned with loading, not CLI conventions).

I still have a requested change though, and that is _not_ to combine this with `data_dir`.  There are two separate concerns here:

1. What is the structure of Spin's own on-disk data (registry cache, template store, plugin store) - the internal layout of Spin's private info, abstracted away from the user
2. What are the UI conventions of the Spin CLI (in regard to interpreting paths entered by users) - specific to user interactions

I'd prefer not to mix those concerns into a single "paths" module.  (We could rename the `data_dir` module to something like `layout` over time, but that's not a matter for this PR.)

(also, if I _did_ mix the two, I would call it `paths` not `spin_paths`... we already know it's Spin _grin_)

Apologies again for the mistake and for the back-and-forth.

fibonacci1729: So TL;DR the change you are requesting is to: 

1. Leave `data_dir` unchanged
2. Move `resolve_manifest_file_path` to `paths.rs` 

I'm fine with that.

itowlson: Works for me - thanks!

fibonacci1729: @itowlson done!

itowlson (approved):
LGTM and apologies again for the initial bad comment

PR #1585, opened 2023-06-13
feat: make bindle optional in spin-loader crate

Mossaka:
Add local and bindle features to spin-loader so that spin embedder can opt-out of bindle dep completely

Signed-off-by: Jiaxiao Zhou (Mossaka) <duibao55328@gmail.com>

### DIFF
--- a/crates/loader/Cargo.toml
+++ b/crates/loader/Cargo.toml
@@ -7,7 +7,7 @@ edition = { workspace = true }
 [dependencies]
 anyhow = "1"
 async-trait = "0.1.52"
-bindle = { workspace = true }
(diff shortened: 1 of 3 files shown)


crates/loader/Cargo.toml
```diff
@@ -32,3 +32,8 @@ tokio-util = "0.6"
 toml = "0.5"
 tracing = { workspace = true }
 walkdir = "2.3.2"
+
+[features]
+default = ["local"]
```
lann, line 37: Let's make the default backward-compatible for now:

```suggestion
default = ["bindle", "local"]
```

lann (approved):
Thanks, LGTM

PR #1587, opened 2023-06-13
Adjust build script dependencies

alexcrichton:
* Always depend on the `BUILD_SPIN_EXAMPLES` env var instead of conditionally to always watch for changes.

* Don't depend on the entire directory tree of examples which includes the `target` directory in each example. This currently causes `cargo build` to always rebuild when issued, but `cargo-target-dep` is already taking care of recording dependencies by reading the `*.d` output of Cargo.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -30,9 +30,9 @@ fn main() {
     let build_spin_tests = env::var("BUILD_SPIN_EXAMPLES")
         .map(|v| v == "1")
         .unwrap_or(true);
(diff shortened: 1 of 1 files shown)


rylev (approved):
👍 @alexcrichton you'll need to sign and sign-off your commits with `git commit -sS`

alexcrichton:
Ok I think I figured out the right bits for signing and such

radu-matei (approved):

PR #1588, opened 2023-06-14
Add constructor for HostComponentInitData

itowlson:
This is a purely technical change to better support embedding scenarios.  At the moment, an embedder wanting to leverage `TriggerExecutorBuilder::build()` can only provide empty initialisation data; this provides for embedders to support `spin up --key-value`-style storage setup features.

### DIFF
--- a/crates/trigger/src/lib.rs
+++ b/crates/trigger/src/lib.rs
@@ -161,6 +161,21 @@ pub struct HostComponentInitData {
     sqlite: Vec<String>,
 }
 
+impl HostComponentInitData {
+    /// Create an instance of `HostComponentInitData`.  `key_value_init_values`
+    /// will be added to the default key-value store; `sqlite_init_statements`
+    /// will be run against the default SQLite database.
+    pub fn new(
+        key_value_init_values: impl Into<Vec<(String, String)>>,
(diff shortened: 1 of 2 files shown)


crates/trigger/src/lib.rs
```diff
@@ -161,6 +161,21 @@ pub struct HostComponentInitData {
...
+    /// Create an instance of `HostComponentInitData`.  `key_value_init_values`
+    /// will be added to the default key-value store; `sqlite_init_statements`
+    /// will be run against the default SQLite database.
+    pub fn new(
+        key_value_init_values: &[(String, String)],
+        sqlite_init_statements: &[String],
+    ) -> Self {
+        Self {
+            kv: key_value_init_values.to_vec(),
+            sqlite: sqlite_init_statements.to_vec(),
```
lann, line 174: A little more flexible (and equivalent for those specific argument types):

```suggestion
        key_value_init_values: impl Into<Vec<(String, String)>>,
        sqlite_init_statements: impl Into<Vec<String>>,
    ) -> Self {
        Self {
            kv: key_value_init_values.into(),
            sqlite: sqlite_init_statements.into(),
```

itowlson: Oh, of course - thanks!

PR #1590, opened 2023-06-15
fix(build): Remove git worktree requirements for build to succeed

### LINKED ISSUE #1589: Cannot Build Spin from Released Source Code
Building Spin from a Spin release source code fails for all releases from v1.2.0 and onwards due to required Git information needed in `build.rs`:

```rs
fn main() {
    vergen::EmitBuilder::builder()
        .build_date()
        .build_timestamp()
        .cargo_target_triple()
        .cargo_debug()
        .git_branch()
        .git_commit_date()
        .git_commit_timestamp()
        .git_sha(true)
        .fail_on_error()
        .emit()
        .expect("failed to extract build information");
...
```

This was added in https://github.com/fermyon/spin/pull/1492/files in order to populate more information to plugins. We should set empty defaults for these values if the do not exist.

## To repro

```sh
curl -LRO https://github.com/fermyon/spin/archive/refs/tags/v1.3.0.tar.gz
tar -xvf v1.3.0.tar.gz
cd spin-1.3.0
cargo build
```
Output

```sh
error: failed to run custom build command for `spin-cli v1.3.0 (/Users/kagold/Programs/temp/spin-1.3.0)`

Caused by:
  process didn't exit successfully: `/Users/kagold/Programs/temp/spin-1.3.0/target/release/build/spin-cli-022ecc32c217a5a5/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'failed to extract build information: not within a suitable 'git' worktree!', build.rs:28:10
warning: build failed, waiting for other jobs to finish...
error: failed to compile `spin-cli v1.3.0 (/Users/kagold/Programs/temp/spin-1.3.0)`, intermediate artifacts can be found at `/Users/kagold/Programs/temp/spin-1.3.0/target`
```

kate-goldenring:
fixes #1589

The git environment variables passed to plugins will be set to [`vergen`'s defaults](https://github.com/rustyhorde/vergen/blob/8e0f393f59696b68b533e2fc94eba22ff7c2e377/vergen/src/feature/git/gix.rs#L244) if Spin is being built from source code rather than a git environment.

Aside: The fact that the Rust SDK was preventing successful builds of Spin provides impetus for maybe moving the SDKs to their own repositories

### DIFF
--- a/sdk/rust/build.rs
+++ b/sdk/rust/build.rs
@@ -8,17 +8,14 @@ fn main() {
         format!("-{pre}")
     };
 
-    let output = Command::new("git")
+    let commit = Command::new("git")
         .arg("rev-parse")
         .arg("HEAD")
         .output()
-        .expect("failed to execute `git`");
-
-    let commit = if output.status.success() {
-        String::from_utf8(output.stdout).unwrap()
-    } else {
-        panic!("`git` failed: {}", String::from_utf8_lossy(&output.stderr));
-    };
+        .ok()
+        .filter(|o| o.status.success())
+        .map(|o| String::from_utf8_lossy(&o.stdout).to_string())
+        .unwrap_or(String::from("unknown"));
 
     let commit = commit.trim();
 
--- a/build.rs
+++ b/build.rs
@@ -14,6 +14,9 @@ const RUST_OUTBOUND_REDIS_INTEGRATION_TEST: &str = "tests/outbound-redis/http-ru
 const TIMER_TRIGGER_INTEGRATION_TEST: &str = "examples/spin-timer/app-example";
 
 fn main() {
+    // Extract environment information to be passed to plugins.
+    // Git information will be set to defaults if Spin is not
+    // built within a Git worktree.
     vergen::EmitBuilder::builder()
         .build_date()
         .build_timestamp()
@@ -23,7 +26,6 @@ fn main() {
         .git_commit_date()
         .git_commit_timestamp()
         .git_sha(true)
-        .fail_on_error()
         .emit()
         .expect("failed to extract build information");
 


sdk/rust/build.rs
```diff
@@ -17,7 +17,7 @@ fn main() {
     let commit = if output.status.success() {
         String::from_utf8(output.stdout).unwrap()
     } else {
```
endocrimes, line 19: Can't comment on the correct line, but if git isn't installed I think line 15 will fail if `git` isn't available in your PATH

sdk/rust/build.rs
```diff
@@ -8,16 +8,19 @@ fn main() {
...
+                String::from_utf8(output.stdout).unwrap()
+            } else {
+                // Spin is not a part of a Git worktree
+                String::from("unknown")
+            }
+        }
+        Err(_e) => {
+            // Git is not installed
+            String::from("unknown")
+        }
```
rylev, line 23:
```suggestion
let commit = Command::new("git")
        .arg("rev-parse")
        .arg("HEAD")
        .output()
        .filter(|o| o.status.success())
        .map(|o| String::from_utf8_lossy(output.stdout))
        .unwrap_or("unknown");
   
```

kate-goldenring: Thanks @rylev. This is much more concise! I slightly modified the mapping to first map `Result<Output, Error>` to an option.

sdk/rust/build.rs
```diff
@@ -8,17 +8,14 @@ fn main() {
         format!("-{pre}")
     };
 
-    let output = Command::new("git")
+    let commit: String = Command::new("git")
```
rylev, line 11: Nit: is there any reason for the `: String` annotation?

kate-goldenring: No. Thanks for pointing it out. it was remnant from me playing around with the mapping.

rylev (approved):

PR #1544, opened 2023-05-26
Update to the upcoming WITs

Mossaka:
Update the WITs to follow the upcoming package format. This will not be able to land until https://github.com/bytecodealliance/wit-bindgen/pull/580 and https://github.com/bytecodealliance/wasmtime/pull/6390 land and possible a new version for wit-bindgen and wasmtime. 

I used the package name `spin:reactor` for the entire WIT package. Happy to change it

PR #1594, opened 2023-06-16
Upgrade wasmtime 10.0

rylev:
This updates Spin to wamstime 10.0. 

I'm marking this as a draft for now as we'll want to wait until wasmtime 10.0 is officially released before making a decision on whether to merge this. 

This also supersedes #1544 (cc @Mossaka - thanks for making those changes!)


crates/core/src/store.rs
```diff
@@ -366,14 +389,14 @@ impl StoreBuilder {
...
-    fn with_wasi(&mut self, f: impl FnOnce(&mut Wasi)) {
+    fn with_wasi(&mut self, f: impl FnOnce(&mut WasiCtxBuilder)) {
         let _ = self.try_with_wasi(|wasi| {
             f(wasi);
             Ok(())
         });
     }
 
-    fn try_with_wasi(&mut self, f: impl FnOnce(&mut Wasi) -> Result<()>) -> Result<()> {
+    fn try_with_wasi(&mut self, f: impl FnOnce(&mut WasiCtxBuilder) -> Result<()>) -> Result<()> {
```
lann, line 399: Might be nice to update these to take two closures e.g.

```suggestion
    fn try_with_wasi(
      &mut self,
      preview1: impl FnOnce(&mut WasiCtxBuilder) -> Result<()>,
      preview2: impl FnOnce(WasiCtxBuilder) -> Result<WasiCtxBuilder>)
    ) -> Result<()> {
```

Which would let you avoid the `take` juggling in all the methods above. :shrug:

rylev: I tried this out because it seemed like a good idea, but it turns out to require even more code due to the fact that some methods can be implemented using a closure based approach (due to their arguments not being moveable into two separate closures). Here's what the code looks like: https://github.com/fermyon/spin/compare/upgrade-wasmtime-10-0...try-suggestion

rylev:
Moving this to ready review now that wasmtime 10.0.0 has been released.

Before actually merging this we'll want to merge:
* https://github.com/fermyon/spin-componentize/pull/10
* https://github.com/fermyon/wit-bindgen-backport/pull/11

And then we should update the git shas referring to these dependencies to be the newly merged commit.

dicej (approved):

PR #1595, opened 2023-06-16
Warn when directly using a WebAssembly component

### LINKED ISSUE #1571: Don't quietly accept a prebuilt component as a source
We lazily wrap wasm modules into components with `spin-componentize` because components aren't stable yet and we don't want to provide compatibility for every revision of the component spec.

We currently quietly accept prebuilt components and attempt to run them as if they were produced by the linked version of `spin-componentize`, which will cause older prebuilt components to fail eventually.

At a minimum we should print a warning if a prebuilt component is used. We may even want to disallow them entirely or gate loading them behind some kind of flag.

rylev:
Fixes #1571

This emits a warning when loading a WebAssembly component directly instead of a core module.

> Warning: spin component at path /Users/rylev/Code/fermyon/min-spin/target/wasm32-unknown-unknown/release/min.wasm is already a WebAssembly component. Directly using WebAssembly components is an experimental feature.

This warning is a bit hard to see. We may want to highlight the warning in bold yellow much like we do in red for errors and other important things in green.

### DIFF
--- a/crates/trigger/src/loader.rs
+++ b/crates/trigger/src/loader.rs
@@ -49,18 +49,23 @@ impl Loader for TriggerLoader {
             .as_ref()
             .context("LockedComponentSource missing source field")?;
         let path = parse_file_url(source)?;
-        spin_core::Component::new(
-            engine,
-            spin_componentize::componentize_if_necessary(&fs::read(&path).await.with_context(
-                || {
-                    format!(
-                        "failed to read component source from disk at path '{}'",
-                        path.display()
-                    )
-                },
-            )?)?,
-        )
-        .with_context(|| format!("loading module {path:?}"))
+        let bytes = fs::read(&path).await.with_context(|| {
+            format!(
+                "failed to read component source from disk at path '{}'",
+                path.display()
+            )
+        })?;
+        let component = spin_componentize::componentize_if_necessary(&bytes)?;
+        let was_already_component = matches!(component, std::borrow::Cow::Borrowed(_));
+        if was_already_component {
+            println!(
+                "Warning: spin component at path {} is already a WebAssembly component. \
+                Directly using WebAssembly components is an experimental feature.",
+                path.display()
+            )
+        }
+        spin_core::Component::new(engine, component)
+            .with_context(|| format!("loading module {path:?}"))
     }
 
     async fn load_module(
--- a/crates/trigger/src/loader.rs
+++ b/crates/trigger/src/loader.rs
@@ -59,8 +59,8 @@ impl Loader for TriggerLoader {
         let was_already_component = matches!(component, std::borrow::Cow::Borrowed(_));
         if was_already_component {
             println!(
-                "Warning: spin component at path {} is already a WebAssembly component. \
-                Directly using WebAssembly components is an experimental feature.",
+                "Warning: Spin component at path {} is a WebAssembly component instead of a \
+                WebAssembly module. Use of the WebAssembly component model is an experimental feature.",
                 path.display()
             )
         }


lann:
> We may want to highlight the warning in bold yellow much like we do in red for errors and other important things in green.

👍

lann (approved):

radu-matei (approved):
LGTM

Extending the color output to warnings sounds good to me.

crates/trigger/src/loader.rs
```diff
@@ -49,18 +49,23 @@ impl Loader for TriggerLoader {
...
-        )
-        .with_context(|| format!("loading module {path:?}"))
+        let bytes = fs::read(&path).await.with_context(|| {
+            format!(
+                "failed to read component source from disk at path '{}'",
+                path.display()
+            )
+        })?;
+        let component = spin_componentize::componentize_if_necessary(&bytes)?;
+        let was_already_component = matches!(component, std::borrow::Cow::Borrowed(_));
```
itowlson, line 59: This feels like a very fragile dependency on an implementation detail of `componentize_if_necessary`.  Would it be prohibitive to modify `spin-componentize` to return "was it originally a component or module" as well as the componentized bytes?

crates/trigger/src/loader.rs
```diff
@@ -49,18 +49,23 @@ impl Loader for TriggerLoader {
...
+        if was_already_component {
+            println!(
+                "Warning: spin component at path {} is already a WebAssembly component. \
```
itowlson, line 62:
```suggestion
                "Warning: Spin component at path {} is already a WebAssembly component. \
```

Consider also dropping `already` - it makes sense only if the user is aware that Spin wraps core modules with components (but maybe we reckon that anybody playing with CM components will know that).

itowlson: Given the heavy overloading of "component," there is a risk that readers will pick out "what is problematic" as "WebAssembly" rather than "component" (because "component" has become such a blah word that it can mean pretty much "any unit of code").  "Hmm, so it doesn't like that it's already WebAssembly?  Maybe it wants me to pass in source code...?"

Maybe we are confident that only users who are already using Spin and are consciously playing with the CM (and so realise the significance of "WebAssembly component") will ever hit this.  But if there is a way to phrase this so that it's clearly a technical term of art, a la "Wasm Component Model component" except less awkward, then that would be helpful.  "...is built to the Wasm Component Model specification" or something like that maybe?

crates/trigger/src/loader.rs
```diff
@@ -49,18 +49,23 @@ impl Loader for TriggerLoader {
...
+                Directly using WebAssembly components is an experimental feature.",
```
itowlson, line 63: Maybe "Component Model components"?

crates/trigger/src/loader.rs
rylev: I'm sort of disagree that it's fragile as it is directly tied to the fact that the bytes that go into `componentize_if_necessary` are the same that came out which is exactly what we want to test for. We could make a bespoke enum that frames this in terms of components and modules, but that enum would essentially just be a reimplementation of `Cow`. However, I'll still try to take a look into making this clearer.

crates/trigger/src/loader.rs
rylev: I've changed the warning to

> Warning: Spin component at path {} is a WebAssembly component instead of a WebAssembly module. Use of the WebAssembly component model is an experimental feature.

itowlson: Perfect - the clear contrast of component vs module hits the spot.  Thank you!

PR #1597, opened 2023-06-16
Fix brokwn like to Key Value Store

giodamelio:
Just a small fix for a broken link in your README


radu-matei (approved):
Thank you for your contribution, @giodamelio!

LGTM

PR #1600, opened 2023-06-20
(fix/oci-client): allow adding a token for OCI client

radu-matei:
This commit updates the underlying OCI client from the OCI crate used by Spin to be public. This is so that consumers of the client can configure the client (in particular, the token cache).

### DIFF
--- a/crates/oci/Cargo.toml
+++ b/crates/oci/Cargo.toml
@@ -11,7 +11,7 @@ dkregistry = { git = "https://github.com/camallo/dkregistry-rs", rev = "37acecb4
 docker_credential = "1.0"
 dirs = "4.0"
 futures-util = "0.3"
-oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "64986855ef0d692df3b270d23c4bee8c41d97c27" }
+oci-distribution = { git = "https://github.com/fermyon/oci-distribution", branch = "access-token-cache" }
 reqwest = "0.11"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
@@ -23,3 +23,4 @@ tempfile = "3.3"
 tokio = { version = "1", features = ["fs"] }
 tracing = { workspace = true }
 walkdir = "2.3"
+
--- a/crates/oci/Cargo.toml
+++ b/crates/oci/Cargo.toml
@@ -11,7 +11,7 @@ dkregistry = { git = "https://github.com/camallo/dkregistry-rs", rev = "37acecb4
 docker_credential = "1.0"
 dirs = "4.0"
 futures-util = "0.3"
-oci-distribution = { git = "https://github.com/fermyon/oci-distribution", branch = "access-token-cache" }
+oci-distribution = { git = "https://github.com/fermyon/oci-distribution", rev = "05022618d78feef9b99f20b5da8fd6def6bb80d2" }
 reqwest = "0.11"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -4,6 +4,8 @@ use anyhow::{bail, Context, Result};
 use docker_credential::DockerCredential;
 use futures_util::future;
 use futures_util::stream::{self, StreamExt, TryStreamExt};
+use oci_distribution::token_cache::RegistryTokenType;
+use oci_distribution::RegistryOperation;
 use oci_distribution::{
     client::{Config, ImageLayer},
     manifest::OciImageManifest,
@@ -37,6 +39,7 @@ const CONTENT_REF_INLINE_MAX_SIZE: usize = 128;
 pub struct Client {
     /// Global cache for the metadata, Wasm modules, and static assets pulled from OCI registries.
     pub cache: Cache,
+    /// Underlying OCI client.
     oci: oci_distribution::Client,
 }
 
@@ -324,6 +327,16 @@ impl Client {
         let mut auth = AuthConfig::load_default().await?;
         auth.insert(server, username, password)?;
         auth.save_default().await
+    }
+
+    /// Insert a token in the OCI client token cache.
+    pub fn insert_token(
+        &mut self,
+        reference: &Reference,
+        op: RegistryOperation,
+        token: RegistryTokenType,
+    ) {
+        self.oci.tokens.insert(reference, op, token);
     }
 
     /// Validate the credentials by attempting to send an authenticated request to the registry.
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -30,7 +30,10 @@ use spin_http::{
     config::{HttpExecutorType, HttpTriggerConfig},
     routes::{RoutePattern, Router},
 };
-use spin_trigger::{locked::DESCRIPTION_KEY, EitherInstancePre, TriggerAppEngine, TriggerExecutor};
+use spin_trigger::{
+    locked::{BINDLE_VERSION_KEY, DESCRIPTION_KEY, OCI_IMAGE_DIGEST_KEY, VERSION_KEY},
+    EitherInstancePre, TriggerAppEngine, TriggerExecutor,
+};
 use tls_listener::TlsListener;
 use tokio::net::{TcpListener, TcpStream};
 use tokio_rustls::server::TlsStream;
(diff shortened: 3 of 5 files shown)


crates/oci/src/client.rs
```diff
@@ -34,7 +34,8 @@ const MAX_PARALLEL_PULL: usize = 16;
 pub struct Client {
     /// Global cache for the metadata, Wasm modules, and static assets pulled from OCI registries.
     pub cache: Cache,
-    oci: oci_distribution::Client,
+    /// Underlying OCI client.
+    pub oci: oci_distribution::Client,
```
rylev, line 38: I don't have full context here, but would it make sense to keep this private and expose the relevant customization function through methods on the wrapper `Client` struct instead?

crates/trigger/src/locked.rs
```diff
@@ -23,6 +23,7 @@ pub const NAME_KEY: MetadataKey = MetadataKey::new("name");
 pub const VERSION_KEY: MetadataKey = MetadataKey::new("version");
 pub const DESCRIPTION_KEY: MetadataKey = MetadataKey::new("description");
 pub const BINDLE_VERSION_KEY: MetadataKey = MetadataKey::new("bindle_version");
+pub const OCI_IMAGE_DIGEST_KEY: MetadataKey = MetadataKey::new("oci_image_digest");
```
itowlson, line 26: I can't see where this is being set... seems like it would need an update to `ApplicationOrigin` and `LockedAppBuilder`?

itowlson (dismissed):
This looks like a separate (and not yet complete change) has gotten included in here maybe?

crates/oci/src/client.rs
radu-matei: Done.

crates/trigger-http/src/lib.rs
```diff
@@ -373,6 +374,8 @@ pub struct AppInfo {
     pub version: Option<String>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub bindle_version: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub oci_image_digest: Option<String>,
```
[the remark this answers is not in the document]
radu-matei: Updated the commit message to include details about its usage.

itowlson:
@radu-matei the outstanding issue for me on this is https://github.com/fermyon/spin/pull/1600/files#r1237937342 (I mentioned it in chat a little while ago but should have noted it here as well)

vdice:
Eager to get this in for more convenient OCI testing.  @radu-matei when you have a chance could you rebase and take a look at https://github.com/fermyon/spin/pull/1600#issuecomment-1613824925?

crates/trigger/src/locked.rs
vdice: Indeed, it seems like we're missing updating the locked app metadata eg in this bit of code: https://github.com/fermyon/spin/blob/main/crates/trigger/src/locked.rs#L61-L69

Shall I add a commit doing so?  @radu-matei

vdice:
(rebased with main)

crates/trigger/src/locked.rs
vdice: Okay, I dug into the code around locked apps, app info, etc. and came up with a few items.

- It now appears to make sense that we didn't update `ApplicatonOrigin` and `LockedAppBuilder`, as the oci image digest isn't currently relevant to any of the [build_locked_app](https://github.com/search?q=repo%3Afermyon%2Fspin+build_locked_app&type=code) code paths, which appear to be either 1. building a locked app from the filesystem or 2. ditto, albeit in preparation for pushing to an oci registry (but then, we won't the digest until after the push; we do know the reference, as provided by the user, yet this feels odd to set as the 'origin'/`ApplicationOrigin` as it is more like the 'destination' in this case)
- In the case where we are running from an oci reference (eg pulling, as in `spin up -f vdice/foo:latest`), we are currently [setting the origin](https://github.com/fermyon/spin/blob/main/crates/oci/src/loader.rs#L45-L47) but we could/should also set the `OCI_IMAGE_DIGEST_KEY` as well.  I did so in this commit: https://github.com/vdice/spin/commit/e332b1f471171be1cd1717e8eac8bccaf4e78832
  - (This is where the const mentioned here could/would be used)
- Speaking of the origin key, we didn't appear to be exposing it on the app info endpoint (i.e. `.well-known/spin/info`) but we could certainly add it (in a follow-up), as I've sketched out in this commit: https://github.com/vdice/spin/commit/eaedad5e52fd00f36e625a042e1d4a894f10c724

I do have further questions regarding why we have duplicated consts for similar metadata both in the [spin app crate and the trigger crate](https://github.com/search?q=repo%3Afermyon%2Fspin%20BINDLE_VERSION_KEY&type=code) as well as how/if the [bindle loader code path](https://github.com/search?q=repo%3Afermyon%2Fspin%20from_bindle&type=code) is still being used (in the context of `ApplicationOrigin` here... the oci path no longer follows the similar code paradigm set in the `loader` crate), but perhaps those are Qs for a different time.

This might have muddied the waters further. To me it seems like we may want to add the [aforementioned commit](https://github.com/vdice/spin/commit/e332b1f471171be1cd1717e8eac8bccaf4e78832) adding the oci image digest to the locked app metadata where applicable (eg on pull) and relegate other enhancements/Qs to follow-ups.  What do we think?

vdice: @itowlson if convenient, could I get your thoughts on https://github.com/fermyon/spin/pull/1600#discussion_r1309417022 above?

crates/oci/Cargo.toml
```diff
@@ -11,7 +11,7 @@ dkregistry = { git = "https://github.com/camallo/dkregistry-rs", rev = "37acecb4
 docker_credential = "1.0"
 dirs = "4.0"
 futures-util = "0.3"
-oci-distribution = { git = "https://github.com/krustlet/oci-distribution", rev = "64986855ef0d692df3b270d23c4bee8c41d97c27" }
+oci-distribution = { git = "https://github.com/fermyon/oci-distribution", branch = "access-token-cache" }
```
lann, line 14: We should pin to specific revs generally, and I'd really like those to be on `main` (or at least a protected branch) so they aren't accidentally deleted.

crates/trigger-http/src/lib.rs
```diff
@@ -254,7 +257,12 @@ impl HttpTrigger {
 
     /// Returns spin status information.
     fn app_info(&self) -> Result<Response<Body>> {
-        let info = AppInfo::new(self.engine.app());
+        let info = AppInfo {
+            name: self.engine.app_name.clone(),
+            version: self.engine.app().get_metadata(VERSION_KEY)?,
+            bindle_version: self.engine.app().get_metadata(BINDLE_VERSION_KEY)?,
+            oci_image_digest: self.engine.app().get_metadata(OCI_IMAGE_DIGEST_KEY)?,
```
lann, line 264: This change can be reverted. `oci_image_digest` is already populated by `AppInfo::new`

crates/trigger/src/locked.rs
lann: This line can be removed.

crates/oci/Cargo.toml
vdice: The `access-token-cache` branch is now protected; updated to use a specific revision.

lann (approved):

crates/trigger/src/locked.rs
itowlson: @vdice Sorry for the slow reply and maybe it's now moot since this is merged.  My only concern was that nothing seemed to set this thing and it looks like it's been removed so all good.

Re exposing the origin key on the app info endpoint, not sure if I understand this right but it sounds like you are suggesting making the app package's OCI reference publicly visible?  If that's correct, I'd be rather concerned about doing that by default.  Yes, public registries are browsable anyway, and private registries are secured anyway, but giving away the storage location of an app sets my "how will bad guys figure out how to exploit this" spidey sense tingling.  (If I've misunderstood what you propose to add to `.well-known` then ignore this comment.)

vdice: Thanks @itowlson -- it turned out this line wasn't needed, so we proceeded to merge.

> Re exposing the origin key on the app info endpoint, not sure if I understand this right but it sounds like you are suggesting making the app package's OCI reference publicly visible? If that's correct, I'd be rather concerned about doing that by default. Yes, public registries are browsable anyway, and private registries are secured anyway, but giving away the storage location of an app sets my "how will bad guys figure out how to exploit this" spidey sense tingling. (If I've misunderstood what you propose to add to .well-known then ignore this comment.)

Good points.  The sketched-out change wasn't included in this PR.  If I dust off the commit with intentions to PR, I'll create an issue first where we can discuss.

PR #1603, opened 2023-06-21
fix(deps): remove `zlib` dependency

kate-goldenring:
The `flate2` crate allows for selecting which `zlib` backend it uses. Instead of requiring/using a system `zlib` library, `flate2`'s [README](https://github.com/rust-lang/flate2-rs#backends) recommends using the `miniz_oxide` backend which uses the `zlib-ng` C library instead of stock system `zlib` library.

Right now, Spin is using system `zlib` via `flate2` and MySQL crates. This changes all imports to use the embedded `zlib`.

Previously:
```sh
$ cargo tree -i flate2 -e features
...
├── flate2 feature "libz-sys"
│   └── flate2 feature "zlib" (*)
├── flate2 feature "miniz_oxide"
│   └── flate2 feature "rust_backend"
│       └── flate2 feature "default" (*)
├── flate2 feature "rust_backend" (*)
└── flate2 feature "zlib" (*)
```

Now, uses `zlib-ng`:

```sh
$ cargo tree -i flate2 -e features
...
├── flate2 feature "libz-ng-sys"
│   └── flate2 feature "zlib-ng" (*)
├── flate2 feature "miniz_oxide"
│   └── flate2 feature "rust_backend"
│       └── flate2 feature "default" (*)
├── flate2 feature "rust_backend" (*)
└── flate2 feature "zlib-ng" (*)
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,12 +100,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2d098ff73c1ca148721f37baad5ea6a465a13f9573aba8641fbbbae8164a54e"
 
 [[package]]
-name = "arrayvec"
-version = "0.7.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
-
-[[package]]
 name = "async-channel"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


fibonacci1729 (approved):

alexcrichton (approved):

rylev:
Looks like cmake needs to be installed in the e2e testing docker image since that's required to build zlib and thus to build Spin. Perhaps we should also note this requirement in the README?

kate-goldenring:
Added the `cmake` dependency and created a PR to add it to our contributing docs as that is where we currently list dependencies for building Spin: https://github.com/fermyon/developer/pull/727

itowlson:
@kate-goldenring under what circumstances will mortals need `cmake` to build Spin locally - e.g. will things like `cargo build` and `make test-unit` still work without it?

PR #1607, opened 2023-06-22
Make some sqlite types pub

rylev:
We were accidentally exploiting a hole in the Rust module system that allows you to return types from public functions that cannot be directly referred to. This fixes that issue.

### DIFF
--- a/sdk/rust/src/sqlite.rs
+++ b/sdk/rust/src/sqlite.rs
@@ -1,14 +1,19 @@
+#![allow(missing_docs)]
+
 wit_bindgen_rust::import!("../../wit/ephemeral/sqlite.wit");
 
 use sqlite::Connection as RawConnection;
 
 /// Errors which may be raised by the methods of `Store`
-pub type Error = sqlite::Error;
-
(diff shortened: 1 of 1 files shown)


sdk/rust/src/sqlite.rs
```diff
@@ -1,14 +1,19 @@
+#![allow(missing_docs)]
```
itowlson, line 1: I would expect us to want to enforce docs for the public SDK interface?

itowlson (changes requested):

sdk/rust/src/sqlite.rs
rylev: I agree though interestingly if you look at most of the other modules of the sdk they already allow missing docs. It seems we're not getting docs from the code generated by wit-bindgen. I need to look more into what specifically is happening though.

itowlson: Ah, okay, no worries then, we can assess that as s separate issue then.  :shipit:

itowlson (approved):

PR #1611, opened 2023-06-24
also delete release when deleting tag

rajatjindal:
this attempts to fix issue where sometimes our canary release is marked as 'draft' and re-runs won't help as it will just update the existing release (which is already draft).

I still don't have a good answer as to why the release is getting marked as draft (except someone reporting same issue here https://github.com/ncipollo/release-action/issues/336#issue-1728946267). may be if we push to main while previous release is being tagged?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -223,6 +223,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: canary
+          delete_release: true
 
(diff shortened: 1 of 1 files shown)


itowlson (approved):
Longer term it would be great to find a better solution, but if this gets us out of the "all CI fails and everything is terrible" hole we're in then :shipit: :shipit: :shipit: and we'll figure it out later!

rajatjindal:
this just happened again (before the merge), so I am going to merge this and monitor if it improves the state.

PR #1613, opened 2023-06-26
List installed plugins in `spin plugins list`

### LINKED ISSUE #1580: `spin plugin list` does not show all installed plugins
If I install a plugin from a local package and file, it does not show up in the list, when running `spin plugin list`, unless I use the `--installed` flag. I think it would also be good to have an option to show the source, similar to how this works with templates.

```
> spin plugin list
cloud 0.1.0
cloud 0.1.1
js2wasm 0.1.0
js2wasm 0.2.0
js2wasm 0.3.0
js2wasm 0.4.0 [installed]
py2wasm 0.1.0
py2wasm 0.1.1
py2wasm 0.2.0 [installed]

> spin plugin list --installed
cloud 1.3 [installed]
js2wasm 0.4.0 [installed]
py2wasm 0.2.0 [installed]
```

itowlson:
Fixes #1580.

I have mixed feelings about this but that's probably because I've got so much test garbage in my system.  I'd welcome feedback from someone with a more realistic environment!

```
$ spin plugins list
arm-only 0.1.0 [incompatible]
cloud 0.1.0
cloud 0.1.1
cloud 0.4 [installed]
js2wasm 0.1.0
js2wasm 0.2.0
js2wasm 0.3.0
js2wasm 0.4.0 [installed]
not-spin-ver 10.2.0
pluginify 0.4 [installed]
py2wasm 0.1.0
py2wasm 0.1.1
py2wasm 0.2.0 [installed]
timer 0.1.0 [installed]
trigger-armstuff 0.1.0 [incompatible]
trigger-composite 0.4 [installed]
trigger-finger 0.1 [installed]
trigger-madeup 0.1.0
trigger-message 0.0.7 [installed]
trigger-pg-event 0.3 [installed]
trigger-sqs 0.3 [installed]
trigger-tcpline 0.1 [installed]
trigger-timer 0.1.0 [installed]
```

Also, this is an incompatible change in a command that we have documented as stable.  Should we put it on a flag (e.g. `--all`) for now?

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -307,7 +307,7 @@ impl List {
         let mut plugins = if self.installed {
             Self::list_installed_plugins()
         } else {
-            Self::list_catalogue_plugins()
+            Self::list_catalogue_and_installed_plugins()
         }?;
 
         plugins.sort_by(|p, q| p.cmp(q));
@@ -321,12 +321,13 @@ impl List {
         let store = manager.store();
         let manifests = store.installed_manifests()?;
         let descriptors = manifests
-            .iter()
+            .into_iter()
             .map(|m| PluginDescriptor {
                 name: m.name(),
                 version: m.version().to_owned(),
                 installed: true,
-                compatibility: PluginCompatibility::for_current(m),
+                compatibility: PluginCompatibility::for_current(&m),
+                manifest: m,
             })
             .collect();
         Ok(descriptors)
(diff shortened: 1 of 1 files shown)


mikkelhegn:
The current help text from `spin plugin list` is `List available or installed plugins`. I would consider it a bug to not include plugins not in the catalog, but installed. That would make it consistent with the `--installed` flag, which is documented (in help) to `List only installed plugins`.

Completing #1614 to include the source, will reveal what's in the catalog and what's not.

mikkelhegn (approved):
LGTM

Issue #1608, opened 2023-06-23
Cargo feature to use only "interface" bits from HTTP crates


itowlson:
The `spin cloud deploy` command needs to understand the app info endpoint for checking update deployment readiness.  The schema for this is defined in the `spin-http-trigger` crate.  This currently brings in a whole lot of Spin runtime goodness from `spin-core` all the way up to the Postgres crate.  It would be good to offer a Cargo feature on `spin-http-trigger` to allow the plugin to import the app info type without the entire runtime.

fibonacci1729:
This is addressed by #1634

michelleN:
@itowlson @fibonacci1729 is this issue resolved?

fibonacci1729:
I would say so! I'll close. @itowlson feel free to reopen if you desire.

PR #1623, opened 2023-06-28
Validate variable and config key names at app load

### LINKED ISSUE #1622: Detect malformed configuration variable keys before up/deploy
The documentation for [dynamic configuration variables](https://github.com/fermyon/developer/blob/main/content/spin/dynamic-configuration.md) specifies that `Configuration keys may only contain lowercase letters and underscores between letters`.

Currently, if i deploy an application with a malformed configuration key (e.g. `LOG_LEVEL`) my application will serve `500` and my logs will contain angry shouting about what the proper format is for keys.

I think it would be ideal if we somehow validate configuration keys pre-deploy/up or ask the doctor to do it.

itowlson:
This fixes #1622 as far as we can go within Spin.  I think that the cloud plugin takes the manifest before validation (the local load pipeline doesn't come apart at the right place) so this would need to be ported to the cloud plugin separately.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5260,6 +5260,7 @@ dependencies = [
  "serde_json",
  "shellexpand 3.1.0",
  "spin-common",
+ "spin-config",
  "spin-manifest",
  "tempfile",
  "terminal",
--- a/crates/loader/Cargo.toml
+++ b/crates/loader/Cargo.toml
@@ -25,6 +25,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 shellexpand = "3.1"
 spin-common = { path = "../common" }
+spin-config = { path = "../config" }
 spin-manifest = { path = "../manifest" }
 tempfile = "3.3.0"
 terminal = { path = "../terminal" }
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -27,7 +27,10 @@ use spin_manifest::{
 };
 use tokio::{fs::File, io::AsyncReadExt};
 
-use crate::{cache::Cache, validation::validate_key_value_stores};
+use crate::{
+    cache::Cache,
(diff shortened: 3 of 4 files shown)


itowlson:
Looking deeper at this, the added dependency on `spin-config` drags `spin-app` and `spin-core` as indirect dependencies of `spin-loader`, which is a lot of weight to add to any plugin that just wants to read a dashed manifest.  So maybe it would be better to duplicate the validation criteria into `spin-loader`.  Or we could wham a Cargo feature into `spin-config`.  @lann do you have opinions?

fibonacci1729:
While discussing this earlier, @lann and I had envisioned a `check` crate that could be shared between `spin`, `lhc-prepare`, and the `cloud-plugin`. This feels more sippable as we'd likely want to gather consensus on what is in the realm of doctor and this validation step (or atleast provide guidance) so perhaps not something to try to address in this PR. 

This is my fault though I should have updated the issue with some more info re: sharing this validation logic with lhc so my apologies Ivan! 

I think it makes sense in the interim to implement something like what you are proposing here though because I hit this issue earlier on when building a demo app for config (as I forgot to consult the dynaconfig docs)

itowlson:
@fibonacci1729 Yeah, the cloud plugin at least currently depends on the runtime through other avenues (see https://github.com/fermyon/spin/issues/1608) so it's not a huge deal for that.  I'm happy to let this stand and re-assess in a future PR.

Thanks for sharing your discussion - I agree it would be good to figure out more broadly what might be useful to alternate tools and hosts and how to surface it.  E.g. a `check` crate could be useful but keeping validation logic with the validatees is also useful; what about things like the info endpoint schema; etc.  But the feature route risks being more complicated to manage.  So 🤷

itowlson:
Would somebody be willing to either approve or ask for changes please?  Thanks!

fibonacci1729 (approved):
LGTM! This is useful today even if we decide to factor things differently in the near future. Thanks @itowlson !

PR #1624, opened 2023-06-28
Update libsql-client dependency

rylev:
Update the libsql-client dependency to the latest version.

### DIFF
--- a/crates/sqlite-inproc/src/lib.rs
+++ b/crates/sqlite-inproc/src/lib.rs
@@ -3,6 +3,8 @@ use std::{
     sync::{Arc, Mutex},
 };
 
+use anyhow::Context;
+use async_trait::async_trait;
 use spin_sqlite::Connection;
 use spin_world::sqlite;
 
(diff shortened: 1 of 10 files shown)


crates/sqlite-libsql/src/lib.rs
```diff
@@ -1,16 +1,18 @@
...
-    client: libsql_client::reqwest::Client,
+    inner: libsql_client::http::Client,
 }
 
 impl LibsqlClient {
-    pub fn new(url: String, token: String) -> Self {
-        Self {
-            client: libsql_client::reqwest::Client::new(url, token),
-        }
+    pub fn create(url: &str, token: String) -> anyhow::Result<Self> {
```
itowlson, line 9: (for my own education, not a problem)  Why does this change from `new` to `create` - is it because returning a `Result` is unidiomatic for `new`?

crates/sqlite-libsql/src/lib.rs
```diff
@@ -59,7 +61,7 @@ impl spin_sqlite::Connection for LibsqlClient {
         // As in `query`, the shenanigans just wrap a call to libsql's `Client::batch()`.
         std::thread::spawn(move || {
             let rt = tokio::runtime::Runtime::new().unwrap();
-            rt.block_on(client.batch(stmts))
+            rt.block_on(client.raw_batch(stmts))
```
itowlson, line 64: Do we know of any implications to changing from `batch` to `raw_batch` here?  (Apart from `batch` would require us to _further_ wrap our `http::Client` in a `client::Client`.)  The docs are a bit ambiguous.

I don't know if it matters but the comment on line 61 still refers to `batch()`.

itowlson (approved):
LGTM, comments are mostly for my own learning

crates/sqlite-libsql/src/lib.rs
rylev: Yes exactly - I believe clippy has a (maybe off by default) warning for this.

crates/sqlite-libsql/src/lib.rs
rylev: Thanks for asking this. I took another look and the `raw_batch` doesn't collect errors - so essentially we would swallow errors if the user's migration file was not valid. On top of that, we would not be running everything in a transaction which is also not good. I've switched back to using `batch` by wrapping the http client in a generic client.

This also made me look at some of the other code around and clean it up some hacks from the original implementation. This includes making the `Connection` trait async. This improves over what is currently in main since we'll now coordinate more appropriately with the tokio runtime.

crates/sqlite-inproc/src/lib.rs
```diff
@@ -31,77 +33,99 @@ impl InProcConnection {
...
+    let rows = rows
+        .into_iter()
+        .map(|r| r.map_err(|e| sqlite::Error::Io(e.to_string())))
+        .collect::<Result<_, sqlite::Error>>()?;
+    Ok(sqlite::QueryResult { columns, rows })
+}
+
 fn convert_data(
-    arguments: impl Iterator<Item = spin_world::sqlite::Value>,
+    arguments: impl Iterator<Item = sqlite::Value>,
```
itowlson, line 105: The reason I used `spin_world` qualified naming was that I got really really confused about plain `sqlite` - at any give time, was it referring to it the SQLite library, the `spin-sqlite` crate, or the `sqlite` WIT interface?  I know it is actually unambiguous, but for me the unqualified form adds a lot of cognitive load.

I appreciate though that the qualified form creates a lot of visual clutter which is its own form of cognitive load...!

crates/sqlite-inproc/src/lib.rs
```diff
@@ -31,77 +33,99 @@ impl InProcConnection {
...
+            conn.execute_batch(&statements)
+                .context("failed to execute batch statements")
+        })
+        .await?
+        .context("failed to spawn blocking task")?;
         Ok(())
     }
 }
 
+fn make_query(
```
itowlson, line 66: "make" reads ambiguously in this context - you intend it in the sense of "the client makes a call to the database" but I first read it was "create a query object".  May I suggest something like `execute_query` or `perform_query`?

crates/sqlite-libsql/src/lib.rs
```diff
@@ -1,68 +1,57 @@
...
-        // error case). (It is tricky to make a function to name it, though, because
-        // of Send constraints.)
-        let result = std::thread::spawn(move || {
-            let rt = tokio::runtime::Runtime::new().unwrap();
-            rt.block_on(client.execute(stmt))
-        })
-        .join()
-        .unwrap_or_else(|_| Err(anyhow::anyhow!("internal thread error")))
-        .map_err(|e| sqlite::Error::Io(e.to_string()))?;
+        let result = client
```
itowlson, line 30: This is so much nicer!

itowlson (approved):
A couple of minor notes / suggestions - up to you if you want to pick them up.

I did want to confirm that we are re-testing `spin up --sqlite` against libsql.  The e2e tests check the in-proc case but not the libsql case, and that, unfortunately, is where tokio threw a fit last time - something was different between setup time and application run time.  I have no reason to expect a regression, and the changes look like they would be less likely to be prone to tokio state errors - I am just rather cautious about this area having been bitten before!

crates/sqlite-inproc/src/lib.rs
rylev: Renamed to `execute_query`

rylev:
I have tested against a libsql based database and it works as expected. 👍

PR #1626, opened 2023-06-29
Don't deploy old docs site on every merge to main

itowlson:
We can still manually deploy it if it ever changes

### DIFF
--- a/.github/workflows/deploy-website.yml
+++ b/.github/workflows/deploy-website.yml
@@ -2,6 +2,8 @@ name: Deploy Spin Docs Website
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):
Should we make a note somewhere, something like "MANUALLY DEPLOY DOCS IF MAKING A CHANGE OR ELSE".

itowlson:
@fibonacci1729 @endocrimes learneded me about keeping the auto action but only triggering on relevant changes so I'm gonna do that instead

fibonacci1729 (approved):

PR #1632, opened 2023-07-05
check for missing template/sdk tags

rajatjindal:


.github/gh-checktags.sh
```diff
@@ -0,0 +1,42 @@
...
+set -e
+
+which gh &> /dev/null || {
+  echo "This script requires the Github CLI 'gh'"
+  exit 1
+}
+
+tags=$(gh release list --repo fermyon/spin --exclude-pre-releases --exclude-drafts | grep -v TITLE | awk '{print $1}')
+
+failed=0
```
rylev, line 16: nit: can we just call this `exit` or `exit_code`?

.github/gh-checktags.sh
```diff
@@ -0,0 +1,42 @@
...
+  fi
+
+  # check golang sdk tag
+  if [[ -z "$(git tag -l sdk/go/$tag)" ]]; then
+    echo "tag sdk/go/$tag does not exist"
+    failed=1
+  fi
+
+  # remove trailing .\d+ from the version
+  tag_1=`echo $tag | sed 's/\.[0-9]*$//g'`
```
rylev, line 33: Nit:

```suggestion
  major_minor=`echo $tag | sed 's/\.[0-9]*$//g'`
```

rylev (commented):
Looks good. I'm not sure a cron based approach is the ideal approach, but I can't think of a better strategy.

.github/gh-checktags.sh
rajatjindal: done 👍

.github/gh-checktags.sh
```diff
@@ -0,0 +1,42 @@
...
+
+tags=$(gh release list --repo fermyon/spin --exclude-pre-releases --exclude-drafts | grep -v TITLE | awk '{print $1}')
+
+failed=0
+for tag in $tags; do
+  if [[ $tag != v1* ]]; then
+    continue
+  fi
+
+  if [[ $tag == *rc* ]]; then
```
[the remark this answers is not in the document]
rajatjindal: done 👍

.github/gh-checktags.sh
rajatjindal: done 👍

rylev (approved):

rajatjindal:
the tests failures seems unrelated to this PR. so we can merge it, if it looks ok otherwise.

itowlson (approved):
As written, this will need updating when we move to v2.  But who know whether we will reassess how we do versioning at that point anyway, so it's probably safest to keep this v1-only for now.  Thanks!

itowlson:
@rylev Re the cron approach: I get where you're coming from.  The workflow also has a `workflow_dispatch` trigger, so we could add "run the sanity check action" to the end of the release process.  And the cron at least gives us the "ambulance at the bottom of the cliff" if something goes amiss.

@rajatjindal Yes, the test failures are because of the known template/SDK mismatch issue.

PR #1633, opened 2023-07-05
Include actual and expected values in config test error message

rajatjindal:
this allows to debug config vars test by sending expected and actual value back

### DIFF
--- a/tests/testcases/config-variables/src/lib.rs
+++ b/tests/testcases/config-variables/src/lib.rs
@@ -12,14 +12,16 @@ fn handle_request(req: Request) -> Result<Response> {
         .query()
         .expect("Should have a password query string");
     let query: std::collections::HashMap<String, String> = serde_qs::from_str(query)?;
-    let provided_password = query
+    let expected_password_value = query
         .get("password")
         .expect("Should have a password query string");
-    let expected_password = config::get("password")?;
+    let actual_password_value = config::get("password")?;
 
     ensure!(
-        provided_password == &expected_password,
-        "password must match expected"
+        expected_password_value == &actual_password_value,
+        "actual password value from config store '{}' must match expected password value '{}'",
+        &actual_password_value,
+        expected_password_value
     );
 
     Ok(http::Response::builder().status(200).body(None)?)


tests/testcases/config-variables/src/lib.rs
```diff
@@ -12,15 +12,21 @@ fn handle_request(req: Request) -> Result<Response> {
...
-        "password must match expected"
-    );
+    if expected_password != &actual_password {
+        let error_msg = format!(
+            "expected password: '{}', actual password: '{}'",
+            expected_password, &actual_password
+        );
+
+        return Ok(http::Response::builder()
+            .status(500)
```
kate-goldenring, line 27: should this be a `401` instead of `500`? Does `ensure!` always throw a `500`?

rajatjindal: the return code does not really matter as long as it is not `200`. I think ensure panics and that triggers a 500 error code

kate-goldenring: That makes sense that a panic is 500. Thanks. I still would think 401 makes more sense here since it is a known potential error. i understand that it does not matter in the sense that the test will check the error code regardless

rajatjindal: sure, updated it to 401 :)

kate-goldenring (approved):

tests/testcases/config-variables/src/lib.rs
```diff
@@ -12,15 +12,21 @@ fn handle_request(req: Request) -> Result<Response> {
         .query()
         .expect("Should have a password query string");
     let query: std::collections::HashMap<String, String> = serde_qs::from_str(query)?;
-    let provided_password = query
+    let expected_password = query
```
itowlson, line 15: I initially found this rename very very confusing because in terms of _password_ logic, this is not "expected", this is indeed the user's attempt.  But in terms of what we are testing, I agree this is the "value we expect to get from config."

So I'd suggest decoupling the names from the "password" fiction, and call this something like "expected_config_value".

tests/testcases/config-variables/src/lib.rs
```diff
@@ -12,15 +12,21 @@ fn handle_request(req: Request) -> Result<Response> {
...
         .get("password")
         .expect("Should have a password query string");
-    let expected_password = config::get("password")?;
+    let actual_password = config::get("password")?;
```
itowlson, line 18: Again I found this renaming confusing in terms of the fiction.  Maybe `actual_config_value`?

tests/testcases/config-variables/src/lib.rs
```diff
@@ -12,15 +12,21 @@ fn handle_request(req: Request) -> Result<Response> {
...
 
-    ensure!(
-        provided_password == &expected_password,
-        "password must match expected"
-    );
+    if expected_password != &actual_password {
+        let error_msg = format!(
+            "expected password: '{}', actual password: '{}'",
```
itowlson, line 22: As this is a test case, I suggest phrasing this in terms of the "what went wrong" e.g. "value we expected to find in configuration / value actually found in configuration".  (This also precludes problems of people imitating out of context: "well Fermyon said I should return the correct password in auth failure messages, for debugging you know.")

tests/testcases/config-variables/src/lib.rs
```diff
@@ -12,15 +12,21 @@ fn handle_request(req: Request) -> Result<Response> {
...
+            expected_password, &actual_password
+        );
+
+        return Ok(http::Response::builder()
+            .status(401)
```
itowlson, line 27: I'd leave this as a 500 - it's "something is wrong in our test, something that shouldn't have happened," not an actual auth failure.  (Not sure if this means we could go back to using `ensure!` though.)

itowlson (changes requested):
The behaviour is fine but the naming is confusing in the context of the password fiction.  As this is specifically a test case, not something that is meant to motivate understanding, I'd suggest dropping the fiction, and being absolutely clear in the naming and the error message about it being a _config value that was not what it should have been._

itowlson:
@rajatjindal Also could you please amend the commit message?  It makes sense close up, but to someone scanning the history it won't be clear what it refers to.

tests/testcases/config-variables/src/lib.rs
itowlson: Oops, I see you and @kate-goldenring have already been round and round on this one... sorry for sticking another opinion in!  I agree with the previous discussion that it doesn't much matter!

tests/testcases/config-variables/src/lib.rs
rajatjindal: done.

tests/testcases/config-variables/src/lib.rs
rajatjindal: done

tests/testcases/config-variables/src/lib.rs
rajatjindal: we can't use `ensure!` as it panics and returns `500` without useful message. though to remove the fiction part, I would change it to return `500` instead of `401`

how I have done in some other tests is to let the `lib.rs` just return the value and do assertion in test case assertions itself. 

@kate-goldenring @itowlson - what do you think about that?

itowlson: @rajatjindal That's fine.  I don't care vastly about the code.  Happy with what you and @kate-goldenring have decided.

tests/testcases/config-variables/src/lib.rs
rajatjindal: done

tests/testcases/config-variables/src/lib.rs
rajatjindal: I've changed it back to 500. just because 401 generally means unauthorized and in the context of the test, the user may get confused with the error code. 

if we don't agree to 500, a better code would be `400` (bad request).

rajatjindal:
> @rajatjindal Also could you please amend the commit message? It makes sense close up, but to someone scanning the history it won't be clear what it refers to.

done.

tests/testcases/config-variables/src/lib.rs
```diff
@@ -10,17 +10,23 @@ fn handle_request(req: Request) -> Result<Response> {
...
         .query()
-        .expect("Should have a password query string");
+        .expect("Should have a expected_config_value query string");
     let query: std::collections::HashMap<String, String> = serde_qs::from_str(query)?;
-    let provided_password = query
-        .get("password")
-        .expect("Should have a password query string");
-    let expected_password = config::get("password")?;
+    let expected_config_value = query
+        .get("expected_config_value")
```
itowlson, line 16: If you change this, I think you need to change the test that passes the query string?  It might be easier to leave the actual query string as was, as you did for the config key.  It's only the variables and the message that I found confusing - sorry if I was overemphatic in my suggestions for change.

itowlson (approved):

rajatjindal:
ok, sorry for the back and forth on the PR. current state:

- ensure returns the error msg, but it was not clear. so I've updated it to include expected/actual values
- changed the variable names to reflect expected/actual values correctly
- I've removed other unnecessary changes.

PR #1634, opened 2023-07-06
Move AppInfo into spin-http

fibonacci1729:


crates/http/src/app_info.rs
```diff
@@ -0,0 +1,36 @@
+use serde::{Deserialize, Serialize};
+use spin_app::{App, MetadataKey};
+
+pub const APP_NAME_KEY: MetadataKey<String> = MetadataKey::new("name");
+pub const APP_VERSION_KEY: MetadataKey = MetadataKey::new("version");
+pub const BINDLE_VERSION_KEY: MetadataKey = MetadataKey::new("bindle_version");
+pub const OCI_IMAGE_DIGEST_KEY: MetadataKey = MetadataKey::new("oci_image_digest");
```
lann, line 7: I think these probably just belong in `spin-app`

lann (approved):

rylev:
I missing some context here - why should this live in spin-http and not spin-app?

lann:
> I missing some context here - why should this live in spin-http and not spin-app?

This is the serde definition of the JSON served at `/.well-known/spin/info` for both Spin and Cloud.

PR #1636, opened 2023-07-07
Check libsql url is in the right shape

### LINKED ISSUE #1635: Handle `libsql://` scheme in SQLite runtime config
Some libSQL documentation shows connections being set up using the `libsql` URL scheme.  For example, someone reading the Turso docs could easily come across their Spin tutorial and think, "aha, that's how I should set up my runtime config."

image

However, the HTTP client that Spin uses rejects this prefix (or did when I used it, though we have revved since), and expects only a host name.  If a user provides a `libsql://` URL in their runtime config, Spin will error on it.

The draft docs do warn people to specify only the host name.  However, we should consider being more forgiving in Spin, and automatically stripping the `libsql://` prefix if present.

A potential risk is if we later do have a reason to support the `libsql://` scheme and need a way to disambiguate the protocols.

rylev:
Fixes #1635 

I've decided to not transparently turn `libsql://` urls into `https://` urls because `libsql://` urls are a thing for talking to libsql databases over a websocket API. This just improves the error message so users know we only support http or https.

### DIFF
--- a/crates/trigger/src/runtime_config/sqlite.rs
+++ b/crates/trigger/src/runtime_config/sqlite.rs
@@ -132,9 +132,26 @@ pub struct LibsqlOpts {
 
 impl LibsqlOpts {
     fn build(&self) -> anyhow::Result<Arc<dyn Connection>> {
-        let client = spin_sqlite_libsql::LibsqlClient::create(&self.url, self.token.clone())
+        let url = &check_url(&self.url).with_context(|| {
+            format!(
+                "unexpected libsql url '{}' in runtime config file ",
+                self.url
+            )
+        })?;
+        let client = spin_sqlite_libsql::LibsqlClient::create(url, self.token.clone())
             .context("failed to create SQLite client")?;
         Ok(Arc::new(client))
+    }
+}
+
+// Checks an incoming url is in the shape we expect
+fn check_url(url: &str) -> anyhow::Result<&str> {
+    if url.starts_with("https://") || url.starts_with("http://") {
+        Ok(url)
+    } else {
+        Err(anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


crates/trigger/src/runtime_config/sqlite.rs
```diff
@@ -132,12 +132,29 @@ pub struct LibsqlOpts {
...
     }
 }
 
+// Checks an incoming url is in the shape we expect
+fn check_url(url: &str) -> anyhow::Result<&str> {
+    if url.starts_with("https://") || url.starts_with("http://") {
+        Ok(url)
+    } else {
+        Err(anyhow::anyhow!(
+            "url does not start with 'https://' or 'http://'. Spin currently only supports talking to libsql databases over http(s)"
```
itowlson, line 153:
```suggestion
            "URL does not start with 'https://' or 'http://'. Spin currently only supports talking to libSQL databases over HTTP(S)"
```

crates/trigger/src/runtime_config/sqlite.rs
```diff
@@ -132,12 +132,29 @@ pub struct LibsqlOpts {
 
 impl LibsqlOpts {
     fn build(&self) -> anyhow::Result<Arc<dyn Connection>> {
-        let client = spin_sqlite_libsql::LibsqlClient::create(&self.url, self.token.clone())
+        let url = &check_url(&self.url).with_context(|| {
+            format!(
+                "unexpected libsql url '{}' in runtime config file ",
```
itowlson, line 137:
```suggestion
                "unexpected libSQL URL '{}' in runtime config file ",
```

itowlson (approved):
I agree with the explanatory error rather than silent substitution.  A couple of fixes to the presentation of the messages, but otherwise good to go - thanks!

PR #1639, opened 2023-07-10
spin doctor component config diagnostic

itowlson:
```
$ spin doctor
📟 The Spin Doctor is in.
🩺 Checking spin.toml...

❗ Diagnosis: config entry 'ouch' in component 'vartest2' has invalid template: unmatched '{{' in template

❗ Diagnosis: config entry 'spork' in component 'vartest2' has invalid template: unknown variable "honk2"

❗ Diagnosis: config key 'plo____nk' in component 'vartest2' is invalid: must not contain multiple consecutive underscores
```

(This does not currently diagnose application variable validity, and it probably should.)

(Also not sure whether this is useful enough to be worth including, given that you get reasonable errors for these at `spin up` time... thoughts?)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5165,6 +5165,8 @@ dependencies = [
  "reqwest",
  "serde",
  "similar",
+ "spin-app",
+ "spin-config",
  "spin-loader",
  "tempfile",
  "terminal",
--- a/crates/config/src/host_component.rs
+++ b/crates/config/src/host_component.rs
@@ -92,7 +92,7 @@ impl config::Host for ComponentConfig {
 impl From<Error> for config::Error {
     fn from(err: Error) -> Self {
         match err {
-            Error::InvalidKey(msg) => Self::InvalidKey(msg),
+            Error::InvalidKey(key, reason) => Self::InvalidKey(format!("{key:?}: {reason}")),
             Error::InvalidSchema(msg) => Self::InvalidSchema(msg),
             Error::Provider(msg) => Self::Provider(msg.to_string()),
             other => Self::Other(format!("{}", other)),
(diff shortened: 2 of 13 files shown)


kate-goldenring:
I think it is worth it because it helps avoid runtime errors by diagnosing the issue before the spin app is used. I agree that it would be good in the future to also assess variable availability but at the same time, i dont think of it as an issue if the provider is not configured properly by `spin doctor` time but is by `spin up` time (i.e. an environment variable is not set yet)

itowlson:
@kate-goldenring ah, to be clear, I wasn't thinking about variable availability (I don't think that's feasible for spin doctor), I just meant the variable syntax (valid names) - this only checks component config.  Sorry for the confusion.

lann:
Given that this doesn't (can't) offer to fix the problem, I would tend to leave it to manifest validation. That said, it would be great if `spin doctor` could run the "normal" manifest validation as a first diagnosis.

itowlson:
@lann My vague thought was that it was surprising if a user ran `spin doctor` and it said "no problems found" and so they ran the app and it went "wait we have a big problem right here"  I agree that running manifest validations would be a simpler way to manage this, but means we need to reorganise the validations to report _all_ problems rather than stop on the first one.  Which, to be clear, would be a good thing to do anyway!

(Another consideration with that is that `spin doctor` tries to keep reporting things even in the face of malformed manifests, whereas the loader validations rely on the manifest being schema-compliant.  But that's less of an issue: hopefully the early fixes make the manifest well-formed enough to load.)

lann:
I was thinking of something like this: https://github.com/fermyon/spin/pull/1646

Combined perhaps with better validation error messages it seems like we could get the same effect without as much special doctor code.

itowlson:
Yeah and more complete validation (not stopping at the first problem).  Very happy to take that route.

Issue #1650, opened 2023-07-12
Redis handlers broken in Rust SDK 1.4


itowlson:
With Spin 1.4 templates:

```
ivan@hecate:~/testing$ spin new -a redis-rust reddysteddygo
ivan@hecate:~/testing$ cd reddysteddygo/
ivan@hecate:~/testing/reddysteddygo$ spin build
# ... succeeds ...
ivan@hecate:~/testing/reddysteddygo$ spin up
Logging component stdio to ".spin/logs/"
Error: Failed to instantiate component 'reddysteddygo'

Caused by:
   0: failed to validate exported interface `fermyon:spin/inbound-http`
   1: module does not export required function `fermyon:spin/inbound-http#handle-request`
Error: exit status: 1
```

If I revert the Spin SDK reference in Cargo.toml to 1.3 (and reinstate the wit-bindgen reference) then it works in the Spin 1.4 runtime, so I think this is specifically a SDK issue.

PR #1657, opened 2023-07-12
Make a test that exercises the canary Rust SDK's redis support

rylev:
This would have caught the issue in https://github.com/fermyon/spin/issues/1650.

We still need to do some thinking about how to make our e2e tests easier to manage, but this should hopefully make the likelihood of repeating the mistake from #1650 much harder in the future.


build.rs
```diff
@@ -130,8 +130,9 @@ fn run<S: Into<String> + AsRef<std::ffi::OsStr>>(
     cmd.stdout(process::Stdio::piped());
     cmd.stderr(process::Stdio::piped());
 
-    if let Some(dir) = dir {
-        cmd.current_dir(dir.into());
+    let dir = dir.map(Into::into);
```
rylev, line 133: All of the changes in this file are not strictly necessary for the fix. I just did some refactoring as I was trying to debug the issue.

tests/testcases/simple-redis-rust/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+bytes = "1"
+# General-purpose crate with common HTTP types.
+http = "0.2"
+spin-sdk = { path = "../../../sdk/rust"}
+
+[workspace]
+
+# Metadata about this component.
+[package.metadata.component]
+name = "spinhelloworld"
```
fibonacci1729, line 22: Is this metadata needed since we aren't using `cargo component` yet?

tests/testcases/simple-redis-rust/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name    = "simple-redis"
```
itowlson, line 2: Maybe `head-redis` or `current-branch-redis`, to make the purpose clear?

tests/testcases/simple-redis-rust/Cargo.toml
itowlson, line 22: Use the app name?  We have a tendency to copy-paste this name everywhere and it causes woes for the Rust SDK (although this position is probably benign).

tests/testcases/simple-redis-rust/spin.toml
```diff
@@ -0,0 +1,14 @@
+spin_version = "1"
+authors = ["Fermyon Engineering <engineering@fermyon.com>"]
+description = "A simple application"
+name = "spin-hello-world"
```
itowlson, line 4: Name corresponding to directory and purpose?

tests/testcases/mod.rs
```diff
@@ -734,6 +734,38 @@ pub async fn simple_spin_rust_works(controller: &dyn Controller) {
     tc.run(controller).await.unwrap()
 }
 
+pub async fn simple_redis_rust_works(controller: &dyn Controller) {
+    async fn checks(
+        _: AppMetadata,
+        _: Option<Pin<Box<dyn AsyncBufRead>>>,
+        stderr_stream: Option<Pin<Box<dyn AsyncBufRead>>>,
+    ) -> Result<()> {
+        let stderr = utils::get_output_stream(stderr_stream, Duration::from_secs(5)).await?;
```
itowlson, line 743: Our other Redis Rust test uses a 20 second timeout.  We should maybe pick a good timeout and give it a const name.

tests/spinup_tests.rs
```diff
@@ -93,6 +93,11 @@ mod spinup_tests {
         testcases::simple_spin_rust_works(CONTROLLER).await
     }
 
+    #[tokio::test]
+    async fn simple_redis_rust_works() {
```
itowlson, line 97: Name for purpose e.g. refer to main or head or current branch

build.rs
```diff
@@ -167,3 +168,9 @@ fn get_os_process() -> String {
         String::from("bash")
     }
 }
+
+fn current_dir() -> String {
+    std::env::current_dir()
+        .map(|d| d.display().to_string())
+        .unwrap_or_else(|_| String::from("<CURRENT DIR>"))
```
itowlson, line 175: Given the usage, maybe `<COULDN'T GET CURRENT DIR>`?  To a reader of test failure messages, it might be a bit clearer?

tests/spinup_tests.rs
itowlson, line 97: Do we want one of these for HTTP too?  It's less likely to go unnoticed, but in terms of root cause it could just as easily have been the HTTP SDK that went awry.

itowlson (approved):
Comments are mostly naming suggestions.  Otherwise LGTM and thanks for taking care of this!

tests/testcases/mod.rs
rylev: We always use 5 second for getting the output stream. The 20 seconds is to wait for spin to become ready. I can try to make that clearer, but I don't want to go too beyond what this PR is for.

build.rs
rylev: I don't think so. The current directory will be used. If we say "<COULDN'T GET CURRENT DIR>" the user has no idea what directory. If the say "<CURRENT DIR>" maybe they'll understand that it means whatever the current directory is.

tests/spinup_tests.rs
rylev: We already have one for HTTP: `simple_spin_rust_test`. I'll change the name so it's more clear.

rylev:
@itowlson I've made some of the naming more clear. I think the entire test suite might need a similar clean up at some point.

tests/testcases/mod.rs
itowlson: Nope, no worries, apologies for missing the different contexts.

PR #1667, opened 2023-07-17
fix `spin-timer` example regression

### LINKED ISSUE #1662: Timer trigger sample fails to resolve import
As part of ongoing component and Wasmtime infrastructure improvements, the `examples/spin-timer` trigger sample was updated in #1615 to refer to the Spin `config` API in a new way.  Unfortunately, although both the trigger and example guest build okay, the example guest fails at `spin up` with:

```
$ spin up
Error: Failed to instantiate component 'three'

Caused by:
    0: Failed to instantiate component 'three'
    1: import `fermyon:example/config` has the wrong type
    2: instance export `get-config` has the wrong type
    3: expected func found nothing
```

I am not sure how to go about fixing this.  @rylev @dicej do you have any ideas please?  I'm not sure how to share the diff in a text format so here's ye olde GitHub screenshot:

image

The workaround is to revert the sample to an earlier commit (Spin itself doesn't need to be reverted as the trigger is an independent process).

Incidentally, this highlights a gap in CI - it tests that the sample _builds_, but doesn't test that it loads and runs - it would be good to improve our coverage here.

dicej:
The recent `wit-bindgen` update broke this example due to changes in how WIT represents interface and world namespaces.  Specifically, the host defines the `config` interface under `fermyon:spin`, whereas the `spin-timer` world is defined under `fermyon:examples`, so we need to tell WIT they're in different packages.

Fixes #1662

### DIFF
--- a/examples/spin-timer/Cargo.lock
+++ b/examples/spin-timer/Cargo.lock
@@ -88,12 +88,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e2d098ff73c1ca148721f37baad5ea6a465a13f9573aba8641fbbbae8164a54e"
 
 [[package]]
-name = "arrayvec"
-version = "0.7.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
-
-[[package]]
 name = "async-channel"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


rylev:
Is there some way we can catch breakage like this in CI?

itowlson:
@rylev It might be a bit laborious, but probably worth doing.  The test would need to:

1. Build the plugin (we already do this I think)
2. Tar the plugin, and update the manifest
3. Install the plugin
4. Build and run the guest
5. Monitor that the guest ran for at least X seconds without exiting (10-15 seconds should suffice), and that it printed the expected text (this is probably pretty similar to the Redis e2e tests)

I'd suggest we do this as a separate piece of work so it doesn't drag this PR out.

itowlson (approved):
Works On My Machine (TM) - thanks for figuring this one out!

Issue #1548, opened 2023-06-01
`spin plugins upgrade -v` does not work for latest plugin version


itowlson:
```
$ spin plugins update
Plugin information updated successfully

$ spin plugins list
js2wasm 0.1.0
js2wasm 0.2.0
js2wasm 0.3.0
js2wasm 0.4.0 [installed]
py2wasm 0.1.0 [installed]
py2wasm 0.1.1
py2wasm 0.2.0

$ spin plugins upgrade py2wasm -v 0.2.0
Error: plugin 'py2wasm' not found at expected location /home/ivan/.local/share/spin/plugins/.spin-plugins/manifests/py2wasm/py2wasm@0.2.0.json: No such file or directory (os error 2)

$ ls -1 /home/ivan/.local/share/spin/plugins/.spin-plugins/manifests/py2wasm/
py2wasm@0.1.0.json
py2wasm@0.1.1.json
py2wasm.json        # the culprit!
```

Why does this matter?  Well, maybe it doesn't.  If you want the latest version, save yourself some keystrokes and skip the `-v`.  But it is a funny little inconsistency.

(The reason I came across this was planning for what I needed to do if the upgrade badger wanted to seek out the "last compatible" or "last in your major version" versions, rather than just "most recent or nothing."  For my own laziness purposes, I was planning to construct the "how to upgrade" instruction to always display the `-v`.  But this behaviour means I would have to construct the command differently for the truly latest version.  Which is not a problem, just a quirk.)

itowlson:
Apparently this got fixed somewhere along the line.  Good times

PR #1668, opened 2023-07-18
`spin plugins install` without version: install latest compatible not latest ever

### LINKED ISSUE #1661: `spin plugin install` fails if latest plugin version doesn't support current Spin version
on trying to install the plugins with spin 1.3, we get following error:

```
$) spin --version
spin 1.3.0 (9fb8256 2023-06-12)

$) spin plugin install js2wasm --yes

Error: Plugin is not compatible with this version of Spin (supported: >=1.4, actual: 1.3.0). Try running `spin plugins update && spin plugins upgrade --all` to install latest or override with `--override-compatibility-check`.
```

It seems because the latest release of js2wasm plugin has compatibility with spin version `>=1.4` only.

It might be useful to identify and install the `latest compatible` version automatically rather than failing the installation.

itowlson:
Fixes #1661 and #1548.

### DIFF
--- a/crates/plugins/src/error.rs
+++ b/crates/plugins/src/error.rs
@@ -14,6 +14,9 @@ pub enum Error {
 
     #[error("URL parse error {0}")]
     UrlParseError(#[from] url::ParseError),
+
+    #[error("{0}")]
+    Other(#[from] anyhow::Error),
 }
 
 /// Contains error details for when a plugin resource cannot be found at expected location
--- a/crates/plugins/src/lookup.rs
+++ b/crates/plugins/src/lookup.rs
@@ -30,7 +30,34 @@ impl PluginLookup {
         }
     }
 
-    pub async fn get_manifest_from_repository(
+    pub async fn resolve_manifest(
+        &self,
+        plugins_dir: &Path,
+        skip_compatibility_check: bool,
+        spin_version: &str,
+    ) -> PluginLookupResult<PluginManifest> {
+        let exact = self.resolve_manifest_exact(plugins_dir).await?;
+        if skip_compatibility_check
+            || self.version.is_some()
+            || exact.is_compatible_spin_version(spin_version)
+        {
+            return Ok(exact);
+        }
+
+        let store = crate::store::PluginStore::new(plugins_dir.to_owned());
+
+        // TODO: This is very similar to some logic in the badger module - look for consolidation opportunities.
+        let manifests = store.catalogue_manifests()?;
+        let relevant_manifests = manifests.into_iter().filter(|m| m.name() == self.name);
+        let compatible_manifests = relevant_manifests
+            .filter(|m| m.has_compatible_package() && m.is_compatible_spin_version(spin_version));
+        let highest_compatible_manifest =
+            compatible_manifests.max_by_key(|m| m.try_version().unwrap_or_else(|_| null_version()));
+
+        Ok(highest_compatible_manifest.unwrap_or(exact))
+    }
+
+    pub async fn resolve_manifest_exact(
         &self,
         plugins_dir: &Path,
     ) -> PluginLookupResult<PluginManifest> {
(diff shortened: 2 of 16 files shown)


crates/plugins/src/lookup.rs
```diff
@@ -41,37 +72,74 @@ impl PluginLookup {
...
+                    Some(self.name.clone()),
+                    expected_path.display().to_string(),
+                    e.to_string(),
+                ))
+            }),
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound && self.version.is_some() => {
+                // If a user has asked for a version by number, and the path doesn't exist,
+                // it _might_ be because it's the latest version. This checks for that case.
+                let latest = Self::new(&self.name, None);
+                if let Ok(manifest) = latest.resolve_manifest_exact_from_good_repo(plugins_dir) {
```
rajatjindal, line 108: should we assert for the version number here (to ensure latest-version == version-requested)?

itowlson: Oops, yes, I think we should.  I did a lengthy series of refactorings on this mess and lost it along the way.  Thank you!

crates/plugins/src/lookup.rs
```diff
@@ -30,7 +30,38 @@ impl PluginLookup {
...
+            let store = crate::store::PluginStore::new(plugins_dir.to_owned());
+
+            // This is very similar to some logic in the badger module
+            let manifests = store.catalogue_manifests()?;
+            let relevant_manifests = manifests.into_iter().filter(|m| m.name() == self.name);
+            let compatible_manifests = relevant_manifests.filter(|m| {
+                m.has_compatible_package() && m.is_compatible_spin_version(spin_version)
+            });
+            let highest_compatible_manifest =
+                compatible_manifests.max_by_key(|m| semver::Version::parse(&m.version).unwrap());
```
rylev, line 55: Is unwrapping the right thing to do here? Can we do `unwrap_or_else(|| semver::Version::new(0,0,0))`?

crates/plugins/src/lookup.rs
```diff
@@ -30,7 +30,38 @@ impl PluginLookup {
...
+
+            match highest_compatible_manifest {
+                Some(m) => Ok(m),
+                None => Ok(exact), // We can't do any better than latest after all
+            }
```
rylev, line 60:
```suggestion
           Ok(highest_compatible_manifest.unwrap_or(exact))
```

crates/plugins/src/lookup.rs
```diff
@@ -41,37 +72,78 @@ impl PluginLookup {
...
                 Some(self.name.clone()),
                 expected_path.display().to_string(),
                 e.to_string(),
-            ))
-        })?;
-        Ok(manifest)
+            )))
+        };
+
+        let manifest = match File::open(&expected_path) {
```
rylev, line 96: Doing I/O in async is usually a no/no. It's fine here though I suppose.

crates/plugins/src/lookup.rs
```diff
@@ -41,37 +72,78 @@ impl PluginLookup {
...
+                    Some(self.name.clone()),
+                    expected_path.display().to_string(),
+                    e.to_string(),
+                ))
+            }),
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound && self.version.is_some() => {
+                // If a user has asked for a version by number, and the path doesn't exist,
+                // it _might_ be because it's the latest version. This checks for that case.
+                let latest = Self::new(&self.name, None);
+                if let Ok(manifest) = latest.resolve_manifest_exact_from_good_repo(plugins_dir) {
```
rylev, line 108: Can this be refactored into a `match` with two patterns:

```rust
match latest.resolve_manifest_exact_from_good_repo(plugins_dir) {
  Ok(manifest) if semver::Version::parse(manifest.version()).ok() == self.version => Ok(manifest),
  _ => not_found(e)
}

Comment (itowlson) [reply crates/plugins/src/lookup.rs:108]: aha, yes, I was sure this could be simplified and this is the way - thanks!

File: crates/plugins/src/error.rs [hunk 6]
```diff
@@ -14,6 +14,9 @@ pub enum Error {
 
     #[error("URL parse error {0}")]
     UrlParseError(#[from] url::ParseError),
+
+    #[error("{0}")]
+    Other(#[from] anyhow::Error),
```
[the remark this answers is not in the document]
itowlson: I tried this but it gives

```
`?` couldn't convert the error to `error::Error`
the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
the trait `From<url::ParseError>` is implemented for `error::Error`
required for `Result<PluginManifest, error::Error>` to implement `FromResidual<Result<Infallible, anyhow::Error>>`
```

so keeping it as is

rajatjindal (approved):

kate-goldenring (approved):

rajatjindal:
Hi @itowlson can we backport this to 1.3.x?

PR #1670, opened 2023-07-19
Don't create directory when `spin add`-ing fileserver

### LINKED ISSUE #1669: `spin add static-fileserver` creates folder based on component name (still have to create `assets` dir manually)
Version: 
`spin 1.4.0 (7aab1fe 2023-07-11)`

Issue:
When adding a `static-fileserver` component to an application, Spin creates a new directory based on the component name which the user provides. For example:

```console
$ spin add static-fileserver
Enter a name for your new component: component-name
HTTP path: /static/...
Directory containing the files to serve: assets
```
If a user only provides a name `component-name` and then presses inter twice (accepting the defaults) Spin creates the following application structure. The component name directory is empty by default:

```console
$ tree .
.
├── Cargo.toml
├── component-name
├── spin.toml
└── src
    └── lib.rs
```

If the application is built and run as is, the following error message is returned:

```console
$ spin build --up

// --snip--

Error: Failed to collect file mounts for assets

Caused by:
      Cannot place ~/http-rust-app/assets: source must be a directory
```

If the user creates a new assets directory manually, the build works:

```console
$ mkdir assets
$ spin build --up

// --snip--

Available Routes:
  http-rust-app: http://127.0.0.1:3000 (wildcard)
  component-name: http://127.0.0.1:3000/static (wildcard)
```

Problem is that we a) had to manually create the `assets` directory and b) the application has an empty directory called `component-name`.

```console
.
├── Cargo.lock
├── Cargo.toml
├── assets
├── component-name
├── spin.toml
└── src
    └── lib.rs
```

Related: 
- [Create directory for static-fileserver template ](https://github.com/fermyon/spin/issues/1022)

itowlson:
Fixes #1669

### DIFF
--- a/crates/templates/src/lib.rs
+++ b/crates/templates/src/lib.rs
@@ -24,3 +24,6 @@ pub use manager::*;
 pub use run::{Run, RunOptions};
 pub use source::TemplateSource;
 pub use template::{Template, TemplateVariantInfo};
+
+#[cfg(test)]
+mod test_built_ins;
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -107,7 +107,11 @@ impl TemplateManager {
     /// Creates a `TemplateManager` for the default install location.
     pub fn try_default() -> anyhow::Result<Self> {
         let store = TemplateStore::try_default()?;
-        Ok(Self { store })
+        Ok(Self::new(store))
+    }
+
+    pub(crate) fn new(store: TemplateStore) -> Self {
+        Self { store }
     }
 
     /// Installs templates from the specified source.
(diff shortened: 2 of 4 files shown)


fibonacci1729 (approved):

michelleN (commented):
Is there a test we could add with this change?

itowlson:
@michelleN Yeah, we don't have a great venue for tests like this.  We do use some of the standard templates in the templates crate unit tests, but only for testing general properties of the templates system, not for their functionality.  Functionality tends to go more in the e2e tests, which are very heavyweight, and more geared around testing that apps run correctly.  I will have a play and see what I can do.

itowlson:
@michelleN Added a test - unfortunately it's dramatically larger than the actual fix...!

Issue #1559, opened 2023-06-08
`spin up` prints `spin trigger http` usage on bad flag


itowlson:
```
$ spin up --fie
error: Found argument '--fie' which wasn't expected, or isn't valid in this context

        If you tried to supply `--fie` as a value rather than a flag, use `-- --fie`

USAGE:
    spin trigger http [OPTIONS]

For more information try --help
Error: exit status: 2
```

dipankardas011:
@itowlson how to reproduce it?

itowlson:
@dipankardas011 Change to a directory with a Spin HTTP application (`spin.toml`) in it, and run `spin up --fie`, and note the `USAGE` line in the output:

```
# In a directory with a (HTTP trigger) spin.toml
ivan@hecate:~/testing/dioxus-test$ ls
Cargo.lock  Cargo.toml  images  spin.toml  src  target

# Run spin up with a bad flag
ivan@hecate:~/testing/dioxus-test$ spin up --fie
error: Found argument '--fie' which wasn't expected, or isn't valid in this context

        If you tried to supply `--fie` as a value rather than a flag, use `-- --fie`

USAGE:
    spin trigger http [OPTIONS]

For more information try --help
Error: exit status: 2
```

USAGE should refer to `spin up` rather than the internal `spin trigger http` command.

dipankardas011:
@itowlson what was the use of -i before

from the spin up -h
```
-e for environment vars
-f for spin.tmol specify
```

dipankardas011:
Currently these things  are present in spin up
```bash
spin up -f spin.toml -e SECRET=342we3 -L logs --listen 0.0.0.0:9090
```

dipankardas011:
Also found this https://developer.fermyon.com/spin/cli-reference#spin-up-http-trigger

itowlson:
This is an error case.  `--fie` is an invalid flag.  `--bibbly-bobbly` would do equally well.

You're right to identify that Spin is trying to invoke the HTTP trigger, and that this is why the USAGE section references it.  The (minor) bug is that the error message - which comes from the HTTP trigger - is inviting users who enter a bad flag to run `spin trigger http` (which is an internal command that users should never run) rather than `spin up` (which is the user-facing interface).

dipankardas011:
okay got it
So to summarize for any invalid command it is suggesting a internal command (gets exposed) for the users to run which is not meant to be used by them

dipankardas011:
[...]

PR #1671, opened 2023-07-19
fix: added error handling `spin trigger [USAGE]`

dipankardas011:
## Problem
all the subcommand for up and watch don't have validation for the `Upcommand.trigger_args` and so at the final step it gets invoked by  invalid args thus causing error

## Solution

Workaround
add usage macro to the `struct TriggerExecutorCommand`

### Issue linked

- [x] Closes: #1559

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -28,7 +28,10 @@ pub const SPIN_WORKING_DIR: &str = "SPIN_WORKING_DIR";
 
 /// A command that runs a TriggerExecutor.
 #[derive(Parser, Debug)]
-#[clap(next_help_heading = "TRIGGER OPTIONS")]
+#[clap(
+    usage = "spin [COMMAND] [OPTIONS]",
+    next_help_heading = "TRIGGER OPTIONS"
+)]
 pub struct TriggerExecutorCommand<Executor: TriggerExecutor>
 where
     Executor::RunConfig: Args,


itowlson:
@dipankardas011 It looks like this is early work-in-progress at the moment - is that right?  If so, is it okay to convert to draft so people know it's not complete?

dipankardas011:
> @dipankardas011 It looks like this is early work-in-progress at the moment - is that right? If so, is it okay to convert to draft so people know it's not complete?

By what I found using simple println statements that during final command run the args passed in command line also getting send here that's why it's returning this message
It's nothing to do with usage thing
If it fails in the trigger http|redis
It will return it's not valid thing

dipankardas011:
Due to which I am not sure how to correct it

itowlson:
Yes, the arguments are passed through to the trigger, which tries to make sense of them.  The USAGE message is printed by clap (the command line argument parser we use) when it fails.

And yeah, I'm not sure which steps to use to fix it either.  It's possible we could get clap to lie about the command name, so that it prints `up` instead of `trigger http`, but I don't know what other sorts of confusion that might cause...!  If you have an easy solution in mind, we'd be delighted to have it; but if not, it may be more rewarding to turn your efforts to something both easier and more impactful.

dipankardas011:
> Yes, the arguments are passed through to the trigger, which tries to make sense of them. The USAGE message is printed by clap (the command line argument parser we use) when it fails.
> 
> And yeah, I'm not sure which steps to use to fix it either. It's possible we could get clap to lie about the command name, so that it prints `up` instead of `trigger http`, but I don't know what other sorts of confusion that might cause...! If you have an easy solution in mind, we'd be delighted to have it; but if not, it may be more rewarding to turn your efforts to something both easier and more impactful.

possible solutions (but open for more solutions)
- make the trigger http | redis to lie in usage thing
- do some validation in the Upcommand impl and then send the request to the backend trigger http
- or add the validation to the trigger http

itowlson:
We can't validate in the `up` command, unless we provide some additional protocol between `up` and the trigger, which I don't think is worth it for this - it would be a significant design exercise for a minor cosmetic issue.

I'm not sure what you mean by "add the validation to the trigger http".  `spin trigger http` is already validating, isn't it?

The only fix I can think of is getting clap in the trigger to lie: I don't know how feasible that is, I reckon either super easy (if clap supports it) or absurdly difficult (otherwise).  Feel free to give it a go - but if it doesn't fall out quickly then don't spend time banging your head against a wall...!

dipankardas011:
> We can't validate in the `up` command, unless we provide some additional protocol between `up` and the trigger, which I don't think is worth it for this - it would be a significant design exercise for a minor cosmetic issue.
> 
> I'm not sure what you mean by "add the validation to the trigger http".  `spin trigger http` is already validating, isn't it?
> 
> The only fix I can think of is getting clap in the trigger to lie: I don't know how feasible that is, I reckon either super easy (if clap supports it) or absurdly difficult (otherwise).  Feel free to give it a go - but if it doesn't fall out quickly then don't spend time banging your head against a wall...!

Sure clap has usage macro
But thanks for this issue learned how to navigate project based on rust

dipankardas011:
```bash
/spin/target/debug/spin up wedcwer
error: Found argument 'wedcwer' which wasn't expected, or isn't valid in this context

USAGE:
    spin [COMMAND] [OPTIONS]

Error: exit status: 2
```

itowlson (approved):
Ha, I like that the usage string doesn't actively lie, just leaves it non-specific.  Thanks for this!

itowlson:
Could I ask that once you've dealt to the lint issue you also squash the commits please?  Thanks!  (We can't squash-merge because of the signature requirement.)

dipankardas011:
> Could I ask that once you've dealt to the lint issue you also squash the commits please?  Thanks!  (We can't squash-merge because of the signature requirement.)

Sure

dipankardas011:
@itowlson the second commit was for the failed lint test

itowlson:
Sorry for the delay @dipankardas011 we have all been away a bit (and I have been away a lot) - merged now!  Thank you!

dipankardas011:
> Sorry for the delay @dipankardas011 we have all been away a bit (and I have been away a lot) - merged now!  Thank you!

Np

PR #1674, opened 2023-07-20
Print available routes as links

### LINKED ISSUE #1672: Underline the application endpoint to make it obvious it's a link
It would be nice to have an underline on the application endpoint in the output of `spin deploy` to make it clearer it is a hyperlink. 

currently: 

Screenshot 2023-07-19 at 15 07 22

potential alternative formatting

Screenshot 2023-07-19 at 15 08 13

itowlson:
Fixes #1672.

(Though actually this only addresses `spin up` - a `cloud-plugin` port would be needed to address @ingride's original issue.)

The implementation I ended up with is not very tidy at the call site, because it needs multiple print statements for each output line.  I looked at using alternate formatting to linkify routes (allowing e.g. `println!("  {id}: {url:#}{wild}");`) but I couldn't see a way to skip the formatting if the output stream wasn't a terminal.  I also wondered about a `terminal::print_link!` macro that would accept the whole line but only linkify the URL bit, but I couldn't see how to handle that level of generality.  So, Rust whizzes, if you have any ideas, I am eager to learn!

Anyway here is what the result looks like - this is WSL in Windows Terminal, other terminals may vary!

image

(I didn't linkify the base URL, because I was too lazy to scan the components to see if the base URL was actually mapped - I always click on a component rather than the base.  But if other folks use the base then I can do it too.)

### DIFF
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -170,6 +170,14 @@ impl RoutePattern {
         format!("{}{}", Self::sanitize(base.into()), Self::sanitize(path))
     }
 
+    /// Build a URL from a base prefix and the pattern
+    pub fn urlify(&self, base_url: &str) -> (String, impl fmt::Display) {
+        match self {
+            Self::Exact(path) => (format!("{base_url}{path}"), Wildness::Exact),
+            Self::Wildcard(pattern) => (format!("{base_url}{pattern}"), Wildness::Wild),
+        }
+    }
+
     fn absolutize<S: Into<String>>(s: S) -> String {
         let s = s.into();
         if s.starts_with('/') {
@@ -197,6 +205,20 @@ impl fmt::Display for RoutePattern {
         match &self {
             RoutePattern::Exact(path) => write!(f, "{}", path),
             RoutePattern::Wildcard(pattern) => write!(f, "{} (wildcard)", pattern),
+        }
+    }
+}
+
+enum Wildness {
+    Exact,
+    Wild,
+}
+
+impl fmt::Display for Wildness {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match &self {
+            Self::Exact => write!(f, ""),
+            Self::Wild => write!(f, " (wildcard)"),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


crates/http/src/routes.rs
```diff
@@ -201,6 +209,20 @@ impl fmt::Display for RoutePattern {
     }
 }
 
+enum Wildness {
```
rylev, line 212: This is cute, but can we think of a clearer name?  I'm not sure what it would be, but I'm pretty sure it isn't this.

crates/terminal/src/lib.rs
```diff
@@ -129,6 +129,24 @@ macro_rules! ceprint {
     };
 }
 
+pub fn print_link(url: &str) {
+    if atty::is(atty::Stream::Stdout) {
```
rylev, line 133: Just a note that we should eventually move the std library support for this: https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html#tymethod.is_terminal

crates/terminal/src/lib.rs
```diff
@@ -129,6 +129,24 @@ macro_rules! ceprint {
...
+        print!("{}", link(url));
+    } else {
+        print!("{url}");
+    }
+}
+
+#[cfg(unix)]
+fn link(url: &str) -> String {
+    format!("\x1B]8;;{url}\x1B\\{url}\x1B]8;;\x1B\\")
```
rylev, line 142: Is there a reason we're not using termcolor for this? Is there not support for what we're trying to achieve?

crates/terminal/src/lib.rs
```diff
@@ -129,6 +129,24 @@ macro_rules! ceprint {
     };
 }
 
+pub fn print_link(url: &str) {
```
rylev, line 132: Would it be cleaner if this didn't directly do the printing and just returned a `Cow<str>` of the escaped string so that it can be used at the callsite by a `println!` macro?

crates/terminal/src/lib.rs
dipankardas011: yes by adding color would also help

dipankardas011: not an expert but can we use https://docs.rs/colored/latest/colored/

dipankardas011: or else we can us the color as raw hexadecimal format
which also works when not want to add more dependencies

PR #1675, opened 2023-07-20
Exit early when trigger process fails

### LINKED ISSUE #966: Expunge `Error: exit status: 1`
Some `spin up` errors get an extra line `Error: exit status: 1` appended to them.  I think this happens when `spin up` succeeded in handing off to `spin trigger ...` but the trigger command then errored out.

Is is possible to remove this?  It's no biggie, just a bit cluttery.

rylev:
Fixes #966 

This exits the main process if the trigger subprocess fails early.

### DIFF
--- a/crates/terminal/src/lib.rs
+++ b/crates/terminal/src/lib.rs
@@ -75,7 +75,6 @@ fn color_choice(stream: atty::Stream) -> termcolor::ColorChoice {
 #[macro_export]
 macro_rules! step {
     ($step:expr, $($arg:tt)*) => {{
-
         $crate::cprint!($crate::colors::bold_green(), $step);
         print!(" ");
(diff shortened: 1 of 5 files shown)


lann (changes requested):
I'm pretty sure this will leave tempfiles behind.

lann:
As an alternative, we could introduce a `ExitStatusError` (mirroring the unstable `std::process::ExitStatusError`), downcast to it in our outer `main`, and `exit(err.code)` there.

rylev:
Good point. I'll move to your suggestion.

lann (approved):
Thanks! Non-blocking suggestion to add a comment at the downcast site.

lann (approved):

PR #1678, opened 2023-07-24
Update to Wasmtime 11

tschneidereit:
Mostly this update was uneventful. However there's a change to the WasiCtxBuilder for Preview2 that made using it much more annoying. @alexcrichton, maybe you have ideas for how to do this in a better way?

### DIFF
--- a/crates/core/src/store.rs
+++ b/crates/core/src/store.rs
@@ -120,6 +120,7 @@ pub struct StoreBuilder {
     engine: wasmtime::Engine,
     epoch_tick_interval: Duration,
     wasi: std::result::Result<WasiCtxBuilder, String>,
+    preview2_dummy_ctx: Option<wasi_preview2::WasiCtxBuilder>,
     host_components_data: HostComponentsData,
     store_limits: StoreLimitsAsync,
 }
@@ -136,6 +137,7 @@ impl StoreBuilder {
             engine,
             epoch_tick_interval,
             wasi: Ok(wasi.into()),
+            preview2_dummy_ctx: Some(wasi_preview2::WasiCtxBuilder::new()),
             host_components_data: host_components.new_data(),
             store_limits: StoreLimitsAsync::default(),
         }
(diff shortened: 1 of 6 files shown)


rylev:
@tschneidereit I think you'll need to update https://github.com/fermyon/spin-componentize and https://github.com/fermyon/wit-bindgen-backport as well. I'll take a closer look later today.

alexcrichton:
> @alexcrichton, maybe you have ideas for how to do this in a better way?

I'll send a PR for updating the upstream APIs to be more amenable to patterns like this

crates/core/src/store.rs
```diff
@@ -410,6 +432,19 @@ impl StoreBuilder {
...
+    let owned = std::mem::replace(ctx, dummy);
+    let owned = f(owned);
+    let dummy = std::mem::replace(ctx, owned);
+    assert_eq!(ctx_addr, std::ptr::addr_of!(*ctx));
+    dummy
+}
+
 /// A builder of a `WasiCtx` for all versions of Wasi
 #[allow(clippy::large_enum_variant)]
 enum WasiCtxBuilder {
```
alexcrichton, line 450: The conclusion of https://github.com/bytecodealliance/wasmtime/pull/6770 was that this should probably use a `WasiCtxBuilder` in both arms to use the move-style everywhere instead of on just one half to avoid some of the weirdness here. Otherwise the other possible fix is to update to `&mut`-style builders but that has other API drawbacks that aren't a clear win at this time.

The major difference from preview1 is that the preview2 `WasiCtx` doesn't have the mutators that preview1's context has I believe.

tschneidereit: I commented over there, because I'm still not entirely convinced, but we can certainly live with things as they are now

tschneidereit: I'm also not too sure how using a builder in both arms would actually make things easier, but I'm probably missing something there

tschneidereit:
> @tschneidereit I think you'll need to update [fermyon/spin-componentize](https://github.com/fermyon/spin-componentize) and [fermyon/wit-bindgen-backport](https://github.com/fermyon/wit-bindgen-backport) as well. I'll take a closer look later today.

Thanks Ryan! I had updated the wit-bindgen backport, but missed spin-componentize. That's https://github.com/fermyon/spin-componentize/pull/16 now.

rylev:
Closing this as I believe we want to wait until wasmtime 13 to upgrade. @tschneidereit let me know if that's not a mistake and we can reopen this.

PR #1679, opened 2023-08-02
Write componentized component to temp directory

rylev:
At the small one-time cost of writing the component to disk, we can more easily allow folks access to the component for experimental use with component tooling like wasm-tools compose.

### DIFF
--- a/crates/trigger/src/loader.rs
+++ b/crates/trigger/src/loader.rs
@@ -56,7 +56,12 @@ impl Loader for TriggerLoader {
             )
         })?;
         let component = spin_componentize::componentize_if_necessary(&bytes)?;
-        if let Err(e) = std::fs::write(self.working_dir.join("component.wasm"), &component) {
+
+        let file_name = path
+            .file_name()
+            .unwrap_or_else(|| std::ffi::OsStr::new("component.wasm"));
+        let component_dump_path = self.working_dir.join(file_name);
+        if let Err(e) = std::fs::write(component_dump_path, &component) {
             tracing::warn!("Could not write component bytes to working directory: {e}")
         }
         let was_already_component = matches!(component, std::borrow::Cow::Borrowed(_));
--- a/crates/trigger/src/loader.rs
+++ b/crates/trigger/src/loader.rs
@@ -56,6 +56,9 @@ impl Loader for TriggerLoader {
             )
         })?;
         let component = spin_componentize::componentize_if_necessary(&bytes)?;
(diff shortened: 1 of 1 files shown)


crates/trigger/src/loader.rs
```diff
@@ -56,6 +56,9 @@ impl Loader for TriggerLoader {
             )
         })?;
         let component = spin_componentize::componentize_if_necessary(&bytes)?;
+        if let Err(e) = std::fs::write(self.working_dir.join("component.wasm"), &component) {
```
itowlson, line 59: In a multi-component application, this will (as far as I can tell) result in each component overwriting the previous one.

itowlson (changes requested):
A question about the user experience: the 99% case is that the working directory is a temp directory which gets deleted as soon as Spin exits (and is not displayed unless you crank up the log level).  Is the presumption that if somebody wants to capture the componentised wrapper then they would specify a working directory in `spin up`?

Other than that, one bug but otherwise looks fine!

crates/trigger/src/loader.rs
rylev: Good catch - unfortunately fixing this won't be easy as we don't have the name of the component accessible at this point. We could use the file name of the module but that's not guaranteed to be unique (e.g., `/some/path/module.wasm` and `/some/other/path/module.wasm). I'll need to noodle a bit on this...

rylev:
>  Is the presumption that if somebody wants to capture the componentised wrapper then they would specify a working directory in spin up?

@itowlson that is my thought currently. I don't expect this to be a feature many will use. It's a temporary escape hatch for component testing until we figure out a more permanent solution.

crates/trigger/src/loader.rs
```diff
@@ -56,6 +56,14 @@ impl Loader for TriggerLoader {
             )
         })?;
         let component = spin_componentize::componentize_if_necessary(&bytes)?;
+
+        let file_name = path
+            .file_name()
```
itowlson, line 61: As you note, there could be clashes here, and I believe with an OCI app this will be some unhelpful digest string.  Is there any downside to adding a component ID parameter to the `load_component` function, and using that to generate the dump file name?  I had a poke at it and the only internal caller seems to have ready access to the ID so the blast radius is low; I guess it depends on if there are any external callers.

rylev:
I'm going to close this for now - this was originally just meant as a small QOL fix for me, but I think there's enough open questions that impact others that I'll need to have a deeper think about it.

PR #1683, opened 2023-08-08
feat(sdk/zig): start off Zig SDK with config component

tensorush:
Hi, I'd love for Spin to have a fully-fledged Zig SDK!

This PR only implements the config component, please let me know if there's something else I should add to start us off.


rylev:
@tensorush thanks for this! Looks like a really great start. 🎉

While in the long run I can certainly see there being an official Zig SDK maintained by the Spin project, I don't think we're ready for that today for the simple reason that the Spin project doesn't have enough Zig expertise to ensure this is maintained properly. What's more, the path forward for SDKs is for them to live outside of the Spin repo which we've done for other SDKs that came after Go and Rust like [Python](https://github.com/fermyon/spin-python-sdk) and [JS](https://github.com/fermyon/spin-js-sdk) and which we eventually hope to do for Rust and Go.

I think the right path forward is for this to live in a separate repo (perhaps on your personal GitHub org to start). We have the [ability to promote](https://developer.fermyon.com/hub) such projects even when they aren't under the Fermyon GitHub org. If the day comes that it makes more sense for this to live in the Fermyon GitHub org then we can consider that then. 

How does that sound?

PR #1684, opened 2023-08-09
core: Replace epoch deadline with yield

lann:
With async yielding, instance execution can be implemented by dropping the async call future, with e.g. tokio::time::timeout.

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -39,7 +39,7 @@ pub use io::OutputBuffer;
 pub use store::{Store, StoreBuilder, Wasi, WasiVersion};
 
 /// The default [`EngineBuilder::epoch_tick_interval`].
-pub const DEFAULT_EPOCH_TICK_INTERVAL: Duration = Duration::from_millis(10);
+pub const DEFAULT_EPOCH_TICK_INTERVAL: Duration = Duration::from_millis(1);
 
 const MB: u64 = 1 << 20;
 const GB: u64 = 1 << 30;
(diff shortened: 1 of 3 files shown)


crates/core/src/store.rs
```diff
@@ -56,35 +56,13 @@ pub enum WasiVersion {
...
     }
-
-    /// Sets the execution deadline.
-    ///
-    /// This is a rough deadline; an instance will trap some time after this
-    /// deadline, determined by [`EngineBuilder::epoch_tick_interval`] and
-    /// details of the system's thread scheduler.
-    ///
-    /// See [`wasmtime::Store::set_epoch_deadline`](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.set_epoch_deadline).
-    pub fn set_deadline(&mut self, deadline: Instant) {
```
lann, line 75: Based on some searching I'm reasonably confident that this method isn't being used anywhere (public).

crates/core/src/lib.rs
```diff
@@ -37,7 +37,7 @@ pub use io::OutputBuffer;
 pub use store::{Store, StoreBuilder, Wasi, WasiVersion};
 
 /// The default [`EngineBuilder::epoch_tick_interval`].
-pub const DEFAULT_EPOCH_TICK_INTERVAL: Duration = Duration::from_millis(10);
+pub const DEFAULT_EPOCH_TICK_INTERVAL: Duration = Duration::from_micros(100);
```
lann, line 40: This arbitrarily updates the original arbitrary value, which "felt too slow" to me. :shrug:

crates/core/src/store.rs
```diff
@@ -370,16 +364,22 @@ impl StoreBuilder {
 
         inner.limiter_async(move |data| &mut data.store_limits);
 
-        // With epoch interruption enabled, there must be _some_ deadline set
-        // or execution will trap immediately. Since this is a delta, we need
-        // to avoid overflow so we'll use 2^63 which is still "practically
-        // forever" for any plausible tick interval.
-        inner.set_epoch_deadline(u64::MAX / 2);
+        let delta = self.yield_interval.as_nanos() / self.epoch_tick_interval.as_nanos();
```
rylev, line 367: This will panic if someone passes a `Duration` of 0 for the tick interval. We probably want to prevent that.

lann: Added `ensure!` above.

lann:
Uhh I think I lost track of this back when e2e tests were broken for a while. Will refresh.

lann:
Yep, but it could still be reviewed and then I could just merge it after the next rebase :slightly_smiling_face:

dicej:
Sorry I'm a year and a half late on this; is it still relevant?

lann:
Oof I'll have to set a reminder to come back and check in the new year :grimacing:

dicej:
I'm going to go ahead and close this so it doesn't keep showing up on my GitHub "reviews requested" list.  Feel free to reopen if it's still relevant.

PR #1685, opened 2023-08-10
minor ux improvement when validation on config fails

rajatjindal:
while playing around with config variables, I ran into some problems and it was not easy to identify which config was having problem.

e.g.

when i had config like below in `spin.toml`
```
allowed_channel = { required = true, secret = true, default = "x" }

OR

allowed_channel = { secret = true }
```

I get error 

```
$) spin up

Error: variable has both `required` and `default` set

```

with the fix, we get following error msg in both the scenarios:

```
$) spin up

Error: variable 'allowed_channel': variable should either have `required` set to true OR a non-empty default value
```

### DIFF
--- a/crates/loader/src/local/mod.rs
+++ b/crates/loader/src/local/mod.rs
@@ -167,7 +167,13 @@ async fn prepare(
     let variables = raw
         .variables
         .into_iter()
-        .map(|(key, var)| Ok((key, var.try_into()?)))
+        .map(|(key, var)| {
+            Ok((
+                key.clone(),
+                var.try_into()
+                    .map_err(|err| anyhow!("variable '{}': {}", key, err))?,
+            ))
+        })
         .collect::<Result<_>>()?;
 
     Ok(Application {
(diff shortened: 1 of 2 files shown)


crates/loader/src/local/mod.rs
```diff
@@ -167,7 +167,13 @@ async fn prepare(
     let variables = raw
         .variables
         .into_iter()
-        .map(|(key, var)| Ok((key, var.try_into()?)))
+        .map(|(key, var)| {
+            Ok((
+                key.clone(),
+                var.try_into()
+                    .map_err(|err| anyhow!("variable '{}': {}", key.clone(), err))?,
```
rylev, line 174: This last call to `key.clone()` shouldn't be be necessary

rylev (approved):
Looks good!

kate-goldenring (approved):

crates/loader/src/local/mod.rs
rajatjindal: 👍

PR #1690, opened 2023-08-14
Allow getting http trigger metadata from (locked)app from spin-http

rylev:
Allows getting http trigger metadata from an app and locked app without needing to pull in the entire `http-trigger` crate (by instead using the more lightweight `spin_http` crate). 

This also extends `LockedApp` to be able to easily deserialize metadata based on a given metadata key.


lann (approved):

vdice:
If you'd like to rebase, e2e tests should now be running successfully.

PR #1695, opened 2023-08-15
fix(registry/push): skip pushing empty blobs

radu-matei:
This commit checks the length of a blob before adding it as a layer to be uploaded to a container registry.

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -106,7 +106,10 @@ impl Client {
                 // locked application file with the file digest.
                 for entry in WalkDir::new(&source) {
                     let entry = entry?;
-                    if entry.file_type().is_file() && !entry.file_type().is_dir() {
+                    if entry.file_type().is_file()
+                        && !entry.file_type().is_dir()
+                        && entry.metadata()?.len() > 0
+                    {
                         tracing::trace!(
                             "Adding new layer for asset {:?}",
                             spin_loader::to_relative(entry.path(), &source)?


vdice (approved):

crates/oci/src/client.rs
```diff
@@ -106,7 +106,10 @@ impl Client {
                 // locked application file with the file digest.
                 for entry in WalkDir::new(&source) {
                     let entry = entry?;
-                    if entry.file_type().is_file() && !entry.file_type().is_dir() {
+                    if entry.file_type().is_file()
```
endocrimes, line 109: We should at least include a DEBUG log here for skipped blobs to help future-us with debugging

itowlson:
Suppose I have an app that opens a file, and that file can legitimately be empty.  Suppose further that, during a push, that file is indeed empty, so it is not pushed.

When I then run the pushed app from the registry, what will happen:

* The file open fails because the file doesn't exist? or
* The pull/load process reconstitutes the empty file?

radu-matei:
Currently, the push/pull process will not reconstruct the empty file.
We can definitely work on reconstructing the file, but I would not block this PR on it.

vdice:
Wondering if we're closer to alignment on what we'd want to get in with this PR and what we might be okay with as follow-ups?

My two cents: perhaps adding a debug log as mentioned in https://github.com/fermyon/spin/pull/1695#discussion_r1294805714 would be prudent for this PR.  The topic around reconstruction of empty files seems a good candidate for a follow-up.

PR #1697, opened 2023-08-15
ci(.github): discontinue use of unmaintained actions

vdice:
This update was motivated by the slew of deprecation warnings spotted in recent CI runs (for example, this [build workflow run](https://github.com/fermyon/spin/actions/runs/5868420592) and this [release workflow run](https://github.com/fermyon/spin/actions/runs/5868420584)).  It turns out they all stem from the [actions-rs/toolchain](https://github.com/actions-rs/toolchain) and [actions-rs/cargo](https://github.com/actions-rs/cargo) actions, which unfortunately no longer appear to be maintained (ref https://github.com/actions-rs/toolchain/issues/216 and https://github.com/actions-rs/cargo/issues/216).

While there may be alternative actions (such as https://github.com/dtolnay/rust-toolchain), I propose we just invoke the `rustup` and `cargo` commands directly, seeing as how our use case is fairly straightforward.

Example runs from this changeset: [build](https://github.com/vdice/spin/actions/runs/5872410262) and [release](https://github.com/vdice/spin/actions/runs/5872410261).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -110,10 +110,8 @@ jobs:
           echo 'linker = "aarch64-linux-gnu-gcc"' >> ${HOME}/.cargo/config.toml
 
       - name: build release
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: "--all-features --release ${{ matrix.config.extraArgs }}"
+        shell: bash
+        run: cargo build --all-features --release ${{ matrix.config.extraArgs }}
 
       - name: Sign the binary with GitHub OIDC token
         shell: bash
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -85,14 +88,18 @@ jobs:
...
 
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: 1.68
-          default: true
-          target: ${{ matrix.config.target }}
+        shell: bash
+        run: |
+          rustup toolchain install ${{ env.RUST_VERSION }}
```
michelleN, line 93: for my own curiosity: how do you have access to the rustup binary here? does it come standard with the os?

vdice: Yes, Rust tools like rustup and cargo come pre-installed in the GitHub runners.

michelleN (approved):

Issue #1700, opened 2023-08-21
e2e tests running out of space


vdice:
The e2e test job has been failing for the last week+.  Based on errors in the workflow, the host is running out of space.  Whether or not this is the actual error or just prevents us from seeing other errors isn't known at this time, as we don't get any logs.

Sample: https://github.com/fermyon/spin/actions/runs/5884367775 with messages:

> The hosted runner: GitHub Actions 2 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

> You are running out of disk space. The runner will stop working when the machine runs out of disk space. Free space left: 63 MB

vdice:
When running the tests locally, I see the following failures.  Granted, I haven't run these in some time.  Is there any sqlite-related setup I may be missing on my machine?  And/or can others repro as well?
```
test spinup_tests::assets_routing_works ... ok
test spinup_tests::http_c_works ... ok
test spinup_tests::http_go_works ... ok
test spinup_tests::http_php_works ... ignored
test spinup_tests::header_dynamic_env_works ... ok
test spinup_tests::header_env_routes_works ... ok
thread 'spinup_tests::http_python_works' panicked at 'called `Result::unwrap()` on an `Err` value: running app

Caused by:
    error running spin up.
    stdout ["Logging component stdio to \".spin/logs/\""]
    stderr: ["Error: Failed to instantiate component 'http-py-generated'", "", "Caused by:", "    0: failed to reduce input adapter module to its minimal size", "    1: adapter module does not have export `sqlite:open`", "Error: exit status: 1"]
    ', tests/testcases/mod.rs:92:34
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test spinup_tests::http_python_works ... FAILED
thread 'spinup_tests::http_js_works' panicked at 'called `Result::unwrap()` on an `Err` value: running app

Caused by:
    error running spin up.
    stdout ["Logging component stdio to \".spin/logs/\""]
    stderr: ["Error: Failed to instantiate component 'http-js-generated'", "", "Caused by:", "    0: failed to reduce input adapter module to its minimal size", "    1: adapter module does not have export `sqlite:open`", "Error: exit status: 1"]
    ', tests/testcases/mod.rs:404:34
test spinup_tests::http_js_works ... FAILED
test spinup_tests::http_rust_outbound_pg_works ... ok
test spinup_tests::http_swift_works ... ok
[...]

PR #1702, opened 2023-08-22
ci(build.yml): run e2e job on larger runner

vdice:
Run the e2e job on a larger runner.

This will incur a $0.016 per minute rate for using this Ubuntu/4-core larger runner, per https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates.  ~~If a given e2e run lasts 45 minutes, this would incur a cost of $0.72.  (Note: minutes used for a run _should_ decrease noticeably on the larger runner, but we don't know how much yet.)~~ _see below for estimated costs._  This may be more than we'd like to spend depending on level of usage, so we'll want to monitor costs.  Still, I am curious to see if it helps the e2e job run to completion; also keen to use this as an exercise on how/where we might utilize these larger runners.

Ref https://github.com/fermyon/spin/issues/1700

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,7 +150,8 @@ jobs:
         run: make test-sdk-go
 
   e2e-tests:
-    runs-on: ubuntu-22.04
+    # run on a larger runner for more SSD/resource access
+    runs-on: ubuntu-22.04-4core-spin
     needs: build-rust-ubuntu
     steps:
       - uses: actions/checkout@v3


vdice:
Okay, the e2e run on the larger runner for this PR completed in 17m 19s, accruing a cost of ~$0.28.  Assuming the run times are similar on runners of this type, that should be a reasonable guage for per-run costs.

In other news, it completed successfully as well, so we were definitely running into resource constraints as mentioned in https://github.com/fermyon/spin/issues/1700.  There still can and should be a discussion of slimming down resource needs, but this PR just focuses on performance on a larger runner.

kate-goldenring:
@vdice to further reduce costs, would it be worth canceling any previous runs of e2e tests anytime a new commit is pushed to a PR instead of running the workflow for every commit. Maybe we already do this, but here is an example workflow that can cancel previous ones that are running whenever a new commit is pushed: https://github.com/project-akri/akri/blob/main/.github/workflows/cancel-previous-pr-workflows.yml. It is helpful for avoiding running the tests many times in the scenario where a PR author prefers many small commits.

vdice:
@kate-goldenring great suggestion, I think we should add logic along the lines of what you've mentioned.  It looks like the particular [cancel-workflow-action](https://github.com/styfle/cancel-workflow-action) used in the linked workflow now suggests to use the [first-class concurrency support](https://docs.github.com/en/actions/using-jobs/using-concurrency) in GH actions to do the same.  A great candidate for a follow-up PR.

radu-matei (approved):

PR #1703, opened 2023-08-22
ci(build.yml): add concurrency config

vdice:
Update the build.yml GH workflow with [concurrency](https://docs.github.com/en/actions/using-jobs/using-concurrency) configuration.

- Serializes runs for a given ref

- In-flight run(s) for a given ref (besides the `main` branch) are cancelled if out-of-date (i.e. when newer commits are pushed to the ref, eg PR branch, etc).  This becomes especially important when some parts of CI may utilize larger runners at an increased per-minute cost.

Ref https://github.com/fermyon/spin/pull/1702#issuecomment-1688730028 for further context

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,12 @@ on:
       - "README.md"
       - "tests/README.md"
 
+# Serialize workflow runs per ref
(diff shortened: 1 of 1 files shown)


kate-goldenring (approved):
Thanks, @vdice! The world just got a little greener 🌲

PR #1708, opened 2023-08-24
Update release process to include updating homebrew formula

kate-goldenring:
Spin is now installable via `brew`! During releases, we should update the [homebrew formula](https://github.com/fermyon/homebrew-tap) to point to the latest release artifacts.

ref #641

### DIFF
--- a/docs/content/release-process.md
+++ b/docs/content/release-process.md
@@ -80,6 +80,10 @@ To cut a release of Spin, you will need to do the following:
    `--certificate-identity` value should match this release, e.g.
    `https://github.com/fermyon/spin/.github/workflows/release.yml@refs/tags/v1.1.0`.
 
+1. Create a Pull Request into Fermyon's Hombrew tap repository updating the [Spin
+   formula](https://github.com/fermyon/homebrew-tap/blob/main/Formula/spin.rb). In the formula,
+   update the version, point to the latest release artifacts, and set their correct sha256 digests.
+
 The release is now complete!
 
 [release action]: https://github.com/fermyon/spin/actions/workflows/release.yml


itowlson (approved):

vdice (commented):
Will we want to preserve formulas for previous Spin releases?  eg when 1.5 comes out, we'd move the current `spin.rb` to `spin.rb@1.4.1` and then update `spin.rb` with 1.5 values.  Or, are we thinking just maintaining and updating the one formula?  I assume with the former, users could `brew install fermyon/tap/spin@1.4.1` whereas with the latter, no such version-pinning would be supported and users would always get the latest - is that correct?

kate-goldenring:
@vdice From my understanding, it is common practice to only support one Homebrew formula, namely the latest version of the project. If you want to support earlier versions ("pinned version formulas"), they become separate formula (like you mentioned) and you cannot simply brew upgrade to get the latest. For example, notice how Nomad's install instructions for macos only provide homebrew as an option for latest release; otherwise, you have to do a binary install: https://developer.hashicorp.com/nomad/docs/install.

in contrast, postgres releases each version as it's own package: https://www.postgresql.org/download/macosx/. Latest postgres is version 15; however if i install 14 and then brew upgrade, i stay on 14:
```sh
$ brew install postgresql@14
$ brew upgrade
$ brew list | grep postgresql
postgresql@14
$ brew install postgresql@15
$ brew list | grep postgresql
postgresql@14
postgresql@15
```
The only way to upgrade is to uninstall the old version and instal the new one (otherwise you get two installs) -- very unideal

My main concern here is that for some reason someone installs an older version and never moves up. But like you mention, we could take the hybrid approach and have latest always in `spin.rb` and then keep older versions as pinned. I'm happy with that. We should also have a retirement plan in that case -- aka we will only support the last X versions... but we probably arent there yet

docs/content/release-process.md
```diff
@@ -80,6 +80,11 @@ To cut a release of Spin, you will need to do the following:
    `--certificate-identity` value should match this release, e.g.
    `https://github.com/fermyon/spin/.github/workflows/release.yml@refs/tags/v1.1.0`.
 
+1. Update the Spin [Homebrew
```
vdice, line 83: Should we mention the specific mechanism envisioned for updating the formula?  eg via PR or...?

kate-goldenring:
@vdice I don't think it would be too hard to manage (just extra `.rb` files) -- i just feel there may be a reason that many other projects only support latest ... maybe to encourage that upgrade story? I do feel like we will hit this question actually in a week when we release 1.5

vdice (approved):
One non-blocking though, otherwise LGTM!

kate-goldenring:
@vdice issue to track on supporting other versions https://github.com/fermyon/homebrew-tap/issues/7

PR #1710, opened 2023-08-26
allow outbound http to components in same app by default

### LINKED ISSUE #957: Option to allow HTTP request to host running the current application
I am trying to build the following scenario: I have two components, and am trying to make requests from one to the other.

Because I don't know the actual URL the application will be available at before hand, every time the application is moved I need to change the `allowed_http_hosts` configuration if I want to continue to be able to send request to another component of the same application.

There are two options I see:

- always allow requests to component of the same application
- similar to the `insecure:allow-all` option, add a `self` option

### LINKED ISSUE #1533: outbound-http: Support "relative" URIs to current app
This has been requested a couple of times and should be _relatively_ straightforward:

- Add a field to `OutboundHttp` e.g. `origin: Uri` which gets updated from each inbound request
- If an outbound request starts with a `/` (has no scheme or authority), join with the origin

michelleN:
+ add ability for one component to make outbound calls
    to another component in the same Spin app using special value `self` in `allowed_http_hosts`
+ make requests with relative routes in app once self is set
+ update example for rust outbound http to demonstrate capability
+ adds e2e
+ resolves #957 
+ resolves #1533


crates/outbound-http/src/lib.rs
```diff
@@ -31,8 +33,19 @@ impl OutboundHttp {
...
+
+        if self.is_same_origin(&url) {
+            return Ok(true);
+        }
+
         Ok(self.allowed_hosts.allow(&url))
     }
+
+    /// Check if the outbound request host and port are the same as origin
+    fn is_same_origin(&self, url: &Url) -> bool {
```
itowlson, line 45: Maybe 
```suggestion
    fn is_same_as_origin(&self, url: &Url) -> bool {
```

?  "Same origin" implies to me whether two things have the same origin, whereas you are interested in whether the _URL_ (the _destination_) is the same as the origin.

crates/outbound-http/src/lib.rs
```diff
@@ -31,8 +33,19 @@ impl OutboundHttp {
     /// If `None` is passed, the guest module is not allowed to send the request.
     fn is_allowed(&self, url: &str) -> Result<bool, HttpError> {
         let url = Url::parse(url).map_err(|_| HttpError::InvalidUrl)?;
+
+        if self.is_same_origin(&url) {
```
itowlson, line 37: There is a possible sneaky gap here, if two applications are hosted on the same host but at different base URLs. I don't think that's possible with any current Spin implementation, so not a blocker (mentioning it mostly because this was the justification for have `base` in the first place).

crates/outbound-http/src/lib.rs
```diff
@@ -21,6 +21,8 @@ pub const ALLOWED_HTTP_HOSTS_KEY: MetadataKey<Vec<String>> = MetadataKey::new("a
 pub struct OutboundHttp {
     /// List of hosts guest modules are allowed to make requests to.
     pub allowed_hosts: AllowedHttpHosts,
+    /// origin is the host of the component that created this outbound HTTP request.
```
itowlson, line 24: This struct isn't associated with a specific HTTP request, so we could make this more explicit:

```suggestion
    /// During an incoming HTTP request, origin is set to the host of that incoming HTTP request.
    /// This is used to automatically allow outbound requests to the same host.
```

itowlson (changes requested):
This is great - thanks for making it happen!

My comments on the functional side of it are, I think, entirely suggestions or nits, not blockers.  I did, though, get a bit lost in the sample - could we change / rationalise some of the naming there please?  Thanks!

itowlson:
@karthik2804 noted that Cloud might need a bit of work to support this behaviour.  It would be good if that could happen before this goes into a Spin release, so as to avoid any "works locally but not in cloud" issues.

karthik2804:
This feature looks great! I have a question about how this works. From looking at the PR, component-to-component HTTP requests for the same app are enabled by default. This is a deviation from what Spin usually does as in that to allow something, some explicit configuration is needed. I am wondering if that is the desired behavior we would want?

From the linked issue I see there is a proposed option for a `self` configuration option that would allow for an explicit configuration (ref: https://github.com/fermyon/spin/issues/957#issuecomment-1688944798).

I wanted to bring this up and see what people thought.

crates/outbound-http/src/lib.rs
michelleN: that's really great wording thank you

examples/http-rust-outbound-http/outbound-http-to-same-app/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+            .uri(outbound_uri.as_str())
+            .body(None)?,
+    )?;
+    res.headers_mut()
+        .insert("spin-component", "rust-outbound-http".try_into()?);
+    println!("{:?}", res);
+    Ok(res)
+}
+
+fn base_url(req: Request) -> Result<Url> {
```
mikkelhegn, line 27: Is it an option to provide a function like this in the SDKs, so users don't have to implement this in every component?

itowlson (approved):

examples/http-rust-outbound-http/outbound-http-to-same-app/src/lib.rs
lann: Alternatively we could reserve some fake host name to represent "self", e.g. `https://self.local`.

crates/outbound-http/src/lib.rs
```diff
@@ -63,7 +53,15 @@ impl outbound_http::Host for OutboundHttp {
             }
 
             let method = method_from(req.method);
-            let url = Url::parse(&req.uri).map_err(|_| HttpError::InvalidUrl)?;
+            let mut url = Url::parse(&req.uri).map_err(|_| HttpError::InvalidUrl)?;
+
+            if url.host_str() == Some("self.local") {
+                url.set_host(self.origin.host())
```
lann, line 59: Should we update the scheme here too? Thinking about testing on `http://localhost` and then switching to `https://my.fermyon.app`.

michelleN: yea right now you'd have to change the scheme in the request from the app and that seems like a papercut.

michelleN: Updated the implementation so you don't need to use `self.local` in the uri when making a request and instead you can use relative paths and the correct origin info will be added to the uri as you originally suggested in #1533

examples/http-rust-outbound-http/outbound-http-to-same-app/src/lib.rs
mikkelhegn: `https://my.ego` :-) Demo time "So here I call my ego" 👍🏻

crates/outbound-http/src/lib.rs
michelleN: `self.local` is now `self` because... one word is better than two!

lann: We may still want to _allow_ `http://self.local/abc` (in addition to just `/abc`) for code that expects an absolute URL, like (I think?) the Go SDK.

michelleN: I'm not seeing where the Go SDK requires an absolute URL. Could you point me in right direction?

lann: I was thinking that since we replace the default http client ([here](https://github.com/fermyon/spin/blob/main/sdk/go/http/http.go#L36)) it might break something in that path if you pass a relative URL, but I haven't tested.

examples/http-tinygo-outbound-http/outbound-http-to-same-app/main.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	spinhttp "github.com/fermyon/spin/sdk/go/http"
+)
+
+func init() {
+	spinhttp.Handle(func(w http.ResponseWriter, r *http.Request) {
+		resp, err := spinhttp.Get("/hello")
```
lann, line 12: If convenient, could you try this as just `http.Get` ([like here](https://github.com/fermyon/spin/blob/main/sdk/go/integration_test.go#L26))?

michelleN: that worked as well

crates/outbound-http/src/lib.rs
```diff
@@ -21,15 +21,27 @@ pub const ALLOWED_HTTP_HOSTS_KEY: MetadataKey<Vec<String>> = MetadataKey::new("a
...
     /// Check if guest module is allowed to send request to URL, based on the list of
-    /// allowed hosts defined by the runtime. If the list of allowed hosts contains
+    /// allowed hosts defined by the runtime. If the url passed in is a relative path,
+    /// only allow if allowed_hosts contains `self`. If the list of allowed hosts contains
     /// `insecure:allow-all`, then all hosts are allowed.
     /// If `None` is passed, the guest module is not allowed to send the request.
-    fn is_allowed(&self, url: &str) -> Result<bool, HttpError> {
+    fn is_allowed(&mut self, url: &str) -> Result<bool, HttpError> {
+        if url.starts_with('/') {
+            if self.allowed_hosts.includes(AllowedHttpHost::host("self")) {
```
itowlson, line 38: Rather than requiring callers to know how AllowedHttpHosts represents the self-link permission (and requiring them to construct an AllowedHttpHost just in order to ask), consider putting this into a `AllowedHttpHosts::allow_relative_url` (or some better name) method.

```suggestion
            return self.allowed_hosts.allow_relative_url();
```

(and remove the `includes` method)

crates/outbound-http/src/lib.rs
```diff
@@ -49,7 +61,14 @@ impl outbound_http::Host for OutboundHttp {
             }
 
             let method = method_from(req.method);
-            let url = Url::parse(&req.uri).map_err(|_| HttpError::InvalidUrl)?;
+
+            let req_url: Url = if req.uri.starts_with('/') {
```
itowlson, line 65: To avoid having the parse and error map on both branches, I wonder if this could be reorganised as something like

```suggestion
            let abs_url = if req.uri.starts_with('/') {
                format!(...)
            } else {
                req.uri.to_string()
            };
            let req_url = Url::parse(&abs_uri).map_err(...)?;
```

crates/trigger-http/src/spin.rs
```diff
@@ -27,11 +27,27 @@ impl HttpExecutor for SpinHttpExecutor {
...
         let EitherInstance::Component(instance) = instance else {
             unreachable!()
         };
 
+        if let Some(authority) = req.uri().authority() {
+            if let Some(scheme) = req.uri().scheme_str() {
+                if let Some(outbound_http_handle) = engine
+                .engine
+                .find_host_component_handle::<std::sync::Arc<outbound_http::OutboundHttpComponent>>(
+                )
```
itowlson, line 40: Oh rustfmt no

(no action for you, I am just sad at the tool)

crates/trigger-http/src/spin.rs
```diff
@@ -27,11 +27,27 @@ impl HttpExecutor for SpinHttpExecutor {
...
+            {
+                let outbound_http_data = store
+                    .host_components_data()
+                    .get_or_insert(outbound_http_handle);
+
+                outbound_http_data.origin = format!("{}://{}", scheme, authority);
+            }
+            }
```
itowlson, line 48: Two consecutive close braces at the same indent feels weird to me but I suppose this is another rustfmt special...

crates/trigger-http/src/spin.rs
```diff
@@ -27,11 +27,27 @@ impl HttpExecutor for SpinHttpExecutor {
             component_id
         );
 
-        let (instance, store) = engine.prepare_instance(component_id).await?;
+        let (instance, mut store) = engine.prepare_instance(component_id).await?;
         let EitherInstance::Component(instance) = instance else {
             unreachable!()
         };
 
+        if let Some(authority) = req.uri().authority() {
```
itowlson, line 35: This has grown to a big enough block that it rather breaks up the flow of the function - consider pulling it out into a `set_http_origin_from_request(&mut store, &req)` (or otherwise meaningfully named) function.

examples/http-tinygo-outbound-http/tinygo-hello/go.sum
```diff
@@ -0,0 +1,2 @@
+github.com/julienschmidt/httprouter v1.3.0 h1:U0609e9tgbseu3rBINet9P48AI/D3oJs4dN7jwJOQ1U=
```
itowlson, line 1: Is it correct that we are adding these in this PR?

itowlson (approved):
Looks fine - I left some style suggestions but nothing blocking.  Thanks!

crates/trigger-http/src/spin.rs
lann: I don't think `rustfmt` produced this, I think it gave up on this code for some reason...

lann: Looks like it gave up because the function call was too long. If you add `use`s for `Arc` and `OutboundHttpComponent` it will happily reformat this.

examples/http-tinygo-outbound-http/tinygo-hello/go.sum
michelleN: good catch. removed!

crates/trigger-http/src/spin.rs
michelleN: good tip! Moving it to its own function helped but I added the uses anyway for readability.

examples/http-tinygo-outbound-http/tinygo-hello/go.sum
michelleN: turns our I did need that go.sum file so I added this back in.

PR #1711, opened 2023-08-28
When `spin watch` terminates `spin up`, allow it to delete its temp dir

### LINKED ISSUE #1698: What is the recommended solution to clear spin-related cache files in the /tmp directory?
Loving spin otherwise btw!

I've been using spin locally via `spin watch` during development of an app using 3 components (the spin-fileserver and 2 custom ones) on a Debian Linux system for weeks. Noticed today my `/tmp` partition was nearly out of space. Seeing thousands of directories called `.tmpRANDOM` (where `RANDOM` is a random 6-digit alphanumeric string) with `spin.lock` (so I know they're related to spin) in `/tmp` totaling to hundreds of GB!

Not observing any subcommand in the `spin` CLI addressing cleaning up cache, etc...

Presuming I can kill `spin watch` and any other instances of spin, and delete all of these via `fd '^spin.lock$' /tmp |sed 's/\/spin\.lock$//' |xargs rm -rf`?

Is there any other solution you would recommend?

Are there any methods to prevent this behavior... such as... spin automatically deleting cache files in /tmp when it exits?

Thanks!

---

Spin version (`spin --version`): `spin 1.4.1 (e0bd911 2023-07-12)`

Installed plugin versions (`spin plugins list --installed`)

```
cloud 0.1.2 [installed]
js2wasm 0.5.1 [installed]
py2wasm 0.3.0 [installed]
```

itowlson:
Fixes #1698.

The heart of this fix simply changes Unix builds to send SIGINT (Ctrl+C) to stop `spin up` instead of just nuking it.  This allows `spin up` to exit gracefully, cleaning up its turds.  And all is grand, except, wee small problem, this by itself results in builds getting nuked before they can complete, so the user no longer sees their changes, and hey, if I want that, I can just use `spin up` amirite?

The reason for this is that the `Building` and `Running` states get out of sync with the actual process.  On a change, the state machine moves into `Building` (and starts the build), but `spin up` is still cleaning up.  When `up` exits, all watch see is "the child process completed", and it takes that to be the current child process i.e. `build`.  Great, so we can nuke `build` and start `up`.  Uh oh.

There are probably various ways of solving this, and I am a bit wary of the one this PR goes for, but that's why we have review cycles.  What it does is add a new state to watch's internal state machine, "waiting for `spin up` to exit".  When something changes while we are in Running, we go into waiting first, and when we get a ChildProcessCompleted in waiting, _then_ we go into building and do a start.

Waiting is not a great state because it doesn't map to a Spin command, and watch really wants states to map to commands.  And if means all the effect clauses now need to think about waiting as well as building and running, and I am not sure I have them doing the right thing...

...BUT...

...it _seems_ to work, and I _think_ the transitions make sense.

Longer term, this exposes possible pain points with `watchexec`.  It seems very much designed around "there is a command and we will restart that command."  We do some magic to intercept those restarts and go "not so fast, I think you will find that the command is actually `spin up` today", but things like the ChildProcessCompleted event are clearly designed for an environment where you are not juggling multiple commands and processes (like CPC doesn't provide _any_ info about the process, because you're meant to know what the process is because there's only one of them).

@calebschoepp this would really benefit from your deep knowledge of spin watch and watchexec, especially in terms of whether the transitions and the tests for those transitions make sense (the tests have a fair bit of "just change the test to match the error" I'm afraid).  I have not yet really tested with "skip build" or "clear" turned on so there might be some glitches there and I am not sure I will know how to fix them if there are.  And of course I'm very open to better ways of managing the graceful exit, this is just something that kinda made sense to me and seemed to work...!  Thanks!

### DIFF
--- a/src/watch_state.rs
+++ b/src/watch_state.rs
@@ -42,33 +42,50 @@ impl WatchState {
         // https://docs.rs/watchexec/latest/watchexec/action/enum.Outcome.html
         let outcome = match (effect, *state) {
             (Effect::Exit, _) => {
-                Outcome::if_running(Outcome::both(Outcome::Stop, Outcome::Exit), Outcome::Exit)
+                Outcome::if_running(Outcome::both(stop_outcome(), Outcome::Exit), Outcome::Exit)
             }
             (Effect::ChildProcessFailed, _) => {
-                Outcome::if_running(Outcome::Stop, Outcome::DoNothing)
+                Outcome::if_running(stop_outcome(), Outcome::DoNothing)
             }
             (Effect::ChildProcessCompleted, State::Building) => {
                 *state = State::Running;
                 self.restart_outcome(PREVENT_CLEAR)
             }
-            (Effect::ChildProcessCompleted, State::Running) => Outcome::DoNothing,
+            (Effect::ChildProcessCompleted, State::Running) => {
+                *state = State::Building;
+                self.restart_outcome(PREVENT_CLEAR)
+            }
+            (Effect::ChildProcessCompleted, State::WaitingForSpinUpToExit) => {
+                *state = State::Building;
+                self.restart_outcome(PREVENT_CLEAR)
+            }
             (Effect::ManifestChange, State::Building) => self.restart_outcome(ALLOW_CLEAR),
+            (Effect::ManifestChange, State::WaitingForSpinUpToExit) => Outcome::DoNothing,
             (Effect::ManifestChange, State::Running) => {
                 if !self.skip_build {
-                    *state = State::Building;
+                    *state = State::WaitingForSpinUpToExit;
+                    Outcome::if_running(stop_outcome(), Outcome::DoNothing)
+                } else {
+                    self.restart_outcome(ALLOW_CLEAR)
                 }
-                self.restart_outcome(ALLOW_CLEAR)
             }
             (Effect::SourceChange, State::Building) => self.restart_outcome(ALLOW_CLEAR),
+            (Effect::SourceChange, State::WaitingForSpinUpToExit) => Outcome::DoNothing,
             (Effect::SourceChange, State::Running) => {
                 if !self.skip_build {
-                    *state = State::Building;
-                    return self.restart_outcome(ALLOW_CLEAR);
+                    *state = State::WaitingForSpinUpToExit;
+                    Outcome::if_running(stop_outcome(), Outcome::DoNothing)
+                } else {
+                    Outcome::DoNothing
                 }
-                Outcome::DoNothing
             }
             (Effect::ArtifactChange, State::Building) => Outcome::DoNothing,
-            (Effect::ArtifactChange, State::Running) => self.restart_outcome(ALLOW_CLEAR),
+            (Effect::ArtifactChange, State::WaitingForSpinUpToExit) => Outcome::DoNothing,
+            (Effect::ArtifactChange, State::Running) => {
+                *state = State::WaitingForSpinUpToExit;
+                Outcome::if_running(stop_outcome(), Outcome::DoNothing)
+            }
+            (Effect::DoNothing, State::WaitingForSpinUpToExit) => Outcome::DoNothing,
             (Effect::DoNothing, _) => Outcome::if_running(Outcome::DoNothing, Outcome::Start),
         };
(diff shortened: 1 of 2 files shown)


itowlson:
I messed up `--clear`.  If I pass `--clear` it terminates `spin up` but never kicks off the restart.  Investigating...

itowlson:
Oh good adding instrumentation makes the bug go away.  So maybe there's still some sort of race condition that's dependent on timing...

src/watch_state.rs
```diff
@@ -42,33 +42,48 @@ impl WatchState {
...
+                    return Outcome::if_running(stop_outcome(), Outcome::DoNothing);
                 }
                 Outcome::DoNothing
             }
             (Effect::ArtifactChange, State::Building) => Outcome::DoNothing,
-            (Effect::ArtifactChange, State::Running) => self.restart_outcome(ALLOW_CLEAR),
+            (Effect::ArtifactChange, State::WaitingForSpinUpToExit) => Outcome::DoNothing,
+            (Effect::ArtifactChange, State::Running) => {
+                *state = State::WaitingForSpinUpToExit;
+                Outcome::if_running(stop_outcome(), Outcome::DoNothing)
```
calebschoepp, line 86: We produce this outcome in three locations. I wonder if it would read a little clearer if we created it with a function.

src/watch_state.rs
```diff
@@ -42,33 +42,50 @@ impl WatchState {
...
                 if !self.skip_build {
-                    *state = State::Building;
+                    *state = State::WaitingForSpinUpToExit;
+                    Outcome::if_running(stop_outcome(), Outcome::DoNothing)
+                } else {
+                    self.restart_outcome(ALLOW_CLEAR)
                 }
-                self.restart_outcome(ALLOW_CLEAR)
             }
             (Effect::SourceChange, State::Building) => self.restart_outcome(ALLOW_CLEAR),
```
calebschoepp, line 72: Does this not cause a race condition. Since we now use SIGINT the build will take awhile to finish and so a `Effect::ChildProcessCompleted` will come while the build is running and that will prematurely kill the build and jump to the running step?

I've been staring at this a long time and struggling to think it through so I could very well be wrong.

calebschoepp (commented):
Nothing immediatly obvious is jumping out to me as a better way to do this. I know watchexec supports multiple commands that you run in a sequence ([docs](https://docs.rs/watchexec/latest/watchexec/command/struct.Supervisor.html)) so maybe we could use/abuse that primitive to handle the sequencing? Although I can't remember, I'd have to imagine there was a reason I chose to not implement it that way though.

If we do reach a point where Watchexec fundamentally doesn't allow us to do what we need I would advocate for forking it to add the functionality we need or upstreaming the changes we would need.

src/watch_state.rs
itowlson: Yeah I am going to write out all the start scenarios "inline" and think through whether they are correct.

itowlson:
It is sometimes missing the ChildProcessCompleted for `spin up` which is preventing it moving to the next operation.  I don't understand why; the child process should still be being monitored at this point (and sometimes is... just sometimes not).

I'm now wondering if `Outcome::Wait` would help instead of extra states.  The docs imply it would, but when I tried it didn't.  But maybe I was using it wrong.

lann:
Have you tried replicating [what the `watchexec` cli does](https://github.com/watchexec/watchexec/blob/96bf3d231e6312beda9a589b52e8b64138514293/crates/cli/src/config/runtime.rs#L174-L177)?

itowlson:
Yeah, I thought about an "estimate how long `spin up` should take to tidy up then force-kill".  But I am not sure it will help with watchexec getting confused about process completions.  Some of the difficulty is the interaction between watchexec's design around "run or restart a process when the thing happens" vs our "switch between two processes when the thing happens."  But it can't hurt to try.

Another possibility I'm exploring is having two separate watchers - one to run `spin build` and one to run `spin up`, suspending the `spin up` watcher whenever the build watcher runs.  This should make each watcher much simpler because better aligned because what watchexec likes to do, and I think the synchronisation could be simpler too.  It's a bigger change though.

itowlson:
Parking this one in favour of #1723 - if that runs into objections I will re-open this

PR #1714, opened 2023-08-28
feat(sdk/go): implement sqlite SDK for TinyGo

adamreese:
Implements Golang sql database interface for Spin's sqlite.

### DIFF
--- a/sdk/go/sqlite/sqlite.go
+++ b/sdk/go/sqlite/sqlite.go
@@ -1,0 +1,133 @@
+// Package sqlite provides access to database stores within Spin
+// components.
+package sqlite
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"errors"
+	"io"
+)
+
+// Open returns a new connection to the database.
+func Open(name string) *sql.DB {
+	return sql.OpenDB(&connector{name: name})
+}
+
+// Conn represents a database connection.
+type Conn struct {
+	_ptr uint32
+}
+
+// Close the connection.
+func (db *Conn) Close() error {
+	db.close()
+	return nil
+}
+
+// Query executes a query that may return rows, such as a SELECT.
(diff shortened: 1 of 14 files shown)


sdk/go/sqlite/sqlite.go
```diff
@@ -0,0 +1,133 @@
...
+
+// QueryContext executes a query that may return rows, such as a SELECT.
+func (c *Conn) QueryContext(_ context.Context, query string, args []driver.Value) (driver.Rows, error) {
+	return c.Query(query, args)
+}
+
+// Exec isn't implemented. Use Query method.
+func (c *Conn) Exec(_ context.Context, _ string, _ []driver.NamedValue) (driver.Result, error) {
+	return nil, errors.New("Exec method not implemented")
+}
```
rylev, line 46: I there a reason we're not implementing this?

adamreese: Correct me if I'm wrong... the returned object should be `LastInsertId` and `RowsAffected` which I don't think is returned from our implementation. https://pkg.go.dev/database/sql/driver#Result

adamreese:
Pushed up some changes to support `Prepare()`

sdk/go/sqlite/sqlite.go
rylev: Ah you're correct. Sorry for the noise!

rylev (approved):
This looks good to me - I think we should get this in and have others write some code against it.

sdk/go/sqlite/sqlite.go
adamreese: I missed the note in the doc that says `Not every database or database driver may support this.` So I'll add it in

adamreese:
@rylev I added support for the `Exec` methods and removed some unneeded Context methods.  Can I get a fresh LGTM?

PR #1721, opened 2023-08-30
Enable Wasmtime's pooling allocator by default

alexcrichton:
This PR continues the work from https://github.com/fermyon/spin/pull/998 (thanks @tschneidereit!) to enable the pooling allocator by default for Spin. The pooling allocator is tailor-made for Spin's use case of many concurrent short-lived instances and I've updated the benchmark in this repository to showcase it.

The new benchmark measures the latency of a single request when there are N tasks infinitely handling requests in the background. The numbers before this PR are:

|           | 1 concurrency | 80 concurrency | 320 concurrency |
|-----------|---------------|----------------|-----------------|
| noop      | 111us         | 12.0ms         | 13.5ms          |
| sleep=1   | 1.33ms        | 15.8ms         | 15.7ms          |
| cpu=1     | 2.46ms        | 14.3ms         | 15.9ms          |

and the numbers after this PR are

|           | 1 concurrency | 80 concurrency | 320 concurrency |
|-----------|---------------|----------------|-----------------|
| noop      | 50.2us        | 570us          | 546us           |
| sleep=1   | 1.26ms        | 7.7ms          | 8.08ms          |
| cpu=1     | 2.39ms        | 2.42ms         | 2.7ms           |

The `noop` benchmark performs no work and simply returns success from the HTTP handler. Here it can be seen that as the concurrency increases the latency drastically increases. This is primarily due to kernel contention on locks for the address space, page tables, etc. With the pooling allocator it's still not scaling well (80 concurrency is 10x slower than 1 concurrency), but it's performing much better than the on-demand allocator. One reason for this is that async stacks are pooled so require zero syscalls to manage. Another reason is that the memory image of the module in question gets to stay in place in memory meaning there are fewer syscalls. The remaining syscalls, however, are why the 10x slowdown happens as there's still lots of kernel contention in page faults.

The `sleep=1` benchmark is an HTTP handler that issues a WASI sleep for 1ms. This is shows similar scaling behavior as `noop` but also probably isn't a great benchmark at this time as the current preview2 implementation of `poll_oneoff` actually uses the `poll` syscall so this benchmark isn't properly cooperative.

The `cpu=1` benchmark is an HTTP request that performs a fixed amount of CPU-bound work during its execution. This showcases how CPU-heavy work which previously did not scale well now scales much better. The reason for this is that once running there's very few syscalls so CPU is itself the starved resource as opposed to locks.

I'll also note that these benchmarks are basically just "yet another" way to view the results of the pooling allocator. @tschneidereit's original benchmarks in https://github.com/fermyon/spin/pull/998 I believe are additionally representative of the performance of the pooling allocator.

Relative to https://github.com/fermyon/spin/pull/998 I've merged in the latest changes and additionally switched the defaults to have the pooling allocator on-by-default. I have additionally removed parameters for the pooling allocator from configuration in the hopes that exposing them isn't necessary in the long run. Environment variable escape hatches are provided but otherwise my hope is that they won't need to be used.


crates/core/src/lib.rs
```diff
@@ -53,6 +58,28 @@ impl Config {
...
         &mut self.inner
     }
+
+    /// Enable the Wasmtime compilation cache. If `path` is given it will override
+    /// the system default path.
+    ///
+    /// For more information, see the [Wasmtime cache config documentation][docs].
+    ///
+    /// [docs]: https://docs.wasmtime.dev/cli-cache.html
+    pub fn configure_cache(&mut self, config_path: &Option<PathBuf>) -> Result<()> {
```
lann, line 68: Maybe `enable_cache`? `configure_cache(None)` seems like it should disable cache.

lann (approved):
Non-blocking method name suggestion; otherwise lgtm

alexcrichton:
Looks like the CI failure is a switft-generated module which has 10k+ table elements. I'm going to increase the current limit from 10k to 30k, but that limit in particular is one we should keep an eye on to increase. Increasing it causes a larger virtual address space reservation to happen for tables but compared to memories it's still paltry at 30k (e.g. a memory is 6G virtual address space, and a 30k element table is 240k of address space.

PR #1724, opened 2023-08-31
add make install command

michelleN:
+ for cargo install --path . --locked
+ helps me speed up local dev experience

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,10 @@ endif
 build:
 	cargo build --release
 
+.PHONY: install
+install:
+	cargo install --path . --locked
+
 .PHONY: test
 test: lint test-unit test-integration
 


Makefile
```diff
@@ -27,6 +27,10 @@ endif
 build:
 	cargo build --release
 
+.PHONY: install
+install:
+	cargo install --path .
```
itowlson, line 32: Forgive the persnicketing, but `make build` does a release build, while this does a debug build - should they be consistent?

itowlson, line 32: ooh might be worth specifying `--locked` as well, otherwise it could pick up random versions of dependencies

fibonacci1729: I believe `cargo install` uses the release profile by default (opposite to `build` which uses `dev`, i.e. `--debug`)

fibonacci1729 (approved):

Makefile
itowlson: Doh, you're right @fibonacci1729.  Apologies!

michelleN: done! thanks

PR #1725, opened 2023-08-31
Add more conversions to sqlite SDK ValueResult

rylev:
Fixes an issue reported by @mikkelhegn where we didn't have a conversion between `ValueResult` and `&[u8]`. This works for both `ValueResult::Blob` and `ValueResult::Text`. 

This adds such a conversion along with conversions for all basic integer types (signed and unsigned) and `f64`. It also allow casting from `ValueResult::Blob` to `&str` which will fail if the blob is not utf8. 

This is technically a breaking change as previously the conversion from `ValueResult` to `u32` would overflow but now it returns an error: [see here](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=b29303f6f6e65cee8299fe794eadc687). I personally think this is an acceptable breaking change, but I'd like to hear others' thoughts.

### DIFF
--- a/sdk/rust/src/sqlite.rs
+++ b/sdk/rust/src/sqlite.rs
@@ -81,12 +81,29 @@ impl<'a> TryFrom<&'a ValueResult> for bool {
     }
 }
 
-impl<'a> TryFrom<&'a ValueResult> for u32 {
+macro_rules! int_conversions {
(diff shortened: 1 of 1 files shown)


itowlson (approved):
Even I can't get fretful about changing the integer overflow behaviour.  :shipit:

PR #1726, opened 2023-08-31
Fix clippy errors from 1.72

rylev:
This fixes clippy errors with the version of clippy shipped with 1.72

### DIFF
--- a/crates/app/src/metadata.rs
+++ b/crates/app/src/metadata.rs
@@ -23,10 +23,7 @@ impl<T> MetadataKey<T> {
 
 impl<T> Clone for MetadataKey<T> {
     fn clone(&self) -> Self {
-        Self {
-            key: self.key,
-            _phantom: PhantomData,
-        }
+        *self
     }
 }
 
--- a/crates/core/src/host_component.rs
+++ b/crates/core/src/host_component.rs
@@ -106,7 +106,7 @@ impl<HC: HostComponent> HostComponentDataHandle<HC> {
 
 impl<HC: HostComponent> Clone for HostComponentDataHandle<HC> {
     fn clone(&self) -> Self {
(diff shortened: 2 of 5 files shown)


crates/app/src/metadata.rs
```diff
@@ -23,10 +23,7 @@ impl<T> MetadataKey<T> {
 
 impl<T> Clone for MetadataKey<T> {
     fn clone(&self) -> Self {
-        Self {
-            key: self.key,
-            _phantom: PhantomData,
-        }
+        *self
```
itowlson, line 26: Hmm, I don't much like this lint.  I see why it's correct but it's less clear what it's doing and why - if `key` ever changed to a non-`Copy` type, the maintainer would need to reconstruct the clone logic.  I suppose that is too trivial to worry about though.

My preference would be to exclude this lint and keep the code explicit.  But if other folks prefer the elided version then it's no biggie..

crates/core/src/host_component.rs
```diff
@@ -106,7 +106,7 @@ impl<HC: HostComponent> HostComponentDataHandle<HC> {
 
 impl<HC: HostComponent> Clone for HostComponentDataHandle<HC> {
     fn clone(&self) -> Self {
-        Self::from_any(self.inner)
+        *self
```
itowlson, line 109: Same comment - prefer to exclude the lint but not a blocker

itowlson (approved):
There's one of the lints where I'd prefer to exclude it keep the affected code as is, but it's not a blocker.

rylev:
I don't really mind either version - the newer version is arguably a bit clearer to me, but I can see why the more verbose might be preferable. Honestly, given the toss up, I'd prefer to just merge this as is and not have to wait for another CI run.

PR #1731, opened 2023-09-04
Large Language Model (LLM) Support

rylev:
This adds unstable support for a new Large Language Model "llm" interface to Spin that allows users to performance inferencing requests and generate embeddings. 

The user can get this support by first opting into which models they want a component to be able to use in their `spin.toml` manifest:

```toml
ai_models = ["llama2-chat","all-minilm-l6-v2"]
```

> Note: We will support "llama2-chat" and "codellama-instruct"  for inferencing and "all-minilm-l6-v2" for generating embeddings.

They then must supply the model files themselves in a well known location: `.spin/ai_models/$NAME_OF_MODEL`. 

> Note: embeddings models are expected to be directories containing a tokenizer.json and model.safetensors file while inferencing is just the model file. 

In their apps, they can use the `llm` interface through the Spin SDK:

```rust
let inference = llm::infer(llm::InferencingModel::Llama2Chat, "Tell me a story about Slats the cat.".into())?;
```

You can read more about this feature in SIP: [[here]](https://github.com/fermyon/spin/blob/llm-support/docs/content/sips/015-llm.md).


radu-matei (approved):
This needs a rebase and to get CI to be happy, then LGTM!

Issue #1733, opened 2023-09-05
Templates on spin for rust need to point at main for their spin-sdk dependency


macolso:
Reported on Discord [here](https://discord.com/channels/926888690310053918/1148522244277014568/1148654246787555378) by @ThorstenHans

ThorstenHans:
I looked up the source code for the install script at [https://github.com/fermyon/developer/blob/main/downloads/install.sh](https://github.com/fermyon/developer/blob/main/downloads/install.sh). It executes `spin templates install --git url --upgrade` in [line 130](https://github.com/fermyon/developer/blob/7f99d957c7d3498b75ca48da97991dbcd88df908/downloads/install.sh#L130)

So my first naive attempt was to execute 

```bash
spin templates intall --git "https://github.com/fermyon/spin" --upgrade --branch main
```

Which will force cloning the main branch as implemented in `crates/templates/src/source.rs#L141-L147`

https://github.com/fermyon/spin/blob/089e135f7fc4e76bf67561c079cc6bfd184a9b4c/crates/templates/src/source.rs#L141-L147

So at this point, users will end up with templates from the `main` branch. Unfortunately, those pin the `spin-sdk` also to `tag = "v1.4.2"`

https://github.com/fermyon/spin/blob/089e135f7fc4e76bf67561c079cc6bfd184a9b4c/templates/http-rust/content/Cargo.toml#L19

After that I tried digging deeper to find the proper branch 😄. So I stumbled upon `llm-sdk`, however `http-rust` template of this branch pins `spin-sdk` to `tag = "v1.4.1`:

https://github.com/fermyon/spin/blob/ca520925df9dd57524041a0cecbed704341253a9/templates/http-rust/content/Cargo.toml#L19

mikkelhegn:
This should be
```
spin-sdk = { git = "https://github.com/fermyon/spin", tag = "canary" }
```

kate-goldenring:
Our templates system does not currently support pointing to canary IIRC so this is not a simple install script update (unless we want to do some wild sed parsing). We could add a note to the script when the canary flag is passed that templates will need to be updated to use the canary tag.

itowlson:
I don't understand the difference between `canary` and `main`.  If you install canary Spin, you _get_ `main`, unless you hit it in the brief window between a commit landing in main and the resulting canary being built.

[...]

PR #1734, opened 2023-09-05
point templates at canary tag

karthik2804:
Update templates to point at the canary tag to allow for installing templates for `llm`.

### DIFF
--- a/templates/http-rust/content/Cargo.toml
+++ b/templates/http-rust/content/Cargo.toml
@@ -16,6 +16,6 @@ bytes = "1"
 # General-purpose crate with common HTTP types.
 http = "0.2"
 # The Spin SDK.
-spin-sdk = { git = "https://github.com/fermyon/spin", tag = "v1.4.2" }
+spin-sdk = { git = "https://github.com/fermyon/spin", tag = "canary" }
 
 [workspace]


kate-goldenring:
Won't this make all installs of the templates point to canary? (even if you want to target main?)

karthik2804:
Only Canary Spin installs templates from `main`

kate-goldenring:
@karthik2804 i am not sure i see where that happens -- do you mind pointing it out? Are we checking the semver of Spin in the templates command to see if it is beyond a release somewhere?

karthik2804:
@kate-goldenring . When installing templates, spin looks for a tag `spin/templates/<major>.<minor>` first, if available - it installs that else it picks up templates from `main`.

kate-goldenring:
Thanks @karthik2804. So sounds like we want to add to our release checklist to point templates to canary tags. We'd want to do the same for the go template too. Is there a way to do this toggling smoothly?

itowlson:
The canary/main distinction confuses the heck out of me here.  Okay, so I grab canary Spin.  Then I try to install templates, which finds no corresponding version tag, so gets the templates from main.  But the templates on main point to canary, which is who knows what - maybe the same commit as the templates, maybe not - and is in no way guaranteed to have any relation to the canary of Spin that I have because time has elapsed.

I'd prefer to just refer to main (with "canary" being the latest binary built from main, not a separate tag).  Or if we want to have a way to stamp a consistent set, then tag that set, or fetch templates via a commit SHA instead of main (Spin knows the commit SHA it was built from), and have the the templates refer to SDKs at the corresponding SHA.

itowlson:
Also note if we do this then we need to fix the template makefile before the next minor release as per https://github.com/fermyon/spin/issues/1733#issuecomment-1716754811

vdice:
@karthik2804 I think we'll want to nail down a general approach in https://github.com/fermyon/spin/issues/1733 as opposed to the changes in this PR.  Shall we close?

karthik2804:
@vdice I agree - we can close this

PR #1741, opened 2023-09-06
Try building aarch64 linux canary build

rylev:
This tries turning back on the aarch64 linux build.

It does this by passing the `+fp16` CPU feature to rustc to build with. This CPU feature is needed for simd instructions carried out by candle. I don't think this will present problems for users, but if we ever hear about anyone who wants to run Spin on a Linux aarch64 machine without that CPU feature, we can decide what to do then.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,15 +34,14 @@ jobs:
               target: "",
               targetDir: "target/release",
             }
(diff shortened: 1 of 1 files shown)


vdice:
Adding this to the 1.5.0 milestone (anticipated for this week) in the hopes that we'll have an aarch64 linux binary per usual, though with the caveat mentioned above.

radu-matei (approved):

PR #1746, opened 2023-09-07
fix(oci): Respect trigger base path in locked app

adamreese:
Adds trigger base path to locked apps when deploying using oci


adamreese:
Moving this change to LHC because Spin expects the route to be without the base prepended.

Issue #1116, opened 2023-02-07
prefix log lines with component ID


michelleN:
It's hard to tell which log lines come from the Spin runtime vs from the actual application (a component). It'd be nice to prefix log lines with this information: `[Spin]` and `[Component-ID]`

For example, if we were to the run the [example http rust app](https://github.com/fermyon/spin/tree/main/examples/http-rust) which has a [single component called `hello`,](https://github.com/fermyon/spin/blob/main/examples/http-rust/spin.toml#L9) the logs look like this:
```
Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000 (wildcard)
{"host": "127.0.0.1:3000", "connection": "keep-alive", ... "spin-path-info": "/", "spin-full-url": "http://127.0.0.1:3000/", "spin-matched-route": "/...", "spin-base-path": "/", "spin-raw-component-route": "/...", "spin-component-route": ""}
```

With the component-id and spin prefixes, the logs would look like this.

```
$ spin up
[Spin] Serving http://127.0.0.1:3000
[Spin] Available Routes:
  hello: http://127.0.0.1:3000 (wildcard)
[hello] {"host": "127.0.0.1:3000", "connection": "keep-alive", ... "spin-path-info": "/", "spin-full-url": "http://127.0.0.1:3000/", "spin-matched-route": "/...", "spin-base-path": "/", "spin-raw-component-route": "/...", "spin-component-route": ""}
```

Would appreciate any feedback here.

itowlson:
Apologies for the pedantry but... the specific Spin output lines you quote are _not_ logs - they are written directly as actionable information for the user.  They should not receive logging-style decoration and their layout should be preserved (or at least modified only consciously).

Obviously this does not affect the broader issue - it relates purely to the example.

michelleN:
> the specific Spin output lines you quote are not logs - they are written directly as actionable information for the user. 

good point. I updated the description to be more clear.

itowlson:
I'm not in favour of decorating normal user interaction this way.  If we're going to decorate stuff, it should be the stuff you only use for diagnostics when something's gone wrong, not the stuff you see as part of normal operation and that is intended to be interacted with.

lann:
Request on Discord: https://discord.com/channels/926888690310053918/1081227946976616528/1081234846430535721

[...]

PR #1753, opened 2023-09-11
sip: Update SIP 005

lann:


docs/content/sips/005-manifest-redesign.md
```diff
@@ -71,8 +72,7 @@ id = "world"
 ### Manifest version
 
 Since this proposal represents a backward-incompatible change to the manifest format, the manifest
-version changes. The existing `spin_version = "1"` field name leaves some room for misinterpretation
-as "compatible with Spin project version 1". As an optional part of this proposal, change that name:
```
lann, line 75: We already renamed this key, separately.

michelleN (commented):
- Would it be possible to move trigger information into the component trigger section?

- I'm for removing base but could we call out why folks would want to use it? I'm still not super clear on why someone would want to use `base` over adding that to the path by hand with Spin.

lann:
I left out a bunch of changes that I meant to include, now added.

lann:
> I'm still not super clear on why someone would want to use base over adding that to the path by hand with Spin.

Me neither...

docs/content/sips/005-manifest-redesign.md
```diff
@@ -61,31 +85,41 @@ In order to decouple triggers and components, we move triggers to new top-level
 ```toml
 [[trigger.http]]
 route = "/world"
-component = "world"
-
-[[component]]
-id = "world"
+handler = "world"
```
mikkelhegn, line 88: Should this be `handler_component`? Handler makes sense, because that's what it does :-), but below the proposal still refers to it as a component `[component.world]`.

docs/content/sips/005-manifest-redesign.md
```diff
@@ -61,31 +85,41 @@ In order to decouple triggers and components, we move triggers to new top-level
...
 ```
 
-### Manifest version
+### Component inline config
 
-Since this proposal represents a backward-incompatible change to the manifest format, the manifest
-version changes. The existing `spin_version = "1"` field name leaves some room for misinterpretation
-as "compatible with Spin project version 1". As an optional part of this proposal, change that name:
+In addition to being defined in their own top-level sections and referenced by ID,
+component configs may be inlined; allowing this:
```
mikkelhegn, line 96: Why does the proposal want to allow this?

lann: It keeps the manifest smaller and linked component/trigger configs visually close together.

docs/content/sips/005-manifest-redesign.md
lann: Sure, it could be. Just a tradeoff between being concise and being explicit I guess :shrug:

docs/content/sips/005-manifest-redesign.md
```diff
@@ -97,30 +131,17 @@ is equivalent to:
...
-Instead of updating the manifest version we could not do that. It would be nice to include
-error message guidance and/or a `spin doctor` tool to upgrade the manifest format. This
-would be easy to detect if we switch the `spin_version` key.
+This upgrade could be performed automatically by `spin doctor`.
 
-### Trigger IDs
-
-We might want to be able to reference individual triggers, for example to implement a
-`spin up --trigger=world`. It would save some compatibility headache if we require IDs now:
+## Design options
```
fibonacci1729, line 141: I think another change that might mitigate future confusion would be to rename `[component.config]` to `[component.variables]`. Especially if/when preview2 value imports and templates (which may be too far out) become a thing.

lann: Oh yeah I'll add that.

docs/content/sips/005-manifest-redesign.md
mikkelhegn: I think it makes it easier to talk about if it's explicit.

docs/content/sips/005-manifest-redesign.md
```diff
@@ -104,11 +104,11 @@ id = "hello"
 
 ## Design options
 
-### Don't update the manifest version
```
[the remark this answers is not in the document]
radu-matei: +1

docs/content/sips/005-manifest-redesign.md
radu-matei: Opinion alert: I don't think there is a danger of confusion with just `handler`.

fibonacci1729: +1 to @radu-matei ....want to avoid `component` studder everywhere.

mikkelhegn: Can you elaborate on that Brian? Are we trying to get rid of component as a Spin concept, and/or align with webassembly concepts and/or wasi-http concepts?

fibonacci1729: I think what motivates my lean towards the concise here is that conceptually a trigger, as modeled in WIT (or generally as a component), might `import` an instance of whatever `http` component (`hello` in this example) and that `import` name expressed in WIT might look like `import handler: ...` or `import fermyon:spin/handler`. You probably wouldn't say in WIT `import handler_component: ...`.

I definitely tend to prefer explicitness but in this case (albeit without great justification) I lean towards the concise I suppose. Seeing `*_component` throughout a larger example manifest would be hard on the eyes, me thinks.

This is not a hill I'm willing to die on though. 

Alternative suggestion, what if we named the field simply `component` which would ameliorate the concise vs explicitness tension. 

🤷‍♂️

mikkelhegn: Thanks for the input. Rereading the combined example, I can see how the 'handler' as part of a 'trigger' section, as a reference to 'who is handling this trigger' (which is a reference to a component; either inline source, or a 'component' section.) is easy enough to explain. Thx!

mikkelhegn (approved):
LGTM

docs/content/sips/005-manifest-redesign.md
```diff
@@ -61,31 +85,47 @@ In order to decouple triggers and components, we move triggers to new top-level
 ```toml
 [[trigger.http]]
 route = "/world"
```
rylev, line 87: super nit: can we use a different example here than "world" for the route and handler component? At first I got confused by the lines below about whether this referred to a wit world. Using "hello-world" instead might make it even more explicit that this just an arbitrary string.

docs/content/sips/005-manifest-redesign.md
```diff
@@ -61,31 +85,47 @@ In order to decouple triggers and components, we move triggers to new top-level
...
+
+```toml
+[[trigger.http]]
+route = "/hello"
+handler = "<generated-id>"
+[component.<generated-id>]
+source = "hello.wasm"
+```
+
+### Rename `config`
```
rylev, line 114: 🎉 +1000

karthik2804:
Does this [issue discussion](https://github.com/fermyon/spin/issues/1768) come into the scope of this PR as well?

lann:
> Does this issue discussion come into the scope of this PR as well?

Not necessarily: it looks like the proposal there would fit into either a v1 or v2 manifest in pretty much the same way.

radu-matei (approved):

fibonacci1729 (approved):

michelleN (commented):
All of these changes look great. The only thing I'm a little confused about is generated IDs. Can you expand a bit more on that? What is generating these IDs? What does that experience look like? Why would you go for using generated IDs?

lann:
> generated IDs

Generated IDs would only apply to "inline" component defs like

```toml
[[trigger.http]]
handler = { path = "main.wasm" }
```

In this case the trigger doesn't need an explicit component ID to reference, but we might still want the component to have an ID for things like https://github.com/fermyon/spin/issues/1116.

I think we can leave the exact generated IDs up to the implementation but I would imagine something like `inline-1` or maybe something more clever like `inline-main`.

michelleN (approved):

PR #1760, opened 2023-09-13
SQLite vector add-in

itowlson:
This proposed feature is in support of the LLM work.  I am shamefully ignorant of the details, something like storing and querying embeddings but I am a mere pawn of @karthik2804 who is better placed than me to explain why it's worth having.

The trouble is that the library we want to use provides static binaries only for Linux/x64, Mac/x64, and Max/aarch64.  This leaves us with a problem of what to do about Linux/arm64 and Windows/x64.  We could say it's not supported on those, but then we have different capabilities on different builds for no reason that is logically apparent to Spin users (it's an implementation detail of an upstream library).  Or we could try to PR support into the [upstream sqlite-vss repo](https://github.com/asg017/sqlite-vss), if Fermyon was willing to invest in that work (and the upstream author was willing to maintain it) - I have no idea how hard it would be though.

This probably comes down to a community / business decision.  For the time being, this PR is mostly about seeing how badly it breaks on the platforms where it _ought_ to work.  But at least it compiled on _my_ machine...

...once.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,7 +112,11 @@ jobs:
       # # - name: show me the libomp
       # #   run: which libomp
       # - name: seriously show me the libomp
-      #   run: tree /opt/homebrew
+      #   run: ls -l -R /usr/local/opt
+      - name: seriously show me the blas I REALLY MEAN THIS
+        run: ls -l -R /usr/local/Cellar/openblas
+      - name: AND LAPACK COME ON
+        run: ls -l -R /usr/local/Cellar/lapack
       # - name: also show me the libomp in a sane way
       #   run: tree /opt/homebrew | grep libomp
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -90,6 +90,13 @@ jobs:
           rust-wasm: true
           rust-cache: true
 
+      - name: Install sqlite-vss dependencies
+        if:  ${{ matrix.os == 'macos-latest' }}
+        run: |
+          brew install llvm
+          brew install libomp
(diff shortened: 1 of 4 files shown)


radu-matei:
ref https://github.com/asg017/sqlite-vss/issues/82

radu-matei:
As far as I can tell, Linux ARM can be built (ref https://github.com/asg017/sqlite-vss/issues/82), so it's a matter of generating the binary as part of the release.
It would be great to contribute upstream this feature.

Windows seems to be somewhat more difficult - https://github.com/asg017/sqlite-vss/issues/82 and https://github.com/facebookresearch/faiss/issues/2551.
Not supporting it for now (and pointing to running in WSL) would make sense for now.

itowlson:
Even the plain Ubuntu build fails on linking to libatlas.  The sqlite-vss documentation warns the "on Linux machines, you'll have to install some packages to make these options work:

```
sudo apt-get update
sudo apt-get install -y libgomp1 libatlas-base-dev liblapack-dev
```

These are evidently not on the GH runner, and weren't on my WSL installation, which makes me suspect they'll likely not be on many contributors' machines.  So that's going to create a speed bump for contributors.  As always, this isn't a blocker, just something to factor into the decision.

itowlson:
With those packages in, Ubuntu builds. Yay!

On to Mac and we get a lot of:

```
warning: object file (/Users/runner/work/spin/spin/target/release/deps/libsqlite_vss-630e52b931658673.rlib(AutoTune.cpp.o)) was built for newer macOS version (12.6) than being linked (12.0)
```

which is odd because we build on `macos-latest`.  But the error is that it's looking for good ol' `omp` again so trying some brew installs.

mikkelhegn:
Desired in 1.6 :-)

PR #1763, opened 2023-09-13
Update to wasmtime 13

rylev:
Marking this as a draft until the official release of wasmtime 13.

This uses [this commit](https://github.com/bytecodealliance/wasmtime/commit/409bdd0330d9525846cc199f5bfe95b6a695f735) off of the `release-13.0.0` branch. When the wasmtime 13 release happens, we should update this to point to that release instead of this commit.

This relies on two other changes:
* https://github.com/fermyon/spin-componentize/pull/21
* https://github.com/fermyon/wit-bindgen-backport/pull/14

Additionally, we need to make some decisions:
* [x] Wasi now requires io buffers to specify their capacity. We should decide what a good capacity for these buffers should be
* [x] The io handle setters require specifying whether the given input or output stream is for a terminal or not. I believe this is used in `terminal` interface. We don't really track whether to given streams are the terminal or not, and some are a combination of the terminal and a file. I've done the conservative thing and marked these all as `IsATTY::No`, but we probably want to review that.

### DIFF
--- a/crates/core/src/io.rs
+++ b/crates/core/src/io.rs
@@ -1,32 +1,33 @@
-use std::sync::{Arc, RwLock};
-
-use wasmtime_wasi::preview2::pipe::WritePipe;
+use wasmtime_wasi::preview2::{pipe::MemoryOutputPipe, HostOutputStream};
 
 /// An in-memory stdio output buffer.
-#[derive(Default)]
-pub struct OutputBuffer(Arc<RwLock<Vec<u8>>>);
+pub struct OutputBuffer(MemoryOutputPipe);
 
 impl OutputBuffer {
-    /// Takes the buffered output from this buffer.
-    pub fn take(&mut self) -> Vec<u8> {
-        std::mem::take(&mut *self.0.write().unwrap())
+    /// Clones the buffered output from this buffer.
+    pub fn contents(&self) -> bytes::Bytes {
+        self.0.contents()
     }
 
-    pub(crate) fn writer(&self) -> WritePipe<Vec<u8>> {
-        WritePipe::from_shared(self.0.clone())
+    pub(crate) fn writer(&self) -> impl HostOutputStream {
+        self.0.clone()
+    }
+}
+
+impl Default for OutputBuffer {
+    fn default() -> Self {
+        Self(MemoryOutputPipe::new(usize::MAX))
     }
 }
 
 #[cfg(test)]
 mod tests {
-    use wasmtime_wasi::preview2::OutputStream;
-
     use super::*;
 
     #[tokio::test]
     async fn take_what_you_write() {
-        let mut buf = OutputBuffer::default();
-        buf.writer().write(b"foo").await.unwrap();
-        assert_eq!(buf.take(), b"foo");
+        let buf = OutputBuffer::default();
+        buf.writer().write(b"foo".to_vec().into()).unwrap();
+        assert_eq!(buf.contents().as_ref(), b"foo");
     }
 }
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -97,21 +97,29 @@ impl Default for Config {
         // knobs for each of these settings just yet and instead they're
         // generally set to defaults. Environment-variable-based fallbacks are
         // supported though as an escape valve for if this is a problem.
-        //
-        // NB: much of this will change in Wasmtime 13 as the settings are
-        // different. Ping @alexcrichton for assistance in updating this if
-        // needed (and delete this comment after the 13 update).
         let mut pooling_config = PoolingAllocationConfig::default();
         pooling_config
-            .instance_count(env("SPIN_WASMTIME_INSTANCE_COUNT", 1_000))
-            .instance_size(env("SPIN_WASMTIME_INSTANCE_SIZE", (10 * MB) as u32) as usize)
-            .instance_tables(env("SPIN_WASMTIME_INSTANCE_TABLES", 2))
-            .instance_table_elements(env("SPIN_WASMTIME_INSTANCE_TABLE_ELEMENTS", 30_000))
-            .instance_memories(env("SPIN_WASMTIME_INSTANCE_MEMORIES", 1))
+            .total_component_instances(env("SPIN_WASMTIME_INSTANCE_COUNT", 1_000))
+            // This number accounts for internal data structures that Wasmtime allocates for each instance.
+            // Instance allocation is proportional to the number of "things" in a wasm module like functions,
+            // globals, memories, etc. Instance allocations are relatively small and are largely inconsequential
+            // compared to other runtime state, but a number needs to be chosen here so a relatively large threshold
+            // of 10MB is arbitrarily chosen. It should be unlikely that any reasonably-sized module hits this limit.
+            .max_component_instance_size(
+                env("SPIN_WASMTIME_INSTANCE_SIZE", (10 * MB) as u32) as usize
+            )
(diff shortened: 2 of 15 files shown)


crates/core/src/lib.rs
```diff
@@ -97,21 +97,17 @@ impl Default for Config {
...
-        // different. Ping @alexcrichton for assistance in updating this if
-        // needed (and delete this comment after the 13 update).
         let mut pooling_config = PoolingAllocationConfig::default();
         pooling_config
-            .instance_count(env("SPIN_WASMTIME_INSTANCE_COUNT", 1_000))
-            .instance_size(env("SPIN_WASMTIME_INSTANCE_SIZE", (10 * MB) as u32) as usize)
-            .instance_tables(env("SPIN_WASMTIME_INSTANCE_TABLES", 2))
-            .instance_table_elements(env("SPIN_WASMTIME_INSTANCE_TABLE_ELEMENTS", 30_000))
-            .instance_memories(env("SPIN_WASMTIME_INSTANCE_MEMORIES", 1))
+            .total_component_instances(env("SPIN_WASMTIME_INSTANCE_COUNT", 1_000))
```
rylev, line 102: ping @alexcrichton since it was noted that this would change. I think the changes here are correct?

crates/core/src/lib.rs
```diff
@@ -97,21 +97,17 @@ impl Default for Config {
...
+            .max_core_instance_size(env("SPIN_WASMTIME_INSTANCE_SIZE", (10 * MB) as u32) as usize)
+            .total_tables(env("SPIN_WASMTIME_INSTANCE_TABLES", 2))
+            .table_elements(env("SPIN_WASMTIME_INSTANCE_TABLE_ELEMENTS", 30_000))
+            .total_memories(env("SPIN_WASMTIME_INSTANCE_MEMORIES", 1))
```
alexcrichton, line 106: `total_{tables,memories}` will want to change here to be on the order of `instance_count`, probably with their own separate env vars to configure them

alexcrichton (approved):

crates/core/src/io.rs
```diff
@@ -1,32 +1,34 @@
...
-    pub(crate) fn writer(&self) -> WritePipe<Vec<u8>> {
-        WritePipe::from_shared(self.0.clone())
+    pub(crate) fn writer(&self) -> impl HostOutputStream {
+        self.0.clone()
+    }
+}
+
+impl Default for OutputBuffer {
+    fn default() -> Self {
+        Self(MemoryOutputPipe::new(1024 * 1024))
```
lann, line 20: I don't know what this _should_ be, but 1MB seems too small; if I'm reading this right it would cap the WAGI output size, right?

alexcrichton: I believe that's correct yeah. To get equivalent behavior as before you'd use `usize::MAX`, so that's a possibility too.

rylev: Well it's the size of the buffer for a single write not for the entire instance, but I think we should set it to `usize::MAX` since that's what we effectively had before.

@lann this type doesn't seem to be used except for in tests. Would it make sense to just get rid of it?

lann: Oh sure enough. It used to be used by wagi but it looks like it was removed at some point. :shrug: 

I don't see it used anywhere else, so should be fine to remove.

crates/core/src/lib.rs
rylev: @alexcrichton does this look right now?

crates/core/src/lib.rs
```diff
@@ -97,21 +97,21 @@ impl Default for Config {
...
-            .instance_tables(env("SPIN_WASMTIME_INSTANCE_TABLES", 2))
-            .instance_table_elements(env("SPIN_WASMTIME_INSTANCE_TABLE_ELEMENTS", 30_000))
-            .instance_memories(env("SPIN_WASMTIME_INSTANCE_MEMORIES", 1))
+            .total_component_instances(env("SPIN_WASMTIME_INSTANCE_COUNT", 1_000))
+            .max_component_instance_size(
+                env("SPIN_WASMTIME_INSTANCE_SIZE", (10 * MB) as u32) as usize
+            )
+            .max_tables_per_component(env("SPIN_WASMTIME_INSTANCE_TABLES", 2))
+            .table_elements(env("SPIN_WASMTIME_INSTANCE_TABLE_ELEMENTS", 30_000))
+            .max_memories_per_component(env("SPIN_WASMTIME_INSTANCE_MEMORIES", 1))
```
alexcrichton, line 108: While this will allow instantiating components this would notably disallow instantiating composed components so the default here may want to be higher than 1. This limit is for the entire component loaded into Wasmtime not each individual component itself within that.

(same for tables above)

alexcrichton (commented):
Just one comment, everything else looks good 👍

crates/core/src/lib.rs
rylev: Set this to 10, but I'm not really sure what a good default would be here...

tschneidereit: @alexcrichton does this limit make each component use the specified number of memory slots in the pooling allocator, or do we only allot the number actually required? If the former, that seems pretty unfortunate, because I think we'll see components with larger compositions pretty soon.

Assuming it's the latter, I think we should bump this up, maybe to 32 or so, because otherwise we'll soon run into issues with at least demos where we want to have more than 10 components in a composition

lann (commented):
Woops, this review has been pending for a bit...

crates/core/src/lib.rs
alexcrichton: The latter (after Nick's recent refactoring). So yes bumping it to 32 should be fine, that just limits the max concurrent instances if everything happens to use 32

crates/core/src/lib.rs
```diff
@@ -97,21 +97,21 @@ impl Default for Config {
...
         let mut pooling_config = PoolingAllocationConfig::default();
         pooling_config
-            .instance_count(env("SPIN_WASMTIME_INSTANCE_COUNT", 1_000))
-            .instance_size(env("SPIN_WASMTIME_INSTANCE_SIZE", (10 * MB) as u32) as usize)
-            .instance_tables(env("SPIN_WASMTIME_INSTANCE_TABLES", 2))
-            .instance_table_elements(env("SPIN_WASMTIME_INSTANCE_TABLE_ELEMENTS", 30_000))
-            .instance_memories(env("SPIN_WASMTIME_INSTANCE_MEMORIES", 1))
+            .total_component_instances(env("SPIN_WASMTIME_INSTANCE_COUNT", 1_000))
+            .max_component_instance_size(
+                env("SPIN_WASMTIME_INSTANCE_SIZE", (10 * MB) as u32) as usize
```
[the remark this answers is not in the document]
alexcrichton: Since this is carried over from my original changes, some documentation for this could be:

> This number accounts for internal data structures that Wasmtime allocates for each instance. Instance allocation is proportional to the number of "things" in a wasm module like functions, globals, memories, etc. Instance allocations are relatively small and are largely inconsequential compared to other runtime state, but a number needs to be chosen here so a relatively large threshold of 10MB is arbitrarily chosen. It should be unlikely that any reasonably-sized module hits this limit.

crates/core/src/lib.rs
rylev: Set to 32 and added a comment.

crates/core/src/lib.rs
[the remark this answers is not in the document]
lann: Sorry I actually completely misread this before and thought it was `SPIN_WASMTIME_ISNTANCE_SIZE * 10MB`.

PR #1766, opened 2023-09-13
Use github form syntax for issue templates

bacongobbler:
This helps provide a bit more structure around new bug reports, including:

- automatically attaching labels to new issues
- providing discrete forms for each suggestion

This follows a similar structure to the issue templates provided by the GitHub docs team: https://github.com/github/docs/tree/main/.github/ISSUE_TEMPLATE


itowlson (commented):
Apart from wording nits and things, I get the sense that a lot of this is aimed at prompting concrete thoughts about the problem or wish - but I fear it could feel like a quiz you have to complete successfully, and that may put people off.  It would be great to find ways of offering this as "Here are some ideas that for things that would be helpful if you know them and if they make sense for your issue," but I'm not sure how to do that in a way that doesn't very quickly come back to a scary list of questions.

I am actually wondering if we should aim for something looser rather than tighter, maybe even ditching the current bug/suggestion split.  Make it as easy as possible for people to engage even if that means more organising and following up on our part.  I'd be interested to hear what other folks think.

PR #1775, opened 2023-09-18
Point templates on `main` at SDKs on `main`

### LINKED ISSUE #1733: Templates on spin for rust need to point at main for their spin-sdk dependency

itowlson:
Fixes #1733.

~~To determine: whether / what process changes are needed to support this (specifically making sure the change to numbered happens on the release branch rather than on main).~~ commits added below - @vdice

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - main
+      - "v[0-9]+.[0-9]+"
     tags:
       - "v*"
 
@@ -289,7 +290,14 @@ jobs:
 
       - name: Set the spin tag
         shell: bash
-        run: echo "SPIN_TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+        run: |
+          echo "SPIN_TAG=${GITHUB_REF/refs\/tags\//}" >> $GITHUB_ENV
+
+      - name: Set the PR base branch
+        shell: bash
+        run: |
+          IFS=. read -r major minor patch <<< "${{ env.SPIN_TAG }}"
+          echo "RELEASE_BRANCH=$major.$minor" >> $GITHUB_ENV
 
       - name: Change sdk version
         shell: bash
@@ -312,7 +320,7 @@ jobs:
           title: "feat(templates): update sdk to ${{ env.SPIN_TAG }}"
           body: Update the SDK version used by the templates
           branch: update-sdk-${{ env.SPIN_TAG }}
-          base: main
+          base: ${{ env.RELEASE_BRANCH }}
           delete-branch: true
           committer: fermybot <103076628+fermybot@users.noreply.github.com>
           author: fermybot <103076628+fermybot@users.noreply.github.com>
--- a/docs/content/release-process.md
+++ b/docs/content/release-process.md
@@ -14,7 +14,6 @@ To cut a major / minor release of Spin, you will need to do the following:
 
 1. Switch to the release branch locally and update versions (e.g. `1.1.0-pre0` could `1.1.0`).
    - Bump the version in Spin's `Cargo.toml`
-   - Update SDK_VERSION in `templates/Makefile`
    - Run `make build` so that `Cargo.lock` and other associated files are updated
 
    PR these changes to the release branch ensuring that pull request has a base corresponding to the release branch (e.g. `v1.1`).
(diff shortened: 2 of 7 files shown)


itowlson:
Looks like the templates want `branch = "main"` rather than `tag = "main"` _sigh_

vdice:
> To determine: whether / what process changes are needed to support this (specifically making sure the change to numbered happens on the release branch rather than on main).

I think the following changes could be the appropriate supplements:
https://github.com/fermyon/spin/commit/99a601d0b7eb7a1664e76f22a7ccb804a01591f9
https://github.com/fermyon/spin/commit/0b69dfe87218606b9051c1e1984956bb17f72621

The updates to the [release workflow](https://github.com/fermyon/spin/commit/99a601d0b7eb7a1664e76f22a7ccb804a01591f9) would:
1. Change the `create-template-sdk-update-pr` job to use the `v[0-9].[0-9]` release branch as its PR base, not `main`
2. Also trigger on pushes to the `v[0-9].[0-9]` release branches (so that the `push-templates-tag` job runs when the automated PR mentioned above is merged)

Then, we can [remove the sections of the release process doc](https://github.com/fermyon/spin/commit/0b69dfe87218606b9051c1e1984956bb17f72621) around manually bumping the `SDK_VERSION` in `templates/Makefile`, since `latest`/`branch = "main"` will be the default (other than `v*`/`spin/templates/v*` tags/branches) and the only agent needing to run `SDK_VERSION=v* make` would be the automation mentioned above.

How does that look @itowlson?  Am I missing any details?

itowlson:
@vdice That all looks plausible!  What do you reckon, do you want to merge those onto this PR, or shall I merge this (when approved) and you do those as a separate PR?

Also, I'm not sure how to test a change like this which is so bound up with the release process - any ideas?

vdice:
Okay, the CI changes [added here](https://github.com/itowlson/spin/compare/main-templates-point-to-main-sdk...vdice:spin:itowlson-main-templates-release-updates) are working as we envisioned in https://github.com/fermyon/spin/pull/1775#issuecomment-1724514871.

A few notes/Qs:

- When I run `go mod tidy` on a go/tinygo-based app that uses `latest` for its version, it appears to resolve to `v1.5.0` -- though don't we expect/hope it would resolve to a specific main commit?  eg
    ```
    require github.com/fermyon/spin/sdk/go latest
    ```
    turns into the following
    ```
    require github.com/fermyon/spin/sdk/go v1.5.0
    ```
    Or perhaps it is working as intended and that's how go versions/mod works?
    
- With these changes when I download a release binary of Spin (eg v1.6.0), the logic on `spin templates install [--upgrade]` will pull templates from the eg `sdk/templates/v1.6.0` tag, correct?  Alternatively, if I have a non release binary of Spin (eg built from main or a commit other than a tag), it will pull templates from... `main` in all cases?  I'm fuzzy on this latter case.

itowlson:
Ugh, I don't know how go.mod refers to main then - that was my inference from reading the Go docs but maybe `latest` means "latest tagged."  This sort of thing is why I tried to rope @adamreese into the madness...

Your last paragraph is partially correct.  _All_ builds of Spin have version numbers - 1.4.2, 1.5.0, 1.6.0-preNNN.  What matters is not whether the version is "pre" (i.e. from main) but whether there is a template tag matching the version (to major-minor level).  So if you are on 1.5 release, it will look for the 1.5 templates tag.  But if you are on a 1.5 prerelease, _it will also look for the 1.5 template tags_.  If you are on the 1.6 prerelease, it will look for the 1.6 template tags, and (at the time of writing) will not find them, so will fall forward to main.  But if we later do a 1.6 release, and you re-install the templates _even without changing your Spin binary_ it will then find the 1.6 templates, because it cares only about the existence of tags, not about releases and non-releases.  Hope that makes sense!  (I think it's simpler than I've made it sound!  But maybe I'm just used to it...)

vdice:
re: go.mod version.  using `main` appears to work, with a caveat:

```
require github.com/fermyon/spin/sdk/go main
```
turns into the following after `go mod tidy`:
```
require github.com/fermyon/spin/sdk/go v1.3.1-0.20230919215209-a0239d34a2a8
```
Which I'm assuming is what we want, as `a0239d34a2a8` matches the commit at HEAD of main at time of writing.

The caveat: `v1.3.1` prefix.  Where is this coming from?  How can we ensure it is the actual most recent tag of Spin?

adamreese:
I tried hacking a solution but I don't think you can reference `main` in go.mod without doing a `go mod tidy`. One unelegant option would be to add `go mod tidy` in the build command in spin.toml.

I have no idea where the `v1.3.1` prefix is coming from. Spin doesn't even have a `v1.3.1` release. ¯\_(°ペ)_/¯

itowlson:
It's probably okay if people using canary Spin and canary templates have to manually run `go mod tidy`.  As long as that weird 1.3.1 thing isn't actually causing any problems.

itowlson:
Can we get a review on this please?  I can't approve it because I'm one of the authors.

adamreese (approved):

PR #1776, opened 2023-09-18
Add `spin up --build` option

### LINKED ISSUE #1735: spin up  --build / spin deploy --build

itowlson:
Fixes #1735.

> A separate cloud plugin PR would be needed for `spin deploy --build`.

```
$ spin up --build
Building component hello3 with `cargo build --target wasm32-wasi --release`
   Compiling hello3 v0.1.0 (/home/ivan/testing/hello3)
    Finished release [optimized] target(s) in 0.38s

Serving http://127.0.0.1:3000
Available Routes:
  hello3: http://127.0.0.1:3000 (wildcard)
```

Buildception: I tried it with `spin build --up --build` and `spin watch --build` although it did some wasted work there was no recursion or unexpected CPU / disk usage.  At least not in my simple test case.  I am a bit concerned that there might be lurking possibilities for buildception in `spin watch --build`, and that it looks enough like the opposite of `spin watch --skip-build` for people to think that it's a good idea to pass it, but not enough to really worry about it for now!

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -85,6 +85,12 @@ pub struct UpCommand {
     #[clap(long, takes_value = false)]
     pub direct_mounts: bool,
 
+    /// For local apps, specifies to perform `spin build` before running the application.
+    ///
+    /// This is ignored on remote applications, as they are already built.
+    #[clap(long, takes_value = false, env = "SPIN_ALWAYS_BUILD")]
+    pub build: bool,
+
     /// All other args, to be passed through to the trigger
     #[clap(hide = true)]
     pub trigger_args: Vec<OsString>,
@@ -123,6 +129,10 @@ impl UpCommand {
             } else {
                 bail!("Default file '{DEFAULT_MANIFEST_FILE}' not found. Run `spin up --from <APPLICATION>`, or `spin up --help` for usage.");
             }
+        }
+
+        if self.build {
+            app_source.build().await?;
         }
 
         // Get working dir holder and hold on to it for the rest of the function.
@@ -437,6 +447,13 @@ impl AppSource {
                 None
             }),
             _ => None,
+        }
+    }
+
+    async fn build(&self) -> anyhow::Result<()> {
+        match self {
+            Self::File(path) => spin_build::build(path, &[]).await,
+            _ => Ok(()),
         }
     }
 }


karthik2804:
This feature is interesting! I am a little concerned about the potential confusion this may cause with both `spin build --up` and `spin up --build` being 2 different ways to do the same thing.  Spin watch emulates a part of the `spin up --build` behavior, which could mean 3 different ways to do the same thing.

I do not have very strong opinions here but wanted to bring it up to see if other people had similar concerns.

calebschoepp:
I share some of @karthik2804 and @itowlson's concerns. What is the use case we're trying to support here? I'm a little wary to be littering `--build` flags through every sub-command in Spin.

mikkelhegn:
Is he real "solution" here to remove (or-deprecate) `spin build --up` now that we have `spin watch`? If I remember correctly that was the order in which they were implemented. So we went from `spin build && spin up` to `spin build --up` and now `spin watch`.

Maybe @michelleN can give some more nuances to the original issue, to help understand the motivation.

itowlson:
`spin build --up` != `spin watch` and we should not force people to use the latter when they want the former.

Not meaning to speak for @michelleN but my impression from discussion at the time was that people were asking for a way to make sure their applications were up to date before they ran or (especially) deployed them.  That's why I defined an environment variable to this, so that people who wanted to _always_ build when they upped or deployed could set it in `.profile` and forget it.

calebschoepp:
> `spin build --up` != `spin watch` and we should not force people to use the latter when they want the former.

Agreed.

> Not meaning to speak for @michelleN but my impression from discussion at the time was that people were asking for a way to make sure their applications were up to date before they ran or (especially) deployed them. That's why I defined an environment variable to this, so that people who wanted to _always_ build when they upped or deployed could set it in `.profile` and forget it.

I'm realizing it was actually me who raised the original issue on Slack:

> I would love some sort of warning to be displayed if you `spin cloud deploy` without having run `spin build` first. The amount of time I've wasted wondering why my code isn't working only to realize that I forgot to rebuild the app is silly.
Lots of ways we could do this but one might be to have `spin cloud deploy` keep a cache of the last deployed app and then warn you if you're redeploying the same app

I was only looking for a warning from `spin deploy` and then it morphed into the ability to have `spin deploy` build your code with a `--build` flag. I guess given the original issue I'm not really sure how we arrived at adding a `--build` flag to `spin up`.

michelleN:
> Maybe @michelleN can give some more nuances to the original issue, to help understand the motivation.

Yea, I should have added more context to the original issue. It's been bothering me that I have to think about the build step every time I want to iterate on my application so my hope is we can replace `spin build --up` with `spin up --build`. `up` is the action that the user most likely wants and it's more intuitive (to me) to make build an option of up rather than the other way around. It's much less likely that the user wants to `build` and not `up`. If we add `--build` as an option to `up` then the user only really needs to know about one command, one action: `spin up`. Linking `--build` to an environment variable that we recommend setting further reduces cognitive load.

When @calebschoepp mentioned his issue, I thought this might resolve both of our concerns so my bad for trying to morph those together without adding context!

karthik2804:
I agree with `--build` being more appropriate on `spin up` than `--up` on `spin build`. I would prefer we potentially remove `--up` on `spin build` with `2.0` maybe? 

I totally agree with `spin deploy --build`.

calebschoepp:
> When @calebschoepp mentioned his issue, I thought this might resolve both of our concerns so my bad for trying to morph those together without adding context!

No worries. I guess stylistically I'm not the biggest fan of `--build` and `--up` flags for my own workflow, but I totally see how others would find them helpful. I'm content to move forward with adding these things.

I didn't see any issues on the code side of things.

src/commands/up.rs
```diff
@@ -85,6 +85,12 @@ pub struct UpCommand {
     #[clap(long, takes_value = false)]
     pub direct_mounts: bool,
 
+    /// For local apps, specifies to perform `spin build` before running the application.
+    ///
+    /// This is ignored on remote applications, as they are already built.
+    #[clap(long, takes_value = false, env = "SPIN_BUILD_BEFORE_UP")]
```
michelleN, line 91: might be nice to make this env var more generic in case we want to use it for other commands like deploy in the future.
`SPIN_ALWAYS_BUILD` wdyt?

michelleN (commented):
THANK YOU. Left a question about the env var

src/commands/up.rs
itowlson: @michelleN That was my first thought too!  It would be spot on for `.profile` and could be shared with `spin cloud deploy`.  But I got cold feet thinking "it looks silly writing `SPIN_ALWAYS_BUILD=true spin up`".  BUT NOBODY WOULD EVER WRITE THAT IVAN.  Right? ...right?

michelleN (approved):

PR #1780, opened 2023-09-19
Manifest V2

lann:
> "I have only made this PR larger because I have not had the time to make it smaller."

Implements SIP-005 as updated in https://github.com/fermyon/spin/pull/1753.

Some highlights:

- Completely refactored `spin-manifest`
  - `Application` et al. are no more; there is only the `AppManifest`, which closely resembles to the parsed `spin.toml`
  - There is also an `AppManifestV1`. Backward compat is provided by `spin_manifest::compat::v1_to_v2_app`.
- Mostly-rewritten `spin-loader` crate takes an `AppManifest` and converts it into a `LockedApp`, ready to execute
- A new `ui-testing` crate, supporting expanded "UI" (a.k.a. "golden file") test suites for both `spin-manifest` and `spin-loader`

There are some (intentional) changes open for debate:

- Validation of component IDs is now more strict, requiring them to be valid component model `kebab-case`. The v1->v2 conversion should make this mostly transparent (search `component_id_from_string` if you care), but there may be some IDs floating around that won't parse now.
- Rules for `files` patterns have been relaxed a bit; you can now use the "placement" form for individual files, e.g. `{ source = "content/some.txt", destination = "/other.txt" }`
- Added a `"triggers"` key to locked app metadata which is currently redundant but should help with forward-compatibility for multiple-trigger apps
- Probably others that I have already forgotten or haven't yet noticed

Still to do is the `spin doctor` manifest upgrade feature, though hopefully that is pretty simple application of the `v1_to_v2_app` code. There are examples of what that should look like in `crates/manifest/tests/ui/v1/*.toml.v2`


src/commands/up.rs
```diff
@@ -458,6 +461,17 @@ impl AppSource {
     }
 }
 
+impl Display for AppSource {
```
itowlson, line 464: This reads more like a Debug implementation than a Display one.

src/commands/up.rs
```diff
@@ -458,6 +461,17 @@ impl AppSource {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AppSource::None => write!(f, "<no source>"),
+            AppSource::File(path) => write!(f, "local app {path:?}"),
+            AppSource::OciRegistry(reference) => write!(f, "remote app {reference:?}"),
+            AppSource::Unresolvable(s) => write!(f, "unknown app source {s:?}"),
```
itowlson, line 470: Unless I've missed where you changed it, the `Unresolvable` case carries an error message, not the original `app_source` flag text from the CLI.

crates/manifest/Cargo.toml
```diff
@@ -5,7 +5,21 @@ authors = { workspace = true }
 edition = { workspace = true }
 
 [dependencies]
-indexmap = "1"
-serde = { version = "1.0", features = [ "derive" ] }
+anyhow = "1.0.75"
+indexmap = { version = "1", features = ["serde"] }
+outbound-http = { path = "../outbound-http" }
+serde = { version = "1.0", features = ["derive"] }
+spin-app = { path = "../app" }
```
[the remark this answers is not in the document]
lann: Oh actually yes. I did a "late refactor" that should make that entirely possible but didn't think about it.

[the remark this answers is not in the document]
lann: OK, I removed the one remaining dependency by refactoring common serde helpers into a new `spin-serde` crate.

crates/manifest/tests/ui/invalid/bad_component_id.err
```diff
@@ -0,0 +1,5 @@
+TOML parse error at line 6, column 12
+  |
+6 | [component.bad_id]
```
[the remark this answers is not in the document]
lann: Yeah, this was my main motivation for moving more validation into the serde schema.

src/commands/new.rs
```diff
@@ -213,7 +212,7 @@ impl FromStr for ParameterValue {
 /// returns the resulting hashmap of key-value pairs.
 async fn values_from_file(path: impl AsRef<Path>) -> Result<HashMap<String, String>> {
     // Get the absolute path of the file we're reading.
-    let path = absolutize(path)?;
+    let path = path.as_ref();
```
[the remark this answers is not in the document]
lann: I believe it was cargo-culted from elsewhere :shrug:

src/commands/up.rs
```diff
@@ -311,15 +313,14 @@ impl UpCommand {
         manifest_path: &Path,
         working_dir: &Path,
     ) -> Result<LockedApp> {
-        let asset_dst = if self.direct_mounts {
-            None
+        let files_mount_strategy = if self.direct_mounts {
```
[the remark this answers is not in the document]
lann: I started off with the original `Option<PathBuf>` and then immediately wrote a bug because it was confusing :sweat_smile:

src/commands/up.rs
lann: This is responsible for the updated error message here: https://github.com/fermyon/spin/pull/1780/files#diff-82f776c9616e3a5ecda575278ca8de8a920780afe36b0fff8ab7252292425aad

src/commands/up.rs
lann: Added a colon to clarify that it is a "reason". I think in practice this case is actually unreachable anyway.

crates/manifest/src/schema/common.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// `{ ... }`
+    Remote {
+        /// `url = "https://example.test/remote.wasm"`
+        url: String,
+        /// `digest = `"sha256:abc123..."`
+        digest: String,
+    },
+}
+
+/// Component source
```
vdice, line 33:
```suggestion
/// WASI Files mount configuration
```

vdice (commented):
All I can contribute are a couple eagle-eyed comment fixes 😂 

Also wondering what our strategy is around updating example and template app manifests to v2?  Perhaps as a follow-up PR before the 2.0 release?  (I suppose we'll want to coordinate with developer docs updates as well.)

itowlson:
I ~set out to cause unrest~ had a play with this as part of trying out deployment, and found a couple of minor oddities.  These are with the "overnight" version - I know you are still fixing things so these may already be done in your working copy! - and are nits for the sad path not problems with the "normal" flow.

### `type` in app trigger settings?

I wrote:

```
[application.trigger.http]
bayes = "/"
```

It correctly erred, but the message was "Error: metadata error: invalid metadata value for "trigger": Error("unknown field `bayes`, expected `type` or `base`", line: 0, column: 0)".  Seems like `type` shouldn't be allowed, but here we go:

```
[application.trigger.http]
type = "redis"
base = "/"
```

And it was perfectly happy with that.  I wondered if the `http` part was giving a name to the trigger which was of type `redis`, but no, the trigger was still type `http`.

### Mismatched application and component triggers

When I wrote `redis` in my application trigger instead of `http`:

```
[application.trigger.redis]
base = "/"

[[trigger.http]]
component = "hello3"
route = "/..."
```

...it ran without warning.

I am guessing the concept is that all triggers are available via the `trigger` table array, and that the `application.trigger` merely applies global settings to those you happen to use, and if you choose to apply global settings that are never used, then what of it?  Still it felt a bit surprising, especially with a property that doesn't apply to Redis (since HTTP rejected unknown settings).  I'll maybe try to get a better sense of the intended mental model at some point, so we can communicate this correctly.

### Mismatched trigger and configuration

I wrote this (to explore a theory about names and types):

```
[application.trigger.redis]
type = "http"
base = "/"

[[trigger.redis]]
component = "hello3"
route = "/..."
```

It understandably errored, with the message:

```
Error: invalid trigger configuration for "redis-1"

Caused by:
   0: json error: unknown field `route`, expected one of `component`, `channel`, `executor`
   1: unknown field `route`, expected one of `component`, `channel`, `executor`
```

The name "redis-1" does not refer to anything in my TOML file, and it was puzzling to see a "JSON error" when I was working only with TOML.

All of these can absolutely be deferred to issues for after the happy-path functionality has landed - they're probably better than the corresponding manifest v1 errors...!

lann:
> Also wondering what our strategy is around updating example and template app manifests to v2? Perhaps as a follow-up PR before the 2.0 release? (I suppose we'll want to coordinate with developer docs updates as well.)

Yeah I want to make sure there aren't any major changes to the implementation first. There will be at least a couple of followups.

vdice:
> Suggestions for improving this? The name is secretly meaningful in that it is assigned to the 1st redis trigger.

Would it be possible to point to the line in the manifest/toml causing the error (as follow-up, if even feasible)?

lann:
> Would it be possible to point to the line in the manifest/toml causing the error (as follow-up, if even feasible)?

Its tricky currently because the trigger config gets validated by the trigger executor which only gets the locked app JSON. I have a couple of ideas for how to improve the error messages with varying degrees of accuracy and complexity, like injecting parse span information from the original manifest into the locked app metadata. If I can't come up with something that works reasonably well for _all_ trigger types (including custom), I can at least (as a fallback) go back to special-casing validation of http triggers.

crates/http/src/config.rs
```diff
@@ -33,10 +33,9 @@ pub enum HttpExecutorType {
 
 /// Wagi specific configuration for the http executor.
 #[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(deny_unknown_fields)]
+#[serde(default, deny_unknown_fields)]
```
fibonacci1729, line 36: TIL

crates/build/src/manifest.rs
```diff
@@ -1,39 +1,48 @@
...
+    })
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(rename_all = "snake_case")]
-pub(crate) struct RawComponentManifest {
+#[derive(Deserialize)]
+pub struct ComponentBuildInfo {
+    #[serde(default)]
     pub id: String,
```
rylev, line 34: I'd don't love the pattern of deserializing with a defaulted, invalid field that gets updated later. I don't understand the code well enough to offer a suggestion and maybe this just the way it has to be, but if there's some way to rearrange it to not have to do this, that would be awesome.

fibonacci1729 (approved):

crates/loader/tests/ui/invalid-digest.err
```diff
@@ -0,0 +1,6 @@
+Failed to load Spin app from "<test-dir>/invalid-digest.toml"
+
+Caused by:
+    0: Failed to load component `web`
+    1: Failed to load Wasm source Remote { url: "https://example.com/wasm.wasm.wasm", digest: "not-a-digest" }
```
rylev, line 5: nit: can we improve the error message to not print the debug output of the `Remote` type? Maybe like:

Failed to load Wasm source: https://example.com/wasm.wasm.wasm with digest: "not-a-digest"

crates/loader/tests/ui/invalid-url-in-allowed-http-hosts.err
```diff
@@ -0,0 +1,7 @@
+Failed to load Spin app from "<test-dir>/invalid-url-in-allowed-http-hosts.toml"
+
+Caused by:
+    0: Failed to load component `hello`
+    1: One or more allowed_http_hosts entries was invalid:
+       ftp://random-data-api.fermyon.app isn't a valid host or host:port string
+       example.com/wib/wob contains a path, should be host and optional port only
```
rylev, line 7: Does this fail because of the lack of scheme? Perhaps we could make the error message clearer?

crates/loader/src/local.rs
```diff
@@ -0,0 +1,486 @@
...
+            } => {
+                let src = Path::new(source);
+                let dest = dest_root.join(destination.trim_start_matches('/'));
+                self.copy_file_or_directory(src, &dest, exclude_files).await
+            }
+        }
+    }
+
+    // Copy files matching glob pattern or single file/directory path.
+    async fn copy_glob_or_path(
```
rylev, line 243: nit: these copy methods might be best in a separate module.

crates/loader/tests/ui.rs
```diff
@@ -0,0 +1,65 @@
+use std::{ffi::OsStr, path::Path};
+
+use futures::Future;
+use spin_loader::cache::Cache;
+use ui_testing::{Failed, Normalizer, UiTestsRunner};
+
+fn main() -> anyhow::Result<()> {
+    // Insert dummy wasm into cache to avoid network traffic
```
rylev, line 8: I'm confused by this. Why would the network be hit if we don't add the dummy?

crates/manifest/src/normalize.rs
```diff
@@ -0,0 +1,72 @@
+//! Manifest normalization functions.
+
+use std::collections::HashSet;
+
+use crate::schema::v2::{self, AppManifest, ComponentSpec, KebabId};
+
+/// Extracts each `ComponentSpec::Inline` into `AppManifest::component`s,
+/// replacing it with a `ComponentSpec::Reference`.
+pub fn normalize_inline_components(manifest: &mut AppManifest) {
```
rylev, line 9: Perhaps I'm confused by the name `ComponentSpec`, but I'm confused by what this function and `normalize_trigger_ids` actually do a high-level (I get that they set some missing values but why?).

rylev (commented):
Looks good!! I have some questions but mostly because I'm unfamiliar with a lot of this code.

crates/build/src/manifest.rs
lann: It could be done differently with more code, but this bit is somewhat of hack to begin with, extracting partial data from a otherwise-possibly-invalid manifest.

crates/loader/tests/ui/invalid-url-in-allowed-http-hosts.err
lann: :shrug: this is unchanged by this PR; I just added the test

crates/loader/tests/ui.rs
lann: `valid-manifest.toml` has a remote source which the loader would (try and fail to) download.

crates/manifest/src/normalize.rs
lann: It makes code that consumes manifests simpler by not dealing with the flexibility that these normalize.

itowlson:
@lann the "redis-1" error happens in this situation:

```
[[trigger.redis]]
component = "hello3"
route = "/..."
```

I suspect most users are going to think of that as 'the "hello3" trigger' or 'the "hello3" component'.

crates/loader/tests/ui/invalid-url-in-allowed-http-hosts.err
itowlson: @rylev The first one fails because there _is_ a scheme (and it's not a HTTP scheme).  The second fails because it contains a _path,_ not just a host.  Maybe they look too run-together at the moment which confuses matters?  But as Lann says this error message already exists, I can noodle on it separately.

lann:
@itowlson I added some nicer ID normalization; see [`normalization.toml`](https://github.com/fermyon/spin/pull/1780/files#diff-e2d635e7a284960e906680aa98a7b855366cc1b9fab2254e89db606a8ad1961a) (and adjacent `normalization.toml.norm`).

itowlson:
@lann Thanks.  That feels more helpful.  I have ideas for further heuristics on identifying unnamed components, but the sort of errors I was raising were the result of malicious tinkering - probably better to wait and see if people ever run into them in the wild, and iterate on them at that point.

itowlson (approved):

PR #1784, opened 2023-09-20
Reconfigure `spin watch` on manifest change

itowlson:
This went through many extremely untidy stages and I surely need to read over it again (and do a bunch more testing), but it seems to work correctly.  Yay!  Now let us never speak of it again

### DIFF
--- a/src/commands/watch.rs
+++ b/src/commands/watch.rs
@@ -1,17 +1,13 @@
 use std::{
-    convert::Infallible,
     path::{Path, PathBuf},
     sync::Arc,
     time::Duration,
 };
 
-use anyhow::{bail, Context, Result};
+use anyhow::{Context, Result};
 use clap::Parser;
 use itertools::Itertools;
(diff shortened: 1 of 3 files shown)


calebschoepp:
I will take a look at this when I'm back from Peru in a week and a half.

calebschoepp (approved):
Great work! I tested it out locally and it worked for me.

Nothing blocking on the code front. However I do think it would be helpful if you add some more comment "documentation" to some of the new constructs like `ReconfigurableWatcher`, `RuntimeConfigFactory` etc. The design wasn't immediately obvious to me and the comments you've previously added really helped.

itowlson:
Good call.  There are a lot of moving parts and the names don't express well how they fit together.  Thank you for taking the time to review and to test!  (And I hope Peru was great!)

calebschoepp (approved):
Unless you want someone else's eyes on this I think it is ready to ship 🚢

PR #1790, opened 2023-09-20
Refactor llm and add http compute option

karthik2804:
This PR refactors the `llm` host component to make it easier to add alternate implementations. This PR also introduces an alternate implementation where the LLM compute workload is moved to a remote server through HTTP.  This allows for a better local development experience by enabling access to AI-grade GPUs. 

The runtime-config looks like the following 
```
[llm_compute]
type = "remote_http"
url = "<server URL>"
auth_token = "Auth token"
```

The remote server needs to implement 2 endpoints - `/infer` and `/embed`.  Both the endpoints consume the `auth_token` through the `Authorization: Bearer <token>` header. 

Both the end points take are of `POST` method. The response and request bodies contain JSON of the following types. 

**Infer**
```
// Request
{
interface InferenceParams {
  model: string;
  prompt: string;
  options: {
    maxToken: u32,
    repeatPenalty: f32,
    repeatPenaltyLastNTokenCount: u32,
    temperature: f32,
    topK: u32,
    topP: f32,
  };
}
}

// Response
{
  text: string, 
  usage: {
    PromptTokenCount: u32,
    GeneratedTokenCount: u32
}
```

**embed**
```
// Request
{
  model: string, 
  input: Vec<string>
}

// Response
{
  embeddings: Vec<Vec<f32>,
   usage: {
    PromptTokenCount: u32
  }
}
```

One such implementation for it is [spin-cloud-gpu](https://github.com/fermyon/spin-cloud-gpu). It is a plugin that deploys an app to the cloud and generates an auth-token that can be used in the runtime config.


crates/llm-local/src/lib.rs
```diff
@@ -0,0 +1,395 @@
...
+        }
+        let model = self.embeddings_model(model).await?;
+        generate_embeddings(data, model).await.map_err(|e| {
+            wasi_llm::Error::RuntimeError(format!("Error occurred generating embeddings: {e}"))
+        })
+    }
+}
+
+impl LocalLlmEngine {
+    pub async fn init(&mut self) {
```
itowlson, line 121: I don't quite understand why this is needed as a separate step from `new`?

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+#[derive(Clone)]
+pub struct RemoteHttpLlmEngine {
+    auth_token: String,
+    url: Url,
+    client: Option<Client>,
+    allowed_models: HashSet<String>,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all(serialize = "camelCase"))]
```
itowlson, line 22: The PR description had some of the fields looking a bit non-camel-cased e.g. `repeatPenalty_last_n_token_count` or `PromptTokenCount`.  I assume from the serde annotations that those were typos and that the interface is consistent?

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+
+#[derive(Deserialize)]
+struct EmbeddingResponseBody {
+    embeddings: Vec<Vec<f32>>,
+    usage: EmbeddingUsage,
+}
+
+#[async_trait]
+impl LlmEngine for RemoteHttpLlmEngine {
+    fn add_allowed_models(&mut self, models: HashSet<String>) {
```
itowlson, line 58: Seems like this is "set" rather than an "add" operation - if I call it twice it is not additive right?

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+        self.allowed_models = models;
+    }
+
+    async fn infer(
+        &mut self,
+        model: wasi_llm::InferencingModel,
+        prompt: String,
+        params: InferencingParams,
+    ) -> Result<wasi_llm::InferencingResult, wasi_llm::Error> {
+        let client = self.client.get_or_insert_with(Default::default);
```
itowlson, line 68: This doesn't seem to check `self.allowed_models`? (which could result in an inference working in dev but not when deployed)

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+            Err(err) => Err(wasi_llm::Error::RuntimeError(err.to_string())),
+        }
+    }
+
+    async fn generate_embeddings(
+        &mut self,
+        model: wasi_llm::EmbeddingModel,
+        data: Vec<String>,
+    ) -> Result<wasi_llm::EmbeddingsResult, wasi_llm::Error> {
+        let client = self.client.get_or_insert_with(Default::default);
```
itowlson, line 119: check `allowed_models`?

crates/trigger/src/runtime_config/llm.rs
```diff
@@ -1,13 +1,39 @@
...
+            let path = runtime_config
+                .state_dir()
+                .unwrap_or_default()
+                .join("ai-models");
+            let mut engine = LocalLlmEngine::new(path, use_gpu);
+            engine.init().await;
+            spin_llm::LlmComponent::new(move || Box::new(engine.clone()))
+        }
+        LlmComputeOpts::RemoteHttp(config) => {
+            println!("Using remote compute for LLMs");
```
itowlson, line 20: Could we remove this line please?  The Spin prologue is already long enough that nobody will ever read it... prologue items really only have value when we are telling people where to find a thing that they might want to look at outside of Spin (like the logs or the database), and this doesn't meet that criterion.  You could put it as an `info!` or `debug!` trace if you feel it's useful?

itowlson (commented):
Thanks for this!  It looks solid to me though I'm no expert for sure!

itowlson:
There is something a bit structurally odd about implementing a feature in the Spin core to support a plugin.  I think it's unavoidable for now but it would be a good one to revisit once we have WASI2 host components.

crates/llm-remote-http/src/lib.rs
karthik2804: Good catch - will fix it!

crates/llm-remote-http/src/lib.rs
karthik2804: Good catch - will fix it!

crates/llm-remote-http/src/lib.rs
karthik2804: Agree - will rename.

crates/llm-remote-http/src/lib.rs
karthik2804: Yea - those are typos. The JSON response is `camelCased`.

crates/llm-local/src/lib.rs
karthik2804: It is a step to warm up the cache but yes can move it into `new` if that is a better solution.

crates/trigger/src/runtime_config/llm.rs
karthik2804: done

crates/llm-local/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+    inferencing_models: HashMap<(String, bool), Arc<dyn llm::Model>>,
+    embeddings_models: HashMap<String, Arc<(tokenizers::Tokenizer, BertModel)>>,
+}
+
+#[async_trait]
+impl LlmEngine for LocalLlmEngine {
+    fn set_allowed_models(&mut self, models: HashSet<String>) {
+        self.allowed_models = models;
+    }
+    async fn infer(
```
rylev, line 43: nit: can you add spaces between methods?

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+    top_p: f32,
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all(deserialize = "camelCase"))]
+struct InferUsage {
+    prompt_token_count: u32,
+    generated_token_count: u32,
+}
+#[derive(Deserialize)]
```
rylev, line 38: nit:
```suggestion
}

#[derive(Deserialize)]
```

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    async fn infer(
+        &mut self,
+        model: wasi_llm::InferencingModel,
+        prompt: String,
+        params: InferencingParams,
+    ) -> Result<wasi_llm::InferencingResult, wasi_llm::Error> {
+        if !self.allowed_models.contains(&model) {
+            return Err(access_denied_error(&model));
+        }
```
rylev, line 70: Can we move this check outside of the specific implementations and into `LlmDispatch`? It would be nice if we didn't have to repeat this check across all implementors.

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,186 @@
+use anyhow::Result;
+use reqwest::{
+    header::{HeaderMap, HeaderValue},
+    Client, Url,
+};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use spin_core::async_trait;
+use spin_llm::{access_denied_error, LlmEngine};
+use spin_world::llm::{self as wasi_llm, InferencingParams};
```
rylev, line 10: nit: why not just also prefix `InferencingParams` like everything else from `wasi_llm`?. In fact, since there's nothing else called `llm` here we could probably just import this as `llm`, no?

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+                    usage: wasi_llm::InferencingUsage {
+                        prompt_token_count: val.usage.prompt_token_count,
+                        generated_token_count: val.usage.generated_token_count,
+                    },
+                }),
+                Err(err) => Err(wasi_llm::Error::RuntimeError(format!(
+                    "Failed to deserialize response: {err}"
+                ))),
+            },
+            Err(err) => Err(wasi_llm::Error::RuntimeError(err.to_string())),
```
rylev, line 120: We might want to provide a bit of additional context to this error since it might not be clear that this is from the call to `POST /infer`

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+            .request(
+                http::Method::POST,
+                self.url.join("/infer").map_err(|_| {
+                    wasi_llm::Error::RuntimeError("Failed to create URL".to_string())
+                })?,
+            )
+            .headers(headers)
+            .body(body)
+            .send()
+            .await;
```
rylev, line 105: I think we should use `map_err` plus `?` here to avoid the double nested `match` expression below.

crates/llm-remote-http/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+impl RemoteHttpLlmEngine {
+    pub fn new(url: Url, auth_token: String) -> Self {
+        RemoteHttpLlmEngine {
+            url,
+            auth_token,
+            allowed_models: Default::default(),
+            client: None,
```
rylev, line 183: Why is client an optional value?

rylev (approved):
Looking good! I have some nits, but this turned out really nice!

crates/llm-remote-http/src/lib.rs
karthik2804: I was following the pattern in  `outbound-http`

crates/llm-remote-http/src/lib.rs
karthik2804: I kept it consistent across the codebase, I will change it to have the `wasi_llm` prefix.

crates/llm-local/src/lib.rs
rylev: I would be in favor of that.

crates/llm-remote-http/src/lib.rs
karthik2804: done!

crates/llm-remote-http/src/lib.rs
karthik2804: Done

crates/llm-remote-http/src/lib.rs
```diff
@@ -102,22 +94,20 @@ impl LlmEngine for RemoteHttpLlmEngine {
...
-                        generated_token_count: val.usage.generated_token_count,
-                    },
-                }),
-                Err(err) => Err(wasi_llm::Error::RuntimeError(format!(
-                    "Failed to deserialize response: {err}"
-                ))),
-            },
-            Err(err) => Err(wasi_llm::Error::RuntimeError(err.to_string())),
+            .await
+            .map_err(|err| wasi_llm::Error::RuntimeError(err.to_string()))?;
```
rylev, line 98: Still missing the context from this error. For example if the error is "HTTP 500" the user will have no idea what call resulted in the 500. We should do the same kind of context adding as we do with the deserialization response below.

crates/llm-remote-http/src/lib.rs
```diff
@@ -155,21 +142,19 @@ impl LlmEngine for RemoteHttpLlmEngine {
...
-                        prompt_token_count: val.usage.prompt_token_count,
-                    },
-                }),
-                Err(err) => Err(wasi_llm::Error::RuntimeError(format!(
-                    "Failed to deserialize response: {err}"
-                ))),
-            },
-            Err(err) => Err(wasi_llm::Error::RuntimeError(err.to_string())),
+            .await
+            .map_err(|err| wasi_llm::Error::RuntimeError(err.to_string()))?;
```
rylev, line 146: Same as above

crates/llm-remote-http/src/lib.rs
karthik2804: fixed it.

crates/llm-remote-http/src/lib.rs
karthik2804: done.

rylev (approved):

rylev:
Going to merge this. We haven't had any objections, and the implementation looks good!

PR #1793, opened 2023-09-22
syntax fix to run only on linux/amd64

rajatjindal:
This seems like a weird syntax thing with GitHub actions

`runner.os == 'linux' && matrix.config.arch == 'amd64' && github.repository_owner == 'fermyon'` works but `runner.os == 'linux' && matrix.config.arch == 'amd64' && ${{ github.repository_owner == 'fermyon' }}`  does not.

tried it here: https://github.com/rajatjindal/test-action/blob/main/.github/workflows/arch2.yaml


itowlson (approved):
Weird!  Thanks for figuring this out!

PR #1795, opened 2023-09-22
Allow other ai models besides while known ones

rylev:
This allows other AI models besides the two well known models: llama2-chat and codellama. 

In order to use other models, the user must first place their model in the right location in the model registry (i.e., in `.spin/ai-models`). The well known models continue to work if the are placed at the top-level of the registry. However, any other model can be placed in a subdirectory named after their model architecture (e.g., llama based models go in `.spin/ai-models/llama`, bloom based models in `.spin/ai-models/bloom`, etc.). The model binary must be named after the model itself. So if you're using a llama based model called "llama2-norwegian" it should be at the path `.spin/ai-models/llama/llama2-norwegian`. 

Spin first checks for well known models and then traverses the model registry for the model file. 

All models must be in the GGML format, and the supported models can be found at [this link](https://github.com/rustformers/llm/blob/84800b02a7a96f62c0c9c03a38c36cb23bf4b2ec/crates/llm/src/lib.rs#L174-L181).

**Note**: this is currently only for inferencing. I would imagine that the exact same scheme could be used for embeddings, but I simply ran out of time to implement it today. If everyone is happy with this approach, I will add the same support for embeddings models. 

Note also that some SDKs still have an allow list of models that are supported. Those will need to be updated as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3174,7 +3174,12 @@ version = "0.2.0-dev"
 source = "git+https://github.com/rustformers/llm?rev=2f6ffd4435799ceaa1d1bcb5a8790e5b3e0c5663#2f6ffd4435799ceaa1d1bcb5a8790e5b3e0c5663"
 dependencies = [
  "llm-base",
+ "llm-bloom",
+ "llm-gpt2",
+ "llm-gptj",
+ "llm-gptneox",
  "llm-llama",
+ "llm-mpt",
  "serde",
  "tracing",
 ]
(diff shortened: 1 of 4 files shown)


karthik2804:
@rylev Looks like the features need to be enabled in the `Cargo.toml` for the other architectures.

I was testing out GPT2 and had to add the feature. 

```toml
llm = { git = "https://github.com/rustformers/llm", rev = "2f6ffd4435799ceaa1d1bcb5a8790e5b3e0c5663", features = [
    "tokenizers-remote",
    "llama",
    "gpt2",
], default-features = false }
```

rylev:
@karthik2804 I pushed a new commit that turns on support in the `llm` crate for all the known models.

karthik2804:
@rylev Now it works! I did run into issues with the `llm-metal` features turned on. 

```
GGML_ASSERT: /Users/karthik_ganeshram/Work/fermyon/spin/target/release/build/ggml-sys-c4c1041cf5cbd8ac/out/ggml-metal.m:905: false && "not implemented"
```

rylev:
@karthik2804 I guess that ggml on metal is just not supported by the llm crate. Do you think we should still merge this PR? I think we should give it a try.

karthik2804:
@rylev I agree - having the ability to run other models will be useful even if we cannot run them using the `metal` builds.

karthik2804 (approved):

Issue #1737, opened 2023-09-06
Add Azure KeyVault support to Spin


ThorstenHans:
We've support to load secrets from HashiCorp Vault for quite some time now. I read through the corresponding parts of the source code yesterday evening and thought that adding support for Azure Key Vault shouldn't be that big of an effort.

There are several crates that wrap AuthN and necessary Key Vault API:

- [https://crates.io/crates/azure_security_keyvault](https://crates.io/crates/azure_security_keyvault) and [https://crates.io/crates/azure_identity](https://crates.io/crates/azure_identity)
- [https://crates.io/crates/azure-key-vault-client](https://crates.io/crates/azure-key-vault-client)

Considering both approaches, I would go with the crates mentioned first. Athough, they're not official SDKs provided by Microsoft, those crates receive regular updates and are owned by [Cameron Taggart](https://crates.io/users/cataggar) who maintians a bunch of Azure crates...

IMO it would be a great addition to allow users loading sensitive configuration data from different sources. 

Any thoughts regarding this one?

vdice:
I think adding another runtime config provider alongside Vault (and the default env provider) would be great.  I imagine there willl be other users keen on integrating Azure KeyVault for this purpose.  👍

itowlson:
Fixed via https://github.com/spinframework/spin/pull/2724

PR #1796, opened 2023-09-22
Add Azure Key Vault config provider

ThorstenHans:
This PR adds **Azure Key Vault** as config provider to `spin` as suggest in #1737

## Authentication

Authentication is implemented using `ClientCredentialFlow` leveraging `clientId` and `clientSecret`  of an Azure Service Principal (SP).

The SP must have the `Key Vault Secrets User` role assigned on the scope of the desired Azure Key Vault instance. Additionally, RBAC must be enabled on the Azure Key Vault instance.

## Loading Secret Values

Although Azure Key Vault supports multiple versions per secret, this implementation loads the latest version of a particular secret. 

## Runtime Configuration File

Once necessary SP and Azure Key Vault have been provisioned and Role Assignment is in place, users must provide a corresponding runtime config file:

```toml
[[config_provider]]
type = "azure_key_vault"
url = "https://mysecrets.vault.azure.net/"
client_id = "00000000-0000-0000-0000-000000000000"
client_secret = "SomeSauce"
tenant_id = "00000000-0000-0000-0000-000000000000"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,6 +214,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-lock"
+version = "2.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "287272293e9d8c41773cec55e365490fe034813a2f172f502d6ddcf75b2f582b"
+dependencies = [
+ "event-listener",
+]
+
+[[package]]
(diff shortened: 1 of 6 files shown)


ThorstenHans:
@melissaklein24 I'll close this one and will come up with a new PR

PR #1803, opened 2023-09-25
http: Make `base = "/"` the default

lann:


radu-matei (approved):
Yeaah, this should have been the default.
LGTM!

fibonacci1729 (approved):

crates/manifest/src/lib.rs
```diff
@@ -205,6 +205,7 @@ impl From<ApplicationTrigger> for ApplicationTriggerSerialised {
 
 /// HTTP trigger configuration.
 #[derive(Clone, Debug, Deserialize, PartialEq, Eq, Serialize)]
+#[serde(default)]
```
michelleN, line 208: what does this change do?

michelleN (approved):

crates/manifest/src/lib.rs
lann: It uses `Self::default` to set default values for any missing fields.

PR #1804, opened 2023-09-25
sip: component imports

fibonacci1729:
[Rendered](https://github.com/fibonacci1729/spin/blob/component-imports-sip/docs/content/sips/016-component-imports.md)

### DIFF
--- a/docs/content/sips/016-component-imports.md
+++ b/docs/content/sips/016-component-imports.md
@@ -1,0 +1,58 @@
+title = "SIP 016 - Spin Component Imports" 
+template = "main" 
+date = "2023-09-22T16:27:43Z"
+---
+
+Summary: A proposal to introduce component imports to the Spin 2.0 manifest.
+
+Owner: <brian.hardock@fermyon.com>
+
+Created: Sep 19, 2023
+
+## Background / Goals
+The Spin 2.0 framework will support executing [preview2 WebAssembly components](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#component-model-explainer). These components will include `import` and `export` statements. This manifest proposal is to enable Spin developers the ability to expressly configure how to satisfy the imports of their components. Furthermore, developers should be enabled to use the exports of a component to satisfy the imports of another.
+
+This SIP is concerned with what's known as "static composition" meaning that a new component binary is produced via composing components together. Static composition here is defined as the process of instantiating components with the exports of other components effectively erasing the imports from the composed component.
+
+## Proposal
+
+### Changes to Spin Manifest 
+To support component imports, a new section will be added to the Spin 2.0 manifest design, `[component.<component-id>.import.<import-id>]`. The following sections describe the various options for specifying how to satisfy imports.
+
+> NOTE: The format for each `import` is required to be a kebab-cased name (e.g. `foo-bar`) or interface id (e.g. `foo:bar/baz`) as described by the [Component Model Explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions).
+
+### Reference by ID
+Imports can be satisfied by associating an import with a component ID which links to a component defined elsewhere in the manifest.
+
+```toml
+[component.foobar.import."foo"]
+source = "other"
+
+[component.other]
+source = { path = "bar.wasm" }
+```
+
+### Reference components in a registry
+With a future component registry (i.e. `warg`), developers will be able to satisfy their
+component's imports using registry references:
+
+```toml
+[component.foobar.import."foo"]
+source = { registry = "bytecodealliance.org:foobar-dep", version = "1.2.3" }
+```
+
+### Satisfy imports using the exports of another component
+Below is an example of a component's import declaration. Developers can use the `export` of a component specified by `source` to satisfy the named `import` of the named component.
+
+```toml
+[component.foobar.import."foo"]
+source = "path/to/component.wasm"
+export = "bar"
+```
+
+This example demonstrates how to satisfy the `foobar` component's import `foo` using the export named `bar` found in the component at `"path/to/component.wasm"`.
+
+> NOTE: The format for each `export` is required to be a kebab-cased name (e.g. `foo-bar`) or interface id (e.g. `foo:bar/baz`) as described by the [Component Model Explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions).
+
+## Future design options
+TODO


docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Reference by ID
+Imports can be satisfied by associating an import with a component ID which links to a component defined elsewhere in the manifest.
+
+```toml
+[component.foobar.import."foo"]
+source = "other"
+
+[component.other]
+source = { path = "bar.wasm" }
```
lann, line 30: What would happen if I specified `files` here?

fibonacci1729: I was imagining using [wasi-virt](https://github.com/bytecodealliance/WASI-Virt); similarly for `environment`. I haven't experimented with this just yet so not entirely sure how viable this is.

Do you have any thoughts?

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+### Changes to Spin Manifest 
+To support component imports, a new section will be added to the Spin 2.0 manifest design, `[component.<component-id>.import.<import-id>]`. The following sections describe the various options for specifying how to satisfy imports.
+
+> NOTE: `import-id` refers to one of the several optional for naming imports as per the [Component Model Explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions).
+
+### Reference by ID
+Imports can be satisfied by associating an import with a component ID which links to a component defined elsewhere in the manifest.
+
+```toml
+[component.foobar.import."foo"]
```
kate-goldenring, line 26: Would it be worth adding an example world with a `foo` interface to the SIP to complete the example?

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+Summary: A proposal to introduce component imports to the Spin 2.0 manifest.
+
+Owner: <brian.hardock@fermyon.com>
+
+Created: Sep 19, 2023
+
+## Background / Goals
+The Spin 2.0 framework will support executing preview2 WebAssembly components. These components will include `import` and `export` statements. This manifest proposal is to enable Spin developers the ability to expressly configure how to satisfy the imports of their components. Furthermore, developers should be enabled to use the exports of a component to satisfy the imports of another.
```
michelleN, line 13: Can we add a link to components for those unfamiliar?

How do we concisely represent where the component work is now in Spin and what needs to be done before we can support component imports in the manifest?

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+
+```toml
+[component.foobar.import."foo"]
+source = "path/to/component.wasm"
+export = "bar"
+```
+
+This example demonstrates how to use an export name `bar` from `"path/to/component.wasm"` to satisfy component `"foobar"`'s import named `"foo"`.
+
+> NOTE: The format for each `export` is required to be a kebab-cased name or interface id (e.g. `foo:bar/baz`) as described by the [Component Model Explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions).
```
itowlson, line 53: Is there anything to say about failure modes here?  What sort of things can go wrong during the import fulfilment process?

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+[component.foobar.import."foo"]
+source = { registry = "bytecodealliance.org:foobar-dep", version = "1.2.3" }
+```
+
+### Satisfy import by projecting into a component
+By specifying an export on a component import declaration, developers can project into the component specified by `source` to use a named export to satisfy the import.
+
+```toml
+[component.foobar.import."foo"]
+source = "path/to/component.wasm"
```
itowlson, line 47: Same question about whether the `source` here might have unsatisfied imports.

docs/content/sips/016-component-imports.md
itowlson, line 53: What is the output of resolution e.g. if I have some import sections and do a `spin registry push`, what goes into the registry?  Is it unfulfilled components with the same import directives, or is it fully linked and self-contained components (guaranteed to be the same as what ran during `spin up`)?

itowlson (changes requested):

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+To support component imports, a new section will be added to the Spin 2.0 manifest design, `[component.<component-id>.import.<import-id>]`. The following sections describe the various options for specifying how to satisfy imports.
+
+> NOTE: `import-id` refers to one of the several optional for naming imports as per the [Component Model Explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions).
+
+### Reference by ID
+Imports can be satisfied by associating an import with a component ID which links to a component defined elsewhere in the manifest.
+
+```toml
+[component.foobar.import."foo"]
+source = "other"
```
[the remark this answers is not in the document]
fibonacci1729: The latter case; as far as i'm aware there hasn't been any discussion around supporting dangling `[component]` entries (not associated with a trigger).

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+### Reference components in a registry
+With a future component registry (i.e. `warg`), developers will be able to satisfy their
+component's imports using registry references:
+
+```toml
+[component.foobar.import."foo"]
+source = { registry = "bytecodealliance.org:foobar-dep", version = "1.2.3" }
+```
+
+### Satisfy import by projecting into a component
```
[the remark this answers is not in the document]
fibonacci1729: That's correct; "projecting into" is meant to convey that components are graphs and you can walk a "path" into a component (or one of its children) to get to an export. In any case, I'll reword this.

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Background / Goals
+The Spin 2.0 framework will support executing preview2 WebAssembly components. These components will include `import` and `export` statements. This manifest proposal is to enable Spin developers the ability to expressly configure how to satisfy the imports of their components. Furthermore, developers should be enabled to use the exports of a component to satisfy the imports of another.
+
+## Proposal
+
+### Changes to Spin Manifest 
+To support component imports, a new section will be added to the Spin 2.0 manifest design, `[component.<component-id>.import.<import-id>]`. The following sections describe the various options for specifying how to satisfy imports.
+
+> NOTE: `import-id` refers to one of the several optional for naming imports as per the [Component Model Explainer](https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md#import-and-export-definitions).
```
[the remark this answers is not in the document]
fibonacci1729: I agree, great point! For Spin (re: `spin.toml`) I think it's safe/fair to assume (for now) a kebab-name or interface id (similar to exports). As in, we can assume that for a Spin component associated with a trigger that the Wasm component is "unlocked".

docs/content/sips/016-component-imports.md
itowlson: Related on failure modes: when would an import fulfilment failure be detected and reported?  During "load and lock" time, before the trigger process gets launched?

fibonacci1729: I think there are two primary failure modes:

1. The `import`/`export` specified doesn't exist in the component
2. The `export` specified is of the wrong type expected by the `import`

I was thinking (re: when can these happen) that compositions are constructed when the trigger runs but it would be fantastic if we could report during load and lock. I think once we have agreement on the general experience, this will be the next question to  explore / lock down.

docs/content/sips/016-component-imports.md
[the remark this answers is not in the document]
thangchung: Indeed. When reading through this `foo`, `foobar`, and `other` might make me confused to fully understand it. That would be great if we had a meaningful name for these terms.

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+
+Created: Sep 19, 2023
+
+## Background / Goals
+The Spin 2.0 framework will support executing preview2 WebAssembly components. These components will include `import` and `export` statements. This manifest proposal is to enable Spin developers the ability to expressly configure how to satisfy the imports of their components. Furthermore, developers should be enabled to use the exports of a component to satisfy the imports of another.
+
+## Proposal
+
+### Changes to Spin Manifest 
+To support component imports, a new section will be added to the Spin 2.0 manifest design, `[component.<component-id>.import.<import-id>]`. The following sections describe the various options for specifying how to satisfy imports.
```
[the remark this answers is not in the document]
rylev: For simple use cases this could be achieved through static component composition (i.e., taking two components and making them 1 component). However ,there are pros and cons to this approach as opposed to satisfying imports by instantiating the other component and using the instantiated instance in the `Host` implementation (what I call "dynamic composition"). 

**Pros of static composition (compared to "dynamic composition")**
* **It's more performant**: there's a penalty paid for exciting the guest and entering the host. Static composition avoids this cost since both components are now the same 1 guest component instead of two separate guest instances.
* **It can be shared**: static composition creates a single new component that can be shared through a component registry and even run in other environments. Dynamic composition requires a Spin runtime to hook everything up at runtime. 

**Cons of static composition**
* **it's less flexible**: dynamic composition allows the user to compose components in different ways. As an off the cuff example: imagine a "request sampling" component that we wouldn't to intercept every 10th request to do some sort of sampled metrics but otherwise not touch the other 9 requests. That's not possible with static composition but would be with a host aware dyanmic composition. 

I think in general we probably don't want the user to know or care about which approach we take. Exposing composed components that the user can use outside of `spin up` should be a separate feature IMO.

docs/content/sips/016-component-imports.md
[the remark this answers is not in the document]
rylev: This concerns me - If we're choosing to only support components that conform to the Spin world for simplicity sake, it would be good for us to have a solid plan of what it would take to support components that don't conform to that world (and instead only expose the export that satisfies the required import).

docs/content/sips/016-component-imports.md
rylev: If we don't go down the route of static composition then each component could have their own instance of wasi.

docs/content/sips/016-component-imports.md
```diff
@@ -0,0 +1,53 @@
...
+source = { path = "bar.wasm" }
+```
+
+### Reference components in a registry
+With a future component registry (i.e. `warg`), developers will be able to satisfy their
+component's imports using registry references:
+
+```toml
+[component.foobar.import."foo"]
+source = { registry = "bytecodealliance.org:foobar-dep", version = "1.2.3" }
```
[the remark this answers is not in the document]
rylev: Indeed. This makes me wonder if what we want instead is for the user to declare "dangling" components (i.e., components that don't have a trigger associated with them) and those components can reference other dangling components that expose some exports that satisfy the component's imports. We can then build up a graph of components. 

For example: imagine a use case of a pass through key value logger: every time the user uses the key-value store, the key and value are logged to some logging interface.

```toml
[[component]]
id = "key_value_logger"
source = { path = "key_value_logger.wasm" }
import.logger = logger

[[component]]
id = "logger"
source = { path = "logger.wasm" }

[[component]]
id = "my-app"
[component.trigger]
route = "/hello"
component.import."fermyon:spin/key-value" = "key_value_logger"
```

docs/content/sips/016-component-imports.md
lann: > each component could have their own instance of wasi.

They can't currently; `wasmtime-wasi` (and bindgen in general?) doesn't support linking to anything other than root instances. That is something that could be fixed though.

docs/content/sips/016-component-imports.md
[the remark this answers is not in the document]
lann: The v2 manifest supports components without a trigger.

docs/content/sips/016-component-imports.md
lann: Today it wouldn't be reported until the trigger runs, but that isn't an inherent limitation.

docs/content/sips/016-component-imports.md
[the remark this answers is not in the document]
fibonacci1729: Thanks for the feedback! I'll update the examples to be a little more intuitive (naming is hard).

docs/content/sips/016-component-imports.md
rylev: If you do dynamic linking, each instance can be backed by a completely separate store so yes it can have different wasi instances. This is how wepl works.

lann: Separate stores would mean we can't use resources, right?

docs/content/sips/016-component-imports.md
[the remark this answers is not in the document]
fibonacci1729: @itowlson update the language here, let me know what you think.

docs/content/sips/016-component-imports.md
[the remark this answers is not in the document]
fibonacci1729: @itowlson Updated the language to constrain the types of `import-id`s supported in this UX. Let me know how it sounds to you, thanks!

fibonacci1729:
Closing in favor of #2543

PR #1811, opened 2023-09-26
fix(registry): use rpassword when taking registry password from stdin

radu-matei:
cc @michelleN


itowlson:
Am I right in thinking this is addressing the problem identified in #1191?

PR #1813, opened 2023-09-26
If command matches non-installed plugin, offer install

itowlson:
When the user invokes a command which is not recognised, but does match a plugin, this feature offers to install that plugin:

```
# User can of course nope out
$ spin check-for-update
Warning: `check-for-update` is not a known Spin command, but there is a plugin with that name.
Would you like to install and run it now? no

# YOLO
$ spin check-for-update
Warning: `check-for-update` is not a known Spin command, but there is a plugin with that name.
Would you like to install and run it now? yes
Plugin 'check-for-update' was installed successfully!

Description:
        Command to verify if you're using latest spin CLI or not.

Homepage:
        https://github.com/ThorstenHans/spin-plugin-check-for-update

⠉⠙ Checking for latest spin CLI version...
# ^^ or whatever the plugin output is
```

(I am now wondering if I should print the description as part of the prompt.)

Conceptually, this PR just adds another case to the "not found" match arm in `execute_external_subcommand` - there's really only a few lines of new code.  But that function was getting so gnarly that I chose to break out the "make sure the plugin is there" stuff into a separate function... which was _still_ gnarly, and anyway long story short here we are with a diff the length of your arm and yep, _even now_ one of the functions is not free of the gnarl.

### DIFF
--- a/src/commands/external.rs
+++ b/src/commands/external.rs
@@ -6,6 +6,7 @@ use spin_plugins::{
     badger::BadgerChecker, error::Error as PluginError, manifest::warn_unsupported_version,
     PluginStore,
 };
+use std::io::{stderr, IsTerminal};
 use std::{collections::HashMap, env, process};
 use tokio::process::Command;
 use tracing::log;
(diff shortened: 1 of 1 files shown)


src/commands/external.rs
```diff
@@ -123,6 +88,98 @@ pub async fn execute_external_subcommand(
...
+    app: clap::App<'_>,
+    override_compatibility_check: bool,
+) -> anyhow::Result<Option<String>> {
+    let plugin_version = match plugin_store.read_plugin_manifest(plugin_name) {
+        Ok(manifest) => {
+            if let Err(e) =
+                warn_unsupported_version(&manifest, SPIN_VERSION, override_compatibility_check)
+            {
+                eprintln!("{e}");
+                // TODO: consider running the update checked?
```
vdice, line 103: Hmmm, that could be handy.  But potentially also another spot to check in with user 'no plugin manifest is found for foo, shall we update your local plugins list'?  Though I think just proceeded seems like a sane default eg 'no plugin manifest is found for foo, pulling the latest available plugins from fermyon/spin-plugins' (or other).

src/commands/external.rs
```diff
@@ -123,6 +88,98 @@ pub async fn execute_external_subcommand(
...
+    Ok(plugin_version)
+}
+
+async fn consider_install(
+    plugin_name: &str,
+    plugin_store: &PluginStore,
+    app: clap::App<'_>,
+    e: &spin_plugins::error::NotFoundError,
+) -> anyhow::Result<Option<String>> {
+    if predefined_externals()
```
vdice, line 122: A thought for a potential follow-up as I acquaint myself with this function.  I see it currently tracks `cloud`: https://github.com/fermyon/spin/blob/362c75d98e3ac2cb1b760b661fbcb18f0eced118/src/commands/external.rs#L39-L42

Should we add js2wasm and py2wasm per https://github.com/fermyon/developer/blob/main/downloads/install.sh#L148?

src/commands/external.rs
```diff
@@ -123,6 +88,98 @@ pub async fn execute_external_subcommand(
...
+    } else {
+        tracing::debug!("Tried to resolve {plugin_name} to plugin, got {e}");
+        terminal::error!("'{plugin_name}' is not a known Spin command. See spin --help.\n");
+        print_similar_commands(app, plugin_name);
+        process::exit(2);
+    }
+}
+
+fn offer_install(name: &str) -> anyhow::Result<bool, anyhow::Error> {
+    terminal::warn!("`{name}` is not a known Spin command, but there is a plugin with that name.");
```
vdice, line 155: Alluding to your thought in the PR description; perhaps printing the plugin description around here would be useful to the user.  Link to repo as well?  (Or getting to crowded at this point?)

kate-goldenring (approved):
LGTM! Thanks for breaking this code out more. One nit question

src/commands/external.rs
```diff
@@ -123,6 +88,98 @@ pub async fn execute_external_subcommand(
...
+                update().await?;
+            }
+            plugin_installer.run().await?;
+        }
+        Ok(None) // No update badgering needed if we just updated/installed it!
+    } else if stderr().is_terminal() && matches_catalogue_plugin(plugin_store, plugin_name) {
+        if offer_install(plugin_name)? {
+            let plugin_installer = installer_for(plugin_name);
+            plugin_installer.run().await?;
+            eprintln!();
```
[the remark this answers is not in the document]
itowlson: The reason for this was to create a visual separation between the "would you like to install it" prompt and the plugin output.  Without the blank line, I felt they ran together rather.  But this is very subjective for sure.

src/commands/external.rs
itowlson: We predefine `cloud` because it is the backing implementation for two hardwired commands, `spin deploy` and `spin login`.  This isn't the case for `js2wasm` and `py2wasm`.

src/commands/external.rs
itowlson: Yes, that was what was on my mind - give them a bit more info to make the decision.  But for now I think I will just merge this and we can iterate on the description thing separately.

PR #1816, opened 2023-09-27
Fix Postgres NULL attempting type conversion

### LINKED ISSUE #1808: WrongType error when inserting NULL into VARCHAR nullable column
**Spin version**: spin 1.5.0 (ca08dd9 2023-09-13); **NO** plugins installed

I'm trying to insert NULL into column with type VARCHAR, like this:

```rust
pg::query(
    ADDRESS,
    "INSERT INTO names (name) VALUES ($1)",
    &vec![ParameterValue::DbNull],
)?;
```

It doesn't work, I'm getting error:

```
Error { kind: ToSql(0), cause: Some(WrongType { postgres: Varchar, rust: "core::option::Option<i32>" }) }
```

It does not work for any type except INT4.

## sample app

database schema:

```sql
CREATE TABLE names (name VARCHAR);
```

spin component sources:

```rust
use anyhow::Result;
use spin_sdk::{
    http::{Request, Response},
    http_component,
    pg::{self, ParameterValue},
};

const ADDRESS: &str = "postgres://postgres@localhost:5432/testdb";

/// A simple Spin HTTP component.
#[http_component]
fn handle_test_pg2(_req: Request) -> Result<Response> {
    pg::query(
        ADDRESS,
        "INSERT INTO names (name) VALUES ($1) RETURNING name",
        &vec![ParameterValue::DbNull],
    )?;

    Ok(http::Response::builder()
        .status(200)
        .header("foo", "bar")
        .body(Some("Hello, Fermyon".into()))?)
}
```

itowlson:
Fixes #1808.

The underlying issue is that, while the Rust Postgres driver does map `Option::None` to database `NULL`, it still cares about the _type_ of the `None`.  This replaces the typeful `None` with a custom value that tells the driver to send it as NULL no matter what.

### DIFF
--- a/crates/outbound-pg/src/lib.rs
+++ b/crates/outbound-pg/src/lib.rs
@@ -103,8 +103,6 @@ impl postgres::Host for OutboundPg {
     }
 }
 
-const DB_NULL: Option<i32> = None;
-
 fn to_sql_parameter(value: &ParameterValue) -> anyhow::Result<&(dyn ToSql + Sync)> {
     match value {
         ParameterValue::Boolean(v) => Ok(v),
@@ -120,7 +118,7 @@ fn to_sql_parameter(value: &ParameterValue) -> anyhow::Result<&(dyn ToSql + Sync
         | ParameterValue::Uint64(_) => Err(anyhow!("Postgres does not support unsigned integers")),
         ParameterValue::Str(v) => Ok(v),
         ParameterValue::Binary(v) => Ok(v),
-        ParameterValue::DbNull => Ok(&DB_NULL),
+        ParameterValue::DbNull => Ok(&PgNull),
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/outbound-pg/src/lib.rs
```diff
@@ -285,3 +283,40 @@ where
         }
     });
 }
+
+struct PgNull;
```
rylev, line 287: Can you add doc comments to this type explaining exactly what you explain in the PR description?

crates/outbound-pg/src/lib.rs
```diff
@@ -285,3 +283,40 @@ where
...
+
+impl ToSql for PgNull {
+    fn to_sql(
+        &self,
+        _ty: &Type,
+        _out: &mut tokio_postgres::types::private::BytesMut,
+    ) -> std::result::Result<tokio_postgres::types::IsNull, Box<dyn std::error::Error + Sync + Send>>
```
rylev, line 294:
```suggestion
    ) -> Result<tokio_postgres::types::IsNull, Box<dyn std::error::Error + Sync + Send>>
```

rylev (changes requested):

rylev (approved):
🎉

Issue #1819, opened 2023-09-27
Spin Docs website on Cloud


vdice:
Update deployment of the Spin docs website (at `/docs`, a redirect app from spin.fermyon.dev to developer.fermyon.com/spin) to run on Fermyon Cloud.

itowlson:
@vdice is this fixed via #1820?

vdice:
@itowlson No, this issue is around overhauling the deployment of this website to run on Fermyon Cloud (today, it runs on a separate Nomad cluster directly).

itowlson:
@vdice The issue history says "vdice moved this from TODO to Done in  Spin Triage on Mar 14, 2024."  Does this mean it can be closed, or was that one of our occasional project board weirdnesses?

vdice:
@itowlson thank you for checking on this olde issue.  This can now be closed as the Spin documentation website moved to Fermyon Cloud as part of the CNCF migration earlier this year.

PR #1820, opened 2023-09-27
chore(docs): re-add base config to spin.toml

vdice:
We're still pinned to an earlier version of spin when [publishing (via bindle)](https://github.com/fermyon/spin/blob/main/deploy/publish-spin-docs.nomad) and deploying the spin docs 'website' (comprised of redirects from spin.fermyon.dev -> developer.fermyon.com/spin), so for now, we need the `base` config.

Fixes deployment error as seen recentlly in https://github.com/fermyon/spin/actions/runs/6317502067

Otherwise, the issue to move the website deployment to cloud is https://github.com/fermyon/spin/issues/1819 (and we can then use latest spin, cloud plugin, etc via https://github.com/fermyon/actions)

### DIFF
--- a/docs/spin.toml
+++ b/docs/spin.toml
@@ -3,7 +3,7 @@ name = "spin-docs"
 version = "0.1.0"
 description = "The Spin documentation website running on... Spin."
 authors = [ "Fermyon Engineering <engineering@fermyon.com>" ]
-trigger = { type = "http" }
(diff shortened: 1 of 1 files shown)


vdice:
(this PR updates a file in `docs` so many of the required checks aren't run, so will be stuck in pending; I can merge once an approval is in.)

itowlson (approved):

Issue #1781, opened 2023-09-19
Make `SPIN_OCI_SKIP_INLINED_FILES` the default


lann:
This was made opt-in via env var for testing and integration here: https://github.com/fermyon/spin/blob/631e32ff28477c01013a621ea41edb21ccabfdc8/crates/oci/src/client.rs#L78

We should make this the default "soonish" so that `spin registry push` will work for those registries that don't handle tiny layers.

lann:
@vdice FYI

PR #1821, opened 2023-09-27
feat(oci): inline content when applicable by default

vdice:
Closes https://github.com/fermyon/spin/issues/1781

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -100,12 +100,6 @@ impl Client {
         auth: RegistryAuth,
         reference: Reference,
     ) -> Result<Option<String>> {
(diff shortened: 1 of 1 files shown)


vdice:
As I think on it, is there any chance we'd still want users to be able to turn this 'feature' off (eg `SPIN_OCI_SKIP_INLINED_FILES=false`)?  Perhaps if there are other consumers of spin apps-as-represented-in-oci-object-form who may be disrupted by missing layers (when they've been inlined into the locked app manifest).

lann (approved):
I don't feel strongly either way.

vdice:
> I don't feel strongly either way.

We can always add support at a later date for turning off this default behavior, should it be desired.

PR #1822, opened 2023-09-28
Remove template custom filters feature

### LINKED ISSUE #1581: templates: Remove wit-bindgen-backport dependency
This is the last vestage of "old" wit-bindgen in spin; app usage of the old ABI is now handled by the spin-componentize adaptation.

After some discussion, I think one of the following should be done:

- Remove the custom filters feature; the only known usage of it is [this simple filter](https://github.com/fermyon/spin-dotnet-sdk/tree/main/tools/dotted-pascal-case-filter) which could be moved into a regular filter impl in spin-templates.

- Commit the wit-bindgen-generated source for the `custom-filter.wit` rather than using the wit-bindgen macro. Breaking changes in the wasmtime APIs that this generated code uses should be rare but would require manual fixes to that code. Guest code would be stuck with old wit-bindgen versions, but that isn't necessarily a major problem for this use.

- Convert custom filters to use the component model. I think as long as the custom filter "world" doesn't include WASI this might be fine.

itowlson:
Fixes #1581.

This removes the template custom filters feature, allowing removal of the wit-bindgen-backport dependency.  **This is a breaking change and should be merged only into a major version.**  Templates containing custom filters will be skipped and reported on installation; if a user has an old template which users the feature, it will fail when run.

I retained the `custom_filters` field in the TOML deserialisation so that we could provide a more meaningful error message than serde alone would do.  Its contents are ignored though.

For some reason the change set off a "large enum variant" lint even though it was making the large variant smaller, so there is an unrelated change in the PR for that.


lann (approved):

lann:
We've branched `v1.6`; this should be good to merge now.

PR #1823, opened 2023-09-28
Plugins search

### LINKED ISSUE #1716: `spin plugins list` does not list latest plugins
Currently, `spin plugins list` acts as a way to both list available and installed plugins, the former being the default and the latter enabled with `--available`. However, it does not fetch latest before listing, so it does not fully act as a "search". Would it be cleaner to instead break these into two subcommands:
- `spin plugins search`: Look up available plugins
- `spin plugins list`: List installed plugins
The current flow matches `apt` while adding `search` matches the `brew` CLI.

The confusion may lie in the fact that `spin plugin list` does not fetch the latest plugins -- this can only be done with `spin plugins update`, so in that way, it does not suffice for "searching for latest plugins". We could fetch the latest plugins on list (implicit update) or fetch the latest to a tempdir to appropriately display available. However, if we do the second approach, there may be confusion with a subsequent install as the plugin wont exist in the local plugins repository.

itowlson:
Fixes #1716.

* `spin plugins list` now fetches latest before printing list
* `spin plugins list --filter` for filtering list
* `spin plugins search <text>` as alias for `spin plugins list --filter <text>`

#1716 also proposes that `spin plugins list` be changed so that it always behaves per the current `--installed` option.  This is a breaking change, so would need to wait for a major version, so is not included in this PR; it would be good to capture that as a separate issue.

#1716 also includes a proposal to add tags to plugins, and to have filtering check them as well as the name; it would be good to capture that as a separate issue too (as it requires other design decisions about how tags are shown and used).

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -29,6 +29,9 @@ pub enum PluginCommands {
     /// List available or installed plugins.
     List(List),
 
+    /// Search for plugins by name.
+    Search(Search),
+
     /// Remove a plugin from your installation.
     Uninstall(Uninstall),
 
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):
🎉

kate-goldenring (approved):
Thanks @itowlson!

PR #1824, opened 2023-09-28
Remove Hippo dependency

itowlson:
I can't find anywhere these are used in the code.  Hippo _is_ alluded to in the CI dependencies (e.g. https://github.com/fermyon/spin/blob/268883366d57712de1a9cc33135238b46013fe86/.github/workflows/build.yml#L119) - I am not sure if that signals that the Hippo crates are still needed somewhere?  (Or if that's redundant too?)

### DIFF
--- a/.github/actions/spin-ci-dependencies/action.yml
+++ b/.github/actions/spin-ci-dependencies/action.yml
@@ -45,17 +45,6 @@ inputs:
     required: false
     type: string
 
-  hippo:
-    description: 'setup hippo'
-    required: false
-    default: 'false'
-    type: bool
-  hippo-version:
-    description: 'Hippo version to setup'
-    default: 'v0.19.0'
-    required: false
-    type: string
-
   golang:
     description: 'setup golang'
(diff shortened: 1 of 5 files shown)


bacongobbler (approved):

bacongobbler:
> Or if that's redundant too?

I think that's redundant. It's there to install the `hippo` CLI as needed for a CI run. That flag along with [the parameters in the github action](https://github.com/fermyon/spin/blob/268883366d57712de1a9cc33135238b46013fe86/.github/actions/spin-ci-dependencies/action.yml#L48-L57) should be safe to remove.

bacongobbler:
And here as well: https://github.com/fermyon/spin/blob/268883366d57712de1a9cc33135238b46013fe86/.github/actions/spin-ci-dependencies/action.yml#L127-L132

itowlson:
Thanks @bacongobbler - I've removed it from the various places it popped up in CI, as a separate commit so easy to revert if needed.

PR #1828, opened 2023-10-01
sqlite sdk throws "database full" error when too many connections are opened

rajatjindal:
It seems like we maintain the list of sqlite connections and it has limited size. Due to a bug in golang sdk, I was opening connection for each execute statement, and was hitting this issue.

The error msg `Database full` didn't help either and I was searching internet for what could be causing that. But was finally able to figure out the connection issue in the golang sdk.

This PR changes the error msg to make it more clear as to what the problem is. I am happy to change the error msg if there are other suggestions. thank you


rylev (approved):

itowlson:
@rajatjindal could you rebase this please so the tests can (hopefully) run?

PR #1830, opened 2023-10-02
Respect `SPIN_ALWAYS_BUILD` on registry push

### LINKED ISSUE #1826: Should `spin registry push` respect `SPIN_ALWAYS_BUILD`?
The Cloud plugin now (well, at next release) recognises `SPIN_ALWAYS_BUILD`.  I am guessing it would be surprising if a user had set that EV, did a registry push, and found they had pushed an old build.  How do others feel about it?

itowlson:
Fixes #1826.

(okay #1826 was a "should we" question but nobody said "are you mad" so I made an executive decision)

### DIFF
--- a/src/commands/registry.rs
+++ b/src/commands/registry.rs
@@ -49,6 +49,10 @@ pub struct Push {
     )]
     pub insecure: bool,
 
+    /// Specifies to perform `spin build` before pushing the application.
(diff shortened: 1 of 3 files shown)


fibonacci1729 (approved):
👍

PR #1833, opened 2023-10-02
Always print warning on blocked outbound HTTP

### LINKED ISSUE #1090: Better inner loop experience for blocked HTTP hosts
If Spin blocks an outbound HTTP request for not being on the allow list, it writes an `info` log, which nobody sees, and returns a DestinationNotAllowed error to the application, which the application probably buries in _its_ log.

However, in the inner dev loop, the likely cause of this is that the user has forgotten to add the host to the component's allow list.  It would be good if, in the local dev experience (in the terminal and running from a `spin.toml` file), we could provide users more specific and actionable feedback.  Some possible ideas:

* Bypass the normal logging and print a message directly to stderr, stating the component and host, and pointing them to `allowed_http_hosts` - natural but risks getting lost in spew
* Collect blocked hosts and print a list as an "after action report" when the user terminates Spin - bit more complicated and provides less feedback in the moment, but very easily accessible
* When a host is blocked, prompt the user if they want to allow it / add it to the allow list - very slick, but when I prototyped it I kept going "why isn't my app responding, oh, it's because it's waiting on a prompt over in my terminal" (and this was me _knowing_ I was testing the prompting behaviour!)
* A hybrid of the last two options, where you get the after action report and can go through it approving items - ooh nice, but remember at this point we have already had a Ctrl+C from the user and are meant to be tearing down the hated process
* SPINTERM

Any other offers?

itowlson:
Fixes #1090.

With the proposed text, it looks like this:

image

Get painting those bikesheds.

(Note that the host component doesn't know which component it's attached to, so we can't provide a component ID without more work (heaven forfend.)

### DIFF
--- a/crates/outbound-http/src/lib.rs
+++ b/crates/outbound-http/src/lib.rs
@@ -53,6 +53,9 @@ impl outbound_http::Host for OutboundHttp {
                 .map_err(|_| HttpError::RuntimeError)?
             {
                 tracing::log::info!("Destination not allowed: {}", req.uri);
+                terminal::warn!("A component tried to make a HTTP request to '{}'.", req.uri);
+                eprintln!("The component was configured not to allow requests to this host.");
+                eprintln!("To allow requests, add the host to `allowed_http_posts`.");
                 return Err(HttpError::DestinationNotAllowed);
             }
 
--- a/crates/outbound-http/src/lib.rs
+++ b/crates/outbound-http/src/lib.rs
@@ -53,9 +53,10 @@ impl outbound_http::Host for OutboundHttp {
                 .map_err(|_| HttpError::RuntimeError)?
             {
                 tracing::log::info!("Destination not allowed: {}", req.uri);
-                terminal::warn!("A component tried to make a HTTP request to '{}'.", req.uri);
-                eprintln!("The component was configured not to allow requests to this host.");
-                eprintln!("To allow requests, add the host to `allowed_http_posts`.");
+                if let Some(host) = host(&req.uri) {
+                    terminal::warn!("A component tried to make a HTTP request to non-allowed host '{host}'.");
+                    eprintln!("To allow requests, add 'allowed_http_hosts = [\"{host}\"]' to the manifest component section.");
+                }
                 return Err(HttpError::DestinationNotAllowed);
             }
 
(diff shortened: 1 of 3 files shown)


crates/outbound-http/src/lib.rs
```diff
@@ -53,6 +53,9 @@ impl outbound_http::Host for OutboundHttp {
                 .map_err(|_| HttpError::RuntimeError)?
             {
                 tracing::log::info!("Destination not allowed: {}", req.uri);
+                terminal::warn!("A component tried to make a HTTP request to '{}'.", req.uri);
+                eprintln!("The component was configured not to allow requests to this host.");
+                eprintln!("To allow requests, add the host to `allowed_http_posts`.");
```
mikkelhegn, line 58:
```suggestion
                eprintln!("To allow requests, add the host name to the `allowed_http_posts` key/value pair in the component section of spin.toml.");
```

mikkelhegn (approved):
LGTM 👍🏻

tschneidereit (approved):
Looks great, thank you! I think adding either Mikkel's suggestion or something like mine (or a combination of both) would be good, but otherwise 🚢 it.

crates/outbound-http/src/lib.rs
itowlson: I did wonder to what extent to make this a "explain what is happening" vs "write exhaustive documentation on how to fix it."  E.g. I thought about warning them to restart after making the change, but was wary of turning it into a wall of text that the eyes just slid off.  It may be too much in that direction already; I'd have preferred to keep it to 1 or 2 lines.

itowlson:
Based on feedback and angst I'm trying a new message:

image

I sent this as a separate commit so it's easy to back out if people hate this more than the previous one.

lann:
> `to the component manifest.`

I'm not sure that people will know what this means. Reference `spin.toml` in some form?

lann (approved):
Specific wording aside, lgtm

itowlson:
I avoided mentioning `spin.toml` because we don't know where the app came from.  I was thinking I should check if it came from a local manifest and not print the message at all if not, because if it comes from OCI then they can't edit it anyway.  It wasn't obvious to me how to determine if a LockedApp `origin` was OCI or file but I guess I could punt in the ol' "looks like an OCI reference" heuristic.

lann:
Ah yeah. Maybe something like "manifest component section" then? We _have_ been discussing having a separate "component manifest" thing...

itowlson:
SOLD

PR #1842, opened 2023-10-03
Move manifest tests from spin-loader to spin-trigger

lann:
Also convert them to "ui tests" which stick their expected output in a file next to the input which can be updated with BLESS=1.


fibonacci1729 (approved):
Nice!

PR #1843, opened 2023-10-03
Replace foo-bar header with content-type

itowlson:
YOU HAVE TAUNTED ME LONG ENOUGH RUST-HTTP TEMPLATE

(All the other HTTP templates already use `content-type` as their example header.)


itowlson:
I am excited for the moment when the integration tests turn out to depend critically on the `foo: bar` header

kate-goldenring (approved):

fibonacci1729 (approved):

PR #1844, opened 2023-10-03
ref(sdk/go): Refactor the Redis Go SDK to be more idiomatic

adamreese:
- Introduces a client struct responsible for handling the connection
- Refactor the `Execute()` arguments


sdk/go/redis/redis.go
```diff
@@ -23,55 +24,112 @@ func Handle(fn func(payload []byte) error) {
...
-// or more results.  This is a general-purpose function which should work with
-// any Redis command.
-func Execute(addr string, command string, arguments []RedisParameter) ([]RedisResult, error) {
-	return execute(addr, command, arguments)
+// Execute runs the specified Redis command with the specified arguments,
+// returning zero or more results.  This is a general-purpose function which
+// should work with any Redis command.
+//
+// Arguments must be string, []byte, int, int64, or int32.
+func (c *Client) Execute(command string, arguments ...any) ([]*Result, error) {
```
adamreese, line 129: I would love some opinions on using an `any` rather than a typed argument. I don't know how I feel about it.

lann: Its pretty standard for better or worse :shrug:

lann: I would probably have this return an error if an unsupported type is passed in though.

adamreese: Added an error

lann (approved):

Issue #1787, opened 2023-09-20
OCI: Max layer count per app


vdice:
Spin's OCI client currently pushes a layer per file for a Spin app.  Some registry implementations have a lower max for layers per 'image' than others.  For instance, ECR enforces a 500 layer per image max (as of recent testing).

We should cap the number of layers for a given Spin app to a configured maximum.  500 sounds like a good max for now, at least until we encounter a lower limit in another registry implementation.

This means we'll need to potentially bundle multiple files into a given layer.

We should avoid bundling component wasm files as these tend to be larger and duplicated between apps/deploys.

PR #1847, opened 2023-10-04
fix(oci): push archive layers if layer count exceeds max

vdice:
Closes https://github.com/fermyon/spin/issues/1787

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -4,6 +4,7 @@ use anyhow::{bail, Context, Result};
 use docker_credential::DockerCredential;
 use futures_util::future;
 use futures_util::stream::{self, StreamExt, TryStreamExt};
+use oci_distribution::errors::OciDistributionError;
 use oci_distribution::token_cache::RegistryTokenType;
 use oci_distribution::RegistryOperation;
 use oci_distribution::{
@@ -14,6 +15,7 @@ use oci_distribution::{
 };
 use reqwest::Url;
 use spin_app::locked::{ContentPath, ContentRef, LockedApp};
+use spin_common::sha256;
 use spin_loader::cache::Cache;
 use spin_manifest::Application;
 use tokio::fs;
@@ -21,16 +23,21 @@ use walkdir::WalkDir;
 
 use crate::auth::AuthConfig;
 
-// TODO: the media types for application, wasm module, and data layer are not final.
+// TODO: the media types for application, wasm module, data and archive layer are not final.
 const SPIN_APPLICATION_MEDIA_TYPE: &str = "application/vnd.fermyon.spin.application.v1+config";
 const WASM_LAYER_MEDIA_TYPE: &str = "application/vnd.wasm.content.layer.v1+wasm";
 const DATA_MEDIATYPE: &str = "application/vnd.wasm.content.layer.v1+data";
+const ARCHIVE_MEDIATYPE: &str = "application/vnd.wasm.content.bundle.v1.tar+gzip";
 
 const CONFIG_FILE: &str = "config.json";
 const LATEST_TAG: &str = "latest";
 const MANIFEST_FILE: &str = "manifest.json";
 
 const MAX_PARALLEL_PULL: usize = 16;
+/// Maximum layer count allowed per app, set in accordance to the lowest
+/// known maximum per image in well-known OCI registry implementations.
+/// (500 appears to be the limit for Elastic Container Registry)
+const MAX_LAYER_COUNT: usize = 500;
 
 // Inline content into ContentRef iff < this size.
 const CONTENT_REF_INLINE_MAX_SIZE: usize = 128;
@@ -100,10 +107,13 @@ impl Client {
         auth: RegistryAuth,
         reference: Reference,
     ) -> Result<Option<String>> {
-        // For each component in the application, add layers for the wasm module and
-        // all static assets and update the locked application with the file digests.
+        // For each component in the application, add a layer for the wasm module and
+        // separate layers for all static assets if application total will be under MAX_LAYER_COUNT,
+        // else an archive layer for all static assets per file entry if not.
+        // Finally, update the locked application with the layer digests.
         let mut layers = Vec::new();
         let mut components = Vec::new();
+        let archive_layers: bool = layer_count(locked.clone()).await? > MAX_LAYER_COUNT;
 
         for mut c in locked.components {
             // Add the wasm module for the component as layers.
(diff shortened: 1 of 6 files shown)


crates/app/src/locked.rs
```diff
@@ -126,6 +126,9 @@ pub struct ContentRef {
     /// If set, the content must have the given SHA-256 digest.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub digest: Option<String>,
+    /// If the content represents a compressed archive.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub archive: Option<bool>,
```
lann, line 131: I'd really like to avoid having this in the lockfile; I think ultimately it would keep things simpler. That _does_ require some refactoring in `Client::pull` but I actually want to make some other changes in there soon so I'm happy to tackle that if you'd like.

vdice: Sure, we can remove.  Were you thinking we'd differentiate based on directory where we pull layers into eg we currently parse the media type and store content in wasm or data dirs today... would we add an archive dir or similar?

lann: We can probably just stick them in the data dir. Even the split between the wasm and data dirs isn't _really_ necessary.

vdice:
Ok @lann, ready for another review.  Per our sync, https://github.com/fermyon/spin/pull/1847/commits/b9ce749e89ce4675d38659a47dc62bcd8cc3ff06 brings in the following changes:
- in `push`, add component.file entry for every file to be archived but only push the archive layer to the oci registry
- in `pull`, when we have an archive layer, unarchive and write files to expected locations in cache (by digest)
  - this logic wasn't quite as clean as I had hoped -- any pointers to simplify?
- revert changes in loader.rs

crates/loader/src/cache.rs
```diff
@@ -38,12 +38,12 @@ impl Cache {
...
 
     /// The Wasm bytes directory for the current cache.
-    fn wasm_dir(&self) -> PathBuf {
+    pub fn wasm_dir(&self) -> PathBuf {
         self.root.join(WASM_DIR)
     }
 
     /// The data directory for the current cache.
-    fn data_dir(&self) -> PathBuf {
+    pub fn data_dir(&self) -> PathBuf {
```
lann, line 46: Unless I'm missing something I think these can be un-`pub`bed again.

crates/oci/src/client.rs
```diff
@@ -230,10 +300,51 @@ impl Client {
...
                         {
                             Err(e) => return Err(e),
+                            // TODO: handle error(s), avoid unwraps below
                             _ => match layer.media_type.as_str() {
                                 WASM_LAYER_MEDIA_TYPE => {
                                     let _ = this.cache.write_wasm(&bytes, &layer.digest).await;
                                 }
+                                ARCHIVE_MEDIATYPE => {
+                                    // Write archive layer as usual to get cache check
+                                    let _ = this.cache.write_data(&bytes, &layer.digest).await;
```
lann, line 310: I see you just copied from above, but this should definitely be checking its result (here and above):

```suggestion
                                    this.cache.write_data(&bytes, &layer.digest).await?;
```

crates/oci/src/client.rs
```diff
@@ -230,10 +300,51 @@ impl Client {
...
+                                            if this.cache.data_file(&digest).is_ok() {
+                                                tracing::debug!("Skipping unpacked file {:?}; file already exists", file_path);
+                                            } else {
+                                                let _ = this.cache.write_data(&file_bytes, digest).await;
+                                            }
+                                        }
+                                    }
+                                }
                                 _ => {
                                     let _ = this.cache.write_data(&bytes, &layer.digest).await;
```
lann, line 349: ...this one too

```suggestion
                                    this.cache.write_data(&bytes, &layer.digest).await?;
```

crates/oci/src/client.rs
```diff
@@ -230,10 +300,51 @@ impl Client {
                             .await
                         {
                             Err(e) => return Err(e),
+                            // TODO: handle error(s), avoid unwraps below
```
lann, line 303: Hmm yeah there is all kinds of weirdness going on in here (already)...

crates/oci/src/utils.rs
```diff
@@ -0,0 +1,44 @@
+use anyhow::{Context, Result};
+use async_compression::tokio::write::GzipEncoder;
+use std::path::{Path, PathBuf};
+
+/// Create a compressed archive of source, returning its path in working_dir
+pub async fn compressed_archive(source: &Path, working_dir: &Path) -> Result<PathBuf> {
```
lann, line 6: If it would help simplify things, it would probably be fine if this _wasn't_ `async`. You can drop blocking work into a separate thread pool with tokio's [`spawn_blocking`](https://docs.rs/tokio/1.32.0/tokio/task/fn.spawn_blocking.html).

crates/oci/src/client.rs
vdice: I'm seeing the following when adding the `?` to bubble up the result:
Screenshot 2023-10-10 at 3 41 00 PM

Which led me to something like the following, but that doesn't look pretty:
Screenshot 2023-10-10 at 3 41 10 PM

There has to be a better way! (Right?)

crates/oci/src/client.rs
vdice: I won't claim it isn't still weird, but at least I got rid of the `unwrap()`s by breaking it out into a separate fn in https://github.com/fermyon/spin/pull/1847/commits/71640b3837e5da7dd776f2aecf263188a82a0934

crates/oci/src/utils.rs
vdice: But I was so excited to figure out how to do it all async!  I wouldn't say it has been getting in the way but I do realize it might be safer to go sync/spawn_blocking...

lann: I give up on async any time I can as soon as it becomes even mildly inconvenient 😅

crates/oci/src/client.rs
vdice: Slightly nicer... maybe: [64f0ae3](https://github.com/fermyon/spin/pull/1847/commits/64f0ae3b2e92d89f814331a60c5c7208c15da35a)

vdice:
https://github.com/fermyon/spin/pull/1847/commits/35dbd8494a27151c41f303c68a08a27803577f2b brings a slight improvement: pairs archive and unarchive logic into the one utils file, for (hopefully) easier refactoring (eg async -> sync, diff compression algorithm...)

lann (approved):
I think `Client` generally needs some TLC but that's not this PR's fault.

vdice:
Thanks @lann.  https://github.com/fermyon/spin/pull/1847/commits/3f7d2ce3b3b0b65d0a6c228fee11edc7b23b2e85 just moves the debug log.

Going to squash/clean up commits next...

Issue #1841, opened 2023-10-03
Use resources in Spin world interfaces


itowlson:
Instead of using pseudo resources like we currently do (e.g., modeling a kv store as a `u32`) we can switch to using wit resources. This should have minimal impact on how the code looks, but will make maintenance much easier.

rylev:
Here are the subtasks I think we need to do to complete this:
* [ ] Move pseudo resource usage (i.e., sqlite and kv) over to real resources
* [ ] Add resources to other interfaces that currently leak their resources (psql, mysql, redis)
* [ ] Create an adapter module for Spin 1.0 worlds to this new world
* [ ] Move all language SDKs over to this new world

PR #1854, opened 2023-10-05
Resources in Spin

rylev:
Part of #1841 

This PR moves sqlite and key-value storage to use component model resources. This is a significant breaking change to the Spin World but the user facing changes are more subtle:
* sqlite use to have `ValueParam` and `ValueResult` for values that are passed into function vs being returned back. We now only have `Value` (a direct translation of the wit) that gets passed to functions by reference and returned by ownership). So while one would have previously done `conn.execute("select * from table where foo = ?", &[ValueParam::Blob(&[1,2,3]))`, you now do `conn.execute("select * from table where foo = ?", &[Value::Blob(vec![1,2,3]))` (notice the move from a slice to an owned vector. 
* We are using the implementations more directly and thus are a bit more at the mercy of the quality of codegen from wit-bindgen. For example, `key_value::Store::get` used to take `key: impl AsRef<str>` allow for anything as long as it could be referenced as a str, but now it only takes `key: &str` allowing only for string slices (and anything dereferenced to a string slice). We could go back to wrapping these types, but I do believe one goal we've had is move to thinner and thinner SDKs so perhaps this is something we could improve on the wit-bindgen side. 

This PR only moves the Rust SDK over - we'll want to move all other SDKs over to wit-bindgen 12 as well. 

This also relies on https://github.com/fermyon/spin-componentize/pull/24


sdk/rust/src/key_value.rs
```diff
@@ -5,91 +5,31 @@
...
-    /// If `name` is empty, open the default store.  Other stores may also be available depending on the app
-    /// configuration.
-    pub fn open(name: impl AsRef<str>) -> Result<Self, Error> {
-        key_value::open(name.as_ref()).map(Self)
-    }
-
-    /// Open the default store.
-    ///
-    /// This is equivalent to `Store::open("default")`.
-    pub fn open_default() -> Result<Self, Error> {
```
itowlson, line 32: Could we bring this back please?

sdk/rust/src/llm.rs
```diff
@@ -7,19 +7,21 @@ pub use crate::wit::fermyon::spin::llm::{
...
+#[derive(Debug, Clone, Copy)]
 pub enum InferencingModel<'a> {
     Llama2Chat,
     CodellamaInstruct,
     Other(&'a str),
 }
 
 impl<'a> InferencingModel<'a> {
-    fn as_str(&self) -> llm::InferencingModel<'a> {
+    fn to_string(&self) -> llm::InferencingModel {
```
itowlson, line 18: Not sure why this change is needed?

sdk/rust/macro/wit/reactor.wit
```diff
@@ -0,0 +1,6 @@
+package fermyon:spin
+
+world exports {
```
itowlson, line 3: Nothing actually exports both inbound-http and inbound-redis does it?  I recall something about this causing problems in custom triggers where the SDK wouldn't work unless you faked up an export... is this an appropriate time to address that?  (Or right time but wrong PR? _grin_)

crates/trigger/src/loader.rs
```diff
@@ -56,6 +56,7 @@ impl Loader for TriggerLoader {
             )
         })?;
         let component = spin_componentize::componentize_if_necessary(&bytes)?;
+        std::fs::write("COMPONENT.wasm", &component).unwrap();
```
itowlson, line 59: Left-over diagnostic code?

sdk/rust/macro/src/lib.rs
```diff
@@ -6,17 +6,28 @@ use quote::quote;
...
+    const HTTP_COMPONENT_WIT_PATH: &str = concat!(env!("CARGO_MANIFEST_DIR"), "/wit");
 
     quote!(
         #func
 
         mod __spin_http {
+            #![allow(missing_docs)]
+            ::spin_sdk::wit_bindgen::generate!({
+                world: "exports",
+                path: #HTTP_COMPONENT_WIT_PATH,
```
dicej, line 18: I don't think this will work in general, given that the wit directory may not be present on the machine building the application code that is invoking the macro, right?

dicej: TIL this should work since the wit directory will be packaged as part of the crate.

sdk/rust/src/llm.rs
rylev: It's not need but the type changed and `as_str` seemed wrong given it was returning a String and not a `str`. I've since changes this to a `Display` impl.

sdk/rust/src/key_value.rs
rylev: Whoops!

sdk/rust/macro/wit/reactor.wit
rylev: I was able to fix this 🎉

crates/key-value/src/lib.rs
```diff
@@ -69,24 +65,33 @@ impl Default for KeyValueDispatch {
...
-    async fn get(&mut self, store: StoreHandle, key: String) -> Result<Result<Vec<u8>, Error>> {
+    async fn get(
+        &mut self,
+        store: Resource<key_value::Store>,
+        key: String,
+    ) -> Result<Result<Vec<u8>, Error>> {
         Ok(async {
             self.stores
-                .get(store)
+                .get(store.rep())
```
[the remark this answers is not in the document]
dicej: Though I suppose that would make sharing code between the new and legacy versions more difficult.

dicej (commented):
Where are the crates/trigger/wasi and crates/trigger/wit files used?  Sorry if it's obvious and I missed it.

crates/key-value-sqlite/src/lib.rs
```diff
@@ -162,7 +163,7 @@ mod test {
         );
 
         assert!(matches!(
-            kv.exists(42, "bar".to_owned()).await?,
+            kv.exists(Resource::new_own(42), "bar".to_owned()).await?,
```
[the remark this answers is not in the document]
rylev: I was planning on doing this in a follow up with other changes to the wit world. I'm nervous about how big this PR has already become.

crates/key-value/src/lib.rs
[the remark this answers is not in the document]
rylev: I'd like to see if I can remove this in a follow up PR.

dicej (approved):
LGTM once we get the Wasmtime deps switched over to a snapshot of the `release-14.0.0` branch.

rylev:
> Where are the crates/trigger/wasi and crates/trigger/wit files used? Sorry if it's obvious and I missed it.

Whoops! Those were from the component composition approach which was removed in favor of the multiple component implementation approach. I've removed those files.

PR #1855, opened 2023-10-06
add `wasi-http` trigger support based on `wasmtime-wasi-http`

dicej:
This required upgrading to a snapshot of what will eventually be Wasmtime 14, as well as upgrading `wit-bindgen` to 0.12.

Note that I had to make a copy of wit/preview2 under sdk/rust/macro in order to allow the `http_component` and `redis_component` macros to generate a `wit_bindgen::generate` invocation that actually works.  Ideally we'd avoid this duplication, e.g. by using a registry.


build.rs
```diff
@@ -90,6 +91,7 @@ error: the `wasm32-wasi` target is not installed
     cargo_build(RUST_HTTP_VAULT_CONFIG_TEST);
     cargo_build(RUST_OUTBOUND_REDIS_INTEGRATION_TEST);
     cargo_build(TIMER_TRIGGER_INTEGRATION_TEST);
+    cargo_build(WASI_HTTP_INTEGRATION_TEST);
```
rylev, line 94: This is fine for now since it's following an existing pattern, but I just want to note we should move this out of the build script and into the tests like we did for spin-componentize.

crates/http/src/lib.rs
```diff
@@ -1,7 +1,19 @@
...
 
 pub const WELL_KNOWN_PREFIX: &str = "/.well-known/spin/";
+
+pub fn full(bytes: Bytes) -> Body {
+    BoxBody::new(Full::new(bytes).map_err(|_| unreachable!()))
+}
+
+pub fn empty() -> Body {
+    BoxBody::new(Empty::new().map_err(|_| unreachable!()))
+}
```
rylev, line 19: Nit: we should probably put these in a `body` module so that it reads more nicely at the call site: e.g. `body::empty()`

crates/trigger-http/src/lib.rs
```diff
@@ -612,7 +624,7 @@ mod tests {
             .build_trigger()
             .await;
 
-        let body = Body::from("Fermyon".as_bytes().to_vec());
+        let body = spin_http::full(Bytes::copy_from_slice("Fermyon".as_bytes()));
```
rylev, line 627: Why don't we use [`Bytes::from_static`](https://docs.rs/bytes/1.5.0/bytes/struct.Bytes.html#method.from_static) for these?

examples/wasi-http-rust-async/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+
+mod wakers;
+
+const READ_SIZE: u64 = 16 * 1024;
+const MAX_CONCURRENCY: usize = 16;
+
+struct Component;
+
+impl IncomingHandler for Component {
+    fn handle(request: IncomingRequest, response_out: ResponseOutparam) {
```
rylev, line 39: The handling here is definitely not as straight-forward as before. What do we plan to do to make this more approachable?

examples/wasi-http-rust-async/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+    wakers: Wakers,
+    response: &OutgoingResponse,
+) -> impl Sink<Option<Vec<u8>>, Error = Error> {
+    outgoing_body(
+        wakers,
+        response.write().expect("response should be writable"),
+    )
+}
+
+fn outgoing_body(wakers: Wakers, body: OutgoingBody) -> impl Sink<Option<Vec<u8>>, Error = Error> {
```
rylev, line 175: Is the plan to move this into some shared library?

crates/trigger-http/src/lib.rs
dicej: Because I didn't know (or forgot) about it.  I'll change them.

examples/wasi-http-rust-async/src/lib.rs
dicej: Yeah, doing async without an SDK is going to be a bit ugly until Preview 3's first-class async is available.  My plan in the meantime is to provide wasi-http support in the Rust, JS, and Python SDKs that hides details like these.  Not sure how the Go SDK is going to look -- I'm leaving that up to the TinyGo/Go folks to figure out.

examples/wasi-http-rust-async/src/lib.rs
dicej: Yeah, specifically the Spin Rust SDK.

dicej: I've moved the executor and helper functions into the SDK.  Specifically, they're now generated by the `wasi_http_component` proc macro I've added due to limitations in `wit-bindgen`.  I'm hoping we can port `wasmtime-wit-bindgen`'s `with` directive to `wit-bindgen` in time for Spin 2.0, which will allow us to move that stuff out of the proc macro.

Cargo.toml
```diff
@@ -107,10 +108,13 @@ members = ["crates/*", "sdk/rust", "sdk/rust/macro"]
 
 [workspace.dependencies]
 tracing = { version = "0.1", features = ["log"] }
-wasmtime-wasi = { version = "13.0.0", features = ["tokio"] }
-wasi-common-preview1 = { version = "13.0.0", package = "wasi-common" }
-wasmtime = { version = "13.0.0", features = ["component-model"] }
-spin-componentize = { git = "https://github.com/fermyon/spin-componentize", rev = "f29cdb26a6b1700ae1fe48c94bbef26a8069d566" }
+wasmtime-wasi-http = { git = "https://github.com/bytecodealliance/wasmtime", rev = "17505667fa2895b3bd31bb9f5f4c544250090926" }
```
dicej, line 111: Using a snapshot of Wasmtime 14 until it is released, since that will be the earliest version that supports WIT resources and inbound `wasi-http`.

crates/core/src/lib.rs
```diff
@@ -213,12 +227,12 @@ impl<T: Send + Sync> EngineBuilder<T> {
         let engine = wasmtime::Engine::new(&config.inner)?;
 
         let mut linker: Linker<T> = Linker::new(&engine);
-        wasmtime_wasi::preview2::command::add_to_linker(&mut linker)?;
+        wasmtime_wasi_http::proxy::add_to_linker(&mut linker)?;
```
dicej, line 230: Note that `wasmtime_wasi_http::proxy::add_to_linker` includes a call to `wasmtime_wasi::preview2::command::add_to_linker`.

Cargo.toml
rylev: Should we switch to using [the release branch](https://github.com/bytecodealliance/wasmtime/tree/release-14.0.0)?

Cargo.toml
```diff
@@ -107,10 +108,13 @@ members = ["crates/*", "sdk/rust", "sdk/rust/macro"]
...
+wasmtime-wasi = { git = "https://github.com/bytecodealliance/wasmtime", rev = "17505667fa2895b3bd31bb9f5f4c544250090926", features = ["tokio"] }
+wasi-common-preview1 = { git = "https://github.com/bytecodealliance/wasmtime", rev = "17505667fa2895b3bd31bb9f5f4c544250090926", package = "wasi-common" }
+wasmtime = { git = "https://github.com/bytecodealliance/wasmtime", rev = "17505667fa2895b3bd31bb9f5f4c544250090926", features = ["component-model"] }
+spin-componentize = { git = "https://github.com/fermyon/spin-componentize", branch = "wasmtime-17505667" }
```
rylev, line 115: Let's make sure to get the changes we need on `main` in spin-componentize.

rylev (approved):
This looks good. I have some nits, but I think those will largely be resolved by working through the conflicts between this PR and the resources PR.

Cargo.toml
dicej: That Wasmtime commit _is_ from the `release-14.0.0` branch (as it existed when I did the update).  I considered just pointing to the `release-14.0.0` branch, but I was a bit paranoid about various Cargo.lock files getting out of sync.

dicej: Also, the Wasmtime version really has to match the adapter _exactly_, or you get build errors, so I didn't want to imply that any snapshot of `release-14.0.0` would necessarily work.

dicej:
Closing in favor of #1877

PR #1859, opened 2023-10-06
ci: Embiggen build runner

lann:
This build step has run out of disk space recently, so we're bumping up to a large runner to give us some breathing room.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,11 +100,11 @@ jobs:
 
   test-rust:
     name: Test Spin SDK - Rust
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.runner }}
     needs: [lint-rust]
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        runner: [ubuntu-22.04-4core-spin, macos-latest]
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


vdice (approved):

radu-matei (approved):

itowlson:
Something a bit weird with required checks that it thinks aren't starting.  @vdice The pending checks are all ones that have changed name as a result of the embiggening e.g. from "Rust / Build Spin (windows-latest) (pull_request)" to "Rust / Build Spin (windows-latest, windows-server-latest-8core-spin, .exe) (pull_request)" - could the required checks checker be not recognising the new names as being the checks it cares about?

lann:
I'm going to roll back to the minimal fix here. We're still somewhat bottlenecked by the macos runner now.

lann:
Given that I just went back to a previously-validated commit and that the branch protection rules need to be updated anyway, I'm going to bypass checks now.

vdice:
>The pending checks are all ones that have changed name as a result of the embiggening e.g. from "Rust / Build Spin (windows-latest) (pull_request)" to "Rust / Build Spin (windows-latest, windows-server-latest-8core-spin, .exe) (pull_request)" - could the required checks checker be not recognising the new names as being the checks it cares about?

Ah, good catch.  I believe you are right @itowlson.

I can help update the req'd checks config when we are ready to update the rest of the runner configs.

PR #1864, opened 2023-10-09
Allow `spin add` to add variables to the manifest

itowlson:
Syntax:

```
// spin-template.toml
[add_component]
[add_component.snippets]
variables = "variables.txt"

// snippets/variables.txt
secret = { required = true }
url = { default = "{{ service-url }}" }
```

Questions:

* Should the variables be merged into `component.txt`?  Splitting them makes it easier to implement because we don't need to analyse which bits of the snippet need to be merged and which appended - we can instead use the snippet key to infer the semantics.
* Should `variables.txt` include the `[variables]` header?  It is not needed for the implementation but `component.txt` has it so maybe for consistency?
* If the manifest being added to already has a variable with the same name as one given in the snippet, the one from the snippet takes precedence.  Do we just need to document this, or do we need to ask the user if we are about to change an existing value?  (The danger being that the user may not know how to decide.)

### DIFF
--- a/crates/templates/src/run.rs
+++ b/crates/templates/src/run.rs
@@ -11,6 +11,7 @@ use walkdir::WalkDir;
 use crate::{
     cancellable::Cancellable,
     interaction::{InteractionStrategy, Interactive, Silent},
+    renderer::MergeTarget,
     template::TemplateVariantInfo,
 };
 use crate::{
@@ -248,9 +249,20 @@ impl Run {
                         Err(anyhow::anyhow!("Spin doesn't know what to do with a 'component' snippet outside an 'add component' operation")),
                 }
             },
+            "variables" => {
+                match &self.options.variant {
+                    TemplateVariantInfo::AddComponent { manifest_path } =>
+                        Ok(RenderOperation::MergeToml(
+                            manifest_path.clone(),
+                            MergeTarget::Application("variables"),
+                            content,
+                        )),
+                    TemplateVariantInfo::NewApplication =>
+                        Err(anyhow::anyhow!("Spin doesn't know what to do with a 'variables' snippet outside an 'add component' operation")),
+                }
+            },
             _ => Err(anyhow::anyhow!(
-                "Spin doesn't know what to do with snippet {}",
-                id
+                "Spin doesn't know what to do with snippet {id}",
             )),
         }
     }
--- a/crates/templates/src/writer.rs
+++ b/crates/templates/src/writer.rs
@@ -9,6 +9,7 @@ pub(crate) struct TemplateOutputs {
 pub(crate) enum TemplateOutput {
     WriteFile(PathBuf, Vec<u8>),
     AppendToml(PathBuf, String),
+    MergeToml(PathBuf, &'static str, String), // only have to worry about merging into root table for now
 }
 
 impl TemplateOutputs {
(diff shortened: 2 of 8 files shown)


fibonacci1729:
> Should the variables be merged into component.txt? Splitting them makes it easier to implement because we don't need to analyse which bits of the snippet need to be merged and which appended - we can instead use the snippet key to infer the semantics.

+1 to keeping them separate files. Separate files seems more flexible especially when you consider overlapping sets of variables declarations shared across components. 

> Should variables.txt include the [variables] header? It is not needed for the implementation but component.txt has it so maybe for consistency?

As much as I love consistency, it would seem like adding the `variables` header would require some extra processing complexity to merge the contents of each snippet section into a singular `variables` section (which isn't the absolute worst)? I think I would lean towards omitting the `variables` header from the snippet unless it is really compelling to others.

> If the manifest being added to already has a variable with the same name as one given in the snippet, the one from the snippet takes precedence. Do we just need to document this, or do we need to ask the user if we are about to change an existing value? (The danger being that the user may not know how to decide.)

I would vote to document plus a warning that a previously defined variable is going to be overwritten.  I think the more concerning question is how does a developer reconcile these conflict? 
* If accepting the overwriting then a component previously wired up to that variable may misbehave. 
* If not, then is one simply forced to not use the target of `add_component` (which obviously is the worst experience).

Apologies if i'm missing something obvious or this isn't the place to raise this question!

fibonacci1729 (approved):
Code looks great and this current iteration makes sense pending decision around handling conflict behavior?

crates/templates/src/run.rs
```diff
@@ -248,6 +249,18 @@ impl Run {
...
+            "variables" => {
+                match &self.options.variant {
+                    TemplateVariantInfo::AddComponent { manifest_path } =>
+                        Ok(RenderOperation::MergeToml(
+                            manifest_path.clone(),
+                            MergeTarget::Application("variables"),
+                            content,
+                        )),
+                    TemplateVariantInfo::NewApplication =>
+                        Err(anyhow::anyhow!("Spin doesn't know what to do with a 'variables' snippet outside an 'add component' operation")),
```
kate-goldenring, line 261: This error assumes that a Spin CLI user is familiar enough with the templates system to know what a snippet is. Maybe instead:

```suggestion
                        Err(anyhow::anyhow!("Cannot create new applications from templates with variables")),
```

crates/templates/src/run.rs
```diff
@@ -248,6 +249,18 @@ impl Run {
...
             },
+            "variables" => {
+                match &self.options.variant {
+                    TemplateVariantInfo::AddComponent { manifest_path } =>
+                        Ok(RenderOperation::MergeToml(
+                            manifest_path.clone(),
+                            MergeTarget::Application("variables"),
+                            content,
+                        )),
+                    TemplateVariantInfo::NewApplication =>
```
kate-goldenring, line 260: Could we add support for new templates with variables in a future PR?

crates/templates/src/writer.rs
```diff
@@ -9,6 +9,7 @@ pub(crate) struct TemplateOutputs {
 pub(crate) enum TemplateOutput {
     WriteFile(PathBuf, Vec<u8>),
     AppendToml(PathBuf, String),
```
kate-goldenring, line 11: Would it be worth documenting in line  here what each of the `MergeToml` parts are?

crates/templates/src/writer.rs
```diff
@@ -47,7 +48,110 @@ impl TemplateOutput {
...
+    let merging = merging.as_table();
+    match doc.get_mut(target) {
+        Some(item) => {
+            let Some(table) = item.as_table_mut() else {
+                anyhow::bail!("Cannot merge template data into {target} as it is not a table");
+            };
+            for (key, value) in merging {
+                match table.entry(key) {
+                    Entry::Occupied(mut e) => {
+                        let existing_val = e.get_mut();
```
kate-goldenring, line 83: What is the thought process on the added component's variables trumping existing component vars? I'd think the other way around would be preferred. Better to break new functionality than existing functionality IMO

kate-goldenring (commented):
This is a great approach! Just a few clarifying questions

crates/templates/src/writer.rs
```diff
@@ -9,6 +9,7 @@ pub(crate) struct TemplateOutputs {
 pub(crate) enum TemplateOutput {
     WriteFile(PathBuf, Vec<u8>),
     AppendToml(PathBuf, String),
+    MergeToml(PathBuf, &'static str, String), // only have to worry about merging to top level for now
```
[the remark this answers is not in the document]
itowlson: top-level = merging into the root TOML table rather than having to merge into the `component` table or the `component.trigger` table.  Is there a better way to express this?  I couldn't think of one

crates/templates/src/run.rs
itowlson: That particular error would be misleading - we can totally create new applications from templates with variables - but yeah I can try to think of something better.  (This is meant to be a can't-happen scenario; it can only happen if the template author ships an always-borked template.)

crates/templates/src/run.rs
itowlson: New templates can have variables - I'm not sure what you're getting at here.

crates/templates/src/writer.rs
itowlson: The thought process was roughly "if there is a clash, it's likely because I'm re-running the same template, in which case I probably want the new definitions rather than the old ones."  And if I do want to keep the old definitions then I can still refer back to them through source control.  That's not to say the thought process was right of course!

crates/templates/src/writer.rs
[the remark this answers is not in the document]
itowlson: I'm trying "root table" instead because at least that hints to the reader that it's a TOML term of art

crates/templates/src/run.rs
itowlson: Trying "Invalid template: template wants to merge variables into existing manifest when creating a new manifest"

crates/templates/src/writer.rs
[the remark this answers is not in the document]
kate-goldenring: Maybe "app-level"?

crates/templates/src/run.rs
kate-goldenring: I was missing the deeper dive into the code base of how we are defining the snippets  only under the AddComponent variant. I now see how this path isn't executed during adds. Thanks!

crates/templates/src/run.rs
itowlson: No worries, Kate, and thanks for digging into this - I will return to the original message.

kate-goldenring (approved):
Learned a lot more about the template system through reviewing. Thanks Ivan!

itowlson:
Okay I'm going to merge this and we can iterate on conflict behaviour

PR #1868, opened 2023-10-09
Change `spin plugins list` to be installed only

itowlson:
This is the breaking part of the changes requested in #1716.  It changes the default for `spin plugins list` from "all" to "installed."

Considerations:

* This results in different signatures for filtering according to whether you are doing installed (`spin plugins list --filter cloud`) vs all plugins (`spin plugins search cloud`).  This follows from the different naming but is a bit weird when testing them side by side.  Is this a concern?
* I dithered for ages about whether to support `spin plugins list --all` (or `--available`).  It feels like a nice option to me (and would mitigate the asymmetry between `list` and `search`), but in the end I didn't because it seemed like a lot of people didn't like having more than one way to do the same thing.
* I kept `--installed` as a hidden no-op option on `spin plugins list` so that any existing scripts that relied on `spin plugins list --installed` didn't break, but maybe that's not worth it?

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -26,7 +26,7 @@ pub enum PluginCommands {
     /// plugins directory.
     Install(Install),
 
-    /// List available or installed plugins.
+    /// List installed plugins.
     List(List),
 
     /// Search for plugins by name.
@@ -320,12 +320,13 @@ impl Upgrade {
     }
 }
 
-/// List available or installed plugins.
+/// List installed plugins.
 #[derive(Parser, Debug)]
 pub struct List {
-    /// List only installed plugins.
-    #[clap(long = "installed", takes_value = false)]
-    pub installed: bool,
+    // This is always treated as set, but is retained for backward
+    // compatibility.
+    #[clap(long = "installed", takes_value = false, hide = true)]
+    _installed: bool,
 
     /// Filter the list to plugins containing this string.
     #[clap(long = "filter")]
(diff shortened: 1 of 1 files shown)


itowlson:
Incidentally the original ask in #1716 was just for `spin plugins list` to get latest before listing.  I added `search` as part of that work.  I am not sold that this final change to make `list` installed _only_, and to require users to use a separate verb for available, is truly beneficial, but here it is if we decide we want it!

src/commands/plugins.rs
```diff
@@ -320,12 +320,13 @@ impl Upgrade {
     }
 }
 
-/// List available or installed plugins.
+/// List installed plugins.
```
vdice, line 323: It looks like the following line needs to be updated as well: https://github.com/fermyon/spin/blob/c3c1de2fd0fcd33aab6d345e4273d2c526c77222/src/commands/plugins.rs#L29

src/commands/plugins.rs
```diff
@@ -334,10 +335,30 @@ pub struct List {
...
+
+struct ListCommandCore {
+    which: WhichPlugins,
+    filter: Option<String>,
+}
+
+impl ListCommandCore {
+    pub async fn run(self) -> Result<()> {
+        let mut plugins = match &self.which {
+            WhichPlugins::Installed => Self::list_installed_plugins(),
```
vdice, line 360: Now that `list` only shows installed, I wonder if it is redundant to keep the `[installed]` labels next to each listing?
```
$ spin plugins list
check-for-update 0.1.0 [installed]
cloud 0.4.0 [installed]
js2wasm 0.6.0post.1694728162 [installed]
platform 0.1.0post.1694720740 [installed]
pluginify 0.6.0 [installed]
py2wasm 0.3.0post.1693957212 [installed]
trigger-sqs 0.3 [installed]
```

vdice (commented):
> This results in different signatures for filtering according to whether you are doing installed (spin plugins list --filter cloud) vs all plugins (spin plugins search cloud). This follows from the different naming but is a bit weird when testing them side by side. Is this a concern?

I do find that slightly weird as well.  Possible to _also_ allow `--filter` for search, just in case users try to carry over the same use from list?  Not sure, let's see what others think.

src/commands/plugins.rs
itowlson: We could.  Keeping it keeps the presentations unified, but we can remove it if people feel strongly.

itowlson:
This wasn't approved in time for v2, and as it's a breaking change, we have to put it on ice.

Issue #604, opened 2022-06-27
feature: "startup" entrypoint


lann:
@ecumene is [looking for a way to make an external service registration request on Spin app startup](https://github.com/fermyon/spin/pull/521#issuecomment-1167712593).

This might be a reasonable feature to add to _all_ trigger types, perhaps as a separate optional func export, e.g. `spin_start: func()`. Need to discuss how it might fit into a production environment though; would it trigger just once at "deploy time"?

radu-matei:
What if instead of an optional function export, this was an optional _component_?
This way you could technically reuse them across multiple applications without code duplication?

I like "deployment hook" as opposed to "startup" — it's clear that it happens once when you deploy the application (although that might make running locally a bit more confusing).

lann:
> What if instead of an optional function export, this was an optional component?

I suppose in Spin's current design it would make the most sense for it to be a separate trigger type, which would make more sense with the future manifest redesign, i.e.:

```toml
[[trigger.deployment]]
component = "discord-bot"
```

itowlson:
The advantage of an entry point - at least in the Spin 1.0 model - is that it travels with the component.  So if an application references a component at a URL or in a registry, and the component requires a database schema, the component can expose a "startup" entry point which performs its migrations.  The application author does not need to worry about adding a separate "migration companion" component, or about keeping the migration component in sync with the handler.

In the component model, I imagine a composed component could achieve the same result; but we'd still want the component author to express that dependency without the Spin app author needing to know about it.  It would cause maintenance woes if Spin app authors had to specify the composition explicitly using Manifest2 entries, and remember to update things in unison.  This composition is of course do-able in the component model but I am not sure how Spin would then invoke the migration component except through an entry point - but perhaps more componentised brains than mine know a way?

(Using 'migration' here as representative of the various use cases mentioned.)

PR #1869, opened 2023-10-10
Application startup entrypoint

itowlson:
This was a vague experiment in relation to #604 that accidentally worked and now I am not sure what to do with it.

My test case wasn't in the Spin repo but basically it was:

```
// WIT
use fermyon:spin/inbound-http
use fermyon:spin/app-start
use fermyon:spin/sqlite

world example {
    export inbound-http
    export app-start
    import sqlite
}
```

```rust
// Rust
impl crate::bindings::exports::fermyon::spin::app_start::Guest for Component {
    fn handle_app_start() -> Result<(), String> {
        let conn = crate::bindings::fermyon::spin::sqlite::open("default")
            .map_err(|e| format!("{e:?}"))?;
        crate::bindings::fermyon::spin::sqlite::execute(conn, "CREATE TABLE IF NOT EXISTS test (message TEXT)", &[])
            .map_err(|e| format!("{e:?}"))?;
        Ok(())
    }
}
```

...and then the HTTP handler implementation was able to use the SQLite table without me needing to initialise it in `spin up`.

So I guess I am interested in knowing whether people feel this is the right way / place to do this (with suitable tidying), and whether we would like to productionise it (which, apart from agreeing a WIT and defining desired behaviour, would of course have Cloud implications that are likely the real difficulty with the feature!).

### DIFF
--- a/crates/trigger/src/lib.rs
+++ b/crates/trigger/src/lib.rs
@@ -155,8 +155,26 @@ impl<Executor: TriggerExecutor> TriggerExecutorBuilder<Executor> {
             .iter_mut()
             .try_for_each(|h| h.app_loaded(app.borrowed(), &runtime_config))?;
 
+        let app_engine = TriggerAppEngine::new(engine, app_name, app, self.hooks).await?;
+
+        for component_id in app_engine.app().components().map(|c| c.id().to_string()) {
+            let (instance, mut store) = app_engine.prepare_instance(&component_id).await?;
+            if let EitherInstance::Component(instance) = instance {
+                let start_func = instance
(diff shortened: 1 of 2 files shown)


lann:
Yeah, I think the hard part here is the policy for when the entrypoint is executed. It would probably need to have a lower bound of "at least once per [`spin up`/`spin deploy`/how do you describe this generically?]". The upper bound is trickier. We can't really do "exactly once" in any sort of distributed system, but if you set it as high as "as often as once per request" then you aren't much better off than just running the init code in your handler. We need something between those extremes but I'm not sure what it ought to be.

Issue #16, opened 2022-01-26
Initializing new applications


radu-matei:
Currently, Spin has no concept of "initializing" a new application without using a template (re #15).

Is there a need for a `spin init` command? What would it do?

mikkelhegn:
I can only see two reasons:
1. Do not need source, but will just reference a pre-built .wasm module for components.
2. A very opinionated directory layout is preferred.

For 1. I believe the feature to re-use components with a URL reference is the best option to promote. We could have a template supporting this scenario, where the template would require the URL and Digest as part of the template creation flow.

For 2. The only thing we coudl do was to create a `spin.toml` file with no components (a few lines). This could also be a template, e.g., `empty app`.

itowlson:
FWIW there is a `http-empty` template, which is a `spin.toml` with no components, but it does set the application `trigger` to HTTP.

michelleN:
@radu-matei What is the use case you're targeting for initializing a new application? What kind of workflow are you envisioning?

itowlson:
Additional context: https://github.com/fermyon/spin/issues/1525#issuecomment-1756104398

michelleN:
Considering we have an `--init` flag on spin new now, do we still want to have a new top level command for 2.0? Should we try out `spin new --init` for a while and come back to this post 2.0?

itowlson:
At this stage I would be strongly in favour of not attempting this for 2.0.  With Manifest 2, it becomes a lot more practical than it was with Manifest 1 though, so getting rid of what is now a slightly artificial `http-empty` template seems like a good plan!

calebschoepp:
Is this still something we want @radu-matei @michelleN? I happened across this issue and it feels like possibly this use case is no longer valid and we could close this issue.

radu-matei:
Yeah, I think it's safe to close this; I haven't seen more people ask for changes to the new experience; we can reopen if needed.

PR #1872, opened 2023-10-10
Changes to the `spin new` experience

### LINKED ISSUE #1525: `spin new`: provide application name without providing template
When creating a Spin application, I usually know the name I want to give the application, but I often cannot remember the template name I'd like to use (is it `http-rust`, `rust-http`, or something else?). I really wish I could `spin new $SOME_NAME` and then be presented with a picker for the template. 

This might not be possible to add though given that `spin new` first expects a template *then* a name.

itowlson:
Fixes #1525.

~**WIP:** There are more changes requested in #1525, but I'm getting it under way immediately because the e2e tests use `spin new` quite a bit and I want to know where they blow up sooner rather than later.  I fixed one place but I'm sure there will be others!~  This now has everything except the proposal to generate a blank component on `spin new` with no template, on which I'd like to gather more feedback before implementing.  That can be a follow-up PR if necessary.

And you don't know how much I wanted to call this PR "The new `new` thing."

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -17,12 +17,13 @@ use crate::opts::{APP_MANIFEST_FILE_OPT, DEFAULT_MANIFEST_FILE};
 /// Scaffold a new application based on a template.
 #[derive(Parser, Debug)]
 pub struct TemplateNewCommandCore {
-    /// The template from which to create the new application or component. Run `spin templates list` to see available options.
-    pub template_id: Option<String>,
-
     /// The name of the new application or component.
     #[clap(value_parser = validate_name)]
     pub name: Option<String>,
+
+    /// The template from which to create the new application or component. Run `spin templates list` to see available options.
+    #[clap(short = 't', long = "template")]
+    pub template_id: Option<String>,
 
     /// Filter templates to select by tags.
     #[clap(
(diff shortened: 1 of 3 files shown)


fibonacci1729 (approved):
🎉 Excellent!

src/commands/new.rs
```diff
@@ -34,9 +35,14 @@ pub struct TemplateNewCommandCore {
 
     /// The directory in which to create the new application or component.
     /// The default is the name argument.
-    #[clap(short = 'o', long = "output")]
+    #[clap(short = 'o', long = "output", group = "location")]
     pub output_path: Option<PathBuf>,
 
+    /// The directory in which to create the new application or component.
```
fibonacci1729, line 41:
```suggestion
    /// Create the application in the current working directory.
```

... Or something (comment appears to be the same from `-o`)

itowlson: oopsie boopsie!  Thanks for the catch

itowlson:
@melissaklein24 I unlinked #16 because this doesn't currently address that.  See https://github.com/fermyon/spin/issues/1525#issuecomment-1756104398 and https://github.com/fermyon/spin/issues/1525#issuecomment-1756138632 for more context on what fixing #16 would involve.

rylev (approved):
🎉

Issue #1866, opened 2023-10-09
Changes to WIT for Spin 2.0 [mini-SIP]


rylev:
The following are my suggestions for changes to the Spin WIT from the Guest perspective. I'm opening this as an issue/mini-SIP as the changes are largely unrelated and are more a grab bag of "small improvements" rather than a single unified approach for improvement. 

## Interfaces with no changes

The following interfaces will not change at all:
* `config`
* `http`
* `llm` (while stream support would be great to add, the lack of first-class streams makes it very difficult for us to provide an API that is user friendly enough)

### `key-value`

The `key-value` interface will change in the following ways:
* Move to using an actual `resource` instead of psuedo resources. 
  * This moves from free standing functions to functions that take an implicit `resource` parameter. This actually matches almost 1 to 1 with how the SDKs expose these interfaces so user change should be minimal.
  * The `close` function is removed in favor of `resource` dropping
* `error::no-such-key` is removed and `store::get` returns an `result<option<list<u8>, error>` instead of a `result<list<u8>, error>`

```wit
interface key-value {
  /// An open key-value store
  resource store {
    /// Open the store with the specified name.
    ///
    /// If `name` is "default", the default store is opened.  Otherwise,
    /// `name` must refer to a store defined and configured in a runtime
    /// configuration file supplied with the application.
    ///
    /// `error::no-such-store` will be raised if the `name` is not recognized.
    open: static func(name: string) -> result<store, error>

    /// Get the value associated with the specified `key` from the specified
    /// `store`.
    ///
    /// `ok(none)` will be returned if the key does not exist. `error::invalid-store` 
    /// will be raised if `store` is not a valid handle to an open store.
    get: func(key: string) -> result<option<list<u8>>, error>

    /// Set the `value` associated with the specified `key` in the specified
    /// `store`, overwriting any existing value.
    ///
    /// `error::invalid-store` will be raised if `store` is not a valid handle
    /// to an open store.
    set: func(key: string, value: list<u8>) -> result<_, error>

    /// Delete the tuple with the specified `key` from the specified `store`.
    ///
    /// `error::invalid-store` will be raised if `store` is not a valid handle
    /// to an open store.  No error is raised if a tuple did not previously
    /// exist for `key`.
    delete: func(key: string) -> result<_, error>

    /// Return whether a tuple exists for the specified `key` in the specified
    /// `store`.
    ///
    /// `error::invalid-store` will be raised if `store` is not a valid handle
    /// to an open store.
[...]

PR #1873, opened 2023-10-10
Changes to outbound redis interfaces

rylev:
Part of #1866 

The changes to the interface are:

* Uses a connection oriented resource rather than free functions.
* Updates the error type from just being `error` to 4 different error possibilities.
* Moves the functions that returned to `s64` to return `u32` since they should never return anything by positive numbers
* `get` returns `result<option<list<u8>>, error>` rather than `result<list<u8>, error>` returning `ok(none)` when the key does not exist - this mirrors how Redis's server API.

### DIFF
--- a/wit/preview2/redis.wit
+++ b/wit/preview2/redis.wit
@@ -37,17 +37,17 @@ interface redis {
 
     /// Removes the specified keys.
     ///
-    /// A key is ignored if it does not exist.
-    del: func(keys: list<string>) -> result<s64, error>
+    /// A key is ignored if it does not exist. Returns the number of keys deleted.
+    del: func(keys: list<string>) -> result<u32, error>
 
     /// Add the specified `values` to the set named `key`, returning the number of newly-added values.
-    sadd: func(key: string, values: list<string>) -> result<s64, error>
+    sadd: func(key: string, values: list<string>) -> result<u32, error>
 
     /// Retrieve the contents of the set named `key`.
     smembers: func(key: string) -> result<list<string>, error>
 
     /// Remove the specified `values` from the set named `key`, returning the number of newly-removed values.
-    srem: func(key: string, values: list<string>) -> result<s64, error>
+    srem: func(key: string, values: list<string>) -> result<u32, error>
 
     /// Execute an arbitrary Redis command and receive the result.
     execute: func(command: string, arguments: list<redis-parameter>) -> result<list<redis-result>, error>
--- a/wit/preview2/redis.wit
+++ b/wit/preview2/redis.wit
@@ -3,6 +3,8 @@ interface redis {
 
   /// Errors related to interacting with Redis
   variant error {
+      /// An invalid address string
+      invalid-address,
       /// There are too many open connections
       too-many-connections,
       /// A retrieved value was not of the correct type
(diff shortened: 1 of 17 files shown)


wit/preview2/redis.wit
```diff
@@ -0,0 +1,55 @@
...
+  /// Errors related to interacting with Redis
+  variant error {
+      /// An invalid address string
+      invalid-address,
+      /// There are too many open connections
+      too-many-connections,
+      /// A retrieved value was not of the correct type
+      type-error,
+      /// An I/O error occurred
+      io(string),
```
lann, line 13: Should we just call this `other(string)` or something? It seems unlikely that all our surprise errors would be IO-related...

(I don't really care for Redis in particular, but for (re)establishing a precedent)

wit/preview2/redis.wit
```diff
@@ -0,0 +1,55 @@
...
+    ///
+    /// If the key does not exist, it is set to 0 before performing the operation.
+    /// An `error::type-error` is returned if the key contains a value of the wrong type
+    /// or contains a string that can not be represented as integer.
+    incr: func(key: string) -> result<s64, error>
+
+    /// Removes the specified keys.
+    ///
+    /// A key is ignored if it does not exist.
+    del: func(keys: list<string>) -> result<s64, error>
```
lann, line 41: Should this return a `u64`? Also maybe update the docs that it returns "the number of keys removed"?

wit/preview2/redis.wit
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Add the specified `values` to the set named `key`, returning the number of newly-added values.
+    sadd: func(key: string, values: list<string>) -> result<s64, error>
```
lann, line 44: `u64`?

wit/preview2/redis.wit
```diff
@@ -0,0 +1,55 @@
...
+
+    /// Retrieve the contents of the set named `key`.
+    smembers: func(key: string) -> result<list<string>, error>
+
+    /// Remove the specified `values` from the set named `key`, returning the number of newly-removed values.
+    srem: func(key: string, values: list<string>) -> result<s64, error>
```
lann, line 50: `u64`?

wit/preview2/redis.wit
rylev: Good catch - this is copied from the first version. It seems that technically the returned number is a signed integer so the conservative thing to do would be to pass that along. However, we can be sure (right?!) that the number of keys has to be positive...

wit/preview2/redis.wit
lann: Yeah I suspect that pretty much any time we'd want another specific case we could be more precise than `io` :shrug:

wit/preview2/redis.wit
lann: Unless you start deleting anti-keys...

lann:
Looks like the underlying `get` it can return any `FromRedisValue`, which could be an `Option<Vec<u8>>` to preserve the nil: https://docs.rs/redis/latest/src/redis/types.rs.html#1516

lann:
Unless I'm missing something, the underlying `redis` API supports both, depending on whether your call to `get` uses `RV = Vec<u8>` or `RV = Option<Vec<u8>>`. It is just our current implementation that (implicitly, via inference) uses `RV = Vec<u8>`.

wit/preview2/redis.wit
rylev: I think it's fine to `u32`. Any concerns with that?

radu-matei (approved):

PR #1875, opened 2023-10-11
Enable finding plugins in a user-selected directory

### LINKED ISSUE #1870: Configurable (or additional) plugin directory search paths
Because Spin Plugin binaries are precompiled binaries with a variety of system dependencies, they may not run as-built-by-Fermyon on all systems - for example `spin cloud` doesn't run on `NixOS` because the ELF interpreter and openssl have a different path (see also our GLIBC version issues in the past).

It would be helpful to allow package managers to distribute Spin Plugins to account for this (and also means that plugins can be reliably updated in the same manner as people otherwise manage their systems).

itowlson:
Fixes #1870.

This is a candidate fix based on my current understanding of the requirement.  If we validate that it meets the need, then I'll tidy it and put it ready for review.

### DIFF
--- a/src/commands/external.rs
+++ b/src/commands/external.rs
@@ -7,6 +7,7 @@ use spin_plugins::{
     PluginStore,
 };
 use std::io::{stderr, IsTerminal};
+use std::path::PathBuf;
 use std::{collections::HashMap, env, process};
 use tokio::process::Command;
 use tracing::log;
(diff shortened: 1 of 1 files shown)


itowlson:
This is potentially relevant to #2555, although I'm not sure the strategy of requiring an environment variable would fit with Rancher Desktop's needs.  That feels like it's closer to the Homebrew situation, which had a similar requirement to Rancher and for which we did a one-off fix: maybe we should explore a more wide-ranging, configurable equivalent to that.  cc @kate-goldenring

itowlson:
also cc @jandubois for visibility - looking at the issue, Rancher's needs are broader than just plugins, but it does sound like the environment variable approach would meet their needs

jandubois:
> I'm not sure the strategy of requiring an environment variable would fit with Rancher Desktop's needs

I was hoping for a minimal addition to `default_data_dir` in `crates/common/src/data_dir.rs`.

I'm not a Rust person, so I'm just making up stuff, but I would hope for something like this:

```rust
/// Return the default data directory for Spin
pub fn default_data_dir() -> Result<PathBuf> {
    if let Ok(spin_appdata) = std::env::var("SPIN_APPDATA") {
        return Some(spin_appdata);
    }
    …
}
```

Don't care about the name of the variable, but it would disable all heuristics/rules and always return the path set in the environment. It would work for both plugins and templates (and everything else you will store in the data directory in the future).

I think the issue of having an **additional** plugin directory is orthogonal to this feature, but of course the way these features would interact should be considered (I have not reviewed the code in this PR in detail, nor the issue that is to be fixed by it).2

lann:
#2563

PR #1880, opened 2023-10-11
Key value interface changes

rylev:
Part of #1866 

The following changes are made to the `key-value` interface:
* No `invalid-store` error - if the user forges a store (which is not possible without unsafe code in the resource based interface), the implementation will trap instead of returning an error. This also changes the old interface to do the same meaning that users will never see `invalid-store` errors again even if they use the old interface.
* `get` returns `result<option<list<u8>>, error>` instead of `result<list<u8>, error>`. This also means the `no-such-key` error is removed in the new interface.

### DIFF
--- a/crates/key-value-redis/src/lib.rs
+++ b/crates/key-value-redis/src/lib.rs
@@ -55,13 +55,7 @@ struct RedisStore {
 impl Store for RedisStore {
     async fn get(&self, key: &str) -> Result<Option<Vec<u8>>, Error> {
         let mut conn = self.connection.lock().await;
-        let result: Vec<u8> = conn.get(key).await.map_err(log_error)?;
-
-        if result.is_empty() {
-            Ok(None)
-        } else {
-            Ok(Some(result))
-        }
+        conn.get(key).await.map_err(log_error)
     }
 
     async fn set(&self, key: &str, value: &[u8]) -> Result<(), Error> {
--- a/crates/key-value-redis/src/lib.rs
+++ b/crates/key-value-redis/src/lib.rs
@@ -53,14 +53,14 @@ struct RedisStore {
 
 #[async_trait]
 impl Store for RedisStore {
-    async fn get(&self, key: &str) -> Result<Vec<u8>, Error> {
+    async fn get(&self, key: &str) -> Result<Option<Vec<u8>>, Error> {
         let mut conn = self.connection.lock().await;
         let result: Vec<u8> = conn.get(key).await.map_err(log_error)?;
 
         if result.is_empty() {
-            Err(Error::NoSuchKey)
+            Ok(None)
         } else {
-            Ok(result)
+            Ok(Some(result))
         }
     }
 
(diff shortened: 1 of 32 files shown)


wit/preview2/key-value.wit
```diff
@@ -1,48 +1,30 @@
 interface key-value {
   /// An open key-value store
   resource store {
-    /// Open the store with the specified name.
+    /// Open the store with the specified label.
```
itowlson, line 4: Should we also change the terminology in the function declaration (line 9)?

radu-matei (approved):

crates/key-value-redis/src/lib.rs
```diff
@@ -53,14 +53,14 @@ struct RedisStore {
...
+    async fn get(&self, key: &str) -> Result<Option<Vec<u8>>, Error> {
         let mut conn = self.connection.lock().await;
         let result: Vec<u8> = conn.get(key).await.map_err(log_error)?;
 
         if result.is_empty() {
-            Err(Error::NoSuchKey)
+            Ok(None)
         } else {
-            Ok(result)
+            Ok(Some(result))
```
lann, line 63: It looks like this will return `None` for empty values. Is that what we want here or should this instead be just:

```rust
   conn.get(key).await.map_err(log_error)
}
```

crates/key-value/src/lib.rs
```diff
@@ -193,7 +165,9 @@ impl spin_world::v1::key_value::Host for KeyValueDispatch {
     async fn get(&mut self, store: u32, key: String) -> Result<Result<Vec<u8>, LegacyError>> {
         let this = Resource::new_borrow(store);
         let result = <Self as key_value::HostStore>::get(self, this, key).await?;
-        Ok(result.map_err(to_legacy_error))
+        Ok(result
+            .map_err(to_legacy_error)
+            .map(|v| v.unwrap_or_default()))
```
lann, line 170: Should this map a `None` value to a `LegacyError::NoSuchKey`?

lann (commented):
A couple of questions about the "no such key" changes. Its little hard to follow the new logic in diff form :sweat_smile:

crates/key-value-redis/src/lib.rs
rylev: Great catch! The previous behavior was wrong it seems.

crates/key-value/src/lib.rs
rylev: Another good catch!

wit/preview2/key-value.wit
```diff
@@ -52,23 +34,13 @@ interface key-value {
...
-    /// The store handle provided is not recognized, i.e. it was either never
-    /// opened or has been closed.
-    invalid-store,
-
-    /// No key-value tuple exists for the specified key in the specified
-    /// store.
-    no-such-key,
-
     /// Some implementation-specific error has occurred (e.g. I/O)
     io(string)
```
lann, line 45: `other`?

lann (commented):
Sorry, didn't scroll alllll the way to the bottom before...

lann (approved):

Issue #1857, opened 2023-10-06
OCI configuration updates to support containerd


jsturtevant:
Based on the issue created in runwasi https://github.com/containerd/runwasi/issues/108 and the proposal to support [WASI  OCI artifacts in Containerd](https://docs.google.com/document/d/11shgC3l6gplBjWF1VJCWvN_9do51otscAm0hBDGSSAc) spins current OCI format won't work with out some minor modifications.

The approach outlined in the doc is to add `"artifactType": "application/vnd.bytecodealliance.component.v1+wasm"` and use  `config.mediaType` as `application/vnd.oci.image.config.v1+json`.  This allows for runtimes to still properly configure and launch the wasm applications (see https://github.com/containerd/runwasi/pull/147 as an example).  

This means the current spin configuration of `application/vnd.fermyon.spin.application.v1+config` would need to move to be a layer in the OCI artifact and artifact type would be specified.

An example of the image manifest might look like:

``` 
regctl manifest get localhost:5000/spin:oci
Name:                                localhost:5000/spin:oci
MediaType:                           application/vnd.oci.image.manifest.v1+json
ArtifactType:                        application/vnd.bytecodealliance.module.v1+wasm
Digest:                              sha256:c74cb8cb19ee2632b4ae825c71780f48e3424cae6fc1e865cb34b1f03da35c58
Total Size:                          839.916kB

Config:
  Digest:                            sha256:30beeddd35f0d22d3c068f29fc6fb41bb9461807e427d3ebdf34d3b922704a4b
  MediaType:                         application/vnd.oci.image.config.v1+json
  Size:                              136B

Layers:

  Digest:                            sha256:499237ead273693b70fb4110b179226fe857efb6da0865fd8f8c98437f8c4467
  MediaType:                         application/vnd.bytecodealliance.wasm.component.layer.v0+wasm
  Size:                              146.992kB

  Digest:                            sha256:270f42cf04ef6dd5b5d223d14bf9a6c703f1f0df7dded6db897997006fa2ca26
  MediaType:                         application/vnd.bytecodealliance.wasm.component.layer.v0+wasm
  Size:                              692.031kB

  Digest:                            sha256:96485f14df303540d3f687be5b663a629a150c96e27697e27b4d3d75f31910b7
  MediaType:                         application/vnd.fermyon.spin.application.v1+config
  Size:                              893B
```

[...]

PR #1882, opened 2023-10-11
feat(oci): manifest/config updates to support containerd

vdice:
Updates per https://github.com/fermyon/spin/issues/1857

- ~~Adds `artifactType: application/vnd.bytecodealliance.component.v1+wasm` to the OCI manifest~~ (https://github.com/fermyon/spin/pull/1882#discussion_r1362574046)
- Adds the locked spin config data as a separate layer
- Updates the OCI manifest config to represent the structure [expected by the spec](https://github.com/opencontainers/image-spec/blob/main/config.md)
  - ⚠️  **This is a breaking change:** older Spin clients won't be able to run Spin apps published by clients with this updated logic.
  - Currently this config uses defaults and we don't yet attempt to push any data specific to a Spin app.  Open to thoughts on what we might want to include.
- ~~Updates the wasm layer media type to `application/vnd.bytecodealliance.wasm.component.layer.v0+wasm`~~ (see https://github.com/fermyon/spin/pull/1882#issuecomment-1773232960)
  - ~~⚠️ **This is a breaking change:** older Spin clients won't be able to run Spin apps published by clients with this updated logic.~~
  - ~~Note that the [corresponding upstream PR](https://github.com/bytecodealliance/registry/pull/146) is still in flight -- is it too early to update the media type?~~
- ~~Adds a notion of the 'legacy' wasm layer media type such that new Spin clients can pull Spin apps published by older clients~~

### DIFF
--- a/crates/oci/Cargo.toml
+++ b/crates/oci/Cargo.toml
@@ -14,7 +14,7 @@ dkregistry = { git = "https://github.com/camallo/dkregistry-rs", rev = "37acecb4
 docker_credential = "1.0"
 dirs = "4.0"
 futures-util = "0.3"
-oci-distribution = { git = "https://github.com/fermyon/oci-distribution", rev = "639c907b7c0c4e74716356585410d4abe4aebf4d" }
+oci-distribution = { git = "https://github.com/fermyon/oci-distribution", rev = "63cbb0925775e0c9c870195cad1d50ac8707a264" }
 reqwest = "0.11"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/crates/oci/Cargo.toml
+++ b/crates/oci/Cargo.toml
@@ -14,7 +14,7 @@ dkregistry = { git = "https://github.com/camallo/dkregistry-rs", rev = "37acecb4
 docker_credential = "1.0"
 dirs = "4.0"
 futures-util = "0.3"
-oci-distribution = { git = "https://github.com/fermyon/oci-distribution", rev = "05022618d78feef9b99f20b5da8fd6def6bb80d2" }
+oci-distribution = { git = "https://github.com/fermyon/oci-distribution", rev = "639c907b7c0c4e74716356585410d4abe4aebf4d" }
 reqwest = "0.11"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -7,11 +7,8 @@ use docker_credential::DockerCredential;
 use futures_util::future;
 use futures_util::stream::{self, StreamExt, TryStreamExt};
 use oci_distribution::{
-    client::{Config, ImageLayer},
-    manifest::{OciImageManifest, OCI_IMAGE_MEDIA_TYPE},
-    secrets::RegistryAuth,
-    token_cache::RegistryTokenType,
-    Reference, RegistryOperation,
+    client::ImageLayer, config::ConfigFile, manifest::OciImageManifest, secrets::RegistryAuth,
+    token_cache::RegistryTokenType, Reference, RegistryOperation,
 };
 use reqwest::Url;
 use spin_common::sha256;
@@ -174,14 +171,15 @@ impl Client {
         );
         layers.push(locked_config_layer);
 
-        let oci_config = Config {
-            // TODO: now that the locked config bytes are pushed as a layer, what should data here be?
-            // Keeping as locked config bytes would make it feasible for older Spin clients to pull/run
-            // apps published by newer Spin clients
-            data: serde_json::to_vec(&locked)?,
-            media_type: OCI_IMAGE_MEDIA_TYPE.to_string(),
-            annotations: None,
+        // Construct empty/default OCI config file. Data may be parsed according to
+        // the expected config structure per the image spec, so we want to ensure it conforms.
+        // (See https://github.com/opencontainers/image-spec/blob/main/config.md)
+        // TODO: Explore adding data applicable to a Spin app.
+        let oci_config_file = ConfigFile {
+            ..Default::default()
         };
+        let oci_config =
+            oci_distribution::client::Config::oci_v1_from_config_file(oci_config_file, None)?;
         let manifest = OciImageManifest::build(&layers, &oci_config, None);
 
         let response = self
@@ -292,7 +290,6 @@ impl Client {
         // Older published Spin apps feature the locked app config *as* the OCI manifest config layer,
         // while newer versions publish the locked app config as a generic layer alongside others.
         // Assume that these bytes may represent the locked app config and write it as such.
-        // TODO: update this assumption if we change the data we write to the OCI manifest config layer.
         let mut cfg_bytes = Vec::new();
         self.oci
             .pull_blob(&reference, &manifest.config.digest, &mut cfg_bytes)
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -28,11 +28,8 @@ pub const SPIN_APPLICATION_MEDIA_TYPE: &str = "application/vnd.fermyon.spin.appl
 pub const DATA_MEDIATYPE: &str = "application/vnd.wasm.content.layer.v1+data";
 /// Media type for a layer representing a compressed archive of one or more files used by a Spin application
 pub const ARCHIVE_MEDIATYPE: &str = "application/vnd.wasm.content.bundle.v1.tar+gzip";
-// Legacy wasm layer media type used by pre-2.0 versions of Spin
-const WASM_LAYER_MEDIA_TYPE_LEGACY: &str = "application/vnd.wasm.content.layer.v1+wasm";
-
-// TODO: use canonical types defined upstream; see https://github.com/bytecodealliance/registry/pull/146
-const WASM_LAYER_MEDIA_TYPE: &str = "application/vnd.bytecodealliance.wasm.component.layer.v0+wasm";
+// Note: this will be updated with a canonical value once defined upstream
+const WASM_LAYER_MEDIA_TYPE: &str = "application/vnd.wasm.content.layer.v1+wasm";
 
 const CONFIG_FILE: &str = "config.json";
 const LATEST_TAG: &str = "latest";
@@ -326,7 +323,7 @@ impl Client {
                                 .await
                                 .with_context(|| "unable to write locked app config to cache")?;
                         }
-                        WASM_LAYER_MEDIA_TYPE | WASM_LAYER_MEDIA_TYPE_LEGACY => {
+                        WASM_LAYER_MEDIA_TYPE => {
                             this.cache.write_wasm(&bytes, &layer.digest).await?;
                         }
                         ARCHIVE_MEDIATYPE => {
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -8,7 +8,6 @@ use futures_util::future;
 use futures_util::stream::{self, StreamExt, TryStreamExt};
 use oci_distribution::{
     client::{Config, ImageLayer},
-    errors::OciDistributionError,
     manifest::{OciImageManifest, OCI_IMAGE_MEDIA_TYPE},
     secrets::RegistryAuth,
     token_cache::RegistryTokenType,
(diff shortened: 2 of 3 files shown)


crates/oci/Cargo.toml
```diff
@@ -25,7 +25,7 @@ spin-manifest = { path = "../manifest" }
 spin-trigger = { path = "../trigger" }
 tempfile = "3.3"
 tokio = { version = "1", features = ["fs"] }
-tokio-util = "0.7.9"
+tokio-util = { version = "0.7.9", features = ["compat"] }
```
vdice, line 28: This cropped up seemingly with the oci-distribution version bump, oddly.  (`compat` not found unless declared as a feature).  I can see about PRing it separately...

vdice: (Added in https://github.com/fermyon/spin/pull/1883)

crates/oci/src/client.rs
```diff
@@ -25,15 +24,19 @@ use walkdir::WalkDir;
...
 /// Media type for a layer representing a generic data file used by a Spin application
 pub const DATA_MEDIATYPE: &str = "application/vnd.wasm.content.layer.v1+data";
 /// Media type for a layer representing a compressed archive of one or more files used by a Spin application
 pub const ARCHIVE_MEDIATYPE: &str = "application/vnd.wasm.content.bundle.v1.tar+gzip";
+// Legacy wasm layer media type used by pre-2.0 versions of Spin
+const WASM_LAYER_MEDIA_TYPE_LEGACY: &str = "application/vnd.wasm.content.layer.v1+wasm";
+
+// TODO: use canonical types defined upstream; see https://github.com/bytecodealliance/registry/pull/146
+const WASM_LAYER_MEDIA_TYPE: &str = "application/vnd.bytecodealliance.wasm.component.layer.v0+wasm";
+const COMPONENT_ARTIFACT_TYPE: &str = "application/vnd.bytecodealliance.component.v1+wasm";
```
jsturtevant, line 39: Based on feedback from the OCI image spec maintainers, we should not be including the artifact type on the OCI Image manifest: https://github.com/opencontainers/image-spec/pull/1137#issuecomment-1760065985

crates/oci/src/client.rs
```diff
@@ -162,12 +165,24 @@ impl Client {
...
+        // Push layer for locked spin application config
+        let locked_config_layer = ImageLayer::new(
+            serde_json::to_vec(&locked).context("could not serialize locked config")?,
+            SPIN_APPLICATION_MEDIA_TYPE.to_string(),
+            None,
+        );
+        layers.push(locked_config_layer);
+
         let oci_config = Config {
+            // TODO: now that the locked config bytes are pushed as a layer, what should data here be?
```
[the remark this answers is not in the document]
vdice: Thanks @jsturtevant.  We're now constructing a default/empty config, albeit via the same [oci-distribution](https://github.com/fermyon/oci-distribution) crate we've been using elsewhere.  Should adhere to the spec (I'd hope!).  Would this suit parsing requirements on containerd's side?

vdice:
I'm happy with these changes if @jsturtevant is and we'd like to get them in before a forthcoming 2.0 release as they represent breaking changes.  My only reservation is around updating the wasm layer media type while its canonical value seems to still be under discussion.  Thoughts?

crates/oci/src/client.rs
```diff
@@ -164,12 +163,25 @@ impl Client {
...
+            SPIN_APPLICATION_MEDIA_TYPE.to_string(),
+            None,
+        );
+        layers.push(locked_config_layer);
+
+        // Construct empty/default OCI config file. Data may be parsed according to
+        // the expected config structure per the image spec, so we want to ensure it conforms.
+        // (See https://github.com/opencontainers/image-spec/blob/main/config.md)
+        // TODO: Explore adding data applicable to a Spin app.
+        let oci_config_file = ConfigFile {
```
jsturtevant, line 178: We should also set the `OS` and `ARCH`, on the image. Because we are moving away from using the `ArtifactType` (see discussions in the [spec repo](https://github.com/opencontainers/image-spec/pull/1137#issuecomment-1767144964)), we need a way to identify this image.  These values are [recommended ](https://github.com/opencontainers/image-spec/blob/main/config.md#properties) to be set to Go Language values.  

> Configurations SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [GOARCH](https://golang.org/doc/install/source#environment).

https://github.com/containerd/runwasi/blob/828f1cce03feb6a3b2d894c27624e7d7c5cef0c2/crates/wasi-demo-app/build.rs#L40-L43

In Runwasi, to start, we will be looking for `ARCH=wasm` in order to know we should read and pass the custom layers to the runtime.  We will read `OS` part and pass it to the runtime so it can be used to determine whether to use modules/components.

crates/oci/src/client.rs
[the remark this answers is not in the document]
jsturtevant: Woops just saw this I left a comment in https://github.com/fermyon/spin/pull/1882#pullrequestreview-1690420796

crates/oci/src/client.rs
vdice: In the linked runwasi code, I see "wasi" is used as OS, though the official valid combo from https://go.dev/doc/install/source#environment appears to be "wasip1" (w/ ARCH of `wasm`).  I gather for containerd's needs, "wasi" is okay?  I assume "wasip2" support (both in go and containerd/etc) is in the near future?

jsturtevant: The image was originally written before golang released the support for wasi and set the OS/ARCH values so we didn't know what it would be.  

As of right now, the OS flag isn't going to be used by runwasi/containerd but passed to the runtime to use to make discussions so TBH it doesn't matter _yet_ since runwasi/containerd are not triggering off it (for now).  That said, the code linked will be updated to be wasip1 (have pr open for it) since the module being built there is actually wasip1 via rust. With the wasmtime/wasmedge shims we will use this to determine if it is module or component and set up the runtime appropriately.

In the future (once things stabilize more) the OS value and possibly `os_features` will become more important as we hope to have more consistency across the eco system and to have containerd do more based on that information.  Right now, containerd simply passes this info onto the shim (runwasi).  I am working on drafting up a doc that will point to how this looks in the future and will share with you folks once ready.

Does Spin adhere to the P1 or have a custom format?

jsturtevant:
>  My only reservation is around updating the wasm layer media type while its canonical value seems to still be under discussion. Thoughts?

For the same reasons in https://github.com/fermyon/spin/pull/1882#discussion_r1367383430, runwasi and containerd are not currently opinionated about the layer media types.  Runwasi, will grab any layers that are not image types and pass them to the runtime, with the media type.  This may change in the future but for now while the eco system is still stabilizing I think it might be ok to keep them layers in the way you have them now in order to not break existing users.  

In the WARG/runwasi the plan is to start to use `application/vnd.bytecodealliance.wasm.component.layer.v0+wasm`.  Once we have a 1.0 for WASI and specific way to use this via containerd we will get to a canonical value.  Till then I think it this will be in flux, and as mentioned runwasi shims will use the OS (p1/p2) to do most of the distinction between the two formats.

crates/oci/src/client.rs
vdice: > Does Spin adhere to the P1 or have a custom format?

Well, that's a bit tricky.  The `.wasm` file(s) included in a Spin app may be preview 1 or may even be preview 2 depending on the language-specific build toolchain used.  Most likely preview 1 at this point.  But then, the higher-level Spin application itself, which this OCI image represents (eg consisting of one or more p1 or p2 `.wasm` files plus zero or more data files/layers and the locked spin config layer), would be considered a 'custom' format and require the Spin runtime (or conformant runtime) to run successfully.

jsturtevant: Since there is a mix, I think either wasip1 or wasip2 would make most sense.  And this is an example where I think we have a little time before everything stabilizes.  At that point I would think there would be "everything is a component" as mentioned when discussing layer types https://github.com/fermyon/spin/issues/1857#issuecomment-1755925324

vdice:
> This may change in the future but for now while the eco system is still stabilizing I think it might be ok to keep them layers in the way you have them now in order to not break existing users.

Thanks @jsturtevant; I've kept it at the current value for now and created another PR for Spin to [futureproof this a bit](https://github.com/fermyon/spin/pull/1941) so that when we do update the media type it won't necessarily be a breaking change.

jsturtevant:
great!  give me a day or so and I will give this a go with https://github.com/deislabs/containerd-wasm-shims/pull/164

vdice:
@jsturtevant thank you.  We have some urgency to get breaking changes merged before this week's Spin 2.0 release, so do let us know if anything else crops up when integrating with the shim.  I'm hoping any other required changes will be additive/non-breaking.

radu-matei:
End-to-end testing this branch looks great, including running existing apps from the registry that were previously pushed with Spin V1.

jsturtevant:
I was able to validate this with the latest runwasi in https://github.com/deislabs/containerd-wasm-shims/pull/164

```
spin registry push docker.io/jsturtevant/spin-wasm-shim:latest-2.0

 sudo ./ctr run --net-host --rm --runtime=io.containerd.spin.v1 docker.io/
jsturtevant/spin-wasm-shim:latest-2.0 testwasm shim.wasm

Serving http://0.0.0.0:3000
Available Routes:
  hello: http://0.0.0.0:3000/hello
  go-hello: http://0.0.0.0:3000/go-hello
Hello, world! You should see me in pod logs
```

crates/oci/src/client.rs
```diff
@@ -164,12 +160,27 @@ impl Client {
...
+        layers.push(locked_config_layer);
+
+        // Construct empty/default OCI config file. Data may be parsed according to
+        // the expected config structure per the image spec, so we want to ensure it conforms.
+        // (See https://github.com/opencontainers/image-spec/blob/main/config.md)
+        // TODO: Explore adding data applicable to the Spin app being published.
+        let oci_config_file = ConfigFile {
+            architecture: oci_distribution::config::Architecture::Wasm,
+            os: oci_distribution::config::Os::Wasip1,
+            ..Default::default()
```
jsturtevant, line 178: One minor but user experience would be to provide a default entry point https://github.com/containerd/runwasi/blob/2fb77889d5d7b0e590a574cfbc44089b54036811/crates/oci-tar-builder/src/bin.rs#L61C39-L61C39

if this isn't set then when running with ctr need to provide one: 



which is the `shim.wasm` in the command `sudo ./ctr run --net-host --rm --runtime=io.containerd.spin.v1 docker.io/
jsturtevant/spin-wasm-shim:latest-2.0 testwasm shim.wasm`

otherwise end up with 

```
time="2023-10-30T22:55:30.029486688Z" level=error msg="on non-Windows, at least one process arg entry is required"

time="2023-10-30T22:55:30.029555687Z" level=error msg="failed to initialize container process: missing args in the process spec"
```

radu-matei: @jsturtevant -- what exactly would be the meaning for the "default entrypoint" for a Spin app, which can have multiple components (which means no _one_ wasm component)?

jsturtevant: This value is passed as arg0 in wasmtime shim.  Maybe it isn't used spin shim?

radu-matei: My assumption is that it wouldn't be used in the Spin shim, as the information about components is part of the locked app manifest.

jsturtevant: Yes, I don't think it is blocking.  I was using the applications name in runwasi. I am afk, but can look what happens in the docker container scenario later

radu-matei (approved):

crates/oci/src/client.rs
jsturtevant: Spin currently ignores this value, runwasi parses it and sends to the shim via `RuntimeContext` but it is ignored by spin's shim: https://github.com/deislabs/containerd-wasm-shims/blob/6ca8df8247d94cf1dc6ce5e529b00ee728aee4ec/containerd-shim-spin-v1/src/engine.rs#L123

The reason this is working when packages as a container is because `/` is the command is set on the yaml: https://github.com/deislabs/containerd-wasm-shims/blob/6ca8df8247d94cf1dc6ce5e529b00ee728aee4ec/README.md?plain=1#L95

In the case of a shim such as wasmtime/wasmedge it makes sense to have pass the entrypoint info as `arg0` and we even do some additional parsing to allow for entering other functions in the wasm module https://github.com/containerd/runwasi/blob/642cafacde77d25762c8c0c0ca78ff1010caff16/crates/containerd-shim-wasm/src/container/context.rs#L17-L25

So in this case we can continue to use `/` in the command for the pod yaml or you could set this value in the config (maybe image name?) and users wouldn't need to set the command to `/` in the yaml.  In the future maybe it can be useful otherwise we can revisit this in the next iteration of the config support in containerd.

PR #1884, opened 2023-10-12
Run lints on Rust test cases

rylev:
Runs linting over "tests/testcases/*/Cargo.toml" and changes Rust linting in CI to happen all together.

Unfortunately the linting job takes much longer now which would delay the other jobs from getting started. Instead of gating on linting passing, we now let everything run and cancel if linting fails.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,13 +35,9 @@ jobs:
           rust-wasm: true
           rust-cache: true
 
-      - name: Cargo Format
+      - name: Run lints
         run:
-          BUILD_SPIN_EXAMPLES=0 cargo fmt --all -- --check
-
-      - name: Cargo Clippy
-        run:
-          BUILD_SPIN_EXAMPLES=0 cargo clippy --workspace --all-targets --features all-tests -- -D warnings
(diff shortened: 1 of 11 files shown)


dicej (approved):

.github/workflows/build.yml
```diff
@@ -35,19 +35,18 @@ jobs:
...
-      - name: Cargo Format
+      - name: Run lints
         run:
-          BUILD_SPIN_EXAMPLES=0 cargo fmt --all -- --check
+          BUILD_SPIN_EXAMPLES=0 make lint
 
-      - name: Cargo Clippy
-        run:
-          BUILD_SPIN_EXAMPLES=0 cargo clippy --workspace --all-targets --features all-tests -- -D warnings
+      - name: Cancel everything if this fails
```
michelleN, line 42: Super optional nit: `Cancel everything if lint fail`

michelleN (approved):

.github/workflows/build.yml
rylev: 👍 I'll add this to another PR so I don't need to run CI again just for that.

PR #1885, opened 2023-10-12
Check in Cargo.lock files for examples and test apps

rylev:
We currently inconsistently check in Cargo.lock files across examples and test apps. Whether to check in Cargo.lock files is an area of debate because there are advantages and disadvantages to both. [The Cargo team has officially said ¯\_(ツ)_/¯  to this question.](https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html). 

For our example and test apps, the ability to do reproducible builds is the biggest reason to check in the lock files. Everyone will now build these examples and test apps exactly the same way (even if a new semver compatible version of a dependency is released in the mean time). This , however, might make PRs more noisy as updating a dependency (especially through the Rust SDK) might cause a cascading effect that causes all the Cargo.lock files to need to update. I believe this cost is worth it.


dicej (approved):
Thanks for doing this!

michelleN (approved):

PR #1886, opened 2023-10-12
Use resources for Postgres API

rylev:
In much the same vane as #1873 - this changes the postgres interface to use a connection oriented resource. Other than that the API is largely the same with the minor exception of removing the "success" variant on the error type (which was renamed from `pg-error` to `error`).


crates/outbound-pg/src/lib.rs
```diff
@@ -35,27 +44,43 @@ impl HostComponent for OutboundPg {
...
+
+#[async_trait]
+impl v2::HostConnection for OutboundPg {
+    async fn open(&mut self, address: String) -> Result<Result<Resource<Connection>, v2::Error>> {
+        Ok(async {
+            self.connections
+                .push(
+                    build_client(&address)
+                        .await
+                        .map_err(|e| v2::Error::ConnectionFailed(format!("{:?}", e)))?,
```
lann, line 57:
```suggestion
                        .map_err(|e| v2::Error::ConnectionFailed(format!("{e:?}")))?,
```
:shrug:

crates/outbound-pg/src/lib.rs
```diff
@@ -35,27 +44,43 @@ impl HostComponent for OutboundPg {
...
+                )
+                .map(Resource::new_own)
+                .map_err(|_| v2::Error::OtherError("too many connections".into()))
```
lann, line 60: I know it was just a reading failure, but it took me a minute to realize that the `map_err` wasn't applying to `Resource::new_own`

```suggestion
                .map_err(|_| v2::Error::OtherError("too many connections".into()))
                .map(Resource::new_own)
```

crates/outbound-pg/src/lib.rs
```diff
@@ -325,3 +344,47 @@ impl std::fmt::Debug for PgNull {
...
+    ($self:ident.$name:ident($address:expr, $($arg:expr),*)) => {{
+        let connection = match <Self as v2::HostConnection>::open($self, $address).await? {
+            Ok(c) => c,
+            Err(e) => return Ok(Err(to_legacy_error(e))),
+        };
+        Ok(<Self as v2::HostConnection>::$name($self, connection, $($arg),*)
+            .await?
+            .map_err(|e| to_legacy_error(e)))
+    }};
+}
```
lann, line 359: I'm on the fence with these macros. It's a nice pattern when you're _writing_ the delegation, but I'm not sure its worth it for the next person who needs to maintain these (even if thats just you in 6 months).

wit/preview2/postgres.wit
```diff
@@ -0,0 +1,24 @@
+interface postgres {
+  use fermyon:spin/rdbms-types.{parameter-value, row-set}
+
+  /// Errors related to interacting with Postgres.
+  variant error {
+      connection-failed(string),
+      bad-parameter(string),
+      query-failed(string),
+      value-conversion-failed(string),
+      other-error(string)
```
lann, line 10: Just `other` to match the redis change?

lann (commented):
Mostly non-blocking nit/style comments. `other-error` vs `other` is the only one I really care about

wit/preview2/postgres.wit
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  /// A connection to a postgres database.
+  resource connection {
+    /// Open a connection to the Redis instance at `address`.
```
calebschoepp, line 15:
```suggestion
    /// Open a connection to the Postgres instance at `address`.
```

crates/outbound-pg/src/lib.rs
rylev: Perhaps I've written too many macros, but I personally find this one easy to understand. I also find it easier to understand the intent of the code which is that it really is meant to mechanically delegate. 

Perhaps we leave this in and see how the age? Inlining them should be easy to do if folks find them hard to understand in the future.

PR #1892, opened 2023-10-16
Wasi HTTP in SDK

rylev:
This is an extension of #1887 that targets the wasi-http handler. It moves most of the code for `wasi-http` handling into the SDK, and improves the programming model. 

This PR enables use of `wasi-http` inside of a handler the looks just like a regular http handler. The user can therefore adopt wasi-http features piecemeal and only be confronted with the complexity as they need to be.

The `wasi-http`-only handler style with a `ResponseOutparam` should only need to be used by users who require streaming outbound response bodies. For all other users, the more straightforward `fn handler(request: Request) -> Response` style handler can be used instead. 

I'm going to merge this into #1887 since those are intrinsically linked and fully reviewing one without the other is not really possible. I hope that even though the resulting PR is huge, it will still be possible to review, but hopefully this PR gives some insight into some of the WASI


examples/wasi-http-rust-streaming-outgoing-body/src/lib.rs
```diff
@@ -1,34 +1,32 @@
...
-    sha2::{Digest, Sha256},
-    spin_sdk::wasi_http_component,
-    std::str,
-    url::Url,
+use anyhow::{bail, Result};
+use futures::{stream, SinkExt, StreamExt, TryStreamExt};
+use spin_sdk::wasi_http::send;
+use spin_sdk::wasi_http::{
+    Fields, IncomingRequest, IncomingResponse, Method, OutgoingBody, OutgoingRequest,
+    OutgoingResponse, ResponseOutparam, Scheme,
```
itowlson, line 6: I know this already existed but would it be possible to change the name of `ResponseOutparam` before freezing this?  (Or is this low-level WIT detail and the SDK will provide a more conventionally named wrapper?)

rylev: This comes directly from `wasi-http`. We could alias to another name, but I want to be careful about creating a potentially confusing rename.

PR #1893, opened 2023-10-16
Update wasm-tools deps

rylev:
Depends on https://github.com/fermyon/spin-componentize/pull/29

Unfortunately the wasm-tools breakage is not fully resolved and this update is needed.

### DIFF
--- a/crates/core/tests/core-wasi-test/Cargo.toml
+++ b/crates/core/tests/core-wasi-test/Cargo.toml
@@ -7,6 +7,6 @@ edition = "2021"
 debug = true
 
 [dependencies]
-wit-bindgen = { git = "https://github.com/bytecodealliance/wit-bindgen", rev = "3fac9c9f4a06061a1cfae83de356cecad2d642d1" }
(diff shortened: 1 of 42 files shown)


crates/core/tests/core-wasi-test/Cargo.toml
```diff
@@ -7,6 +7,6 @@ edition = "2021"
 debug = true
 
 [dependencies]
-wit-bindgen = { git = "https://github.com/bytecodealliance/wit-bindgen", rev = "3fac9c9f4a06061a1cfae83de356cecad2d642d1" }
+wit-bindgen = { git = "https://github.com/bytecodealliance/wit-bindgen", rev = "2405a79c74c5d61b9bc88c378d475c6c21ed6a9f" }
```
endocrimes, line 10: Can we add a comment as to why we point to a rev rather than a release?

rylev: Added it to the SDK repo which is where we actually need this version - the rest of the uses are to avoid having multiple versions of wit-bindgen floating around. Hopefully wit-bindgen 0.13 will be released in the new few days and this will all go away.

lann (approved):
whyyyy

rylev:
> whyyyy

https://github.com/bytecodealliance/wit-bindgen/pull/701

PR #1901, opened 2023-10-17
No more http and `inbound-http` interfaces

### LINKED ISSUE #1871: Consider deprecation of old style `outbound-http` and `inbound-http`
With `wasi-http` affording the ability to do inbound and outbound http, there is little reason to have an the old style `outbound-http` and `inbound-http` interfaces. We should consider getting rid of them - potentially in Spin 2.0. Of course, we'll need to keep support around for legacy components, but perhaps this can be implemented in terms of wasi-http?

The only advantage those interfaces afford is that non-streaming bodies are simplier to use. However, we'll want to ensure that we make usage of the streaming APIs as simple as possible and having an additional two interfaces just for this is confusing and hard to maintain.

rylev:
Resolves #1871 (marking as a daft until that PR is merged)

## What's going on here?

Instead of having two interfaces (`wasi-http` and `http`) that both do the same thing, we are removing the `http` interface (and the `inbound-http` interface) in favor of only using `wasi-http`. **This only impacts the guest** - the host continues to support the `http` interface for older apps. 

This has little impact on the user's actual programming experience (at least for Rust) as the Rust SDK continues to allow a `Request` -> `Response` style handler even though `wasi-http` is built on a `Request` and `ResponseOutParam` style. This should make it considerably less confusing for users interacting with Spin WebAssembly components directly as they won't have to wonder why there is both `wasi-http` and `http`.

## Concerns

For other languages with less ability to neatly hide implementation details, we are forcing users to use the considerably more complex `wasi-http` (which is more complex because it supports streaming HTTP bodies), but I don't see why even languages like C shouldn't be able to put a simpler non-streaming-body API in front of the `wasi-http` API. 

## Open Questions

* Instead of the macro being `#[wasi_http_component]`, we could keep it as `#[http_component]` since there is no longer another http style to contend with. Should we?
* Should the manifest v2 default to `executor = { type = "wasi" }`?
* Should the SDK's module be `http` instead of `wasi_http`? We originally used the name `wasi_http` to differentiate from the Spin `http` interface but this is no longer necessary. Perhaps the shorter `http` is better?


itowlson:
> Instead of the macro being `#[wasi_http_component]`, we could keep it as `#[http_component]` since there is no longer another http style to contend with. Should we?

The `wasi` sure feels redundant if it's not being contrasted with another macro.  For me one consideration is the upgrade experience:

* If we _remove_ `http_component` and have _only_ `wasi_http_component`, upgrading the SDK immediately breaks the code with a "macro not found" error.  What detail does the user see?  Does the error steer them in the right direction to fix it?
* If we _keep_ the `http_component` name but change the behaviour to be the current `wasi_http_component`, upgrading the SDK... does what?  Does the app continue to work as before because we still support the request-response paradigm?  Or does it break in subtle ("hey this needs to return a Whatsit instead of a Doodad") or unsubtle (weird error, runtime malfeasance) ways?

A similar comment about `wasi` being redundant applies to the third open question.  It seems like it has value if contrasting with something else, but not if "this is just how we do HTTP now."

rylev:
### Decision and new status quo

@itowlson I've switched to having everything called `http` as I agree that the name `wasi_http` works best when its in contrast but when there is only module related to http, it should just be called `http`. 

As for what the upgrade experience is like - unfortunately the SDK has several breaking changes in it so users will most likely just get type errors. We should make sure we have docs on the changes to the SDK so that users know how to upgrade. That being said, the changes are mostly small and trivial and the overall "shape" of handlers remains the same. I was able to upgrade one of my apps very quickly just following Rust compiler errors. 

Once the user has fixed the type errors, there is one runtime issue that they will run into: 

```
Error processing request: no fermyon:spin/inbound-http instance found
```

They must update the `[component.trigger]` config to include `executor = { type = "wasi" }`. I've added more detail to the error if we encounter that they indeed have a wasi-http handler exported but are still configured to use `fermyon:spin/inbound-http`:

```
Error processing request: no fermyon:spin/inbound-http instance found but wasi:http/incoming-handler was found instead.
Make sure to put `executor = { type = "wasi" }` under [component.trigger] in the spin.toml manifest.
```

### How can we improve?

There are two ways we can make this better:
* Default to `executor = { type = "wasi" }` for manifest v2. Unfortunately, I don't this is possible right now since that config is [shared between both manifest versions](https://github.com/fermyon/spin/blob/06528c5e2b754baadadf2a80ac10d5f34c174f7a/crates/http/src/config.rs#L25). I'm not sure how hard this would be to change so that the config could have a different default depending on the version. cc @lann 
* My preferred solution: we remove `executor = { type = "wasi" }` completely and rely on inspecting the component binary to know whether to use `fermyon:spin/inbound-http` or `wasi:http/incoming-handler` (with a default to `wasi:http/incoming-handler` should both be present). cc @dicej 

Thoughts?

itowlson:
@rylev "follow the compiler errors" is what I hoped for - delighted to see this and +1 to inspecting the binary if we can

rylev:
Tests are failing because we haven't implemented the outbound http calls to the same service yet in wasi-http. I'll do that tomorrow.

crates/http/src/config.rs
```diff
@@ -21,14 +21,14 @@ pub struct HttpTriggerConfig {
...
 #[serde(deny_unknown_fields, rename_all = "lowercase", tag = "type")]
 pub enum HttpExecutorType {
-    /// The component implements the Spin HTTP interface.
+    /// The component implements an HTTP based interface.
+    ///
+    /// This can be either `fermyon:spin/inbound-http` or `wasi:http/incoming-handler`
     #[default]
-    Spin,
+    #[serde(alias = "spin")]
+    Http,
```
rylev, line 29: @lann  do you hate this?

lann: I considered hating it briefly but decided not to. In practice its really just "wagi or default" anyway.

crates/loader/tests/ui/valid-manifest.lock
```diff
@@ -24,7 +24,7 @@
       "trigger_config": {
         "component": "four-lights",
         "executor": {
-          "type": "spin"
+          "type": "http"
```
lann, line 27: Hmm actually I might need to hate this if its going to be backward compat problem.

rylev: I guess this does mean that Spin 1.x can't read locked apps from Spin 2.x but I think that's fine?

lann: Oh yeah, I guess the alias makes it forward compatible. :+1:

PR #1902, opened 2023-10-17
Adjust outbound HTTP Requests in wasi-http

rylev:
Changes the outbound HTTP API.

The `send` function takes any type that implements `Sendable` which are all request like types. It then returns the result of trying to turn the response into the desired type. See some examples for how this will work. 

This API isn't the cleanest unfortunately, but so far I've not come up with anything better.

### DIFF
--- a/examples/http-rust-outbound-http/outbound-http-to-same-app/src/lib.rs
+++ b/examples/http-rust-outbound-http/outbound-http-to-same-app/src/lib.rs
@@ -13,7 +13,7 @@ async fn send_outbound(_req: Request) -> Result<impl IntoResponse> {
             .uri("/hello") // relative routes are not yet supported in cloud
             .body(())?,
     )
-    .await?;
+    .await??;
     res.headers_mut()
         .insert("spin-component", "rust-outbound-http".try_into()?);
     println!("{:?}", res);
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -3,6 +3,7 @@ pub mod conversions;
 
 use std::collections::HashMap;
 
+use async_trait::async_trait;
 #[doc(inline)]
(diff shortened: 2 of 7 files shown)


sdk/rust/src/http.rs
```diff
@@ -482,46 +510,119 @@ impl ResponseOutparam {
...
         let mut body = response.take_body();
         self.set(response);
         body.send(buffer).await
     }
 }
 
-/// Send an outgoing request
-pub async fn send<I, O>(request: I) -> Result<O, SendError>
+/// Send a request and get a response
+pub async fn send<S, R>(request: S) -> Result<Result<R, R::Error>, Error>
```
lann, line 522: What made this necessary? Are there conversions that can't do the boxed errors?

examples/http-rust-outbound-http/outbound-http-to-same-app/src/lib.rs
```diff
@@ -13,7 +13,7 @@ async fn send_outbound(_req: Request) -> Result<impl IntoResponse> {
             .uri("/hello") // relative routes are not yet supported in cloud
             .body(())?,
     )
-    .await?;
+    .await??;
```
itowlson, line 16: I worry the double question mark is going to be a paper cut for users..?

sdk/rust/src/http.rs
rylev: No, it was a (admittedly fairly unsuccessful) attempt at simplifying the signature.

lann: Yeah doesn't seem worth the doubled Results.

michelleN (commented):
Just tested `/outbound-to-hello-component` and the body is not coming through in the response

rylev:
Closing this since most of the changes were landed in #1975

PR #1903, opened 2023-10-17
ci: Print dirty lint paths

lann:
This should make it easier to fix "Git working tree dirtied by lints".

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Run lints
         run: |
             BUILD_SPIN_EXAMPLES=0 make lint
-            git diff --quiet . || (echo "Git working tree dirtied by lints. Run `make lint` and check in changes" && exit 1)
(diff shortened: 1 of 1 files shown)


rylev (approved):

rylev:
Just FYI: running `make lint` should automatically update `Cargo.lock` files so hopefully you shouldn't need to know which files are dirtied?

lann:
> Just FYI: running `make lint` should automatically update `Cargo.lock` files so hopefully you shouldn't need to know which files are dirtied?

`make lint` can take a long time.

michelleN (approved):

itowlson:
I had an action fail on the lints and the output was:

image

As far as I can tell, this commit was included (action was https://github.com/fermyon/spin/actions/runs/6555744599/job/17804658720, PR https://github.com/fermyon/spin/pull/1905).  Should I have seen the paths needing amending?

(Yes I am running `make lint` now to find out what I did wrong, I am just checking in that this is the expected behaviour of the "print dirty lint paths" tweak.)

PR #1906, opened 2023-10-18
Decouple LockedApp schema from `spin-core`

itowlson:
This is another experiment in decoupling the all-important `LockedApp` object from `spin-core` (and therefore from wasmtime and all its instruments of CPU heating).  The previous unpublished efforts were (1) feature flag to compile a version of `spin-app` without the `spin-core` dependency (intrusive and felt like a maintenance catastrophe waiting to happen) and (2) similar to this but without `spin-app` re-exporting its existing API (larger blast radius).

I don't know if this is the right answer - a whole crate seems pretty heavyweight, and maybe harder to understand than a feature - so if other folks have better ideas then bring 'em on, I'm happy to do what works.


crates/locked-app/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+pub const OCI_IMAGE_DIGEST_KEY: MetadataKey = MetadataKey::new("oci_image_digest");
+
+/// Type alias for a [`Result`]s with [`Error`].
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Errors returned by methods in this crate.
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// An error propagated from the [`spin_core`] crate.
+    #[error("spin core error: {0:#}")]
```
rylev, line 33: Maybe just `#[transparent]`?

rylev (approved):
This seems reasonable to me (though I think I might generally be more comfortable with small crates than others).

One thing we could consider doing in addition at some point is breaking off the wasmtime related functionality from `spin-core` and putting that into a `runtime` crate. We could potentially re-export it from `spin-core` under a feature flag to easily allow dependents to decide whether they need the runtime or not.

lann:
I'm generally fine with this, and in the long term I'd like to get rid of the `spin-app` crate entirely; I'm not really happy with the `App` abstraction in there.

crates/locked-app/src/lib.rs
itowlson: This was intended as no change to existing behaviour, but if @lann feels this would be a good change then I can bundle it in here for sure.

lann: Probably fine :shrug:

crates/app/src/lib.rs
```diff
@@ -7,22 +7,21 @@
 #![deny(missing_docs)]
 
 mod host_component;
-pub mod locked;
-mod metadata;
-pub mod values;
+pub use spin_locked_app::locked;
+pub use spin_locked_app::values;
+pub use spin_locked_app::{Error, MetadataKey, Result};
```
lann, line 12: Do we need these to be `pub`?

crates/locked-app/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+//! Spin internal application interfaces
+//!
+//! This crate contains interfaces to Spin application configuration to be used
+//! by crates that implement Spin execution environments: trigger executors and
+//! host components, in particular.
+
+#![deny(missing_docs)]
+
+pub mod locked;
```
lann, line 9: {C,Sh}ould this module be flattened into the crate root? (perhaps be `pub use`ing all its interesting bits)

crates/locked-app/src/lib.rs
itowlson: Thanks, done

crates/app/src/lib.rs
itowlson: The intent here was to re-export the things that `spin-app` previously exported, so that other crates didn't need to import both and change a bunch of references.  (My previous crack at this did _not_ re-export, and resulted in niggly dependency changes throughout the codebase, which is why I went this route this time.)

crates/locked-app/src/lib.rs
itowlson: It could; again, this seemed like a way to minimise code changes elsewhere.  Most Spin crates continue to use the full-fat `spin-app` crate, and it just seemed easier to keep things very much the same.

lann (approved):
Modules paths can always be mucked about with another time.

PR #1907, opened 2023-10-18
Add make command for updating all lock files

rylev:
Instead of doing the full `clippy` and `rustfmt` run, we can now run `make update-cargo-locks` which will do the minimal work to update the lock files.


itowlson:
(seriously, thanks - this is way friendlier)

PR #1913, opened 2023-10-19
Break `loader` dependency on wasmtime

itowlson:
I think this will leave cloud plugin (after `manifest-v2`) depending on `wasmtime` only via `http`.

I didn't want to use features for this, because it's impossible to test within Spin that the featured subset is consistent.  But it felt awkward breaking `AllowedHttpHosts` out into its own crate, and no other crate felt like a good home for it.  (It really is a specific concern of the outbound HTTP subsystem - `loader` only wants to validate that it's legal.)  I'd love for someone to suggest a better way!

The big file change is just a move of the diffed text, no changes within it, to allow me to feature at the module level instead of featuring individual bits.


crates/outbound-http/src/lib.rs
```diff
@@ -1,179 +1,12 @@
 pub mod allowed_http_hosts;
+#[cfg(feature = "runtime")]
 mod host_component;
+#[cfg(feature = "runtime")]
+mod outbound_http;
```
rylev, line 5: The module name is a bit awkward since it's also the name of the crate. Perhaps `host_impl`?

rylev (approved):
This seems reasonable. I'm approving but I would love to see my one nit addressed!

lann (approved):

crates/outbound-http/src/lib.rs
itowlson: Your nit is my command - done!

PR #1914, opened 2023-10-19
Fix "duplicate package" warning when referencing Spin crates

itowlson:
When two Rust examples in the Spin repo have the same package name, anyone compiling Rust against _any_ Spin repo GitHub reference - whether Spin crates or the Rust SDK - gets a "warning: skipping duplicate package `...`" message.

We had another `http-rust` one sneak in recently: this should fix it.


rylev (approved):
Whoops! Thank you!

PR #1918, opened 2023-10-19
Rename config to variables

fibonacci1729:
* Renames environment variables `SPIN_CONFIG_*` to `SPIN_VARIABLE_*` for the `EnvVariablesProvider`.
* Renames runtime config manifest key from `config_provider` to `variables_provider`.
* Moves `crates/config` to `crates/variables`.
* All tests/examples updated.


fibonacci1729:
On a less important note, the environment runtime variable provider is now confusingly named `EnvVariableProvider`. Open to naming suggestions if this irks others as well.

crates/trigger/src/runtime_config/variable_provider.rs
```diff
@@ -32,7 +34,7 @@ impl ConfigProviderOpts {
 
 #[derive(Debug, Default, Deserialize)]
 #[serde(deny_unknown_fields)]
-pub struct EnvConfigProviderOpts {
+pub struct EnvVariableProviderOpts {
```
fibonacci1729, line 37: Referring to this: https://github.com/fermyon/spin/pull/1918#issuecomment-1771426654

crates/variables/src/lib.rs
```diff
@@ -140,7 +140,7 @@ impl<'a> Key<'a> {
                 .chars()
                 .find(|c| !(c.is_ascii_lowercase() || c.is_ascii_digit() || c == &'_'))
             {
-                Err(format!("invalid character {:?}. Variable names and config keys may contain only lower-case letters, numbers, and underscores.", invalid))
+                Err(format!("invalid character {:?}. Variable names and keys may contain only lower-case letters, numbers, and underscores.", invalid))
```
itowlson, line 143: "and keys" is probably redundant now

crates/variables/src/lib.rs
```diff
@@ -163,31 +163,27 @@ impl<'a> AsRef<str> for Key<'a> {
...
 type Result<T> = std::result::Result<T, Error>;
 
-/// A config resolution error.
+/// A variable resolution error.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
-    /// Invalid config key.
-    #[error("invalid config key: {0}")]
+    /// Invalid variable key.
+    #[error("invalid variable key: {0}")]
```
itowlson, line 170: key -> name now?

fibonacci1729: This tempts me to rename the `Key` wrapper struct to `VariableName`. Thoughts or am I just OCD?

itowlson: I'd probably leave it for now.  It's purely internal, it's not doing any harm, and it would create even more churn in a PR that already has wide impact.  But conversely, it's a fairly localised change so won't make rebases around this PR any harder.  So if you are keen then go for it!

crates/trigger/src/runtime_config/variable_provider.rs
lann: Seems fine.

crates/trigger/src/runtime_config/variable_provider.rs
```diff
@@ -1,28 +1,30 @@
...
-use spin_config::provider::{env::EnvProvider, vault::VaultProvider};
+use spin_variables::provider::{env::EnvProvider, vault::VaultProvider};
 
 use super::RuntimeConfig;
 
-pub type ConfigProvider = Box<dyn spin_config::Provider>;
+pub type VariableProvider = Box<dyn spin_variables::Provider>;
 
-const DEFAULT_ENV_PREFIX: &str = "SPIN_CONFIG";
+const DEFAULT_ENV_PREFIX: &str = "SPIN_VARIABLE";
```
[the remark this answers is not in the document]
lann: I think we should first try `SPIN_VARIABLE_`, then try `SPIN_CONFIG_` and _if found, print a deprecation notice_.

lann:
:bike::derelict_house:: Variable**s**Provider?

crates/trigger/src/runtime_config.rs
```diff
@@ -198,8 +198,8 @@ pub struct RuntimeConfigOpts {
     #[serde(default)]
     pub llm_compute: Option<LlmComputeOpts>,
 
-    #[serde(rename = "config_provider", default)]
-    pub config_providers: Vec<ConfigProviderOpts>,
+    #[serde(rename = "variable_provider", default)]
```
lann, line 201: :bike::derelict_house::
```suggestion
    #[serde(rename = "variables_provider", default)]
```

fibonacci1729: I originally did this but decided to drop the plurality because it started to sound funny in my head.

lann: I think it reads better to someone who hasn't recently gazed into the variables abyss.

fibonacci1729: Ok i'm just going to undo all the switches from `variables*` to `variable*` that i made this morning. So going back to `VariablesProvider`.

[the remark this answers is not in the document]
fibonacci1729: Updated this to use an alias and added a test.

crates/variables/src/host_component.rs
```diff
@@ -89,11 +90,41 @@ impl config::Host for ComponentConfig {
...
+            let component_id = self.component_id.as_deref().unwrap();
+            let key = Key::new(&key)?;
+            Ok(self
+                .resolver
+                .get()
+                .unwrap()
+                .resolve(component_id, key)
+                .await?)
+        }
+        .await)
```
rylev, line 110:
```suggestion
        Ok(Ok(<Self as variables::Host>::get(self, key).await??))
```

crates/variables/src/host_component.rs
```diff
@@ -89,11 +90,41 @@ impl config::Host for ComponentConfig {
...
+}
+
+impl From<Error> for spin_world::v1::config::Error {
     fn from(err: Error) -> Self {
         match err {
-            Error::InvalidKey(msg) => Self::InvalidKey(msg),
-            Error::InvalidSchema(msg) => Self::InvalidSchema(msg),
+            Error::InvalidName(msg) => Self::InvalidKey(msg),
             Error::Provider(msg) => Self::Provider(msg.to_string()),
             other => Self::Other(format!("{}", other)),
```
rylev, line 129: Why do we have a catch-all pattern here instead of exhaustively matching?

crates/variables/src/host_component.rs
```diff
@@ -89,11 +90,41 @@ impl config::Host for ComponentConfig {
...
+    }
+}
+
+impl From<Error> for spin_world::v1::config::Error {
     fn from(err: Error) -> Self {
         match err {
-            Error::InvalidKey(msg) => Self::InvalidKey(msg),
-            Error::InvalidSchema(msg) => Self::InvalidSchema(msg),
+            Error::InvalidName(msg) => Self::InvalidKey(msg),
             Error::Provider(msg) => Self::Provider(msg.to_string()),
```
rylev, line 128: Does this not work?

```suggestion
            Error::Provider(msg) => Self::Provider(msg),
```

crates/variables/src/host_component.rs
```diff
@@ -89,11 +90,41 @@ impl config::Host for ComponentConfig {
...
+    fn from(err: Error) -> Self {
+        match err {
+            Error::InvalidName(msg) => Self::InvalidName(msg),
+            Error::Provider(msg) => Self::Provider(msg.to_string()),
+            other => Self::Other(format!("{}", other)),
+        }
+    }
+}
+
+impl From<Error> for spin_world::v1::config::Error {
```
rylev, line 124: I know this existed before this PR, but I find it confusing that we have an additional `Error` type beyond the wit defined one even though it seems to have the same shape as the wit error.

examples/config-rust/.env
```diff
@@ -1 +0,0 @@
-SPIN_CONFIG_DOTENV = "dotenv"
```
rylev, line 1: Is this file still there but empty? Is that what we want?

examples/spin-timer/deps/spin/variables.wit
```diff
@@ -0,0 +1,13 @@
+package fermyon:spin@2.0.0
+
+interface variables {
+    // Get an application variable value for the current component.
+    // The name must match one defined in in the component manifest.
```
rylev, line 5:
```suggestion
    /// Get an application variable value for the current component.
    ///
    /// The name must match one defined in in the component manifest.
```

examples/spin-timer/deps/spin/variables.wit
```diff
@@ -0,0 +1,13 @@
...
+    get: func(name: string) -> result<string, error>
+
+    variant error {
```
rylev, line 8: Can you add doc comments for this error and all the variants?

examples/spin-timer/deps/spin/variables.wit
```diff
@@ -0,0 +1,13 @@
...
+        invalid-name(string),
+        provider(string),
```
rylev, line 10: I don't love the name `provider`, but I'm not sure I have a better name. The verbose option (that I'm not really suggesting) would be `provider-specific-error`. Hmm....

examples/spin-timer/deps/spin/variables.wit
```diff
@@ -0,0 +1,13 @@
...
+        other(string),
+    }
```
rylev, line 12: Do we want to have a `not-found` error for when the requested variable is not found? Right now this gets encoded as a `provider` error which doesn't seem right.

sdk/rust/src/lib.rs
```diff
@@ -95,12 +95,8 @@ pub mod pg;
 /// Implementation of the Spin MySQL database interface.
 pub mod mysql;
 
-/// Implementation of the spin config interface.
-#[allow(missing_docs)]
-pub mod config {
-    /// Exports the generated Spin config items.
-    pub use super::wit::v1::config::{get_config as get, Error};
-}
+/// Implementation of the spin variables interface.
```
rylev, line 98: Can we do `#[doc(inline)]` instead and use doc comments on the variables interface to populate this?

examples/spin-timer/deps/spin/variables.wit
```diff
@@ -0,0 +1,13 @@
+package fermyon:spin@2.0.0
```
rylev, line 1: All of these comments obviously mostly apply to the wit file in "wit/preview2" (which I originally thought this file was).

rylev (changes requested):
Looking good!! I have some small things I think we should address before we merge this though.

examples/spin-timer/deps/spin/variables.wit
lann: I left it out because I didn't want to imply that `not-found` was a "normal" condition (like the KV v1 equivalent); not being able to resolve a variable is a programming or operational mistake.

examples/config-rust/.env
fibonacci1729: `config-rust` was renamed to `variables-rust` so this file exists there and the change is to use `SPIN_VARIABLE_DOTENV`.

examples/spin-timer/deps/spin/variables.wit
fibonacci1729: I don't personally have a strong preference here (i like `provider` but probably just comfortable staring at by now :p) but if you have a concrete suggestion i'm willing to use it!

examples/spin-timer/deps/spin/variables.wit
fibonacci1729: Yep (sorry it's confusing this WIT exists in an example because pre-registry days). But nonetheless they should be identical barring the package statement.

crates/variables/src/host_component.rs
fibonacci1729: Similar to my comment below about the `Error` type, it contains variants that aren't actionable by (and will never reach) the guest code.

crates/trigger/src/runtime_config/variable_provider.rs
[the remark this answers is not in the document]
fibonacci1729: @itowlson added a deprecation check, let me know your thoughts on the language!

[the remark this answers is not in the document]
fibonacci1729: Sample output:

```
❯ SPIN_CONFIG_OBJECT="teapot" spin up

Serving http://127.0.0.1:3000
Available Routes:
  spin-variales-rust: http://127.0.0.1:3000 (wildcard)
Use of environment provided variables prefix SPIN_CONFIG is deprecated
```

crates/variables/src/provider/env.rs
```diff
@@ -5,31 +5,58 @@ use async_trait::async_trait;
...
+
+        let upper_key = key.as_ref().to_ascii_uppercase();
+        let env_key = format!("{prefix}_{upper_key}");
+
+        match self.query_env(&env_key)? {
+            None if use_fallback => {
+                let env_key = format!("{LEGACY_ENV_PREFIX}_{upper_key}");
+                match self.query_env(&env_key) {
+                    result @ Ok(Some(_)) => {
+                        eprintln!("Use of environment provided variables prefix SPIN_CONFIG is deprecated");
```
rylev, line 53: Nit: I personally find this clearer

```suggestion
                        eprintln!("Use of deprecated prefix 'SPIN_CONFIG' for Spin variables provided by environment variables.");
```

crates/variables/src/provider/env.rs
```diff
@@ -5,31 +5,58 @@ use async_trait::async_trait;
...
+                        result
+                    }
+                    other => other,
+                }
```
rylev, line 57:
```suggestion
                let result = self.query_env(&env_key);
                if let Ok(Some(_)) = &result {
                    eprintln!("Use of environment provided variables prefix SPIN_CONFIG is  deprecated");
                }
                result
```

examples/spin-timer/deps/spin/variables.wit
rylev: It's strange that this error is labeled as a provider error when it's fairly likely *not* a provider error. I think the most likely cause will be users thinking there is a variable name `foo` and there not actually being one. 

It's not a huge deal that this gets rendered as `provider("no variable named 'foo'")` - the error is clear enough - it's just a bit strange is all. 

I personally don't really see potential confusion from having another error case, but I guess I'm fine letting this go if everyone disagrees with me.

rylev (approved):

lann (approved):

crates/trigger/src/runtime_config/variable_provider.rs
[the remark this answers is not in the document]
itowlson: I wonder if it would be okay to be a bit more specific than that e.g.

> Warning: variable 'foo': SPIN_VARIABLE_FOO was not set, so used SPIN_CONFIG_FOO. The SPIN_CONFIG prefix is deprecated; please switch to the SPIN_VARIABLE prefix.

At some point we should write a docs page on v1-to-v2 changes, and when we do that we can link it in the warning.

(I do worry it might be a bit wordy if it was printed repeatedly!  But we can see how that feels in practice and get clever if it proves bothersome.)

crates/loader/Cargo.toml
```diff
@@ -27,6 +27,7 @@ shellexpand = "3.1"
 spin-locked-app = { path = "../locked-app" }
 spin-common = { path = "../common" }
 spin-manifest = { path = "../manifest" }
+spin-variables = { path = "../variables" }
```
itowlson, line 30: This loader dependency appears to be a new dependency rather than a rename.  Have we really introduced this dependency?  There seem to be no associated code changes so I'm not sure why it would be needed.

crates/trigger/src/runtime_config.rs
```diff
@@ -57,12 +57,12 @@ impl RuntimeConfig {
         Ok(())
     }
 
-    /// Return a Vec of configured [`spin_config::Provider`]s.
-    pub fn config_providers(&self) -> Vec<ConfigProvider> {
-        let default_provider = ConfigProviderOpts::default_provider_opts(self).build_provider();
-        let mut providers: Vec<ConfigProvider> = vec![default_provider];
+    /// Return a Vec of configured [`spin_variables::Provider`]s.
```
itowlson, line 60: This was existing code but it's a bit odd that the doc string says the Vec is of `spin_variables::Provider` and the very next line says it's of `VariablesProvider`!

crates/trigger/src/runtime_config.rs
```diff
@@ -198,8 +198,8 @@ pub struct RuntimeConfigOpts {
     #[serde(default)]
     pub llm_compute: Option<LlmComputeOpts>,
 
-    #[serde(rename = "config_provider", default)]
-    pub config_providers: Vec<ConfigProviderOpts>,
+    #[serde(alias = "config_provider", rename = "variables_provider", default)]
```
itowlson, line 201: super nit: consider swapping the order so the "main" name is first 

```suggestion
    #[serde(rename = "variables_provider", alias = "config_provider", default)]
```

itowlson (approved):
A couple of nits but otherwise LGTM!

crates/loader/Cargo.toml
fibonacci1729: Oh! Good catch, must have missed this when rebasing onto the v2 manifest.

crates/trigger/src/runtime_config.rs
fibonacci1729: Updated to `VariablesProvider`.

PR #1927, opened 2023-10-23
doctor: Unbreak `spin doctor` for Manifest V2

lann:
This doesn't add any V2 functionality, but prevents spurious errors on valid V2 manifests.

Also simplifies `Checkup::for_each_diagnosis` into `Checkup::next_diagnosis`, which is much less finicky and not really any harder to use.

### DIFF
--- a/crates/doctor/tests/data/manifest_version_wrong_value.toml
+++ b/crates/doctor/tests/data/manifest_version_wrong_value.toml
@@ -1,4 +1,4 @@
-spin_manifest_version = 2
+spin_manifest_version = 1
 
 # comment preserved
 name = "app-name"
--- a/crates/doctor/tests/ui.rs
+++ b/crates/doctor/tests/ui.rs
@@ -1,0 +1,50 @@
+use std::path::PathBuf;
+
+use spin_doctor::{Checkup, PatientDiagnosis};
+use ui_testing::{Failed, UiTestsRunner};
+
+fn main() -> anyhow::Result<()> {
+    let mut runner = UiTestsRunner::default();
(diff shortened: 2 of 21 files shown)


crates/doctor/tests/ui.rs
lann, line 1: For now this just captures one change to the version diagnosis plus some smoke tests, but I think I'd like to migrate all the tests in this crate to this style.

crates/doctor/tests/data/manifest_version_wrong_value.toml
```diff
@@ -1,4 +1,4 @@
-spin_manifest_version = 2
+spin_manifest_version = 1
```
lann, line 1: Just the wrong _type_ now.

src/commands/doctor.rs
```diff
@@ -35,55 +34,48 @@ impl DoctorCommand {
             icon = Emoji("🩺 ", "")
         );
 
-        let count = spin_doctor::Checkup::new(manifest_file)
-            .for_each_diagnosis(move |diagnosis, patient| {
-                async move {
-                    show_diagnosis(&*diagnosis);
+        let mut checkup = spin_doctor::Checkup::new(manifest_file)?;
+        let mut has_problems = false;
+        while let Some(PatientDiagnosis { diagnosis, patient }) = checkup.next_diagnosis().await? {
```
lann, line 39: Most of the changes below are just reformatting due to un-nesting here.

itowlson (approved):
anything involving un-nesting is an instant approve

PR #1930, opened 2023-10-24
Add example for streaming a file in Rust

radu-matei:
This commit adds a simple example showing file streaming in Rust.


examples/wasi-http-streaming-file/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[http_component]
+async fn handler(req: IncomingRequest, res: ResponseOutparam) {
+    stream_file(req, res).await.unwrap();
+}
+
+async fn stream_file(_req: IncomingRequest, res: ResponseOutparam) -> Result<()> {
+    let response = OutgoingResponse::new(
+        200,
+        &Fields::new(&[(
```
rylev, line 20: FYI: `Headers` is an alias for `Fields` that we should probably prefer.

dicej (approved):

rylev (approved):

rylev:
Approved this but I do want us to come up with a better way to handle errors other than unwrapping.

Issue #1929, opened 2023-10-24
HeaderValues in the Rust SDK are not Strings anymore


mikkelhegn:
In Canary (as of Oct 24th), the header values returned by the Rust SDK are `Vec<u8>` instead of `String`.

According to @tschneidereit it's because:

> Header values can contain arbitrary values and don't need to be valid unicode strings. Hence, wasi-http encodes them as Vec<u8> instead of String.
> We should probably handle this the same [way](https://docs.rs/http/latest/http/header/struct.HeaderValue.html) the http crate does.

lann:
I believe #1939 addresses this as well as possible for now.

PR #1931, opened 2023-10-24
No custom `Request`/`Response` type

rylev:
Fixes https://github.com/fermyon/spin/issues/1929

This gets rid of the custom `Request` and `Response` types in favor of using the `http` crate. 

**Why**

This change avoids custom `Requst` and `Response` types. Creating general purpose `Request` and `Response` types is not trivial and would potentially just see us recreating the `http` crate. Instead of this, we rely on the ecosystem standard `http` crate while allowing for future expansion beyond that crate (through other types implementing the various conversion traits). 

**Considerations**

* The SDK does not re-export the `http` crate so users will need to directly depend on the crate themselves.
* Usage of the `http` crate is not *mandatory* though with the current API, it's very difficult to avoid. This is because the request and response types can be any types that implement the appropriate conversion traits. In the future as we support more types that implement these traits, user reliance on the `http` crate may decrease. 
* The `http` crate is not the world's most user-friendly crate. Unfortunately, this means that the Spin SDK de-facto inherits any design worts of the `http` crate. 
* In particular, `http::Response` relies on a builder pattern without very many nice shortcuts. So while previously the user could write `Request::new(200, "my body")` they must now write `Request::builder().status(200).body("my body")?`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6023,7 +6023,6 @@ dependencies = [
  "form_urlencoded",
  "futures",
  "http",
- "once_cell",
  "routefinder",
  "serde",
  "serde_json",
--- a/examples/http-rust-outbound-http/http-hello/Cargo.lock
+++ b/examples/http-rust-outbound-http/http-hello/Cargo.lock
@@ -224,12 +224,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f665ee40bc4a3c5590afb1e9677db74a508659dfd71e126420da8274909a0167"
 
 [[package]]
-name = "once_cell"
-version = "1.18.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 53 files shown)


tschneidereit (approved):
This is excellent, thank you!

I haven't looked at the implementation in detail, so this r+ is only about the approach and design, which look great to me.

radu-matei (approved):

fibonacci1729 (approved):

rylev:
I'm working on an alternative to this that I'm starting to like. I'll open a PR tomorrow so that we can compare.

itowlson:
When do we expect this to stabilise @rylev?  I appreciate we want to get it right, but I was trying yesterday to make a start on docs and it seems like the Rust API is still in flux?

rylev:
Trying my best to get this done. I agree it's a balance of getting it right and having to actually ship things. I hope that this will have settled by end of day tomorrow my time.

rylev:
Going to close this in favor of https://github.com/fermyon/spin/pull/1939

PR #1933, opened 2023-10-24
Require user to add redis host to list of allowed listed hosts

rylev:
~~This makes it required for users using the Redis v2 interface to add `allowed_redis_hosts` to their manifest specifying which hosts are allowed to be reached. This mirrors how key/value, llm, and sqlite all work.~~

~~Support is added in manifest v1 and v2 for the `allowed_redis_hosts`  key which defaults to the empty list. However, apps targeting the old Redis interface do not need to update their configs as only the new interface respects this configuration.~~ 

~~Next, we will move MySQL and PostgreSQL to do the same.~~

**Update**:

Introduce: `allowed_outbound_hosts` as a general way to control outbound networking. This is implemented for Redis for now, but will be introduced for MySQL, PostgreSQL, and http in future PRs.

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -114,6 +114,7 @@ impl LocalLoader {
         let metadata = ValuesMapBuilder::new()
             .string("description", component.description)
             .string_array("allowed_http_hosts", component.allowed_http_hosts)
+            .string_array("allowed_redis_hosts", component.allowed_redis_hosts)
             .string_array("key_value_stores", component.key_value_stores)
             .string_array("databases", component.sqlite_databases)
             .string_array("ai_models", component.ai_models)
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -114,7 +114,7 @@ impl LocalLoader {
         let metadata = ValuesMapBuilder::new()
             .string("description", component.description)
             .string_array("allowed_http_hosts", component.allowed_http_hosts)
-            .string_array("allowed_redis_hosts", component.allowed_redis_hosts)
+            .string_array_option("allowed_redis_hosts", component.allowed_redis_hosts)
             .string_array("key_value_stores", component.key_value_stores)
             .string_array("databases", component.sqlite_databases)
             .string_array("ai_models", component.ai_models)
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -111,6 +111,10 @@ impl LocalLoader {
         component: v2::Component,
     ) -> Result<LockedComponent> {
         outbound_http::allowed_http_hosts::parse_allowed_http_hosts(&component.allowed_http_hosts)?;
+        if let Some(hosts) = &component.allowed_outbound_hosts {
+            spin_outbound_networking::AllowedHosts::parse(hosts)
+                .context("`allowed_outbound_hosts` is malformed")?;
+        }
         let metadata = ValuesMapBuilder::new()
             .string("description", component.description)
             .string_array("allowed_http_hosts", component.allowed_http_hosts)
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -114,7 +114,7 @@ impl LocalLoader {
         let metadata = ValuesMapBuilder::new()
             .string("description", component.description)
             .string_array("allowed_http_hosts", component.allowed_http_hosts)
-            .string_array_option("allowed_redis_hosts", component.allowed_redis_hosts)
+            .string_array_option("allowed_outbound_hosts", component.allowed_outbound_hosts)
             .string_array("key_value_stores", component.key_value_stores)
             .string_array("databases", component.sqlite_databases)
             .string_array("ai_models", component.ai_models)
--- a/crates/manifest/src/schema/v2.rs
+++ b/crates/manifest/src/schema/v2.rs
@@ -102,6 +102,9 @@ pub struct Component {
     /// `allowed_http_hosts = ["example.com"]`
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub allowed_http_hosts: Vec<String>,
+    /// `allowed_redis_hosts = ["myredishost.com"]`
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub allowed_redis_hosts: Vec<String>,
     /// `key_value_stores = ["default"]`
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub key_value_stores: Vec<SnakeId>,
--- a/crates/manifest/src/schema/v2.rs
+++ b/crates/manifest/src/schema/v2.rs
@@ -103,8 +103,8 @@ pub struct Component {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub allowed_http_hosts: Vec<String>,
     /// `allowed_redis_hosts = ["myredishost.com"]`
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub allowed_redis_hosts: Vec<String>,
+    #[serde(default, skip_serializing_if = "is_none_or_empty")]
+    pub allowed_redis_hosts: Option<Vec<String>>,
     /// `key_value_stores = ["default"]`
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub key_value_stores: Vec<SnakeId>,
@@ -117,6 +117,11 @@ pub struct Component {
     /// Build configuration
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub build: Option<ComponentBuildConfig>,
+}
+
+/// Used to skip serializing if the value is either `None` or `Some(v)` where `v` is empty
+fn is_none_or_empty<T>(value: &Option<Vec<T>>) -> bool {
+    value.as_ref().map(|s| !s.is_empty()).unwrap_or_default()
 }
 
 mod one_or_many {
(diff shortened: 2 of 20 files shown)


crates/outbound-redis/src/lib.rs
```diff
@@ -58,6 +66,23 @@ impl v2::HostConnection for OutboundRedis {
...
+impl v2::Host for OutboundRedis {}
+
+#[async_trait]
+impl v2::HostConnection for OutboundRedis {
+    async fn open(&mut self, address: String) -> Result<Result<Resource<RedisConnection>, Error>> {
+        if !self.allowed_hosts.contains(&address) {
+            terminal::warn!(
+                "A component tried to make a HTTP request to non-allowed address '{address}'."
+            );
+            eprintln!("To allow requests, add 'allowed_redis_hosts = [\"{address}\"]' to the manifest component section.");
```
lann, line 80: :shrug: 
```suggestion
            eprintln!("To allow requests, add 'allowed_redis_hosts = [{address:?}]' to the manifest component section.");
```

lann (approved):

lann:
It seems like _ideally_ the Redis v1 interface would enforce `allowed_redis_hosts` _if present_ (including disabling on `allowed_redis_hosts = []`), but I don't know how much extra pain that would be to implement.

rylev:
I've updated this PR with the following:

* If users set `allowed_outbound_hosts` in a manifest v1, the v1 redis interface will respect that - otherwise all hosts are allowed. V2 is deny by default.
* `outbound-networking` has been introduced to allow for more general control than just host matching:
  * This mirrors the outbound http controls (and will likely replace it in the futute).
  * The `allowed_outbound_hosts` can contain url-like strings that contain a host and port (both required) and an optional scheme. The check determines whether the outgoing request has the same host and port and if the manifest included a scheme it also checks that that matches. 
* If the user passes a url to the implementation without scheme, the implementation will try to parse the url with the scheme that makes sense given the context. For example, if a user passes "localhost:6379" to the redis implementation, the implementation will assume that the scheme is "redis://".

lann (approved):

crates/loader/src/local.rs
```diff
@@ -111,9 +111,14 @@ impl LocalLoader {
         component: v2::Component,
     ) -> Result<LockedComponent> {
         outbound_http::allowed_http_hosts::parse_allowed_http_hosts(&component.allowed_http_hosts)?;
+        if let Some(hosts) = &component.allowed_outbound_hosts {
+            spin_outbound_networking::AllowedHosts::parse(hosts)
```
itowlson, line 115: If we allow `parse` to accept an Option then we can avoid having to check at the call site

crates/outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+        let url_str = url.as_ref();
+        let url: anyhow::Result<Url> = url_str
+            .try_into()
+            .with_context(|| format!("could not convert {url_str:?} into a url"));
+        let (url, has_scheme) = match url {
+            Ok(url) if url.has_host() => (url, true),
+            first_try => {
+                // If the url doesn't successfully parse try again with an added scheme.
+                // This resolves the ambiguity between 'spin.fermyon.com:80' and 'unix:80'.
+                // Technically according to the spec a valid url *must* contain a scheme. However,
```
itowlson, line 81: A URL must, but a host doesn't.  It won't be long before we get sockets, so this could really be operating on a URL _or_ host/IP address.  (Which I know you are saying will work... just flagging that we can't appeal to the URL spec anyway. _grin_)

crates/outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+                // we allow url-like strings without schemes, and we interpret the first part as the host.
+                let second_try = format!("scheme://{url_str}")
+                    .as_str()
+                    .try_into()
+                    .context("could not convert into a url");
```
itowlson, line 86: It seems like the remaining errors here don't say _which_ host entry committed the crime, and I don't see it being added further up.  Can you confirm that a parse failure does tell the user which host(s?) failed?

crates/outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+                match (second_try, first_try) {
+                    (Ok(u), _) => (u, false),
+                    // Return an error preferring the error from the first attempt if present
+                    (_, Err(e)) | (Err(e), _) => return Err(e),
+                }
+            }
+        };
+        let host = url.host_str().context("the url has no host")?.to_owned();
```
itowlson, line 94: Errors referring to "the URL" may be surprising to the user who thought they were entering a host.

crates/outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,255 @@
...
+    host: String,
+    port: u16,
+}
+
+impl AllowedHost {
+    fn parse<U: AsRef<str>>(url: U) -> anyhow::Result<Self> {
+        let url_str = url.as_ref();
+        let url: anyhow::Result<Url> = url_str
+            .try_into()
+            .with_context(|| format!("could not convert {url_str:?} into a url"));
```
itowlson, line 75: Leaving this comment once for all the places: URL should be capitalised in user-facing messages and I will suffer a mild toe injury on this hill.

crates/manifest/src/schema/v2.rs
```diff
@@ -102,6 +102,9 @@ pub struct Component {
     /// `allowed_http_hosts = ["example.com"]`
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub allowed_http_hosts: Vec<String>,
+    /// `allowed_outbound_hosts = ["myredishost.com"]`
+    #[serde(default, skip_serializing_if = "is_none_or_empty")]
+    pub allowed_outbound_hosts: Option<Vec<String>>,
```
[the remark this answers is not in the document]
rylev: I realize I never came out and said this, but the plan in my head is that `allowed_http_hosts` is deprecated even http goes through `allowed_outbound_hosts`. That will be in a follow-up PR though.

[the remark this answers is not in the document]
rylev: It's my understanding that v1 semantics have to be representable in the v2 manifest struct. This is because, we only deserialize into v1. V1 is then immeadiately converted into v2 and the rest of the code only deals with v2 manifests. 

For the host implementation of redis, we need to know whether the `allowed_outbound_hosts` was set or not , because there is a semantic difference between not set at all and set and empty (i.e., if it's set and empty, we deny everything, but if it's not set at all, we allow all). 

cc @lann for any corrections to my understanding.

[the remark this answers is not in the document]
itowlson: Yes, I assumed that HTTP would respect `allowed_outbound_hosts` but also allow entries from `allowed_http_hosts` for ease of transition - I hadn't actually clocked that wasn't in this PR!  When you say "in a follow-up PR" is that still in the Spin 2 timeframe?  I need to understand the expected 2.0 behaviour down for docs (acknowledging that there may be changes ongoing but hopefully small ones).

crates/outbound-networking/src/lib.rs
rylev: I believe though that most clients still operate on URLs even if the host of that URL is an IP address so I think the point still stands. Maybe we need to implement our own parser though...

itowlson:
Further to that timing question, are MySQL and Postgres also in scope for 2.0?

crates/manifest/src/schema/v2.rs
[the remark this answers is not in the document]
lann: True, though we could translate v1 `None` to `["redis://*:*"]`?

examples/rust-outbound-redis/spin.toml
```diff
@@ -8,6 +8,7 @@ version = "0.1.0"
 environment = { REDIS_ADDRESS = "redis://127.0.0.1:6379", REDIS_CHANNEL = "messages" }
 id = "outbound-redis"
 source = "target/wasm32-wasi/release/rust_outbound_redis.wasm"
+allowed_outbound_hosts = ["127.0.0.1:6379"]
```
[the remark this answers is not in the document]
lann: My intent has been to allow `[variables]` to be interpolated into these fields but it is current architecturally very awkward to do so, requiring a pretty major refactor of ~~`spin-config`~~ `spin-variables`.

rylev:
@itowlson yes - once this is merged I will do the other two. I wanted to make sure that the approach is agreed upon before doing the others.

crates/manifest/src/schema/v2.rs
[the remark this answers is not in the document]
rylev: We currently don't support '*' globs so this isn't currently possible.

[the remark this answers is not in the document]
rylev: My intent was to remove `allowed_http_hosts` from v2 manifests entirely. How this works with v1 manifests I'm not sure. I'm going to try my best to speed through this and figure out the right shape. Sorry for the last minute changes!

crates/loader/src/local.rs
rylev: `parse` taking `Option` seems a bit too tied to the fact that the manifest file currently is an optional field. We're discussing in the other thread about this not being the case potentially so I'd rather keep the slightly more verbose version here.

crates/outbound-networking/src/lib.rs
rylev: Circling back here - Currently our implementation requires more than just a host. Let me try my hand at a grammar:

```
<config> ::= allowed_outbound_hosts = <list>
<list> ::= ["<item>",*]
<item> ::= <host>:<port> | <scheme>://<host><:port>?
```

So the name  `allowed_outbound_hosts ` is a slight misnomer as the strings are more than hosts.

PR #1939, opened 2023-10-25
Improve the builtin Request/Response types

rylev:
This is meant as an alternative to #1931. It improves the `Request`/`Response` types with an API that is useable, flexible, and powerful and hopefully enough to cover 80% of use cases. For more advanced use cases, we would recommend using a third party crate like `http`. 

Note: The one slightly awkward API choice is that `RequestBuilder::uri` is infallible so that even if the user passes a garbage URI it still "works" - they'll likely only see any error when the try to use the uri. I think this is the right balance of easy of use without sacrificing too much protection against misuse.

### DIFF
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -289,7 +289,9 @@ impl HeaderValue {
     /// Construct a `HeaderValue` from a bag of bytes
     pub fn bytes(bytes: Vec<u8>) -> HeaderValue {
         HeaderValue {
-            inner: HeaderValueRep::Bytes(bytes),
+            inner: String::from_utf8(bytes)
+                .map(HeaderValueRep::String)
+                .unwrap_or_else(|e| HeaderValueRep::Bytes(e.into_bytes())),
         }
     }
 
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -28,17 +28,75 @@ pub struct Request {
     body: Vec<u8>,
 }
 
-enum HeaderValue {
+/// A header value.
+///
+/// Since header values do not have to be valid utf8, this allows for
+/// both utf8 strings and bags of bytes.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct HeaderValue {
+    inner: HeaderValueRep,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+enum HeaderValueRep {
+    /// Header value encoded as a utf8 string
     String(String),
+    /// Header value as a bag of bytes
     Bytes(Vec<u8>),
 }
 
 impl HeaderValue {
+    /// Construct a `HeaderValue` from a string
+    pub fn string(str: String) -> HeaderValue {
+        HeaderValue {
+            inner: HeaderValueRep::String(str),
+        }
+    }
+
+    /// Construct a `HeaderValue` from a bag of bytes
+    pub fn bytes(bytes: Vec<u8>) -> HeaderValue {
+        HeaderValue {
+            inner: HeaderValueRep::Bytes(bytes),
+        }
+    }
+
+    /// Get the `HeaderValue` as a utf8 encoded string
+    ///
+    /// Returns `None` if the value is a non utf8 encoded header value
+    pub fn as_str(&self) -> Option<&str> {
+        match &self.inner {
+            HeaderValueRep::String(s) => Some(s),
+            HeaderValueRep::Bytes(_) => None,
+        }
+    }
+
+    /// Get the `HeaderValue` as bytes
+    pub fn as_bytes(&self) -> &[u8] {
+        self.as_ref()
+    }
+
+    /// Turn the `HeaderValue` into a String (in a lossy way if the `HeaderValue` is a bag of bytes)
+    pub fn into_utf8_lossy(self) -> String {
+        match self.inner {
+            HeaderValueRep::String(s) => s,
+            HeaderValueRep::Bytes(b) => String::from_utf8_lossy(&b).into_owned(),
+        }
+    }
+
     /// Turn the `HeaderValue` into bytes
-    fn into_bytes(self) -> Vec<u8> {
-        match self {
-            HeaderValue::String(s) => s.into_bytes(),
-            HeaderValue::Bytes(b) => b,
+    pub fn into_bytes(self) -> Vec<u8> {
+        match self.inner {
+            HeaderValueRep::String(s) => s.into_bytes(),
+            HeaderValueRep::Bytes(b) => b,
+        }
+    }
+}
+
+impl AsRef<[u8]> for HeaderValue {
+    fn as_ref(&self) -> &[u8] {
+        match &self.inner {
+            HeaderValueRep::String(s) => s.as_bytes(),
+            HeaderValueRep::Bytes(b) => b,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


sdk/rust/src/http.rs
```diff
@@ -30,38 +149,94 @@ pub struct Request {
...
-            body: body.into_body(),
+    /// The response status
+    pub fn status(&self) -> &StatusCode {
+        &self.status
+    }
+
+    /// The response headers
+    ///
+    /// Technically headers do not have to be utf8 encoded but this type assumes they are.
+    /// If you know you'll be dealing with non-utf8 encoded headers, reach more a more powerful
```
tschneidereit, line 177:
```suggestion
    /// If you know you'll be dealing with non-utf8 encoded headers, reach for a more powerful
```

sdk/rust/src/http.rs
```diff
@@ -30,38 +149,94 @@ pub struct Request {
...
+    /// response type like that found in the `http` crate.
+    pub fn headers(&self) -> &[(String, String)] {
```
tschneidereit, line 179: could we maybe instead introduce a `headers_raw` (and `headers_raw_mut`) that returns `(String, Vec<u8>)`?

I'm imagining that we'd internally store headers as something like `(String, HeaderVal)`, where `HeaderVal` is something like
```rust
enum HeaderVal {
  Bytes(Vec<u8>),
  String(String),
  Both(Vec<u8>, String),
}
```

I bet there's some more optimal representation of this for the case where `Bytes` contains valid unicode, too.

sdk/rust/src/http.rs
```diff
@@ -15,13 +15,132 @@ pub use super::wit::wasi::http::types::*;
...
 pub struct Request {
     /// The method of the request
-    pub method: Method,
-    /// The path together with the query string
-    pub path_and_query: String,
+    method: Method,
+    /// The uri for the request
+    ///
+    /// The first item is set to `None` if the supplied uri is malformed
+    uri: (Option<hyperium::Uri>, String),
```
tschneidereit, line 22: if we haven't use hyper's `Uri` type before, then I think we should strongly consider using [rust-url](https://crates.io/crates/url) instead. It's about as battle-tested as can be. Importantly it's what Spin uses on the host-side as well, meaning there's no risk of security issues caused by varying interpretations of URLs.

tschneidereit (changes requested):
This looks great to me!

I left a couple of comments inline, but I agree that this is the right approach to take.

sdk/rust/src/http.rs
rylev: I am somewhat worried about this. I think I've managed to [push a change](https://github.com/fermyon/spin/pull/1939/commits/011576ce83b13108d750a19e5c9a656011cd2a02) that supports this without being a maintenance burden, be we do have to get comfortable at some point saying that these types will never cover 100% of use cases and a 3rd party crate might be necessary.

sdk/rust/src/http.rs
tschneidereit: Ok, I guess we can live with this for now. The security issues I'm talking about are real, unfortunately, but are much more likely to affect very naive, untested implementations, neither of which Hyper's is.

sdk/rust/src/http/conversions.rs
```diff
@@ -293,6 +294,58 @@ impl IntoStatusCode for hyperium::StatusCode {
     }
 }
 
+/// A trait for any type that can be turned into `Response` headers
+pub trait IntoHeaders {
+    /// Turn `self` into a `Response` body
```
lann, line 299: Comments need updating

sdk/rust/src/http.rs
```diff
@@ -15,13 +17,165 @@ pub use super::wit::wasi::http::types::*;
...
     /// The request headers
-    pub headers: Vec<(String, Vec<u8>)>,
+    headers: HashMap<String, HeaderValue>,
```
lann, line 26: I know this was my suggestion but since we aren't exposing the `Map`ness of this directly we could retain a bit of HTTP compliance by storing this as a `Vec` and e.g. `self.headers.iter().rev().find()`ing in `self.header`.

lann (approved):
A couple of suggestions, all of which could be handled in follow-up.

sdk/rust/src/http.rs
```diff
@@ -15,13 +17,165 @@ pub use super::wit::wasi::http::types::*;
...
+    }
+
+    /// Set the uri
+    pub fn uri(&mut self, uri: impl Into<String>) -> &mut Self {
+        self.request.uri = Request::parse_uri(uri.into());
+        self
+    }
+
+    /// Set the headers
+    pub fn headers(&mut self, headers: impl conversions::IntoHeaders) -> &mut Self {
```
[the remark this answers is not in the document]
rylev: I think we probably want both. In most of my usage I've already had a headers collection and just wanted that collection to be the headers, but we probably also want the ability to add headers one by one, perhaps with a `header` function?

sdk/rust/src/http.rs
rylev: Ah good catch - do you mean because a header can have more than one value? How about `HashMap<String, Vec<HeaderValue>>`? It seems worth it to save having to iterate through all headers to find a specific one.

sdk/rust/src/http.rs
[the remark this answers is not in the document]
lann: Yeah its quite nice to be able to do e.g. `.header("Authorization", format!("Bearer {token}"))`

sdk/rust/src/http.rs
```diff
@@ -15,13 +17,165 @@ pub use super::wit::wasi::http::types::*;
...
+        self.headers.iter().filter_map(|(k, v)| match v {
+            HeaderValue::String(v) => Some((k.as_str(), v.as_str())),
+            HeaderValue::Bytes(_) => None,
+        })
+    }
+
+    /// Return a header value
+    ///
+    /// Will return `None` if the header does not exist or if it is not utf8
+    pub fn header(&self, name: &str) -> Option<&str> {
```
lann, line 93: This should be case-insensitive, either by downcasing everywhere or by explicitly searching.

sdk/rust/src/http.rs
lann: Yeah, multiple values are _almost_ uncommon, but basically mandatory for `set-cookie`.

lann: uuuuugghhhh we're probably better off using `http::HeaderMap` internally :neutral_face:

rylev: `HeaderMap` is annoying in that it checks the bytes in `HeaderValue` for validity which may be more spec compliant but means we have to deal with that instead of letting it be someone else's problem.

lann: Yeah OK. Well, we do still need to handle case-sensitivity. Downcasing is probably the simplest way to manage that.

lann (approved):
Looks good; thanks!

michelleN (approved):

lann (approved):

sdk/rust/src/http.rs
```diff
@@ -28,17 +28,75 @@ pub struct Request {
...
+enum HeaderValueRep {
+    /// Header value encoded as a utf8 string
     String(String),
+    /// Header value as a bag of bytes
     Bytes(Vec<u8>),
 }
 
 impl HeaderValue {
+    /// Construct a `HeaderValue` from a string
+    pub fn string(str: String) -> HeaderValue {
```
tschneidereit: could `HeaderValue` implement `FromStr` instead?

sdk/rust/src/http.rs
```diff
@@ -15,13 +17,215 @@ pub use super::wit::wasi::http::types::*;
...
+    fn as_ref(&self) -> &[u8] {
+        match &self.inner {
+            HeaderValueRep::String(s) => s.as_bytes(),
+            HeaderValueRep::Bytes(b) => b,
+        }
+    }
+}
+
+impl Request {
+    fn new(method: Method, uri: impl Into<String>) -> Self {
```
rylev, line 105: AGH!

sdk/rust/src/http.rs
```diff
@@ -28,17 +28,75 @@ pub struct Request {
...
+        }
+    }
+
+    /// Get the `HeaderValue` as a utf8 encoded string
+    ///
+    /// Returns `None` if the value is a non utf8 encoded header value
+    pub fn as_str(&self) -> Option<&str> {
+        match &self.inner {
+            HeaderValueRep::String(s) => Some(s),
+            HeaderValueRep::Bytes(_) => None,
```
tschneidereit: I think it'd be good to try to return a string here, even if the underlying storage is a vec. Probably using
```rust
HeaderValueRep::Bytes(b) => b.from_utf8().ok(),
```

tschneidereit: Alternatively, that could of course be done when creating the `HeaderValue`, same as what happens inside `into_header_rep` right now

tschneidereit (commented):
A slight nit and one comment of some substance—but almost there!

sdk/rust/src/http.rs
rylev: `FromStr` is for fallible operations (i.e., it's what powers the `parse` function). I don't think it's appropriate here. We could implement `From<String>` and `From<Vec<u8>>`.

sdk/rust/src/http.rs
```diff
@@ -172,7 +212,7 @@ impl RequestBuilder {
     pub fn header(&mut self, key: impl Into<String>, value: impl Into<String>) -> &mut Self {
```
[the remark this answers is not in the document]
rylev: The builder allows setting raw headers, so worse case scenario users can use that.

sdk/rust/src/http.rs
```diff
@@ -99,6 +384,23 @@ impl std::fmt::Display for Method {
 }
 
 impl IncomingRequest {
+    /// The incoming request Uri
+    pub fn uri(&self) -> String {
```
itowlson, line 388: Can the user get this as a Url object or would they need to parse the string?

rylev: They would need to parse. The thought by behind it is to not make the `http` crate part of the public API for these types.

PR #1940, opened 2023-10-25
Improve the OutParam API for wasi-http

rylev:
This API is a bit awkward but this makes it slightly better by allowing method call syntax. I'm not sure I fully understand why the wit-bindgen generated version does allow for method call syntax.


dicej (approved):
WIT has three types of functions that can be associated with resources: constructors, methods, and static methods.  Methods always take their implicit `self`/`this` parameter by `borrow` handle.  Static methods don't have an implicit `self`/`this` parameter.  There's simply no way (currently) to associate a non-static method with a resource that takes its `self`/`this` parameter by `own` handle, so the closest thing you can do is define a static method that has an explicit `self`/`this` parameter as an `own` handle.

We _could_ add some kind of heurstic to `wit-bindgen` to detect that pattern and turn it into a method that takes `self` by owned value, but for now we treat all static methods as associated functions that do not have a real `self` parameter.

itowlson (approved):

PR #1941, opened 2023-10-25
feat(loader/cache): add data dir fallback if wasm layer not found

vdice:
Pre-emptive 'futureproofing' in anticipation of the wasm layer media type changing in the future (see https://github.com/fermyon/spin/pull/1882#issuecomment-1773232960).

This change is intended for the forthcoming 2.0 release.  With this logic in place, 2.0 clients should continue to be able to pull Spin apps published by a future version of Spin with the wasm layer media type changed.


lann (approved):
We could probably just merge the wasm and data dirs together anyway...

vdice:
> We could probably just merge the wasm and data dirs together anyway...

Good point... we should be able to do that in a refactor post-2.0 if we don't get to it beforehand.

PR #1942, opened 2023-10-25
ci/docs(releasing): updates for release candidates

vdice:
CI and docs updates around release candidates.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,8 @@ jobs:
       id-token: write
       # needed to upload artifacts to a GH release
       contents: write
+    outputs:
+      prerelease: ${{ steps.release-version.outputs.prerelease }}
     strategy:
       matrix:
         config:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -264,7 +276,9 @@ jobs:
     name: create tag sdk/go/v*
     runs-on: ubuntu-latest
     needs: build-and-sign
-    if: startsWith(github.ref, 'refs/tags/v')
+    if: |
+      startsWith(github.ref, 'refs/tags/v') &&
+      needs.build-and-sign.outputs.prerelease == 'false'
```
vdice, line 281: @itowlson I've opted to skip: pusthing the `sdk/go/v*` tag, creating the template SDK updates PR and pushing the templates tag if the v* release is a pre-release/release candidate.  WDYT?

itowlson: That will make them use main, which is _potentially_ problematic if the main templates/SDKs diverge from the stabilising templates/SDKs.  But hopefully any such changes would be planned for backport anyway.

vdice: Good point.  We _could_ do all of the above as usual for release candidates but I am unsure about the last one: pushing the templates tag.  We'd push eg `spin/templates/v2.0` on our first v2.0.0-rc.1 release as it stands now.  What might be the best strategy for this bit?

vdice: Updated to push sdk + templates tags as usual in https://github.com/fermyon/spin/pull/1942/commits/36e78370b759357b545f97fce792a60fc9bad4aa

michelleN (approved):

fibonacci1729 (approved):
Looks great, thanks @vdice !

PR #1944, opened 2023-10-25
Break `http` crate dependency on `wasmtime`

itowlson:
I think this might be the last one folks... again I resorted to a feature but I can look at splitting out crates if people think that's better.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5758,6 +5758,7 @@ dependencies = [
  "percent-encoding",
  "serde",
  "spin-app",
+ "spin-locked-app",
  "spin-testing",
  "toml 0.8.2",
  "tracing",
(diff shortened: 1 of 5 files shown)


dicej (approved):

lann:
This is fine but I want to obnoxiously note that if there isn't something checking for this dependency it will likely creep back in.

itowlson:
@lann Yeah that's my concern.  I am not sure how to mitigate it - I could put a silly check into cloud-plugin but that would be a severely lagging indicator, and Spin will never see it because it compiles app/core in anyway.  I guess I could write a thingy that references the dependency-broken crates and rings a bell if it ever finds itself importing wasmtime, and include that in Spin CI.  It seems awfully specific though.

PR #1946, opened 2023-10-25
trigger-http: Fix `base` routing without leading slash

### LINKED ISSUE #1943: Validate HTTP `base` to be absoluate
If I accidentally enter a base without the leading slash:

```
# v1
trigger = { type = "http", base = "boop" }

# v2
[application.trigger.http]
base = "boop"
```

then `spin up` appears to succeed, but it prints a route of `http://127.0.0.1:3000boop`, and `localhost:3000/boop` 404s.

(`spin deploy` also prints a badly concatenated URL, but Fermyon Cloud handles it like a champ regardless.)

lann:
This seems to take care of the printed routes as well.

Fixes #1943

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -93,10 +93,13 @@ impl TriggerExecutor for HttpTrigger {
     type RunConfig = CliArgs;
 
     async fn new(engine: TriggerAppEngine<Self>) -> Result<Self> {
-        let base = engine
+        let mut base = engine
             .app()
             .require_metadata(spin_http::trigger::METADATA_KEY)?
             .base;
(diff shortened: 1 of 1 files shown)


itowlson (approved):
Yes, this should fix local printing because the "available routes" rubric is printed by the HTTP trigger.  I'll make a separate fix for `deploy`.

I am not sure if any other tools need to be updated to handle the possibility.  I do wonder if doing it at load time might mitigate the need to have this code in all consumers, but if it gets problematic we can always add that later.  Thanks!

itowlson:
I think this only needs a rebase to pass (or you could admin your way past it, as it's a known issue and resolved in main)

PR #1948, opened 2023-10-25
Fix broken template

fibonacci1729:
The `http-rust` template appears to have been broken by #1939. This PR updates the templates to use the new headers API.

### DIFF
--- a/templates/http-rust/content/src/lib.rs
+++ b/templates/http-rust/content/src/lib.rs
@@ -1,10 +1,10 @@
-use spin_sdk::http::{IntoResponse, Request};
+use spin_sdk::http::{HeaderValue, IntoResponse, Request};
 use spin_sdk::http_component;
 
 /// A simple Spin HTTP component.
 #[http_component]
 fn handle_{{project-name | snake_case}}(req: Request) -> anyhow::Result<impl IntoResponse> {
-    println!("{:?}", req.headers);
+    println!("Handling request to {:?}", req.header("spin-full-url"));
     Ok(http::Response::builder()
         .status(200)
         .header("content-type", "text/plain")


templates/http-rust/content/src/lib.rs
```diff
@@ -1,12 +1,12 @@
-use spin_sdk::http::{IntoResponse, Request};
+use spin_sdk::http::{IntoResponse, Request, HeaderValue};
 use spin_sdk::http_component;
 
 /// A simple Spin HTTP component.
 #[http_component]
 fn handle_{{project-name | snake_case}}(req: Request) -> anyhow::Result<impl IntoResponse> {
-    println!("{:?}", req.headers);
+    println!("{:?}", req.headers().collect::<Vec<(&str, &HeaderValue)>>());
```
itowlson, line 7: Oh man... I don't want this to be the first thing a new user sees when they pop open a newly created Spin app...

itowlson: I guess we could turn it into a loop, which would also make for more readable output. But then the output is pretty verbose.  Maybe we should figure out what we're trying to achieve with this line.

itowlson (changes requested):

templates/http-rust/content/src/lib.rs
itowlson: What if we printed one of the headers from https://developer.fermyon.com/spin/http-trigger#additional-request-information?  That would:

* Limit the amount of output
* Show how to access a specific header
* Teach people about the existence of the SPECIAL headers

E.g. `println!("Handling request to {}", req.header("spin-full-url"))` or something like that?

fibonacci1729: Updated @itowlson

itowlson (approved):
Nice one, thanks for picking this up Brian!

michelleN (approved):

templates/http-rust/content/src/lib.rs
```diff
@@ -1,10 +1,10 @@
-use spin_sdk::http::{IntoResponse, Request};
+use spin_sdk::http::{HeaderValue, IntoResponse, Request};
```
rylev, line 1: Is this new import necessary?

itowlson: @rylev no, we exterminated it in #1956

PR #1951, opened 2023-10-25
Fix v1 config error propagation

fibonacci1729:
When updating to `variables` I implemented the `spin_world::v1::config::Host` trait in terms of the `variables::Host` implementation. However (and i'm not sure how/why) the `variables::Error` (defined in `variables.wit`) was being propagated back to the ABI and causing a panic deep in the bowels of something. In any case, i decided to simply remap`variables::Error` to a variant in `spin_world::v1::config::Error` which appears to correct the issue.

### DIFF
--- a/crates/variables/src/host_component.rs
+++ b/crates/variables/src/host_component.rs
@@ -96,7 +96,14 @@ impl spin_world::v1::config::Host for ComponentVariables {
         &mut self,
         key: String,
     ) -> Result<Result<String, spin_world::v1::config::Error>> {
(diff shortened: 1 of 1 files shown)


itowlson (approved):
Thanks!  Worth testing with the redirect component if you've not already done so?

fibonacci1729:
@itowlson have done so and no error so seemingly fallback kicks in as expected.

itowlson:
@fibonacci1729 merged this because Karthik's bartholomew build hit it too, hope that's okay

PR #1953, opened 2023-10-26
Support v1 syntax for spin new and spin add

itowlson:
By popular demand, we changed the `spin new/add` syntax in Spin 2 to `spin new [NAME] [-t TEMPLATE]`.

The trouble is we have a _lot_ of content out there, such as videos and blogs, which walks people through using the v1 syntax, and would be extremely expensive to remake.

This PR introduces an ugly hack by which if a user uses the old syntax, `spin new TEMPLATE NAME`, then we spot that there are two positional args and interpret them accordingly.  This does leave us in the ugly position that in `spin new foo` foo is the app name and in `spin new foo bar` foo is the template name but I don't know how to resolve that.  It is going to be a vexing experience for people who forget the `-t` because the error will be not guide them to "missing flag".

The behaviour now is:

```
# V1 syntax, creates a Rust app named tnpos
ivan@hecate:~/testing/tpltest$ spin new http-rust tnpos
Description:
HTTP path: /...

# Treated as V2 syntax, prompts for template, warns in case they meant V1 syntax
ivan@hecate:~/testing/tpltest$ spin new http-rust
This will create an app called http-rust. If you meant to use the http-rust template, write '-t http-rust'.

# Treated as V2 syntax, prompts for template
ivan@hecate:~/testing/tpltest$ spin new myapp

# V2 syntax, prompts for name
ivan@hecate:~/testing/tpltest$ spin new -t http-rust
Enter a name for your new application: fie
Description:
HTTP path: /...

# V2 syntax, creates Rust app called ntarg
ivan@hecate:~/testing/tpltest$ spin new ntarg -t http-rust
Description:
HTTP path: /...

# Treated as V1 syntax because two positional parameters, but fails
ivan@hecate:~/testing/tpltest$ spin new myapp http-rust
Error: No templates matched 'myapp'

# Bad syntax, berates user
ivan@hecate:~/testing/tpltest$ spin new myapp http-rust -t http-rust
Error: Cannot supply both positional and named template id
```


mikkelhegn (approved):
LGTM

rylev:
Can we add some warnings so that people are educated about the new syntax? Preferably this will allow us to eventually remove this hack.

itowlson:
@rylev added

```
$ spin new http-rust biscuits
Using Spin 1 command syntax. The recommended syntax in Spin 2 is 'spin new biscuits -t http-rust'
```

PR #1955, opened 2023-10-26
add global value convertor for mysql and sqlite go sdk 

Patrick0308:
If statement didn't implement ColumnConverter interface, go's db driver will use driver.DefaultParameterConverter to covert value, such as coverting int to int64. 
I think it's not expected.

May be share a ValueConverter which should be on a internal pkg?

### DIFF
--- a/sdk/go/mysql/mysql.go
+++ b/sdk/go/mysql/mysql.go
@@ -7,10 +7,9 @@ import (
 	"errors"
 	"io"
 	"reflect"
-)
 
-// globalValueConv a valueConv instance
-var globalValueConv = &valueConv{}
+	spindb "github.com/fermyon/spin/sdk/go/internal/db"
+)
 
 // Open returns a new connection to the database.
 func Open(address string) *sql.DB {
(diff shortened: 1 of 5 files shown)


sdk/go/mysql/mysql.go
```diff
@@ -92,6 +96,18 @@ func (s *stmt) Exec(args []driver.Value) (driver.Result, error) {
 	return &result{}, err
 }
 
+// ColumnConverter return globalValueConv to don't use driver.DefaultParameterConverter
+func (s *stmt) ColumnConverter(_ int) driver.ValueConverter {
+	return globalValueConv
+}
+
+// valueConv a convertor not convert value
+type valueConv struct{}
```
adamreese, line 105:
```suggestion
type valueConverter struct{}
```

sdk/go/sqlite/sqlite.go
```diff
@@ -8,6 +8,9 @@ import (
 	"io"
 )
 
+// globalValueConv a valueConv instance
+var globalValueConv = &valueConv{}
```
adamreese, line 12:
```suggestion
var globalValueConverter = &valueConverter{}
```

sdk/go/mysql/mysql.go
```diff
@@ -9,6 +9,9 @@ import (
 	"reflect"
 )
 
+// globalValueConv a valueConv instance
+var globalValueConv = &valueConv{}
```
adamreese, line 13:
```suggestion
var globalValueConverter = &valueConverter{}
```

sdk/go/sqlite/sqlite.go
```diff
@@ -167,6 +171,18 @@ func (s *stmt) Exec(args []driver.Value) (driver.Result, error) {
 	return &result{}, err
 }
 
+// ColumnConverter return globalValueConv to don't use driver.DefaultParameterConverter
+func (s *stmt) ColumnConverter(_ int) driver.ValueConverter {
+	return globalValueConv
+}
+
+// valueConv a convertor not convert value
+type valueConv struct{}
```
adamreese, line 180:
```suggestion
type valueConverter struct{}
```

adamreese (commented):
Sorry for the huge amount of nits. Refactoring to an internal package would be great. Thanks again for all this work you have been doing. We really appreciate you!

adamreese (approved):

PR #1958, opened 2023-10-26
Potentially final wit changes for 2.0.0

rylev:
This is the last of the wit changes I believe we'll need for 2.0. These changes are largely cosmetic and meant just to tidy things up:

* The 2.0.0 package longer has a dependency on the previous (un-versioned) Spin wit package. 
  * This means moving the `llm`, `rdms-types`, and `redis-types` interfaces into the Spin 2.0 package instead 
  * These interfaces will go from looking unversioned to versioned (e.g., `spin:fermyon/llm` to `spin:fermyon/llm@2.0.0`).
* Move the mysql and postgres error types into the `rdms-types` package to share their definition
* Move everything in `redis-types` into `redis`.
* Remove the unused `redis-types#error` enum (the `redis` interface already has its own error and didn't use this other error type).
* Remove the `redis-trigger` world

# Open questions

## Should we remove the `redis-trigger` world?

I removed the `redis-trigger` world in anticipation of this moving to a plugin. In reality this doesn't really change anything. Users of the redis trigger will continue to be able to use Spin, but this sends a signal that we don't plan on supporting this natively and users can't use this wit package to build a redis trigger handler using wit based tooling (e.g., cargo component). 

This seems like the most conservative move since even if we decide that we actually do want to natively support Redis triggers, we can add this back with a minor version bump but removing this support would require a major version bump. 

## What should the package version be?

We've versioned this new Spin package as 2.0 to match the new version of the Spin runtime, but the wit world and Spin runtime are not necessarily linked. There will almost certainly be a future where a new minor version of the Spin runtime is released with no changes to the wit which will mean these versions come unlocked. I could also imagine a future where major changes are made to the wit world leading to a new major version release of the wit but the Spin runtime does not have a major version bump. 

With this in mind, do we want to already start versioning things differently and call this new package `fermyon:spin@1.0.0`?

For users of the SDK this will have 0 impact (versioning is already done separately through the SDK packages). However, this is visible to those interacting directly with Spin through component model tooling.


tschneidereit:
This all makes sense to me, thank you! ❤️ 

> Should we remove the `redis-trigger` world?

I think that might make sense, in light of your argument that it's easier to reintroduce than to remove later.

> What should the package version be?

I agree with your current decision to start with 2.0, given that while these are the same APIs as in the past, they're not strictly semver compatible because they actually have a version number.

In general one thing we could consider is always giving Spin APIs the version of the first Spin release supporting them. That way, the version numbers would readily tell developers the MSSV (Minimum Supported Spin Version) of the API.

wit/preview2/world.wit
```diff
@@ -1,25 +1,21 @@
...
-world redis-trigger {
-  include platform
-  export fermyon:spin/inbound-redis
-}
-
+// The full world of a guest targeting an http-trigger
 world http-trigger {
   include platform
-  export fermyon:spin/inbound-http
+  export wasi:http/outgoing-handler@0.2.0-rc-2023-10-18
```
michelleN, line 12: how come we are exporting outgoing-handler and not incoming-handler here?

wit/preview2/deps/spin@unversioned/config.wit
lann, line 1: Is `deps/spin@versioned` going to make some WIT tooling sad at some point?

rylev: Up until now wit tooling hasn't cared about directory names. But I think we'll be able to change this in the future if we need to. `0.0.0` isn't correct though so I'd rather not do the change unless we have to.

wit/preview2/world.wit
rylev: Good catch!

wit/preview2/deps/spin@unversioned/config.wit
lann: `0.0.0` is a somewhat conventional stand-in for "unversioned", but yeah I suppose we don't have to change unless it becomes a problem

wit/preview2/world.wit
```diff
@@ -1,25 +1,21 @@
...
-}
-
+// The full world of a guest targeting an http-trigger
 world http-trigger {
   include platform
-  export fermyon:spin/inbound-http
+  export wasi:http/incoming-handler@0.2.0-rc-2023-10-18
 }
 
+// The imports needed for a guest to run on a Spin host
```
fibonacci1729, line 15:
```suggestion
/// The imports needed for a guest to run on a Spin host
```

wit/preview2/world.wit
```diff
@@ -1,25 +1,21 @@
...
   include fermyon:spin/host
   include platform
 }
 
-world redis-trigger {
-  include platform
-  export fermyon:spin/inbound-redis
-}
-
+// The full world of a guest targeting an http-trigger
```
fibonacci1729, line 9:
```suggestion
/// The full world of a guest targeting an http-trigger
```

wit/preview2/world.wit
```diff
@@ -1,25 +1,21 @@
 package fermyon:spin@2.0.0
 
+// All worlds the Spin host supports
```
fibonacci1729, line 3:
```suggestion
/// All worlds the Spin host supports
```

lann (approved):

PR #1961, opened 2023-10-26
[wasi-http handler] Handle error and send response if not yet sent

rylev:
Allows http handlers to return `Result<_, impl IntoResponse>`. 

If the handler returns an `Err` and they've not yet set a response, we will set one for them. 

Some pros/cons of this:
* `ResponseOutparam::set` can now panic if it gets called twice (before we prevented this by taking `self` instead of `&self`). This necessary to dynamically decide if the handler will set the response or if the macro code will. 
* If a response has already been set but the handler then returns an error, all we can do is log. 
  * IMO this makes the 'streaming-outgoing-body' example worse as the user is lulled into thinking their errors will be handled but they won't be. They'll just be logged.
* If the user returns `Result<T, impl IntoResponse>` and they return a `Ok(T)` we currently just drop that value on the floor.
  * We could accept any type T that implements `IntoResponse` (returning it if the response hasn't yet been set), but then we would be forced to implement `IntoResponse` for `()` which is a big footgun that people always complain about in `axum`.

I'm personally unsure if I actually want to see this merged, but I thought I would open it up for comment from others.

### DIFF
--- a/examples/wasi-http-rust-streaming-outgoing-body/src/lib.rs
+++ b/examples/wasi-http-rust-streaming-outgoing-body/src/lib.rs
@@ -11,7 +11,10 @@ use url::Url;
 const MAX_CONCURRENCY: usize = 16;
 
 #[http_component]
-async fn handle_request(request: IncomingRequest, response_out: ResponseOutparam) {
+async fn handle_request(
+    request: IncomingRequest,
+    response_out: ResponseOutparam,
+) -> anyhow::Result<()> {
     let headers = request.headers().entries();
 
     match (request.method(), request.path_with_query().as_deref()) {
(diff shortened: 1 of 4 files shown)


lann:
Could we instead have something like:

```rust
pub struct EmptyResponse;
pub type EmptyResult<E = anyhow::Error> = Result<EmptyResult, E>;
```

lann:
```rust
Ok(IPromiseIDidntReturnErrAfterCallingResponseOutparamSet)
```

:neutral_face:

lann:
> `ResponseOutparam::set` can now panic if it gets called twice (before we prevented this by taking self instead of &self). This necessary to dynamically decide if the handler will set the response or if the macro code will.

Could you avoid this by using a `#[doc(hidden)]` e.g. `ResponseOutparamInternal` in the macro and wrapping it in a `struct ResponseOutparam(ResponseOutparamInternal)` (which still has `fn set(self, ...)`) and passing _that_ into the handler? (more types more better)

rylev:
@lann I think that's a good idea and basically what @dicej had suggested before! I was originally intending this to PR to allow users to call `set` dynamically, but that is clearly not a good idea.

lann:
@rylev given your last comment I moved this to draft to aid in 2.0 triage.

rylev:
Closing this since I'm not so sure this a good idea. It is potentially possible to add something like this post 2.0 in a backwards compatible way.

PR #1972, opened 2023-10-27
Implement HostOutputStream for MyStdoutStream

fibonacci1729:
TODO: The same should be done for `MyStdinStream`.

### DIFF
--- a/crates/core/src/store.rs
+++ b/crates/core/src/store.rs
@@ -1,8 +1,9 @@
 use anyhow::{anyhow, Result};
+use bytes::Bytes;
 use std::{
     io::{Read, Write},
     path::{Path, PathBuf},
-    sync::Mutex,
+    sync::{Arc, Mutex},
     time::{Duration, Instant},
 };
 use system_interface::io::ReadReady;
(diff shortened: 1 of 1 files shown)


crates/core/src/store.rs
```diff
@@ -264,7 +269,7 @@ impl StoreBuilder {
                 ctx.set_stderr(Box::new(wasi_preview1::pipe::WritePipe::new(w)))
             }
             WasiCtxBuilder::Preview2(ctx) => {
-                ctx.stderr(MyStdoutStream(Mutex::new(Some(Box::new(
+                ctx.stderr(MyStdoutStream(Arc::new(Mutex::new(Box::new(
```
dicej, line 272: I think we need to make this a `tokio::sync::Mutex` instead of a `std::sync::Mutex` since we'll need to lock it during an `await` in the `Subscribe` implementation.

crates/core/src/store.rs
```diff
@@ -442,15 +447,31 @@ impl StdinStream for MyStdinStream {
...
+    }
+
+    fn check_write(&mut self) -> Result<usize, StreamError> {
+        self.0.lock().unwrap().check_write()
+    }
+}
+
+#[async_trait]
+impl Subscribe for MyStdoutStream {
+    async fn ready(&mut self) {}
```
dicej, line 469: This should call `self.0.lock().await.ready().await` (and see the note about using `tokio::sync::Mutex` above).

dicej (changes requested):

dicej:
The CI failures indicate that this whole approach (which was my suggestion) might be flawed.  Each component expects to have its own handle to stdout, and since `HostOutputStream` has mutually-exclusive asynchronous (from the guest perspective) operation which we can't feasibly manage concurrently, I think we'll need a different solution.  @fibonacci1729 let's huddle about this when you have time.

fibonacci1729:
Superseded by https://github.com/fermyon/spin/pull/1982

Issue #1954, opened 2023-10-26
Examples: http-rust-outbound-http bugs/updates


vdice:
Encountered the following for the [http-rust-outbound-http](https://github.com/fermyon/spin/tree/main/examples/http-rust-outbound-http) example.

### 1. self request

Making a `self` request to another component doesn't appear to work:

```
$ curl -i http://127.0.0.1:3000/outbound-to-hello-component
HTTP/1.1 200 OK
date: Thu, 26 Oct 2023 01:20:32 GMT
transfer-encoding: chunked
spin-component: rust-outbound-http


```
Note the `rust-outbound-http`; this doesn't match up with the name that I'd currently expect it to be, eg `hello-component` (see [here](https://github.com/fermyon/spin/blob/main/examples/http-rust-outbound-http/spin.toml#L41)) -- however, when I update this value, it still doesn't appear to work (eg I expect to see ["Hello, Fermyon!" returned](https://github.com/fermyon/spin/blob/main/examples/http-rust-outbound-http/http-hello/src/lib.rs#L9) but don't)

### 2. outbound requests

Curling the `/outbound` and `/wildcard` endpoints results in the following:

```
$ curl -i http://127.0.0.1:3000/outbound
HTTP/1.1 200 OK
spin-component: rust-outbound-http
content-type: application/json
content-length: 75
date: Thu, 26 Oct 2023 01:23:50 GMT

curl: (18) transfer closed with 75 bytes remaining to read

$ curl -i http://127.0.0.1:3000/wildcard
HTTP/1.1 200 OK
spin-component: rust-outbound-http
content-type: application/json
content-length: 78
date: Thu, 26 Oct 2023 01:23:53 GMT

curl: (18) transfer closed with 78 bytes remaining to read
```

App logs show:

```
Response { status: 200, version: HTTP/1.1, headers: {"content-length": "75", "content-type": "application/json", "date": "Thu, 26 Oct 2023 01:23:50 GMT", "spin-component": "rust-outbound-http"}, body: () }
Response { status: 200, version: HTTP/1.1, headers: {"content-length": "78", "content-type": "application/json", "date": "Thu, 26 Oct 2023 01:23:53 GMT", "spin-component": "rust-outbound-http"}, body: () }
```
(By the way, for these routes, we shouldn't be setting the `spin-component` headers, right?  The component makes a request to an external API.  Perhaps they aren't hurting things, though.)

### 3. supported in cloud?

Are the following comments still valid?
[...]

PR #1975, opened 2023-10-27
Fix outbound http in the Rust SDK (take 2)

rylev:
This is a subset of the changes from #1902.

It keeps the API for `send` the same (with a slight change to the trait bounds). The internals are tweaked to fix outgoing requests which were not working when a body was sent.

There still seems to be some bugs in wasmtime-wasi-http that prevent true streaming of outbound request bodies. 

Also fixes https://github.com/fermyon/spin/issues/1954. @vdice, the issue was that the response had a non-zero content-length header but the body was actually empty. This was fixed by actually returning the body - [fixed here](https://github.com/fermyon/spin/compare/http-fix?expand=1#diff-060f2a7c3367a6ec24a7825ca6eac497707e2451843adf4b4c623b11ae783bf3L10).

### DIFF
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -535,6 +535,9 @@ impl ResponseOutparam {
 }
 
 /// Send an outgoing request
+///
+/// If `request`` is an `OutgoingRequest` and you are streaming the body to the
+/// outgoing request body sink, you need to ensure it is dropped before awaiting this function.
 pub async fn send<I, O>(request: I) -> Result<O, SendError>
 where
     I: TryIntoOutgoingRequest,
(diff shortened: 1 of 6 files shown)


sdk/rust/src/http.rs
```diff
@@ -498,18 +537,25 @@ impl ResponseOutparam {
...
-        request
-            .try_into()
-            .map_err(|e| SendError::RequestConversion(e.into()))?,
-    )
-    .await
-    .map_err(SendError::Http)?;
+    let (request, body_buffer) = I::try_into_outgoing_request(request)
+        .map_err(|e| SendError::RequestConversion(e.into()))?;
+    if let Some(body_buffer) = body_buffer {
+        // We assume that implementors of `TryIntoOutgoingRequest` do not call write so this
```
dicej, line 548: incomplete sentence?

dicej (approved):

vdice (commented):
> Also fixes https://github.com/fermyon/spin/issues/1954. @vdice, the issue was that the response had a non-zero content-length header but the body was actually empty. This was fixed by actually returning the body - [fixed here](https://github.com/fermyon/spin/compare/http-fix?expand=1#diff-060f2a7c3367a6ec24a7825ca6eac497707e2451843adf4b4c623b11ae783bf3L10).

Thanks @rylev.  That was the missing bit for the [outbound-http-to-same-app](https://github.com/fermyon/spin/tree/main/examples/http-rust-outbound-http/outbound-http-to-same-app) component as well.  Can we update its [lib.rs](https://github.com/fermyon/spin/blob/main/examples/http-rust-outbound-http/outbound-http-to-same-app/src/lib.rs#L10) to set the type as `http::Response<String>` in this PR as well?

PR #1959, opened 2023-10-26
Add address checks for mysql and postgres

rylev:
Continuation of #1933 but for MySQL and PostgreSQL. 

Users of the v2 `mysql` and `postgres` interfaces must now add the hosts they are allowed to call to the manifest.

PR #1978, opened 2023-10-27
chore(examples/tinygo-outbound-*): add allowed_outbound_hosts config

vdice:
Re: https://github.com/fermyon/spin/pull/1933 and https://github.com/fermyon/spin/pull/1959

### DIFF
--- a/examples/tinygo-outbound-mysql/spin.toml
+++ b/examples/tinygo-outbound-mysql/spin.toml
@@ -13,6 +13,7 @@ component = "tinygo-mysql"
 [component.tinygo-mysql]
 environment = { DB_URL = "mysql://spin:spin@127.0.0.1/spin_dev" }
 source = "main.wasm"
+allowed_outbound_hosts = ["mysql://127.0.0.1"]
 [component.tinygo-mysql.build]
 command = "tinygo build -target=wasi -gc=leaking -o main.wasm main.go"
 watch = ["**/*.go", "go.mod"]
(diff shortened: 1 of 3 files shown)


examples/tinygo-outbound-mysql/spin.toml
```diff
@@ -13,6 +13,7 @@ component = "tinygo-mysql"
 [component.tinygo-mysql]
 environment = { DB_URL = "mysql://spin:spin@127.0.0.1/spin_dev" }
 source = "main.wasm"
+allowed_outbound_hosts = ["mysql://127.0.0.1:3306"]
```
lann, line 16: fwiw `:3306` should be the default for `mysql://`

rylev: We currently require the port even if there is a scheme. Changing this is super easy, but I wanted to start with the most conservative approach and expand it if/when we think it's sensible.

itowlson: I thought `allowed_outbound_hosts` entries were meant to be just host + port rather than scheme?  Is scheme allowed but ignored?

itowlson: (`allowed_http_hosts` allowed the HTTP/S scheme for backward compatibility because there was already existing code that used it, but we tried to remove the scheme from the docs in favour of "just the host".)

lann: #1988 makes default ports optional

rylev: To expand on what @lann said: scheme is always required and the port can omitted when they scheme has a default well known port. Since MySQL has well known default port (i.e., 3306), the port can be omitted.

vdice: Thanks all; updated.

rylev (approved):

radu-matei (approved):

PR #1980, opened 2023-10-27
Fix `spin up --help` for unbuilt apps

### LINKED ISSUE #1957: `spin up --help` does not print trigger options unless all Wasm files exist
Create a new Spin app:

```
spin new emptyish -t http-rust
cd emptyish
```

DO NOT BUILD IT.

Now run `spin up --help`.  It prints the `up` options but not the trigger options section (not even the generic trigger options).

Now do `spin build` and run `spin up --help` again.  It now _does_ print the trigger options.

I think this is because the manifest load operation now fully materialises the app, which fails if a component `source` doesn't resolve (in this case, because the file hasn't been built).

It would be good to be able to see help regardless of build status - the Wasm files are not needed in order to to know what options are available.  (Although yes, this sort of multiple competing needs load pipeline is how the v1 loader got into the mess it did, so care is needed!)

It seems like this also happens if the manifest has no components in it (e.g. `http-empty` template), though I'm not sure I'm reproducing this consistently - first time I tried this seemed to give me the generic trigger options as I expected, now it gives me no trigger options at all.  So maybe it's not just missing Wasm files causing this - I've not dug into it yet.

Thanks to @tpmccallum for spotting the missing options!

lann:
This shuffles the `spin up` logic around a bit to allow `--help` calls to execute before fully loading local apps. A nice side benefit is that `--help` should be a bit faster, especially for apps with lots of `files`. Note that `OCI` behavior is (should be) unchanged as the current OCI codebase doesn't really allow for splitting up "get the manfiest" and "load everything".

Fixes #1957

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -1,13 +1,12 @@
 mod app_source;
 
 use std::{
-    collections::HashSet,
     ffi::OsString,
     fmt::Debug,
     path::{Path, PathBuf},
 };
 
-use anyhow::{anyhow, bail, ensure, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use clap::{CommandFactory, Parser};
 use reqwest::Url;
 use spin_app::locked::LockedApp;
@@ -18,7 +17,7 @@ use tempfile::TempDir;
 
 use crate::opts::*;
 
-use self::app_source::AppSource;
+use self::app_source::{AppSource, ResolvedAppSource};
 
 const APPLICATION_OPT: &str = "APPLICATION";
 
@@ -124,7 +123,7 @@ impl UpCommand {
     }
 
     async fn run_inner(self) -> Result<()> {
-        let app_source = self.resolve_app_source();
+        let app_source = self.app_source();
 
         if app_source == AppSource::None {
             if self.help {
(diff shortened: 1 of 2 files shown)


src/commands/up.rs
```diff
@@ -308,29 +292,54 @@ impl UpCommand {
         Ok(locked_url)
     }
 
-    async fn prepare_app_from_file(
+    // Take the AppSource and do the minimum amount of work necessary to
+    // be able to resolve the trigger executor.
+    async fn resolve_app_source(
```
itowlson, line 297: (a note for the future, not for now)  This function and `load_resolved_app_source` now have almost nothing to do with `self` - it might be interesting to explore making AppSource and ResolvedAppSource more self-contained at some future point.  Not a consideration for now though

itowlson (approved):
I left a nit about naming but it's all internal and we can always noodle on it later - this works great, thanks!

src/commands/up/app_source.rs
```diff
@@ -0,0 +1,101 @@
...
+    File {
+        manifest_path: PathBuf,
+        manifest: AppManifest,
+    },
+    OciRegistry {
+        locked_app: LockedApp,
+    },
+}
+
+impl ResolvedAppSource {
```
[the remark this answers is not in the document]
lann: This is the 3rd unsatisfactory name I came up with.

src/commands/up.rs
lann: Indeed `up.rs` seems ripe for major refactoring "pretty soon".

PR #1983, opened 2023-10-29
Fix watchexec missing files in subdirectories

### LINKED ISSUE #1981: Bug: spin watch broken for nested files
Using spin canary (built from main as of https://github.com/fermyon/spin/commit/47e40b9b38950c55355db575cd1b7ac681292621), I'm seeing behavior where `spin watch` doesn't trigger a rebuild upon changes to nested files.

Any app with a nested file entry in the `watch` config should repro, eg `watch = ["src/**/*.rs", "Cargo.toml"]` as seen in the [http-rust example](https://github.com/fermyon/spin/tree/main/examples/http-rust).  When running this example, a re-build is triggered when the `Cargo.toml` file is amended but not triggered when `src/lib.rs` is amended.

itowlson:
Fixes #1981.

### DIFF
--- a/src/commands/watch.rs
+++ b/src/commands/watch.rs
@@ -7,6 +7,7 @@ use std::{
 use anyhow::{Context, Result};
 use clap::Parser;
 use itertools::Itertools;
+use path_absolutize::Absolutize;
 use spin_common::paths::parent_dir;
 use uuid::Uuid;
 use watchexec::Watchexec;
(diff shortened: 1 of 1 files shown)


radu-matei (approved):

src/commands/watch.rs
```diff
@@ -93,6 +94,7 @@ impl WatchCommand {
 
         let spin_bin = std::env::current_exe()?;
         let manifest_file = spin_common::paths::resolve_manifest_file_path(&self.app_source)?;
+        let manifest_file = manifest_file.absolutize()?.to_path_buf(); // or watchexec misses files in subdirectories
```
lann, line 97: BUT WHY

vdice (approved):

src/commands/watch.rs
itowlson: BECAUSE REASONS LANN REASONS

We could probably absolutise at the last moment before we pass the directory to `watchexec` - I just get paranoid about capturing the true location as early as possible (a) so I don't need to remember to do it at _all_ the last moments and (b) in case some other piece of code sneakily switches out the current directory from underneath me.  _thousand yard stare_

We can iterate on this for sure though.

PR #1988, opened 2023-10-30
Get rid of `allowed_http_hosts` in v2 manifest

rylev:
This removes `allowed_http_hosts` in favor of `allowed_outbound_hosts`.

`allowed_outbound_hosts` has become strict in that it is required to follow the form "$SCHEME://$HOST:$PORT" where $SCHEME can be "\*" or a specific scheme, $HOST can be "\*" or a specific host, and $PORT can be "\*", a specific port, or a port range. In the future we will allow lists of schemes, hosts, and ports. 

Some notes:
* "insecure:allow-all" is done away with in favor of `*://*:*`.
* Schemes are required but if you don't care you can use globs: `*://example.com`


crates/loader/src/local.rs
```diff
@@ -110,15 +110,12 @@ impl LocalLoader {
         id: &KebabId,
         component: v2::Component,
     ) -> Result<LockedComponent> {
-        outbound_http::allowed_http_hosts::parse_allowed_http_hosts(&component.allowed_http_hosts)?;
-        if let Some(hosts) = &component.allowed_outbound_hosts {
-            spin_outbound_networking::AllowedHosts::parse(hosts)
-                .context("`allowed_outbound_hosts` is malformed")?;
-        }
+        spin_outbound_networking::AllowedHostsConfig::parse(&component.allowed_outbound_hosts)
+            .context("`allowed_outbound_hosts` is malformed")?;
```
lann, line 114: Definitely low priority but it might be nice to eventually move this parsing into serde for the nice errors you get out of toml deserialization.

crates/manifest/src/allowed_http_hosts.rs
lann, line 1: minor: `crates/manifest/src/compat/allowed_http_hosts.rs` perhaps?

crates/manifest/tests/ui/maximal.json
```diff
@@ -59,8 +59,8 @@
       "exclude_files": [
         "**/secret"
       ],
-      "allowed_http_hosts": [
-        "example.com"
+      "allowed_outbound_hosts": [
+        "https://example.com:443"
```
lann, line 63: Requiring _both_ a port and a scheme does seem pretty odd, especially for http(s).

examples/rust-outbound-pg/spin.toml
```diff
@@ -12,6 +12,6 @@ component = "outbound-pg"
 [component.outbound-pg]
 environment = { DB_URL = "host=localhost user=postgres dbname=spin_dev" }
 source = "target/wasm32-wasi/release/rust_outbound_pg.wasm"
-allowed_outbound_hosts = ["localhost:5432"]
+allowed_outbound_hosts = ["postgres://localhost:5432"]
```
itowlson, line 15: I'm not sure where this scheme comes from?  There's no scheme used in the connection string.

examples/wasi-http-rust-streaming-outgoing-body/spin.toml
```diff
@@ -12,7 +12,7 @@ component = "wasi-http-async"
 
 [component.wasi-http-async]
 source = "target/wasm32-wasi/release/wasi_http_rust_streaming_outgoing_body.wasm"
-allowed_http_hosts = ["insecure:allow-all"]
+allowed_outbound_hosts = ["https://*:*"]
```
itowlson, line 15: Would this still allow HTTP (not HTTPS) for localhost scenarios?

examples/http-tinygo-outbound-http/spin.toml
```diff
@@ -16,9 +16,9 @@ component = "outbound-http-to-same-app"
 
 [component.tinygo-hello]
 source = "tinygo-hello/main.wasm"
-allowed_http_hosts = [
-    "https://random-data-api.fermyon.app",
-    "https://postman-echo.com",
+allowed_outbound_hosts = [
+    "https://random-data-api.fermyon.app:443",
```
itowlson, line 20: I liked the ergonomics of not having to specify the default port.

tschneidereit:
We just discussed this in the Spin Up meeting, and I think we should hold off on removing the existing way of doing this this shortly before the release.

Landing this change requires a lot of documentation update, solid work on devUX to ensure that developers aren't left stranded because they don't know how to migrate, and potentially also some deeper conversation about the exact design and ergonomics of `allowed_outbound_hosts`.

examples/http-tinygo-outbound-http/spin.toml
rylev: We could make `https://random-data-api.fermyon.app` an alias for `https://random-data-api.fermyon.app:443` (i.e., if the port is not specified than the default port for the scheme is assumed).

examples/wasi-http-rust-streaming-outgoing-body/spin.toml
rylev: I think in the current implementation it would but that seems a bit odd. The most consistent thing would be to require `{http, https}://*:*` (note that the list syntax is not yet implemented), but that is admittedly a bit annoying.

examples/rust-outbound-pg/spin.toml
rylev: We're not doing literal equality checking on strings: the scheme in this config signifies that you are allowed to make postgres outbound calls to that host and port (and not, for example, http calls).

rylev:
Ok. I've pushed some changes that I think address most of the issues:

## `allowed_http_hosts` in v2

We will continue to support `allowed_http_hosts` in v2 manifests but we will emit a warning when it is used.

For example, say I have the following in my manifest:

```toml
allowed_http_hosts = ["foo.com"]
```

This will work as expected (http traffic to foo.com will be allowed), but the following warning will be emitted:

```
Warning: Use of the deprecated field `allowed_http_hosts` - to fix, replace the use of `allowed_http_hosts` with `allowed_outbound_hosts = ["https://foo.com:443"]
```

## Default ports

The `allowed_outbound_hosts` syntax now allows for omitting the port. If the port is omitted, the scheme is used to determine which port is allowed. For example, an `https` scheme allows for port 443 and a `redis` scheme allows for port 6379. 

## A full description of the `allowed_outbound_hosts` syntax

The `allowed_outbound_hosts` syntax is composed of three parts: a scheme, a host, and an optional port in the following form:

`$SCHEME://$HOST:$PORT`

Each of these elements can be either a specific string (e.g., scheme can be 'http' or 'redis') or a glob which allows anything. Additionally ports allow for port ranges using the syntax ($START..$END). In the future, we can easily extend this to more flexible syntax such as lists (e.g., `{redis, mysql}://example.com:{3000, 4000, 5000..5999}`). 

* **Scheme**: The scheme decides which type of request is allowed. For example, if you specify 'redis' as the scheme than redis requests are allowed but postgres, mysql, http, etc. are not. Note that this is independent of whether the url the user is using in their code has the scheme or not. Making a request from the user's code to 'example.com' or 'redis://example.com' are both permitted when the `allowed_outbound_hosts` list includes 'redis://example.com'.  
* **Host**: the host must match exactly between the url the user provides to an outbound networking interface being used and the configuration in `allowed_outbound_hosts`. For example Redis requests to 'example.com' are permitted by `allowed_outbound_hosts = ["redis://example.com"]` but requests to 'subdomain.example.com' are not (even if 'subdomain.example.com' redirects to 'example.com'). 
* **Port**: The port used to make the request must be in the range specified by the config. For example, a request to 'example.com:4001' is allowed by `allowed_outbound_hosts = ["*://example.com:4001"]` and `allowed_outbound_hosts = ["*://example.com:4000..4999"]` but is not allowed by `allowed_outbound_hosts = ["*://example.com:4000"]`.

examples/http-tinygo-outbound-http/spin.toml
rylev: This has been done (see [this comment](https://github.com/fermyon/spin/pull/1988#issuecomment-1786949253))

crates/loader/tests/ui/insecure-allow-all-with-invalid-url.err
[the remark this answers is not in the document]
rylev: Ah yes - saw this yesterday, and I didn't commit the delete I guess.

crates/manifest/src/schema/v2.rs
```diff
@@ -119,9 +119,31 @@ pub struct Component {
...
+        if !normalized.is_empty() {
+            terminal::warn!(
+                "Use of the deprecated field `allowed_http_hosts` - to fix, \
+            replace the use of `allowed_http_hosts` with `allowed_outbound_hosts = [{}]`",
+                normalized
+                    .iter()
+                    .map(|s| format!("\"{s}\""))
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            )
```
lann, line 137: :shrug: 
```suggestion
            replace the use of `allowed_http_hosts` with `allowed_outbound_hosts = {normalized:?}`")
```

crates/outbound-networking/src/lib.rs
```diff
@@ -125,194 +254,285 @@ fn well_known_port(scheme: &str) -> Option<u16> {
...
     fn default() -> Self {
         Self::SpecificHosts(Vec::new())
     }
 }
 
-#[derive(PartialEq, Eq, Debug)]
-pub struct AllowedHost {
-    scheme: Option<String>,
+#[derive(Debug, Clone)]
+pub struct Url {
```
lann, line 306: Probably shouldn't have a crate export named `Url`; could cause deep confusion if `use`d elsewhere (like how I was just confused seeing `Url::parse(url, "https")` above :sweat_smile:).

crates/outbound-http/src/host_impl.rs
```diff
@@ -28,11 +27,13 @@ impl OutboundHttp {
     /// If `None` is passed, the guest module is not allowed to send the request.
     fn is_allowed(&mut self, url: &str) -> Result<bool, HttpError> {
         if url.starts_with('/') {
-            return Ok(self.allowed_hosts.allows_relative_url());
+            return Ok(self.allowed_hosts.allows_relative_url("http"));
```
lann, line 30: Is this going to fail if "self" is using https?

crates/outbound-networking/src/lib.rs
```diff
@@ -1,122 +1,251 @@
...
-                    // Return an error preferring the error from the first attempt if present
-                    (_, Err(e)) | (Err(e), _) => Err(e),
+    pub fn parse(url: impl Into<String>) -> anyhow::Result<Self> {
+        let original = url.into();
+        let url = original.trim();
+        let scheme_end = url.find("://").with_context(|| {
+            format!("{url:?} does not contain a scheme (e.g., 'http://' or '*://'")
+        })?;
+        let scheme = &url[..scheme_end];
+        let host_and_port = &url[(scheme_end + 3)..];
```
lann, line 64: Nice convenience method `split_once` for this:
```suggestion
        let (scheme, host_and_port) = url.split_once("://").with_context(|| {
            format!("{url:?} does not contain a scheme (e.g., 'http://' or '*://'")
        })?;
```

examples/http-tinygo-outbound-http/spin.toml
```diff
@@ -16,9 +16,9 @@ component = "outbound-http-to-same-app"
...
+    "https://postman-echo.com:443",
```
lann, line 21: Can drop the ports here now?

crates/manifest/src/schema/v2.rs
lann: I guess if I'm going to double-check my own suggestion I may as well post it: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c73cc30e708a249bae003282a604097a

crates/outbound-http/src/host_impl.rs
rylev: Yes, We should probably allow either.

crates/manifest/src/schema/v2.rs
```diff
@@ -119,9 +119,26 @@ pub struct Component {
...
+impl Component {
+    /// Combine `allowed_outbound_hosts` with the deprecated `allowed_http_hosts` into
+    /// one array all normalized to the syntax of `allowed_outbound_hosts`.
+    pub fn normalized_allowed_outbound_hosts(&self) -> anyhow::Result<Vec<String>> {
+        let normalized =
+            crate::compat::convert_allowed_http_to_allowed_hosts(&self.allowed_http_hosts, false)?;
+        if !normalized.is_empty() {
+            terminal::warn!(
+                "Use of the deprecated field `allowed_http_hosts` - to fix, \
+            replace the use of `allowed_http_hosts` with `allowed_outbound_hosts = {normalized:?}`",
```
vdice, line 131: suggestion: thoughts on reducing to just the following?

```suggestion
            replace with `allowed_outbound_hosts = {normalized:?}`",
```

lann (approved):

crates/manifest/src/schema/v2.rs
itowlson: @vdice maybe 
```suggestion
            replace `allowed_http_hosts` with `allowed_outbound_hosts = {normalized:?}`",
```
so as to be absolutely specific?

rylev: If there are no other changes, I might sneak this in to a follow-up PR to avoid having to run CI again, but otherwise 👍

itowlson (approved):
Thanks for working through what must at times have been frustrating feedback.  I've not fully grasped the internals of the PR, but it looks good, and the behaviour gives us the long-term plan going forward while making users following older content successful.  The ideal outcome.

itowlson:
I will update the docs to align with this.  (Thanks for picking up the templates!)

itowlson:
@vdice heads up for the JS and Python templates

Issue #1985, opened 2023-10-30
I/O error with outbound HTTP and streaming responses


itowlson:
Maybe this is user error but I can't figure it out, sorry.

I have a server which streams a response over a period of 2 seconds:

```rust
// UPSTREAM SERVER
use futures::SinkExt;
use spin_sdk::http::{Fields, IncomingRequest, OutgoingResponse, ResponseOutparam};
use spin_sdk::http_component;

#[http_component]
async fn handle_stmtest(_req: IncomingRequest, response_out: ResponseOutparam) {
    let response = OutgoingResponse::new(
        200,
        &Fields::new(&[("content-type".to_string(), b"text/plain".to_vec())]),
    );
    let mut body = response.take_body();

    response_out.set(response);

    for i in 1..20 {
        let payload = format!("Hello {i}\n");
        if let Err(e) = body.send(payload.into()).await {
            eprintln!("Error sending payload: {e:#}");
            return;
        }
        std::thread::sleep(std::time::Duration::from_millis(100));
    }
}
```

When I curl this, it works fine; I see "Hello 1" to "Hello 19" returned over a short period.

Now I try to use it from a Spin client, essentially proxying it as it streams:

```rust
// PSEUDO-PROXY
use futures::{TryStreamExt, SinkExt};
use http::Request;
use spin_sdk::http::{IncomingRequest, ResponseOutparam, OutgoingResponse, Headers, IncomingResponse, send};
use spin_sdk::http_component;

#[http_component]
async fn handle_httpouttest(_req: IncomingRequest, response_out: ResponseOutparam) {
    let outbound_req = Request::get("http://localhost:7071").body(()).unwrap();  // running it here

    // Wire up the downstream response sink
    let response = OutgoingResponse::new(200, &Headers::new(&[]));
    let mut sink = response.take_body();
    response_out.set(response);

    // Send the request to upstream, and stream the response body
    let out_response: IncomingResponse = send(outbound_req).await.unwrap();
    let mut rbody = out_response.into_body_stream();

    // Every time data is available from upstream, send it downstream
    loop {
        let Ok(maybe_chunk) = rbody.try_next().await else {
            return;
        };
        if let Some(chunk) = maybe_chunk {
            sink.send(" *** CHUNK INCOMING *** ".into()).await.unwrap();
            sink.send(chunk).await.unwrap();
        } else {
            println!("nowt");
            std::thread::sleep(std::time::Duration::from_millis(200));
        }
    }
}
```

[...]

PR #1991, opened 2023-10-30
add test/example for outbound request body streaming

dicej:


sdk/rust/src/http.rs
```diff
@@ -461,14 +461,24 @@ impl IncomingResponse {
...
     ///
     /// Panics if the body was already consumed.
-    pub fn into_body_stream(self) -> impl futures::Stream<Item = Result<Vec<u8>, streams::Error>> {
+    // TODO: This should ideally take ownership of `self` and be called `into_body_stream` (i.e. symmetric with
+    // `IncomingRequest::into_body_stream`).  However, as of this writing, `wasmtime-wasi-http` is implemented in
+    // such a way that dropping an `IncomingResponse` will cause the request to be cancelled, meaning the caller
+    // won't necessarily have a chance to send the request body if they haven't started doing so yet (or, if they
+    // have started, they might not be able to finish before the connection is closed).  See
+    // https://github.com/bytecodealliance/wasmtime/issues/7413 for details.
+    pub fn take_body_stream(&self) -> impl futures::Stream<Item = Result<Vec<u8>, streams::Error>> {
```
itowlson, line 470: @dicej is this a planned change to the 2.0 SDK?

(Also, would this explain why your sample works and #1985 drops?)

examples/wasi-http-rust-streaming-outgoing-body/src/lib.rs
```diff
@@ -83,18 +89,121 @@ async fn handle_request(request: IncomingRequest, response_out: ResponseOutparam
...
-            response_out.set(response);
+            if let Some(url) = headers.iter().find_map(|(k, v)| {
+                (k == "url")
+                    .then_some(v)
+                    .and_then(|v| std::str::from_utf8(v).ok())
+                    .and_then(|v| Url::parse(v).ok())
+            }) {
+                match double_echo(request, &url).await {
+                    Ok((request_copy, response)) => {
+                        let mut stream = response.take_body_stream();
```
rylev, line 103: Finding it a bit hard to follow with the generic naming: perhaps using slightly more verbose names might help?

```suggestion
                        let mut incoming_response_body = response.take_body_stream();
```

examples/wasi-http-rust-streaming-outgoing-body/src/lib.rs
```diff
@@ -83,18 +89,121 @@ async fn handle_request(request: IncomingRequest, response_out: ResponseOutparam
...
+
+                        let response = OutgoingResponse::new(
+                            200,
+                            &Headers::new(
+                                &headers
+                                    .into_iter()
+                                    .filter_map(|(k, v)| (k == "content-type").then_some((k, v)))
```
rylev, line 110: Is this not equivalent?
```suggestion
                                    .filter(|(k, v)| k == "content-type")
```

examples/wasi-http-rust-streaming-outgoing-body/src/lib.rs
```diff
@@ -83,18 +89,121 @@ async fn handle_request(request: IncomingRequest, response_out: ResponseOutparam
...
+                                    .collect::<Vec<_>>(),
+                            ),
+                        );
+
+                        let mut body = response.take_body();
```
rylev, line 115: Same as above

```suggestion
                        let mut outgoing_response_body = response.take_body();
```

sdk/rust/src/http.rs
```diff
@@ -461,14 +461,24 @@ impl IncomingResponse {
     /// # Panics
     ///
     /// Panics if the body was already consumed.
-    pub fn into_body_stream(self) -> impl futures::Stream<Item = Result<Vec<u8>, streams::Error>> {
+    // TODO: This should ideally take ownership of `self` and be called `into_body_stream` (i.e. symmetric with
```
rylev, line 464: Sigh... I think this whole API might need another look - it's hard to get the flow right. Oh well...

sdk/rust/src/http.rs
dicej: Yes and yes.  The API change is necessary because there's no other way to work around https://github.com/bytecodealliance/wasmtime/issues/7413, and I'm not sure when that will be fixed or whether it will be backported to Wasmtime 14 (and if it is, probably not in time for Spin 2.0).

dicej: Update: here's a PR with the fix: https://github.com/bytecodealliance/wasmtime/pull/7426

PR #1998, opened 2023-10-31
rust: Small improvements to HTTP types

lann:
- Better `Debug` for bodies
- `impl IntoResponse for ResponseBuilder`

### DIFF
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -150,6 +150,17 @@ impl Request {
     }
 }
 
+impl std::fmt::Debug for Request {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Request")
+            .field("method", &self.method)
+            .field("uri", &self.uri)
+            .field("headers", &self.headers)
+            .field("body", &DebugBody(&self.body))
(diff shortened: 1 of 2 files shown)


sdk/rust/src/http.rs
```diff
@@ -705,6 +716,21 @@ impl std::fmt::Display for streams::Error {
 
 impl std::error::Error for streams::Error {}
 
+struct DebugBody<'a>(&'a [u8]);
```
rylev, line 719: I know I suggested this, and we're already debug printing headers, but are we worried about security implications of this? This can easily leak sensitive information into logs.

rylev (commented):
I like the changes, but this does make me wonder if we've fully thought through the data privacy aspects of such a change.

sdk/rust/src/http.rs
lann: :shrug: I'm ambivalent on this one. As an appeal to convention, `http::Request` appears to  dump everything: https://docs.rs/http/0.2.9/src/http/request.rs.html#704-715

rylev (approved):

dicej:
Is this still relevant?  If so, we'll want to move it over to the `spin-rust-sdk` repo.

Issue #1979, opened 2023-10-27
Outbound Redis: not always deny by default


vdice:
https://github.com/fermyon/spin/pull/1933 added logic intended to deny outbound requests to redis for apps using the v2 manifest, unless the host is explicitly allowed via `allowed_outbound_hosts` config.

Indeed, this is the case for some apps, eg the [rust-outbound-redis](https://github.com/fermyon/spin/tree/main/examples/rust-outbound-redis) example.

However, other apps _don't_ see the same deny-by-default behavior.  This can be reproduced via the [tinygo-outbound-redis app](https://github.com/fermyon/spin/tree/main/examples/tinygo-outbound-redis) (you may need to remove the `allowed_outbound_hosts` configuration to repro if it has been applied via https://github.com/fermyon/spin/pull/1978)

lann:
The intention was to treat the absence of `allowed_outbound_hosts` as "allow all" for backward compatibility, at least for now.

itowlson:
I thought that was only for v1 _manifests_ (because if you were updating the manifest you could add the allow at that point).  Are we doing allow-all for v1 SDKs in v2 manifests as well?

vdice:
Even after https://github.com/fermyon/spin/pull/1987, I'm still seeing the following inconsistency between outbound-mysql/pg behavior.

Take [rust-outbound-mysql](https://github.com/fermyon/spin/tree/main/examples/rust-outbound-mysql) and [tinygo-outbound-mysql](https://github.com/fermyon/spin/tree/main/examples/tinygo-outbound-mysql) (both now on v2 manifest):

- With `allowed_outbound_hosts=["mysql://127.0.0.1"]`: both behave as expected (requests allowed)
- With `allowed_outbound_hosts=[]`: both behave as expected (requests denied)
- With no `allowed_outbound_hosts` config: rust denies by default, tinygo allows

I'm not sure if this is expected given current conditions, etc, but thought I'd double check.

rylev:
This is the currently expected behavior. deny-by-default is controlled by which interfaces you use. Rust uses the v2 interfaces and thus always respects `allowed_outbound_hosts`. Go uses the v1 interfaces and thus only respects `allowed_outbound_hosts` if its explicitly set.

The thought behind this was we don't want to break old apps - i.e., if it worked in Spin 1.0, it should work in Spin 2.0.

[...]

PR #2004, opened 2023-10-31
Move to deny-by-default for everything but v1 SDKs and v1 manifests

rylev:
Potential fix for https://github.com/fermyon/spin/issues/1979 specifically implementing [this idea](https://github.com/fermyon/spin/issues/1979#issuecomment-1787797365).

### DIFF
--- a/crates/manifest/src/schema/v2.rs
+++ b/crates/manifest/src/schema/v2.rs
@@ -128,7 +128,7 @@ impl Component {
         if !normalized.is_empty() {
             terminal::warn!(
                 "Use of the deprecated field `allowed_http_hosts` - to fix, \
-            replace the use of `allowed_http_hosts` with `allowed_outbound_hosts = {normalized:?}`",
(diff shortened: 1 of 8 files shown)


itowlson (approved):
Can't comment on the implementation but I'm in favour of making v2 manifests deny-by-default regardless of SDK version.

vdice (commented):
> Can't comment on the implementation but I'm in favour of making v2 manifests deny-by-default regardless of SDK version.

Ditto ^^.  Note: a quick test with the tinygo-outbound examples (eg [tinygo-outbound-mysql](https://github.com/fermyon/spin/tree/f2bdcdb88ced6f194aa88285065d5c82662a67e0/examples/tinygo-outbound-mysql)) showed allowed-by-default behavior (v2 manifest albeit v1 interface).  Perhaps this can be double-checked.

vdice (approved):
👍 Re-tested and looking good!

PR #2007, opened 2023-11-01
chore(*): bump to 2.0.0-rc.1

vdice:


itowlson (approved):
I will confess to not reading _all_ of the `Cargo.lock` files

PR #2019, opened 2023-11-01
log error if `wasi:http/incoming-handler#handle` fails post-response

dicej:
Previously, we were simply ignoring the result of the guest invocation once it produced a response, which meant if it trapped after that point nothing would be logged.  Now we always log if there's any kind of error.

### DIFF
--- a/crates/trigger-http/src/handler.rs
+++ b/crates/trigger-http/src/handler.rs
@@ -3,6 +3,7 @@ use std::{net::SocketAddr, str, str::FromStr};
 use crate::{Body, HttpExecutor, HttpTrigger, Store};
 use anyhow::bail;
 use anyhow::{anyhow, Context, Result};
+use futures::TryFutureExt;
 use http::{HeaderName, HeaderValue};
 use http_body_util::BodyExt;
 use hyper::{Request, Response};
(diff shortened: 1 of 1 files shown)


crates/trigger-http/src/handler.rs
```diff
@@ -186,7 +187,23 @@ impl HttpHandlerExecutor {
...
+                            .await
+                            .context("guest invocation panicked")?
+                            .context("guest invocation failed")?;
+
+                        Ok(())
+                    }
+                    .map_err(|e: anyhow::Error| {
+                        tracing::warn!("component error after response: {e:?}");
+                    }),
+                );
```
rylev, line 203: I'm confused about how we expect this to work, and I realize I must have not understood how this was supposed to work before either.

`response` is what we ultimately get back from the guest call to `ResponseOutparam::set`, right? Which means after this point the guest might spend an unbounded amount of time writing the response body.

Is this not a race condition where `execute_wasi` might return before the handler is actually finished? Wouldn't we need to await `handle` to know when the guest is actually done?

So, I guess my question is: why do we spawn a task here again? Shouldn't we be awaiting this on the current task and not spawning a new one?

rylev (changes requested):

crates/trigger-http/src/handler.rs
dicej: Yes, `execute_wasi` will normally return before the response is fully written.  And yes, we need to await `handle` to know when the guest is actually done, which is what this patch does.  We have to spawn a task to do that because we *need* to return from `execute_wasi` as soon as the response is set -- that's how we hand the response to Hyper so it can start delivering it back to the client.  If we don't return, we'll deadlock due to backpressure when the guest tries to write the response body but Hyper has not yet had a chance to start reading it and pass it on to the client.

lann (approved):

rylev (approved):

vdice:
Is this a v2.0 backport candidate?

dicej:
Yes; it's on my todo list to create a backport PR, but haven't gotten to it yet.

vdice:
I got u fam: https://github.com/fermyon/spin/pull/2034

dicej:
Thanks @vdice!  I wish there was a `backport` button in GitHub.

PR #2025, opened 2023-11-01
Don't allow v2 components to be added to v1 manifest

### LINKED ISSUE #2016: `spin add` doesn't check for manifest version mis-match
When using the `spin add` command with a template using the manifest v1 format, with an application, which already has the manifest in v2, the command completes, and you end up with a non-working manifest.

It would be great to have something checking if there's a match between the manifest used in the app, and the manifest version being brought by the template.

itowlson:
Fixes #2016

### DIFF
--- a/crates/templates/src/template.rs
+++ b/crates/templates/src/template.rs
@@ -358,6 +358,43 @@ impl Template {
             }
         }
     }
+
+    pub(crate) fn check_compatible_manifest_format(
+        &self,
+        manifest_format: u32,
+    ) -> anyhow::Result<()> {
+        let Some(content_dir) = &self.content_dir else {
+            return Ok(());
+        };
+        let manifest_tpl = content_dir.join("spin.toml");
+        if !manifest_tpl.is_file() {
+            return Ok(());
+        }
+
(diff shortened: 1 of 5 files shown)


crates/templates/src/template.rs
```diff
@@ -358,6 +358,43 @@ impl Template {
...
+        if !manifest_tpl.is_file() {
+            return Ok(());
+        }
+
+        // We can't load the manifest template because it's not valid TOML until
+        // substituted, so GO BIG or at least GO CRUDE.
+        let Ok(manifest_tpl_str) = std::fs::read_to_string(&manifest_tpl) else {
+            return Ok(());
+        };
+        let is_v1_tpl = manifest_tpl_str.contains("spin_manifest_version = \"1\"");
```
michelleN, line 379: why do we escape quotes here and not for spin_manifest_version = 2?

michelleN (approved):

crates/templates/src/template.rs
itowlson: In v1 there are quotes around the 1.  In v2 there are no quotes around the 2.

fibonacci1729 (approved):

PR #2027, opened 2023-11-02
`spin add` check trigger compatibilty for v2

itowlson:
`spin add` has traditionally complained if you added a HTTP trigger to a Redis app.  That currently doesn't happen in v2, because the parsing code doesn't work with the templatised version of the v2 manifest (it's no longer valid TOML), and the system defaults to "if we can't figure it out, do what the user asked."

This PR recreates the trigger check.

The parsing code is terrible, and there is some duplication.  The idea for now is to catch the common case and improve it in a later pass.


crates/templates/src/app_info.rs
```diff
@@ -85,5 +136,34 @@ struct ManifestV1TriggerProbe {
 #[derive(Deserialize)]
 struct ManifestV2TriggerProbe {
     // `[trigger.<type>]`
-    trigger: toml::value::Table,
+    trigger: Option<toml::value::Table>, // empty will not have a trigger table in v2
```
rylev, line 139: nit: can you make the comment here and the comment above into doc comments?

crates/templates/src/app_info.rs
```diff
@@ -67,15 +72,61 @@ impl AppInfo {
         })
     }
 
+    fn from_existent_template(manifest_path: &Path) -> anyhow::Result<Self> {
+        // This has to be cruder, because (with the v2 style of component) a template
+        // is no longer valid TOML, so `from_existent_file` fails at manifest
+        // version inference.
+        let manifest_tpl_str = std::fs::read_to_string(manifest_path)?;
```
rylev, line 79: Bit of a shame we have to read the whole file here, but the alternative is probably not worth it.

crates/templates/src/app_info.rs
```diff
@@ -67,15 +72,61 @@ impl AppInfo {
...
+
+        // TODO: a better way
```
rylev, line 81: Not sure this comment is actually helpful...

crates/templates/src/app_info.rs
```diff
@@ -12,13 +15,13 @@ use crate::store::TemplateLayout;
 
 pub(crate) struct AppInfo {
     manifest_format: u32,
-    trigger_type: String,
+    trigger_type: Option<String>,
```
rylev, line 18: Can we comment on what a `None` for `trigger_type` signifies.

rylev (commented):
I don't understand the `spin add` flow enough to really comment on this, but the code itself looks fine.

kate-goldenring (commented):
I am assuming we will need to remove the `>= 1` trigger check once we support multiple triggers right?

itowlson:
@kate-goldenring once we support multiple trigger types, the whole subsystem goes away because there is no need to check that the incoming templates has the same trigger as the existing app, so yeah the `<= 1` check will go away as part of that mass removal.

crates/templates/src/app_info.rs
```diff
@@ -67,15 +72,61 @@ impl AppInfo {
         })
     }
 
+    fn from_existent_template(manifest_path: &Path) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
itowlson: Good idea: it's tested via the manager tests but a more focused test would be helpful for something this kludgy.

crates/templates/src/app_info.rs
```diff
@@ -67,15 +72,61 @@ impl AppInfo {
...
+        // This has to be cruder, because (with the v2 style of component) a template
+        // is no longer valid TOML, so `from_existent_file` fails at manifest
+        // version inference.
+        let manifest_tpl_str = std::fs::read_to_string(manifest_path)?;
+
+        // TODO: a better way
+        let is_v1_tpl = manifest_tpl_str.contains("spin_manifest_version = \"1\"");
```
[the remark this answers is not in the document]
itowlson: Not unless we think someone is going to use Spin 0.9 templates with Spin 2 (or unless there are third party templates out there that haven't been updated since Spin 0.9).  And it defaults to forgiving, so if there _are_ very old templates out there, they will be treated as "allow the add" rather than anything blocking.

On the other hand, yes, I would like this check to be more aligned with the actual TOML parser!

kate-goldenring (approved):
lgtm!

PR #2028, opened 2023-11-02
[Backport v2.0] Don't allow v2 components to be added to v1 manifest

itowlson:
Backporting #2025 to v2.0

### DIFF
--- a/crates/templates/src/app_info.rs
+++ b/crates/templates/src/app_info.rs
@@ -11,6 +11,7 @@ use spin_manifest::schema::v1;
 use crate::store::TemplateLayout;
 
 pub(crate) struct AppInfo {
+    manifest_format: u32,
     trigger_type: String,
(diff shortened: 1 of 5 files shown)


itowlson:
@vdice I've put this in because it's something Mikkel raised and does not affect user or developer experience.  But feel free to decline it if you want to keep 2.0 as frozen as possible - I'm completely sympathetic to that!

vdice:
@itowlson I'm fine with backporting if you'd like to see it in.  I didn't get a chance to test this logic -- is the error obvious to the user when they add a componnent with a mismatched manifest version?

itowlson:
@vdice I hope so.  It prints `Error: This template is for a different version of the Spin manifest` - I admit this could be better but hopefully enough.

vdice (approved):

PR #2033, opened 2023-11-02
update seemingly deprecated template flag

goshlanguage:
Hey team,

When kicking the tires, I noticed the readme has a `-t` flag, which I presume is for template. 
I have just downloaded `spin` today, so I don't suspect my binary is out of date, but it's possible.

Here's what I saw:
```sh
spin new --accept-defaults -t http-rust hello-rust
error: Found argument '-t' which wasn't expected, or isn't valid in this context

	If you tried to supply `-t` as a value rather than a flag, use `-- -t`

USAGE:
    spin new [OPTIONS] [ARGS]

```

And here's what I found worked:

```sh
spin new --accept-defaults http-rust hello-rust
```

Updated the README accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ $ rustup target add wasm32-wasi
 First, run the `spin new` command to create a Spin application from a template.
 ```bash
 # Create a new Spin application named 'hello-rust' based on the Rust http template, accepting all defaults
-$ spin new --accept-defaults -t http-rust hello-rust
+$ spin new --accept-defaults http-rust hello-rust
 ```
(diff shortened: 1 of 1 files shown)


vdice:
Hi @goshlanguage apologies for the confusion here.  You're right: the latest [v1.5.1](https://github.com/fermyon/spin/releases/tag/v1.5.1) release of Spin doesn't have the `-t/--template` flag.  This flag has since been added to Spin `main` and is in the [canary release](https://github.com/fermyon/spin/releases/tag/canary).  We merged the README update recently per the forthcoming v2.0 release due out soon.  We realize there will be muscle memory with `spin new` so the old (pre-canary/2.0) invocation style will still be supported along with the new for a period of time.  Hope this helps!

PR #2034, opened 2023-11-02
[Backport v2.0] log error if `wasi:http/incoming-handler#handle` fails post-response

vdice:
Backporting #2019 to v2.0


dicej (approved):
Thanks!

PR #2036, opened 2023-11-02
outbound-networking: More permissive allow-all

lann:
Two changes here that were regressions in LHC's understanding of allow-all:

- `insecure:allow-all` allows http and https
- `*:*` implies `self`

_Should backport for 2.0 if merged_

### DIFF
--- a/crates/manifest/src/compat.rs
+++ b/crates/manifest/src/compat.rs
@@ -120,7 +120,11 @@ pub(crate) fn convert_allowed_http_to_allowed_hosts(
         Vec::new()
     };
     match http_hosts {
-        AllowedHttpHosts::AllowAll => outbound_hosts.extend(["https://*:*".into()]),
+        AllowedHttpHosts::AllowAll => outbound_hosts.extend([
+            "http://*:*".into(),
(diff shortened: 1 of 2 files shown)


crates/manifest/src/compat.rs
```diff
@@ -120,7 +120,11 @@ pub(crate) fn convert_allowed_http_to_allowed_hosts(
         Vec::new()
     };
     match http_hosts {
-        AllowedHttpHosts::AllowAll => outbound_hosts.extend(["https://*:*".into()]),
+        AllowedHttpHosts::AllowAll => outbound_hosts.extend([
+            "http://*:*".into(),
+            "https://*:*".into(),
+            "http://self".into(),
```
itowlson, line 126: Does this also need `https://self` or is `http://self` a special case that allows both?

rylev: It's not that `http://self` is special, but when a request is made to `self` we check that there is either permission for `http://self` or `https://self`.

rylev (approved):
Looks good to me!

lann:
Accidentally pushed to main...

PR #2037, opened 2023-11-02
Use Wasmtime patch for WASI HTTP tasks

radu-matei:
Use Wasmtime patch for WASI HTTP tasks.

_This should be backported to 2.0._


alexcrichton:
I appended a commit here which fixes the e2e tests, although there's still some SDK tests that are failing. Are those perhaps known to be flaky or are those also likely a cause of these changes?

radu-matei:
Not sure, I queued the failing checks for a re-run.

PR #2038, opened 2023-11-02
read incoming request bodies lazily

dicej:
This currently entails terrible code duplication, even more terrible error handling, and miscellaneous crimes of various severity.

The main goal for the time being is to have something to benchmark against.  We can clean up (some of? all of?) the crimes prior to merging if the benchmarks look good.

### DIFF
--- a/sdk/rust/macro/src/lib.rs
+++ b/sdk/rust/macro/src/lib.rs
@@ -106,7 +106,7 @@ pub fn http_component(_attr: TokenStream, item: TokenStream) -> TokenStream {
                     let request: ::spin_sdk::http::IncomingRequest = ::std::convert::Into::into(request);
                     let response_out: ::spin_sdk::http::ResponseOutparam = ::std::convert::Into::into(response_out);
                     ::spin_sdk::http::run(async move {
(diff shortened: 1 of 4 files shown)


rylev (commented):
I have lots of opinions on the individual code crimes here, but I'll resist commenting on those until we decide this is worth doing.

On the point of whether this is actually worth doing, I think I'd really have to see a very large improvement in performance. Given that the `http` crate's body is also eager (unless using a lazy body type), it seems there is precedent for having an eagerly evaluated body.

dicej:
One thing I learned when looking more closely at the `http` crate is that it's pretty much totally agnostic about the body type.  It's just an unbounded generic, so it can be anything you want: `Vec`, `Bytes`, `Lazy`, `impl Stream`, `impl AsyncRead` -- whatever.  So I'd say it leaves the "eagerly evaluated vs. lazy" decision entirely up to the user.

dicej:
Sounds like the current thinking is that the performance issue this is intended to address should be fixed on the host side after all, so I'm going to close this.  We can reopen later if we change our minds.

PR #2039, opened 2023-11-03
[Backport v2.0] outbound-networking: More permissive allow-all

lann:
Backporting #2036 to v2.0


vdice:
@lann any ideas on the test failure?

lann:
Ugh yes. I accidentally pushed the last revision of the PR to `main` but I never pushed the last revision to the PR...

PR #2042, opened 2023-11-03
When translating allowed_http_hosts allow both http and htttps

rylev:
`allowed_http_hosts` allowed both http and https so when translating to `allowed_outbound_hosts` we should respect that.


lann:
Needs more `BLESS`

lann (approved):

lann:
FYI, this will need `BLESS=1 cargo test` before CI will pass.

PR #2045, opened 2023-11-03
Fix incorrect instructions in outbound hosts error message

### LINKED ISSUE #2044: Error message for 'non-allowed host' is incorrect
```
Warning: A component tried to make a HTTP request to non-allowed host 'localhost'.
To allow requests, add 'allowed_http_hosts = ["localhost"]' to the manifest component section.
Cannot send HTTP request: Destination not allowed: http://localhost:8000
```

It should be 
```
To allow requests, add 'allowed_http_hosts = ["http:localhost:8000"]' to the manifest component section.
```

adamreese:
Fixes #2044

### DIFF
--- a/crates/outbound-http/src/host_impl.rs
+++ b/crates/outbound-http/src/host_impl.rs
@@ -166,7 +166,13 @@ fn response_headers(h: &HeaderMap) -> anyhow::Result<Option<Vec<(String, String)
 }
 
 fn host(url: &str) -> Option<String> {
-    url::Url::parse(url)
-        .ok()
-        .and_then(|u| u.host().map(|h| h.to_string()))
+    match url::Url::parse(url) {
+        Err(_) => None,
+        Ok(u) => Some(format!(
+            "{}://{}:{}",
+            u.scheme(),
+            u.host().unwrap(),
+            u.port().unwrap()
+        )),
+    }
 }


crates/outbound-http/src/host_impl.rs
```diff
@@ -166,7 +166,13 @@ fn response_headers(h: &HeaderMap) -> anyhow::Result<Option<Vec<(String, String)
...
-    url::Url::parse(url)
-        .ok()
-        .and_then(|u| u.host().map(|h| h.to_string()))
+    match url::Url::parse(url) {
+        Err(_) => None,
+        Ok(u) => Some(format!(
+            "{}://{}:{}",
+            u.scheme(),
+            u.host().unwrap(),
+            u.port().unwrap()
```
itowlson, line 175: The URL not including a port seems like a highly likely situation.  We should not panic if it happens - just return a host with no port.  (Heck, even if it's unlikely it doesn't merit a panic...)

crates/outbound-http/src/host_impl.rs
```diff
@@ -166,7 +166,13 @@ fn response_headers(h: &HeaderMap) -> anyhow::Result<Option<Vec<(String, String)
...
 fn host(url: &str) -> Option<String> {
-    url::Url::parse(url)
-        .ok()
-        .and_then(|u| u.host().map(|h| h.to_string()))
+    match url::Url::parse(url) {
+        Err(_) => None,
+        Ok(u) => Some(format!(
+            "{}://{}:{}",
+            u.scheme(),
+            u.host().unwrap(),
```
itowlson, line 174: This does not merit a panic.  It should return `None` as the original function did.

itowlson (changes requested):
AB2FB643-F1D7-456F-95A9-DE318FFF01D3-768x432-450245981

itowlson:
@adamreese At the risk of having my own Rust skills forensically disassembled, given that the function returns an Option, I would suggest using the `?` operator e.g. `u.host()?`.  If `host` returns None, this will have the function return None immediately.

For the port thing, though, I think you will need to pattern match, because you don't want to nope out just because the URL doesn't contain  a port. E.g. something like `let port_str = match u.port() { Some(p) => format!(":{p}"), None => "".to_owned() };` (only nicer).  Although as @rylev notes the syntax for `allowed_outbound_hosts` is different anyway.

fibonacci1729:
Closing since it appears #2136 addresses this.

PR #2053, opened 2023-11-04
Update internal command list

karthik2804:
Updating the list of internal command names that need to be checked when installing plugins.

### DIFF
--- a/crates/plugins/src/lib.rs
+++ b/crates/plugins/src/lib.rs
@@ -9,13 +9,20 @@ pub use store::PluginStore;
 
 /// List of Spin internal subcommands
 pub(crate) const SPIN_INTERNAL_COMMANDS: &[&str] = &[
+    "template",
     "templates",
     "up",
     "new",
     "add",
+    "login",
     "deploy",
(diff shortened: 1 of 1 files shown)


kate-goldenring (approved):
Thank you @karthik2804!

crates/plugins/src/lib.rs
```diff
@@ -9,13 +9,19 @@ pub use store::PluginStore;
...
     "add",
+    "login",
     "deploy",
     "build",
     "plugin",
+    "plugins",
     "trigger",
     "external",
+    "doctor",
+    "registry",
```
itowlson, line 25: `oci` is still accepted too. Yes, I had forgotten about it too.

itowlson (approved):
Good catch - thanks for this!  One tiny obscure nit it would be good to add, but not a blocker.

crates/plugins/src/lib.rs
karthik2804: added

michelleN (approved):

PR #2056, opened 2023-11-05
Error at `new` time if name is not usable

### LINKED ISSUE #2052: spin new: trouble with name using number
My first chosen names using cli `spin new` were wrong without warnings :

`2023-11-spin` => `spin build` raised an error : 
```
Building component 2023-11-spin with `cargo build --target wasm32-wasi --release`
error: failed to parse manifest at `/Users/francois/Documents/dev/github/franxois/2023-11-spin/Cargo.toml`

Caused by:
  the name `2023-11-spin` cannot be used as a package name, the name cannot start with a digit
Error: Build command for component 2023-11-spin failed with status Exited(101) 
```

`spin-2023-11` => `spin build` ok but not `spin up`

```
Error: TOML parse error at line 11, column 13
   |
11 | component = "spin-2023-11"
   |             ^^^^^^^^^^^^^^
data did not match any variant of untagged enum ComponentSpec
``` 

If I use `spin new -t http-rust 2023-11-spin` I have an error "Name must start with a letter" => OK
If I use `spin new -t http-rust spin-2023-11` no warning but error with `spin up`
If I use `spin new -t http-rust spin_2023_11` no warning but error with `spin up`
If I use `spin new -t http-rust spin-2` no warning but error with `spin up`

BTW, `spin-spin`, `spin2` or `spin_spin` are ok

`spin 2.0.0 (e4bb235 2023-11-03)`
```
cloud 0.5.1 [installed]
js2wasm 0.6.1 [installed]
py2wasm 0.3.2 [installed]
```

itowlson:
Fixes #2052.

```
# Error cases from original issue
ivan@hecate:~/testing$ spin new -t http-rust 2023-11-spin
error: Invalid value "2023-11-spin" for '<NAME>': Each segment of the name must start with a letter. '2023', '11' do not start with a letter

ivan@hecate:~/testing$ spin new -t http-rust spin-2023-11
error: Invalid value "spin-2023-11" for '<NAME>': Each segment of the name must start with a letter. '2023', '11' do not start with a letter

ivan@hecate:~/testing$ spin new -t http-rust spin_2023_11
error: Invalid value "spin_2023_11" for '<NAME>': Each segment of the name must start with a letter. '2023', '11' do not start with a letter

# Singular form
ivan@hecate:~/testing$ spin new -t http-rust spin-2
error: Invalid value "spin-2" for '<NAME>': Each segment of the name must start with a letter. '2' does not start with a letter

# This is caught but the error message is not ideal: is this required to fix?
ivan@hecate:~/testing$ spin new -t http-rust _foo
error: Invalid value "_foo" for '<NAME>': Each segment of the name must start with a letter. '' does not start with a letter

# Accepted
ivan@hecate:~/testing$ spin new -t http-rust spin_y2023_m11
```

(`add` uses the same logic and gets the same errors.)

Long term it would be good to centralise this logic, but the `new`/`add` name rules are slightly more forgiving than the manifest parser (because the templates kebab-case the things we care about kebab-casing), so I didn't tackle that here.

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -6,6 +6,7 @@ use std::{
 
 use anyhow::{anyhow, bail, Context, Result};
 use clap::Parser;
+use itertools::Itertools;
 use path_absolutize::Absolutize;
 use tokio;
 
@@ -343,7 +344,6 @@ async fn prompt_name(variant: &TemplateVariantInfo) -> anyhow::Result<String> {
 
 lazy_static::lazy_static! {
     static ref PATH_UNSAFE_CHARACTERS: regex::Regex = regex::Regex::new("[^-_.a-zA-Z0-9]").expect("Invalid path safety regex");
-    static ref NAME: regex::Regex = regex::Regex::new("^[a-zA-Z].*").expect("Invalid name regex");
 }
 
 fn path_safe(text: &str) -> PathBuf {
@@ -352,11 +352,25 @@ fn path_safe(text: &str) -> PathBuf {
 }
 
 fn validate_name(name: &str) -> Result<String, String> {
-    if NAME.is_match(name) {
-        Ok(name.to_owned())
-    } else {
-        Err("Name must start with a letter".to_owned())
+    let splits = name.split(|c| !char::is_alphanumeric(c));
+    let invalid_split_displays = splits
+        .filter(|s| !s.starts_with(char::is_alphabetic))
+        .map(|s| format!("'{s}'"))
+        .collect_vec();
+
+    if invalid_split_displays.is_empty() {
+        return Ok(name.to_owned());
     }
+
+    let invalid_text = invalid_split_displays.join(", ");
+    let verb = if invalid_split_displays.len() == 1 {
+        "does"
+    } else {
+        "do"
+    };
+
+    let msg = format!("Each segment of the name must start with a letter. {invalid_text} {verb} not start with a letter");
+    Err(msg)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


src/commands/new.rs
```diff
@@ -352,11 +353,25 @@ fn path_safe(text: &str) -> PathBuf {
 }
 
 fn validate_name(name: &str) -> Result<String, String> {
-    if NAME.is_match(name) {
-        Ok(name.to_owned())
-    } else {
-        Err("Name must start with a letter".to_owned())
+    let splits = name.split(|c| !char::is_alphanumeric(c));
+    let invalid_split_displays = splits
+        .filter(|s| !NAME.is_match(s))
```
rylev, line 358: Do we need the regex? Wouldn't `filter(|s| s.starts_with(char::is_ascii_alphabetic))` work?

src/commands/new.rs
```diff
@@ -352,11 +353,25 @@ fn path_safe(text: &str) -> PathBuf {
 }
 
 fn validate_name(name: &str) -> Result<String, String> {
-    if NAME.is_match(name) {
-        Ok(name.to_owned())
-    } else {
-        Err("Name must start with a letter".to_owned())
+    let splits = name.split(|c| !char::is_alphanumeric(c));
```
rylev, line 356: Can the separator legally be any non-alphanumeric character? Is "foo😴a" a legal name?

itowlson: Well, it works, so I guess it's de facto legal!  The name as written is only used for the app name, which Spin doesn't care about.  Component names get kebab-cased during processing, which turns that into `foo-a` which is a legal kebab-id.  And the generated directory gets snake-cased to `foo_a` which is a legal directory name.

michelleN (approved):

Issue #2057, opened 2023-11-06
Common component test suite


rylev:
Part of https://github.com/fermyon/spin/issues/2055

Testing the Spin runtime involves running through the following code paths:

* CLI arg parsing
* Manifest parsing
* Component instantiation
* Trigger set up (e.g., in the case of http triggers, setting up the http server)
* Individual trigger handling (e.g., in the case of http triggets, handling an incoming http request and invoking the correct component)

Having to run through all of these steps to test individual functionality means we'll be retesting large swaths of indentical functionality over and over again making testing much slower than it needs to be. 

At the core of reducing this duplication is the need for creating a common suite of test components that can be reused by multiple tests without the need for rebuilding each time. 

## Different approaches 

As specified in https://github.com/fermyon/spin/issues/2055, we want to be able to run against some sort of suite of pre-built components. However, there are many ways we could do this each with their own pros and cons. I'd like to outline them here for discussion:

### Full Spin components built with the SDK

This is closest to what we currently have. We create Spin components using the SDK. The only difference to the current approach is that when running tests, we would not build these components but instead use the pre-built WebAssembly binaries.

**Pros**:
* Easy to move to since it's very close to what we're currently doing.
* Easy to write - the SDKs exist to make writing Spin apps easier and so being able to use them will make writing tests easier

**Cons**:
* Conflates SDK and runtime testing - if the SDK changes, we might either want to change the tests or keep around code that uses old SDK versions (which might be confusing). This is all despite that we're not really testing the SDKs at all. 
* Repeats testing of handling code - since all Spin components need an exported handler, each test would have to repeat boiler plate of handler code. 

### Full Spin components built without the SDK

This is similar to the above just with the components not being written using the SDK. 

**Pros**
* Similar to the SDK approach and has some of its pros
* Does not conflate SDK testing with runtime testing

**Cons**:
* Harder to write than components built with an SDK

### Partial Spin components

This approach would require building components that are not fully compliant Spin components as they don't handler exports. This would only test a single imported interface (e.g., only testing SQLite).

**Pros**:
* Avoids repeatedly testing handler exports

**Cons**:
* Makes running tests much more difficult as these components cannot be tested against "off-the-shelf" Spin compliant runtimes. 

## Proposal

The following is my proposal for how we handle runtime testing:

* The test collection is composed of full Spin compliant components that can be used in their pre-built form by test suites for testing Spin compliant runtimes.
* These components will be built without the use of an SDK.
[...]

PR #2062, opened 2023-11-07
Runtime Testing POC

rylev:
An implementation of the ideas in https://github.com/fermyon/spin/issues/2057

This adds a common test component collection (currently consisting of only one component) together with a runtime test suite (currently consistently of 3 tests) that tests this component in various ways. The test component is a reusable component that can be used in many different tests.

The common test component collection currently only has one component that uses the sqlite interface. Of course, this will soon expand to having many more common components. The common component collection uses `cargo-component` instead of the Spin SDK to create the component to avoid needing to update common components when the SDK is updated.

The test suite consists of three tests that test this common sqlite component in various Spin configurations including when it doesn't have permission to the database it uses and when it doesn't first run the migration. The test suite reads from a directory structure which contains test definitions. Spin is then run inside of temporary folder against these definitions to produce a result (by convention either a 200 for successes and 500s for errors). If the result is expected the test passes and otherwise it fails. The test definitions are composed of Spin.toml files as well as optional arguments to Spin, runtime data, and an error.txt file if Spin is expected to fail.

Each test takes about 750ms to run since the components that are being tested no longer need to be built before the test is run. 

You can read more about all of this in the various READMEs.

## Future changes

The following are things I plan on doing immediately once we all agree on this as a good direction:
* Factor out common logic for making common components into their own crate
* Run the test suite against the current version of Spin (right now it runs against the Spin on the $PATH) as a part of `cargo test`
* Replace all of the end to end tests with these type of tests.


tests/runtime-tests/src/main.rs
```diff
@@ -0,0 +1,304 @@
+use std::{
+    io::{Read, Write},
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+
+use anyhow::Context;
+
+fn main() -> anyhow::Result<()> {
```
lann, line 9: A test harness like this would be useful for user app testing too: https://github.com/fermyon/spin/issues/2063

test-components/sqlite/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Tests the SQLite interface.
+
+## Expectations
+
+This test component expects the following to be true:
+* It is given permission to open a connection to the "default" database.
+* That database has one table named `test_data` with two text columns (`key` and `value`)
+* `test_data` has at least one row with `key` set to `my_key` and value set to `my_value`.
+* It does not have permission to access a database named "forbidden".
```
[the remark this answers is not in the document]
rylev: I'm not sure I agree fully. I do agree that code comments tend to remain fresh for longer than READMEs. But the audience of this documentation isn't really expected to read the code. They are authors of test cases who want to use this component in their test. The readme aims to give a general idea of the scope of what the component is testing and what outside expectations it has.

rylev:
@lann I've addressed (most) of your comments. I got the test harness running as a part of `cargo test` but it's currently being ignored until we have more confidence everything is in working order. I wonder if we should merge this now so things don't get stale.

lann:
> @lann I've addressed (most) of your comments. I got the test harness running as a part of `cargo test` but it's currently being ignored until we have more confidence everything is in working order. I wonder if we should merge this now so things don't get stale.

Given that its all new code and not enabled I don't think theres any rush to merge.

rylev:
I've done the following:
* Allowed running the runtime test suite as part of `cargo test` or as a stand alone binary
* Feed in paths to the test suite definitions, component repo, and spin binary so that there aren't any assumptions about where these things live
* When run as part of `cargo test` it takes the built executable (instead of taking whatever Spin is on the path). When running the test suite as a stand alone binary we just take whatever spin is on the path. Eventually we should allow customization for the stand alone use case.
* Updated the test so that it is no longer ignored.
* Improved errors

## What's next? 

### 1. Docker 

There's one thing left to do to make this ready for moving all integration tests over: dockerization. In CI we want to ensure that all runtime dependencies are installed. The easiest way for us to do this is with Docker. Luckily, this is much simpler than the status quo since we don't need any dependencies for building Spin components (as all the components are prebuilt). This means we really just need the following:
* Any runtime dependencies such as postgres, redis, etc.
* The `runtime-tests` crate (we may want to prebuild this so that the docker container doesn't need a Rust toolchain)
* The built spin binary we want to test.

Once we have those things we can run the `runtime-tests` crate binary and point it at this Spin binary we've copied over. 

### 2. Out-of-date component rebuilds 

Currently the repository of pre-built components has no way of knowing when a given component has been changed and needs to be rebuilt. For now we rely on a human to know that if the source of a component changes, the component needs to be rebuilt and checked in. This seems fine for now, but we'll want to soon move to some mechanism that detects this and either automatically rebuilds or alerts that the component needs to be rebuilt.

### 3. More tests!

Once everyone is happy that the direction here is correct, we simply need to add more tests. This should be as simple as converting the integration tests into test components in the test-components repo and than adding some tests to the runtime-tests suite.

rylev:
Update re: docker - thinking a bit more about this, I don't really think dockerization is necessary. All we need is to ensure that the runtime dependencies (e.g., postgres, mysql, redis) are running. We currently do this through docker compose - and we can continue to do so. Running the actual tests in docker isn't really necessary.

tests/runtime-tests/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+}
+
+fn make_http_request(spin: &mut Spin) -> Result<String, anyhow::Error> {
+    if let Some(status) = spin.try_wait()? {
+        anyhow::bail!("Spin exited early with status code {:?}", status.code());
+    }
+    log::debug!("Connecting to HTTP server on port {}...", spin.port());
+    let mut connection = std::net::TcpStream::connect(format!("127.0.0.1:{}", spin.port()))
+        .context("could not connect to Spin web server")?;
+    connection.write_all(b"GET / HTTP/1.1\r\nHost: 127.0.0.1:3000\r\n\r\n")?;
```
rajatjindal, line 232: will it always be a GET request? and can we make other request types from the tests e.g. to add regression test for https://github.com/fermyon/spin/issues/2080

tests/runtime-tests/README.md
```diff
@@ -0,0 +1,27 @@
+# Runtime Tests
+
+The runtime tests ensure that Spin can properly run applications. 
+
+For the purposes of these tests, an "application" is a collection of the following things:
+* A Spin compliant WebAssembly binary
+* A spin.toml manifest
+* A list of arguments to be passed to `spin up`
+* A collection of a files needed for running `spin up` (e.g., a migration file for passing to the `--sqlite` argument)
```
rajatjindal, line 9: would we also need plugins such as `js2wasm` (or `py2wasm`) to run tests for JS/TS/Python etc?

tests/runtime-tests/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+The runtime tests can either be run as a library function (e.g., this is how they are run as part of Spin's test suite using `cargo test`) or they can be run stand alone using the `runtime-tests` crate's binary (i.e., running `cargo run` from this directory).
+
+## How do I add a new test?
+
+To add a new test you must add a new folder to the `tests` directory with at least a `spin.toml` manifest. 
+
+The manifest can reference pre-built Spin compliant WebAssembly modules that can be found in the `test-components` folder in the Spin repo. It does so by using the `{{$NAME}}` where `$NAME` is substituted for the name of the test component to be used. For example `{{sqlite}}` will use the test-component named "sqlite" found in the `test-components` directory.
+
+Optionally, an `args` file can be added with newline separated arguments to be passed to the `spin up` invocation, and a `data` directory can be added which will have all of its contents copied into the temporary folder where `spin up` will be run.
```
rajatjindal, line 21: could we somehow change the content of this data dir on the fly? e.g. for KV tests (especially with cloud), we want to ensure that the new value is indeed updated.

tests/runtime.rs
```diff
@@ -0,0 +1,17 @@
+#[cfg(feature = "e2e-tests")]
+#[test]
```
rajatjindal, line 2: does this annotation here means that all tests will report as a "single test" from "Cargo test" perspective? could we somehow force it to report each test as its own?

tests/runtime-tests/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+
+struct Spin {
+    process: std::process::Child,
+    #[allow(dead_code)]
+    stdout: OutputStream,
+    stderr: OutputStream,
+    port: u16,
+}
+
+impl Spin {
```
rajatjindal, line 278: could we change ths to a trait and have a Spin implementation (and then a cloud implementation). that way we can try to reuse same tests for cloud tests.

tests/runtime-tests/src/lib.rs
rajatjindal, line 232: nit: the hardcoded port #

tests/runtime-tests/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+
+#[derive(Debug)]
+enum Response {
+    Ok,
+    Err(String),
+    /// This happens when we panic
+    ErrNoBody,
+}
+
+impl FromStr for Response {
```
rajatjindal, line 194: I am curious why using raw tcp response manipulation instead of using some http library?

tests/runtime-tests/README.md
rajatjindal, line 1: these tests essentially removes the testing/validation of "spin new" and "spin build" (and thus the spin templates). 

is the intention to remove them completely or to cover them somewhere else?

tests/runtime-tests/README.md
rylev: Added clarification in the README - we are not testing that js2wasm produces the correct WebAssembly components. Tests for that should either remain as end to end tests or IMO they should just be tests in the plugin repo themselves.

tests/runtime-tests/README.md
rylev: I've removed this paragraph as support for `data` directory has actually been removed. Each test component is required to check itself that data can be persisted across multiple calls to the host.

tests/runtime.rs
rajatjindal: thanks Ryan.

tests/runtime-tests/README.md
rajatjindal: I think one of the usecase of having data directory is to test scenarios such as static assets are mounted and served as expected. e.g. static fileserver, so this might be useful to have?

tests/runtime-tests/README.md
rajatjindal: this makes sense. thanks Ryan

tests/runtime-tests/README.md
rylev: That's certainly true - we'll add that functionality when we add a test for file system access.

tests/runtime.rs
lann: or https://docs.rs/libtest-mimic/ which the ui-testing crate uses.

test-components/helper/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+macro_rules! ensure_matches {
+    ($expr:expr, $($arg:tt)*) => {
+        if !matches!($expr, $($arg)*) {
+            $crate::return_error!("`{}` did not match `{}`", stringify!($expr), stringify!($($arg)*))
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! assert_eq {
```
lann, line 146: ensure_eq?

test-components/helper/src/lib.rs
```diff
@@ -0,0 +1,166 @@
...
+    ($expr1:expr, $expr2:expr) => {
+        if $expr1 != $expr2 {
+            $crate::return_error!("`{}` != `{}`", stringify!($expr1), stringify!($expr2));
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! return_error {
```
lann, line 155: bail? :slightly_smiling_face:

lann (approved):

tests/runtime-tests/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+### The testing protocol
+
+The test runner will make a GET request against the `/` path. The component should either return a 200 if everything goes well or a 500 if there is an error. If an `error.txt` file is present, the Spin application must return a 500 with the body set to some error message that contains the contents of `error.txt`.
+
+## When do tests pass?
+
+A test will pass in the following conditions:
+* The Spin web server returns a 200
+* The Spin web server returns a 500 with a body that contains the same text inside of the test's error.txt file.
```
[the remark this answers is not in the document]
rylev: Sort of. Essentially, I've not thought of a good way to abstract away from http and be trigger agnostic. I think long term, if we can get it to work, we would have many tests that are trigger agnostic and then some more trigger specific tests. We've talked about adding a test trigger which might be the right path forward here, but I'd like to get this merged first.

PR #2085, opened 2023-11-10
[Backport v2.0] fix non-streaming outgoing POST requests in Rust SDK

lann:
Backporting #2083 to v2.0


lann:
Just getting CI kicked off; should wait for #2085 to actually merge...

PR #2088, opened 2023-11-10
Fix example lock

lann:


lann:
Logical merge conflict between adding the example and bumping the spin version :facepalm:

fibonacci1729 (approved):

Issue #1173, opened 2023-02-17
Offer a multiselect plugin upgrade experience


itowlson:
You can currently upgrade plugins via `spin plugins upgrade foo` or `spin plugins upgrade --all`.

We could also allow `spin plugins upgrade` without the name or flag to launch an interactive experience similar to the templates one, in which it displays a list of plugins and the user ticks which one to upgrade.

lann:
I would vote for plain `spin plugins upgrade` to be equivalent to `--all`.

itowlson:
Heh... that was what I originally did with templates (so to get the multiselect UI you had to do `spin templates upgrade --ask`).  But other people preferred "ask by default and have a flag for `--all`".  I'd be happy to go back if a wider sample of opinion says otherwise.  But regardless, I think we should use the same default in both places.

PR #2091, opened 2023-11-13
Multiselect plugin upgrade

joaogdemacedo:
Trying to add a multi-select plugin upgrade (issue #1173) as my first PR.
I still haven't tested it via CLI yet, because I'm still wondering how I can run the command with the changes I've made. I can't just run 'spin plugins upgrade', right?

Looking forward for your feedback.

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -63,7 +63,6 @@ pub struct Install {
         name = PLUGIN_NAME_OPT,
         conflicts_with = PLUGIN_REMOTE_PLUGIN_MANIFEST_OPT,
         conflicts_with = PLUGIN_LOCAL_PLUGIN_MANIFEST_OPT,
-        required_unless_present_any = [PLUGIN_REMOTE_PLUGIN_MANIFEST_OPT, PLUGIN_LOCAL_PLUGIN_MANIFEST_OPT],
     )]
     pub name: Option<String>,
 
@@ -230,6 +229,8 @@ impl Upgrade {
     /// Upgrades one or all plugins by reinstalling the latest or a specified
     /// version of a plugin. If downgrade is specified, first uninstalls the
     /// plugin.
+    /// Also, by default, Spin displays the list of installed plugins that are in
+    /// the catalogue and prompts user to choose which ones to upgrade.
     pub async fn run(self) -> Result<()> {
         let manager = PluginManager::try_default()?;
         let manifests_dir = manager.store().installed_manifests_directory();
@@ -242,9 +243,69 @@ impl Upgrade {
 
         if self.all {
             self.upgrade_all(manifests_dir).await
+        } else if self.name.is_none()
+            && self.local_manifest_src.is_none()
+            && self.remote_manifest_src.is_none()
+        {
+            // Default behavior
+            self.upgrade_default().await
         } else {
             self.upgrade_one().await
         }
+    }
+
+    async fn upgrade_default(self) -> Result<()> {
+        let catalogue_plugins = list_catalogue_plugins().await?;
+        let installed_plugins = list_installed_plugins()?;
+
+        let installed_in_catalogue: Vec<_> = installed_plugins
+            .into_iter()
+            .filter(|installed| {
+                catalogue_plugins.iter().any(|catalogue| {
+                    installed.manifest == catalogue.manifest
+                        && matches!(catalogue.compatibility, PluginCompatibility::Compatible)
+                })
+            })
+            .collect();
+
+        eprintln!(
+            "Select repos to upgrade. Use Space to select/deselect and Enter to confirm selection."
+        );
+        let selected_indexes = match dialoguer::MultiSelect::new()
+            .items(&installed_in_catalogue)
+            .interact_opt()?
+        {
+            Some(indexes) => indexes,
+            None => return Ok(()),
+        };
+
+        let plugins_selected = elements_at(installed_in_catalogue, selected_indexes);
+
+        if plugins_selected.is_empty() {
+            eprintln!("No plugins selected");
+            return Ok(());
+        }
+
+        // Upgrade plugins selected
+        for plugin in plugins_selected {
+            let manager = PluginManager::try_default()?;
+            let manifest_location = ManifestLocation::PluginsRepository(PluginLookup::new(
+                &plugin.name,
+                Some(plugin.version.parse::<Version>()?),
+            ));
+
+            try_install(
+                &plugin.manifest,
+                &manager,
+                true,  // not sure
+                true,  // not sure
+                false, // not sure
+                &manifest_location,
+            )
+            .await?;
+        }
+
+        Ok(())
     }
 
     // Install the latest of all currently installed plugins
@@ -320,6 +381,46 @@ impl Upgrade {
     }
 }
 
+// Make list_installed_plugins and list_catalogue_plugins into 'free' module-level functions
+// in order to call them in Upgrade::upgrade_default
+fn list_installed_plugins() -> Result<Vec<PluginDescriptor>> {
+    let manager = PluginManager::try_default()?;
+    let store = manager.store();
+    let manifests = store.installed_manifests()?;
+    let descriptors = manifests
+        .into_iter()
+        .map(|m| PluginDescriptor {
+            name: m.name(),
+            version: m.version().to_owned(),
+            installed: true,
+            compatibility: PluginCompatibility::for_current(&m),
+            manifest: m,
+        })
+        .collect();
+    Ok(descriptors)
+}
+
+async fn list_catalogue_plugins() -> Result<Vec<PluginDescriptor>> {
+    if update_silent().await.is_err() {
+        terminal::warn!("Couldn't update plugins registry cache - using most recent");
+    }
+
+    let manager = PluginManager::try_default()?;
+    let store = manager.store();
+    let manifests = store.catalogue_manifests();
+    let descriptors = manifests?
+        .into_iter()
+        .map(|m| PluginDescriptor {
+            name: m.name(),
+            version: m.version().to_owned(),
+            installed: m.is_installed_in(store),
+            compatibility: PluginCompatibility::for_current(&m),
+            manifest: m,
+        })
+        .collect();
+    Ok(descriptors)
+}
+
 /// List available or installed plugins.
 #[derive(Parser, Debug)]
 pub struct List {
@@ -335,7 +436,7 @@ pub struct List {
 impl List {
     pub async fn run(self) -> Result<()> {
         let mut plugins = if self.installed {
-            Self::list_installed_plugins()
+            list_installed_plugins()
         } else {
             Self::list_catalogue_and_installed_plugins().await
         }?;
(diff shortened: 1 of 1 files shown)


src/commands/plugins.rs
```diff
@@ -63,7 +63,6 @@ pub struct Install {
         name = PLUGIN_NAME_OPT,
         conflicts_with = PLUGIN_REMOTE_PLUGIN_MANIFEST_OPT,
         conflicts_with = PLUGIN_LOCAL_PLUGIN_MANIFEST_OPT,
-        required_unless_present_any = [PLUGIN_REMOTE_PLUGIN_MANIFEST_OPT, PLUGIN_LOCAL_PLUGIN_MANIFEST_OPT],
```
itowlson, line 66: Not sure if it's intentional that this PR changes the `install` command?  I think you might have intended to remove this from line 170 instead?

src/commands/plugins.rs
```diff
@@ -242,11 +243,72 @@ impl Upgrade {
...
+                        && matches!(catalogue.compatibility, PluginCompatibility::Compatible)
+                })
+            })
+            .collect();
+
+        eprintln!(
+            "Select plugins to upgrade. Use Space to select/deselect and Enter to confirm selection."
+        );
+        let selected_indexes = match dialoguer::MultiSelect::new()
+            .items(&installed_in_catalogue)
```
itowlson, line 276: This looks like it currently displays _all_ plugins installed from the catalogue, even those without an eligible upgrade.

src/commands/plugins.rs
```diff
@@ -242,11 +243,72 @@ impl Upgrade {
...
+            .into_iter()
+            .filter(|installed| {
+                catalogue_plugins.iter().any(|catalogue| {
+                    installed.manifest == catalogue.manifest
+                        && matches!(catalogue.compatibility, PluginCompatibility::Compatible)
+                })
+            })
+            .collect();
+
+        eprintln!(
```
itowlson, line 272: We need to handle the case where there are no upgrades available - it should print an informative message rather than an empty selector.

src/commands/plugins.rs
```diff
@@ -242,11 +243,72 @@ impl Upgrade {
...
+            eprintln!("No plugins selected");
+            return Ok(());
+        }
+
+        // Upgrade plugins selected
+        for plugin in plugins_selected {
+            let manager = PluginManager::try_default()?;
+            let manifest_location = ManifestLocation::PluginsRepository(PluginLookup::new(
+                &plugin.name,
+                Some(plugin.version.parse::<Version>()?),
```
itowlson, line 295: I am not sure this is quite right.  `installed_in_catalogue` contains the installed plugins _at their installed versions_.  This therefore appears to use the already-installed version as the version to upgrade to.  I may be misreading though?

src/commands/plugins.rs
```diff
@@ -335,7 +437,7 @@ pub struct List {
 impl List {
     pub async fn run(self) -> Result<()> {
         let mut plugins = if self.installed {
-            Self::list_installed_plugins()
+            list_installed_plugins()
         } else {
             Self::list_catalogue_and_installed_plugins().await
```
itowlson, line 442: Although not required for this PR, I would move `list_catalogue_and_installed_plugins` to be a module function too, so that a reader coming to this in 6 months is not puzzled by "why is one of these functions module level and the other one associated?"

src/commands/plugins.rs
```diff
@@ -474,6 +538,26 @@ impl PluginDescriptor {
     }
 }
 
+impl std::fmt::Display for PluginDescriptor {
```
itowlson, line 541: I'd be a little bit cautious about giving this as _the_ display format for a plugin descriptor.  `List::print`, for example, prints them in its own specific format.

My suggestion for the multiselect display would be either:

* Just the plugin name (e.g. `foo`)
* The plugin name with the current version and what it would upgrade to (e.g. `foo (from 1.0.0 to 1.2.0)`)

You can use strings rather than descriptors in the `Multiselect` `items` list as long as you ensure they keep the same order (have the same indexes) so that indexing back into your descriptor list still works e.g.

```
// using the "just the name" example
let names: Vec<_> = descriptors.iter().map(|d| d.name.to_string()).collect();  // preserves order
let indexes = dialoguer::Multiselect::new().items(&names)...;
let selected = elements_at(descriptors, indexes);
```

itowlson (changes requested):
There's a few things, which I think will be obvious when you're able to run the command (I left a message in the Discord thread but am not sure I understood the difficulty - feel free to ping me), but given you've not been able to run it _and_ are new to Rust, this is a heroic effort!  Thanks for this - I'm sure we can land it soon!

src/commands/plugins.rs
joaogdemacedo: You’re right, obviously. Awkward mistake.

src/commands/plugins.rs
joaogdemacedo: Of course. Condition added in line 293.

src/commands/plugins.rs
joaogdemacedo: You're correct. What I tried to do to check which plugins are eligible for an upgrade was to call `get_manifest()` with the latest version (using `manifest_location(None)`) for each plugin. If the function returns successfully, indicating an update is available, I push it to the new `eligible_plugins` vector with the descriptor and the new manifest to have both the old and latest versions.

src/commands/plugins.rs
joaogdemacedo: You are right. Based on my understanding of the code, it seems that to determine the version for an upgrade, I need to provide `'None'` as the version when calling `manifest_location()`. From the tests I've conducted, I believe I have it correct this time.

src/commands/plugins.rs
joaogdemacedo: Hmm, that makes sense!

src/commands/plugins.rs
joaogdemacedo: I combined the two ideas you gave me and create a `names` vector, and using the macro `format!` I appended the versions to the name to display more information to the user as you suggested.

joaogdemacedo:
I've attempted to address all the comments correctly, and I believe we are closer to getting it right. This time, I was able to test it via the CLI. The tests I performed were as follows:

- Running the command with all plugins in their latest version. It resulted in "No plugins found to upgrade".
- After downgrading plugins and running the command again, the plugins were successfully upgraded to their latest version.
- Not selecting any upgrades from the list resulted in the correct message.

src/commands/plugins.rs
```diff
@@ -242,11 +243,99 @@ impl Upgrade {
...
+    async fn upgrade_multiselect(self) -> Result<()> {
+        let catalogue_plugins = list_catalogue_plugins().await?;
+        let installed_plugins = list_installed_plugins()?;
+
+        let installed_in_catalogue: Vec<_> = installed_plugins
+            .into_iter()
+            .filter(|installed| {
+                catalogue_plugins.iter().any(|catalogue| {
+                    installed.manifest == catalogue.manifest
+                        && matches!(catalogue.compatibility, PluginCompatibility::Compatible)
```
itowlson, line 267: I am not sure we need this check.  We don't really care if the catalogue copy of the installed plugin is compatible.  In fact, a possible case is:

* User installs Plugin 1.0 from catalogue.  1.0 is compatible with Spin 3.0 or below.
* User upgrades to Spin 4.1.
* User tries to run Plugin and it warns about compatibility.
* User tries to upgrade.

In this case we would _not_ want to filter Plugin 1.0 out of the list.

We care about compatibility on _upgrade versions_ but I don't think that's what this is looking at.  I might be misunderstanding though?  (As you've rightly noted, unit tests would be the way to catch cases like this - it's hard to simulate with actual plugins!)

src/commands/plugins.rs
```diff
@@ -242,11 +243,99 @@ impl Upgrade {
...
+                .await
+            {
+                if installed_plugin.version != manifest.version() {
+                    eligible_plugins.push((installed_plugin, manifest));
+                }
+            }
+        }
+
+        if eligible_plugins.is_empty() {
+            eprintln!("No plugins found to upgrade");
```
itowlson, line 294: This message would fit the case where the user has no plugins installed.  But by this point the list could be empty because all their plugins are up to date.  My suggestion would be to add a check for "no plugins installed" earlier in the code, and then change this to 
```suggestion
            eprintln!("All plugins are up to date");
```

src/commands/plugins.rs
```diff
@@ -242,11 +243,99 @@ impl Upgrade {
...
+            return Ok(());
+        }
+
+        let names: Vec<_> = eligible_plugins
+            .iter()
+            .map(|(descriptor, manifest)| {
+                format!(
+                    "{} from version {} to {}",
+                    descriptor.name.clone(),
```
itowlson, line 303: nit: I don't _think_ you need to clone these.  (On the other hand, I assume you're doing it because Rust complained if you didn't!  And there's no harm in it for sure.)

src/commands/plugins.rs
```diff
@@ -242,11 +243,99 @@ impl Upgrade {
...
+                &installed_plugin.name,
+                None,
+            ));
+
+            // Attempt to get the manifest to check eligibility to upgrade
+            if let Ok(manifest) = manager
+                .get_manifest(&manifest_location, false, SPIN_VERSION)
+                .await
+            {
+                if installed_plugin.version != manifest.version() {
```
itowlson, line 287: In theory we should check that the catalogue manifest version is _greater than_ the installed version but I don't know if that's ever an issue in practice.

itowlson (changes requested):
Thanks for this!  I left a few more comments but the two I am concerned about are:

* The compatibility check while getting the list of installed plugins that are also in the catalogue (comment on line 267)
* The empty list messaging (command on line 294 but will need a tweak further up the file as well)

Otherwise this looks great.  And thanks for taking the time to describe the tests you did!

src/commands/plugins.rs
joaogdemacedo: Yes, I was doing it because, when I was trying to use the descriptor in the `dialoguer::MultiSelect`, I couldn't move the values. But since I found a simpler way to display the names and versions it isn't needed anymore!

src/commands/plugins.rs
joaogdemacedo: Yeah, you're right, even if this is an issue in practice, in theory, we should check if the version is greater than the installed one.
What we could do is parse the strings to numbers and compare them. What do you think?

itowlson: If we were to do this, we would not parse to numbers but to the `semver::Version` type (see https://docs.rs/semver/latest/semver/).  The `Manifest` type has a `try_version()` method which can give you this  And `semver::Version` allows you to use comparison operators such as `<` and `>` (see e.g. https://github.com/fermyon/spin/blob/1d662e3774f85106de4b212ae48d0d4760f47ee2/crates/plugins/src/badger/mod.rs#L285).

That said, it's up to you whether you want to wrestle with this.  Unfortunately, it is not quite as easy as just calling `try_version()` and doing a comparison, because `try_version` can return an error (if the version is malformed).  Which means we have to think about error handling.  Rust makes it very easy to simply bomb out (via the `?` operator), but that could mean one bad version string broke the whole experience.  One option would be to write a helper function that handled the error cases by falling back to string comparison, along the lines of

```
fn is_potential_upgrade(current: &PluginManifest, candidate: &PluginManifest) -> bool {
  match (current.try_version(), candidate.try_version()) {
    (Ok(cur_ver), Ok(cand_ver)) => cand_ver > cur_ver,
    _ => current.version() != candidate.version(),
  }
}
```

and then this line would be `if is_potential_upgrade(&installed_plugin, &manifest)`.  _(Disclaimer: written in the GitHub comment box, not tested, unlikely to compile as is!)_

It's up to you whether you want to invest this amount of effort - it's not a blocker as far as I'm concerned.

joaogdemacedo: Why `current.version() != candidate.version()` if something goes wrong? I know this is the condition we have at the moment, but shouldn't we return false if something goes wrong with `try_version()`'s?

src/commands/plugins.rs
joaogdemacedo: To check the compatibility of the _upgrade versions_, I believe using the `PluginCompatibility::for_current()` is the best way to do it. Correct me, if I'm wrong.

So, I'll add the condition below after the versions comparison:

`if let PluginCompatibility::Compatible = PluginCompatibility::for_current(&manifest)`

src/commands/plugins.rs
itowlson: Suppose we returned `false` if one of the versions failed to parse.  Now suppose that somehow (again, should never happen) the user's _installed_ version fails to parse.  Now they don't have the option to upgrade.

This could be avoided by having it return `false` if only the _target_ version fails to parse.  But since we're displaying the from and to versions, I'd recommend erring on the side of leaving the choice with the user.

Again, it should be moot because there should _never_ be a malformed version in the catalogue.  But that was the motivation for the more forgiving fallback.

src/commands/plugins.rs
itowlson: That sounds good - thanks!

As a style detail, `if let PluginCompatibility::Compatible = PC::for_current(...)` might be more idiomatic as `matches!(PC::for_current(...), PC::Compatible)` - the `let` is a bit odd because no actual assignment is possible.  You could also add `PartialEq` to the PluginCompatibility `#[derive(...)]` list, which would allow you to write `if PC::Compatible == PC::for_current`.  None of this is a blocker though!

joaogdemacedo: You're totally right. Thanks for the hint!

itowlson (approved):
LGTM!  Thanks for your patience working though this!

PR #2093, opened 2023-11-13
chore(templates/tests): bump spin fileserver component to v0.2.0

vdice:
Ref https://github.com/fermyon/spin-fileserver/releases/tag/v0.2.0

### DIFF
--- a/templates/static-fileserver/content/spin.toml
+++ b/templates/static-fileserver/content/spin.toml
@@ -11,5 +11,5 @@ route = "{{ http-path | http_wildcard }}"
 component = "{{project-name | kebab_case}}"
 
 [component.{{project-name | kebab_case}}]
-source = { url = "https://github.com/fermyon/spin-fileserver/releases/download/v0.1.0/spin_static_fs.wasm", digest = "sha256:96c76d9af86420b39eb6cd7be5550e3cb5d4cc4de572ce0fd1f6a29471536cb4" }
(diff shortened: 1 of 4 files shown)


vdice:
Looks like a legit error from the assets-test using the fileserver component; will take a look.

vdice:
The test failure is around the expected response body when a non-existent file is requested.  Currently, we expect "Not Found" as the resp body, however with the latest v0.2.0 version of the fileserver, we get an empty body.  The resp code is still 404 for both:

### spin_static_fs.wasm v01.0
```
$ curl -i http://127.0.0.1:3000/static/thisshouldbeexcluded/4
HTTP/1.1 404 Not Found
content-length: 9
date: Tue, 14 Nov 2023 17:42:50 GMT

Not Found
```

### spin_static_fs.wasm v0.2.0
```
$ curl -i http://127.0.0.1:3000/static/thisshouldbeexcluded/4
HTTP/1.1 404 Not Found
cache-control: max-age=60
etag: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
transfer-encoding: chunked
date: Tue, 14 Nov 2023 17:39:58 GMT

```

I'm assuming this is due to v0.2.0 using different http libs per wasi-http?  Do we want to add logic to return a "Not Found" body or is this an acceptable change?  cc @dicej

dicej:
Sorry, I didn't see this earlier.  We can definitely add response bodies in this case -- probably a good idea from a user-friendliness perspective.

vdice:
Thanks @dicej.  Will close this PR and re-open with a forthcoming 0.2.1 release.

PR #2096, opened 2023-11-14
Watch should not serve until there is a build

### LINKED ISSUE #2061: `spin watch` shouldn't start serving an application if the initial `spin build` fails
Executing `spin watch` causes an initial invocation of the `build` command. If that command fails, Spin should abort instead of loading and running the application.

itowlson:
Fixes #2061


calebschoepp (approved):
LGTM!

PR #2101, opened 2023-11-16
Statically linked spin

rajatjindal:
For openshift integration we need statically linked binary of spin. This PR adds the required changes for that.

I think we also need to attach these artifacts to releases, I will look at it next (will add commits to this PR)

cc @0xE282B0

### DIFF
--- a/.github/actions/spin-ci-dependencies/action.yml
+++ b/.github/actions/spin-ci-dependencies/action.yml
@@ -17,6 +17,11 @@ inputs:
     required: false
     default: 'false'
     type: bool
+  rust-cross:
+    description: 'setup rust cross'
+    required: false
+    default: 'false'
+    type: bool
   rust-cache:
     description: 'setup rust cache'
     required: false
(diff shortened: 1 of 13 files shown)


lann (approved):

tschneidereit:
While I wouldn't necessarily expect there to be problems, we should verify that these builds don't meaningfully regress performance. Musl isn't the fastest libc out there, and in particular its malloc implementation isn't fast, so that might result in unexpected regressions if we have hot code somewhere in Spin where that matters. (And if all of this isn't mitigated by Rust's Musl target anyway.)

rajatjindal:
Hi @tschneidereit, thank you for your feedback. Do you happen to have any suggestions about what could be the alternative here? I am happy to explore other options here.

rajatjindal:
also, just to ensure we are on same page, this PR adds additional release assets for statically linked Spin binaries (and does not replace existing releases binaries).

tschneidereit:
Oh sorry, I should've been clearer: I agree that having these builds makes sense! I also don't really have an alternative as such.

That doesn't mean that we shouldn't test performance of this configuration vs others though. If we were to discover meaningful regressions, there might be things we can do to address those without moving away from this target. For example, we could look into using a different allocator, or we could potentially change things in Spin to not be as affected by musl's lower performance.

PR #2116, opened 2023-11-27
Always url encode authority when parsing OutboundUrl

### LINKED ISSUE #2113: outbound-networking does not work as expected for MySQL
## Context

I created a Spin application that interacts with an MySQL database. In my case, the database is hosted in Azure (but this does not matter at all). I recognized outbound connectivity being blocked when the MySQL password contains a `#` (hash sign)! 

In my case, `allowed_outbound_hosts` in `Spin.toml` look like this: 

```toml
allowed_outbound_hosts = ["mysql://mydatabase.mysql.database.azure.com"]
``` 

I use the `Connection` struct from `spin_sdk::mysql` to establish the connection as shown below:

```rust
let c = Connection::open("mysql://thorsten:pass#word001@mydatabase.mysql.database.azure.com/testdb?ssl-mode")?;
```

At runtime, I get the following error:

```bash
Handler returned an error: Error::ConnectionFailed("address mysql://thorsten:pass#word001@mydatabase.mysql.database.azure.com/testdb?ssl-mode=required is not permitted")
```

On top of permitting the outgoing connection, my password is not shown in `stdout`. 

## Root Cause

I checked the source to spot the root cause, which is ultimately here:

https://github.com/fermyon/spin/blob/9e7a68d17f82b51239c4217b57948d8df98db3a0/crates/outbound-networking/src/lib.rs#L302C9-L316C12

My guess is that the `url` crate, considers the `#` from the password introducing a URL hash-fragment. That being said, I think this may also break for outgoing connections with credentials in the connection string and other schemes than `mysql`. 

The behavior could be reproduced by adding the following unit test:

```rust
#[test]
    fn test_hash_char_in_db_password() {
        let allowed = AllowedHostsConfig::parse(&["mysql://xyz.com"]).unwrap();
        assert!(
            allowed.allows(&OutboundUrl::parse("mysql://user:pass#word@xyz.com", "mysql").unwrap())
        );
    }
```

## Environment

os: `Darwin mbp14.local 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:24 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6000 arm6`
spin: `spin 2.0.1 (1d72f1c 2023-11-10)`
spin-sdk: `{ git = "https://github.com/fermyon/spin", tag = "v2.0.0" }`

rylev:
Fixes #2113 

This prevents errors when users don't url encode the authority where the authority accidentally gets viewed as the url itself.

For example, in `mysql://user:pass#word@xyz.com` the `user` will be parsed as the host before this change. With this change the authority is url encoded and correctly parsed as the authority.

### DIFF
--- a/crates/outbound-networking/src/lib.rs
+++ b/crates/outbound-networking/src/lib.rs
@@ -299,7 +299,21 @@ pub struct OutboundUrl {
 
 impl OutboundUrl {
     pub fn parse(url: impl Into<String>, scheme: &str) -> anyhow::Result<Self> {
-        let url = url.into();
+        let mut url = url.into();
+        let original = url.clone();
+
+        // Ensure that the authority is url encoded. Since the authority is ignored after this,
+        // we can always url encode the authority even if it is already encoded.
+        if let Some(at) = url.find('@') {
+            let scheme_end = url.find("://").map(|e| e + 3).unwrap_or(0);
+            let userinfo = &url[scheme_end..at];
+
+            let encoded = urlencoding::encode(userinfo);
+            let prefix = &url[..scheme_end];
+            let suffix = &url[scheme_end + userinfo.len()..];
+            url = format!("{prefix}{encoded}{suffix}");
+        }
+
         let parsed = match url::Url::parse(&url) {
             Ok(url) if url.has_host() => Ok(url),
             first_try => {
@@ -322,7 +336,7 @@ impl OutboundUrl {
                 .with_context(|| format!("{url:?} does not have a host component"))?
                 .to_owned(),
             port: parsed.port(),
-            original: url,
+            original,
         })
     }
 }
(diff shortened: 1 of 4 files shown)


crates/outbound-networking/src/lib.rs
```diff
@@ -299,7 +299,20 @@ pub struct OutboundUrl {
...
+        // Ensure that the authority is url encoded. Since the authority is ignored after this,
+        // we can always url encode the authority even if it is already encoded.
+        if let Some(at) = url.find('@') {
+            let scheme_end = url.find("://").map(|e| e + 3).unwrap_or(0);
+            let userinfo = &url[scheme_end..at];
+
+            let encoded = urlencoding::encode(userinfo);
+            let prefix = &url[..scheme_end];
+            let suffix = &url[scheme_end + userinfo.len()..];
+            url = format!("{prefix}{encoded}{suffix}");
```
itowlson, line 313: This results in `OutboundUrl::original` being the modified string rather than the _original_ original.  As far as I can tell that's only used for the `Display` implementation so I'm not sure if this is a concern - it might be surprising if displayed in a message and the user went "but that's not what I wrote"?

itowlson (commented):
Functionally looks fine, just a question about whether it's okay to use the modified form in the Display implementation.

PR #2118, opened 2023-11-27
chore(docs): manifest v2; redirect module to v0.1.0

vdice:
* Update the manifest to v2
* Update the redirect module to [v0.1.0](https://github.com/fermyon/spin-redirect/releases/tag/v0.1.0)

- [x] Depends on https://github.com/fermyon/spin/pull/2129


docs/spin.toml
```diff
@@ -1,198 +1,250 @@
...
+component = "redirect-contributing"
+route = "/contributing"
+
+[component.redirect-download]
 source = "modules/redirect.wasm"
 environment = { DESTINATION = "https://developer.fermyon.com/downloads/install.sh" }
-[component.trigger]
-route = "/downloads/install.sh"
-executor = { type = "wagi" }
+allowed_outbound_hosts = ["https://developer.fermyon.com"]
```
itowlson, line 135: Are these needed?  They don't seem to be there in the v1 components.  I would expect us just to issue a 3xx to the client, not to need to actually do a request.

vdice: Oh you are right - thank you.  Not needed; removed.

itowlson (approved):

PR #2122, opened 2023-11-29
Add async support to the rust-sdk router

### LINKED ISSUE #2114: Rust SDK: Support async functions in HTTP router
Now that `http_component` functions can be `async` several developers in discord have been asking about using the `HTTP` router with async functions.

fibonacci1729:
This PR adds support for `async` functions/closures to be registered with the rust-sdk router. I update the `http-rust-router` example to demonstrate how to register `async` handlers using the suite of `*_async` registration methods on the router. Additionally the example demonstrates that you can register synchronous handlers along side asynchronous handlers.

A notable omission here is the lack of an `async` equivalent to the `http_router!` macro. Open for feedback on whether folks think this is a worthwhile pursuit. Ideally one could hint in the macro which registration method to use (e.g. `get` or `get_async`), for example:

```
async fn hello_planet(_req: Request, params: Params) -> anyhow::Result<impl IntoResponse> {
        let planet = params.get("planet").expect("PLANET");
        Ok(Response::new(200, planet.to_string()))
}

let router = http_router! {
        @async GET "/hello/:planet" => api::hello_planet, // --> calls router.get_async(...)
        _   "/*" => |_req: Request, params| {
            let capture = params.wildcard().unwrap_or_default();
            Response::new(200, capture.to_string())
        }
};
```

Closes #2114

### DIFF
--- a/examples/http-rust-router/src/lib.rs
+++ b/examples/http-rust-router/src/lib.rs
@@ -8,23 +8,29 @@ use spin_sdk::{
 #[http_component]
 fn handle_route(req: Request) -> Response {
     let mut router = Router::new();
-    router.get("/hello/:planet", api::hello_planet);
-    router.any("/*", api::echo_wildcard);
+    router.get("/goodbye/:planet", api::goodbye_planet);
+    router.get_async("/hello/:planet", api::hello_planet);
+    router.any_async("/*", api::echo_wildcard);
     router.handle(req)
 }
 
 mod api {
     use super::*;
 
-    // /hello/:planet
-    pub fn hello_planet(_req: Request, params: Params) -> Result<impl IntoResponse> {
+    // /goodbye/:planet
+    pub fn goodbye_planet(_req: Request, params: Params) -> Result<impl IntoResponse> {
         let planet = params.get("planet").expect("PLANET");
+        Ok(Response::new(200, planet.to_string()))
+    }
 
+    // /hello/:planet
+    pub async fn hello_planet(_req: Request, params: Params) -> Result<impl IntoResponse> {
+        let planet = params.get("planet").expect("PLANET");
         Ok(Response::new(200, planet.to_string()))
     }
 
     // /*
-    pub fn echo_wildcard(_req: Request, params: Params) -> Result<impl IntoResponse> {
+    pub async fn echo_wildcard(_req: Request, params: Params) -> Result<impl IntoResponse> {
         let capture = params.wildcard().unwrap_or_default();
         Ok(Response::new(200, capture.to_string()))
     }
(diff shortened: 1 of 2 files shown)


examples/http-rust-router/src/lib.rs
```diff
@@ -17,15 +18,17 @@ mod api {
...
-
         Ok(Response::new(200, planet.to_string()))
     }
 
     // /*
-    pub fn echo_wildcard(_req: Request, params: Params) -> Result<impl IntoResponse> {
+    pub fn echo_wildcard(
+        _req: Request,
+        params: Params,
+    ) -> impl Future<Output = Result<impl IntoResponse>> {
```
rylev, line 30: I would make this an `async` function instead of having to do `impl Future`

rylev (commented):
I don't think we can merge the PR as it currently stands as this is a pretty significant breaking change. 

We should try out how much of maintanece nightmare it will be to double the size of the API with either `async_` functions or a separate `AsyncRouter` type. My personal preference is for a separate type, but this isn't based on any concrete reasoning.

examples/http-rust-router/src/lib.rs
fibonacci1729: Yep! Using the `async` keyword here defin8litely has better ergonomics.

sdk/rust/src/http/router.rs
```diff
@@ -1,17 +1,51 @@
...
 use routefinder::{Captures, Router as MethodRouter};
+use std::future::Future;
 use std::{collections::HashMap, fmt::Display};
 
-type Handler = dyn Fn(Request, Params) -> Response;
+type DynHandler = dyn Handler;
+
+/// An HTTP request handler.
+///
+///
```
rylev, line 14:
```suggestion
```

sdk/rust/src/http/router.rs
```diff
@@ -1,17 +1,51 @@
...
+
 use super::conversions::{IntoResponse, TryFromRequest, TryIntoRequest};
 use super::{responses, Method, Request, Response};
+use async_trait::async_trait;
 use routefinder::{Captures, Router as MethodRouter};
+use std::future::Future;
 use std::{collections::HashMap, fmt::Display};
 
-type Handler = dyn Fn(Request, Params) -> Response;
+type DynHandler = dyn Handler;
```
rylev, line 10: I'd prefer not to have type aliases unless they are obvious and make things much easier to read which I don't think is the case here.

rylev (commented):
Looks good to me - I'm still not sure how I feel about duplicating all the methods with `async` variants vs having a different `AsyncRouter` type but I guess the advantage here is that it allows mix/matched routing... Would love to hear what others think.

PR #2123, opened 2023-11-29
allow multiple components to be executed on same trigger

### LINKED ISSUE #1920: redis-trigger: Ony last component gets redis pub message
I've noticed that if I have two components listening to the same channel, only the later one receives published messages, while the first one doesn't receive any.

This is my spin.toml.
```
spin_manifest_version = "1"
authors = ["Seungjin Kim>"]
description = "foo"
name = "foo"
trigger = { type = "redis", address = "redis://foo.com:19539" }
version = "0.1.0"

[[component]]
id = "one"
source = "one/target/wasm32-wasi/release/one.wasm"
allowed_http_hosts = []
[component.trigger]
channel = "test"
[component.build]
workdir = "one"

[[component]]
id = "two"
source = "two/target/wasm32-wasi/release/two.wasm"
allowed_http_hosts = []
[component.trigger]
channel = "test"
[component.build]
workdir = "two"
```
The spin.toml file is set up very simply, and each component includes code at the example level, with both having identical code. However, when my Redis publishes a message, only 'component two' receives the message, while 'component one' receives nothing

karthik2804:
Updates the Redis Trigger implementation to allow multiple components to receive events from the same channel.

Currently - with this PR, there is no way to associate specific errors to component ID (only the generic "`handle-message` returned an error"). Are there any better ways to surface the error? 

fixes #1920

### DIFF
--- a/crates/redis/src/lib.rs
+++ b/crates/redis/src/lib.rs
@@ -5,7 +5,7 @@ mod spin;
 use std::collections::HashMap;
 
 use anyhow::{anyhow, Context, Result};
-use futures::StreamExt;
+use futures::{future::join_all, StreamExt};
 use redis::{Client, ConnectionLike};
 use serde::{de::IgnoredAny, Deserialize, Serialize};
 use spin_app::MetadataKey;
@@ -25,7 +25,7 @@ pub struct RedisTrigger {
     // Redis address to connect to
     address: String,
     // Mapping of subscription channels to component IDs
-    channel_components: HashMap<String, String>,
+    channel_components: HashMap<String, Vec<String>>,
 }
 
 /// Redis trigger configuration.
@@ -58,11 +58,14 @@ impl TriggerExecutor for RedisTrigger {
     async fn new(engine: TriggerAppEngine<Self>) -> Result<Self> {
         let address = engine.app().require_metadata(TRIGGER_METADATA_KEY)?.address;
 
-        let channel_components = engine
-            .trigger_configs()
-            .map(|(_, config)| (config.channel.clone(), config.component.clone()))
-            .collect();
+        let mut channel_components: HashMap<String, Vec<String>> = HashMap::new();
 
+        for (_, config) in engine.trigger_configs() {
+            channel_components
+                .entry(config.channel.clone())
+                .or_default()
+                .push(config.component.clone());
+        }
         Ok(Self {
             engine,
             address,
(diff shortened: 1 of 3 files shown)


crates/redis/src/lib.rs
```diff
@@ -107,15 +110,29 @@ impl TriggerExecutor for RedisTrigger {
...
-                .execute(&self.engine, component_id, channel, msg.get_payload_bytes())
-                .await?
+        if let Some(component_ids) = self.channel_components.get(&channel) {
+            let futures: Vec<_> = component_ids
+                .iter()
+                .map(|id| {
+                    tracing::trace!("Executing Redis component {id:?}");
+                    SpinRedisExecutor.execute(&self.engine, id, &channel, msg.get_payload_bytes())
+                })
+                .collect();
```
rylev, line 123: I think turning this into a `Vec` isn't necessary as `join_all` works on `intoIterator`.

crates/redis/src/lib.rs
```diff
@@ -107,15 +110,29 @@ impl TriggerExecutor for RedisTrigger {
 impl RedisTrigger {
     // Handle the message.
     async fn handle(&self, msg: redis::Msg) -> Result<()> {
-        let channel = msg.get_channel_name();
+        let channel = msg.get_channel_name().to_owned();
```
rylev, line 113: Why does this need to change?

crates/redis/src/lib.rs
```diff
@@ -107,15 +110,29 @@ impl TriggerExecutor for RedisTrigger {
...
+                .collect();
+            let results: Vec<_> = join_all(futures).await.into_iter().collect();
+            let mut errors = Vec::new();
+            for result in results {
+                if result.is_err() {
+                    if let Some(err) = result.err() {
+                        errors.push(err)
+                    }
+                }
+            }
```
rylev, line 132: This is a bit more terse and IMO readable:

```suggestion
            let errors = results.filter_map(|r| r.err()).collect::<Vec<_>>():
           
```

Or you could just return the first error:

```suggestion
           results.collect::<Result<Vec<_>>()?:
```

rylev (approved):
Looks good!

crates/redis/src/lib.rs
karthik2804: Which would be the preferred approach here?

lann: The first option is probably a little better but it doesn't matter much because errors are only checked in tests.

Speaking of which, we should probably log errors here :sweat_smile:

karthik2804: Okay I went with the first option.

PR #2126, opened 2023-11-29
Remove the sqlite e2e tests as the runtime tests make them redundant

rylev:
With #2062 the sqlite e2e tests were made redundant. This PR removes those tests.


rajatjindal (approved):
🎉

PR #2132, opened 2023-11-30
Fix duplicate package warning from Rust SDK

### LINKED ISSUE #2131: The duplicate packages are back, Ted
Actual behaviour referencing Spin SDK on `main`:

```
$ spin build
Building component danglybobs with `cargo build --target wasm32-wasi --release`
    Updating crates.io index
    Updating git repository `https://github.com/fermyon/spin`
warning: skipping duplicate package `sqlite` found at `/home/ivan/.cargo/git/checkouts/spin-91500438ac5656d2/cf8c427/tests/testcases/sqlite`
```

Expected behaviour:

No warning.

This usually means an example or test case got copied and we forgot to change the name.  It might be good to add a test for this given it happens every few months.  _grin_

itowlson:
Fixes #2131

### DIFF
--- a/test-components/sqlite/Cargo.toml
+++ b/test-components/sqlite/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "sqlite"
+name = "sqlite-test-component"
 version = "0.1.0"
 edition = "2021"
 


test-components/sqlite/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "sqlite"
+name = "sqlite-test-component"
```
rylev, line 2: I'm confused - these test-component crates are not in the same workspace as everything else. Why would this cause an issue?

itowlson:
@rylev It's a quirk (or arguably a bug) of Cargo git references - it apparently scans the _entire_ repo for Cargo manifests - I agree it _shouldn't_ cause an issue but it does!  Searching for the original issue I found some stuff implying they had fixed it for Cargo itself, so it may be there's something we can opt into re "published" vs "unpublished" packages?

lann:
> apparently scans the entire repo for Cargo manifests

I was recently wondering how it resolves `{ git = "...", package = "not-the-root" }` which this would explain!

PR #2133, opened 2023-11-30
fix(trigger): don't panic if a component is not attached to a trigger

### LINKED ISSUE #2130: One trigger and multiple components don't work
* Spin version: `spin 2.1.0-pre0 (87b28483 2023-11-29)`

Given this `spin.toml`, it seems like `spin up` doesn't work:
```toml
spin_manifest_version = 2

[application]
name = "hello-wasi-http"
version = "1.0.0"

[[trigger.http]]
route = "/"
component = "hello"

[component.hello]
source = "main_2023_10_18.component.wasm"
[component.hello.build]
command = """go generate && 
    tinygo build -o main_2023_10_18.wasm -target=wasi main_2023_10_18.go && 
    wasm-tools component embed wit/2023_10_18 main_2023_10_18.wasm > main_2023_10_18.embed.wasm && 
    wasm-tools component new main_2023_10_18.embed.wasm -o main_2023_10_18.component.wasm --adapt wasi_snapshot_preview1.reactor.2023_10_18.wasm
"""

[component.hello2]
source = "main_2023_11_10.component.wasm"
[component.hello2.build]
command = """go generate && 
    tinygo build -o main_2023_11_10.wasm -target=wasi main_2023_11_10.go && 
    wasm-tools component embed wit/2023_11_10 main_2023_11_10.wasm > main_2023_11_10.embed.wasm && 
    wasm-tools component new main_2023_11_10.embed.wasm -o main_2023_11_10.component.wasm --adapt wasi_snapshot_preview1.reactor.2023_11_10.wasm
"""
```

The panic message is 
```bash
thread 'main' panicked at /Users/mossaka/.cargo/git/checkouts/spin-91500438ac5656d2/87b2848/crates/trigger/src/lib.rs:248:88:
called `Option::unwrap()` on a `None` value
```

I think the issue is that the `spin.toml` defines multiple components - "hello" and "hello2" in this case, but only one http trigger for "hello". This messed up the trigger hashmap that looks for all the component IDs. 

I've verified this is a valid spin.toml file by running `spin doctor`

```bash
spin doctor   
📟 The Spin Doctor is in.
🩺 Checking spin.toml...
```

## Expectation

Spin runtime shouldn't panic given a valid spin.toml file.

Mossaka:
Instead of instantiate trigger for every component, we should only instantiate trigger that has attached to a component. This fixes #2130

### DIFF
--- a/crates/trigger/src/lib.rs
+++ b/crates/trigger/src/lib.rs
@@ -250,6 +250,12 @@ impl<Executor: TriggerExecutor> TriggerAppEngine<Executor> {
                         .await
                         .with_context(|| format!("Failed to instantiate component '{id}'"))?,
                 );
+            } else {
+                tracing::warn!(
+                    "component '{id}' is not used by any triggers in app '{app_name}'",
+                    id = id,
+                    app_name = app_name
+                )
             }
         }
 
--- a/crates/trigger/src/lib.rs
+++ b/crates/trigger/src/lib.rs
@@ -243,12 +243,14 @@ impl<Executor: TriggerExecutor> TriggerAppEngine<Executor> {
         let mut component_instance_pres = HashMap::default();
         for component in app.borrowed().components() {
             let id = component.id();
-            component_instance_pres.insert(
-                id.to_owned(),
-                Executor::instantiate_pre(&engine, &component, trigger_configs.get(id).unwrap())
-                    .await
-                    .with_context(|| format!("Failed to instantiate component '{id}'"))?,
-            );
+            if let Some(config) = trigger_configs.get(id) {
+                component_instance_pres.insert(
+                    id.to_owned(),
+                    Executor::instantiate_pre(&engine, &component, config)
+                        .await
+                        .with_context(|| format!("Failed to instantiate component '{id}'"))?,
+                );
+            }
         }
 
         Ok(Self {


crates/trigger/src/lib.rs
```diff
@@ -243,12 +243,14 @@ impl<Executor: TriggerExecutor> TriggerAppEngine<Executor> {
         let mut component_instance_pres = HashMap::default();
         for component in app.borrowed().components() {
             let id = component.id();
-            component_instance_pres.insert(
-                id.to_owned(),
-                Executor::instantiate_pre(&engine, &component, trigger_configs.get(id).unwrap())
-                    .await
-                    .with_context(|| format!("Failed to instantiate component '{id}'"))?,
-            );
+            if let Some(config) = trigger_configs.get(id) {
```
itowlson, line 246: It might be good to log a warning if the ID is _not_ found.  At the moment, dangling components are not an expected thing (although this may change with @fibonacci1729's composition work, I'm not when things get composed and what ends up in the lockfile), and I'd be inclined to give the user some way to find out that a surprising thing has happened.

itowlson (approved):
LGTM although I'd be keen to see a log message

crates/trigger/src/lib.rs
Mossaka: Sure

crates/trigger/src/lib.rs
```diff
@@ -243,12 +243,20 @@ impl<Executor: TriggerExecutor> TriggerAppEngine<Executor> {
...
+                    id.to_owned(),
+                    Executor::instantiate_pre(&engine, &component, config)
+                        .await
+                        .with_context(|| format!("Failed to instantiate component '{id}'"))?,
+                );
+            } else {
+                tracing::warn!(
+                    "component '{id}' is not used by any triggers in app '{app_name}'",
+                    id = id,
+                    app_name = app_name
```
rylev, line 257: Nit: this can be removed.

Issue #2128, opened 2023-11-29
add example of how to use spin registry push in CLI help text


michelleN:
Example:
```
spin registry push michellenoorali/spin-hello-world:0.1.0
```

It took me a while to figure out what reference should be. An example in the CLI help text would have made things a little clearer.

ogghead:
This should now be addressed by https://github.com/fermyon/spin/pull/2135 😄

michelleN:
Thank you @ogghead !

PR #2135, opened 2023-12-03
Added examples to `spin registry` CLI commands, added Windows override to Makefile

ogghead:
# Overview
This addresses https://github.com/fermyon/spin/issues/2128. This is my first PR to Spin, please let me know if there are any conventions I have missed!

## Changes
* Added examples to `spin registry` commands
* Added Windows override to Makefile to avoid trying to set RUST_LOG environment variables

## Testing
* Ran all tests with `make test`, verified the only failures were expected due to Windows versioning/ file system differences (several spin doctor, spin-loader, and spin-plugins tests failed locally, but should succeed in build fleet)

### DIFF
--- a/src/commands/registry.rs
+++ b/src/commands/registry.rs
@@ -53,7 +53,8 @@ pub struct Push {
     #[clap(long, takes_value = false, env = ALWAYS_BUILD_ENV)]
     pub build: bool,
 
-    /// Reference of the Spin application
+    /// Reference in the registry of the Spin application.
+    /// This is a string whose format is defined by the registry standard, and generally consists of <registry>/<username>/<application-name>:<version>. E.g. ghcr.io/ogghead/spin-test-app:0.1.0
     #[clap()]
     pub reference: String,
 }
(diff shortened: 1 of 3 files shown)


src/commands/registry.rs
```diff
@@ -126,6 +134,9 @@ pub struct Login {
     )]
     pub password_stdin: bool,
 
+    /// Container registry server
```
itowlson, line 137: We (again generally and inconsistently) tend to avoid "container" in this context, because Spin OCI images are not containers and we don't want to risk ambiguity on that.  If we qualify "registry" then we prefer "OCI" as in the specification (e.g. in the `spin registry` root subcommand help text).  This is all a very long explanation of why I suggest 

```suggestion
    /// OCI registry server (e.g. ghcr.io)
```

(again with the merging paragraphs)

itowlson (changes requested):
Thanks for this and welcome to the project!  Despite the 'request changes', it looks good!  The suggestions are just some minor wordsmithing - absolutely open to alternatives if you dislike them of course!

And thanks for persevering with the Windows side of things - most of the development happens on Mac and Linux and sometimes it shows.  We really appreciate folks helping keep us honest on the Windows side!

itowlson (approved):
Looks good to me - thank you very much!

itowlson:
The e2e test failure looks like we might have changed an error message or something?  I can't see it being related to this PR.

PR #2136, opened 2023-12-04
Improve error message for non-allowed host in outbound-http handler

rylev:
Closes https://github.com/fermyon/spin/issues/2044 and https://github.com/fermyon/spin/pull/2045

### DIFF
--- a/crates/outbound-http/src/host_impl.rs
+++ b/crates/outbound-http/src/host_impl.rs
@@ -46,9 +46,9 @@ impl outbound_http::Host for OutboundHttp {
                 .map_err(|_| HttpError::RuntimeError)?
             {
                 tracing::log::info!("Destination not allowed: {}", req.uri);
-                if let Some(host) = host(&req.uri) {
-                    terminal::warn!("A component tried to make a HTTP request to non-allowed host '{host}'.");
-                    eprintln!("To allow requests, add 'allowed_http_hosts = [\"{host}\"]' to the manifest component section.");
+                if let Some(host_and_port) = host_and_port(&req.uri) {
+                    terminal::warn!("A component tried to make a HTTP request to non-allowed host '{host_and_port}'.");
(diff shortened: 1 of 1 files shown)


crates/outbound-http/src/host_impl.rs
```diff
@@ -46,9 +46,9 @@ impl outbound_http::Host for OutboundHttp {
                 .map_err(|_| HttpError::RuntimeError)?
             {
                 tracing::log::info!("Destination not allowed: {}", req.uri);
-                if let Some(host) = host(&req.uri) {
-                    terminal::warn!("A component tried to make a HTTP request to non-allowed host '{host}'.");
-                    eprintln!("To allow requests, add 'allowed_http_hosts = [\"{host}\"]' to the manifest component section.");
+                if let Some(host_and_port) = host_and_port(&req.uri) {
+                    terminal::warn!("A component tried to make a HTTP request to non-allowed host '{host_and_port}'.");
+                    eprintln!("To allow requests, add 'allowed_outbound_hosts = [\"https://{host_and_port}\"]' to the manifest component section.");
```
vdice, line 51: It looks like `https://` is hardcoded here.  What about cases where the non-allowed URL uses `http://`?

itowlson: We have some similar code for generating error text at https://github.com/fermyon/spin/blob/b98eba25edeecb37385fb77e906ddc04071f2176/crates/outbound-networking/src/lib.rs#L22  Could/should we unify some or all of the code?

rylev: There's definitely room for us to share code here, but it will require a bit of refactoring. Unfortunately, http is a bit of a special snowflake since it allows for things like relative requests.

vdice (approved):

itowlson (approved):

PR #2150, opened 2023-12-07
Outbound postgres runtime tests

rylev:
New outbound postgres runtime tests that replace the now deleted end to end tests.


test-components/outbound-postgres/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        let rowset = ensure_ok!(test_character_types(&conn));
+        ensure!(rowset.rows.iter().all(|r| r.len() == 3));
+        ensure!(matches!(rowset.rows[0][0], rdbms_types::DbValue::Str(ref s) if s == "rvarchar"));
+
+        let pid1 = format!("{:?}", ensure_ok!(pg_backend_pid(&conn)));
+        let pid2 = format!("{:?}", ensure_ok!(pg_backend_pid(&conn)));
+        ensure_eq!(pid1, pid2);
+
+        Ok(())
```
itowlson, line 33: It looks like there are no tests around NULL handling.  We have had bugs around this in the past - it would be good to cover in the runtime tests, as it's hard to unit test robustly.

itowlson (approved):
Looks good - I suggested one thing I'd like to add, but it wasn't in the existing tests so it's not a regression or anything - could easily be added as a separate enhancement.

PR #2152, opened 2023-12-08
Make TimerTrigger example importable in other projects

0xE282B0:
### Description
The `spin-timer` example shows how to create a simple trigger for spin applications. The example is a binary crate that can be used by `spin up` to run a locked application. The goal of this PR is to make the `TimerTrigger` importable in other projects, so that this trigger can be used not only as a standalone binary, but also when embedding spin. So that plugin developers can use this crate as a blueprint.

This PR has two commits:

1. The first commit moves everything but the main function to a `lib.rs`
~~2. The second commit replaces the manual creation of the plugin manifest with pluginify~~.

I'd love to hear your feedback 😊


radu-matei:
cc @karthik2804 and @itowlson 

I know @karthik2804 has been working on a dedicated cron trigger — https://github.com/fermyon/spin-trigger-cron, which we want to support.
(this is compared with the timer trigger from the Spin repo which has historically been an example of how to create one and embed Spin).

0xE282B0:
Awesome, I wasn't aware of the spin-trigger-cron repo! 😅

itowlson (approved):
I'm delighted to see this - it's been in the back of my mind and I really appreciate you dealing to it!

I'm a bit concerned about using the `pluginify` plugin in the context of this example - or specifically about it being the _only_ way to run it.  `pluginify` is kind of magical; and as this is an example, and its job is to build understanding, it would be good to also show the manual way.  But that's not a blocker and others may have different points of view!

itowlson:
@radu-matei For sure.  The intent with the timer trigger, as you note, is educational, to provide a model for trigger devs to follow.  What @0xE282B0 does here is absolutely what we want trigger devs to do, and should be included.

0xE282B0:
>I'm a bit concerned about using the pluginify plugin in the context of this example - or specifically about it being the only way to run it. pluginify is kind of magical; and as this is an example, and its job is to build understanding, it would be good to also show the manual way. But that's not a blocker and others may have different points of view!

@itowlson, maybe I was so excited about how easy pluginify is that I forgot the value of learning the manual way. How about a short setup guide in the readme?

itowlson:
@0xE282B0 sounds good!

examples/spin-timer/readme.md
```diff
@@ -12,3 +12,22 @@ To test:
...
+## Pluginify
+
+To test:
+* create a `spin-pluginify.toml` file as follows:
+```toml
+name= "trigger-timer"
+description= "IT IS I THE GREAT OZ"
+homepage= "www.example.com"
+version= "0.1.0"
+spin_compatibility= ">=0.7"
```
itowlson, line 25: This probably ought to be `>=2.0` at this point though it doesn't much matter for the local dev case

examples/spin-timer/readme.md
```diff
@@ -12,3 +12,22 @@ To test:
...
 
 Then you should be able to `spin build --up` the [guest](./app-example/).
+
+## Pluginify
+
+To test:
+* create a `spin-pluginify.toml` file as follows:
+```toml
+name= "trigger-timer"
+description= "IT IS I THE GREAT OZ"
```
itowlson, line 22: maybe we should change this description
```suggestion
description= "Run Spin components at timed intervals"
```

itowlson (changes requested):
I am sorry to do this (and I know the bad description was mine) but now that it's in the read-me we really have got to change the description!

examples/spin-timer/readme.md
0xE282B0: I changed the description, spin_compatibility, and homepage in both files `REAME.md` and `trigger-timer.json`, right?

itowlson: Perfect - thanks!

itowlson: To clarify, are these changes in progress or are they pushed?  There's no hurry of course, just checking my assumptions!

itowlson (approved):
Thanks - and sorry for the hitch on the description!

examples/spin-timer/readme.md
0xE282B0: Good point, should have been pushed.Good point, should have been pushed. But weren't.

PR #2156, opened 2023-12-11
Basic Socket Support

rylev:
This PR adds basic socket support to Spin through the `allowed_outbound_hosts` mechanism. 

By default any usage of Socket APIs will return a `access-denied` error. The user can opt into Socket usage by adding entries to a component's `allowed_outbound_hosts` array. For example:

```toml
[component.my-component]
source = "my_component.wasm"
allowed_outbound_hosts = ["*://127.0.0.1:5001"]
```

Sockets usage is only allowed when the scheme is set to '*'. This is because we have no way of knowing what type of traffic is happening over a socket and so even if the socket is only handling HTTP traffic, Spin cannot detect this. `allowed_outbound_hosts` already supports lists of hosts (e.g., `["*://{127.0.0.1, 192.168.0.1}:5001"]`), lists of ports, and port ranges. This should work with all of that. . 

### Limitations and future expansions

Currently the host portion of the `allowed_outbound_hosts` string (i.e., the '127.0.0.1' in the example above) must be an IP address. In the future, we will add support for DNS host names but this requires ongoing work in wasmtime. 

When the user allows a `*` host (e.g., `allowed_outbound_hosts = ["*:/*:5001"]`, this will turn on all socket traffic. In the near future, we may want to limit this a bit to not include privileged ports on private network (i.e., all ports from 1-1023) and perhaps even UDP.


test-components/tcp-sockets/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+impl Component {
+    fn main() -> Result<(), String> {
+        let address = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 5001);
+
+        let client = ensure_ok!(tcp_create_socket::create_tcp_socket(IpAddressFamily::Ipv4));
+
+        ensure_ok!(client.start_connect(
+            &instance_network::instance_network(),
+            match address {
+                SocketAddr::V6(address) => {
```
dicej, line 29: This branch won't be hit given the hard-coded IPv4 address above, so we might as well remove it for now.

test-components/tcp-sockets/component.wasm
dicej, line 1: Did you mean to check this file in?  If so, what is it for and how was it created?

dicej (commented):
Looking good so far!

Will we support IPv6 addresses in `allowed_outbound_hosts`?  If so, it would be good to have a test for that.  And maybe support for whatever the IPv6 equivalent of `format!("{host}/32")` is?

test-components/tcp-sockets/component.wasm
rylev: Yes, this is how the test-components repository works. It is simply the result of running `cargo b --target=wasm32-wasi` on this crate (plus it's run through `wasm-tools strip` to reduce binary size).

rylev:
@dicej - IP v6 should be supported, but we might have to make some changes to the `allowed_outbound_hosts` parser first. I'd like to merge this first and tackle that next if you don't mind.

test-components/tcp-sockets/component.wasm
dicej: What's the advantage of checking these files in to the repo vs. building them as needed?  And are there docs or a script anywhere indicating how to regenerate them?  I went looking, but couldn't find anything.

rylev: The advantage is just that it's quicker and less complex to not have to build them from scratch each time. Eventually I'd like to move them a different repo and have CI ensure that the pre-built versions are the same as you would get from building them from scratch. 

I've added some of this context to the README: https://github.com/fermyon/spin/tree/sockets/test-components#test-components

dicej (approved):

PR #2157, opened 2023-12-11
rust-sdk: add more request helpers

### LINKED ISSUE #2155: Rust SDK request helpers
Currently, the Rust SDK offers a few helpers for creating HTTP requests. Specifically, it implements functions for creating GET and POST requests easily, but not anything else.

This means that a program that needs to create different requests ends up looking something similar to:

```rust
        let _res: Response =
            spin_sdk::http::send(Request::post(store_url.clone(), order_json.clone())).await?;

        // Get state from the store
        let res: Response =
            spin_sdk::http::send(Request::get(format!("{}/{}", store_url, order.id))).await?;
        let retrieved_order: Order = serde_json::from_slice(&res.body())?;
        println!("Got order: {:?}", retrieved_order);

        // Delete state from the store
        let _res: Response = spin_sdk::http::send(
            Request::builder()
                .method(spin_sdk::http::Method::Delete)
                .uri(format!("{}/{}", store_url, order.id))
                .build(),
        )
        .await?;
```

Obviously, I can just use the builder approach for all, or just be okay with the inconsistency.

Decided to open just in case others are mildly bothered by this as well.

fibonacci1729:
Adds helpers for `DELETE`, `PUT`, `PATCH`.

Closes #2155

### DIFF
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -61,6 +61,25 @@ impl Request {
         builder
     }
 
+    /// Creates a [`RequestBuilder`] to PUT the given `body` to `uri`
+    pub fn put(uri: impl Into<String>, body: impl conversions::IntoBody) -> RequestBuilder {
+        let mut builder = RequestBuilder::new(Method::Put, uri);
+        builder.body(body);
+        builder
+    }
+
+    /// Creates a [`RequestBuilder`] to PATCH the resource specified by `uri`
+    pub fn patch(uri: impl Into<String>, body: impl conversions::IntoBody) -> RequestBuilder {
+        let mut builder = RequestBuilder::new(Method::Patch, uri);
+        builder.body(body);
+        builder
+    }
+
+    /// Creates a [`RequestBuilder`] to DELETE the resource specified by `uri`
+    pub fn delete(uri: impl Into<String>) -> RequestBuilder {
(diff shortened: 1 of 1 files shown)


sdk/rust/src/http.rs
```diff
@@ -61,6 +61,30 @@ impl Request {
         builder
     }
 
+    /// Creates a [`RequestBuilder`] to fetch the headers from `uri`
+    pub fn head(uri: impl Into<String>) -> RequestBuilder {
+        RequestBuilder::new(Method::Head, uri)
+    }
```
rylev, line 67: Head requests are pretty rare, is this worth the extra method? The work around is pretty straightforward.

sdk/rust/src/http.rs
```diff
@@ -61,6 +61,30 @@ impl Request {
...
+
+    /// Creates a [`RequestBuilder`] to DELETE the resource specified by `uri`
+    pub fn delete(uri: impl Into<String>) -> RequestBuilder {
```
rylev, line 70: Nit: I would put `delete` after `put` and `patch` so that those methods are next to `post`.

rylev (changes requested):
I would personally prefer if we did not add a `head` method.

rylev (approved):

PR #2151, opened 2023-12-08
chore(cargo/sdk): prep for publishing spin-sdk crate

vdice:
Some updates in preparation for publishing the spin-sdk crate per https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate

TODO:
- [x] publish the `spin-macro` crate and update the reference in `spin-sdk`'s Cargo.toml (https://github.com/fermyon/spin/pull/2151#discussion_r1421043540)

Ref https://github.com/fermyon/spin/issues/316

PR #2160, opened 2023-12-11
ci(release.yml): add crates publishing job

vdice:
- Adds a step to publish the `spin-sdk` crate on each v* release tag

Depends on https://github.com/fermyon/spin/pull/2151

TODO:
  - [x] Open Question: What to do re: publishing the `spin-macro` crate, a dependency of the `spin-sdk` crate?
     -  Keep `spin-macro` crate version separate from workspace/spin-sdk version (Currently it is [0.1.0](https://crates.io/crates/spin-macro))
         - Manually publish crate when this version is bumped, manually create PR to update crate version in spin-sdk's Cargo.toml
         - Add automation to publish crate when this version is bumped (could be slightly tricky if not associated with a git tag), potentially also add automation to create a PR to spin-sdk's Cargo.toml to rev the version
     - Align `spin-macro` crate version with workspace version (i.e. release tag version) and publish prior to publishing `spin-sdk` (Haven't tested this, will the approach work via back-to-back jobs in a GH workflow?)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -469,3 +469,23 @@ jobs:
           file: _dist/spin-${{ env.RELEASE_VERSION }}-static-${{ env.RUNNER_OS }}-${{ matrix.config.arch }}.tar.gz
           tag: ${{ github.ref }}
           prerelease: ${{ steps.release-version.outputs.prerelease == 'true' }}
+
+  crates:
+    name: Publish to crates.io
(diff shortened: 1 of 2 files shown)


dicej:
Given that `spin-macro` changes about as often as `spin-sdk`, we might as well give them the same version and publish them at the same time.

dicej (approved):

PR #2161, opened 2023-12-12
Fix Windows not copying globs or directories

### LINKED ISSUE #2111: Asset directory mounts do not work on Windows
The was reported by freaky.ashes on Discord - thanks to them for flagging it.

To reproduce:

* Create a Spin app with an asset directory mount (e.g. `files = [{ source = "assets", destination = "/" }]`)
* Run it on Windows
* You get a "couldn't resolve <temp>\assets\<component>" error for the component asset directory, because the assets directory didn't get created

This happens because of a conflict between canonicalisation and the glob crate: when a glob is prefixed with a canonical (UNC) path, it _does not match any files_.  (And asset directories get created only when files exist to be copied into them.)  To reproduce this, create this Rust app:

```rust
use std::path::PathBuf;

fn main() {
    println!("Using plain dir");
    let bp = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
    let paths = glob::glob(bp.join("src/**/*").to_str().unwrap()).unwrap();
    for path in paths {
        println!("{path:?}");
    }

    println!("Using canonicalised dir");
    let bp = PathBuf::from(env!("CARGO_MANIFEST_DIR")).canonicalize().unwrap();
    let paths = glob::glob(bp.join("src/**/*").to_str().unwrap()).unwrap();
    for path in paths {
        println!("{path:?}");
    }
}
```

It prints something like

```
Using plain dir
Ok("D:\\Code Experiments\\globtest\\src\\main.rs")
Using canonicalised dir
```

This breaks directory mounts on Spin 2 because we canonicalise the source path, and use `<source-path>/**/*` to implement directory copying.  I assume it also breaks glob mounts too but have not tested that case.

This is a known bug in the upstream glob crate: https://github.com/rust-lang/glob/issues/132.  The issue suggests a workaround which, although unsafe for general use, should be fine for us.  Alternatively it might suffice to absolutise rather than canonicalise.  The issue is 6 months old and there's no movement on it so we should not expect a fix from upstream.

itowlson:
Fixes #2111,

An alternative approach would be to remove the UNC prefix as mentioned in https://github.com/rust-lang/glob/issues/132, which I can implement if people prefer.

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -30,8 +30,7 @@ impl LocalLoader {
         files_mount_strategy: FilesMountStrategy,
         cache_root: Option<PathBuf>,
     ) -> Result<Self> {
-        let app_root = app_root
-            .canonicalize()
+        let app_root = safe_canonicalize(app_root)
             .with_context(|| format!("Invalid manifest dir `{}`", app_root.display()))?;
         Ok(Self {
             app_root,
@@ -388,6 +387,20 @@ impl LocalLoader {
             path: dest.into(),
         })
     }
+}
+
+/// This canonicalizes the path in a way that works with globs. On non-Windows
+/// platforms, we can use Path::canonicalize, but on Windows platforms this
+/// expands to a UNC path, and the glob library does not work with UNC paths.
+#[cfg(not(windows))]
+fn safe_canonicalize(path: &Path) -> std::io::Result<PathBuf> {
+    path.canonicalize()
+}
+
+#[cfg(windows)]
+fn safe_canonicalize(path: &Path) -> std::io::Result<PathBuf> {
+    use path_absolutize::Absolutize;
+    Ok(path.absolutize()?.into_owned())
 }
 
 fn locked_metadata(


crates/loader/src/local.rs
```diff
@@ -2,6 +2,7 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{bail, ensure, Context, Result};
 use futures::future::try_join_all;
+use path_absolutize::Absolutize;
```
karthik2804, line 5: should this also be gated to be required only on windows?

crates/loader/src/local.rs
```diff
@@ -390,6 +389,17 @@ impl LocalLoader {
     }
 }
 
+#[cfg(not(windows))]
+fn safe_canonicalize(path: &Path) -> std::io::Result<PathBuf> {
```
rylev, line 393: Can we add docs to one of these functions so there's a bit more context on why this is needed?

rylev (dismissed):

itowlson:
@rylev This is still blocked - I would like to get it into 2.1 if we can - do you still have concerns?

vdice (approved):

PR #2162, opened 2023-12-12
Fix double backslashes in printed Windows paths

### LINKED ISSUE #2158: Doubled backslashes in some Windows messages
E.g. `Logging component stdio to "D:\\Code Experiments\\staticfstest\\.spin\\logs\\"` or  `Failed to load Spin app from "D:\\Code Experiments\\staticfstest\\spin.toml"`.

This ~might be~ is down to using a debug representation of paths ~but I have not confirmed that~.

itowlson:
Fixes #2158.

This changes user-facing messages (errors and prints; I skipped logs) to print paths correctly on Windows.

```
# Before
Logging component stdio to ".\\target\\..\\..\\dirmounttest\\.spin\\logs\\"
# After
Logging component stdio to .\target\..\..\dirmounttest\.spin\logs\
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5910,6 +5910,7 @@ dependencies = [
  "rand 0.8.5",
  "safetensors",
  "serde",
+ "spin-common",
  "spin-core",
  "spin-llm",
  "spin-world",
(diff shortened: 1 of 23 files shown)


rylev (changes requested):
I'm happy with this change overall. We've had some debate about the right way to print paths, and I think this settles that debate. That being said, I do think we likely want to keep double quotes around paths as that usually makes it easier for readers to parse.

rylev (approved):

PR #2163, opened 2023-12-12
Build test components instead of relying on pre-built binaries

rylev:
Until now we've been relying on pre-built test component binaries for the runtime tests. This has the advantage of being very fast (the binaries are already built, and so there is no build phase at all). However, it also has the disadvantage of potentially getting out of sync with the source. This PR changes the test components to be built as a part of building Spin.

This is not without its drawbacks: for example, the test components are a part of the workspace meaning they will be built during `cargo build`. This does have the counter advantage though of making the test components part of rust-analyzer builds so that things like auto-complete work.

Ultimately, I think this is the right way to go for now. In the future, we may want to reconsider whether we move these to a separate repository and pre-build the binaries with some check that those pre-built binaries do not get of sync with their source.


tests/test-components/build.rs
```diff
@@ -0,0 +1,75 @@
...
+    let mut threads = Vec::new();
+    let mut name_to_path = HashMap::new();
+    for package in packages {
+        let o = out_dir.clone();
+        let manifest_path = PathBuf::from("components")
+            .join(&package)
+            .join("Cargo.toml");
+        let manfiest = cargo_toml::Manifest::from_path(&manifest_path).unwrap();
+
+        let handle = std::thread::spawn(move || {
```
dicej, line 22: Nit: does this actually give us concurrency?  If they're all part of the same workspace, won't all but one end up blocked waiting for the package cache file lock?

dicej (approved):
LGTM.  Thanks for doing this!

tests/test-components/build.rs
rylev: Yea, you're right. I'll remove.

.github/workflows/build.yml
```diff
@@ -87,7 +87,7 @@ jobs:
           rust-cache: true
 
       - name: Cargo Build
-        run: cross build --target x86_64-unknown-linux-musl --workspace --release --all-targets --features openssl/vendored
```
rylev, line 90: @rajatjindal FYI: I have removed these flags since they are unnecessary for building a Spin executable. I noticed this when the build was failing due to the `test-components` crate failing to build, but that crate does not need to be built to get a Spin executable.

rajatjindal: thank you for the heads up. these are also used in release workflow here https://github.com/fermyon/spin/blob/d41deb9eb33faeaa069d24a3922f26406c1cd2ca/.github/workflows/release.yml#L429C14-L429C20 

you may want to remove those as well in that case

PR #2164, opened 2023-12-13
Fix "duplicate package `key-value`" warning in Spin SDK git reference

itowlson:
This will drop being an issue once we are fully on `crates.io` (it would be nice, for those times when we want to test with main, but it will stop being user-facing).  So if the crates.io stuff lands in time then we can close this.


rajatjindal (approved):

rylev (commented):
I think this will be fixed by https://github.com/fermyon/spin/pull/2163 when it lands, so perhaps this change isn't necessary?

PR #2166, opened 2023-12-13
add support for `wasi:http/incoming-handler@0.2.0-rc-2023-11-10`

dicej:
This builds upon Alex's excellent work in
https://github.com/fermyon/spin/pull/2108, adding the ability to run guests which target the `0.2.0-rc-2023-11-10` version of `wasi-http`.

I've added a test which uses `wit-bindgen` and `wit-component` directly to ensure it uses the correct snapshot for everything:`wasi-http`, `wasi-cli`, etc.

Note that we're not yet updating any SDKs to use the new snapshot, since we still want people to be able to use the new SDK to target Spin 2.0.


alexcrichton (approved):

rylev (commented):
Is there any reason that the tests aren't runtime tests? If not, I can submit a PR to convert them.

dicej:
> Is there any reason that the tests aren't runtime tests? If not, I can submit a PR to convert them.

I try to avoid Docker when testing and debugging since it adds overhead and (IMO) not a lot of value for simple tests like this.  Can (a subset of) the runtime tests be run without it?

rylev:
@dicej we only run inside of Docker in CI currently because that's where the e2e tests run - the runtime tests have no dependency on Docker and can be run outside of a Docker container (which I do locally). Avoiding one-off tests like this is the exact goal of the runtime tests, so if you're feeling resistance towards adding one, it makes me want to have a conversation again about how we unify testing.

dicej:
I didn't realize Docker was optional -- glad to hear it is.  Maybe we can pair up and you can walk me through converting this to a runtime test?

PR #2171, opened 2023-12-14
Map correct token count for llm:InferencingResult

laurentiustamate94:
Used `generated_token_count` when mapping to `InferencingResult.usage.generated_token_count`


rylev (approved):
Whoops! Thanks 🎉

PR #2179, opened 2023-12-15
fix statically linked build issue for aarch

rajatjindal:
The aarch cross image was using old version of Ubuntu. I had to switch to using image marked as 'edge' in cross-rs repo to fix this.

a new version of cross has not been released in a long time, but using a sha256 of image should be safe until we have a better alternative.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,6 +73,45 @@ jobs:
           name: spin-ubuntu-latest
           path: target/release/spin
 
+  build-spin-static:
+    name: Build Spin static
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        config:
+          - {
+              arch: "aarch64",
+              target: "aarch64-unknown-linux-musl",
+              rustflags: '-C target-feature=+fp16 -C target-feature=+crt-static -C link-self-contained=yes'
+            }
+          - {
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -11,7 +11,7 @@ on:
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 env:
-  RUST_VERSION: 1.71
+  RUST_VERSION: 1.73
```
vdice, line 14: What is the reason for the bump here?  It's drifting from the [current MSRV](https://github.com/fermyon/spin/blob/main/Cargo.toml#L15).

.github/workflows/build.yml
```diff
@@ -73,6 +73,58 @@ jobs:
           name: spin-ubuntu-latest
           path: target/release/spin
 
+  build-rust-static:
```
vdice, line 76: Any reason we're not using a `matrix` strategy for these two, to slim down the code footprint?

rajatjindal: no specific reason. i can change it to use matrix strategy.

rajatjindal: this is now done

vdice (approved):

PR #2187, opened 2023-12-17
Mutable access to request headers

### LINKED ISSUE #2186: Mutate HTTP request headers
A user on Discord wanted to inject info into an HTTP request via the headers before passing it to the router

This seems to be possible via the `http::Request` type using the `headers_mut()` function, but does not seem to be available using the `spin_sdk::http::Request` type which we now recommend.

itowlson:
Fixes #2186.

This ended up being not symmetric with `Request::headers()` but I couldn't see a way to get closer (e.g. `iter_mut()` didn't seem like it would allow for inserts).  It also surfaces the implementation in what feels like a rather raw way.  @rylev Please feel free to close in favour of a better way if you have one, I'd love a learning opportunity!

### DIFF
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -116,6 +116,16 @@ impl Request {
         self.headers.get(&name.to_lowercase())
     }
 
+    /// Set a header
+    pub fn set_header(&mut self, name: impl Into<String>, value: impl Into<String>) {
+        self.headers.insert(
+            name.into(),
+            HeaderValue {
+                inner: HeaderValueRep::String(value.into()),
+            },
+        );
+    }
+
     /// The request body
     pub fn body(&self) -> &[u8] {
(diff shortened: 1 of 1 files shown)


rylev (changes requested):
The function was added to the `Response` type which I don't think was intended? In any case, would it be possible to use `set_header` instead of having mutable access to the headers collection?

lann:
> The function was added to the `Response` type which I don't think was intended? In any case, would it be possible to use `set_header` instead of having mutable access to the headers collection?

:+1: for `set_header` to match the behavior of https://docs.rs/spin-sdk/2.1.0/spin_sdk/http/struct.RequestBuilder.html#method.header

itowlson:
Well, doh.  Thanks for the catch and guidance.  Updated.

rylev (approved):

PR #2189, opened 2023-12-19
print error msg when plugin binary is missing

rajatjindal:


src/commands/external.rs
```diff
@@ -66,7 +67,15 @@ pub async fn execute_external_subcommand(
     )
     .await?;
 
-    let mut command = Command::new(plugin_store.installed_binary_path(&plugin_name));
+    let binary = plugin_store.installed_binary_path(&plugin_name);
+    if !Path::new(binary.as_os_str()).exists() {
+        return Err(anyhow!(
+            "plugin executable {:?} is missing. Try uninstalling and installing the plugin again.",
```
itowlson, line 73: Use the `common::ui::quoted_path()` function (`anyhow!("plugin executable {} ...", quoted_path(&binary))`) - this ensures consistent formatting and correct rendering across platforms.  (We found that the debug format prints double backslashes on Windows.)

src/commands/external.rs
```diff
@@ -66,7 +67,15 @@ pub async fn execute_external_subcommand(
     )
     .await?;
 
-    let mut command = Command::new(plugin_store.installed_binary_path(&plugin_name));
+    let binary = plugin_store.installed_binary_path(&plugin_name);
+    if !Path::new(binary.as_os_str()).exists() {
```
itowlson, line 71: `binary` is already a `PathBuf` so has the `exists` method built in - no need to extract the string or construct a `Path`.

itowlson (changes requested):
The content is fine - thanks!  Just a couple of style niggles that it would be good to clear up - sorry.

src/commands/external.rs
rajatjindal: thank you ❤️

src/commands/external.rs
rajatjindal: 💯  thank you

rajatjindal:
thanks for the review Ivan. I have addressed the review comments. made a small change in error msg to include the plugin name explicitly. I hope that is ok.

itowlson (approved):
Looks grand now - thanks!

PR #2195, opened 2023-12-21
Stop testing kv and sqlite twice

rylev:
This PR simplifies the Makefile and stops us from testing kv and sqlite twice. Currently the kv and sqlite spinup_tests will be tested both during integration testing and e2e tests.

It seems that this makes testing in CI faster on ubuntu by ~16 minutes and macOS by ~30 minutes (mainly due to us not running the build step which was not necessary and very costly).


radu-matei (approved):
This is excellent, thanks @rylev!

PR #2199, opened 2023-12-21
Patched build.rs to run on systems without rustup. Added flake for ni…

benwis:
In order to build spin on systems without rustup, you can't use rustup to check for targets. I've replaced that check with one from rustc. If it's not installed it still fails for nix users as it can't install it, but if it is it passes at least

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,6 @@
 
 {
-  description = "A basic Rust devshell";
+  description = "Nix Devshell for Spin";
 
   inputs = {
     nixpkgs.url      = "github:NixOS/nixpkgs/nixos-unstable";
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,39 @@
+
+{
+  description = "A basic Rust devshell";
+
+  inputs = {
+    nixpkgs.url      = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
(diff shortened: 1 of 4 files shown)


itowlson:
@benwis Could you sign off the commit (and GPG sign if not already done) please?  Thanks!

flake.nix
```diff
@@ -0,0 +1,39 @@
+
+{
+  description = "A basic Rust devshell";
```
endocrimes, line 3: We should probably update this to reference spin, but that can be done as a follow up

endocrimes (commented):
mostly lgtm, I'm ooo rn but I can give this a try tonight

flake.nix
benwis: I mentioned spin in the new commit

benwis:
> @benwis Could you sign off the commit (and GPG sign if not already done) please? Thanks!

I will try, I haven't setup code signing before, shouldn't be too bad

benwis:
> @benwis Could you sign off the commit (and GPG sign if not already done) please? Thanks!

Signed!

endocrimes (approved):
Tested that this works correctly on linux - Darwin test is taking a longer because well... a laptop is not a server 😅 - I'll follow up with anything needed for nix-darwin builds though.

benwis:
> Tested that this works correctly on linux - Darwin test is taking a longer because well... a laptop is not a server 😅 - I'll follow up with anything needed for nix-darwin builds though.

Thank you!

PR #2201, opened 2023-12-22
Wasi HTTP runtime test

rylev:
This adds a runtime test for `wasi:http/outgoing-handler` specifically for version `0.2.0-rc-2023-11-10`.

The purpose of this PR is:
* Add a runtime test that could replace the integration test that was already testing this interface
* Explore the addition of a feature of the test component repository for explicitly adapting test components with a given version of the wasi adapter.

## Removing the integration test

I have not yet removed the integration test though I think one can make the argument that we can. The integration test exercises both the outgoing and incoming handlers while this runtime test (like all runtime tests) does not concern itself with triggers and therefore only really exercises the outgoing handler.

To be more precise, the runtime test runner uses whatever http handler is defined in the top-level `wit` directory which for now happens to be version `0.2.0-rc-2023-10-18`. However, this is considered an implementation detail. At some point in the future we will add an explicit testing framework for testing trigger handlers.

I will leave it up to @dicej whether he thinks removing the integration test is fine to do or not.  At the very least, I think we can remove most of the code from the integration test since there's no really need to test the outgoing handler in both an integration test *and* a runtime test.

## Handling adapter versions

Since this test is explicitly about ensuring support for a specific version of the adapter, this PR also introduces a way to adapt test components with a specific adapter version.

Previously, the test components were not actually components but rather wasm modules that Spin knows how to convert to components during the loading process using `spin_componentize`. This PR keeps this status quo for all test components with the exception of the newest one. Now any time a component's name ends in `v0.2.0-rc-2023-11-10`, we run the module through the appropriate component adapter for that version.

In the future, I think it would be good to *always* adapt components since we're explicitly not testing Spin's ability to do adaption of modules to components in runtime tests, but this is left to a future change.

## http-echo server

Lastly, this PR adds an `http-echo` server runtime test service written in Python. Any test in the future that wants to exercise outbound http, can use this service.


lann:
This is a bit of a tricky one since outgoing wasi uses pollables. I think we'll certainly want to keep at least some integration tests around proxy-style forwarding of request/response bodies; iirc there were a number of subtle bugs around those during Spin 2.0 development.

tests/test-components/build.rs
```diff
@@ -30,19 +35,42 @@ fn main() {
...
-        let wasm = out_dir
+        let package_name = manifest.package.expect("manifest has no package").name;
+        let binary_name = package_name.replace(['-', '.'], "_");
+        let wasm_path = out_dir
             .join("wasm32-wasi")
             .join("debug")
             .join(format!("{binary_name}.wasm"));
 
+        let adapter_version = package.split('v').last().and_then(|v| match v {
+            "0.2.0-rc-2023-11-10" => Some("15.0.1"),
```
lann, line 49: Why this mapping rather than naming the adapter with the full rc version?

tests/test-components/build.rs
```diff
@@ -3,23 +3,28 @@ use std::{collections::HashMap, path::PathBuf, process::Command};
 fn main() {
     println!("cargo:rerun-if-changed=components");
     println!("cargo:rerun-if-changed=helper");
```
lann, line 5: `println!("cargo:rerun-if-changed=adapters");`?

tests/test-components/build.rs
```diff
@@ -30,19 +35,42 @@ fn main() {
...
+                .expect("failed to encode component");
+            std::fs::write(&wasm_path, new_bytes).expect("failed to write new wasm binary");
+        }
+
         // Generate const with the wasm binary path
-        generated_code += &format!("pub const {const_name}: &str = {wasm:?};\n");
-        name_to_path.insert(package, wasm);
+        generated_code += &format!(
+            "pub const {const_name}: &str = \"{}\";\n",
+            wasm_path.display()
```
lann, line 71:
```suggestion
            "pub const {const_name}: &str = {wasm_path:?};\n"
```

:shrug:

lann (approved):

dicej (commented):
I think this is a good exploration of how to make the runtime test framework more flexible, but I'm having trouble seeing the advantages of _replacing_ the existing integration test in this case.

90% of the reason for the integration test is to test the inbound handler -- that's the part that was previously missing and which I added, so that's what I was primarily focused on testing.  The outbound part was something I tacked on because it was easy.  So I don't want to remove coverage from the inbound side, even temporarily.

I do agree that we should strive to minimize overlap among tests, and I also think the integration test is way heavier than it needs to be -- it shouldn't have to start a whole spin process to test this kind of thing.  But I also feel that it has a couple of distinct advantages over the runtime test (inbound coverage and use of ephemeral ports) which I'd rather not lose.

tests/test-components/components/wasi-http-v0.2.0-rc-2023-11-10/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+use helper::{ensure_eq, ensure_ok};
+use wasi::http::outgoing_handler;
+use wasi::http::types::{Headers, Method, OutgoingBody, OutgoingRequest, Scheme};
+use wasi::io::streams::StreamError;
+
+helper::define_component!(Component);
+
+impl Component {
+    fn main() -> Result<(), String> {
+        let url = url::Url::parse("http://localhost:8080").unwrap();
```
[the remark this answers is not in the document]
lann: It would probably be robust _enough_ for the test runner to bind a random port, unbind, then pass that port in to both the service and test.

There is also a hack for this which can _mostly_ resolve races where you bind a random port, open a connection to that port, then close the port without closing the connection which puts the socket into TIME_WAIT, then SO_REUSEADDR in the service.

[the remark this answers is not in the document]
rylev: Yes, support for ephemeral ports is added in https://github.com/fermyon/spin/pull/2193, but I didn't want to have to reimplement it here. I will add that once #2193 is merged.

rylev:
@dicej makes a ton of since. Looks like we'll want to replace the integration test with another test, but we should wait until we have a better test for the inbound trigger exports. 👍

> use of ephemeral ports

As noted in the other comment, support for ephemeral ports was added in https://github.com/fermyon/spin/pull/2193 and so we can take advantage of that when that PR lands.

tests/test-components/build.rs
rylev: As noted in https://github.com/fermyon/spin/issues/2158, using the debug implementation does the wrong thing in Windows.

lann: Are these tests actually going to be runnable in windows? :grimacing:

lann: I guess the service-less ones ought to be...

tests/test-components/build.rs
rylev: The boring answer is that the version string is long and makes me sad, and the adapter file name was named after the wasmtime version so I decided to do the mapping. It's an easy change, but I'd like to take a step back and think a bit deeper about how we want to do the automagical adapter step, but I think that can wait for a PR in the near future. How does that sound?

lann: `15.0.1` just seems more confusing if you don't already have deep wasmtime knowledge. :shrug: 

Certainly not a blocker.

tests/test-components/build.rs
rylev: The Python ones should also be runnable, and if they're not that's a bug. I believe the docker containers should also be runnable on Windows, but that's not something I have too much experience with so we'll have to see.

tests/test-components/build.rs
rylev: Yea I hear you. I can change it to your suggestion. I agree it's less confusing. This might change in the future though.

tests/test-components/build.rs
rylev: lul - actually Windows failed because we _weren't_ using the debug output... Switched back.

PR #2203, opened 2023-12-23
ignore asset globs from being watched when using direct-mounts

### LINKED ISSUE #2202: Spin watch - Do not restart on changes to mounted files when using `--direct-mounts`
It would be nice if Spin does not restart when started using the `watch` command with the argument of `--direct-mounts`.  It could be a nice optimization to avoid having to restart as sometimes, spin can take a few moments to restart if it has to load in larger wasm modules.

karthik2804:
fixes #2202

### DIFF
--- a/src/commands/watch.rs
+++ b/src/commands/watch.rs
@@ -132,8 +132,13 @@ impl WatchCommand {
         let source_code_tx = Arc::new(source_code_tx);
         let manifest_tx = Arc::new(manifest_tx);
 
+        // No need to restart on asset changes if direct-mounting
+
+        let contains_direct_mounts = self.up_args.contains(&"--direct-mounts".to_owned());
+
         let artifact_filterer = Box::new(ArtifactFilterFactory {
             skip_build: self.skip_build,
+            skip_assets: contains_direct_mounts,
         });
         let (artifact_watcher, artifact_watcher_handle) = self
             .spawn_watchexec(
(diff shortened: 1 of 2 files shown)


src/commands/watch.rs
```diff
@@ -132,8 +132,14 @@ impl WatchCommand {
         let source_code_tx = Arc::new(source_code_tx);
         let manifest_tx = Arc::new(manifest_tx);
 
+        // Check if the --direct-mounts flag is set We can skip watching assets
+        // if we are direct mounting instead of a tempfs
```
calebschoepp, line 136:
```suggestion
        // Check if the --direct-mounts flag is set. We can skip watching assets
        // if we are direct mounting instead of a tempfs
```

Nit: Seems like you used a plugin to wrap the comments but I think you want a period in that sentence.

calebschoepp (approved):
Lets make sure Ivan looks at this too.

src/commands/watch.rs
```diff
@@ -132,8 +132,14 @@ impl WatchCommand {
         let source_code_tx = Arc::new(source_code_tx);
         let manifest_tx = Arc::new(manifest_tx);
 
+        // Check if the --direct-mounts flag is set We can skip watching assets
```
itowlson, line 135: Nit: the first part of the comment restates the code - you can probably simplify the comment to e.g. "No need to restart on asset changes if direct-mounting"

itowlson (commented):
This all LGTM although I see it is marked as draft?

src/commands/watch.rs
karthik2804: good catch. I simplified the comment as suggested by @itowlson

src/commands/watch.rs
karthik2804: Done.

karthik2804:
@itowlson updated the comment and marked it as ready for review.

itowlson (approved):

PR #2206, opened 2024-01-02
Add random port support to Python services

rylev:
With https://github.com/fermyon/spin/pull/2193 we introduced Docker services for runtime testing that can assign random ports on the host to well know ports on the guest. This PR does the same for Python based services which can have well known ports mapped to randomly assigned ports.

The Python services do this by printing a `PORT=(P1,P2)` message to stdout where P1 is the well known port and P2 is the randomly assigned port. The runtime testing runtime parses this message and knows how to replace the well known ports in the test spin.toml files with the random ports.

With this change, we're at a point where runtime tests should run on most host machines without any modification.

### DIFF
--- a/tests/runtime-tests/src/services/python.rs
+++ b/tests/runtime-tests/src/services/python.rs
@@ -1,6 +1,9 @@
+use crate::OutputStream;
+
 use super::Service;
 use anyhow::Context as _;
 use std::{
+    cell::OnceCell,
     collections::HashMap,
     path::Path,
     process::{Command, Stdio},
@@ -8,6 +11,8 @@ use std::{
 
 pub struct PythonService {
     child: std::process::Child,
+    stdout: OutputStream,
+    ports: OnceCell<HashMap<u16, u16>>,
     _lock: fslock::LockFile,
 }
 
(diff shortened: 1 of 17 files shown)


tests/runtime-tests/src/services/python.rs
```diff
@@ -17,18 +22,23 @@ impl PythonService {
...
             )
-            // Ignore stdout
-            .stdout(Stdio::null())
+            .stdout(Stdio::piped())
             .spawn()
             .context("service failed to spawn")?;
-        Ok(Self { child, _lock: lock })
+        std::thread::sleep(std::time::Duration::from_millis(1000));
+        Ok(Self {
+            stdout: OutputStream::new(child.stdout.take().expect("TODO")),
```
rajatjindal, line 37: is this TODO expected here?

tests/runtime-tests/src/services/python.rs
```diff
@@ -53,8 +72,35 @@ impl Service for PythonService {
...
+            Some(ports) => Ok(ports),
+            None => {
+                let ports = stdout
+                    .lines()
+                    .filter_map(|l| l.trim().split_once('='))
+                    .map(|(k, v)| -> anyhow::Result<_> {
+                        let k = k.trim();
+                        let v = v.trim();
+                        if k == "PORT" {
+                            let err = "malformed service kv pair - PORT values should be in the form PORT(80,8080)";
```
rajatjindal, line 90: nit: should be in form PORT=(80,8080)

rajatjindal: also **maybe** say "port pair" instead of "kv pair".

rajatjindal (approved):

PR #2209, opened 2024-01-03
feat(cli.rs): allow setting app log dir via SPIN_LOG_DIR env var

vdice:
- Allows setting the app's log directory via an `SPIN_LOG_DIR` env var

Unsure if it was an intentional decision to _not_ allow setting via env var, though?

I happened to notice this when deploying to the OSS Fermyon Platform, using Hippo, which attempts to set this env var: https://github.com/deislabs/hippo/blob/main/src/Infrastructure/Services/NomadJobService.cs#L189

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -44,6 +44,7 @@ where
         name = APP_LOG_DIR,
         short = 'L',
         long = "log-dir",
+        env = "SPIN_LOG_DIR",
     )]
     pub log: Option<PathBuf>,
 


crates/trigger/src/cli.rs
```diff
@@ -43,6 +43,7 @@ where
     #[clap(
         name = APP_LOG_DIR,
         short = 'L',
+        env = APP_LOG_DIR,
```
itowlson, line 46: For an environment variable I'd consider `SPIN_LOG_DIR` (as Hippo used) rather than `APP_LOG_DIR`.  Reduces the risk of a clash, and makes it easier for someone poring through the environment to understand what the variable relates to.

itowlson: (to be clear `APP_LOG_DIR` is fine for the _name_ because that only gets used when printing help strings, at which point the user is strictly in the context of `spin up`... I am just wary of it as a name in the _global_ environment)

itowlson (approved):

fibonacci1729 (approved):

PR #2211, opened 2024-01-05
Abstract the runtime used in runtime tests so that it's not hardcoded as Spin

rylev:
The next step after this is to see if we can put this to use against another Spin compliant runtime.

### DIFF
--- a/tests/runtime-tests/src/lib.rs
+++ b/tests/runtime-tests/src/lib.rs
@@ -1,26 +1,29 @@
+pub(crate) mod io;
 mod services;
+pub mod spin;
 
 use std::{
-    io::Read,
     path::{Path, PathBuf},
-    process::{Command, Stdio},
     sync::OnceLock,
 };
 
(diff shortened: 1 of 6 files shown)


tests/runtime-tests/src/lib.rs
```diff
@@ -210,169 +197,11 @@ fn copy_manifest(
...
-    fn output_as_str(&mut self) -> Option<&str> {
-        std::str::from_utf8(self.output()).ok()
-    }
+pub enum TestResult {
+    /// The test passed
+    Pass,
+    /// Wasm errored (the wasm error, additional error info)
+    Err(String, String),
+    /// Wasm failed to run (additional error info)
+    Fatal(String),
```
lann, line 206: Bikeshedding, maybe something like `Err` -> `Fail` and `Fatal` -> `RuntimeError` might be more clear?

lann (approved):

PR #2212, opened 2024-01-05
SIP 016 - Inbound WebSocket Support

dicej:


docs/content/sips/016-inbound-websockets.md
```diff
@@ -0,0 +1,189 @@
...
+}
+
+interface inbound-websocket-receive {
+  use wasi:http/types.{incoming-request, response-outparam};
+  use websocket-types.{frame, close-frame};
+
+  // Construct a new handler for the specified request, which the
+  // client has requested to be upgraded to a WebSocket.
+  //
+  // - `id`: a globally unique identifier which may be used to send
```
lann, line 124: Maybe name this `token` to make it more obvious that it is sensitive?

docs/content/sips/016-inbound-websockets.md
```diff
@@ -0,0 +1,189 @@
...
+  // - `response-out`: a resource which may be used to send a
+  //   response to the client. If (and only if) the response has a
+  //   status code of 101, then the connection will be upgraded to a
+  //   WebSocket.
+  handle-new: func(id: string, request: incoming-request, response-out: response-outparam);
+
+  // Handle the most recent frames received for the specified
+  // WebSockets. `frames` consists of a mapping from WebSocket ids
+  // to ordered lists of frames.
+  handle-frames: func(frames: list<tuple<string, list<received-frame>>);
```
lann, line 143: I'm a little skeptical about batching frames from different connections in the same call since it implies processing requests from multiple end users in the same instance sandbox. On the flip side I'd be fine with (opportunistically) reusing an instance for consecutive `handle-frames` calls from the same websocket (side note: we could potentially do the same for pipelined http requests...🤔).

docs/content/sips/016-inbound-websockets.md
```diff
@@ -0,0 +1,189 @@
...
+
+interface inbound-websocket-send {
+  use websocket-types.{frame, future-send-frame};
+
+  // Attempt to send the specified frame to specified WebSocket.
+  //
+  // This will fail if the WebSocket is no longer open. Otherwise,
+  // if the frame is an instance of `close`, the connection will
+  // be closed after sending the frame.
+  send-frame: func(id: string, frame: frame) -> future-send-frame;
```
lann, line 154: Should this allow batched sends?

docs/content/sips/016-inbound-websockets.md
```diff
@@ -0,0 +1,189 @@
...
+  // passed to `incoming-websocket-receive#handle-new`.
+  export inbound-websocket-receive;
+
+  import inbound-websocket-send;
+}
+```
+
+### Spin Implementation
+
+We’ll add a new trigger type called `websocket` which will be handled by the `spin-trigger-http` crate. This crate will handle normal (non-WebSocket) incoming requests as usual: instantiating the component, calling the `wasi:http/incoming-handler#handle` method, and discarding the instantiation when the entire response has been sent. For WebSocket requests (i.e. those with headers requesting a connection upgrade per the WebSocket protocol), it will instead call `fermyon:spin/inbound-websocket-receive#handle-new`, and if that produces a valid 101 ("Switching Protocols") response, the host component will discard the instance and upgrade the connection to a WebSocket.
```
lann, line 172: Do we need a new trigger type? The http trigger already supports multiple worlds (spin, wasi, wagi) chosen at runtime based on the shape of the component. Any inherent reason why it couldn't also choose based on the shape of the request?

Alternatively, maybe this could _extend_ the WASI http inbound world with this `handle-frames` export and a new e.g. `upgrade-websocket: function(inbound-request) -> token`.

docs/content/sips/016-inbound-websockets.md
dicej: @rylev had concerns that reusing an instance for multiple calls could be confusing to developers.  I don't have a strong opinion either way.

docs/content/sips/016-inbound-websockets.md
dicej: I considered that, but I figured that if an app has multiple messages to send to a given client and wants to do that efficiently, they can bundle them into a single frame (i.e. handle it in the app-level protocol).  Also, we can always add new `send-frames` function later, if desired.  I don't have strong feelings about it, though.

docs/content/sips/016-inbound-websockets.md
dicej: To be clear, my thinking was that the same code would be shared between `http` and `websocket` trigger types.  And yes, extending the wasi-http world is exactly what I'm proposing -- note the `include wasi:http/proxy;` line in the world.

The reason I'm proposing a new trigger type named `websocket` is to make it clear to Spin that it can rely on the component to export `inbound-websocket-receive`.  It's not necessary, since Spin can always ask the component what it exports, but I figured it's best to be explicit -- the same way we're explicit about other trigger types even though we could arguably guess based on the exports.

dicej:
Note that I've added a new `State Management` section near the bottom; please take a look if you have a chance.

docs/content/sips/016-inbound-websockets.md
rylev: There's currently the implied assumption that when a component is entered, it's a new instance. If we change that assumption, I think we should at least document that and examine that tradeoffs that that has.

lann: Ignoring instance reuse for the moment, my main concern here is processing messages from multiple users in a single instance. One of the main advantages of Spin's instance-per-request model is that an e.g. buffer overflow is sandboxed to the requestor's instance.

lann (approved):

PR #2213, opened 2024-01-08
feat(up): Spawn multiple trigger commands

### LINKED ISSUE #1755: Support for multiple trigger types in a single Spin application
This has been frequently mentioned in conversations, and also in issues (E.g., #1005), but I couldn't find an issue to track the request, so here it is.

A frequent Discord thread laid out a relevant scenario as well: https://discord.com/channels/926888690310053918/1149614859185569852

carlokok:
This PR does a few things and fixes #1755:
- Fix the HTTP/Redis plugin to use the lock file's metadata.triggers.http/redis path instead of metadata.trigger field.
- Use tokio to async wait for the sub processes to end (I didn't want to introduce a new dependency on async processes)

End result:
WindowsTerminal_fSadslpDYA

One note: I cannot get the integration tests to run well on Windows. Probably a local issue, so I was unable to verify if the integration tests work.

### DIFF
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -7,6 +7,7 @@
 #![deny(missing_docs)]
 
 mod host_component;
+use serde_json::Value;
 pub use spin_locked_app::locked;
 pub use spin_locked_app::values;
 pub use spin_locked_app::{Error, MetadataKey, Result};
@@ -212,16 +213,29 @@ impl<'a, L> App<'a, L> {
         &'this self,
         trigger_type: &'a str,
     ) -> Result<Option<T>> {
-        self.locked.metadata.get("triggers").map_or(Ok(None), |t| {
-            t.get(trigger_type)
-                .map(T::deserialize)
-                .transpose()
-                .map_err(|err| {
-                    Error::MetadataError(format!(
-                        "invalid metadata value for {trigger_type:?}: {err:?}"
-                    ))
-                })
-        })
+        let Some(value) = self.get_trigger_metadata_value(trigger_type) else {
+            return Ok(None);
+        };
+        let metadata = T::deserialize(value).map_err(|err| {
+            Error::MetadataError(format!(
+                "invalid metadata value for {trigger_type:?}: {err:?}"
+            ))
+        })?;
+        Ok(Some(metadata))
+    }
+
+    fn get_trigger_metadata_value(&self, trigger_type: &str) -> Option<Value> {
+        if let Some(trigger_configs) = self.locked.metadata.get("triggers") {
+            // New-style: `{"triggers": {"<type>": {...}}}`
+            trigger_configs.get(trigger_type).cloned()
+        } else if self.locked.metadata["trigger"]["type"] == trigger_type {
+            // Old-style: `{"trigger": {"type": "<type>", ...}}`
+            let mut meta = self.locked.metadata["trigger"].clone();
+            meta.as_object_mut().unwrap().remove("type");
+            Some(meta)
+        } else {
+            None
+        }
     }
 
     /// Returns an iterator of [`AppTrigger`]s defined for this app with
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -207,6 +207,23 @@ impl<'a, L> App<'a, L> {
             .map(|locked| AppTrigger { app: self, locked })
     }
 
+    /// Returns the trigger metadata for a specific trigger type.
+    pub fn get_trigger_metadata<'this, T: Deserialize<'this> + Default>(
+        &'this self,
+        trigger_type: &'a str,
+    ) -> Result<Option<T>> {
+        self.locked.metadata.get("triggers").map_or(Ok(None), |t| {
+            t.get(trigger_type)
+                .map(T::deserialize)
+                .transpose()
+                .map_err(|err| {
+                    Error::MetadataError(format!(
+                        "invalid metadata value for {trigger_type:?}: {err:?}"
+                    ))
+                })
+        })
+    }
+
     /// Returns an iterator of [`AppTrigger`]s defined for this app with
     /// the given `trigger_type`.
     pub fn triggers_with_type(
--- a/crates/http/src/trigger.rs
+++ b/crates/http/src/trigger.rs
@@ -7,8 +7,6 @@ pub const METADATA_KEY: MetadataKey<Metadata> = MetadataKey::new("trigger");
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct Metadata {
-    // The type of trigger which should always been "http" in this case
-    pub r#type: String,
     // The based url
     #[serde(default = "default_base")]
     pub base: String,
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -237,7 +237,7 @@ impl UpCommand {
         #[cfg(not(windows))]
         {
             // https://github.com/nix-rust/nix/issues/656
-            let pid = nix::unistd::Pid::from_raw(child.id() as i32);
+            let pid = nix::unistd::Pid::from_raw(child.id() as u32);
             ctrlc::set_handler(move || {
                 if let Err(err) = nix::sys::signal::kill(pid, nix::sys::signal::SIGTERM) {
                     tracing::warn!("Failed to kill trigger handler process: {:?}", err)
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -16,6 +16,8 @@ use spin_oci::OciLoader;
 use spin_trigger::cli::{SPIN_LOCAL_APP_DIR, SPIN_LOCKED_URL, SPIN_WORKING_DIR};
 use tempfile::TempDir;
 
+use futures::future::try_join_all;
+
 use crate::opts::*;
 
 use self::app_source::{AppSource, ResolvedAppSource};
@@ -154,7 +156,11 @@ impl UpCommand {
             .with_context(|| format!("Couldn't find trigger executor for {app_source}"))?;
 
         if self.help {
-            return self.run_trigger(trigger_cmd, None).await;
+            for cmd in trigger_cmd
+            {
+                    self.run_trigger(cmd.clone(), None).await?;
+            }
+            return Ok(());
         }
 
         let mut locked_app = self
@@ -165,13 +171,16 @@ impl UpCommand {
 
         let local_app_dir = app_source.local_app_dir().map(Into::into);
 
-        let run_opts = RunTriggerOpts {
-            locked_app,
-            working_dir,
-            local_app_dir,
-        };
+        try_join_all(trigger_cmd.iter().map(|cmd| {
+            let run_opts = RunTriggerOpts {
+                locked_app: locked_app.clone(),
+                working_dir: working_dir.clone(),
+                local_app_dir: local_app_dir.clone(),
+            };
 
-        self.run_trigger(trigger_cmd, Some(run_opts)).await
+            self.run_trigger(cmd.clone(), Some(run_opts))
+        }))
+        .await.map(|_| ())
     }
 
     fn get_canonical_working_dir(&self) -> Result<WorkingDirectory, anyhow::Error> {
@@ -191,13 +200,14 @@ impl UpCommand {
     }
 
     async fn run_trigger(
-        self,
+        &self,
         trigger_cmd: Vec<String>,
         opts: Option<RunTriggerOpts>,
     ) -> Result<(), anyhow::Error> {
         // The docs for `current_exe` warn that this may be insecure because it could be executed
         // via hard-link. I think it should be fine as long as we aren't `setuid`ing this binary.
-        let mut cmd = std::process::Command::new(std::env::current_exe().unwrap());
+        
+        let mut cmd = tokio::process::Command::new(std::env::current_exe().unwrap());
         cmd.args(&trigger_cmd);
 
         if let Some(RunTriggerOpts {
@@ -235,7 +245,7 @@ impl UpCommand {
             })?;
         }
 
-        let status = child.wait()?;
+        let status = child.wait().await?;
         if status.success() {
             Ok(())
         } else {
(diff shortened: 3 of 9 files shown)


carlokok:
Fixed the print! that should never have been there.

src/commands/up.rs
```diff
@@ -165,13 +174,18 @@ impl UpCommand {
...
         let local_app_dir = app_source.local_app_dir().map(Into::into);
 
-        let run_opts = RunTriggerOpts {
-            locked_app,
-            working_dir,
-            local_app_dir,
-        };
-
-        self.run_trigger(trigger_cmd, Some(run_opts)).await
+        join_all(trigger_cmd.iter().map(|cmd| {
```
lann, line 177: What happens if one of the trigger executors fails to run? If I'm not missing anything I believe `spin up` would happily continue executing the other triggers which probably (?) isn't what we want.

crates/app/src/lib.rs
```diff
@@ -207,6 +207,22 @@ impl<'a, L> App<'a, L> {
             .map(|locked| AppTrigger { app: self, locked })
     }
 
+    /// Returns the trigger metadata for a specific trigger type.
+    pub fn get_trigger_metadata<'this, T: Deserialize<'this> + Default>(
```
lann, line 211: Unfortunately we need to maintain backwards-compatibility in the locked app manifest, so we can't just change things around like this.

crates/http/src/trigger.rs
```diff
@@ -2,13 +2,11 @@ use serde::{Deserialize, Serialize};
 use spin_locked_app::MetadataKey;
 
 /// Http trigger metadata key
-pub const METADATA_KEY: MetadataKey<Metadata> = MetadataKey::new("trigger");
+pub const METADATA_KEY: MetadataKey<Metadata> = MetadataKey::new("triggers.http");
```
lann, line 5: Ditto re: backward compat.

lann (changes requested):
Thanks for the contribution! I think this would benefit from some discussion around a few key points:

- `spin up` UX:
  - How should `--help` work for multiple trigger types
  - If a single trigger executor dies does `spin up` quit completely?
- Backwards compatibility

src/commands/up.rs
```diff
@@ -154,7 +156,14 @@ impl UpCommand {
             .with_context(|| format!("Couldn't find trigger executor for {app_source}"))?;
 
         if self.help {
-            return self.run_trigger(trigger_cmd, None).await;
+            return join_all(
```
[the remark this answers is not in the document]
carlokok: I can run them sequentially, but that will still print a lot more than it should.

src/commands/up.rs
carlokok: I'll look into it.

crates/app/src/lib.rs
carlokok: Note that I'm not changing the format here. the .lock file has something like (for a single trigger, multiple triggers are supported but don't write out the single trigger item):
```
{
  "spin_lock_version": 0,
  "metadata": {
    "authors": [
      "itowlson <ivan.towlson@fermyon.com>"
    ],
    "name": "cron-sample",
    "origin": "file:///C:/projects/spintest/spin.toml",
    "triggers": {
      "http": {
        "base": "/shop"
      }
    },
    "trigger": {
      "type": "http": 
      "base": "/shop"
    },
    "version": "0.1.0"
  },
```

crates/http/src/trigger.rs
carlokok: This isn't used anymore in my pr, bad leftover. If we can wrap up above I can remove this one.

src/commands/up.rs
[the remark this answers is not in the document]
lann: For `--help` in particular I think the ideal UX would be to print trigger-specific flags in separate sections, but that would be a bigger chunk of work. Also I'm realizing (as I think @itowlson alluded to earlier) that there is a more fundamental problem here with how to parse trigger-specific flags and I'm really not sure how to handle that.

crates/app/src/lib.rs
lann: Ah sorry, I added the `"triggers"` key as part of the migration to Spin 2.0 but we still need to be able to read the (redundant) old `"trigger"` keys for backward-compat. I need to figure out how to document all of this somewhere...

crates/app/src/lib.rs
```diff
@@ -207,6 +207,22 @@ impl<'a, L> App<'a, L> {
...
+        &'this self,
+        trigger_type: &'a str,
+    ) -> Result<T> {
+        Some(&self.locked.metadata["triggers"][trigger_type])
```
lann, line 215: This will panic if `"triggers"` or the specific `trigger_type` don't exist. This whole function should probably return `Result<Option<T>>` instead, with `Ok(None)` if the key(s) are missing.

src/commands/up.rs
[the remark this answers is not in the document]
itowlson: If it comes to it, I'd argue we can punt on help printing more than it should - it will be untidy, but we can track it as a "known issue with the first release" and it's not worth blocking getting the feature into people's hands.

Re trigger specific flags, one thing I tried in the composite trigger plugin experiment was prefixing flags with the type of the trigger for which they were intended e.g. `--http:listen / --http-listen`.  The plugin would look for prefixes matching the composed triggers and would selectively pass the de-munged flags on.  It's probably tedious to require that on all flags, but perhaps it could be used to disambiguate when you want to pass different flag values with the same name e.g. `--http-listen 3001 --websocket-listen 3002`?  Just thinking out loud here though.

crates/app/src/lib.rs
itowlson: There's potentially also an issue where plugins which are interested in trigger types may not know about the `metadata/triggers` key, but rely on `metadata/trigger`.  (Although my main worry was the Cloud plugin, and that looks directly at the trigger entries rather than the metadata.)

This may be addressable by continuing to populate `metadata/trigger` if there is a single type, and setting it to something like `type: composite` if there are multiple types?  That way, existing "check trigger type" code will still work, and the error if it encounters a composite app should be semi-meaningful.

src/commands/up.rs
itowlson, line 177: I believe this should exit as soon as _any_ trigger exits - otherwise the user could end up with only half their application running.

That said, there has been vague talk of a `run` or `once` trigger, which would invoke its guest and then exit.  If that trigger existed, we'd need to be more nuanced about cancelling other triggers in the same application.  That's speculative at the moment though!

src/commands/up.rs
```diff
@@ -235,7 +249,7 @@ impl UpCommand {
             })?;
         }
 
-        let status = child.wait()?;
+        let status = tokio::task::spawn(async move { child.wait() }).await??;
```
itowlson, line 252: nit: Would it be more natural to change this function to use a `tokio::process::Command`, so that you merely had to await `child.wait()` rather than explicitly spawn a task and awaiting that?

src/commands/up/app_source.rs
```diff
@@ -83,7 +83,7 @@ pub enum ResolvedAppSource {
 }
 
 impl ResolvedAppSource {
-    pub fn trigger_type(&self) -> anyhow::Result<&str> {
+    pub fn trigger_type(&self) -> anyhow::Result<Vec<&str>> {
```
itowlson, line 86: I know I am getting into the weeds at this point, but this will need to be renamed `trigger_types`

itowlson (commented):
Just a couple of comments, mostly at the nit level.  I'm really excited to see this!

src/commands/up.rs
carlokok: That would be a lot better, had I known that existed :) This is pretty much my first serious venture into Rust. I'll fix.

src/commands/up.rs
carlokok: Changed to try_join_all that seems to give this effect.

crates/app/src/lib.rs
carlokok: Fixed this with a fallback to look at either.

src/commands/up.rs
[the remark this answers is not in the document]
carlokok: So I think this one should be something you guys have to get to agreement on. It's now not properly working due to timers not supporting listen, but this is more of a choice on approach than something I can fix myself.

carlokok:
I did a new push and *think* I fixed all comments you provided. Let me know if you see anything else that needs fixing.

dicej:
Lann and Ivan asked me to comment on this from the perspective of the [WebSocket proposal](https://github.com/fermyon/spin/pull/2212).  It looks fine to me, and I think it's pretty much orthogonal to the addition of a `websocket` trigger.

The current plan is for the `websocket` trigger type to be an alias for the `http` trigger type with an additional feature enabled (the ability to upgrade HTTP connections to WebSocket connections).  I don't expect that supporting multiple trigger types in a single `spin up` command will affect that one way or another.

itowlson:
@carlokok I think I see what is happening with the timer trigger.  Because `TRIGGER_METADATA_KEY` has switched over to `triggers`, the deserialisation type now need to reflect the `timer` nesting e.g.

```rust
// This is the "triggers" element of the lockfile, which can contain multiple
// objects e.g. "http", "timer", etc. Ignore all the ones the timer trigger
// doesn't care about.
#[derive(Clone, Debug, Default, Deserialize, Serialize)]
struct TriggerMetadataParent {
    timer: TriggerMetadata,
}

#[derive(Clone, Debug, Default, Deserialize, Serialize)]
#[serde(deny_unknown_fields)]
struct TriggerMetadata {
    speedup: Option<u64>,
}

// This now needs to point to the *parent* schema not the timer-specific schema
const TRIGGER_METADATA_KEY: MetadataKey<TriggerMetadataParent> = MetadataKey::new("triggers");

#[async_trait]
impl TriggerExecutor for TimerTrigger {
    // ...
    async fn new(engine: spin_trigger::TriggerAppEngine<Self>) -> anyhow::Result<Self> {
        let speedup = engine
            .app()
            .require_metadata(TRIGGER_METADATA_KEY)?
            .timer               // <------- new element here
            .speedup
            .unwrap_or(1);
        // ...
    }
}
```

Looking into some of the other troubles you noted as well - thanks for persevering with this.  The devil was always going to end up in the details...!

carlokok:
Thanks! I'll try to fix the linting issue when I get back in the office tomorrow.

itowlson:
@carlokok I think that 'linting' issue is actually a compiler error that `child.id()` is now an `Option` rather than a plain number (a side effect I didn't realise when suggesting changing to `tokio::process::Command`, sorry).

Although this is reasonably easy to fix (e.g. using `if let` or `Option::map`), there's also going to be a problem with callint `ctrlc::set_handler` in a loop.  That function can only be called _once_ for a process: when there are two trigger types, the second call panics.

I _think_ the way this ends up looking is moving the Ctrl+C handling out of the `run_trigger` function, and instead collecting the futures from the calls to `run_trigger` and having a single Ctrl+C handler that loops over them.  This may have its own ramifications in the design of `run_trigger` though.

Does that all make sense?  I'm happy to sketch this out in more detail if that would be useful.  (Or I can send you a PR for it, if that's better for you - I already had to rough this out because I am on WSL, so had to make these tweaks anyway!)

carlokok:
If you can help me a bit with a PR that would be greatly appreciated. I'm not currently setup with Linux development and the error does not trigger on Windows.

itowlson:
@carlokok Is there any further support we can give you to land this?  I ask this not in order to hound you, but just wanted to check in, in case you're stuck or need more information or discussion from us.

carlokok:
I was actually wondering if anything else needs to be fixed or changed. But now I see some of the checks failed so I will try to reproduce it locally tomorrow and fix it

lann:
I also created a PR against your branch that reorganizes the metadata code: https://github.com/carlokok/spin/pull/3

carlokok:
Would it be possible that someone does another quick review & run the workflows? Iann did a good fix I think and it shows at least what testcases I need ot look at.

carlokok:
@itowlson I think @lann s pr might have fixed your concerns too. All tests pass too now. Is there anything else you want me change further?

itowlson:
@carlokok Thanks for the update.  I've kicked the tyres on this and it seems good.  What I tried:

* HTTP (built in) single and with timer
* Timer (updated) single and with HTTP
* SQS (not updated) single and with HTTP and timer

Everything worked except for SQS in the multi-trigger configuration.  This is because SQS, being linked to the existing trigger crates, looks at `trigger` instead of `triggers`, which is not set in the multi-trigger case.  There's nothing we can do about this, and unfortunately I don't think we can even improve the error handling because the `unwrap()` is baked into the existing trigger crate.

There is still something amiss in the SQS + timer + HTTP case where, in one of my tests, timer did not exit after SQS crashed.  (HTTP _did_ exit, so this is likely a race condition.)  I am okay with continuing to work on this as a separate PR - I don't think it's a blocker for this one, but it's something to consider for the release.

There are also a few supporting changes that don't block this PR but need picking up, e.g. the template experience for "add a X trigger to a Y application".  (Plus of course docs but we know to do those as part of the relevant release docs.)

@melissaklein24 @lann Given that status, and the upcoming 2.2 release, how do you want to play this?  We can either:

1. Merge now, include ir in 2.2, document any known bugs, and announce it as an experimental feature in 2.2.
2. Merge now, include it in 2.2, but don't announce it immediately - give it a few weeks to shake out bugs and UX stuff like templates, and announce it in 2.3
3. Hold off merging until after 2.2, just in case there are any lurking nasties in the single-trigger scenario.  (Which I don't think there are, but I wouldn't stake my life on it.)  Then merge immediately after 2.2 so we can start the shaking-out process ready for 2.3.
4. Something else

What do you reckon?

itowlson (approved):

itowlson:
@carlokok Could you GPG-sign the commits please?  We require GPG signing, but unfortunately GitHub doesn't seem to check it until the PR is approved and ready to merge.  Sorry about this.

carlokok:
@itowlson I had to rebase the whole thing to resign it but I think it should be good now.

itowlson:
@carlokok Thanks - all green now!

@lann 👍

Issue #2170, opened 2023-12-14
Using the `Json` extractor with the 2.1 Rust template


itowlson:
In Spin 2.0.1, the Rust template referenced the `http` crate, and we supported the `Json<T>` extractor as a body type parameter e,g, `http::Request<Json<Person>>`.

In Spin 2.1, the Rust template does _not_ reference the `http` crate and users are directed instead to the `spin_sdk::http::Request` type.  This type is _not_ generic - the body is always presented as bytes - and we do not have seem to have samples of how to use the extractors with it.  I guess you can do something like `Jsion::<Person>::try_from_body(req.into_body())` but it's not as friendly as the extractor syntax!

If the extractors work with the SDK request type, it would be good to provide samples that show how; if they don't, we should provide that functionality.

rylev:
The following used to work in the run up to 2.0 but was removed because of some fears that it was potentially confusing:

```rust
use std::collections::HashMap;
use spin_sdk::{http::{Json, Response}, http_component};

/// A simple Spin HTTP component.
#[http_component]
fn handle(body: Json<HashMap<String, String>>) -> Response {
    println!("Incoming Body: {body:#?}");
    Response::new(200, Vec::new())
}
```

We could make this work again if we want, but of course, the user then doesn't have access to any other parts of the request. 

Alternatively, we could add a `json` method to `Request` that allows users to access the body in some deserialized form.

itowlson:
A `json` method feels very specific (although admittedly the "80% case" kind of specific, with `String` and maybe form-encoded being the other biggies).  I had also wondered about a function called something like `into_body_as()` or `extract`, but the `req.extract::<Json<Person>>()` syntax is a bit spiky.

Another, although maybe hairy, possibility is a typed form of request e.g. `spin_sdk::http::typed::Request<T>`.  It could be a testing and maintenance pain though, and of course a discoverability issue and extra cognitive load for users.

PR #2216, opened 2024-01-09
Typed requests without needing the `http` crate

itowlson:
This explores a possible approach to #2170.

With this PR, a user can write a `http_component` entry point using an extractor, without needing the `http` crate (as this is not referenced in the template), by `use`-ing `spin_sdk::http::typed::Request<T>` instead of `spin_sdk::http::Request`.  E.g.

```
// Note typed::Request rather than Request
use spin_sdk::http::{IntoResponse, typed::Request, Response, Json};

#[derive(Deserialize)]
struct Fie {
    thing: String,
    whatsit: i32,
}

#[http_component]
fn handle_test(req: Request<Json<Fie>>) -> anyhow::Result<impl IntoResponse> {
    let fie = req.body()?;
    let text = format!("The thing was {} and the whatsit was {}", fie.thing, fie.whatsit);
    Ok(Response::builder()
        .status(200)
        .body(text)
        .build())
}
```

I don't love the duplication of methods - I don't know if this can be made more maintainable.  But anyway putting this up for feedback and discussion.

### DIFF
--- a/sdk/rust/src/http.rs
+++ b/sdk/rust/src/http.rs
@@ -484,6 +484,84 @@ impl std::fmt::Display for Method {
     }
 }
 
+/// Provides a typed Request type with automatic conversion of the body.
+pub mod typed {
+    use super::{
+        conversions::{IncomingRequestError, TryFromBody, TryFromIncomingRequest},
+        HeaderValue, IncomingRequest, Method,
+    };
+    use std::marker::PhantomData;
+
+    /// A typed request object that represents an incoming request whose body is expected
+    /// to be in a known format.  For example, if the body is expected to be a string,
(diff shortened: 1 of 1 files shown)


fibonacci1729:
I'm similarly not entirely convinced this approach is worth it either ... I think I would be more in favor of an "extractor" as mentioned in the linked issue in the near term (which although is less aesthetic, provides reasonable convenience around having to go from serialized bytes to some user provided data structure).

Having said that, @itowlson I am amendable to what you're proposing here if you feel strongly about the DX.

itowlson:
The people have spoken!  Thanks for the feedback - I'll close.

PR #2217, opened 2024-01-09
Generalize runtime testing framework to be a general testing Spin framework

rylev:
This PR makes the runtime testing framework  more general and allows its use in integration and e2e testing. It will be a little while still until we can fully move over to this framework for those tests but the idea is that all integration like tests will run against this framework. 

The framework provides a way to:
* instantiate a Spin compliant runtime (for now just Spin but easily extensible to other Spin compliant runtimes)
* run services that the runtime relies on (e.g., Redis, MySQL, etc.)
* populate Spin.toml manifest templates with runtime determined values
* build Spin apps from source without needing to add them manually to build.rs
* run tests against the runtime without needing to hard code things like which port the test is running on or where the component binary lives

## Writing new tests

The testing framework aims to try to make adding new tests easy. Currently, runtime tests are supported in a first class fashion which allows for a convention driven way of declaring tests without writing very much code at all. However, the integration tests still work in an a more imperative way. We'll have to wait and see if it makes since to keep things this way in the long term or not. The point is that the framework can handle various ways to declare tests depending on whatever makes sense for the use case.

The integration tests work in the following way:

Delcare a testing environment and run tests against it

```rust
let mut env = testing_framework::TestEnvironment::up(config)?;
env.test(test)?;
```

The testing environment is configured through a `TestEnvironmentConfig` which declares how the runtime is started and how services are configured. For example a `Spin` runtime environment appropriate for integration tests is declared as follows:

```rust
        let config = testing_framework::TestEnvironmentConfig::spin(
            spin_binary().into(),  // The location of `spin_binary`
            preboot, // What to do before booting the runtime (e.g., moving files into place)
            testing_framework::ServicesConfig::none(),  // The services the runtime depends on (in this case none)
        );
```

The tests are simple functions that get a runtime and return a `TestResult`. 

```rust
      let test = |spin| {
            assert_status(spin, "/test/hello", 200)?;
            assert_status(spin, "/test/hello/wildcards/should/be/handled", 200)?;
            assert_status(spin, "/thisshouldfail", 404)?;
            assert_status(spin, "/test/hello/test-placement", 200)?;
            Ok(())
        };
```

A big chunk of this code is the same across all integration tests and so it can be easily shared but if an integration test needs more flexibility, it has it. 

## What comes next

With a few more tweaks, most of the integration tests can moved over to this allowing us to get rid of a lot of code in the integration test file. We should also be able to run the e2e tests against this as well.


tests/testing-framework/README.md
```diff
@@ -1,13 +1,48 @@
...
+Services allow for tests to be run against external sources. The service definitions can be found in the 'services' directory. Each test directory contains a 'services' file that configures the tests services. Each line of the services file should contain the name of a services file that needs to run. For example, the following 'services' file will run the `tcp-echo.py` service:
+
+```txt
+tcp-echo
+```
+
+Each service is run under a file lock meaning that all other tests that require that service must wait until the current test using that service has finished.
+
+The following service types are supported:
+* Python services (a python script ending in the .py file extension)
```
dicej, line 16: No changes needed here, but I just wanted to note that we could eventually support a third type of service using e.g. https://rust-script.org/, e.g. if there's useful Rust library providing a service we want to test against.

More generally, instead thinking of these as Python services, perhaps we could think of them as script services, allowing any of several extensions (.py, .rb, .rs, .sh, etc.) to be used, with the service author selecting the language according to the libraries available.

tests/testing-framework/src/manifest_template.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// Read a manifest template from a file.
+    pub fn from_file(path: impl AsRef<Path>) -> anyhow::Result<Self> {
+        let path = path.as_ref();
+        let manifest = std::fs::read_to_string(path)
+            .with_context(|| format!("could not read manifest template at '{}'", path.display()))?;
+        Ok(Self { manifest })
+    }
+
+    /// Substitute template variables in the manifest template.
+    pub fn substitute(&mut self, env: &mut TestEnvironment) -> Result<(), anyhow::Error> {
```
dicej, line 22: If this ever gets more complicated, we might want to consider using an existing template library, e.g. https://crates.io/crates/tinytemplate (which Spin already depends on).

dicej (commented):
Looks reasonable to me overall.

I expect we'll end up writing or modifying services fairly often.  For example, if I were to thoroughly test outbound HTTP, I'd want the service to have a whole bunch of endpoints I can hit, with a mix of methods, request body handlers, expected request headers, response bodies, etc..  As I come up with more ideas to test I'll need to add more features to the service.

BTW, say I wanted to write an outbound HTTP test such that the service will return a large-ish file in the response body, and I want to make sure the guest receives exactly the expected bytes.  How would that look in this scheme?  Can we generate assets on the fly (e.g. a file full of random bytes) so we don't have to check them into the repo, but also give both the service and the test access to those assets?

tests/testing-framework/README.md
rylev: Adding support for more types of scripting services is reasonable, but I think it's important that we limit the supported languages to ones that we can be generally sure are widely available across many different dev machines. Python is widely chosen because it's usually available on a dev's machine and if not, it is very easy to install. Other languages like bash might be common on some machines but a pain to install on others (e.g., on Windows outside of WSL contexts). 

So in general, I'm in favor of adding more support, but I'd like to make sure that we don't explode the choice of scripting language without an understanding of how it impacts how easy it is to run these tests.

rylev:
> I expect we'll end up writing or modifying services fairly often. For example, if I were to thoroughly test outbound HTTP, I'd want the service to have a whole bunch of endpoints I can hit, with a mix of methods, request body handlers, expected request headers, response bodies, etc.. As I come up with more ideas to test I'll need to add more features to the service.

This shouldn't be a problem. Changes to services should be picked up automatically on each run so the dev loop for modifying services should be very quick. 

> BTW, say I wanted to write an outbound HTTP test such that the service will return a large-ish file in the response body, and I want to make sure the guest receives exactly the expected bytes. How would that look in this scheme? Can we generate assets on the fly (e.g. a file full of random bytes) so we don't have to check them into the repo, but also give both the service and the test access to those assets?

There's no specific support for coordination between services and runtimes, but the services and runtimes both run in the same temporarily directory and have access to anything that gets moved into that directory. So it should be relatively straight forward to generate the file, move it into the temporary directory, run the service that expects to find that file in its current directory, and write a test that also has access to that file.

lann (approved):

dicej (approved):

Issue #2190, opened 2023-12-20
Rust template 2.1 fails on Windows


itowlson:
Reported by a Discord user (https://discord.com/channels/926888690310053918/1186493596501803059/1186700173397721122).

Follow these steps _on Windows_ (it works fine on Linux):

1. Download and unzip Spin 2.1 from the GH release page, and install the 2.1 templates.

2. Create an application using the `http-rust` template.  The `Cargo.toml` looks like this:

```toml
[package]
name = "spin21test"
authors = ["itowlson <ivan.towlson@fermyon.com>"]
description = ""
version = "0.1.0"
edition = "2021"

[lib]
crate-type = ["cdylib"]

[dependencies]
anyhow = "1"
spin-sdk = { git = "https://github.com/fermyon/spin", tag = "v2.1.0" }

[workspace]
```

3. Run `spin build`.

**Expected behaviour**

It builds.

**Actual behaviour**

A slew of errors, mostly "failed to resolve/cannot find type", but the first of which is:

```text
error: expected `world`, `interface` or `use`, found '.'
            --> C:\Users\ivan\.cargo\git\checkouts\spin-91500438ac5656d2\4ca3a56\sdk\rust\./wit:1:1
             |
           1 | ../../wit
             | ^
  --> C:\Users\ivan\.cargo\git\checkouts\spin-91500438ac5656d2\4ca3a56\sdk\rust\src\lib.rs:24:5
   |
24 | /     wit_bindgen::generate!({
25 | |         world: "platform",
26 | |         path: "./wit",
27 | |     });
   | |______^
```

If I change the SDK reference to the `crates.io` reference (`spin-sdk = "2.1"`) it works fine; but that is not what the template uses.  (Although they should, once we get the scripts properly wrangled!)

ogghead:
I was able to reproduce by following the steps given (thank you!), and I was also able to reproduce the errors by building Spin directly on Windows with `make build`

Changing `path: "./wit"` to `path: "../../wit"` [on this line](https://github.com/fermyon/spin/blob/main/sdk/rust/src/lib.rs#L26) referenced by the build error allowed my local build to succeed. I did not verify whether this change builds on Unix systems though. I do not have context on the [wit](https://github.com/fermyon/spin/blob/main/sdk/rust/wit) file currently used as an intermediate - let me know if there is a reason that is needed! Otherwise, I am happy to publish a PR for this change tomorrow after the 9-5!

[...]

PR #2221, opened 2024-01-10
ci(templates): update rust templates automation to use the crates.io versions

vdice:
- Updates rust templates during versioning automation to use the spin-sdk crate version

Tested on fork via https://github.com/vdice/spin/actions/runs/7480272473; see also bot PR https://github.com/vdice/spin/pull/9

Ref https://github.com/fermyon/spin/issues/2190

Not sure if we'd like this to fully close 2190 since Windows users may still have trouble running the rust templates from main/canary, where we'll still use the `git` reference for the spin-sdk.


itowlson (approved):
Yay!  Thank you!

PR #2222, opened 2024-01-11
update to Wasmtime 17 and WASI 0.2.0-rc-2023-12-05

dicej:
As of this writing, Wasmtime 17 has not yet been released, so we're temporarily pointing at the `release-17.0.0` branch.  By the time it is released, WASI 0.2.0-rc-2023-12-05 will become WASI 0.2.0, at which point we'll need to update the WIT files and code to match.

The upshot is that we'll be supporting three snapshots indefinitely: 2023-10-18, 2023-11-10, and 2023-12-05 AKA the final 0.2.0 release.

Most of the changes here deal with supporting those three snapshots, while the rest deal with the Hyper 1.0 update, which led to a bit of awkwardness when mixing Hyper and Reqwest, since the latter has not yet been updated to use Hyper 1.0.


rylev (approved):
Looks good to me!

lann (approved):

alexcrichton (approved):

dicej:
I haven't been able to repro the CI test failures yet -- will keep investigating.

dicej:
Wasn't able to repro the CI failures on Mac or Linux.  Look like they're hitting a 20 second timeout waiting for `spin up` to start listening on the specified port, but they only take a couple of seconds to run on my machines.  Maybe even our beefy 4-core runner is struggling today?  Maybe we should consider running the integration tests using a release build of Spin instead of a debug build?

dicej:
I just ran the same tests on `main`, and they're not any faster there (when I run them locally), so there doesn't seem to be any actual performance regression here.

PR #2225, opened 2024-01-12
fix(registry/auth): auth validation workaround

radu-matei:
This commit updates the registry login logic to not error if the credential validation fails, but log the message and attempt to continue.

This behavior unblocks the use of registries whose login flow does not accept certain auth types, but if the credentials entered are wrong, could lead to users being appearing to be logged in, but unable to access the registry without logging in again with the right credentials.

Specifically, this most likely needs a --skip-validation flag to be explicitly passed by the user when logging in, but we probably need to rethink the login (and potentially add a logout / retry) flow.

We should most likely not merge this commit in its current state, its goal is to immediately unblock a current use case, but posted for future reference.

### DIFF
--- a/crates/oci/src/auth.rs
+++ b/crates/oci/src/auth.rs
@@ -50,7 +50,7 @@ impl AuthConfig {
         Ok(())
     }
 
-    fn default_path() -> Result<PathBuf> {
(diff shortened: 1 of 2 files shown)


vdice:
We've decided to pursue https://github.com/fermyon/spin/pull/2308 as an alternate fix.

PR #2228, opened 2024-01-16
Move more integration tests to use the testing framework

rylev:
This PR moves more tests over to using the testing frameworks. By and large this is mostly just deleting code that isn't used any more because the testing framework takes care of this. 

I would suggest looking at `tests/spinup_tests.rs` for the most relevant changes. 

cc @dicej who has authored some of these integration tests. As an example of the change that's happening here, the following is the code that is needed for setting up and executing a wasi-http integration test:

```rust
 fn test_wasi_http_rc(test_name: &str) -> anyhow::Result<()> {
        run_test(
            test_name,
            testing_framework::SpinMode::Http,
            [], // No `spin up` args
            testing_framework::ServicesConfig::new(vec!["http-echo".into()])?,
            |env| {
                let port = env
                    .get_port(80)?
                    .context("no http-echo port was exposed by test services")?;
                assert_spin_request(
                    env.runtime_mut(),
                    testing_framework::Request::full(
                        reqwest::Method::GET,
                        "/",
                        &[("url", &format!("http://127.0.0.1:{port}/",))],
                        Some("So rested he by the Tumtum tree".into()),
                    ),
                    200,
                    &[],
                    "Hello, world!",
                )?;
                Ok(())
            },
        )?;

        Ok(())
    }
```

If a new wasi-http test is to be added, the app just needs to be added to the test components repository (where it will be automatically built and run through the appropriate adapter) and then the following test needs to be added:

```rust
 test_wasi_http_rc("wasi-http-$VERSION")
```


dicej (commented):
Looks good overall.  I'm having trouble seeing where the `wasi_snapshot_preview1.reactor.wasm` files for the `rc-2023-11-10` and `rc-2023-12-05` tests are passed to `wit-component` now that the code previously responsible for that has been removed from `build.rs`.  Is there a new way that's done which I'm missing?

rylev:
@dicej I was about to write how this is indeed supported, but then I realized I was missing one change to support `rc-2023-12-05` that I've now pushed. Thanks for catching this!

In general, running test components through adapters is supported. The [test components](https://github.com/fermyon/spin/tree/main/tests/test-components) repository which supplies the test components used in runtime and e2e testing has support for running the test component through an adapter when it gets built. This happens in [the test-components crate's build.rs](https://github.com/fermyon/spin/blob/be88d813f43297744af766642b858577b22b5095/tests/test-components/build.rs#L50-L69).

I'm not 100% happy with how this mechanism works since it requires adding adapters to an allow list, but I am overall really happy that it's trivial to add new adapters and have components run through them. We can tweak this mechanism over time though.

dicej (approved):

PR #2230, opened 2024-01-18
trigger-http: Log TLS startup errors instead of propagating

lann:
This prevents TLS handshake errors from exiting the server loop.

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -344,8 +344,10 @@ impl HttpTrigger {
 
         loop {
             let (stream, addr) = listener.accept().await?;
-            let stream = acceptor.accept(stream).await?;
-            Self::serve_connection(self_.clone(), stream, addr);
+            match acceptor.accept(stream).await {
+                Ok(stream) => Self::serve_connection(self_.clone(), stream, addr),
+                Err(err) => tracing::error!(?err, "Failed to start TLS session"),
+            }
         }
     }
 }


crates/trigger-http/src/lib.rs
```diff
@@ -344,8 +344,10 @@ impl HttpTrigger {
 
         loop {
             let (stream, addr) = listener.accept().await?;
-            let stream = acceptor.accept(stream).await?;
-            Self::serve_connection(self_.clone(), stream, addr);
+            match acceptor.accept(stream).await {
```
rylev, line 347: If acceptor.accept is a bad state where it will always error this might just turn into an infinite loop no? Should we log and return the error?

lann: I don't think that should be an issue with the TLS acceptor. Note that the TCP listener just above does still bail.

rylev: I don't know enough about the TLS acceptor to say whether this is fine to do or not. I can't really give too much more of opinion without knowing why we don't want to propagate errors.

lann: The TLS acceptor handles the TLS handshake. It could fail because of an underlying TCP error but it can also fail because the handshake cannot be completed (for any number of reasons, some of which aren't under the server's control). Any error that is permanently fatal to the TCP listener should fail on the next loop iteration's call to `listener.accept`.

fibonacci1729 (approved):

PR #2231, opened 2024-01-18
Warn instead of trace when errors happen in the spin_redis_engine

karthik2804:
Changes the tracing level to `warn` as otherwise it leads to cases where the module errored and it looks like Spin hanged. I ran into this while using an invalid component. I could not understand its cause until @dicej recommended switching from `debug` to `trace`.

### DIFF
--- a/crates/redis/src/lib.rs
+++ b/crates/redis/src/lib.rs
@@ -94,7 +94,11 @@ impl TriggerExecutor for RedisTrigger {
         let mut stream = pubsub.on_message();
         loop {
             match stream.next().await {
-                Some(msg) => drop(self.handle(msg).await),
+                Some(msg) => {
+                    if let Err(err) = self.handle(msg).await {
+                        tracing::warn!("Error handling message: {err}");
+                    }
+                }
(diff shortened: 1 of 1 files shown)


dicej (approved):
LGTM.

This does raise the question of why we're silently dropping errors here: https://github.com/fermyon/spin/blob/main/crates/redis/src/lib.rs#L97 -- there's a non-trivial amount of code dedicated to tracking and collecting errors in that file, and yet we end up just swallowing them completely.

karthik2804:
@dicej what would be the recommended way of handling it? add some tracing there?

dicej:
I would suggest doing a `tracing::warn` in the `RedisTrigger::run` loop instead of dropping the error (which would make the change you already made unnecessary).  Either that, or stick with your current change and remove the `Result<()>` return types from `RedisTrigger::handle` and `RedisExecutor::execute` since they're just ignored anyway.

karthik2804:
@dicej @itowlson - I have updated the PR.

dicej (approved):

PR #2232, opened 2024-01-19
print active channels on redis trigger

karthik2804:
This PR prints active channels once the Redis trigger starts to improve the Ux slightly. In cases where the wasm modules take a while to prepare, it can look like Spin has hanged if there is no feedback to the user until they decide to send the first request and if it does not have any print statements, it will continue looking like that. 

(e.g.) 
```bash
$ spin up                                                                                                                  
Preparing Wasm modules is taking a few seconds...
```

after the pr it looks like

```bash
$ spin up
Logging component stdio to "examples/redis-rust/.spin/logs/"
Active Channels on redis://localhost:6379:
        messages: echo-message
```

### DIFF
--- a/crates/redis/src/lib.rs
+++ b/crates/redis/src/lib.rs
@@ -85,10 +85,12 @@ impl TriggerExecutor for RedisTrigger {
             .with_context(|| anyhow!("Redis trigger failed to connect to {}", address))?
             .into_pubsub();
 
+        println!("Active Channels on {address}:");
         // Subscribe to channels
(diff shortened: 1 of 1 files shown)


rylev (approved):
Awesome! This will help in testing to know when the Redis trigger has started. We [currently](https://github.com/fermyon/spin/blob/c25b685ed49e471be210138815fec2306093c56a/tests/testing-framework/src/spin.rs#L109-L110) just wait a bit of time and hope that's good enough.

PR #2238, opened 2024-01-22
allow installing select templates

### LINKED ISSUE #2235: install a specific template from a repository
the current ux around templates management is lacking a way to install a specific template from a repository. for example, I want to install https://github.com/fermyon/spin/tree/main/templates/http-go.

maybe something like the following could be implemented?

```bash
spin templates install \
    --git https://github.com/fermyon/spin \
    --branch main \
    --name http-go
```

karthik2804:
Adds the ability to be able to install specific templates using the following:

```
spin templates install --git https://github.com/fermyon/spin --id https-rust,http-go
...
Installed 2 template(s)

+-------------------------------------------------+
| Name        Description                         |
+=================================================+
| http-go     HTTP request handler using (Tiny)Go |
| http-rust   HTTP request handler using Rust     |
+-------------------------------------------------+

Skipped 10 template(s)

+-------------------------------------+
| Name                Reason skipped  |
+=====================================+
| http-c              Id not included |
| http-empty          Id not included |
| http-grain          Id not included |
| http-php            Id not included |
| http-swift          Id not included |
| http-zig            Id not included |
| redirect            Id not included |
| redis-go            Id not included |
| redis-rust          Id not included |
| static-fileserver   Id not included |
+-------------------------------------+
```

~~Todo: fix inconsistency between install and upgrade where upgrade always installs all the existing templates.~~ Currently `upgrade` installs all the templates it finds even if the user has manually deleted some templates. Is this the behavior we want or should it actually check if a template exists before upgrading it?

fixes #2235

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -68,6 +68,8 @@ pub enum SkippedReason {
     AlreadyExists,
     /// The template was skipped because its manifest was missing or invalid.
     InvalidManifest(String),
+    /// The template name does match the provided IDs
+    IdNotIncluded,
 }
 
 /// The results of installing a set of templates.
@@ -121,6 +123,7 @@ impl TemplateManager {
         source: &TemplateSource,
         options: &InstallOptions,
         reporter: &impl ProgressReporter,
+        install_template_list: &Vec<String>,
     ) -> anyhow::Result<InstallationResults> {
         if source.requires_copy() {
             reporter.report("Copying remote template source");
@@ -140,7 +143,13 @@ impl TemplateManager {
 
         for template_dir in template_dirs {
             let install_result = self
-                .install_one(&template_dir, options, source, reporter)
+                .install_one(
+                    &template_dir,
+                    options,
+                    source,
+                    reporter,
+                    install_template_list,
+                )
                 .await
                 .with_context(|| {
                     format!("Failed to install template from {}", template_dir.display())
@@ -163,6 +172,7 @@ impl TemplateManager {
         options: &InstallOptions,
         source: &TemplateSource,
         reporter: &impl ProgressReporter,
+        install_template_list: &Vec<String>,
     ) -> anyhow::Result<InstallationResult> {
         let layout = TemplateLayout::new(source_dir);
         let template = match Template::load_from(&layout) {
@@ -180,6 +190,15 @@ impl TemplateManager {
             }
         };
         let id = template.id();
+
+        if !install_template_list.is_empty() && !install_template_list.contains(&id.to_owned()) {
+            let message = format!("Skipping template {}...", id);
+            reporter.report(message);
+            return Ok(InstallationResult::Skipped(
+                id.to_owned(),
+                SkippedReason::IdNotIncluded,
+            ));
+        }
 
         let message = format!("Installing template {}...", id);
         reporter.report(&message);
@@ -479,7 +498,12 @@ mod tests {
         assert_eq!(0, manager.list().await.unwrap().templates.len());
 
         let install_result = manager
-            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .install(
+                &source,
+                &InstallOptions::default(),
+                &DiscardingReporter,
+                &Vec::new(),
+            )
             .await
             .unwrap();
         assert_eq!(TPLS_IN_THIS, install_result.installed.len());
@@ -510,7 +534,12 @@ mod tests {
         assert_eq!(0, manager.list().await.unwrap().warnings.len());
 
         let install_result = manager
-            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .install(
+                &source,
+                &InstallOptions::default(),
+                &DiscardingReporter,
+                &Vec::new(),
+            )
             .await
             .unwrap();
         assert_eq!(TPLS_IN_THIS, install_result.installed.len());
@@ -530,7 +559,12 @@ mod tests {
         let source = TemplateSource::File(project_root());
 
         manager
-            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .install(
+                &source,
+                &InstallOptions::default(),
+                &DiscardingReporter,
+                &Vec::new(),
+            )
             .await
             .unwrap();
 
@@ -547,7 +581,12 @@ mod tests {
         let source = TemplateSource::File(project_root());
 
         manager
-            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .install(
+                &source,
+                &InstallOptions::default(),
+                &DiscardingReporter,
+                &Vec::new(),
+            )
             .await
             .unwrap();
 
@@ -566,7 +605,12 @@ mod tests {
         let source = TemplateSource::File(project_root());
 
         manager
-            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .install(
+                &source,
+                &InstallOptions::default(),
+                &DiscardingReporter,
+                &Vec::new(),
+            )
             .await
             .unwrap();
 
@@ -585,7 +629,12 @@ mod tests {
         let source = TemplateSource::File(project_root());
 
         manager
-            .install(&source, &InstallOptions::default(), &DiscardingReporter)
+            .install(
+                &source,
+                &InstallOptions::default(),
+                &DiscardingReporter,
+                &Vec::new(),
+            )
             .await
             .unwrap();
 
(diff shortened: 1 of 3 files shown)


crates/templates/src/manager.rs
```diff
@@ -181,6 +191,15 @@ impl TemplateManager {
...
         let id = template.id();
 
+        if !install_template_list.is_empty() && !install_template_list.contains(&id.to_owned()) {
+            let message = format!("Skipping template {}...", id);
+            reporter.report(message);
+            return Ok(InstallationResult::Skipped(
+                id.to_owned(),
+                SkippedReason::IdNotIncluded,
+            ));
+        }
```
karthik2804, line 201: Should this be moved into its function? Feels wrong to be passing in a vec of IDs to the `install_one` method.

src/commands/templates.rs
```diff
@@ -181,18 +186,26 @@ impl Install {
...
 impl Upgrade {
     pub async fn run(&self) -> Result<()> {
+        let template_manager = TemplateManager::try_default()?;
+        let installed: Vec<String> = template_manager
+            .list()
+            .await?
+            .templates
+            .into_iter()
+            .map(|s| s.id().to_owned())
+            .collect();
```
karthik2804, line 196: Do we also want to filter based on the repo it is from?

crates/templates/src/manager.rs
```diff
@@ -68,6 +68,8 @@ pub enum SkippedReason {
     AlreadyExists,
     /// The template was skipped because its manifest was missing or invalid.
     InvalidManifest(String),
+    /// The template name does match the provided IDs
```
itowlson, line 71: Is there meant to be a 'not' here?  Otherwise I'm unsure what this is saying.

crates/templates/src/manager.rs
```diff
@@ -181,6 +191,15 @@ impl TemplateManager {
         };
         let id = template.id();
 
+        if !install_template_list.is_empty() && !install_template_list.contains(&id.to_owned()) {
+            let message = format!("Skipping template {}...", id);
```
itowlson, line 195: If the user has asked for a specific template, do they need to see skip messages for the other templates?  (Currently you print the skipped ones in the list anyway I think?)

crates/templates/src/manager.rs
itowlson: This is still installing one template though, so the name is no less correct.  But yeah, it's a bit weird that we call it for templates we don't want to install...!

crates/templates/src/manager.rs
```diff
@@ -181,6 +191,15 @@ impl TemplateManager {
         };
         let id = template.id();
 
+        if !install_template_list.is_empty() && !install_template_list.contains(&id.to_owned()) {
```
itowlson, line 194: This makes the empty vector a magic value that means "all" rather than "none."  Consider an enum with appropriate methods.

(One of those "the CLI / serialisation / etc. representation is not necessarily the correct _parsed_ representation" situations.)

crates/templates/src/manager.rs
```diff
@@ -121,6 +123,7 @@ impl TemplateManager {
         source: &TemplateSource,
         options: &InstallOptions,
         reporter: &impl ProgressReporter,
+        install_template_list: &Vec<String>,
```
itowlson, line 126: Is there a reason for this to be a separate parameter rather than part of `InstallIOptions`?

src/commands/templates.rs
```diff
@@ -81,6 +81,11 @@ pub struct Install {
     /// If present, updates existing templates instead of skipping.
     #[clap(long = "upgrade", alias = "update")]
     pub update: bool,
+
+    /// By default, Spin upgrades all the templates in the source.
+    /// Pass this flag with a list of template IDs to update specific ones.
+    #[clap(long, value_delimiter = ',')]
```
itowlson, line 87: I don't love the comma separator - I think we usually use allow-multiple args for this sort of thing?  (But I may be wrong!)

src/commands/templates.rs
```diff
@@ -181,18 +186,26 @@ impl Install {
...
+            .map(|s| s.id().to_owned())
+            .collect();
         if self.git.is_some() {
             // This is equivalent to `install --update`
             let install = Install {
                 git: self.git.clone(),
                 branch: self.branch.clone(),
                 dir: None,
                 update: true,
+                id: installed,
```
itowlson, line 204: Doesn't this mean that if we add a new template to the Spin repo, existing users will never get it?  That seems regrettable.

src/commands/templates.rs
```diff
@@ -570,6 +583,7 @@ fn skipped_reason_text(reason: &SkippedReason) -> String {
     match reason {
         SkippedReason::AlreadyExists => "Already exists".to_owned(),
         SkippedReason::InvalidManifest(msg) => format!("Template load error: {}", msg),
+        SkippedReason::IdNotIncluded => "ID not included".to_owned(),
```
itowlson, line 586: I'm a bit torn on this.  Part of me wants to not print it because it wasn't within the scope of what the user asked to install.  But another part of me notes that if the user typoes an ID then it will help them find what it was meant to be.

In any event if we do print it I'd suggest a message such as "Not requested" that reflects user intent (I just want the C template) rather than the implementation (I passed a list of IDs).

crates/templates/src/manager.rs
itowlson, line 194: The problem with a pure filtering approach is error handling.  If I write `install --id http-rsut` it will cheerily report "No templates installed" and the user is going "but where is my Rust template."  The user is unlikely to think of their command as filtering: they think they have _asked for_ a specific template.  So if we can't find the template they asked for, that feels like we should be telling them "hey we didn't find http-rsut" and maybe even "but may we interest you in this charming http-rust, very popular, very in right now."

itowlson (changes requested):

crates/templates/src/manager.rs
karthik2804: Yep - that is a typo. will fix.

crates/templates/src/manager.rs
karthik2804: I agree and will remove it.

crates/templates/src/manager.rs
karthik2804: Good point. Will make the changes required.

crates/templates/src/manager.rs
karthik2804: Nothing more than me just hacking something quickly, I will move it as a part of the `InstallOptions`.

src/commands/templates.rs
karthik2804: I do not have a really strong preference but having to type `--id template-a --id template-b` seemed a bit more verbose but that does seem more semantically correct with it being `id` (singular)

src/commands/templates.rs
karthik2804: Yes but the issue that I see with the current approach is that, If I install all the templates from Spin repo and then decide I do not want to write zig or c code and delete the templates and then upgrade, I end up getting all the templates installed again.

src/commands/templates.rs
karthik2804: I think we can have this table instead of the reporter messages with the reason "Not requested" as suggested.

crates/templates/src/manager.rs
karthik2804: I agree that the preferred approach should be to let them know if we did not find a template that they asked for.

src/commands/templates.rs
itowlson: Yes but deleting C doesn't mean I want to miss out on Befunge when I've been waiting so long for it.  Worse still, with this, even if I _never deleted anything_ I will _still_ never get Befunge, because this implementation _locks me in_ to my current set of templates.

I'm fully in favour of respecting deletes or "I only ever asked for http-rust" across an upgrade - but we must have an answer for that _first._  We need to write down the desired (and undesired) behaviours, and then _design_ something that will behave correctly, and I fear that will have to be rather more complicated and nuanced than this simple initial approach (which I realise is a draft for feedback, but this is the feedback _grin_).

karthik2804: I agree with this needing discussion. My take on this is that `upgrade` only upgrades what I have. If I need something new that I have never had I would have to `install` first. I can definitely be convinced otherwise. 

I am happy to remove the changes to the `upgrade`  part and leave the behavior as it is today if that is a preferred approach.

itowlson: Yes, the trouble is that historically what I have had is a _repo_ rather than a set of individual templates.  (For better or for worse!) 

As for what to do in this PR... the thing is, I agree that if someone has said "I only want http-rust" then an upgrade should only give them http-rust.  The current behaviour is definitely not the preferred approach!  So I am not sure that we can punt on this unfortunately...

karthik2804: I have given it a little more thought and to me, the more I am convinced `upgrade` should not bring in anything new. Even though it has historically done so, I think using the command `spin templates install --update ...` is what I seem to recall in most places so this should not have too much impact potentially (?).

karthik2804:
I am closing this as it will require more work to be put into the redesign of the current template system.

PR #2240, opened 2024-01-23
cargo cfg: fix build on aarch64-linux

endocrimes:
Fixes spin builds on aarch64 linux (we set this via ENV for ci release builds, but that is unhelpful when building spin as a developer on arm)


rylev (approved):
I think this allows us to remove this flag from being set in CI too, but that can be done as a follow up.

PR #2242, opened 2024-01-24
allow redis trigger to connect to multiple servers

karthik2804:
Adds the ability to connect to multiple Redis servers by optionally specifying an address on the individual trigger config.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6488,6 +6488,7 @@ dependencies = [
  "spin-testing",
  "spin-trigger",
  "spin-world",
+ "tokio",
  "tracing",
 ]
 
--- a/crates/trigger-redis/Cargo.toml
+++ b/crates/trigger-redis/Cargo.toml
@@ -19,6 +19,7 @@ spin-trigger = { path = "../trigger" }
 spin-world = { path = "../world" }
 redis = { version = "0.21", features = ["tokio-comp"] }
 tracing = { workspace = true }
+tokio = { version = "1.23", features = ["full"] }
 
 [dev-dependencies]
 spin-testing = { path = "../testing" }
--- a/crates/trigger-redis/src/lib.rs
+++ b/crates/trigger-redis/src/lib.rs
@@ -81,20 +81,15 @@ impl TriggerExecutor for RedisTrigger {
 
     /// Run the Redis trigger indefinitely.
     async fn run(self, _config: Self::RunConfig) -> Result<()> {
-        let addresses: Vec<String> = self
+        let tasks: Vec<_> = self
             .server_channels
-            .keys()
-            .map(|key| key.to_owned())
-            .collect();
-        let tasks: Vec<_> = addresses
+            .clone()
             .into_iter()
-            .map(|address| {
+            .map(|(server_address, channel_components)| {
                 let trigger = self.clone();
-                // safe to unwrap hashmap as only address that exist are passed
-                let channel_components = self.server_channels.get(&address).unwrap().clone();
                 tokio::spawn(async move {
                     trigger
-                        .run_listener(address.clone(), channel_components)
+                        .run_listener(server_address.clone(), channel_components.clone())
                         .await
                 })
             })
--- a/crates/trigger-redis/src/lib.rs
+++ b/crates/trigger-redis/src/lib.rs
@@ -2,27 +2,29 @@
 
 mod spin;
 
-use std::collections::HashMap;
-
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use futures::{future::join_all, StreamExt};
 use redis::{Client, ConnectionLike};
 use serde::{de::IgnoredAny, Deserialize, Serialize};
 use spin_core::async_trait;
 use spin_trigger::{cli::NoArgs, TriggerAppEngine, TriggerExecutor};
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
 
 use crate::spin::SpinRedisExecutor;
 
 pub(crate) type RuntimeData = ();
 pub(crate) type Store = spin_core::Store<RuntimeData>;
 
+type ChannelComponents = HashMap<String, Vec<String>>;
 /// The Spin Redis trigger.
+#[derive(Clone)]
 pub struct RedisTrigger {
-    engine: TriggerAppEngine<Self>,
-    // Redis address to connect to
-    address: String,
-    // Mapping of subscription channels to component IDs
-    channel_components: HashMap<String, Vec<String>>,
+    engine: Arc<TriggerAppEngine<Self>>,
+    // List of Redis address to connect to
+    addresses: Vec<String>,
+    // Mapping of server with subscription channel and associated component IDs
+    server_channels: HashMap<String, ChannelComponents>,
 }
 
 /// Redis trigger configuration.
@@ -33,6 +35,8 @@ pub struct RedisTriggerConfig {
     pub component: String,
     /// Channel to subscribe to
     pub channel: String,
+    /// optional overide address for trigger
+    pub address: Option<String>,
     /// Trigger executor (currently unused)
     #[serde(default, skip_serializing)]
     pub executor: IgnoredAny,
@@ -52,32 +56,80 @@ impl TriggerExecutor for RedisTrigger {
     type RunConfig = NoArgs;
 
     async fn new(engine: TriggerAppEngine<Self>) -> Result<Self> {
-        let address = engine
+        let default_address: String = engine
             .trigger_metadata::<TriggerMetadata>()?
             .unwrap_or_default()
             .address;
-        let address_expr = spin_expressions::Template::new(address)?;
-        let address = engine.resolve_template(&address_expr)?;
+        let default_address_expr = spin_expressions::Template::new(default_address)?;
+        let default_address = engine.resolve_template(&default_address_expr)?;
 
-        let mut channel_components: HashMap<String, Vec<String>> = HashMap::new();
+        let mut addresses = HashSet::new();
+
+        let mut server_channels: HashMap<String, ChannelComponents> = HashMap::new();
 
         for (_, config) in engine.trigger_configs() {
-            channel_components
+            let address = config.address.clone().unwrap_or(default_address.clone());
+            let address_expr = spin_expressions::Template::new(address)?;
+            let address = engine.resolve_template(&address_expr)?;
+            addresses.insert(address.clone());
+            let server = server_channels.entry(address).or_default();
+            server
                 .entry(config.channel.clone())
                 .or_default()
                 .push(config.component.clone());
         }
         Ok(Self {
-            engine,
-            address,
-            channel_components,
+            engine: Arc::new(engine),
+            addresses: addresses.into_iter().collect(),
+            server_channels,
         })
     }
 
     /// Run the Redis trigger indefinitely.
     async fn run(self, _config: Self::RunConfig) -> Result<()> {
-        let address = &self.address;
+        let addresses = self.addresses.clone();
+        let tasks: Vec<_> = addresses
+            .into_iter()
+            .map(|address| {
+                let trigger = self.clone();
+                tokio::spawn(async move { trigger.run_listener(address).await })
+            })
+            .collect();
 
+        let _ = join_all(tasks).await;
+
+        Ok(())
+    }
+}
+
+impl RedisTrigger {
+    // Handle the message.
+    async fn handle(&self, address: String, msg: redis::Msg) -> Result<()> {
+        let channel = msg.get_channel_name();
+        tracing::info!("Received message on channel {:?}", channel);
+
+        if let Some(channels) = self.server_channels.get(&address) {
+            if let Some(component_ids) = channels.get(channel) {
+                let futures = component_ids.iter().map(|id| {
+                    tracing::trace!("Executing Redis component {id:?}");
+                    SpinRedisExecutor.execute(&self.engine, id, channel, msg.get_payload_bytes())
+                });
+                let results: Vec<_> = join_all(futures).await.into_iter().collect();
+                let errors = results
+                    .into_iter()
+                    .filter_map(|r| r.err())
+                    .collect::<Vec<_>>();
+                if !errors.is_empty() {
+                    return Err(anyhow!("{errors:#?}"));
+                }
+            } else {
+                tracing::debug!("No subscription found for {:?}", channel);
+            }
+        }
+        Ok(())
+    }
+
+    async fn run_listener(&self, address: String) -> Result<()> {
         tracing::info!("Connecting to Redis server at {}", address);
         let mut client = Client::open(address.to_string())?;
         let mut pubsub = client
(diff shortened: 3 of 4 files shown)


crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
+        for address in addresses {
+            let trigger = self.clone();
+            // Spawn an asynchronous task with a closure capturing self by reference
+            println!("Starting listener for {address}");
+            tasks.push(tokio::spawn(
+                async move { trigger.run_listener(address).await },
+            ));
+        }
+
+        let _ = join_all(tasks).await;
```
karthik2804, line 98: I think we should possibly exit when one of the listeners exits. thoughts?

itowlson: Either exit so that the alarm goes off, or try to restart the relevant listener.  But yeah, I don't think we want it continuing in a partial state.

(For what it's worth I think the SQS trigger exits, but that may be because I am lazy.)

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
         let mut channel_components: HashMap<String, Vec<String>> = HashMap::new();
 
         for (_, config) in engine.trigger_configs() {
+            let address = config.address.clone().unwrap_or(default_address.clone());
+            if !addresses.contains(&address) {
+                addresses.push(address.clone());
+            }
             channel_components
-                .entry(config.channel.clone())
+                .entry(format!("{}:{}", address, config.channel.clone()))
```
itowlson, line 73: if only Rust had a feature for structured compound data but alas no

crates/redis/src/lib.rs
itowlson: Looking at the SQS trigger `futures::future::select_all` seems to be what I thought would work

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
     /// Run the Redis trigger indefinitely.
     async fn run(self, _config: Self::RunConfig) -> Result<()> {
-        let address = &self.address;
+        let addresses = self.addresses.clone();
+        let mut tasks = Vec::new();
+        // Spawn an asynchronous task for each iteration
+        for address in addresses {
+            let trigger = self.clone();
+            // Spawn an asynchronous task with a closure capturing self by reference
+            println!("Starting listener for {address}");
```
itowlson, line 92: Remove

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
 
     /// Run the Redis trigger indefinitely.
     async fn run(self, _config: Self::RunConfig) -> Result<()> {
-        let address = &self.address;
+        let addresses = self.addresses.clone();
+        let mut tasks = Vec::new();
+        // Spawn an asynchronous task for each iteration
+        for address in addresses {
+            let trigger = self.clone();
+            // Spawn an asynchronous task with a closure capturing self by reference
```
itowlson, line 91: Not sure this comment is terribly helpful?  It seems to repeat the code rather than explaining it

itowlson: I also think it might be wrong.  Given that you clone and move my guess is that a clone is being moved into the closure.

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
             channel_components,
         })
     }
 
     /// Run the Redis trigger indefinitely.
     async fn run(self, _config: Self::RunConfig) -> Result<()> {
-        let address = &self.address;
+        let addresses = self.addresses.clone();
+        let mut tasks = Vec::new();
+        // Spawn an asynchronous task for each iteration
```
itowlson, line 88: This comment seems to repeat the code rather than explain it

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
+        for address in addresses {
+            let trigger = self.clone();
+            // Spawn an asynchronous task with a closure capturing self by reference
+            println!("Starting listener for {address}");
+            tasks.push(tokio::spawn(
```
itowlson, line 93: This is a style thing, but throughout this PR you show a strong preference for loops pushing onto mutable vectors, rather than using iterator functions - is that deliberate?  It's not wrong, and at this scale mostly a style thing, I'm just curious.

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
+            if !errors.is_empty() {
+                return Err(anyhow!("{errors:#?}"));
+            }
+        } else {
+            tracing::debug!("No subscription found for {:?}", channel);
+        }
+
+        Ok(())
+    }
+    async fn run_listener(&self, address: String) -> Result<()> {
```
itowlson, line 132: blank line betwwen functions

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
+
+impl RedisTrigger {
+    // Handle the message.
+    async fn handle(&self, address: String, msg: redis::Msg) -> Result<()> {
+        let channel = msg.get_channel_name();
+        tracing::info!("Received message on channel {:?}", channel);
+
+        if let Some(component_ids) = self
+            .channel_components
+            .get(&format!("{}:{}", address, channel))
```
itowlson, line 112: If you do it this way then have a helper function so that the format doesn't have to be maintained in multiple places

crates/redis/src/lib.rs
```diff
@@ -87,16 +141,19 @@ impl TriggerExecutor for RedisTrigger {
         println!("Active Channels on {address}:");
         // Subscribe to channels
         for (channel, component) in self.channel_components.iter() {
-            tracing::info!("Subscribing component {component:?} to channel {channel:?}");
-            pubsub.subscribe(channel).await?;
-            println!("\t{channel}: [{}]", component.join(","));
+            if channel.contains(&address) {
+                let channel_name = channel.strip_prefix(&format!("{address}:"));
```
itowlson, line 145: I'm sorry but this stringing feels worse and worse.  Please use a tuple or struct to represent structured data.

crates/redis/src/lib.rs
```diff
@@ -87,16 +141,19 @@ impl TriggerExecutor for RedisTrigger {
...
+                tracing::info!("Subscribing component {component:?} to channel {channel_name:?}");
+                pubsub.subscribe(channel_name).await?;
+                println!("\t{channel}: [{}]", component.join(","));
```
itowlson, line 148: the channel may be ambiguous - consider disamiguating if so (or if override? not sure what's best)

itowlson (changes requested):

crates/redis/src/lib.rs
```diff
@@ -52,30 +55,81 @@ impl TriggerExecutor for RedisTrigger {
...
             .trigger_metadata::<TriggerMetadata>()?
             .unwrap_or_default()
             .address;
 
+        let mut addresses = vec![default_address.clone()];
+
         let mut channel_components: HashMap<String, Vec<String>> = HashMap::new();
 
         for (_, config) in engine.trigger_configs() {
+            let address = config.address.clone().unwrap_or(default_address.clone());
```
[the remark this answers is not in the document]
karthik2804: I am not sure I follow this as the address will either be the optional override or the default provided at the application level. I apologize if I am missing something obvious here.

crates/redis/src/lib.rs
karthik2804: Just a force of habit. Will update to use an iterator.

crates/redis/src/lib.rs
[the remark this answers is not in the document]
itowlson: This comment referred to the original code, which kept cloning the default address in order to check it against the vector _which was already initialised with the default address_.  There was no need to take any action unless there was an override, because the default address was already in the vector.  Hence the suggestion to `if let Some(address) =` and do the check-and-insert only if there was an override.

The current code does depend on inserting the default address during the loop, so the comment no longer applies.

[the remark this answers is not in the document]
karthik2804: Ah, that makes sense!

crates/redis/src/lib.rs
karthik2804: This should be fixed now as it prints the `{server_address}:{channel}`

crates/redis/src/lib.rs
```diff
@@ -2,27 +2,29 @@
...
 pub struct RedisTrigger {
-    engine: TriggerAppEngine<Self>,
-    // Redis address to connect to
-    address: String,
-    // Mapping of subscription channels to component IDs
-    channel_components: HashMap<String, Vec<String>>,
+    engine: Arc<TriggerAppEngine<Self>>,
+    // List of Redis address to connect to
+    addresses: Vec<String>,
+    // Mapping of server with subscription channel and associated component IDs
```
itowlson, line 26: I believe `address` and `server` refer to the same thing - could this be confusing?  Is there a term we can helpfully unify on?

crates/redis/src/lib.rs
```diff
@@ -2,27 +2,29 @@
...
 /// The Spin Redis trigger.
+#[derive(Clone)]
 pub struct RedisTrigger {
-    engine: TriggerAppEngine<Self>,
-    // Redis address to connect to
-    address: String,
-    // Mapping of subscription channels to component IDs
-    channel_components: HashMap<String, Vec<String>>,
+    engine: Arc<TriggerAppEngine<Self>>,
+    // List of Redis address to connect to
```
itowlson, line 24: meganit 
```suggestion
    // List of Redis addresses to connect to
```

crates/redis/src/lib.rs
```diff
@@ -52,30 +56,78 @@ impl TriggerExecutor for RedisTrigger {
...
                 .entry(config.channel.clone())
                 .or_default()
                 .push(config.component.clone());
         }
         Ok(Self {
-            engine,
-            address,
-            channel_components,
+            engine: Arc::new(engine),
+            addresses: addresses.into_iter().collect(),
```
itowlson, line 79: Is it necessary / useful to convert them to a vector?  Would it cause any problem to leave them as a HashSet?

crates/redis/src/lib.rs
```diff
@@ -86,56 +138,34 @@ impl TriggerExecutor for RedisTrigger {
 
         println!("Active Channels on {address}:");
```
itowlson, line 139: This will print multiple instances of the Active Channels message now I think?

crates/redis/src/lib.rs
```diff
@@ -86,56 +138,34 @@ impl TriggerExecutor for RedisTrigger {
...
-                return Err(anyhow!("{errors:#?}"));
+                };
             }
-        } else {
-            tracing::debug!("No subscription found for {:?}", channel);
+            return Ok(());
         }
-
-        Ok(())
+        bail!("Server does not have any channels that need to be subscribed to")
```
itowlson, line 168: This should never happen.  If it does it means your `addresses` and `server_channels` collections are out of sync.  Is it possible to design your data structure so that this cannot happen?  For example, in the `server_channels` collection, could you use the keys of that collection as your list of addresses, instead of maintaining a separate collection?

crates/redis/src/lib.rs
```diff
@@ -52,30 +56,78 @@ impl TriggerExecutor for RedisTrigger {
...
+    }
+}
+
+impl RedisTrigger {
+    // Handle the message.
+    async fn handle(&self, address: String, msg: redis::Msg) -> Result<()> {
+        let channel = msg.get_channel_name();
+        tracing::info!("Received message on channel {:?}", channel);
+
+        if let Some(channels) = self.server_channels.get(&address) {
```
itowlson, line 109: This shouldn't change on each message - there is no need to re-look-up the address each time.

There's some things about the code here that make me wonder if there's a listener type (one instance per server, holding that server's channel map, running that server's loop, essentially pulling out `run_listener` to hold its own state) trying to get out.  That may be more of a change than you fancy getting into though!

crates/redis/src/lib.rs
```diff
@@ -52,30 +56,78 @@ impl TriggerExecutor for RedisTrigger {
...
+        println!("exiting");
+        Ok(())
+    }
+}
+
+impl RedisTrigger {
+    // Handle the message.
+    async fn handle(&self, address: String, msg: redis::Msg) -> Result<()> {
+        let channel = msg.get_channel_name();
+        tracing::info!("Received message on channel {:?}", channel);
```
itowlson, line 107: Should this include the address?

crates/redis/src/lib.rs
```diff
@@ -52,30 +56,78 @@ impl TriggerExecutor for RedisTrigger {
...
+            .map(|address| {
+                let trigger = self.clone();
+                tokio::spawn(async move { trigger.run_listener(address).await })
+            })
+            .collect();
+
+        // wait for the first handle to be returned and drop the rest
+        let (_, _, rest) = futures::future::select_all(tasks).await;
+        drop(rest);
+        println!("exiting");
```
itowlson, line 98: Is this needed?  Do other triggers print it?

itowlson (commented):
I don't think there are any blockers here, but I feel a little uneasy about some design details (e.g. two lists of addresses, terminology) and there are a couple of wee tidying things.  Have a look at the comments and if you go "Ivan you are talking nonsense because" then I am probably okay to approve.

crates/redis/src/lib.rs
karthik2804: server url maybe?

crates/redis/src/lib.rs
karthik2804: Yes, that makes sense.

crates/redis/src/lib.rs
karthik2804: I am not sure I follow. Are you suggesting that each listener only get the only the hashmap (channel, components) instead of the double hashmap that it currently gets?

itowlson: Yes.

crates/redis/src/lib.rs
karthik2804: Other triggers do not. I will remove it.

crates/redis/src/lib.rs
karthik2804: done

crates/redis/src/lib.rs
```diff
@@ -52,30 +54,91 @@ impl TriggerExecutor for RedisTrigger {
...
+        let addresses: Vec<String> = self
+            .server_channels
+            .keys()
+            .map(|key| key.to_owned())
+            .collect();
+        let tasks: Vec<_> = addresses
+            .into_iter()
+            .map(|address| {
+                let trigger = self.clone();
+                // safe to unwrap hashmap as only address that exist are passed
```
itowlson, line 89: Rather than doing this, consider iterating the HashMap, which will give you a tuple of the address with its channel map.  No need to extract the address then later go back and look up the channel map.  A la

```
let tasks = self.server_channels.into_iter().map(|server, channel_components| ...)
```

itowlson (approved):
LGTM - thanks for your patience with all the niggling

karthik2804:
cc: @lann for viz

lann (commented):
This is kinda weird but it started out kinda weird so :shrug:

itowlson:
What is required to land this @karthik2804?  (Apart from a rebase.)  Does something need to be done with Lann's comments?  (I'm not sure whether "kinda weird" is saying it needs a rework?)

karthik2804:
@lann can you clarify?

PR #2247, opened 2024-01-25
Move wasi-http integration tests over to new testing framework

rylev:
This moves the wasi-http tests over to the new testing framework. The consequences of this PR are that: 
* We can delete a bunch of helper code that the testing framework now does
* It makes these wasi tests work like all other tests.
* It moves building code out of build.rs and into the test suite which only gets run when tests are run
* The examples can either be deleted or simplified since they don't have to play double duty.
* The testing framework now has more support for streaming bodies

To see how these tests now look, make sure not to miss [this](https://github.com/fermyon/spin/pull/2247/files#diff-66b1ad6fa2ee7241ab510f68f5cd9e98455d9f8c8a99aaedb99377e60b6cc9cdR877-R1055).


dicej (approved):
LGTM.  Thanks, as always, for working on this.

I'm wondering if we should remove the `wasi-http-rust-streaming-outgoing-body` example entirely given the overlap with the integration test and the fact that it's no longer being tested (so might regress without us noticing).  My favorite kind of examples are also tests, and it seems like `integration-wasi-http-streaming` fits in that category, making the old example redundant.

rylev:
> I'm wondering if we should remove the wasi-http-rust-streaming-outgoing-body example entirely given the overlap with the integration test and the fact that it's no longer being tested (so might regress without us noticing). My favorite kind of examples are also tests, and it seems like integration-wasi-http-streaming fits in that category, making the old example redundant.

I see where you're coming from, but I'm not sure I fully agree. My thinking is that examples are there to be illustrative while tests are there to ensure that Spin functions correctly. These concerns can and sometimes do overlap, but not always, and we shouldn't force them to. What's best for teaching devs how to use Spin is not always going to be best for ensuring Spin continues to function.

With that in mind, the example has been greatly simplified and now only shows one of the previously 3 endpoints. The hope behind is that this can be a good guide for how to do streaming bodies in Spin. The argument that we need to run tests against examples is something I'm not really sure about. We should ensure that the examples continue to build (which we do), but I'm not sure it's super necessary to test their functionality. We currently don't run tests against any other examples so the wasi-http example is hardly the outlier here. 

With all of this said, I would like to soon revisit the wasi-http tests and potentially rewrite them entirely. There are large aspects of wasi-http (and how it interacts specifically with Spin) that we're not currently testing, and I'd like to change that. Perhaps we can revisit this question when we do that work.

PR #2248, opened 2024-01-26
Cancel external triggers if one of multiple fails on start

### LINKED ISSUE #2245: In multi-trigger, if one trigger crashes, others may keep running
I created a multi-trigger app with http, timer and sqs triggers.  I knew that the SQS trigger was not yet compatible with multi-trigger.  The SQS trigger obligingly crashed during startup, but the timer trigger continued to run.

I believe this is due to the timer trigger being a plugin.  To run the trigger, we start a `spin` process, which in turn starts a `timer-trigger` process.  I believe that when we kill that `spin` process, the child `timer-trigger` process continues to run.

Further evidence for this is from logging the PIDs of the trigger processes.  None of these PIDs matched the ID from doing `ps -A | grep timer` after SQS had crashed.

This may be a bit gnarly to fix because when we kill the Spin process it just dies, it doesn't get to say "goodnight sweet prince" or (more relevantly) "die, actual plugin process, die" - we can try setting `kill_on_drop` on the child but I am not sure if finalisers will run in this case.  Anyway I will take a look at it and see if I can make a dent.

cc @kate-goldenring if you have any thoughts on this!

itowlson:
Fixes #2245.

There may still be a nasty if one of the child triggers fails to launch, or if a trigger unexpectedly exits _without_ reporting an error.  But this does fix, albeit uglily, the "typical" case of a trigger failing (and specifically a non-multi-aware trigger crashing during startup).

I'm concerned about using timers here, because they're obviously quite fragile; but I can't think of another way to get at "all `spin trigger-foo` process have had the chance to start their child `trigger-foo` processes and hook them up to Ctrl+C."  Open to ideas!

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -8,7 +8,6 @@ use std::{
 
 use anyhow::{anyhow, bail, Context, Result};
 use clap::{CommandFactory, Parser};
-use itertools::Itertools;
 use reqwest::Url;
 use spin_app::locked::LockedApp;
 use spin_common::ui::quoted_path;
@@ -17,13 +16,22 @@ use spin_oci::OciLoader;
 use spin_trigger::cli::{SPIN_LOCAL_APP_DIR, SPIN_LOCKED_URL, SPIN_WORKING_DIR};
 use tempfile::TempDir;
 
-use futures::StreamExt;
-
 use crate::opts::*;
 
 use self::app_source::{AppSource, ResolvedAppSource};
 
 const APPLICATION_OPT: &str = "APPLICATION";
+
+// If multiple triggers start very close together, there is a race condition
+// where if one trigger fails during startup, other external triggers may
+// not have their cancellation hooked up.  (kill_on_drop doesn't fully solve
+// this because it kills the Spin process but that doesn't cascade to the
+// child plugin trigger process.) So add a hopefully insignificant delay
+// between them to reduce the chance of this happening.
+const MULTI_TRIGGER_START_OFFSET: tokio::time::Duration = tokio::time::Duration::from_millis(20);
+// And just in case wait a few moments before performing the first "have
+// any exited" check.
+const MULTI_TRIGGER_LET_ALL_START: tokio::time::Duration = tokio::time::Duration::from_millis(500);
 
 /// Start the Fermyon runtime.
 #[derive(Parser, Debug, Default)]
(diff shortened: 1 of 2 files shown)


src/commands/up.rs
```diff
@@ -387,25 +413,46 @@ impl UpCommand {
...
+    let pids = pids.to_owned();
+    ctrlc::set_handler(move || {
+        kill_them_all_god_will_know_his_own(&pids);
+    })?;
+    Ok(())
+}
+
+#[cfg(windows)]
+fn get_pids(_trigger_processes: &[tokio::process::Child]) -> Vec<usize> {
+    vec![]
```
rylev, line 431: I'm confused why this would be correct.

rylev (approved):
ooof. It certainly isn't pretty. I think it's fine for now. I wonder if there's some sort of supervisor tree architecture we can adopt that makes some of this book keeping easier, but that's really hard when you have to jump the process boundary... Hmm....

src/commands/up.rs
```diff
@@ -231,6 +250,13 @@ impl UpCommand {
                 .await
                 .context("Failed to start trigger process")?;
             trigger_processes.push(child);
+
+            if is_multi {
+                // Allow time for the child `spin` process to launch the trigger
+                // and hook up its cancellation. Mitigates the race condition
+                // noted on the constant (see there for more info).
+                tokio::time::sleep(MULTI_TRIGGER_START_OFFSET).await;
+            }
```
lann, line 259: It doesn't seem like this should be necessary both here and above?

src/commands/up.rs
```diff
@@ -387,25 +413,46 @@ impl UpCommand {
...
+#[cfg(windows)]
+fn kill_them_all_god_will_know_his_own(_pids: &[usize]) {}
+
+#[cfg(not(windows))]
+fn kill_them_all_god_will_know_his_own(pids: &[nix::unistd::Pid]) {
+    // https://github.com/nix-rust/nix/issues/656
+    for pid in pids {
+        // println!("killing {pid}");
+        if let Err(err) = nix::sys::signal::kill(*pid, nix::sys::signal::SIGTERM) {
+            tracing::warn!("Failed to kill trigger handler process: {:?}", err)
```
lann, line 453: I think this could be simplified a bit by passing the `u32`s from `Child::id` into `kill_...` and doing the cast to `Pid`s in there.

src/commands/up.rs
itowlson: _chuckle_ It's not "correct".  This PR tracks the original Ctrl+C propagation code, which was cfg-ed out on Windows (because the signals stuff doesn't have an equivalent); so although I need to have a function here (and could indeed get OS-level process handles) I can't do anything meaningful with them, at least not without a lot more research than "preserve what the old code did".

I can look at implementing Ctrl+C propagation on Windows for sure.  But I'd be inclined to look at a slightly different design for that, with processes starting suspended and being placed in a job object.  That's more than I want to bite off in this PR though...!

PR #2249, opened 2024-01-26
automatically set content-length on appropriate outbound requests

### LINKED ISSUE #2246: Outbound Requests don't automatically set the `Content-Length` header
It's fairly common practice to automatically set the `Content-Length` header on behalf of the user. We should be doing this for outbound requests where the size of the request is known.

karthik2804:
I tried basing this PR on the comments in #2246 but I am not sure. let me know! 

fixes #2246

### DIFF
--- a/crates/outbound-http/src/host_impl.rs
+++ b/crates/outbound-http/src/host_impl.rs
@@ -1,5 +1,8 @@
 use anyhow::Result;
-use http::HeaderMap;
+use http::{
+    header::{CONTENT_LENGTH, TRANSFER_ENCODING},
+    HeaderMap,
+};
 use reqwest::Client;
 use spin_core::async_trait;
 use spin_outbound_networking::{AllowedHostsConfig, OutboundUrl};
@@ -67,8 +70,8 @@ impl outbound_http::Host for OutboundHttp {
             let body = req.body.unwrap_or_default().to_vec();
 
             // if no content length, set it
-            if headers.get("content-length").is_none() && !body.is_empty() && headers.get("transfer-encoding").is_none() {
-                headers.append("content-length", body.len().into());
+            if !body.is_empty() && !headers.contains_key(CONTENT_LENGTH) && !headers.contains_key(TRANSFER_ENCODING){
+                headers.insert("content-length", body.len().into());
             }
 
             if !req.params.is_empty() {
--- a/crates/outbound-http/src/host_impl.rs
+++ b/crates/outbound-http/src/host_impl.rs
@@ -63,8 +63,13 @@ impl outbound_http::Host for OutboundHttp {
 
             let req_url = reqwest::Url::parse(&abs_url).map_err(|_| HttpError::InvalidUrl)?;
 
-            let headers = request_headers(req.headers).map_err(|_| HttpError::RuntimeError)?;
+            let mut headers = request_headers(req.headers).map_err(|_| HttpError::RuntimeError)?;
             let body = req.body.unwrap_or_default().to_vec();
+
+            // if no content length, set it
+            if headers.get("content-length").is_none() && !body.is_empty() && headers.get("transfer-encoding").is_none() {
+                headers.append("content-length", body.len().into());
+            }
 
             if !req.params.is_empty() {
                 tracing::log::warn!("HTTP params field is deprecated");
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -14,10 +14,14 @@ use std::{
 use anyhow::{Context, Result};
 use async_trait::async_trait;
 use clap::Args;
-use http::{uri::Scheme, StatusCode, Uri};
+use http::{
+    header::{CONTENT_LENGTH, TRANSFER_ENCODING},
+    uri::Scheme,
+    StatusCode, Uri,
+};
 use http_body_util::BodyExt;
 use hyper::{
-    body::{Bytes, Incoming},
+    body::{Body as _, Bytes, Incoming},
     server::conn::http1,
     service::service_fn,
     Request, Response,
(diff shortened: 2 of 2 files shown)


rylev (approved):
This looks correct but is incomplete. This only impacts the old http interface but doesn't impact wasi-http. We would also need to update the wasi-http interface which we hook into [here](https://github.com/fermyon/spin/blob/93d22a9e60f7f13fd49b44a93391673386e1b501/crates/trigger-http/src/lib.rs#L463).

karthik2804:
I will take a look at the wasi-http interface as well.

karthik2804:
@lann it would be good to get your eyes on this as well.

tschneidereit:
I agree that this is a great change, thank you for working on it, @karthik2804! For the wasi-http side, it'd be great if we could make the required change upstream in Wasmtime, so Spin would just inherit it.

crates/outbound-http/src/host_impl.rs
```diff
@@ -63,9 +63,14 @@ impl outbound_http::Host for OutboundHttp {
...
             let req_url = reqwest::Url::parse(&abs_url).map_err(|_| HttpError::InvalidUrl)?;
 
-            let headers = request_headers(req.headers).map_err(|_| HttpError::RuntimeError)?;
+            let mut headers = request_headers(req.headers).map_err(|_| HttpError::RuntimeError)?;
             let body = req.body.unwrap_or_default().to_vec();
 
+            // if no content length, set it
+            if headers.get("content-length").is_none() && !body.is_empty() && headers.get("transfer-encoding").is_none() {
+                headers.append("content-length", body.len().into());
+            }
```
lann, line 72: Constants prevent typos:

```rust
use http::header::{CONTENT_LENGTH, TRANSFER_ENCODING};
```

`contains_key` is a little more idiomatic/comprehensible:
```suggestion
            // set content length if necessary
            if !body.is_empty()
                 && !headers.contains_key(CONTENT_LENGTH)
                 && !headers.contains_key(TRANSFER_ENCODING)
             {
                 headers.insert(CONTENT_LENGTH, body.len().into());
             }
  
```

crates/trigger-http/src/lib.rs
```diff
@@ -535,6 +535,16 @@ impl OutboundWasiHttpHandler for HttpRuntimeData {
...
 
+        let body_size_hint = request.request.body().size_hint().exact();
+        let headers = request.request.headers_mut();
+
+        if let Some(size) = body_size_hint {
+            if headers.get("content-legnth").is_none() && headers.get("transfer-encoding").is_none()
+            {
+                headers.insert("content-length", size.into());
+            }
+        }
```
lann, line 546: ...typos like this :slightly_smiling_face: 
> `"content-legnth"`

```suggestion
        if let Some(size) = request.request.body().size_hint().exact() {
            let headers = request.request.headers_mut();
            if size > 0
                && headers.contains_key(CONTENT_LENGTH)
                && headers.contains_key(TRANSFER_ENCODING)
            {
                headers.insert(CONTENT_LENGTH, size.into());
            }
        }
```

(not sure about `size > 0` but seems right?)

lann (changes requested):
`"content-legnth"` typo is only real blocker

karthik2804:
Thanks for catching those @lann .

Do we still want to include the part for header checks in the wasi-http case until we can upstream the changes?

lann (approved):
> Do we still want to include the part for header checks in the wasi-http case until we can upstream the changes?

No strong opinion. Maybe at least open an issue upstream and link to it here?

itowlson:
What is required to land this @karthik2804?  Is there a decision needed somewhere?

itowlson:
@karthik2804 Who can make that decision?  @tschneidereit is it you? or Alex?  (Sorry Karthik, I know I'm nagging, I just want it not to stall, as I'm sure do you!)

karthik2804:
Definitely not me. Probably @tschneidereit would be the best resource on this?

itowlson:
@tschneidereit Can you help us with a path to land this please, or nominate someone else who can make the decision?  Thanks!

crates/outbound-http/src/host_impl.rs
```diff
@@ -63,9 +66,14 @@ impl outbound_http::Host for OutboundHttp {
...
             let req_url = reqwest::Url::parse(&abs_url).map_err(|_| HttpError::InvalidUrl)?;
 
-            let headers = request_headers(req.headers).map_err(|_| HttpError::RuntimeError)?;
+            let mut headers = request_headers(req.headers).map_err(|_| HttpError::RuntimeError)?;
             let body = req.body.unwrap_or_default().to_vec();
 
+            // if no content length, set it
+            if !body.is_empty() && !headers.contains_key(CONTENT_LENGTH) && !headers.contains_key(TRANSFER_ENCODING){
+                headers.insert("content-length", body.len().into());
+            }
```
tschneidereit, line 75: Initially I thought this was wrong because the body size could change for streaming bodies, but then realized that this interface doesn't support those. I think at least this should contain a comment to that effect, but IIUC it could just use the same approach as in the wasi-http implementation below, right?

And I think in both cases, you don't even need to extract the body: you should be able to use `if let Some(size) = req.size_hint().exact()`, which I think should just not match (because `exact()` will return `None` if no body is present, or the final size isn't known.

tschneidereit (commented):
I agree with @karthik2804 that the divergence from Wasmtime is a bit concerning here. More to the point, I think this is something that should ideally be specified by wasi-http, but isn't.

Karthik, would you mind opening an issue on https://github.com/WebAssembly/wasi-http requesting clarification on the expected / allowed behavior here?

karthik2804:
I will open an issue @tschneidereit and then look at the implementation for the `spin-http` interface.

karthik2804:
Here is the link to the issue https://github.com/WebAssembly/wasi-http/issues/105

karthik2804:
The upstream issue got a response 
https://github.com/WebAssembly/wasi-http/issues/105. How do we want to proceed?

tschneidereit:
I think Luke is unfortunately right, and I think even if we land this patch, it'll end up not actually doing anything for wasi-http, because the case where we'd set `Content-Length` will simply never be hit.

As such, it probably makes sense to do this in our SDKs instead, where we often *do* know the final body length. I'll look into supporting this in JS

karthik2804:
I will update this PR to only address the Spin-specific outbound HTTP interface. Or do we not think that causes deviating behavior?

tschneidereit:
I think it's ok to do it on spin-http, but the question is: do we still want to change that?

itowlson:
@tschneidereit The more differences between spin-http and wasi-http behaviour, the more surprises and pain (and documentation effort).  I'd vote to do it everywhere or do it nowhere.  Especially since the hope is that spin-http will fade away as users migrate to wasi-http-based SDKs.  You don't want a relied-upon behaviour to mysteriously disappear because someone bumped to the new SDK.

karthik2804:
I am happy to close this PR is that is the consensus

tschneidereit:
Yeah, I think given the wasi-http situation it's probably best to close this PR and instead focus on applying the header in our SDKs where we can.

Issue #2236, opened 2024-01-20
spin registry push container image metadata


rgl:
At https://github.com/rgl/spin-http-ts-example/tree/test-release I'm building an example container image. When I build a container image with docker build, I can set container image metadata with:

```dockerfile
...
LABEL org.opencontainers.image.source="$HELLO_SOURCE_URL"
LABEL org.opencontainers.image.revision="$HELLO_REVISION"
LABEL org.opencontainers.image.description="Example Spin HTTP Application written in TypeScript"
LABEL org.opencontainers.image.licenses="ISC"
```

And build/push it with:

```bash
docker build \
    --build-arg "HELLO_SOURCE_URL=$HELLO_SOURCE_URL" \
    --build-arg "HELLO_VERSION=$HELLO_VERSION" \
    --build-arg "HELLO_REVISION=$HELLO_REVISION" \
    -t "$IMAGE" \
    .
docker push "$IMAGE"
```

And the image endsup with the expected metadata, like in https://github.com/rgl/spin-http-ts-example/pkgs/container/spin-http-ts-example/169177958?tag=0.1.0-rc3.

But when using `spin registry push` no metadata is sent, like in https://github.com/rgl/spin-http-ts-example/pkgs/container/spin-http-ts-example/169178514?tag=0.1.0-rc5.

So, can we please have a way to set metadata?

itowlson:
Let me see if I can elicit a scope / functional spec here, because I am not sure how closely this needs to follow Docker.

To literally follow Docker we would need:

* A persistent way to define metadata slots with fixed or parameterised values (the equivalent of the LABEL statements), e.g. in `spin.toml`
* A way to set the parameterised values at push time, e.g. arguments to `spin registry push`

But another approach would be:

* A way to define metadata at push time (no persistent slots), e.g. `spin registry push --metadata org.opencontainers.image.license=mylicence`

The latter would be considerably easier; admittedly the UI would be less convenient but maybe this is mostly done from CI where it doesn't matter?

cc @radu-matei @vdice

vdice:
The first approach (new section in spin.toml, support for overriding at time of publish) brings a host of design/UX questions as this would be the first instance of capturing registry/publishing-related data in a build document.  I can imagine some other values that may be handy to store there (or near) like a default registry reference, etc... but all of this is to say that it probably warrants a wider discussion (and perhaps a [SIP](https://github.com/fermyon/spin/blob/main/docs/content/sips/index.md)?)

[...]

PR #2254, opened 2024-01-27
feat: add support for setting the pushed oci image manifest annotations

rgl:
This closes https://github.com/fermyon/spin/issues/2236.

You can see how GitHub Container Registry shows an image without annotations at:

https://github.com/rgl/spin-http-ts-example/pkgs/container/spin-http-ts-example/171707227?tag=0.2.0

And the one with annotations at:

https://github.com/rgl/spin-http-ts-example/pkgs/container/spin-http-ts-example/171719572?tag=0.0.0-test1

This is how I've pushed it the image with my local spin version:

```console
$ echo my-github-token-with-write-packages-scope | docker login ghcr.io -u rgl --password-stdin
$ ~/Projects/spin/target/debug/spin registry push --annotation "org.opencontainers.image.description=$(jq -r .description package.json)" ghcr.io/rgl/spin-http-ts-example:0.0.0-test1
Pushing app to the Registry...
Pushed with digest sha256:60373ae9983dac0356fc91b1ff016f3580fd7664a9061f71d8a5fd266c646a0c
```

### DIFF
--- a/src/commands/registry.rs
+++ b/src/commands/registry.rs
@@ -2,6 +2,7 @@ use crate::opts::*;
 use anyhow::{Context, Result};
 use clap::{Parser, Subcommand};
 use indicatif::{ProgressBar, ProgressStyle};
+use spin_common::arg_parser::parse_kv;
 use spin_oci::Client;
 use std::{io::Read, path::PathBuf, time::Duration};
 
@@ -61,6 +62,11 @@ pub struct Push {
     /// Cache directory for downloaded registry data.
     #[clap(long)]
     pub cache_dir: Option<PathBuf>,
+
+    /// Specifies the OCI image manifest annotations (in key=value format).
+    /// Any existing value will be overwritten. Can be used multiple times.
+    #[clap(long = "annotation", parse(try_from_str = parse_kv))]
+    pub annotations: Vec<(String, String)>,
 }
 
 impl Push {
@@ -70,11 +76,17 @@ impl Push {
             spin_build::build(&app_file, &[]).await?;
         }
 
+        let annotations = if self.annotations.is_empty() {
+            None
+        } else {
+            Some(self.annotations.iter().cloned().collect())
+        };
+
         let mut client = spin_oci::Client::new(self.insecure, self.cache_dir.clone()).await?;
 
         let _spinner = create_dotted_spinner(2000, "Pushing app to the Registry".to_owned());
 
-        let digest = client.push(&app_file, &self.reference).await?;
+        let digest = client.push(&app_file, &self.reference, annotations).await?;
         match digest {
             Some(digest) => println!("Pushed with digest {digest}"),
             None => println!("Pushed; the registry did not return the digest"),
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -1,6 +1,6 @@
 //! Spin's client for distributing applications via OCI registries
 
-use std::collections::HashMap;
+use std::collections::{BTreeMap, HashMap};
 use std::path::{Path, PathBuf};
 
 use anyhow::{bail, Context, Result};
@@ -97,6 +97,7 @@ impl Client {
         &mut self,
         manifest_path: &Path,
         reference: impl AsRef<str>,
+        annotations: Option<BTreeMap<String, String>>,
     ) -> Result<Option<String>> {
         let reference: Reference = reference
             .as_ref()
@@ -115,7 +116,8 @@ impl Client {
         )
         .await?;
 
-        self.push_locked_core(locked, auth, reference).await
+        self.push_locked_core(locked, auth, reference, annotations)
+            .await
     }
 
     /// Push a Spin application to an OCI registry and return the digest (or None
@@ -124,6 +126,7 @@ impl Client {
         &mut self,
         locked: LockedApp,
         reference: impl AsRef<str>,
+        annotations: Option<BTreeMap<String, String>>,
     ) -> Result<Option<String>> {
         let reference: Reference = reference
             .as_ref()
@@ -131,7 +134,8 @@ impl Client {
             .with_context(|| format!("cannot parse reference {}", reference.as_ref()))?;
         let auth = Self::auth(&reference).await?;
 
-        self.push_locked_core(locked, auth, reference).await
+        self.push_locked_core(locked, auth, reference, annotations)
+            .await
     }
 
     /// Push a Spin application to an OCI registry and return the digest (or None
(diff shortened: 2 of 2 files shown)


src/commands/registry.rs
```diff
@@ -5,6 +5,8 @@ use indicatif::{ProgressBar, ProgressStyle};
 use spin_oci::Client;
 use std::{io::Read, path::PathBuf, time::Duration};
 
+use super::new::ParameterValue;
```
itowlson, line 8: This type is used for template parameters.  I realise that today it happens to have the right shape and parsing behaviour that you need for annotations, but we should not bind the two together.  We should either:

* Create a new type defined by shape and behaviour (a la `EqualsSeparatedStringPair`) in a shared location, and use that in both places; or
* Create an Annotation type in this module, and use that here.

src/commands/registry.rs
```diff
@@ -66,11 +72,18 @@ impl Push {
             spin_build::build(&app_file, &[]).await?;
         }
 
+        let annotations = self.annotations.map(|parameters| {
```
itowlson, line 75: The names `parameters` and `parameter` in this section are misleading.

itowlson (commented):
Thanks for this!  Someone else will review the OCI side; I just had a couple of naming nits for the command code.  We really appreciate the contribution!

src/commands/registry.rs
rgl: I found a common type at `use spin_common::arg_parser::parse_kv`, so I've rewritten to code to use it.

src/commands/registry.rs
rgl: Ah, indeed! In the meantime, this code no longer exists. Please see the new one.

itowlson (approved):
Thanks for the update - this looks great!  I had one suggestion for simplifying the Vector -> HashMap conversion but it's not a blocker.  @vdice do you have a moment to look at the OCI side?

vdice (approved):
LGTM, thanks @rgl!

Will hold off on merging in case any optional changes per https://github.com/fermyon/spin/pull/2254#discussion_r1470064452 come through.

src/commands/registry.rs
```diff
@@ -66,11 +72,24 @@ impl Push {
             spin_build::build(&app_file, &[]).await?;
         }
 
+        let annotations =
+            if self.annotations.is_empty() {
+                None
+            } else {
+                Some(self.annotations.iter().cloned().fold(
```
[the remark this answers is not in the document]
rgl: if there are duplicate `--annotation` arguments with the same key (e.g. `--annotation a=a --annotation a=aa`) won't that cause a `panic`?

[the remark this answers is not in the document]
itowlson: No, last one wins, same as with the fold.

vdice:
@rgl sorry, I often forget this detail on new contributions:  Thanks for completing the DCO sign-off; can you also please ensure that the commit(s) are [GPG-signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)?

src/commands/registry.rs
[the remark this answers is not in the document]
rgl: Oh nice! Thank you for clarifying this!

I've updated the code, please check it now.

rgl:
@vdice I think everything should be ready to go now :-)

rgl:
@vdice please do not yet merge this until https://github.com/project-zot/zot/issues/2210 is addressed.

I'm starting to think that we should have a way to set the image manifest annotations (what the current `--annotation` does in this MR) and the image config annotations. What do you guys think? Maybe have two arguments? `--oci-image-manifest-annotation` and `--oci-image-config-annotation`?

vdice:
@rgl Would it be overkill to take the provided `annotation`s and add them to _both_ the manifest and image config?

A question: Refreshing my knowledge with the [config spec](https://specs.opencontainers.org/image-spec/config/?v=v1.0.1), I see that the config object has a `Labels` field intended to capture "... arbitrary metadata for the container."  Then, the config object, being a [descriptor](https://specs.opencontainers.org/image-spec/descriptor/?v=v1.0.1) (if I'm understanding correctly), can also be assigned generic [annotations](https://specs.opencontainers.org/image-spec/annotations/).  Which would we want to utilize here?

(As an aside, I built on this PR to play around a bit and I'm not immediately seeing generic annotations added to the config layer ([here](https://github.com/fermyon/spin/blob/main/crates/oci/src/client.rs#L189)) show up as intended in the manifest json... so there may be a bit of further work either in our oci client or the underlying [oci-distribution](https://github.com/fermyon/oci-distribution) crate.  I haven't yet tried adding metadata to the `Config.Labels` object on the config...)

vdice:
Hi @rgl checking in on this one.  Wondering what you were thinking re: https://github.com/fermyon/spin/pull/2254#issuecomment-1917343654.  Thanks!

endocrimes:
It looks like `docker` adds them to the config block:

```json
    "Config": {
            "Hostname": "",
            "Domainname": "",
            "User": "65532:65532",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt",
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
            ],
            "Cmd": null,
            "Image": "",
            "Volumes": null,
            "WorkingDir": "/",
            "Entrypoint": [
                "/manager"
            ],
            "OnBuild": null,
            "Labels": {
                "org.opencontainers.image.created": "2024-02-19T16:28:43.720Z",
                "org.opencontainers.image.description": "",
                "org.opencontainers.image.licenses": "NOASSERTION",
                "org.opencontainers.image.revision": "ff68ce40b1d5e202a717483297db0eaa3bbacdc5",
                "org.opencontainers.image.source": "https://github.com/my-org/my-project",
                "org.opencontainers.image.title": "my-project",
                "org.opencontainers.image.url": "https://github.com/my-org/my-project",
                "org.opencontainers.image.version": "main"
            }
    },
    "Metadata": {
        "LastTagTime": "2024-02-20T14:23:56.103651348Z"
    }
```

rgl:
@endocrimes do you known if that is aligned with the oci image-spec? I still didn't get a chance to spend quality time reading the specs, hopefully I can do it soon.

itowlson:
@endocrimes sorry to hound but bark, bark

endocrimes (approved):
I'm happy for this to land as is, and update later if needed for broader compatibility.

itowlson:
Thanks @endocrimes!  @rgl thanks for your patience - it's been a crazy busy time - could you rebase and resolve the merge conflict please?  Then we can land this.  Thanks!

rgl:
@itowlson no worries! I've rebased it.

itowlson:
Thank you @rgl!  Glad to have landed this and thanks again for sticking with it.

Issue #2251, opened 2024-01-27
Cannot pass trigger specific flags on up in multi trigger app


karthik2804:
I have an app that uses both a redis and a HTTP component. I was trying to set the HTTP listener to run on `3001` instead of `3000` by passing in.

```bash
spin up --listen localhost:3001
```

which results in an error as the redis trigger does not recognize the flag.  

```bash
spinrelease up --listen localhost:3001   
error: Found argument '--listen' which wasn't expected, or isn't valid in this context
```

Flags that are common like `--log-dir` continue to work.  

I am not exactly sure what the solution would be to this. One potential idea I can think of is that when multiple triggers are present spin can optionally parse arguments something along the lines of `--http-trigger="--listen localhost:3001" ` and only pass the argument to the HTTP trigger. and anything not in those specific arguments gets passed to both?

radu-matei:
As a user, I would expect passing the `--listen` flag to continue to work and to apply to the HTTP trigger, but not apply to the Redis trigger.

itowlson:
The least worst short-term solution may be simply to have triggers ignore unknown flags, although this does run us into the problem that `spin up --lsiten` would now silently discard and give unexpected default behaviour.  (For a nastier example, consider the TLS flags.)

Longer term we could consider some way for `up` to interrogate triggers about which flags they support and filter (or error) accordingly.  This would also help to construct unified help text (instead of repeating the help text for each trigger).

itowlson:
Semi-related: at some point, we may have a disambiguation issue, e.g. websocket `--listen` vs HTTP `--listen`.  (Please don't nitpick the example - I know about the relationship between web sockets and HTTP - it was just easier than making up two imaginary triggers that legitimately used the same flag name.)

lann:
We have a parallel problem with validating trigger config. Maybe the trigger plugin "interface" should include a mechanism for trigger executor binaries to describe their own config schema and arguments.

PR #2255, opened 2024-01-28
Ignore unknown trigger flags

itowlson:
This is one approach to addressing #2251 - it's not the best, but it may be expedient in the short term.  It has the serious downside that users who mistype flags, even for their existing, single-trigger applications, will not get errors.  This may be unacceptable - I'd be completely on board with that decision!

The alternatives I can think of are:

* Document that multi-trigger applications can't take trigger-specific flags, i.e. a known limitation of an experimental feature.  Easy to do, but few people will read it, and so people will hit the error which is not very self-explanatory.
* Enforce that multi-trigger applications can't take trigger-specific flags during this experimental phase: that is, if `up` detects that it's a multi scenario and the collection of passthrough flags is non-empty then halt with a meaningful error.  Easy to do, self-explanatory, but blocks common flags like `--log-dir` as well.
* Provide a way for triggers to list their flags, and have `up` use that to filter what it passes through to each trigger.  Requires more time and thought, and careful coding in `up` - there will be all sorts of exciting edge cases here (positional arguments, just say no; correct handling of 'old' triggers), but allows for other nice features down the line (help combining, config validation, detection of non-multi-able triggers).  I think we should do this, but I'm wary of trying to jam it into 2.2.

Other ideas welcome!

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -32,7 +32,8 @@ pub const SPIN_WORKING_DIR: &str = "SPIN_WORKING_DIR";
 #[derive(Parser, Debug)]
 #[clap(
     usage = "spin [COMMAND] [OPTIONS]",
-    next_help_heading = "TRIGGER OPTIONS"
+    next_help_heading = "TRIGGER OPTIONS",
+    ignore_errors = true
 )]
 pub struct TriggerExecutorCommand<Executor: TriggerExecutor>
 where


itowlson:
I have also started prototyping option 3 (triggers tell `up` which flags they can accept) but not sure how much I will trust it at 2.2.

rylev:
I wonder if option 2 with a few allow-listed options like `--log-dir` might be best in the short term. Option 3 does sound like the ideal if it's not too much effort to implement.

itowlson:
Closed in favour of #2257

PR #2256, opened 2024-01-29
Include trigger type in trigger options help header

### LINKED ISSUE #2252: Spin up --help on multi trigger not clear
When running the `spin up --help` in a multi-trigger app,  the output prints the help options of both but there is no way to identify which set of options belong to which trigger.

actual: 
```bash
$ spin up --help                 
spin-up 
Start the Spin application

USAGE:
    spin up [OPTIONS]

OPTIONS:
        ...
TRIGGER OPTIONS:
        ...
TRIGGER OPTIONS:
        ...
```

expected:

```bash
$ spin up --help                 
spin-up 
Start the Spin application

USAGE:
    spin up [OPTIONS]

OPTIONS:
        ...
HTTP TRIGGER OPTIONS:
        ...
REDIS TRIGGER OPTIONS:
        ...
```

itowlson:
Fixes #2252.

Note this does not deduplicate common options - it only adds the trigger type as part of the section header.

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -32,7 +32,7 @@ pub const SPIN_WORKING_DIR: &str = "SPIN_WORKING_DIR";
 #[derive(Parser, Debug)]
 #[clap(
     usage = "spin [COMMAND] [OPTIONS]",
-    next_help_heading = "TRIGGER OPTIONS"
+    next_help_heading = help_heading::<Executor>()
 )]
 pub struct TriggerExecutorCommand<Executor: TriggerExecutor>
 where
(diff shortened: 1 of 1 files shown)


karthik2804 (approved):

crates/trigger/src/cli.rs
```diff
@@ -255,6 +255,12 @@ fn warn_if_wasm_build_slothful() -> sloth::SlothGuard {
     sloth::warn_if_slothful(SLOTH_WARNING_DELAY_MILLIS, format!("{message}\n"))
 }
 
+fn help_heading<E: TriggerExecutor>() -> Option<&'static str> {
+    let heading = format!("{} TRIGGER OPTIONS", E::TRIGGER_TYPE.to_uppercase());
+    let as_str = Box::new(heading).leak();
```
rylev, line 260: Using `Box::leak` is fine here IMO, but I think perhaps using a static and `OnceCell` is another, arguably less hacky, solution.

rylev (approved):

crates/trigger/src/cli.rs
lann: Agreed; `leak` is perfectly fine for this purpose.

PR #2257, opened 2024-01-29
Allow only common trigger options in multi-trigger apps

itowlson:
Alternative to #2255 to address #2251

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -166,12 +166,26 @@ impl UpCommand {
         let trigger_cmds = trigger_command_for_resolved_app_source(&resolved_app_source)
             .with_context(|| format!("Couldn't find trigger executor for {app_source}"))?;
 
+        let is_multi = trigger_cmds.len() > 1;
+
         if self.help {
+            if is_multi {
+                // For now, only common flags are allowed on multi-trigger apps.
+                let mut child = self
+                    .start_trigger(trigger_command(HELP_ARGS_ONLY_TRIGGER_TYPE), None)
+                    .await?;
+                let _ = child.wait().await?;
+                return Ok(());
+            }
             for cmd in trigger_cmds {
                 let mut help_process = self.start_trigger(cmd.clone(), None).await?;
                 _ = help_process.wait().await;
             }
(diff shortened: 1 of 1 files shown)


src/commands/up.rs
```diff
@@ -410,6 +423,57 @@ impl UpCommand {
             }
         }
     }
+
+    fn allow_only_common_flags(&self) -> anyhow::Result<()> {
+        allow_only_common_flags(&self.trigger_args)
+    }
+}
+
+const COMMON_FLAGS: &[&str] = &[
```
rylev, line 432: Can we comment that these flags are common *to all well-known trigger types*.

src/commands/up.rs
```diff
@@ -410,6 +423,57 @@ impl UpCommand {
...
+    "--runtime-config-file",
+    "--state-dir",
+    "--key-value",
+    "--sqlite",
+    "--help-args-only",
+];
+
+fn allow_only_common_flags(args: &[OsString]) -> anyhow::Result<()> {
+    for i in 0..(args.len()) {
+        let Some(value) = args[i].to_str() else {
```
rylev, line 453:
```suggestion
    for (i, arg) in args.iter().enumerate() {
        let Some(value) = arg.to_str() else {
```

src/commands/up.rs
```diff
@@ -410,6 +423,57 @@ impl UpCommand {
...
+            // Err on the side of forgiveness
+            continue;
+        };
+        if value.starts_with('-') {
+            // Flag candidate. Is it allowed?
+            if !is_common_flag(value) {
```
rylev, line 459: Nit: I would personally inline this function.

rylev (approved):
This looks good to me!

itowlson:
@rylev Thanks for the review - all good points.  Unfortunately, as we want to land this for 2.2, I'm going to have to punt them to a tidy-up PR - sorry.  Also, as I'm already looking at Option 3 (and have it working on the happy path), which would enable trigger-specific flags, tidyings to this may become moot, but we'll see how it goes.

PR #2260, opened 2024-01-30
chore(release): Spin v2.2 version bumps

vdice:


vdice:
Oops, seeing the following from the Lint Rust job:
```
...
M	tests/test-components/components/Cargo.lock
Git working tree dirtied by lints. Run 'make update-cargo-locks' and check in changes
```

Alas, [make update-cargo-locks](https://github.com/fermyon/spin/blob/main/Makefile#L80-L86) currently only appears to update files under `examples`... Do we need to add `tests` as well?

itowlson:
@vdice Huh. We never used to have to do tests but I guess something changed...

vdice:
I pushed a commit manually running `cargo update -w --offline` in `tests/test-components/components`: https://github.com/fermyon/spin/pull/2260/commits/8f9f5af19e43920585e126934895d4850fb7051a.  We'll see if that suffices for our needs here.

PR #2266, opened 2024-01-31
Multi-trigger: ask triggers which flags they can accept

### LINKED ISSUE #2251: Cannot pass trigger specific flags on up in multi trigger app

itowlson:
Fixes #2251.  Yes.  AGAIN.

Very exploratory. WIP and crime-filled: please do not look yet!


crates/plugins/src/manifest.rs
```diff
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, Context, Result};
+use is_terminal::IsTerminal;
```
fibonacci1729, line 2: Can we just use `std::io::IsTerminal` now that it's part of the `std`?

fibonacci1729 (approved):

crates/plugins/src/manifest.rs
itowlson: We could and should.  This is what I get for mashing the "add a use" tooltip.  Thanks!

PR #2272, opened 2024-02-13
Fix test components build

lann:
- Exclude test-components/components/target from workspace members
- Add test components to vscode rust-analyzer.linkedProjects
- Write adapted test components to new path instead of overwriting modules

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -14,5 +14,9 @@
     },
     "rust-analyzer.files.excludeDirs": [
         "templates"
+    ],
+    "rust-analyzer.linkedProjects": [
+        "./Cargo.toml",
+        "./tests/test-components/components/Cargo.toml"
     ]
 }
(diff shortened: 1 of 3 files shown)


lann:
oops, that last commit doesn't actually...work...

.vscode/settings.json
```diff
@@ -14,5 +14,9 @@
     },
     "rust-analyzer.files.excludeDirs": [
         "templates"
+    ],
+    "rust-analyzer.linkedProjects": [
```
rylev, line 18: 😍

tests/test-components/build.rs
```diff
@@ -25,9 +25,14 @@ fn main() {
     for package in packages {
         let crate_path = PathBuf::from("components").join(&package);
         let manifest_path = crate_path.join("Cargo.toml");
+        if !manifest_path.exists() {
```
rylev, line 28: Nit: can we log when the path doesn't exist? The logs don't show unless a verbosity option is passed, but it could be nice to provide some sort of indication that a directory is being skipped.

rylev (approved):

PR #2275, opened 2024-02-14
update python tests to use componentize-py template

karthik2804:
Updated the test based on the latest update based on https://github.com/fermyon/spin-python-sdk/pull/66

Also, updates the testing readme to include information about running all the tests.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1,12 +1,8 @@
-use std::{fs, path::Path};
-
 mod testcases;
 
 mod integration_tests {
     use sha2::Digest;
-    use std::{collections::HashMap, path::Path};
-
-    use crate::visit_dirs;
+    use std::collections::HashMap;
 
     use super::testcases::{
         assert_spin_request, bootstap_env, http_smoke_test_template, run_test, spin_binary,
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -36,3 +36,9 @@ make test-integration
 ```
 
 Note that this also runs the runtime tests as well.
+
+To run all the tests that include external dependencies use:
+
+```bash
+make test-integration-full
+```
```
rylev, line 44: This is going to conflict with https://github.com/fermyon/spin/pull/2268 which includes a bit more detail

tests/integration.rs
```diff
@@ -417,12 +417,18 @@ Caused by:
     #[cfg(feature = "extern-dependencies-tests")]
     // Ignore until https://github.com/fermyon/spin-python-sdk/pull/65 is merged
```
rylev, line 418: I think this can be removed now

rylev (approved):

tests/README.md
karthik2804: I will remove this.

tests/integration.rs
karthik2804: Will do.

tests/integration.rs
```diff
@@ -415,14 +415,25 @@ Caused by:
 
     #[test]
     #[cfg(feature = "extern-dependencies-tests")]
-    // Ignore until https://github.com/fermyon/spin-python-sdk/pull/65 is merged
     fn http_python_template_smoke_test() -> anyhow::Result<()> {
+        let prebuild = |env: &mut testing_framework::TestEnvironment<_>| {
+            let mut setup_venv = std::process::Command::new("python3");
+            setup_venv.args(["-m", "venv", "venv"]);
+            env.run_in(&mut setup_venv);
+            let mut activate_venv = std::process::Command::new("source");
```
dicej, line 423: Does this actually work?  I.e. does it modify the environment of the current process?  I would think it only works inside a shell.

karthik2804: Good point! I am not sure why the test does not pass in CI though even without it.

itowlson (approved):

karthik2804:
seem to be running into an unrelated failure
```
failures:
    integration_tests::test_vault_config_provider
test result: FAILED. 36 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 371.07s
```

PR #2277, opened 2024-02-14
Migrate spin inbound-http test to integration test suite

lann:
Also sneak in a `use reqwest::Method`; it seemed weird that `reqwest` was appearing all over the test cases where its just an implementation detail.


tests/test-components/components/integration-spin-inbound-http/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+        return Err("request params field is deprecated".into());
+    }
+
+    let headers = req.headers.into_iter().collect::<HashMap<_, _>>();
+    let path = headers
+        .get("spin-path-info")
+        .map(|path| path.as_str())
+        .unwrap_or("");
+    match path {
+        "/echo" => Ok(req.body.unwrap_or_default()),
```
rylev, line 41: One thing I'd like for us to try to do is to remove redundancy in our testing. We have other tests that test echoing capabilities as well as tests that read request headers. The unique elements of this test seem to be asserting that `params` is empty and reading `spin-path-info`. I wonder if this test can be simplified?

rylev (changes requested):
I'd like to explore the other option that parts of this test can simply be deleted.

tests/test-components/components/integration-spin-inbound-http/src/lib.rs
lann: Don't the other tests only cover wasi handlers?

rylev: Ah you're right. Ok - we should probably make tests that test the wasi handler and the Spin handler mirror one another so we can more easily see if there are gaps in test coverage, but that can be done in a follow up.

rylev (approved):

rylev:
@lann you'll need to rebase for the CI fix

PR #2281, opened 2024-02-16
In process runtime tests

rylev:
This adjusts the runtime tests to use a Spin runtime that exists in the same process as the tests instead of as a separate Spin process which should make tests faster and more reliable. 

This is also the second runtime the runtime tests support so it's a stress test that the runtime testing framework is flexible enough to support different runtimes. This should make it easier to add additional runtimes in the future.

*Reviewers note*: you might want to review the first two commits as the third commit is just refactoring which will make the actual substantial changes harder to parse.

### DIFF
--- a/tests/runtime-tests/src/lib.rs
+++ b/tests/runtime-tests/src/lib.rs
@@ -124,7 +124,7 @@ impl RuntimeTest<InProcessSpin> {
             Ok(())
         };
         let services_config = services_config(&config)?;
-        let env_config = TestEnvironmentConfig::in_memory(services_config, preboot);
+        let env_config = TestEnvironmentConfig::in_process(services_config, preboot);
         let env = TestEnvironment::up(env_config)?;
         Ok(Self {
             test_path: config.test_path,
--- a/tests/runtime-tests/src/lib.rs
+++ b/tests/runtime-tests/src/lib.rs
@@ -150,10 +150,27 @@ impl RuntimeTest<InMemorySpin> {
         })
     }
 
-    fn run(&mut self) {
+    pub fn run(&mut self) {
         self.run_test(|env| {
             let runtime = env.runtime_mut();
-            todo!()
+            let response = runtime.make_http_request(testing_framework::Request::new(reqwest::Method::GET, "/"))?;
+            if response.status() == 200 {
+                return Ok(());
+            }
+            if response.status() != 500 {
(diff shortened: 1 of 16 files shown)


tests/runtime-tests/src/lib.rs
```diff
@@ -77,6 +79,108 @@ impl RuntimeTest<Spin> {
...
+
+            Err(TestError::Failure(RuntimeTestFailure {
+                error: text,
+                stderr: runtime.stderr().to_owned(),
+            }))
+        })
+    }
+}
+
+impl RuntimeTest<InMemorySpin> {
```
lann, line 102: micro-nit: maybe "InProcessSpin"? it still reads stuff from disk...

tests/testing-framework/src/test_environment.rs
```diff
@@ -191,3 +191,94 @@ impl TestEnvironmentConfig<Spin> {
...
+                .unwrap();
+            let response = self
+                .trigger
+                .handle(
+                    req,
+                    http::uri::Scheme::HTTP,
+                    std::net::SocketAddr::V4(std::net::SocketAddrV4::new(
+                        std::net::Ipv4Addr::LOCALHOST,
+                        80,
+                    )),
```
lann: FYI
```suggestion
                    (std::net::Ipv4Addr::LOCALHOST, 80).into(),
```

tests/testing-framework/src/test_environment.rs
```diff
@@ -191,3 +191,94 @@ impl TestEnvironmentConfig<Spin> {
...
+                        std::fs::write(env.path().join("locked.json"), json)?;
+
+                        let loader = TriggerLoader::new(env.path().join(".working_dir"), false);
+                        let trigger = TriggerExecutorBuilder::<HttpTrigger>::new(loader)
+                            .build(
+                                format!("file:{}", env.path().join("locked.json").display()),
+                                RuntimeConfig::default(),
+                                HostComponentInitData::default(),
+                            )
+                            .await?;
```
lann, line 228: SEE ITS NOT SO BAD

tests/testing-framework/src/http.rs
```diff
@@ -0,0 +1,131 @@
+//! Utilities for tests that function over HTTP
+
+use std::collections::HashMap;
+
```
lann, line 4: Maybe `pub use reqwest::Method` to expose less impl detail?

lann (approved):

PR #2283, opened 2024-02-16
unify WIT files

dicej:
This addresses two issues:

- When upgrading to WASI 0.2.0, I briefly considered bumping the `fermyon:spin` package version to 2.2.0 and updating the Rust and Go SDKs to use WASI 0.2.0. I later decided against that due to semver concerns, but forgot to change the package version back to 2.0.0.  Complicating things further was that the package version under wit-2023-10-18 was still 2.0.0, and that's the one we were actually using in `spin-core`, so we didn't notice there was a problem until someone tried to use the version under wit-0.2.0 instead.

- Having three separate WIT directories with inconsistent duplicates is obviously confusing, so I've unified them.  This unfortunately resulted in a lot of (superficial) code changes since `wit-bindgen` now needs to generate separate modules for various WASI package versions, but at least we've eliminated the duplication and ambiguity.


lann (approved):

crates/core/src/wasi_2023_10_18.rs
```diff
@@ -100,66 +100,66 @@ wasmtime::component::bindgen!({
     },
 });
 
-use wasi::cli::terminal_input::TerminalInput;
-use wasi::cli::terminal_output::TerminalOutput;
-use wasi::clocks::monotonic_clock::Instant;
-use wasi::clocks::wall_clock::Datetime;
-use wasi::filesystem::types::{
+use wasi::cli0_2_0_rc_2023_10_18::terminal_input::TerminalInput;
```
tschneidereit, line 103: While that won't make the diff less noisy, would it perhaps make the resulting code a bit nicer if we changed the `use` statements to alias away the version? I.e. start the implementation files with a set statement along the lines of
```rust
use wasi::{
    cli0_2_0_rc_2023_10_18 as wasi_cli,
    clocks0_2_0_rc_2023_10_18 as wasi_clocks,
    ...
};
```

tschneidereit: Another advantage of this would be that if the bindings generation changes in the future to omit version strings in more scenarios (as I think it could here, because only one version of each interface is actually included in the list of interfaces) we'd only have to change this one place.

tschneidereit (approved):
Thank you for doing this! ❤️ 

Besides the comment I left inline, one question: is the `wit/deps` folder required to have this layout? And more specifically, would it be possible to organize it to bundle the different versions in sub-folders?

I.e., `wit/deps/cli-2023-10-18` would become `wit/deps/wasi-0.2.0-rc-2023-10-18/cli`, etc.

If that is possible, I think it'd be good to apply it. If not, never mind :)

In any case, r=me with these changes applied or an explanation for why they don't work / make sense.

dicej:
@tschneidereit The directory structure you described was the first thing I tried, but `wit-bindgen` didn't like that, complaining that no package version was found in the first-level directory.  So I don't _think_ there's a way to do it, but maybe @alexcrichton can confirm or refute that.

alexcrichton:
That's correct, `wit-parser` requires a "flat" `deps` folder.

In the future though you can do `deps/wasi-cli@0.2.0.wasm` plus `deps/wasi-cli@0.2.1.wasm` plus ...

(and by future in mean in the future with Wasmtime, I think the latest `wit-bindgen` should support that)

tschneidereit:
ok, that makes sense. @alexcrichton I wonder if that's something we could teach `wit-bindgen` to support as well? I think keeping textual representations of interfaces would be nice, and a bit more flexibility on directory layout for WIT files would be as well :)

crates/core/src/wasi_2023_10_18.rs
```diff
@@ -14,152 +14,172 @@ mod latest {
...
+            "wasi:http/types/outgoing-body": latest::http::types::OutgoingBody,
+            "wasi:http/types/fields": latest::http::types::Fields,
+            "wasi:http/types/response-outparam": latest::http::types::ResponseOutparam,
+            "wasi:http/types/future-incoming-response": latest::http::types::FutureIncomingResponse,
+            "wasi:http/types/future-trailers": latest::http::types::FutureTrailers,
+        },
+    });
+}
+
+mod wasi {
```
tschneidereit, line 108: Ooh, that's much nicer than my suggestion! ❤️

PR #2287, opened 2024-02-20
Feature/mqtt publisher

suneetnangia:
PR to add MQTT publisher in Spin.


wit/mqtt.wit
```diff
@@ -0,0 +1,32 @@
...
+      /// There are too many open connections
+      too-many-connections,
+      /// A retrieved value was not of the correct type
+      type-error,
+      /// Some other error occurred
+      other(string),
+  }
+
+  /// QoS for publishing Mqtt messages
+  /// TODO: will enum result in consistent values across languages?
```
radu-matei, line 15: cc @itowlson @dicej @rylev

lann: Sort of. There is an index associated with each enum variant but it isn't necessarily exposed to guest languages.

suneetnangia:
Could someone from Spin pls assist with cross build failure for aarch64? 
I can locally cross build (for aarch64-unknown-linux-musl) paho with vendored-ssl feature but it seems cross build for Spin is customised and I am not sure if I want to modify that without much background on that customisation. Thanks.

itowlson:
@suneetnangia To be clear, are you saying:

* You can cross build Spin to aarch64 locally but not in CI; or
* You cannot cross build Spin to aarch64 locally _or_ in CI (but you can cross build paho-mqtt locally in other applications)

itowlson:
A colleague notes:

> I did notice that spin uses `openssl` crate but paho-mqtt uses `openssl-sys` crate. Not sure what the difference is but just wanted to point that out.

although I am not sure if that's a fixable thing or why it would specifically affect cross compilation

itowlson:
Another colleague suggests adding `cmake` to the cross-compilation Dockerfile (https://github.com/fermyon/spin/blob/main/cross/Dockerfile) and reports that with this installed it does build on aarch64 (but was unable to test `cargo-cross`).

alexcrichton:
I've debugged this locally and was able to reproduce the failure here in CI and also reproduce the ability that things to actually build successfully locally. After debugging further it appears that this is related to CMake configuration which is specific to the `cross` tool used to cross-compile in CI.

I am unfortunately no CMake expert myself, but [these lines](https://github.com/cross-rs/cross/blob/c87a52a60dbcde069714a8d2ed51570e7fc23cf9/docker/toolchain.cmake#L43-L47) are present as `/opt/toolchain.cmake` in the `cross` container image. That file is then passed through to get down to cmake which then bottoms out with the error you're seeing on CI. I found that if I commented out those lines then the build succeeds in the given container.

I believe what's happening is that with `CMAKE_FIND_ROOT_PATH` set it attempts to append any paths to that. That means that the build script which passes `-DOPENSSL_ROOT_DIR=/target/..` ends up getting probed at the path `/usr/local/aarch.../target/...`, e.g. the root path there is erroneously prepended to the actual path to openssl. That means that cmake is unable to find openssl and the build fails.

Again though I'm not a CMake expert. In that sense I don't know how to fix this, nor do I know why the `toolchain.cmake` file is configured by the `cross` tool. I've found that older images used by `cross` do not specify this cmake file and I'm able to build successfully with cross (tested by changing [this line](https://github.com/fermyon/spin/blob/4cfbac6ced6e6318a464a4055a526c6d3aac55a4/cross/Dockerfile#L1) to `:0.2.5` instead of `@sha...`). That was recently changed in #2179 to specifically not be the older image (I think) so I'm not sure that a viable fix is to "just" change to 0.2.5 of the cross image in this case.

rajatjindal:
Hi @suneetnangia, I have merged a potential workaround for fixing this CI issue. could you please rebase against main and try again.

thank you.

suneetnangia:
@rajatjindal thanks, that fixed it! Thanks all for lending a hand here.

squillace:
Cheers to @suneetnangia and @alexcrichton here.

suneetnangia:
Example app added [here](https://github.com/fermyon/spin-rust-sdk/pull/17) as a subsequent PR.

suneetnangia:
Trigger plugin is developed [here ](https://github.com/suneetnangia/spin-mqtt-trigger-sdk)

crates/outbound-mqtt/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+impl v2::HostConnection for OutboundMqtt {
+    async fn open(
+        &mut self,
+        address: String,
+        username: String,
+        password: String,
+        keep_alive_interval: u64,
+    ) -> Result<Result<Resource<MqttConnection>, Error>> {
+        if !self.is_address_allowed(&address) {
+            return Ok(Err(Error::InvalidAddress));
```
itowlson, line 70: This should specifically indicate that the address is not permitted (cf. e.g https://github.com/fermyon/spin/blob/3ee465e8ae114d627a1441327bca44d5c9ad0084/crates/outbound-mysql/src/lib.rs#L84)

wit/deps/spin@unversioned/mqtt.wit
```diff
@@ -0,0 +1,6 @@
+interface mqtt {
+  use mqtt-types.{payload, error, qos};
+
+  // Publish an MQTT message to the specified topic and return an error, if any.
+  publish: func(address: string, username: string, password: string, keepaliveinterval: u64, topic: string, payload: payload, qos: qos) -> result<_, error>;
```
itowlson, line 5: Could the keep alive interval indicate the units please?

itowlson, line 5: Suggest `keepalive-interval` or `keep-alive-interval` instead of all the words run together.

crates/outbound-mqtt/src/host_component.rs
```diff
@@ -0,0 +1,37 @@
...
+
+pub struct OutboundMqttComponent;
+
+impl HostComponent for OutboundMqttComponent {
+    type Data = OutboundMqtt;
+    fn add_to_linker<T: Send>(
+        linker: &mut spin_core::Linker<T>,
+        get: impl Fn(&mut spin_core::Data<T>) -> &mut Self::Data + Send + Sync + Copy + 'static,
+    ) -> anyhow::Result<()> {
+        spin_world::v1::mqtt::add_to_linker(linker, get)?;
```
itowlson, line 15: The v1 world is for backward compatibility.  Since old SDKs won't have MQTT anyway, you can skip it and just do v2.

crates/outbound-mqtt/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+    async fn publish(
+        &mut self,
+        connection: Resource<MqttConnection>,
+        topic: String,
+        payload: Vec<u8>,
+        qos: Qos,
+    ) -> Result<Result<(), Error>> {
+        Ok(async {
+            let client = self.get_conn(connection).await.map_err(other_error)?;
+            let message = paho_mqtt::Message::new(&topic, payload, qos as i32);
```
itowlson, line 90: `qos as i32` couples the WIT very directly to the Paho implementation.  I'd suggest implementing a conversion function that matches on the enum value.

crates/outbound-mqtt/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+            Err(_) => return Ok(Err(v1::Error::Error)),
+        };
+        Ok(<Self as v2::HostConnection>::$name($self, connection, $($arg),*)
+            .await?
+            .map_err(|_| v1::Error::Error))
+    }};
+}
+
+#[async_trait]
+impl v1::Host for OutboundMqtt {
```
itowlson, line 125: Per discussion on the v1 world, I believe you can skip this.

crates/outbound-mqtt/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+impl OutboundMqtt {
+    async fn get_conn(
+        &mut self,
+        connection: Resource<MqttConnection>,
+    ) -> Result<&mut Client, Error> {
+        self.connections
+            .get_mut(connection.rep())
+            .ok_or(Error::Other(
```
itowlson, line 159: In other places we use a ConnectionFailed variant for this case.  Is that not appropriate here?

wit/deps/spin@unversioned/world.wit
```diff
@@ -24,6 +24,7 @@ world platform {
   import mysql;
   import sqlite;
   import redis;
+  import mqtt;
```
itowlson, line 27: I think this is probably not needed (I think unversioned is for back compat).  Hopefully someone more knowledgeable can chime in...

wit/mqtt.wit
```diff
@@ -0,0 +1,32 @@
+interface mqtt {
+  /// Errors related to interacting with Mqtt
+  variant error {
+      /// An invalid address string
+      invalid-address,
+      /// There are too many open connections
+      too-many-connections,
+      /// A retrieved value was not of the correct type
+      type-error,
```
itowlson, line 9: I'm probably missing something but I don't see where this gets used?

itowlson (commented):
Thanks for this - it looks really good.  I've left a few comments, all pretty minor I think, but otherwise this looks good to go!  I am stoked to get this into Spin!

wit/mqtt.wit
suneetnangia: Will remove, not used indeed.

wit/deps/spin@unversioned/mqtt.wit
suneetnangia: Changing it to keep-alive-interval-in-secs

crates/outbound-mqtt/src/lib.rs
suneetnangia: Was trying to be consistent with Redis implementation which uses `Other` as well? [ here ](https://github.com/fermyon/spin/blob/3ee465e8ae114d627a1441327bca44d5c9ad0084/crates/outbound-redis/src/lib.rs#L331) or did I misunderstand your comment?

crates/outbound-mqtt/src/lib.rs
suneetnangia: Agreed, more appropriate (will update), think we may need to reflect this in [outbound Redis](https://github.com/fermyon/spin/blob/3ee465e8ae114d627a1441327bca44d5c9ad0084/crates/outbound-redis/src/lib.rs#L76) too for consistency?

itowlson: That would be good, but I don't think Redis has a suitable error variant for doing that, and adding or modifying one would have compatibilty considerations.  Longer term for sure though - thanks for the spot.

suneetnangia:
Updated the PR to reflect comments, @itowlson thanks for reviewing, if you can resolve/respond to them that'd be great.

wit/deps/spin@unversioned/world.wit
```diff
@@ -23,7 +23,7 @@ world platform {
   import postgres;
   import mysql;
   import sqlite;
-  import redis;
+  import redis;
```
itowlson, line 26:
```suggestion
  import redis;
```

(accidental whitespace change)

itowlson (approved):
LGTM - thanks!  

I spotted one spurious diff from whitespace, which it would be nice to tidy.  And if you have a moment, it would be great if you could rationalise the commit history a bit.  But none of this is blocking - it looks great.

wit/deps/spin@unversioned/world.wit
suneetnangia: ehh... let me fix that, wish linting could check that as well.

itowlson:
Merging this so I can update #2299 for it.  @lann @radu-matei if any concerns then let's do a follow-up PR.

itowlson:
@suneetnangia We're running into a problem with Windows builds in the release configuration (for some reason PR builds work, but main builds don't) - the openssl-sys dependency on Perl is problematic on Windows generally, and even in the friendly environment of GH runners there's a dependency missing.  We're trying to understand what's going on and how to get this buildable on Windows, but if we can't resolve it soon we **may** have to revert the commits and ask you to recreate the PR (so that we have a releasable main).  We're not at that point yet but just letting you know so it's not a nasty surprise if it happens.  Thanks for being understanding!

squillace:
@itowlson, totes understood. Let us know if there's anything we can do to help.

rajatjindal:
@rylev opened a potential fix/workaround for this: https://github.com/fermyon/spin/pull/2328. @suneetnangia when you get chance, could you please take a look and provide your feedback there.

PR #2290, opened 2024-02-21
Service chaining SIP

itowlson:


docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+We have [prototyped](https://github.com/fermyon/spin/pull/2229) dispatching these requests directly to the component handling the route, but this was in the context of an implementation detail of Spin.  The requirement for this project is for an application to declare that it _requires_ dispatch to run locally on the node, and for hosts that cannot meet that prerequisite to reject the application.  This allows developers to write applications as a network of chained microservices, without having to be concerned about what happens if the microservices are not co-located.
+
+# User experience
+
+## Expressing a requirement for chaining
+
+To express that the application requires chaining for self-requests, add the following to `spin.toml`:
+
+```toml
+[application.prerequisites]
```
fibonacci1729, line 26: Major nit: `prerequisites` threw me at first glance, how do you feel about `requirements` or `runtime_requirements`?

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+service_chaining = "required"
+```
+
+**BIKESHED:** Or something more explicit such as `direct_self_requests`.  Or `prerequisites` could be an array of capabilities that the host must provide (`prerequisites = ["service-chaining"]`).  Also, we _could_ call this something like `environment` or `host` instead, but I'm kind of inclined to leave it generic as to what prerequisites might go in here.
```
fibonacci1729, line 30: +1. I like the idea of declaring host requirements as a list of strings to avoid the frustrating serde level schema errors.

docs/content/sips/017-service-chaining.md
lann: A downside of "requirements": it is synonymous with "dependencies" in some places (Pythonlandia).

`features`? :grimacing:

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+Entries in `prerequisites` are considered must-understand.  For the section itself, that has to happen at a `serde` level since we can't retrofit it to existing builds.  But `serde` errors are not very lovely, and I'd prefer hosts in future to be able to say "hey I don't support your `gpu` requirement" rather than "unknown field serialisation error."
```
lann, line 32: This could be a semi-structured table with any key allowed but values constrained to e.g. `"required"` or `"preferred"` (it may also be nice to support `true` as a synonym for `"required"`.

Runtimes would then bail on `unknown_feature = "required"` but could ignore (perhaps warn on) `unknown_feature = "preferred"`.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Requesting a component via chaining
+
+There are two ways we could express a HTTP request to a chained component:
+
+1. Use the existing self-request style.  The host _may_ chain self-requests, and _must_ chain them if the "require chaining" setting is on.  This is super easy for the application developer, at both the send and receive end.  However, it becomes more complex for the host, as the host now has to route the request - this drags in much more of the HTTP trigger plumbing.  (The prototype did this and it was egregiously hairy.)
```
lann, line 38: It also requires the developer to keep these requests in sync with the trigger routing table, which may not always be desirable.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+2. Address requests directly to the destination components by name. We still need a URL, so this would introduce a special host name such as `spin.internal`.  Then the URL `http://spin.internal/accounts` would invoke the `accounts` component.  This bypasses routing, which should simplify implementation.  It also opens the door to private endpoints, where an internal microservice is not exposed to the network (does not have a HTTP route) and so is accessible only from within the application.  However, chained components which parse URLs may need to take extra care to correctly handle both routed and chaining URLs.
```
lann, line 40: > chained components which parse URLs may need to take extra care to correctly handle both routed and chaining URLs.

This may be an argument for a slightly different URL scheme: `http://<component>.spin.internal/<any path allowed here>`

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+The preferred user experience is option 2: `http://spin.internal/<component-name>`
+
+**DETAIL NEEDED:** How will Spin request headers be set in this case?  One option is that Spin won't set them _at all_ and the calling component has to inject them if the receiver needs them.  Otherwise e.g. Consider a request to `http://spin.internal/accounts/314`.  `spin-path-info` (the 'wildcard match' bit of the path) is probably `/314` (bit after component name), but what about `spin-matched-route` or `spin-component-route`?  Are they just not set?  What of `spin-client-addr` - forwarded or skipped?
+
+**DETAIL NEEDED:** Should Spin automatically set an additional header indicating that this is a chained (unrouted) call?  If so, should it contain any additional data?  Or is it up to the caller to decide what, if anything, they want to disclose?
```
lann, line 46: I think that we should (at least) say that `host: *.spin.internal` appears on requests if and only if they have been routed internally.

Brainstorming, it might be nice to also define a header prefix that similarly can only traverse internal calls, i.e. would be filtered out of any inbound http request: something like `spin-internal-*`.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Permissions
+
+The developer must enable chaining by adding `http://spin.internal` to the list of allowed outbound hosts.  (We cannot use the application flag as permission because not all components may be allowed to make self-requests.)  We could and arguably should consider that any component with `self` permission also has chaining permission: if a component is allowed to make requests to other parts of the application, who cares if that's routed or chained?  However, the idea of private endpoints complicates this a bit - is there a situation in which a component should be allowed to access public routes but not private endpoints?
```
lann, line 50: Extending the suggestion above to use `<component>.spin.internal`, this would also give us a natural mechanism for more fine-grained control over internal requests via `allowed_outbound_hosts`, e.g. `allowed_outbound_hosts = ["*.spin.internal"]` vs `allowed_outbound_hosts = ["specific-component.spin.internal"]`

lann: > However, the idea of private endpoints complicates this a bit - is there a situation in which a component should be allowed to access public routes but not private endpoints?

Given that private endpoints don't exist yet we could just say that `self` only grants access to public endpoints.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+Currently, `self` allows access to all routes in the application.  The initial version of service chaining will be similar; if someone needs to restrict access to specific components, that can be a future option.
+
+# Runtime and framework considerations
+
+## Chaining requirements in the lock file
+
+The lock file format built into existing hosts is designed for extension, and so [avoids using serde "deny unknown fields"](https://github.com/fermyon/spin/blob/663593d1423ed3518744d6f797e6a3970575d617/crates/locked-app/src/locked.rs#L16).  Unfortunately, this means there is no way to add a prerequisites section to the lock file in such a way that existing hosts will reject it if present.  For example, if someone deploys a chaining-required app onto an older version of the containerd shim, it will appear to accept it, but will not chain.
+
+Unfortunately, I don't think there's anything we can do about this, except to introduce a new `spin_lock_version`.  Doing this naively would mean existing hosts would reject the lock file _whether or not it contains prerequisites_.  We could get smart and emit a v.next lock file _only_ if the app expressed prerequisites but hoo boy.
```
lann, line 62: Pragmatically, I think we could still get away with this since the locked app JSON format is only exposed externally in OCI artifacts which are not (yet) in common use outside of Fermyon-maintained systems. We could also improve this problem for the future by adding `deny_unknown_fields` to the top-level `LockedApp` schema (but not `metadata`; forcing careful thought about this kind of compat problem is very much the point there).

docs/content/sips/017-service-chaining.md
lann: This could be a semi-structured table with any key allowed but values constrained to e.g. `"required"` or `"preferred"` (it may also be nice to support `true` as a synonym for `"required"`.

Runtimes would then bail on `unknown_feature = "required"` but could ignore (perhaps warn on) `unknown_feature = "preferred"`.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+In any case, whether as a section or as an entry in `metadata`, the updated format specifies (and the Spin crate enforces) must-understand semantics for all entries under it.
+
+## Spin command line implementation
+
+## Where do invoked components run?
+
+In Spin, all components of an app run within a trigger process (typically a child Spin process, but can also be a plugin).
+
+When the calling component is on the HTTP trigger, life is (relaTively) easy, because the host process has all the HTTP components - that is, all possible invocation targets - loaded and ready to go.
```
lann, line 74: > relaTively

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+When the calling command is on a _non-HTTP_ trigger, things are more challenging.  Either we must host a copy of the HTTP trigger engine (and load HTTP components) in the non-HTTP trigger process, or we must transfer the request to the HTTP trigger for handling.  Neither of these options is very appealing.
+
+**DECISION REQUIRED:** Do we allow service chaining from non-HTTP components to HTTP ones, and if so where do we run the HTTP components?
```
lann, line 78: I think the entire premise of this feature probably requires that the chunk of the http trigger code responsible for invoking guest http handlers be extracted into another crate (probably `spin-http` if that doesn't cause some dependency tree meltdown). That chunk can then be reused between the http trigger and the outbound http host component.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,106 @@
...
+
+**DETAIL NEEDED:** Should Spin automatically set an additional header indicating that this is a chained (unrouted) call?  If so, should it contain any additional data?  Or is it up to the caller to decide what, if anything, they want to disclose?
+
+## Permissions
+
+The developer must enable chaining by adding `http://spin.internal` to the list of allowed outbound hosts.  (We cannot use the application flag as permission because not all components may be allowed to make self-requests.)  We could and arguably should consider that any component with `self` permission also has chaining permission: if a component is allowed to make requests to other parts of the application, who cares if that's routed or chained?  However, the idea of private endpoints complicates this a bit - is there a situation in which a component should be allowed to access public routes but not private endpoints?
+
+**DECISION REQUIRED:** Should `allowed_outbound_hosts` including _either_ of `self` or `spin.internal` be enough to grant permission for self-requests using chaining?
```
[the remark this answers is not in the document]
lann: It may depend on details of how private endpoints are configured. One option would be to say that any component the exports an http handler is "eligible" for private service chaining, in which case it could be very surprising that the `self` permission suddenly applies to new components. :shrug:

docs/content/sips/017-service-chaining.md
itowlson: If this is in reference to URL parsing, it doesn't in itself solve the problem: components will still need to be aware that they could be called via two different paths.  But I do like that it keeps the component name safely separated from the rest of the path, so it gets discarded "naturally" without the receiver needing to strip segments.  As you note elsewhere, it also allows more granular restriction based on hosts, so I'm in favour.

docs/content/sips/017-service-chaining.md
lann: `[thingies]` it is!

docs/content/sips/017-service-chaining.md
itowlson: It's exposed to external triggers: if we allow non-HTTP triggers to chain, then triggers linked against downlevel versions of the trigger crate need to reject applications with the chaining requirement.

lann: Another option would be to "get smart and emit a v.next lock file _only_ if the app expressed prerequisites" but drop that logic in a future major spin release.

docs/content/sips/017-service-chaining.md
lann: `[runtime_features]`?

docs/content/sips/017-service-chaining.md
lann: An option to avoid having to bump the version for similar purposes in the future:

- Update the `spin_lock_version` to `1`
- Add a new top-level field, e.g. `spin_lock_features = ["host_prereqs"]`; any features in this set must be known-supported by host
- Set `host_prereqs` iff there are any `required` host prereqs

docs/content/sips/017-service-chaining.md
lann: See also https://github.com/fermyon/spin/issues/2293 which may need some similar application-global runtime configuration (e.g. outbound http distributed trace headers).

itowlson:
Starting to look at how to get this working and I believe we need to revisit the opt-in options.

The idea of an application saying "chain if the host supports it but not required" was appropriate when we envisaged implementing chaining on routed self-references.  However, now that chaining occurs _only_ on `spin.internal` (and is to component names, not routes), hosts that don't support chaining will have no way to service `spin.internal` requests.

Therefore, the only options are "this application uses and requires service chaining" or "this application does not use service chaining."

(This does not mean that future host prerequisites couldn't have a "preferred" value - just that this one doesn't.)

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,124 @@
...
+service_chaining = "required"
+```
+
+Entries in the `prerequisites` table can have any name, but the value must be one of:
+
+* `"required"` - the host must reject the application if it does not recognise the name, or cannot fulfil the requirement
+* `"preferred"` - the host should warn if it does not recognise the name, or cannot fulfil the requirement, but it shouldn't reject the application
+* `"optional"` (default) - this is the same as omitting the requirement but it might be nice to have a way to state it explicitly
+
+**BIKESHED:** Or something more explicit such as `direct_self_requests`? I'm not sure if "service chaining" will be obvious to readers (it wasn't to me the first time I heard it).
```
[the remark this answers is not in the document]
lann: "Local service chaining"?

docs/content/sips/017-service-chaining.md
```diff
@@ -14,30 +14,26 @@ Created: Feb 21, 2024
...
+If a host doesn't understand the `host_requirements` section, it must reject the application.  We can't retro-fit this to existing hosts so we have to rely on `serde` for this, which means the error will be unlovely.  However, for individual fields within the section, hosts can give better errors/warnings.
 
-**BIKESHED:** `prerequisites` is a pretty grim name. Other ideas: `requirements` or `features` (maybe too generic), or something referring to the host e.g. er `host` or `host_requirements`, which may be close the door to other kinds of precondition but is at least explicit.
-
-If a host doesn't understand the `prerequisites` section, it must reject the application.  We can't retro-fit this to existing hosts so we have to rely on `serde` for this, which means the error will be unlovely.  However, for individual fields within the section, hosts can give better errors/warnings.
+If a host understands the section, but doesn't understand an entry in that section, it _must_ reject the application with an error message explaining the required host features.
```
lann, line 36: Should we bake in a non-mandatory value now (`"preferred"`/`"optional"`)? These could still be disallowed for this specific feature but would permit backward-compatible additions in the future.

docs/content/sips/017-service-chaining.md
```diff
@@ -14,30 +14,26 @@ Created: Feb 21, 2024
...
 
 ## Expressing a requirement for chaining
 
-To express that the application requires chaining for self-requests, add the following to `spin.toml`:
+To express that the application requires service chaining, add the following to `spin.toml`:
 
 ```toml
-[application.prerequisites]
-service_chaining = "required"
+[application.host_requirements]
```
lann, line 28: Continuing the great bike shed tradition:

The term "requirement" doesn't necessarily evoke _enabling_ something for me. More importantly, the _absence_ of a particular "requirement" doesn't necessarily imply that a feature is _disabled_.

lann: I guess while I'm here arguing about the shed color I should probably suggest that we build a garage instead:

```toml
[application.host_features]
local_service_chaining = "enabled"
```

docs/content/sips/017-service-chaining.md
```diff
@@ -61,9 +59,11 @@ To avoid spoofing, Spin must strip any `Host: *.spin.internal` headers from rout
...
+The developer must enable chaining to a component by adding `http://<component-name>.spin.internal`, or `http://*.spin.internal` to allow any destination component, to the list of allowed outbound hosts.  (We cannot use the application flag as permission because not all components may be allowed to make self-requests.)
 
-**DECISION REQUIRED:** Should `allowed_outbound_hosts` including _either_ of `self` or `*.spin.internal` be enough to grant permission for self-requests using chaining?  What about hypothetical private endpoints?
+We could and arguably should consider that any component with `self` permission also has chaining permission: if a component is allowed to make requests to other parts of the application, who cares if that's routed or chained?  However, the idea of private endpoints complicates this a bit - is there a situation in which a component should be allowed to access public routes but not private endpoints?
+
+**The preferred user experience is that the developer must list the `<component-name>.spin.internal` host.**
```
lann, line 66: What about `http://*` or `*://*:*`? On one hand it would be quite odd for these to _not_ match any specific URL, but you also don't want to accidentally permit internal access when you just wanted a proxy or what have you.

docs/content/sips/017-service-chaining.md
```diff
@@ -91,16 +91,14 @@ When the calling command is on a _non-HTTP_ trigger, things are more challenging
 
 2. We somehow transfer the request to the HTTP trigger for handling. This is both difficult _and_ unappealing. Let's not do it.
 
-**DECISION REQUIRED:** Do we allow service chaining from non-HTTP components to HTTP ones?
-
 > _Additional context:_ Today, only HTTP components can make self-requests anyway (because non-HTTP components can't infer the host URL).
 
+**For the initial release we will allow chaining only from HTTP components to HTTP components.**  This does not preclude allowsing chaining from non-HTTP components in a future drop.
```
lann, line 96: > allowsing

docs/content/sips/017-service-chaining.md
itowlson: Yes, wildcards would still match... sorry, I didn't say that explicitly.

docs/content/sips/017-service-chaining.md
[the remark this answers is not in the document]
itowlson: @lann I think the problem is the "service chaining."  Sprinkling "local" on it doesn't help to explain what it is.

docs/content/sips/017-service-chaining.md
lann: It occurs to me that if we required allowed hosts to explicitly list `spin.internal` (not wildcards) we could get rid of this section entirely.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,122 @@
...
+
+1. We must host a large chunk of the HTTP trigger engine (and load HTTP components) in the non-HTTP trigger process. Physically this means moving the HTTP execution engine into a crate that can be shared by the `outbound-http` and `trigger-http` crates. However, this will still need a lot of the same infrastructure that the full trigger does, e.g. loading host components, permissions, etc. - I don't think it will omit much except the hooking up to the server/router. The two different ways of loading HTTP components feels like an exciting area for innovating in subtle surprises, and this is without even thinking about external trigger processes embedding different versions of the HTTP engine...
+
+2. We somehow transfer the request to the HTTP trigger for handling. This is both difficult _and_ unappealing. Let's not do it.
+
+> _Additional context:_ Today, only HTTP components can make self-requests anyway (because non-HTTP components can't infer the host URL).
+
+**For the initial release we will allow chaining only from HTTP components to HTTP components.**  This does not preclude allowing chaining from non-HTTP components in a future drop.
```
itowlson, line 96: Agreed at review

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,122 @@
...
+
+2. Address requests directly to the destination components by name. We still need a URL, so this would introduce a special host name such as `spin.internal`.  Senders address individual components via subdomains.  Thus, the URL `http://accounts.spin.internal` would invoke the `accounts` component.  This bypasses routing, which should simplify implementation.  It also opens the door to private endpoints, where an internal microservice is not exposed to the network (does not have a HTTP route) and so is accessible only from within the application.  However, chained components which parse URLs may need to take extra care to correctly handle both routed and chaining URLs.
+
+**The preferred user experience is option 2: `http://<component-name>.spin.internal/`**
+
+> This is why we cannot have "preferred but not required" for internal requests.  Using the `foo.spin.internal` host requires direct component addressing by the host - we cannot fall back to making a routed request.
+
+## Incoming request headers
+
+To avoid spoofing, Spin must strip any `Host: *.spin.internal` headers from routed requests.
```
itowlson, line 52: Decision: agreed to block this header pattern, and to ensure that chained requests supply the header.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,122 @@
...
+
+**DETAIL NEEDED:** It's proposed that we define a pattern such as `spin-internal-*` such that headers matching this pattern are stripped from routed requests: these headers should be settable only by the chaining infrastructure.  (Or can be set by the sender of a chained request?)  I'm not sure we've identified a concrete use case yet though?
+
+**DETAIL NEEDED:** Should Spin automatically set an additional header indicating that this is a chained (unrouted) call?  If so, should it contain any additional data?  Or is it up to the caller to decide what, if anything, they want to disclose?
```
itowlson, line 56: Decision: No

docs/content/sips/017-service-chaining.md
itowlson: If sender sets Host header to something else then we overwrite it.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,122 @@
...
+## Expressing a requirement for chaining
+
+To express that the application requires service chaining, add the following to `spin.toml`:
+
+```toml
+[application.host_requirements]
+spin_internal_requests = "required"
+```
+
+> Why ask the user to write out `"required"` instead of just listing the names?  In future, we may have features where it makes sense to say "I'd prefer to run on a host that supports this, but it's not a blocker," or where the user can specify a value (for a bad example, consider an "available memory" prerequisite).
```
itowlson, line 32: Decision: we will not surface this in the TOML, but instead infer it from the presence of `spin.internal` in `allowed_outbound_hosts`.  It will still appear in an up-level lock file.

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,122 @@
...
+The `containerd` shim uses the Spin outbound HTTP implementation, so this should Just Work, but we may need to do some analysis to ensure that it meets the scheduling requirements (all components on same node).
+
+The shim links all triggers to be in-process: I am not sure if this has any implications for the "host the HTTP engine inside the host component" strategy.
+
+# Future possibilities
+
+* In the "web of microservices," many chaining targets should be called only by other components, not by external entities.  Such "private endpoints" should be planned for even if they don't happen in the first cut of the design.
+* Service chaining opens the door to structured middleware.  A component acting as middleware would, instead of making a request to a specific endpoint, make a request to "the next component in the chain."
+* The initial proposal allows apps to express "all self-requests must be local."  This suggests two future enhancements:
+  1. Express that only certain self-requests are required to be local.  This becomes significant if the components of an app can have affinity and be scheduled to different nodes; no host supports that today.
```
itowlson, line 121: Remove

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,122 @@
...
+2. We somehow transfer the request to the HTTP trigger for handling. This is both difficult _and_ unappealing. Let's not do it.
+
+> _Additional context:_ Today, only HTTP components can make self-requests anyway (because non-HTTP components can't infer the host URL).
+
+**For the initial release we will allow chaining only from HTTP components to HTTP components.**  This does not preclude allowing chaining from non-HTTP components in a future drop.
+
+## Streaming and concurrency
+
+It's possible for a caller to be streaming a request to a callee and the callee streaming its response at the same time.  The implementation needs to allow for this: it may just fall out for free from everything on the host side being async, but we need to test and make sure the bytes keep flowing!  Chained requests will not have hyper in the flow helpfully spawning tasks so we may end up needing to do it manually - this can be deferred to implementation.
+
```
itowlson, line 101: Failure mode: if a callee traps, what happens to the caller

docs/content/sips/017-service-chaining.md
```diff
@@ -0,0 +1,112 @@
...
+# User experience
+
+## Expressing a requirement for chaining
+
+Spin will infer that an application requires service chaining if any component includes a `spin.internal` subdomain (or subdomain wildcard) in its `allowed_outbound_hosts`.
+
+This has two implications:
+
+1. The "allow all" wildcard (`*://*:*`) will not, by itself, enable service chaining.
+2. A templated host (`https://{{ accounts_service }}:*`), with the variable set to a `spin.internal` subdomain, will not, by itself, allow service chaining.
```
lann, line 30: This seems like a very uncommon scenario but as a workaround we could allow just `spin.internal` to enable service chaining without allowing any specific requests, e.g. 
```toml
allowed_outbound_hosts = ["spin.internal", "{{ maybe_internal_service }}"]
```

lann, line 30: What is the implication here? It seems like service chaining should still be _permitted_ where a templated host resolves to `spin.internal`, but it just wouldn't trigger the "must support service chaining" app validation?

docs/content/sips/017-service-chaining.md
lann: I think we should default to just returning an error to the caller. We can always allow opt-in cascading failures later if needed.

lann (approved):
One question but overall LGTM.

docs/content/sips/017-service-chaining.md
tschneidereit: agreed, yes: if we don't have something that we officially treat as a requirement for the semantics of internal service chaining, we should nevertheless still be free to apply the optimization as we see fit.

tschneidereit (approved):
I think this turned out great, and I'm strongly in favor of accepting this SIP!

docs/content/sips/017-service-chaining.md
itowlson: @lann Correct.  So a non-chaining-aware host would fail at runtime while trying to treat `spin.internal` as a routable address, but a chaining-aware one would just keep chaining away like the proverbial boss.

itowlson:
@vdice or other admin - could you force merge this please?  The required checks don't run for docs-only PRs so merging is disabled for us mortals!

PR #2295, opened 2024-02-26
update plugin install prompt to deal with line wrapping

### LINKED ISSUE #2292: remove plugin installation confirmation stutter
Every time I install a plugin, I get an installation confirmation prompt. When I select yes, the prompt is displayed again with the final confirmation. This happens when the prompt is very long. If I remove the url from the prompt string, I don't get the stutter or second prompt with confirmation. This also doesn't happen if I use a different crate for displaying the confirmation dialogue like [promptly](https://crates.io/crates/promptly). 

Screenshot 2024-02-23 at 11 47 24 AM

spin 2.3.0-pre0 (78f84e7f 2024-02-21)

Would love for us to figure out a way to remove the stutter.

karthik2804:
fixes #2292

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -644,12 +644,13 @@ fn continue_to_install(
 }
 
 fn prompt_confirm_install(manifest: &PluginManifest, package: &PluginPackage) -> Result<bool> {
-    let prompt = format!(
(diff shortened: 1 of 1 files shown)


kate-goldenring (approved):

michelleN (approved):
This looks so good!! Thank you @karthik2804 !

PR #2298, opened 2024-02-26
set host header on self outbound requests

### LINKED ISSUE #2296: Chaining HTTP self-requests gives DNS error
See Discord discussion beginning at https://discordapp.com/channels/926888690310053918/950022897160839248/1211528857572614194

If a component at `/b` makes a `self` HTTP request to `/c`, it works fine.

However, if `/a` makes a self-request to `/b` which in turn makes a self-request to `/c`, it fails with a DNS error.

Logging the `spin-full-url` at `/b`, in the first case it's normal, but in the second case it's missing a host (e.g. `http://:443/b`).  This likely messes up setting the outbound HTTP host component's `origin`, which is required for the self-request to `/c`.

cc @michelleN if you have bandwidth at the moment, otherwise I or someone else can take a look

karthik2804:
Set the host header so that more than one level of `self` requests work. Not entirely sure if this is the correct approach but fixes the test app I had calling `a -> b -> c`

fixes #2296

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -14,7 +14,7 @@ use std::{
 use anyhow::{Context, Result};
 use async_trait::async_trait;
 use clap::Args;
-use http::{uri::Scheme, StatusCode, Uri};
+use http::{header::HOST, uri::Scheme, HeaderValue, StatusCode, Uri};
 use http_body_util::BodyExt;
 use hyper::{
     body::{Bytes, Incoming},
@@ -490,6 +490,9 @@ impl OutboundWasiHttpHandler for HttpRuntimeData {
                 .parse()
                 // origin together with the path and query must be a valid URI
                 .unwrap();
+            let host = format!("{}:{}", uri.host().unwrap(), uri.port().unwrap());
+            let headers = request.request.headers_mut();
+            headers.insert(HOST, HeaderValue::from_str(&host)?);
 
             request.use_tls = uri
                 .scheme()


crates/outbound-http/src/host_impl.rs
```diff
@@ -55,15 +56,20 @@ impl outbound_http::Host for OutboundHttp {
...
 
-            let abs_url = if req.uri.starts_with('/') {
-                format!("{}{}", self.origin, req.uri)
+            let mut headers = request_headers(req.headers).map_err(|_| HttpError::RuntimeError)?;
+            let abs_url;
+            if req.uri.starts_with('/') {
+                abs_url = format!("{}{}", self.origin, req.uri);
+                // If it is a self request - set the host header.
+                let uri = Uri::from_str(&req.uri).map_err(|_| HttpError::RuntimeError)?;
+                let host = format!("{}:{}", uri.host().unwrap(), uri.port().unwrap());
```
itowlson, line 65: I'm a bit puzzled as to how this works on what seems to be the original request URI, the one that starts with `/` - I don't quite get how that would provide you with a host.  On the other hand if it works it works...!

crates/outbound-http/src/host_impl.rs
```diff
@@ -55,15 +56,20 @@ impl outbound_http::Host for OutboundHttp {
...
+                headers.insert("host", HeaderValue::from_str(&host).map_err(|_| HttpError::RuntimeError)?);
```
itowlson, line 66: Are we correct to overwrite an existing host header if there is one?

crates/trigger-http/src/lib.rs
```diff
@@ -490,6 +490,9 @@ impl OutboundWasiHttpHandler for HttpRuntimeData {
                 .parse()
                 // origin together with the path and query must be a valid URI
                 .unwrap();
+            let host = format!("{}:{}", uri.host().unwrap(), uri.port().unwrap());
+            let headers = request.request.headers_mut();
+            headers.insert(HOST, HeaderValue::from_str(&host)?);
```
itowlson, line 495: Same question about potentially overwriting an existing host header.

crates/outbound-http/src/host_impl.rs
karthik2804: I did not test the older outbound-http so I am not sure there, Will write a test for it using the JS SDK.

crates/trigger-http/src/lib.rs
karthik2804: We are only overwriting on self-requests.

itowlson: I understand that but what if some mad, bad person sends a self request and _explicitly sets the host header on it._  (I cannot imagine why they would ever do this... I just wonder if we should respect it if they do.)

karthik2804: I will check if it is a case of overwriting always or only if the user explicitly sets it. In that case, we can check for its existence before setting it.

karthik2804: By default, the host seems to be 
```
host = Some(
    ":443",
)
```

Do we want to do something like 
```
if let Some(":443") = headers.get("host") {
    headers.insert(HOST, HeaderValue::from_str(&host)?);
}
```

itowlson: Egad!  That host looks extremely weird, though I guess it's what we should expect given what we saw in the issue.  It makes me think it's already getting set _somewhere_ - and that makes me wonder if we can tweak wherever it's currently getting set rather than overwriting it later in the pipeline.  It reminded me of the issue @dicej found (https://github.com/fermyon/spin-rust-sdk/issues/8#issuecomment-1961575084) so there may be some subtle difference between main and RTM anyway...

Anyway, "user explicitly sets Host header in a self-request" certainly feels like a wildly obscure edge case.  Given that it's not trivial to determine, I'm happy to punt on it for the time being.  Thanks for the detailed investigation!

karthik2804: I'll try to dig into where the header actually gets set. I will also update the rust SDK that the app uses to  `main` instead of pinning it to `v2.2` to see if that interaction differs.

karthik2804: I did some digging and this originates from the parsing of the `uri` by `hyper`.  The first request goes through fine because the `host` is set properly on the incoming request that we use that to construct the URL and then in the subsequent requests, it fails. 

ref: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fb83cc1730a01a63f2d9c70ff95074a8

karthik2804:
The fix only seems to be needed for the `wasi_http` interface. The `spin_http` interface seems to work as expected.

itowlson (approved):
Nice - thanks!

PR #2299, opened 2024-02-27
Variables in `redis.address` and `allowed_outbound_hosts`

### LINKED ISSUE #2153: Configurable allowed outbound hosts based on application variables
I have the following setup: the host I need my application to send outbound connections to is an application variable that is set at runtime through an application variable.

Is there a way to use a (or more) variable(s) to configure an allowed outbound host?

```toml
[variables]
my_host = { default = "http://localhost" }
my_port = { default = "5003" }

[[trigger.http]]
route = "/checkout/..."
component = "checkout"

[component.checkout]
source = "checkout/target/checkout.wasm"
allowed_outbound_hosts = [
	"{{ my_host }}:{{ my_port }}",
]
```

itowlson:
Fixes #2153.

I am not sure if this is the right approach: it ends up very invasive, as the Resolver ends up being handed out to an alarmingly large number of parties.  Unfortunately, because those parties query the App metadata directly for `allowed_outbound_hosts`, I couldn't see a way to avoid this without making the App object resolver-aware, which felt a bit seismic (and caused ructions for the crate layering).

Because variable resolution is async, I experimented with letting all `allowed_outbound_hosts` checking be async, but it became painful.  Blocking during variable queries is clearly not desirable, but in practice full async is probably safe to punt to future work, at least for the CLI.

I don't yet understand the Cloud impact if we were to pursue this draft - I assume it would be ~terrifying~ non-trivial.  In particular, I'm not sure of the performance characteristics of variable queries: if those cross the network then the current blocking design is likely untenable.

Things that still need looking at:

1. Tests. So many tests
2. Whether this will play nicely with component variables (see also: 1)

Finally:

1. Some of the big changes are from splitting `spin-expressions` out of `spin-variables` - most of this is a straight copy so not as humungous as it looks!
2. I've had a lot of back and forth trying to get this working so if I've left in something badly named or badly explained then please be gentle with me, I tried to tidy it up but undoubtedly missed something...


itowlson:
Okay, I have a build that respects component variables - but now I'm thinking it's not the right thing.  Component variables are about _what the component is allowed to see,_ the view the developer gives to the guest: they implement a trust boundary.  But allow-lists aren't given to the guest, they're used by the host.  You shouldn't have to surface a piece of data to the guest just in order to use it in an allow-list.

In practice if you're configuring your allow-list from variables, you're probably handing those same URLs to the guest via component variables.  So maybe "you shouldn't have to" gets trumped by "but you always will."

Thoughts?

(Also, this reflects a limitation of the current resolver, where you can resolve _either_ at application level _or_ at component level but not both.  We could relax that limitation and give allow-list resolution access to both sets of variables.  But first I want to get a sense for what people think is the _right_ behaviour.)

lann:
I think the variables system should be redesigned. The current pull-based approach works fine (not great) for the way variables were originally designed - guests fetching values dynamically - but it is a bad fit for setting per-request behavior like allowed hosts.

Imagine a `cat_facts_url` variable is updated mid-request; first the host resolves it to `old.cat-facts.com` to set the allowed host, then the guest resolves it to `new.cat-facts.com` and the subsequent request fails. Now this particular scenario could be fixed by caching variable values for the duration of a request but there are other more subtle problems like resolving two interdependent variables at different times or performance issues with network-based variable providers as you point out above.

I think a better design would be for all relevant variables to be pre-resolved before any particular request comes in. This would be trivial for environment variables but for something like Vault it would require some kind of subscription or background refresh task. The way variables are exposed to users wouldn't need to change syntactically but the contract would change to e.g. "variable values represent a snapshot from the 'recent past'" rather than "fresh exactly when you fetch them".

itowlson:
Proposal is go with what we have (application variables, not snapshotted at start of instantiation).  If we want to allow component variables, or redesign the variables subsystem, we can do so later without impacting compatibility.

As @lann notes, this could result in inconsistency during a change, where one value is checked but a different one is used.  I don't believe this creates a security vulnerability, but it could result in spurious errors if a request is in flight at the moment the variable is changed.  So we need consensus that we are okay with that risk for the time being.  And as we both noted it won't be super performant if variables have to be fetched across the network, so we also need agreement that that's acceptable (and I admit that issue is not quantified).

If we decide we do want to do more foundational work around snapshotting, that's also fine - if so, let's make that call and we can get it under way.

itowlson:
I ignored out the Postgres integration test because the underlying one is also ignored (because flaky).  I have tested locally.

crates/outbound-networking/src/lib.rs
```diff
@@ -289,45 +326,104 @@ fn well_known_port(scheme: &str) -> Option<u16> {
...
         }
+        let allowed = if allowed
+            .iter()
+            .all(|h| matches!(h, TemplateableAllowedHost::Literal(_)))
+        {
+            let lit = AllowedHostsList::Templated(allowed, resolver.clone()).snapshot();
+            AllowedHostsList::LiteralsOnly(lit)
+        } else {
+            AllowedHostsList::Templated(allowed, resolver.clone())
+        };
```
lann, line 374: Could we get away with resolving eagerly here? Seems like it could potentially save a fair bit of complexity...

crates/trigger/src/lib.rs
```diff
@@ -369,6 +394,13 @@ impl<Executor: TriggerExecutor> TriggerAppEngine<Executor> {
             )
         })
     }
+
+    pub async fn resolve(
```
lann, line 398: Nit: maybe `resolve_template` here too, since it may not be obvious what an "engine" is "resolving".

crates/outbound-networking/src/lib.rs
itowlson: I tried that in a previous iteration and it got a bit nasty because I was trying to make `DynamicHostComponent::update_data` async to allow for it and long story short the borrow checker got mad as a wet hen.  But now I know the secret sauce of `executor::block_on` I should definitely have another look - thanks!

lann: Oh blarg no you were correct; `executor::block_on` wouldn't be good here because a slow provider would block the entire tokio task thread.

lann: I reiterate that the entire `variables` edifice ought to be burned to the ground.

itowlson: A slow provider would block the Tokio task thread anyway, when it came to checking the hosts. But maybe that's more tolerable since you're about to do networking anyway?

Also, the `network` hook accesses this stuff during store build, so store build would be slowed anyway.

* Does that change your opinion on where to resolve?
* Does it mean that burning the variables edifice to the ground needs to be a prerequisite for this work?

lann: Blocking the thread is different (worse) than stopping the task from making progress. Tokio runs async tasks on a small pool of threads. If you block one of those threads you are taking away a sizable chunk of tokio's capacity to do concurrent work.

crates/outbound-networking/src/lib.rs
```diff
@@ -30,6 +30,43 @@ pub fn check_url(url: &str, scheme: &str, allowed_hosts: &AllowedHostsConfig) ->
...
+        } else {
+            Ok(Self::Template(template))
+        }
+    }
+
+    fn snapshot(&self, resolver: &spin_expressions::Resolver) -> anyhow::Result<AllowedHostConfig> {
+        match self {
+            Self::Literal(host) => Ok(host.clone()),
+            Self::Template(template) => {
+                let resolved = futures::executor::block_on(resolver.resolve_template(template))?;
```
lann, line 54: Ah I hadn't noticed this line before in my haste to suggest that all this code go away. This line would also cause the thread to block...

crates/outbound-networking/src/lib.rs
itowlson: Okay so between this and the newly added comment at https://github.com/fermyon/spin/pull/2299#discussion_r1513572761 it sounds like this needs to either:

* Asyncify _all_ the things, every single touchpoint for this stuff; or
* Be put on hold while we rework variables

Is that correct?

crates/expressions/src/lib.rs
```diff
@@ -19,6 +19,14 @@ pub struct Resolver {
     providers: Vec<Box<dyn Provider>>,
 }
 
+#[derive(Default)]
+pub struct PreparedResolver {
+    variables: HashMap<String, String>,
+}
+
+pub type SharedPreparedResolver =
```
lann, line 27: Sharing is preparing.

crates/expressions/src/lib.rs
```diff
@@ -19,6 +19,14 @@ pub struct Resolver {
     providers: Vec<Box<dyn Provider>>,
 }
 
+#[derive(Default)]
+pub struct PreparedResolver {
```
lann, line 23: Nit-shedding (bike-knitting?): maybe `ResolvedVariables`?

lann (approved):
Thanks for wading through this morass.

crates/expressions/src/lib.rs
itowlson: I prefer a `Resolver`-based name because the purpose of it is to resolve templates; that it holds the resolved variables is an implementation detail.

itowlson:
Holding this until #2287 merges so that @suneetnangia doesn't need to rework the MQTT publisher for it.  (The changes here for MQTT will be mechanical, he said, airily tempting a jealous Fate.)

PR #2303, opened 2024-02-28
Add SIP for configuring and emitting observability

calebschoepp:
[Rendered view](https://github.com/fermyon/spin/pull/2303/files?short_path=3cef17e#diff-3cef17e9cbe59cc774f45b039267df37fb42f83537b050b5889746f240d4240a)


docs/content/sips/017-configuring-and-emitting-observability.md
```diff
@@ -0,0 +1,85 @@
...
+
+The developer must be able to configure the endpoints of OTLP compliant collectors where traces and metrics can be sent. This can be expressed in the `spin.toml` manifest.
+
+```toml
+[application]
+otlp_grpc_traces_endpoint = "http://localhost:4317"
+otlp_grpc_metrics_endpoint = "http://localhost:4317"
+```
+
+**INPUT NEEDED:** How do we want to configure these endpoints? Where do they belong in the manifest? Do we want to allow for separate endpoints for traces and metrics? Do we want to allow for choosing between gRPC and HTTP? Should we let them pass a flag to `spin` to override the manifest value?
```
lann, line 42: This probably belongs in runtime config, not the app manifest.

lann: > Do we want to allow for choosing between gRPC and HTTP?

Probably yes, and we might want to actually start with `http/protobuf`: https://opentelemetry.io/docs/specs/otel/protocol/exporter/#specify-protocol

docs/content/sips/017-configuring-and-emitting-observability.md
```diff
@@ -0,0 +1,85 @@
...
+
+The developer must be able to opt in or out of trace context propagation. This should be possible on a per component basis. By default trace context propagation is disabled for a component. This can be expressed in the `spin.toml` manifest.
+
+```toml
+[component.my-component]
+allow_trace_context_propagation = true
```
lann, line 48: This is probably worth giving its own section, e.g. `[component.my-component.tracing]`. In addition to toggling propagation we may want to have options to e.g. disable tracing entirely for particularly high-frequency components, customize span names, add metadata, etc.

Note that TOML would also let you write this as e.g.

```toml
[component.my-component]
tracing = { context_propagation = true }
```

docs/content/sips/017-configuring-and-emitting-observability.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+**INPUT NEEDED:** Is it correct to put this on the component config as opposed to the trigger config?
```
lann, line 51: There might be options that only make sense as trigger config but I think it makes sense to start at the component.

docs/content/sips/017-configuring-and-emitting-observability.md
```diff
@@ -0,0 +1,85 @@
...
+
+Improving the guest observability of Spin with WASI Observe is outside the scope of this SIP. But, it is worth noting that when we add guest observability we may want to move from a simple binary model of trace context propagation to an allow-list based system.
+
+```toml
+[component.my-component]
+allowed_trace_context_propagation_origins = ["my-component", "my-other-component"]
+```
+
+**INPUT NEEDED:** I know this is something we need but I don't actually have any clue what it would look like.
```
lann, line 60: Fine to punt on this for now I think, though we should clearly document that propagation is all-or-nothing.

docs/content/sips/017-configuring-and-emitting-observability.md
endocrimes: Link vs Trust Remote To Be An Actual Parent is one that probably makes sense trigger-wise, but would be fine component-wise

docs/content/sips/017-configuring-and-emitting-observability.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Improving observability
+
+#### Runtime observability
+
+Currently a few spans are emitted by Spin, but there is no consistent pattern. We should have all host components emit spans when they are called at the debug level. They may emit further spans at the trace level. Each host component should also emit any relevant metrics e.g. count of times called.
```
endocrimes, line 66: Spans don't actually have a concept of "levels" - Rust's `tracing` library only gives them levels because it's primarily a logging interface.

We probably _can't_ actually use Rust `tracing` for implementing wasi-observe because it requires you to specify all attributes up-front (which wasi-observe will _not_ be doing)

calebschoepp:
> Link vs Trust Remote To Be An Actual Parent is one that probably makes sense trigger-wise, but would be fine component-wise

@endocrimes what does "Link Vs Trust Remote To Be An Actual Parent" mean?

docs/content/sips/017-configuring-and-emitting-observability.md
calebschoepp: Maybe the right answer is to using `tracing_opentelemetry` for runtime observability, but then directly use the `opentelemetry` SDK for trigger and guest observability. Then you could treat them as different things easily when exporting them to places.

docs/content/sips/017-configuring-and-emitting-observability.md
calebschoepp: What does "Link Vs Trust Remote To Be An Actual Parent" mean? @endocrimes

docs/content/sips/017-configuring-and-emitting-observability.md
calebschoepp: Good call. I'll make the change and add some of the obvious config to it.

docs/content/sips/017-configuring-and-emitting-observability.md
calebschoepp: > Probably yes, and we might want to actually start with / default to http/protobuf

Sounds fine. Why do you want to default to `http/protobuf` rather than gRPC?

lann: From the link:

> If no configuration is provided the default transport SHOULD be http/protobuf unless SDKs have good reasons to choose grpc as the default (e.g. for backward compatibility reasons when grpc was already the default in a stable SDK release).

docs/content/sips/017-configuring-and-emitting-observability.md
lann: `tracing` span levels would be used to control which spans are emitted via `tracing-opentelemetry`.

docs/content/sips/017-configuring-and-emitting-observability.md
```diff
@@ -31,34 +31,32 @@ This proposal assumes that all observability data produced should be OTEL compli
...
+[observability.tracing]
+endpoint = "http://localhost:4317"
+protocol = "http/protobuf" # "grpc" or "http/protobuf" or "http/json" and it defaults to "http/protobuf"
+[observability.metrics]
+endpoint = "http://localhost:4317"
+protocol = "http/protobuf" # "grpc" or "http/protobuf" or "http/json" and it defaults to "http/protobuf"
 ```
 
-**INPUT NEEDED:** How do we want to configure these endpoints? Where do they belong in the manifest? Do we want to allow for separate endpoints for traces and metrics? Do we want to allow for choosing between gRPC and HTTP? Should we let them pass a flag to `spin` to override the manifest value?
+**BIKESHED:** `observability.tracing`, `otel.tracing`, `telemetry.tracing`....
```
lann, line 45: Maybe something like:

```toml
[telemetry.otlp]  # OTLP is the protocol name
endpoint = "..."
traces = true
metrics = true  # default false? 🤷
```

calebschoepp: Did you purposely remove the ability to configure the protocol?

lann: No, I was just trying to capture an example that covered the protocol name and toggling e.g. traces vs metrics.

calebschoepp:
I've stripped down this SIP to cover what has already been done. I'll give a few days for review and if I don't get any engagement I'll just merge it b/c the intention is basically just to track the work that has already been done.

lann (approved):

PR #2304, opened 2024-02-28
Tweak spin-loader to be able to run in Wasm

rylev:
This allows the `spin-loader` crate to run in WebAssembly. 

This should be a no-op for non-Wasm code with one exception:
* `safe_canonicalize` now always runs through `path_absolutize::Absolutize` instead of only on Windows. If there's a reason this should not happen on Unix systems, let me know.

For WebAssembly, all features are supported except for downloading wasm from source. In a future PR, we'll need to figure out an appropriate wasi friendly replacement for `reqwest`.

For I/O the WebAssembly simply uses std lib's blocking alternatives rather than the tokio APIs.

### DIFF
--- a/crates/loader/Cargo.toml
+++ b/crates/loader/Cargo.toml
@@ -17,7 +17,7 @@ lazy_static = "1.4.0"
 mime_guess = { version = "2.0" }
 outbound-http = { path = "../outbound-http", default-features = false }
 spin-outbound-networking = { path = "../outbound-networking" }
-path-absolutize = "3.0.11"
+path-absolutize = { version = "3.0.11", features = ["use_unix_paths_on_wasm"] }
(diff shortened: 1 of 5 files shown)


lann (approved):
This seems fine but I'm a little concerned about the implied maintenance burden of doing this more broadly across the spin codebase...

rylev:
> This seems fine but I'm a little concerned about the implied maintenance burden of doing this more broadly across the spin codebase...

We don't need to worry about this yet I think. I'm fine doing this on a best effort basis for now (i.e., if this regresses, I'm responsible for noticing and fixing it). This does mean that others shouldn't rely on anything being buildable in wasm (yet), and we can think through the consequences of this more thoroughly when we're ready to commit to that.

PR #2305, opened 2024-02-29
Service chaining

itowlson:
~Extremely WIP and not ready for review!  (This initial commit, for example, is just the core dispatch, none of the validation or precautions, and only minimally tested.)~

Tasks:

* Testing and (inevitably) fixing
  * Paths ✔️ 
  * Headers ✔️ 
  * Streaming bodies ✔️ 
  * Concurrency ✔️ 
* `host_requirements` section and validation
  * `spin.toml` [N/A]
  * Lockfile ✔️ 
  * Propagation ✔️ 
* Strip any `Host: *.spin.internal` header from routed requests ✔️ 
* Headers to provide on chained requests
  * Standard Spin request headers ✔️ 
  * Additional headers for chaining ~(TBD)~ [N/A]

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -12,6 +12,7 @@ use spin_locked_app::{
     values::{ValuesMap, ValuesMapBuilder},
 };
 use spin_manifest::schema::v2::{self, AppManifest, KebabId, WasiFilesMount};
+use spin_outbound_networking::SERVICE_CHAINING_DOMAIN_SUFFIX;
 use tokio::sync::Semaphore;
 
 use crate::{cache::Cache, FilesMountStrategy};
@@ -559,8 +560,10 @@ fn is_chaining_host(pattern: &str) -> bool {
     };
 
     match allowed.host() {
-        HostConfig::List(hosts) => hosts.iter().any(|h| h.ends_with(".spin.internal")),
-        HostConfig::AnySubdomain(domain) => domain == ".spin.internal",
+        HostConfig::List(hosts) => hosts
+            .iter()
+            .any(|h| h.ends_with(SERVICE_CHAINING_DOMAIN_SUFFIX)),
+        HostConfig::AnySubdomain(domain) => domain == SERVICE_CHAINING_DOMAIN_SUFFIX,
         _ => false,
     }
 }
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -79,6 +79,8 @@ impl LocalLoader {
 
         let metadata = locked_metadata(application, triggers.keys().cloned())?;
 
+        let app_requires_service_chaining = components.values().any(requires_service_chaining);
+
         let variables = variables
             .into_iter()
             .map(|(name, v)| Ok((name.to_string(), locked_variable(v)?)))
@@ -104,11 +106,27 @@ impl LocalLoader {
         }))
         .await?;
 
+        let mut host_requirements = ValuesMapBuilder::new();
+        if app_requires_service_chaining {
+            host_requirements.string(
+                spin_locked_app::locked::SERVICE_CHAINING_KEY,
+                spin_locked_app::locked::HOST_REQ_REQUIRED,
+            );
+        }
+        let host_requirements = host_requirements.build();
+
+        let mut must_understand = vec![];
+        if !host_requirements.is_empty() {
+            must_understand.push(spin_locked_app::locked::MustUnderstand::HostRequirements);
+        }
+
         drop(sloth_guard);
 
         Ok(LockedApp {
             spin_lock_version: Default::default(),
             metadata,
+            must_understand,
+            host_requirements,
             variables,
             triggers,
             components,
@@ -523,6 +541,28 @@ fn file_url(path: impl AsRef<Path>) -> Result<String> {
     let abs_path = safe_canonicalize(path)
         .with_context(|| format!("Couldn't resolve `{}`", path.display()))?;
     Ok(Url::from_file_path(abs_path).unwrap().to_string())
+}
+
+fn requires_service_chaining(component: &spin_manifest::schema::v2::Component) -> bool {
+    component
+        .normalized_allowed_outbound_hosts()
+        .unwrap_or_default()
+        .iter()
+        .any(|h| is_chaining_host(h))
+}
+
+fn is_chaining_host(pattern: &str) -> bool {
+    use spin_outbound_networking::{AllowedHostConfig, HostConfig};
+
+    let Ok(allowed) = AllowedHostConfig::parse(pattern) else {
+        return false;
+    };
+
+    match allowed.host() {
+        HostConfig::List(hosts) => hosts.iter().any(|h| h.ends_with(".spin.internal")),
+        HostConfig::AnySubdomain(domain) => domain == ".spin.internal",
+        _ => false,
+    }
 }
 
 const SLOTH_WARNING_DELAY_MILLIS: u64 = 1250;
--- a/crates/locked-app/src/locked.rs
+++ b/crates/locked-app/src/locked.rs
@@ -182,6 +182,24 @@ impl LockedApp {
     ) -> crate::Result<T> {
         self.metadata.require_typed(key)
     }
+
+    /// Checks that the application does not have any host requirements
+    /// outside the supported set. The error case returns a comma-separated
+    /// list of unmet requirements.
+    pub fn ensure_needs_only(&self, supported: &[&str]) -> Result<(), String> {
+        let unmet_requirements = self
+            .host_requirements
+            .keys()
+            .filter(|hr| !supported.contains(&hr.as_str()))
+            .map(|s| s.to_string())
+            .collect::<Vec<_>>();
+        if unmet_requirements.is_empty() {
+            Ok(())
+        } else {
+            let message = unmet_requirements.join(", ");
+            Err(message)
+        }
+    }
 }
 
 /// A LockedComponent represents a "fully resolved" Spin component.
--- a/crates/locked-app/src/locked.rs
+++ b/crates/locked-app/src/locked.rs
@@ -33,11 +33,11 @@ pub enum MustUnderstand {
     HostRequirements,
 }
 
-/// TODO:
+/// Features or capabilities the application requires the host to support.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(rename_all = "snake_case")]
 pub enum HostRequirement {
-    /// TODO:
+    /// The application requires local service chaining.
     LocalServiceChaining,
 }
 
@@ -56,11 +56,11 @@ pub struct LockedApp {
     #[serde(
         default,
         skip_serializing_if = "ValuesMap::is_empty",
-        deserialize_with = "i_cant_even"
+        deserialize_with = "deserialize_host_requirements"
     )]
     pub host_requirements: ValuesMap,
     /// Custom config variables
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    #[serde(default, skip_serializing_if = "LockedMap::is_empty")]
     pub variables: LockedMap<Variable>,
     /// Application triggers
     pub triggers: Vec<LockedTrigger>,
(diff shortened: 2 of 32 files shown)


itowlson:
@lann This has a proposal for the lockfile stuff, which declares v0 and v1 serialisation formats plus a new wrapper that handles deserialisation from either format and serialisation to the v0 or v1 format depending on whether v1 features are needed.  The wrapper format is no longer `Serialize` or `Deserialize` - this was to get around code that performed serialisation or deserialisation directly, which would have bypassed the version understander, although it could be restored as a custom implementation.  The PR as a whole isn't ready yet (and, in particular, the "must-understand" stuff isn't processed), but it would be good to know whether this aligns with your thinking and whether it's going to be friendly to Cloud.  Thanks!

lann:
I had been thinking that the version field could just be relaxed to allow 0 or 1 since the format is (syntactically) backward-compatible.

itowlson:
@lann That would work for deserialising, but would require custom serialisation.  Is that your preference?

lann:
Ugh serde does not make this easy does it?

lann:
After a closer look it does seem like custom serialization might be the least-bad option here from a maintenance perspective.

itowlson:
@lann Okay I have been exploring the custom serialisation (in a different branch, which I can share if it would help) and I am measurably dubious because it feels like either:

1. Everywhere that deserialises one of these things has to check it understands the must-understand stuff; or
2. We also have to do custom deserialisation and check the must-understand stuff within that

I don't trust myself or anybody else to do 1 correctly.  So this approach seems to leave us with custom deserialisation and serde errors, which feels like brittle work for a worse UX.  I am happy to press on with custom serialisation/deserialisation if you feel it's the right thing, but I wanted to touch base and make sure you still felt it was better than the wrapper type.

lann:
I think deserialization can be handled with a serde enum; something like:

```rust
// LockedApp {
    must_understand: Vec<MustUnderstand>,
// }

#[derive(Deserialize)
#[serde(rename_all = "snake_case")]
enum MustUnderstand {
    HostRequirements,
}
```

It also could make sense to have a custom `#[serde(deserialize_with = "deserialize_must_understand")]` for nicer error messages at some point.

itowlson:
Okay this has mostly come together except for streaming.  I have a test case where the front end streams its request to the back end, and streams the response from the back.  When I route the front-to-back request, it works.  When I chain the two, it deadlocks.

I was SO CLOSE.

itowlson:
Okay, I have two-way streaming working.  I understand why my first attempt to test it deadlocked when chained, but I don't understand why it worked when routed - so I'm still concerned, because there _is_ a behavioural difference, and that could trip folks up if they have a working streaming interaction and attempt to chain it.  Still, unblocked for now I think.

crates/loader/src/local.rs
```diff
@@ -525,6 +545,30 @@ fn file_url(path: impl AsRef<Path>) -> Result<String> {
...
+        return false;
+    };
+
+    match allowed.host() {
+        HostConfig::List(hosts) => hosts
+            .iter()
+            .any(|h| h.ends_with(SERVICE_CHAINING_DOMAIN_SUFFIX)),
+        HostConfig::AnySubdomain(domain) => domain == SERVICE_CHAINING_DOMAIN_SUFFIX,
+        _ => false,
+    }
```
lann, line 569: I think all of the service chaining host string logic could plausibly live in `spin-outbound-networking`

crates/trigger-http/src/handler.rs
```diff
@@ -26,7 +26,7 @@ pub struct HttpHandlerExecutor;
 impl HttpExecutor for HttpHandlerExecutor {
     async fn execute(
         &self,
-        engine: &TriggerAppEngine<HttpTrigger>,
+        engine: &Arc<TriggerAppEngine<HttpTrigger>>,
```
lann, line 29: You don't often see `&Arc` params; typically you would use an owned `Arc` and let the caller decide whether it needs to clone or not.

crates/trigger-http/src/lib.rs
```diff
@@ -456,10 +471,76 @@ pub(crate) trait HttpExecutor: Clone + Send + Sync + 'static {
...
+        Ok(data.table().push(HostFutureIncomingResponse::new(handle))?)
+    }
+}
+
+fn parse_chaining_target(request: &wasmtime_wasi_http::types::OutgoingRequest) -> Option<String> {
+    let host = request.request.uri().authority().map(|a| a.host().trim())?;
+
+    let (first, rest) = host.split_once('.')?;
+
+    if rest == SERVICE_CHAINING_DOMAIN {
```
lann, line 537: Ditto "maybe move to `spin-outbound-networking`"

lann (commented):
I'd be happy for someone with a better mental model of how wasi http works to review this. @dicej?

crates/locked-app/src/locked.rs
```diff
@@ -4,21 +4,61 @@ use std::path::PathBuf;
...
+/// Identifies fields in the LockedApp that the host must process if present.
 #[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum MustUnderstand {
+    /// If present in `must_understand`, the host must support all features
+    /// in the app's `host_requirements` section.
+    HostRequirements,
+}
+
+/// TODO:
```
fibonacci1729, line 36: Just flagging this in case you missed it.

crates/locked-app/src/locked.rs
```diff
@@ -28,6 +68,85 @@ pub struct LockedApp {
...
+        {
+            let mut hr = ValuesMapBuilder::new();
+
+            while let Some(key) = map.next_key::<String>()? {
+                let value: serde_json::Value = map.next_value()?;
+                if value.as_str() == Some(HOST_REQ_OPTIONAL) {
+                    continue;
+                }
+                // TODO: something better with errors, but A::Error is not helpful
+                hr.serializable(key, value).unwrap();
```
fibonacci1729, line 95: You could do something like:
```
hr.serializable(key, value).map_err(<A::Error as ::serde::de::Error>::custom)?;
```

Or if you `use serde::de::Error` then:
```
hr.serializable(key, value).map_err(A::Error::custom)?;
```

crates/trigger-http/src/lib.rs
```diff
@@ -456,10 +471,76 @@ pub(crate) trait HttpExecutor: Clone + Send + Sync + 'static {
...
+impl HttpRuntimeData {
+    fn chain_request(
+        data: &mut spin_core::Data<Self>,
+        request: wasmtime_wasi_http::types::OutgoingRequest,
+        component_id: String,
+    ) -> wasmtime::Result<
+        wasmtime::component::Resource<wasmtime_wasi_http::types::HostFutureIncomingResponse>,
+    >
+    where
+        Self: Sized,
```
fibonacci1729, line 489: You can remove the `where Self: Sized` here.

crates/loader/src/local.rs
itowlson: Aaaaaaaaaaggggghhh.  Yes, that was my first thought, but I didn't want to introduce a new dependency for the loader.  But whaddaya know, the loader already depends on `spin-outbound-networking` BUT IT'S NOT LISTED WITH ALL THE OTHER SPIN CRATES.

Ahem.  Yes, I agree.

crates/trigger-http/src/handler.rs
```diff
@@ -347,7 +347,8 @@ impl HandlerType {
 
 fn set_http_origin_from_request(
     store: &mut Store,
-    engine: &TriggerAppEngine<HttpTrigger>,
+    engine: &Arc<TriggerAppEngine<HttpTrigger>>,
+    handler: &HttpHandlerExecutor,
```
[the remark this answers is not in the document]
itowlson: We could instantiate it within the function I guess, but conceptually what this is saying is "I want to use the _exact same_ engine and executor that the HTTP trigger has set up."  That the executor has no state today is an implementation detail that this function shouldn't have to care about.

Really I should probably Arc it rather than cloning it, in case changes are made to the executor after I capture it...

crates/trigger-http/src/lib.rs
```diff
@@ -456,10 +471,76 @@ pub(crate) trait HttpExecutor: Clone + Send + Sync + 'static {
...
+            "Internal error: internal request chaining not prepared (handler not assigned)",
+        ))?;
+
+        let base = "/";
+        let raw_route = "/...";
+
+        let client_addr = std::net::SocketAddr::from_str("0.0.0.0:0")?;
+
+        let between_bytes_timeout = request.between_bytes_timeout;
+        let worker = Arc::new(wasmtime_wasi::preview2::spawn(async {}));
```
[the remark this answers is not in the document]
itowlson: It needs a worker.  In the real wasi-http implementation, that worker is the task for the TCP connection.  I don't have a TCP connection.  I don't know what this should be so I put this in and it worked.  I'll add a comment, but if the "Eh" means "this does not look right and you should be doing something else instead" then I'd be in need of guidance from you or Joel - thanks!

crates/trigger-http/src/lib.rs
```diff
@@ -456,10 +471,76 @@ pub(crate) trait HttpExecutor: Clone + Send + Sync + 'static {
...
+                Ok(resp) => Ok(Ok(IncomingResponseInternal {
+                    resp,
+                    between_bytes_timeout,
+                    worker,
+                })),
+                Err(e) => Err(wasmtime::Error::msg(e)),
+            }
+        };
+
+        let handle = wasmtime_wasi::preview2::spawn(resp_fut);
```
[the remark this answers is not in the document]
itowlson: That doesn't typecheck though.  The WASI method calls `tokio::task::spawn` wraps it in the specific kind of join handle that HostFutureIncomingResponse wants.  I can copy that wrapping if you prefer?

crates/locked-app/src/locked.rs
itowlson: BRIAN

THANK YOU

I had completely not clocked that I needed the trait to do useful things with `A::Error`.  Works like a charm now.

crates/trigger-http/src/lib.rs
itowlson: Copypasta, thanks for the spot!

crates/trigger-http/src/lib.rs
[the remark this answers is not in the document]
lann: Nope just looked odd. I don't know what `worker` is for either :shrug:

crates/trigger-http/src/lib.rs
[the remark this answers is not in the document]
lann: Ah I see. No that seems fine, just more wasmtime-wasi-http stuff I don't understand enough to adequately review :sweat_smile:

fibonacci1729 (approved):

itowlson:
@lann I'm still not happy with the LockedApp stuff.  At the moment I check host requirements during deserialisation and that just doesn't work (because the `spin_locked_app` crate doesn't know what the caller of the deserialiser actually supports).  So I am back to either:

* A wrapper type, such that consumers of LockedApp can only get at it via a function that checks requirements; or
* All consumers making sure to check that they can meet the host requirements after deserialisation.

We've shied away from the first option, which leaves us with the second.  (Or I'm missing something.  Always plausible.)  For triggers I can encapsulate the check in the trigger loader - which is probably the 90% case - but anything else will rely on the programmer knowing they need to do the check.  Are we okay with that?  Or did you already have another plan in mind for this?  _grin_

dicej (approved):
I mostly skimmed the loader bits (assuming Lann knows those best) and focused on the `trigger-http` parts.  Looks good to me, and it's actually simpler than I would have expected.  Nice work!

itowlson:
Added a "check that the app doesn't need anything I don't provide" function and implemented it in the base trigger code.  Example message: `Error: This application requires the following features that are not available in this version of the 'redis' trigger: local_service_chaining`

Note that because capabilities are checked at application level, this can misfire in mixed-trigger apps.  A mixed app whose HTTP component allows `spin.internal` but whose Redis component does not will still fail saying the app _as a whole_ needs service chaining.  This isn't ideal: we need to decide if the compatibility risk of iterating on it later justifies holding this up.

cc @lann for opinioning

itowlson:
(The alternative would be for the Redis trigger _not_ to check, which would enable the "HTTP component chains but Redis does not" scenario, but would mean the "Redis component chains" scenario would sneak through and fail when the request was made. 
 If we think that's better then I will just back out the last commit!)

PR #2311, opened 2024-02-29
fix(example): update variables example to use constant time comparison

kate-goldenring:


lann (approved):

examples/vault-variable-test/src/lib.rs
```diff
@@ -20,3 +20,23 @@ fn handle_vault_variable_test(req: Request) -> Result<impl IntoResponse> {
         .body(response_json)
         .build())
 }
+
+fn constant_time_compare(a: &str, b: &str) -> bool {
+    let mut diff = 0;
+    if a.len() != b.len() {
+        return false;
+    }
+    for (a, b) in a.bytes().zip(b.bytes()) {
```
itowlson, line 29: If we are going to get into this (and I understand why we are), then there could still be sneaky issues here with the optimiser.  Looking at the `constant_time_eq` crate, they do some sneaky shenanigans there to defeat it (https://docs.rs/constant_time_eq/latest/src/constant_time_eq/lib.rs.html#41) - should we do the same?

itowlson: Although the `black_box` docs say "As such, it must not be relied upon to control critical program behavior. This _immediately_ precludes any direct use of this function for cryptographic or security purposes." so 🤷

kate-goldenring: Since it is an example that is meant to illustrate spin rather than how to do constant time comparison, should we just add the dependency? Then, people can reference it's limitations in the library docs.

lann (approved):
👍 thanks

PR #2314, opened 2024-03-03
Subdomain wildcards

### LINKED ISSUE #2309: Subdomain wildcards
The service chaining SIP (#2290) proposes `widgets.spin.internal` as the domain name for chained requests.  To allow chaining to _any_ other component (a la `self`), it proposes allow-listing `*.spin.internal`.  However, subdomain wildcards don't currently resolve - they parse but are matched literally rather than treated as a pattern.

Should we:

1. Allow subdomain wildcards
2. Special case the chaining domain and allow wildcards only there
3. Use a different convention for chaining

itowlson:
Fixes #2309

### DIFF
--- a/crates/outbound-networking/src/lib.rs
+++ b/crates/outbound-networking/src/lib.rs
@@ -143,6 +143,7 @@ impl SchemeConfig {
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum HostConfig {
     Any,
+    AnySubdomain(String),
     ToSelf,
     List(Vec<String>),
     Cidr(ipnet::IpNet),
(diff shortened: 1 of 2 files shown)


crates/outbound-networking/src/lib.rs
```diff
@@ -172,12 +173,19 @@ impl HostConfig {
             bail!("hosts must not contain paths");
         }
 
+        if let Some((subdomain, domain)) = host.split_once('.') {
+            if subdomain == "*" {
+                return Ok(Self::AnySubdomain(format!(".{domain}")));
+            }
```
lann, line 179: Should be equivalent:
```suggestion
        if let Some(domain) = host.strip_prefix("*.") {
            return Ok(Self::AnySubdomain(format!(".{domain}")));
```

crates/outbound-networking/src/lib.rs
```diff
@@ -172,12 +173,19 @@ impl HostConfig {
...
+        }
+
```
lann, line 181: Could we validate that there is no `*` at this point?

lann (approved):

PR #2316, opened 2024-03-04
WIP: Better observability in Spin

calebschoepp:
An extremely WIP branch to improve the observability in Spin as per #2303. This description will be updated once this is ready for review. Mostly just using this to take advantage of Github PR diff view.


crates/trigger-http/src/handler.rs
```diff
@@ -38,6 +38,9 @@ impl HttpExecutor for HttpHandlerExecutor {
             component_id
         );
 
+        let span = tracing::info_span!("execute_wasm", "otel.kind" = "server");
+        let _enter = span.enter();
```
lann, line 42: https://docs.rs/tracing/latest/tracing/struct.Span.html#method.enter

> Warning: in asynchronous code that uses async/await syntax, Span::enter should be used very carefully or avoided entirely. Holding the drop guard returned by Span::enter across .await points will result in incorrect traces.

> For instrumenting asynchronous code, tracing provides the [Future::instrument combinator](https://docs.rs/tracing/latest/tracing/trait.Instrument.html) for attaching a span to a futureFor instrumenting asynchronous code, tracing provides the [Future::instrument combinator](https://docs.rs/tracing/latest/tracing/trait.Instrument.html) for attaching a span to a future

> The [#[instrument] attribute macro](https://docs.rs/tracing/latest/tracing/attr.instrument.html) can automatically generate correct code when used on an async function:

calebschoepp:
Closing in favour of #2348 for a cleaner history

PR #2318, opened 2024-03-04
feat(loader): support loading AOT compiled components

kate-goldenring:


crates/trigger/src/loader.rs
```diff
@@ -13,16 +11,22 @@ use tokio::fs;
...
 use spin_common::{ui::quoted_path, url::parse_file_url};
 
+/// Loader for the Spin runtime
 pub struct TriggerLoader {
+    /// Working directory where files for mounting exist.
     working_dir: PathBuf,
+    /// Set the static assets of the components in the temporary directory as writable.
     allow_transient_write: bool,
+    /// All components have been ahead of time (AOT) compiled (to cwasm) and should be loaded through deserialization.
+    aot: bool,
```
itowlson, line 21: The boolean here is hard to interpret at point of use (e.g. in the `TriggerLoader` constructor call or when checking `self.aot`).  (Particularly since `allow_transient_write` is also a boolean so you end up with a sequence of booleans in the constructor call and what do they mean.)  Consider an enum such as `CompilationStatus::AllAot` or `Aot::AllAotCompiled`.

(and yes, `allow_transient_writes` should have been an enum for the same reasons, but that's out of scope!)

crates/trigger/src/loader.rs
```diff
@@ -49,15 +53,22 @@ impl Loader for TriggerLoader {
...
-            format!(
-                "failed to read component source from disk at path '{}'",
-                path.display()
-            )
-        })?;
-        let component = spin_componentize::componentize_if_necessary(&bytes)?;
-        spin_core::Component::new(engine, component.as_ref())
-            .with_context(|| format!("loading module {}", quoted_path(&path)))
+        if self.aot {
+            unsafe {
```
itowlson, line 57: Should we put any assertions or validations in before calling unsafe code?

fibonacci1729: It might make sense to assert the result of `detect_precompiled` here to ensure the source does in fact actually contain precompiled bytes.

fibonacci1729 (approved):
LGTM!

crates/trigger/src/loader.rs
```diff
@@ -49,15 +66,33 @@ impl Loader for TriggerLoader {
...
-        })?;
-        let component = spin_componentize::componentize_if_necessary(&bytes)?;
-        spin_core::Component::new(engine, component.as_ref())
-            .with_context(|| format!("loading module {}", quoted_path(&path)))
+        match self.compilation_status {
+            CompilationStatus::AllAotComponents => {
+                match engine.detect_precompiled_file(&path)?{
+                    Some(Precompiled::Component) => {
+                        unsafe {
+                            spin_core::Component::deserialize_file(engine, &path)
```
kate-goldenring, line 74: Wasmtime recommends using `deserialize_file` over `deserilize` as it is faster due to [not copying around data](https://github.com/bytecodealliance/wasmtime/blob/149858b111686cc92a8da89397f4f3333023cca5/crates/wasmtime/src/runtime/module.rs#L467)

However, [`detect_precompiled_file`](https://github.com/bytecodealliance/wasmtime/blob/149858b111686cc92a8da89397f4f3333023cca5/crates/wasmtime/src/engine/serialization.rs#L162) reads in the bytes from the file anyways, so i am wondering if it is best to read the bytes and use `detect_precompiled` and `deserialize`.

fibonacci1729: Ah good catch. I think it makes sense to read the bytes and use `detect_precompiled` / `deserialize` to avoid reading the file twice as you note.

kate-goldenring: I still think what we have could be faster since `deserialize_file` mmaps so there is still only one read of bytes -- but this is nitty

itowlson: @kate-goldenring What is the implication if an attacker provides a malicious file - I assume that is why it's marked `unsafe` - e.g. could it result in memory corruption or reads or writes of other modules in the same process?  Or is it guaranteed to fail in a safe manner?

kate-goldenring: @itowlson from the Wasmtime [docs](https://github.com/bytecodealliance/wasmtime/blob/149858b111686cc92a8da89397f4f3333023cca5/crates/wasmtime/src/runtime/module.rs#L432), you're right that we need to take a step back and think about how to handle this:

```
    /// This function is only
    /// designed to receive the previous input from [`Module::serialize`] and
    /// [`Engine::precompile_module`]. If the exact output of those functions
    /// (unmodified) is passed to this function then calls to this function can
    /// be considered safe.
```

lann: You should prefer the `*_file` methods. `detect_precompiled_file` can (potentially) take advantage of `seek`ing in files and `deserialize_file` could be quite important for boot times in some situations.

Don't worry about reading the file twice; if the OS can spare the memory you will generally hit cache on the second read.

kate-goldenring: To make it clear that deserializing can be unsafe, I added another constructor behind a feature gate that enables support for precompiled components. It is also documented. The hope is that any runtime that chooses to use it will ensure they owned the compilation of the component.

crates/trigger/Cargo.toml
```diff
@@ -8,6 +8,12 @@ edition = { workspace = true }
 llm = ["spin-llm-local"]
 llm-metal = ["llm", "spin-llm-local/metal"]
 llm-cublas = ["llm", "spin-llm-local/cublas"]
+# Enables creating a TriggerLoader that loads AOT compiled components. It
+# enables using the unsafe `wasmtime::deserialize` function, which can be unsafe
```
rylev, line 12: Nit: a `wasmtime::deserialize` function doesn't exist. 

Also, I think we shouldn't necessarily mention the wasmtime function used here and instead point users to `TriggerLoader::new_aot_compiled` where we spell out all of the safety concerns.

crates/trigger/src/loader.rs
```diff
@@ -13,16 +11,46 @@ use tokio::fs;
...
 impl TriggerLoader {
     pub fn new(working_dir: impl Into<PathBuf>, allow_transient_write: bool) -> Self {
         Self {
             working_dir: working_dir.into(),
             allow_transient_write,
+            compilation_status: CompilationStatus::NoneAot,
+        }
+    }
+
+    /// Creates a TriggerLoader that loads AOT compiled components. It enables
```
rylev, line 42: I think we'll want to spell out exactly what needs to happen here for this to be safe. i.e., that the `LockedComponentSource`  passed to `Loader::load_component` references a precompiled component compiled using the same version of `spin_core` used here and the same `Engine`.

lann: The most important point to get across is that it is _completely unsafe for untrusted input_.

Technically it is "safe" for almost any non-malicious input; it will just fail to deserialize unless compiled with a compatible engine.

crates/trigger/src/loader.rs
```diff
@@ -13,16 +11,46 @@ use tokio::fs;
...
+    /// uses the unsafe `wasmtime::deserialize` function, which is only safe if
+    /// used on components serialized with `Wasmtime Module::serialize` and
+    /// `Engine::precompile_module`. See the Wasmtime documentation for more
+    /// information:
+    /// https://docs.rs/wasmtime/latest/wasmtime/struct.Module.html#method.deserialize
+    #[cfg(feature = "aot-compilation")]
+    pub fn new_aot_compiled(working_dir: impl Into<PathBuf>, allow_transient_write: bool) -> Self {
```
lann, line 49: I do think we should mark this as `unsafe`. It doesn't immediately do anything unsafe itself but its returned value has potentially unsafe behaviors. It is roughly equivalent to some wasmtime config options like https://docs.rs/wasmtime/18.0.2/wasmtime/struct.Config.html#method.cranelift_flag_enable

fibonacci1729: OoOo TIL. That's a good catch.

kate-goldenring: Thats a great example. I changed this into a `unsafe` setter that sets `CompilationStatus::AllAotComponents` instead of a constructor.

crates/trigger/src/loader.rs
kate-goldenring: I updated the comment a little, but am still only mentioning that the requirement is `wasmtime` compilation (nothing Spin specific). Also, now the function is marked `unsafe`. @rylev do these changes look appropriate?

crates/trigger/Cargo.toml
kate-goldenring: Pointed to the function docs

crates/trigger/src/loader.rs
```diff
@@ -13,16 +11,46 @@ use tokio::fs;
...
         Self {
             working_dir: working_dir.into(),
             allow_transient_write,
+            compilation_status: CompilationStatus::NoneAot,
+        }
+    }
+
+    /// Creates a TriggerLoader that loads AOT compiled components. It enables
+    /// uses the unsafe `wasmtime::deserialize` function, which is only safe if
+    /// used on components serialized with `Wasmtime Module::serialize` and
```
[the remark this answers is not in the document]
kate-goldenring: I did that because i wanted to fast track people to where the safety discussion is in the docs, but we can let them get themselves there

crates/trigger/Cargo.toml
```diff
@@ -8,6 +8,12 @@ edition = { workspace = true }
...
+# if the AOT compiled components are not trusted. See the Wasmtime documentation
+# for more information:
+# https://docs.rs/wasmtime/latest/wasmtime/struct.Module.html#method.deserialize
+aot-compilation = []
```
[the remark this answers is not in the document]
kate-goldenring: I marked the setter for the functionality as unsafe. Do you think that is sufficient or should we also set the feature name as unsafe?

crates/trigger/src/loader.rs
rylev: I don't think marking the constructor as `unsafe` is the right approach. The contract with `unsafe` is that if the caller of the function can verify certain invariants are met *at the time of the function call* then the call is safe. However, it's the call to `Loader::load_component` that needs to verify the invariant (in this case that the component source is valid to be used with `Component::deserialize`). 

`unsafe` is IMO inappropriate because the invariant that needs to be satisfied cannot be satisfied in code. It's up to whoever is running the software to ensure the invariant holds. This is why I think a feature flag that prominently displays the safety risks is the right approach.

rylev: I now see @lann's comment below pointing to similar usage in `cranelift`. However, I think this is categorically different usage. 

The usage in the `cranelift` API can still be verified as correct by users by verifying its usage is correct in combination with the rest of the usage of `cranelift` - in other words, the safety properties can be verified statically albeit manually. The usage here on the other hand, cannot be verified statically as correct - it depends on how the software is run - the same code can be correct when run in one context and incorrect when run in another. That is not what `unsafe` is meant to signal.

lann: > the safety properties can be verified statically albeit manually

By this definition I don't think `deserialize_file` would be `unsafe` either; I don't think any unsafety occurs until the component is instantiated and you certainly can't verify its safety statically...

rylev: `deserialize_file` is different in that the safety check happens at the call site of the function marked `unsafe`. You should only call `deserialize_file` if you're sure the input *to that function* meets some invariants. That's different than calling a function and promising all future input into the system as a whole meets some invariants. 

Taking a step back... we both are trying to achieve the same thing here: ensuring that users of these APIs know what the right thing to do is. I personally think the feature flag is the right approach. It says "don't build and run this software unless you know what you're doing", and I think it does it in a better way than marking the function `unsafe`. If others disagree though that's fine, I won't try to block merging this PR.

lann: I would be fine with the feature flag only if "unsafe" appears in it, but I would personally keep this `unsafe` as well (belt _and_ suspenders), just so `unsafe` appears _somewhere_ in the calling code. :shrug:

fibonacci1729: > just so unsafe appears somewhere in the calling code. 

I tend to agree with this and this was the intent originally.

kate-goldenring: We are aiming to get this in tomorrow before we can cut a release. @rylev can we leave unsafe on the setter? @lann does this LGTY?

rylev: I'm not willing to die on this hill 😉 so :shipit:. It's certainly better than *not* marking a function as `unsafe` when it clearly is.

crates/trigger/src/loader.rs
```diff
@@ -13,18 +11,47 @@ use tokio::fs;
...
     }
+
+    /// Updates the TriggerLoader to load AOT compiled components.
+    ///
+    /// # Safety
+    /// This enables using the unsafe Wasmtime `Component::deserialize` method.
+    /// The method is safe for any Wasmtime precompiled content (components
+    /// serialized with Wasmtime` Component::serialize` or
+    /// `Engine::precompile_module`). It should never be used with untrusted
+    /// input. See the Wasmtime documentation for more information:
```
rylev, line 49: This comment is confusing unless you've read the code. There's no indication of which input we're actually talking about. It would be better if this comment indicated that it is the content source of the `LockedComponentSource` input to `<TriggerLoader as Loader>::load_component` that we're concerned with. Something like:

> This method is safe if it can be guaranteed that `<TriggerLoader as Loader>::load_component` will only ever be called with a trusted `LockedComponentSource`.  

The details about `Component::deserialize` are fine to keep but should be secondary to the point above IMO

lann: I agree that this could use a little wordsmithing but most consumers (by which I just mean spin shim :sweat_smile:) probably won't have context for what `LockedComponentSource` or even `Loader::load_component` are doing.

Maybe something along the lines of:

> Updates the TriggerLoader to load AOT precompiled components
> 
> **Warning: This feature may bypass important security guarantees of the Wasmtime
> security sandbox if used incorrectly! Read this documentation carefully.**
> 
> Usually, components are compiled "just in time" from portable Wasm sources.
> This method causes components to instead be loaded as Wasmtime-precompiled
> native executable binaries. Precompiled binaries must be produced with a compatible
> Wasmtime engine using the same Wasmtime version and compiler target settings -
> typically by a host with the same processor that will be executing them. See the
> Wasmtime documentation for more information:
> https://docs.rs/wasmtime/latest/wasmtime/struct.Module.html#method.deserialize
> 
> # Safety
>
> This method is marked as `unsafe` because it enables potentially unsafe behavior if
> used to load malformed or malicious precompiled binaries. Loading sources from an
> incompatible Wasmtime engine will fail but is otherwise safe. **Precompiled binaries
> must never be loaded from untrusted sources.**

rylev: > most consumers (by which I just mean spin shim 😅) probably won't have context for what LockedComponentSource or even Loader::load_component are doing

This is true, but most Spin users don't need to make this choice. And if you're building a different consumer that would need to make this choice, you have to know these details in order to make the right choice.

Most people should read this and interpret this as just saying "here be dragons". Only those who really need to know will take the time to inform themselves about the details so they can make a decision about whether to enable this feature. How would one even know how to precompile binaries without understanding in detail about how they are eventually loaded?

kate-goldenring: I combined both of your proposed texts, ensuring @rylev's constructive comment on what exact Spin structure must be trusted (`LockedComponentSource`) owned the line of what makes it safe.

lann (approved):

radu-matei (approved):
LGTM from the perspective of a consumer (the containerd shim).

Thank you!

PR #2324, opened 2024-03-05
fix(cmd/up): --help takes precedence over --build

### LINKED ISSUE #2323: `spin up --build --help` incorrectly builds the application
When passing both `--build` and `--help` to spin up, the application builds _while also showing help output_ - you can see this in the details block below. One would expect the application to not build, but instead only show the help output 😅 (although admittedly this is an amusing issue to write).

<details>
<summary>
$ spin up --build --help
</summary>

```console
$ spin up --build --help
spin-up 
Start the Spin application

USAGE:
    spin up [OPTIONS]

OPTIONS:
        --build                 For local apps, specifies to perform `spin build` before running the
                                application [env: SPIN_ALWAYS_BUILD=]
        --direct-mounts         For local apps with directory mounts and no excluded files, mount
                                them directly instead of using a temporary directory
    -e, --env <ENV>             Pass an environment variable (key=value) to all components of the
                                application
    -f, --from <APPLICATION>    The application to run. This may be a manifest (spin.toml) file, a
                                directory containing a spin.toml file, or a remote registry
                                reference. If omitted, it defaults to "spin.toml"
    -h, --help                  
    -k, --insecure              Ignore server certificate errors from a registry
        --temp <TMP>            Temporary directory for the static assets of the components

Building component hello-starling with `npm run build`

> starling-js-app@1.0.0 build
> npx webpack --mode=production && npx mkdirp target && ./tools/componentize.sh ./dist/spin.js ./target/hello-starling.wasm

asset spin.js 3.53 KiB [compared for emit] (name: main)
./src/index.ts 3 KiB [built] [code generated]
webpack 5.90.3 compiled successfully in 628 ms
HTTP TRIGGER OPTIONS:
        --allow-transient-write
            Set the static assets of the components in the temporary directory as writable

        --cache <WASMTIME_CACHE_FILE>
            Wasmtime cache configuration file
            
            [env: WASMTIME_CACHE_FILE=]

        --disable-cache
            Disable Wasmtime cache
            
            [env: DISABLE_WASMTIME_CACHE=]

        --disable-pooling
            Disable Wasmtime's pooling instance allocator

        --follow <FOLLOW_ID>
            Print output to stdout/stderr only for given component(s)

        --key-value <KEY_VALUES>
            Set a key/value pair (key=value) in the application's default store. Any existing value
            will be overwritten. Can be used multiple times

    -L, --log-dir <APP_LOG_DIR>
            Log directory for the stdout and stderr of components. Setting to the empty string
            disables logging to disk
            
            [env: SPIN_LOG_DIR=]

        --listen <ADDRESS>
            IP address and port to listen on
            
            [default: 127.0.0.1:3000]

    -q, --quiet
            Silence all component output to stdout/stderr

        --runtime-config-file <RUNTIME_CONFIG_FILE>
            Configuration file for config providers and wasmtime config
            
            [env: RUNTIME_CONFIG_FILE=]

        --sqlite <SQLITE_STATEMENTS>
            Run a SQLite statement such as a migration against the default database. To run from a
            file, prefix the filename with @ e.g. spin up --sqlite @migration.sql

        --state-dir <STATE_DIR>
            Set the application state directory path. This is used in the default locations for
            logs, key value stores, etc.
            
            For local apps, this defaults to `.spin/` relative to the `spin.toml` file. For remote
            apps, this has no default (unset). Passing an empty value forces the value to be unset.

        --tls-cert <TLS_CERT>
            The path to the certificate to use for https, if this is not set, normal http will be
            used. The cert should be in PEM format
            
            [env: SPIN_TLS_CERT=]

        --tls-key <TLS_KEY>
            The path to the certificate key to use for https, if this is not set, normal http will
            be used. The key should be in PKCS#8 format
            
            [env: SPIN_TLS_KEY=]
```
</details>

endocrimes:
We shouldn't build applications when the user is requesting help.

I also did a quick local test to make sure a missing component binary doesn't print any new errors/cause issues when `--help`ing - and that the re-ordering is safe (`spin up --build` still works) although i assume our integration tests already somewhat cover that path.

fixes #2323

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -662,6 +662,47 @@ route = "/..."
         Ok(())
     }
 
+    #[test]
+    fn spin_up_help_build_does_not_build() -> anyhow::Result<()> {
+        let env = testing_framework::TestEnvironment::<()>::boot(
+            &testing_framework::ServicesConfig::none(),
+        )?;
+
+        // We still don't see full help if there are no components.
+        let toml_text = r#"spin_version = "1"
+name = "unbuilt"
+trigger = { type = "http" }
+version = "0.1.0"
+[[component]]
+id = "unbuilt"
+source = "fake.wasm"
+[component.build]
+command = "echo should_not_print"
+[component.trigger]
+route = "/..."
+"#;
+        env.write_file("spin.toml", toml_text)?;
+        env.write_file("fake.wasm", [])?;
+
+        testing_framework::runtimes::spin_cli::SpinCli::start(
+            &spin_binary(),
+            &env,
+            Vec::<String>::new(),
+            SpinAppType::None,
+        )?;
+
+        let mut up = std::process::Command::new(spin_binary());
+        up.args(["up", "--build", "--help"]);
+        let output = env.run_in(&mut up)?;
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        assert!(stdout.contains("--quiet"));
+        assert!(stdout.contains("--listen"));
+        assert!(!stdout.contains("should_not_print"));
+
+        Ok(())
+    }
+
     // TODO: Test on Windows
     #[cfg(not(target_os = "windows"))]
     #[test]
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -151,10 +151,6 @@ impl UpCommand {
             }
         }
 
-        if self.build {
-            app_source.build().await?;
-        }
-
         // Get working dir holder and hold on to it for the rest of the function.
         // If the working dir is a temporary dir it will be deleted on drop.
         let working_dir_holder = self.get_canonical_working_dir()?;
@@ -184,6 +180,10 @@ impl UpCommand {
                 _ = help_process.wait().await;
             }
             return Ok(());
+        }
+
+        if self.build {
+            app_source.build().await?;
         }
 
         let mut locked_app = self


fibonacci1729 (approved):

karthik2804 (approved):

tests/integration.rs
```diff
@@ -662,6 +662,47 @@ route = "/..."
...
     }
 
+    #[test]
+    fn spin_up_help_build_does_not_build() -> anyhow::Result<()> {
+        let env = testing_framework::TestEnvironment::<()>::boot(
+            &testing_framework::ServicesConfig::none(),
+        )?;
+
+        // We still don't see full help if there are no components.
+        let toml_text = r#"spin_version = "1"
```
itowlson, line 672: Bit surprised to see a v1 manifest here - we're generally using v2 now I think. (It's not a problem though, nothing cares whether it was loaded from v1 or v2.)

itowlson (commented):
As `UpCommand` has an explicit `help` field, have we considered annotating that with suitable "not at the same time as" attributes?  That might make it clearer to the user that `--build --help` is not a valid combination, rather than silently dropping the `--build` bit...?

tests/integration.rs
endocrimes: I copy pasted this from the `up --help` works test 😅

itowlson: _chuckle_ I'll transfer my surprise there, helpfully taking it out of scope for this PR _grin_

PR #2328, opened 2024-03-07
Use the bundled version of paho and remove vendored openssl

rylev:
It seems that we're building openssl from source which causes all kinds of build issues. It's not clear to me why this is needed instead of going to default route of bundling the paho C library which allows us to not have to build openssl from source.

cc @suneetnangia for insights into the original thought behind how we were building `paho-mqtt`.

### DIFF
--- a/.github/actions/spin-ci-dependencies/action.yml
+++ b/.github/actions/spin-ci-dependencies/action.yml
@@ -78,6 +78,12 @@ inputs:
     default: 'false'
     type: bool
 
+  vcpkg:
+    description: 'setup vcpkg'
+    required: false
+    default: 'false'
+    type: bool
+
 runs:
   using: "composite"
   steps:
@@ -129,3 +135,11 @@ runs:
       if: ${{ inputs.tinygo == 'true' }}
       with:
         version: ${{ inputs.tinygo-version }}
+
+    ## Install vcpgk
+    - name: "Install vcpkg"
+      run: |
+        echo "VCPKG_ROOT=$env:VCPKG_INSTALLATION_ROOT" | Out-File -FilePath $env:GITHUB_ENV -Append
+        vcpkg install openssl:x64-windows-static-md
+      shell: pwsh
+      if: ${{ inputs.vcpkg == 'true' }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -124,6 +124,7 @@ jobs:
           rust: true
           rust-wasm: true
           rust-cache: true
+          vcpkg: "${{ matrix.os == 'windows-latest' }}"
 
       - name: Cargo Build
         run: cargo build --workspace --release --all-targets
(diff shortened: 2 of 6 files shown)


suneetnangia:
@rylev thanks, initially, I had paho configured without vendored-ssl feature flag indeed. That was resulting in build failures on some platforms (cannot remember which ones exactly) in Spin GH checks. I suspected that was due to openssl dependency not present on all platforms so building that dependency along with paho seemed to have fixed it but clearly not 🤦‍♂️. Also, I notice in Spin adds[ a flag in build](https://github.com/fermyon/spin/blob/3ee82225c8186be93b6e2a4f23a21f87abdf8ac3/.github/workflows/build.yml#L66C5-L66C110) to enable vendored openssl. 
If nothing works, I can look at replacing the MQTT lib to rumqtt instead.

suneetnangia:
> @rylev thanks, initially, I had paho configured without vendored-ssl feature flag indeed. That was resulting in build failures on some platforms (cannot remember which ones exactly) in Spin GH checks. I suspected that was due to openssl dependency not present on all platforms so building that dependency along with paho seemed to have fixed it but clearly not 🤦‍♂️. Also, I notice in Spin adds[ a flag in build](https://github.com/fermyon/spin/blob/3ee82225c8186be93b6e2a4f23a21f87abdf8ac3/.github/workflows/build.yml#L66C5-L66C110) to enable vendored openssl. If nothing works, I can look at replacing the MQTT lib to rumqtt instead.

Current build failure on windows-latest was the reason for adding vendored-ssl flag i.e. "Could not find directory of OpenSSL installation, and this `-sys` crate cannot
  proceed without this knowledge. If OpenSSL is installed and this crate had
  trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
  compilation process."

rylev:
@suneetnangia it seems the issue was just that the openssl crate was expecting vcpkg to be able to help it find where openssl was installed. I changed CI to do just that, and it's passing now.

suneetnangia:
Thank you so much for this!

.github/workflows/build.yml
```diff
@@ -124,6 +124,7 @@ jobs:
           rust: true
           rust-wasm: true
           rust-cache: true
+          vcpkg: "${{ matrix.os == 'windows-latest' }}"
```
rajatjindal, line 127: we may need similar change in release.yml as well. (I have volunteered to consolidate build/release actions to use same source to be consistent with each other).

.github/actions/spin-ci-dependencies/action.yml
```diff
@@ -129,3 +135,11 @@ runs:
       if: ${{ inputs.tinygo == 'true' }}
       with:
         version: ${{ inputs.tinygo-version }}
+
+    ## Install vcpgk
```
rajatjindal, line 139: nit: this is setting up openssl using vcpgk right?

rylev: Ah yea, bad naming. Perhaps we do that as a follow up to not block this PR more?

rajatjindal (approved):

PR #2330, opened 2024-03-07
Use rumqttc instead of paho-mqtt

itowlson:
Exploring if this gets us past the OpenSSL woes.

Note this performs the release build against the PR, because the main PR build was not reliably catching errors.  This is an interim solution to get us past this woe!

### DIFF
--- a/crates/outbound-mqtt/src/lib.rs
+++ b/crates/outbound-mqtt/src/lib.rs
@@ -38,8 +38,10 @@ impl OutboundMqtt {
         keep_alive_interval: Duration,
     ) -> Result<Result<Resource<MqttConnection>, Error>> {
         Ok(async {
-            let mut conn_opts =
-                rumqttc::MqttOptions::parse_url(address).map_err(|_| Error::InvalidAddress)?;
+            let mut conn_opts = rumqttc::MqttOptions::parse_url(address).map_err(|e| {
+                tracing::error!("MQTT URL parse error: {e:?}");
+                Error::InvalidAddress
+            })?;
             conn_opts.set_credentials(username, password);
             conn_opts.set_keep_alive(keep_alive_interval);
             let (client, event_loop) = AsyncClient::new(conn_opts, MQTT_CHANNEL_CAP);
--- a/crates/outbound-mqtt/src/lib.rs
+++ b/crates/outbound-mqtt/src/lib.rs
@@ -3,7 +3,7 @@ mod host_component;
 use std::time::Duration;
 
 use anyhow::Result;
-use paho_mqtt::Client;
+use rumqttc::AsyncClient;
 use spin_core::{async_trait, wasmtime::component::Resource};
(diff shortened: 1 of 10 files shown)


itowlson:
cc @suneetnangia for visibility and because I didn't fully understand the rumqtt stuff!  E.g. do you know if I need to do anything with the EventLoop for this publish-only case

suneetnangia:
You beat me to it @itowlson :)... I was keeping this rumqttc option in my local repo if we could not get past openssl challenges in Paho. Did we not resolve that using the fix @rylev provided?

itowlson:
@suneetnangia No, unfortunately that failed on a different part of the build!  (https://github.com/fermyon/spin/actions/runs/8192690543)  It's all a bit puzzling.  But this is looking promising and I'm glad to hear you're okay with it!

fibonacci1729 (approved):

crates/outbound-mqtt/src/lib.rs
```diff
@@ -101,11 +105,11 @@ impl v2::HostConnection for OutboundMqtt {
     }
 }
 
-fn convert_to_mqtt_qos_value(qos: Qos) -> i32 {
+fn convert_to_mqtt_qos_value(qos: Qos) -> rumqttc::QoS {
```
suneetnangia, line 108: Probably we can rename this function now that we use rumqttc which takes it's own enum than standard MQTT values i.e. 0,1,2

itowlson: @suneetnangia Good point, though I'm not sure any new name would end up looking very different... _grin_

Given how long builds take, and since I really really want to see if this passes release build, would it be okay if I merged anyway?  We can come back to it in a tidying PR for sure.

suneetnangia: Sounds good.

suneetnangia:
> cc @suneetnangia for visibility and because I didn't fully understand the rumqtt stuff! E.g. do you know if I need to do anything with the EventLoop for this publish-only case

Think we will need Eventloop for publish as well e.g. when connection from the pool needs re-establishing etc, that can go in subsequent PR, I'll take a look tomorrow... also needs to update Readme on main to include outbound MQTT in the list.

itowlson:
Ah, thanks - my test involved only a single publish, so definitely didn't exercise that potential issue.  Thanks for taking the look, I really appreciate your patience with this process.

(and yep good catch on the readme, though given experience I'm going to wait and see if the release build completes...!)

suneetnangia:
> Ah, thanks - my test involved only a single publish, so definitely didn't exercise that potential issue. Thanks for taking the look, I really appreciate your patience with this process.
> 
> (and yep good catch on the readme, though given experience I'm going to wait and see if the release build completes...!)

Issue raised to track it here https://github.com/fermyon/spin/issues/2335

PR #2333, opened 2024-03-07
[Backport v2.3] Use the bundled version of paho and remove vendored openssl

adamreese:
Backporting #2328 to v2.3


itowlson (changes requested):
I don't think you need this.  It was a fix to the MQTT PR that has not been backported.

adamreese:
I add It because it was on the milestone. Happy to remove if it's not needed. https://github.com/fermyon/spin/milestone/23?closed=1

PR #2334, opened 2024-03-07
Bump version to 2.3.1

adamreese:


itowlson (approved):

kate-goldenring:
do we need to update all the versions in tests like in this release? https://github.com/fermyon/spin/pull/2260/files

vdice:
> do we need to update all the versions in tests like in this release? https://github.com/fermyon/spin/pull/2260/files

No, now that the SDKs are in their own repos, the plan is to update tests when/if there are new SDK releases, out-of-band from Spin releases.

vdice (approved):

PR #2344, opened 2024-03-11
Fixed issue around swallowing mqtt publish errors.

### LINKED ISSUE #2335: MQTT Outbound False Positive
@itowlson thanks for adding rumqtt support, I spent sometime today to look at the use of Eventloop in rumqtt in the branch #2330 we merged y'day (as we discussed).

Problem:
We swallow errors when messages are published in the code here https://github.com/fermyon/spin/blob/a8d509a37120737ee21d3b356c638798cb997b59/crates/outbound-mqtt/src/lib.rs#L94 if we do not use Eventloop.

Explanation:
Our code is publishing message to eventloop (not to MQTT broker) which always succeed, but when eventloop attempts to publish the message to MQTT broker, it can fail due various reasons e.g. incorrect password, network issues, which we do not handle. 

Our runtime test for MQTT is passing as we have correct connection parameters but it will also pass if we have incorrect connection parameters, effectively pretending it's all good, making the test moot. It will not surface connection or other failures when publishing messages to MQTT broker. [tested this negative use case].

This will need fixing in subsequent PR, I'll have a look?

suneetnangia:
Fixes #2335

### DIFF
--- a/crates/outbound-mqtt/src/lib.rs
+++ b/crates/outbound-mqtt/src/lib.rs
@@ -88,12 +88,24 @@ impl v2::HostConnection for OutboundMqtt {
         qos: Qos,
     ) -> Result<Result<(), Error>> {
         Ok(async {
-            let (client, _) = self.get_conn(connection).await.map_err(other_error)?;
+            let (client, eventloop) = self.get_conn(connection).await.map_err(other_error)?;
             let qos = convert_to_mqtt_qos_value(qos);
+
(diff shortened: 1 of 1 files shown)


suneetnangia:
@itowlson when you get a moment please. Thanks.

itowlson (approved):
Great stuff.  Thank you so much for spotting and addressing this big gap in my hasty fix.

suneetnangia:
Likewise, thanks much for helping with the PR.

Issue #2293, opened 2024-02-23
Improving observability of Spin


calebschoepp:
[Observability](https://opentelemetry.io/docs/concepts/observability-primer/#what-is-observability) is critical for a great developer experience. We should work to improve the observability of Spin, but that is a very vague statement. What exactly are we improving the observability of? Spin itself? Spin apps?

This issue is meant to act as a meta-issue that clarifies what we mean by "improving observability of Spin". It will provide a lay of the land by describing the different levels of observability within Spin that we want to improve. Other issues, SIPs, and PRs will be used to track the actual work of improving the observability and they can backlink to this meta-issue.

Before we dive in I want to note that [OpenTelemetry](https://opentelemetry.io/) has become the industry standard for observability data and is the standard we would want to conform to.

# Types of observability in Spin

I propose that there are four types of observability in Spin that we want to enable. They exist on a spectrum from host-focused to guest-focused.

## 1) Runtime observability — observing the Spin runtime itself

Developers operating Spin in a production environment want observability into the state of the Spin process itself. This would include among other things:

- Emit spans for any critical background work e.g. garbage collection.
- Emit metrics on things like connection pool sizes.
- Potentially emit Spin logs in an OTEL format?

Some notable non-requirements include:

- Emitting metrics around CPU/memory usage. These should be collected via an agent on the node.
- Emitting metrics around aggregate stats of things like request or error count. These should be created by aggregating the request observability data downstream.

## 2) Trigger observability — observing the requests made to Spin applications

Developers want observability into the requests that are made to their Spin application. This would include among other things:

- Emit span when a Spin application is triggered with metadata about the trigger event.
- Support trace context propagation from incoming headers and pass trace context propagation on outbound calls. Provide configuration in `spin.toml` to enable or disable this for security reasons.
- Emit metrics about Spin application trigger events.
- Potentially emit Spin application logs in an OTEL format? 

## 3) Component observability — observing the interaction between composed components

Developers will create their Spin applications from a composition of components. Ideally we can automatically emit spans as the component composition graph is traversed and components are executed. This would include among other things:

- Emit span when each Wasm component is executed.
- Support trace contexts between each component.

This would require upstream modifications in [Wasmtime](https://github.com/bytecodealliance/wasmtime). 

## 4) Guest observability — observing the code within the guest module

[...]

PR #2348, opened 2024-03-11
feat(*): Implement the skeleton of an OTEL observability system

calebschoepp:
## Summary

This PR introduces OTEL observability into Spin — specifically tracing. Using the nomenclature of #2293 I have added some trigger observability while also cleaning up a bit of the runtime observability. This PR **DOES NOT** provide full tracing across the entirety of the Spin codebase. That would make for a huge PR so I opted to only trace a few parts and full tracing can be added in a followup PR.

The goal of this PR is to land in a place we're happy with the DX around observability leaving room for us to improve it in future PRs.

## Testing
```bash
# Install the otel-cli for easily producing distributed traces
go install github.com/equinix-labs/otel-cli@latest

# Run Jaeger
docker run -d -p16686:16686 -p4317:4317 -p4318:4318 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/all-in-one:latest

# Run the app
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318 spin up

# Call the app with the otel-cli
OTEL_EXPORTER_OTLP_ENDPOINT=localhost:4317 otel-cli exec -- bash -c 'curl localhost:3000-v -H "traceparent: $TRACEPARENT"'
```

### DIFF
--- a/crates/llm-local/src/lib.rs
+++ b/crates/llm-local/src/lib.rs
@@ -21,6 +21,7 @@ use std::{
     sync::{Arc, Mutex},
 };
 use tokenizers::PaddingParams;
+use tracing::instrument;
 
 #[derive(Clone)]
 pub struct LocalLlmEngine {
@@ -32,6 +33,7 @@ pub struct LocalLlmEngine {
 
 #[async_trait]
 impl LlmEngine for LocalLlmEngine {
+    #[instrument(name = "llm_generate_local_inference", skip(self, prompt))]
     async fn infer(
         &mut self,
         model: wasi_llm::InferencingModel,
@@ -91,6 +93,7 @@ impl LlmEngine for LocalLlmEngine {
         Ok(response)
     }
 
+    #[instrument(name = "llm_generate_local_embeddings", skip(self, data))]
     async fn generate_embeddings(
         &mut self,
         model: wasi_llm::EmbeddingModel,
--- a/crates/telemetry/src/detector.rs
+++ b/crates/telemetry/src/detector.rs
@@ -1,0 +1,47 @@
+use std::{env, time::Duration};
+
+use opentelemetry::{Key, KeyValue, Value};
+use opentelemetry_sdk::{
+    resource::{EnvResourceDetector, ResourceDetector},
+    Resource,
+};
+
+const OTEL_SERVICE_NAME: &str = "OTEL_SERVICE_NAME";
+
+/// Custom resource detector for Spin relevant attributes service.name and service.version.
+///
+/// To set service.name this detector will first try `OTEL_SERVICE_NAME` env. If it's not available,
+/// then it will check the `OTEL_RESOURCE_ATTRIBUTES` env and see if it contains `service.name`
+/// resource. If it's also not available, it will use `spin`.
+///
+/// To set service.version, it will use the spin_version passed in new.
+#[derive(Debug)]
+pub struct SpinResourceDetector {
+    spin_version: String,
+}
+
+impl SpinResourceDetector {
+    /// Create a new instance of SpinResourceDetector.
+    pub fn new(spin_version: String) -> Self {
+        SpinResourceDetector { spin_version }
+    }
+}
+
+impl ResourceDetector for SpinResourceDetector {
+    fn detect(&self, _timeout: Duration) -> Resource {
+        let service_name = env::var(OTEL_SERVICE_NAME)
+            .ok()
+            .filter(|s| !s.is_empty())
+            .map(Value::from)
+            .or_else(|| {
+                EnvResourceDetector::new()
+                    .detect(Duration::from_secs(0))
+                    .get(Key::new("service.name"))
+            })
+            .unwrap_or_else(|| "spin".into());
+        Resource::new(vec![
+            KeyValue::new("service.name", service_name),
+            KeyValue::new("service.version", self.spin_version.clone()),
+        ])
+    }
+}
--- a/crates/telemetry/src/lib.rs
+++ b/crates/telemetry/src/lib.rs
@@ -1,0 +1,97 @@
+use std::io::IsTerminal;
+
+use opentelemetry_sdk::propagation::TraceContextPropagator;
+use tracing_subscriber::{fmt, prelude::*, registry, EnvFilter, Layer};
+
+pub mod detector;
+mod propagation;
+mod traces;
+
+pub use propagation::extract_trace_context;
+pub use propagation::inject_trace_context;
+
+/// Initializes telemetry for Spin using the [tracing] library.
+///
+/// Under the hood this involves initializing a [tracing::Subscriber] with multiple [Layer]s. One
+/// [Layer] emits [tracing] events to stderr, and another sends spans to an OTEL collector.
+///
+/// Configuration is pulled from the environment.
+pub fn init(spin_version: String) -> anyhow::Result<ShutdownGuard> {
+    // This layer will print all tracing library log messages to stderr.
+    let fmt_layer = fmt::layer()
+        .with_writer(std::io::stderr)
+        .with_ansi(std::io::stderr().is_terminal())
+        .with_filter(
+            EnvFilter::from_default_env()
+                .add_directive("wasmtime_wasi_http=warn".parse()?)
+                .add_directive("watchexec=off".parse()?),
+        );
+
+    // We only want to build the otel layer if the user passed some endpoint configuration and it wasn't explicitly disabled.
+    let build_otel_layer = !otel_sdk_disabled() && otel_enabled();
+    let otel_layer = if build_otel_layer {
+        // In this case we want to set the error handler to log errors to the tracing layer.
+        opentelemetry::global::set_error_handler(otel_error_handler)?;
+
+        Some(traces::otel_tracing_layer(spin_version)?)
+    } else {
+        None
+    };
+
+    // Build a registry subscriber with the layers we want to use.
+    registry().with(otel_layer).with(fmt_layer).init();
+
+    // Used to propagate trace information in the standard W3C TraceContext format. Even if the otel
+    // layer is disabled we still want to propagate trace context.
+    opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
+
+    Ok(ShutdownGuard)
+}
+
+/// Returns a boolean indicating if the OTEL layer should be enabled.
+///
+/// It is considered enabled if any of the following environment variables are set and not empty:
+/// - `OTEL_EXPORTER_OTLP_ENDPOINT`
+/// - `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
+/// - `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
+///
+/// Note that this is overridden if OTEL_SDK_DISABLED is set and not empty.
+fn otel_enabled() -> bool {
+    const ENABLING_VARS: &[&str] = &[
+        "OTEL_EXPORTER_OTLP_ENDPOINT",
+        "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT",
+        "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT",
+    ];
+    ENABLING_VARS
+        .iter()
+        .any(|key| std::env::var_os(key).is_some_and(|val| !val.is_empty()))
+}
+
+/// Returns a boolean indicating if the OTEL SDK should be disabled for all signals.
+///
+/// It is considered disabled if the environment variable `OTEL_SDK_DISABLED` is set and not empty.
+fn otel_sdk_disabled() -> bool {
+    std::env::var_os("OTEL_SDK_DISABLED").is_some_and(|val| !val.is_empty())
+}
+
+fn otel_error_handler(err: opentelemetry::global::Error) {
+    static FIRST_OTEL_ERROR: std::sync::Once = std::sync::Once::new();
+    FIRST_OTEL_ERROR.call_once(|| {
+        tracing::error!("There has been an error with the OpenTelemetry system, traces and metrics are likely failing to export");
+        tracing::error!("Further OpenTelemetry errors will be logged at DEBUG level")
+    });
+    tracing::debug!(?err, "OpenTelemetry error");
+}
+
+/// An RAII implementation for connection to open telemetry services.
+///
+/// Shutdown of the open telemetry services will happen on `Drop`.
+#[must_use]
+pub struct ShutdownGuard;
+
+impl Drop for ShutdownGuard {
+    fn drop(&mut self) {
+        // Give tracer provider a chance to flush any pending traces.
+        opentelemetry::global::shutdown_tracer_provider();
+    }
+}
--- a/crates/telemetry/src/traces.rs
+++ b/crates/telemetry/src/traces.rs
@@ -1,0 +1,96 @@
+use std::time::Duration;
+
+use opentelemetry_otlp::WithExportConfig;
+use opentelemetry_otlp::{OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_TRACES_ENDPOINT};
+use opentelemetry_sdk::{
+    resource::{EnvResourceDetector, TelemetryResourceDetector},
+    trace::Tracer,
+    Resource,
+};
+use tracing_opentelemetry::OpenTelemetryLayer;
+use tracing_subscriber::{EnvFilter, Layer, Registry};
+
+use crate::detector::SpinResourceDetector;
+
+/// Constructs a layer for the tracing subscriber that sends spans to an OTEL collector.
+///
+/// It pulls OTEL configuration from the environment based on the variables defined
+/// [here](https://opentelemetry.io/docs/specs/otel/protocol/exporter/) and
+/// [here](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration).
+pub(crate) fn otel_tracing_layer(
+    spin_version: String,
+) -> anyhow::Result<
+    tracing_subscriber::filter::Filtered<OpenTelemetryLayer<Registry, Tracer>, EnvFilter, Registry>,
+> {
+    let resource = Resource::from_detectors(
+        Duration::from_secs(5),
+        vec![
+            // Set service.name from env OTEL_SERVICE_NAME > env OTEL_RESOURCE_ATTRIBUTES > spin
+            // Set service.version from Spin metadata
+            Box::new(SpinResourceDetector::new(spin_version)),
+            // Sets fields from env OTEL_RESOURCE_ATTRIBUTES
+            Box::new(EnvResourceDetector::new()),
+            // Sets telemetry.sdk{name, language, version}
+            Box::new(TelemetryResourceDetector),
+        ],
+    );
+
+    // This will configure the exporter based on the OTEL_EXPORTER_* environment variables. We
+    // currently default to using the HTTP exporter but in the future we could select off of the
+    // combination of OTEL_EXPORTER_OTLP_PROTOCOL and OTEL_EXPORTER_OTLP_TRACES_PROTOCOL to
+    // determine whether we should use http/protobuf or grpc.
+    let mut exporter = opentelemetry_otlp::new_exporter().http();
+    if let Some(endpoint) = fix_endpoint_bug() {
+        exporter = exporter.with_endpoint(endpoint);
+    }
+
+    let tracer = opentelemetry_otlp::new_pipeline()
+        .tracing()
+        .with_exporter(exporter)
+        .with_trace_config(opentelemetry_sdk::trace::config().with_resource(resource))
+        .install_batch(opentelemetry_sdk::runtime::Tokio)?;
+
+    let env_filter = match EnvFilter::try_from_env("SPIN_OTEL_TRACING_LEVEL") {
+        Ok(filter) => filter,
+        // If it isn't set or it fails to parse default to info
+        Err(_) => EnvFilter::new("info"),
+    };
+
+    Ok(tracing_opentelemetry::layer()
+        .with_tracer(tracer)
+        .with_threads(false)
+        .with_filter(env_filter))
+}
+
+// This mitigation was taken from https://github.com/neondatabase/neon/blob/main/libs/tracing-utils/src/lib.rs
+//
+// opentelemetry-otlp v0.15.0 has a bug in how it uses the
+// OTEL_EXPORTER_OTLP_ENDPOINT env variable. According to the
+// OpenTelemetry spec at
+// <https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp>,
+// the full exporter URL is formed by appending "/v1/traces" to the value
+// of OTEL_EXPORTER_OTLP_ENDPOINT. However, opentelemetry-otlp only does
+// that with the grpc-tonic exporter. Other exporters, like the HTTP
+// exporter, use the URL from OTEL_EXPORTER_OTLP_ENDPOINT as is, without
+// appending "/v1/traces".
+//
+// See https://github.com/open-telemetry/opentelemetry-rust/pull/950
+//
+// Work around that by checking OTEL_EXPORTER_OTLP_ENDPOINT, and setting
+// the endpoint url with the "/v1/traces" path ourselves. If the bug is
+// fixed in a later version, we can remove this code. But if we don't
+// remember to remove this, it won't do any harm either, as the crate will
+// just ignore the OTEL_EXPORTER_OTLP_ENDPOINT setting when the endpoint
+// is set directly with `with_endpoint`.
+fn fix_endpoint_bug() -> Option<String> {
+    if std::env::var(OTEL_EXPORTER_OTLP_TRACES_ENDPOINT).is_err() {
+        if let Ok(mut endpoint) = std::env::var(OTEL_EXPORTER_OTLP_ENDPOINT) {
+            if !endpoint.ends_with('/') {
+                endpoint.push('/');
+            }
+            endpoint.push_str("v1/traces");
+            return Some(endpoint);
+        }
+    }
+    None
+}
--- a/crates/trigger-http/src/handler.rs
+++ b/crates/trigger-http/src/handler.rs
@@ -17,6 +17,7 @@ use spin_trigger::TriggerAppEngine;
 use spin_world::v1::http_types;
 use std::sync::Arc;
 use tokio::{sync::oneshot, task};
+use tracing::{instrument, Instrument};
 use wasmtime_wasi_http::{proxy::Proxy, WasiHttpView};
 
 #[derive(Clone)]
@@ -24,6 +25,7 @@ pub struct HttpHandlerExecutor;
 
 #[async_trait]
 impl HttpExecutor for HttpHandlerExecutor {
+    #[instrument(name = "execute_wasm", skip_all, fields(otel.kind = "server"))]
     async fn execute(
         &self,
         engine: Arc<TriggerAppEngine<HttpTrigger>>,
@@ -210,29 +212,40 @@ impl HttpHandlerExecutor {
             None => Handler::Latest(Proxy::new(&mut store, &instance)?),
         };
 
-        let handle = task::spawn(async move {
-            let result = match handler {
-                Handler::Latest(proxy) => {
-                    proxy
-                        .wasi_http_incoming_handler()
-                        .call_handle(&mut store, request, response)
-                        .await
-                }
-                Handler::Handler2023_10_18(proxy) => {
-                    proxy.call_handle(&mut store, request, response).await
-                }
-                Handler::Handler2023_11_10(proxy) => {
-                    proxy.call_handle(&mut store, request, response).await
-                }
-            };
+        let span = tracing::debug_span!("execute_wasi");
+        let handle = task::spawn(
+            async move {
+                let result = match handler {
+                    Handler::Latest(proxy) => {
+                        proxy
+                            .wasi_http_incoming_handler()
+                            .call_handle(&mut store, request, response)
+                            .instrument(span)
+                            .await
+                    }
+                    Handler::Handler2023_10_18(proxy) => {
+                        proxy
+                            .call_handle(&mut store, request, response)
+                            .instrument(span)
+                            .await
+                    }
+                    Handler::Handler2023_11_10(proxy) => {
+                        proxy
+                            .call_handle(&mut store, request, response)
+                            .instrument(span)
+                            .await
+                    }
+                };
 
-            tracing::trace!(
-                "wasi-http memory consumed: {}",
-                store.as_ref().data().memory_consumed()
-            );
+                tracing::trace!(
+                    "wasi-http memory consumed: {}",
+                    store.as_ref().data().memory_consumed()
+                );
 
-            result
-        });
+                result
+            }
+            .in_current_span(),
+        );
 
         match response_rx.await {
             Ok(response) => {
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -41,7 +41,7 @@ use tokio::{
     net::TcpListener,
     task,
 };
-use tracing::log;
+use tracing::{log, Instrument};
 use wasmtime_wasi_http::{body::HyperIncomingBody as Body, WasiHttpView};
 
 use crate::{handler::HttpHandlerExecutor, wagi::WagiHttpExecutor};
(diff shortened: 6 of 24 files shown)


crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Provides configuration for the telemetry system.
+///
+/// Largely consists of OpenTelemetry variables defined
+/// [here](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration)
+/// but also has some Spin specific configuration.
+#[derive(Envconfig, Debug)]
+pub struct Config {
+    /// Disable the OpenTelemetry for all signals (traces and metrics).
+    #[envconfig(from = "OTEL_SDK_DISABLED", default = "true")]
+    pub otel_sdk_disabled: bool,
```
lann, line 14: I still think this should be opt-in. See for example what containerd does: https://github.com/containerd/containerd/blob/9c77c389d897ffea482b2d56831b741aa0081eae/pkg/tracing/plugin/otlp.go#L131-L147

calebschoepp: This is opt in. It's a weird double negative to turn it on disabled-false.

lann: Ah that's...pretty weird too; sort of an "opt-out to not opting-in". That containerd approach seems better to me: an explicit otel endpoint configuration enables it but you can still override with `OTEL_SDK_DISABLED`

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// Exporter endpoint for metrics.
+    #[envconfig(from = "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT")]
+    pub otel_exporter_otlp_metrics_endpoint: Option<String>,
+
+    /// Determines the verbosity of the OpenTelemetry tracing layer.
+    ///
+    /// This is a Spin specific value. This is what allows us to have different levels of verbosity
+    /// in the fmt layer and the otel layer. The fmt layer respects `RUST_LOG` and the otel layer
+    /// respects `OTEL_TRACING_LEVEL`.
+    #[envconfig(from = "OTEL_TRACING_LEVEL", default = "info")]
```
calebschoepp, line 41: Happy to bikeshed 🚲 🏠 this name.

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[envconfig(from = "OTEL_EXPORTER_OTLP_ENDPOINT")]
+    pub otel_exporter_otlp_endpoint: Option<String>,
+
+    /// Exporter endpoint for traces.
+    #[envconfig(from = "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")]
+    pub otel_exporter_otlp_traces_endpoint: Option<String>,
+
+    /// Exporter endpoint for metrics.
+    #[envconfig(from = "OTEL_EXPORTER_OTLP_METRICS_ENDPOINT")]
+    pub otel_exporter_otlp_metrics_endpoint: Option<String>,
```
calebschoepp, line 34: We don't technically use these config values other than to see if the user has set them so that we know if we should turn on an OTEL layer or not. `tracing-otlp` automatically reads these values from the environment.

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn inner(self) -> Vec<KeyValue> {
+        self.0
+    }
+}
+
+impl std::str::FromStr for KeyValues {
+    type Err = anyhow::Error;
+
+    fn from_str(_s: &str) -> Result<Self, Self::Err> {
+        // TODO: Get a working implementation of this
```
calebschoepp, line 75: I could use help here. I was having trouble fighting the lifetimes.

crates/telemetry/src/traces.rs
```diff
@@ -0,0 +1,180 @@
...
+        .install_batch(opentelemetry_sdk::runtime::Tokio)?;
+
+    Ok(tracing_opentelemetry::layer()
+        .with_tracer(tracer)
+        .with_threads(false)
+        .with_filter(LevelFilter::from_level(Level::from_str(
+            config.otel_tracing_level.as_str(),
+        )?)))
+    // TODO: Set the filter level so only info spans from spin or spin children crates makes it
+    // TODO: Make sure it doesn't blow up to badly if the exporter doesn't exist
```
calebschoepp, line 86: I still need to do these things before we can merge this.

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,78 @@
...
+/// [here](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration)
+/// but also has some Spin specific configuration.
+#[derive(Envconfig, Debug)]
+pub struct Config {
+    /// Disables OpenTelemetry for all signals (traces and metrics).
+    #[envconfig(from = "OTEL_SDK_DISABLED", default = "false")]
+    pub otel_sdk_disabled: bool,
+
+    /// Value of the `service.name` resource attribute.
+    #[envconfig(from = "OTEL_SERVICE_NAME", default = "spin")]
```
calebschoepp, line 17: It occurs to me that we probably want the `service.name` to be whatever the user set as the app name in their manifest rather than just `Spin`. This would require us to read the manifest immediatley at the start of `spin.rs` since reloading the tracing subscriber isn't working. Are we okay with that?

Side note: How are we going to handle hot-swapping the OTEL resource in cloud?

lann: Lets hold off on setting the service.name to the app name automatically until we understand the impact on shim users.

crates/telemetry/src/config.rs
lann: We might want to prefix _this_ one with `SPIN_` since it isn't a standard OTEL env var.

crates/telemetry/src/config.rs
lann: It might be less confusing to just list the vars that can enable tracing e.g.:

```rust
fn otlp_enabled() -> bool {
    const ENABLING_VARS: &[&str] = &[
        "OTEL_EXPORTER_OTLP_ENDPOINT",
        // ...
    ];
    ENABLING_VARS.iter().any(|key|
        std::env::var_os(key).is_some_and(|val| !val.is_empty()))
}
```

crates/telemetry/src/config.rs
lann: Also I _think_ this can handle at least this var for you: https://docs.rs/opentelemetry_sdk/0.22.0/opentelemetry_sdk/resource/struct.EnvResourceDetector.html (with https://docs.rs/opentelemetry_sdk/0.22.0/opentelemetry_sdk/struct.Resource.html#method.from_detectors)

crates/telemetry/src/config.rs
lann: Could you give more context? The `set_version` thing above seems fine to me...

crates/telemetry/src/traces.rs
```diff
@@ -0,0 +1,180 @@
...
+    // appending "/v1/traces".
+    //
+    // See https://github.com/open-telemetry/opentelemetry-rust/pull/950
+    //
+    // Work around that by checking OTEL_EXPORTER_OTLP_ENDPOINT, and setting
+    // the endpoint url with the "/v1/traces" path ourselves. If the bug is
+    // fixed in a later version, we can remove this code. But if we don't
+    // remember to remove this, it won't do any harm either, as the crate will
+    // just ignore the OTEL_EXPORTER_OTLP_ENDPOINT setting when the endpoint
+    // is set directly with `with_endpoint`.
```
lann, line 60: This is good context but it breaks up the flow of reading this function. I'd recommend extracting the mitigation (and comment) into a e.g. `fix_endpoint_bug(endpoint: &mut String)` function.

crates/telemetry/src/traces.rs
```diff
@@ -0,0 +1,180 @@
...
+        .with_threads(false)
+        .with_filter(LevelFilter::from_level(Level::from_str(
+            config.otel_tracing_level.as_str(),
+        )?)))
+    // TODO: Set the filter level so only info spans from spin or spin children crates makes it
+    // TODO: Make sure it doesn't blow up to badly if the exporter doesn't exist
+}
+
+/// Injects the current W3C TraceContext into the provided request.
+pub fn inject_trace_context<'a>(req: impl Into<HeaderInjector<'a>>) {
```
lann, line 90: Might be nice to split everything from here down into a e.g. `propagation` submodule. :shrug:

crates/trigger-http/src/handler.rs
```diff
@@ -210,29 +212,40 @@ impl HttpHandlerExecutor {
...
+                };
+
+                tracing::trace!(
+                    "wasi-http memory consumed: {}",
+                    store.as_ref().data().memory_consumed()
+                );
+
+                result
+            }
+            .in_current_span(),
```
lann, line 247: Does this work properly when the debug level is disabled?

crates/telemetry/src/config.rs
rylev: You'll have to copy the string data (using something like `to_owned`) - i.e., you can't directly store parts of `s` in a `KeyValues`.

src/bin/spin.rs
```diff
@@ -42,15 +41,15 @@ async fn main() {
...
-        .init();
+    let mut telemetry_config = match spin_telemetry::config::Config::from_env() {
+        Ok(config) => config,
+        Err(err) => {
+            eprintln!("Failed to load telemetry config: {err}");
+            return Ok(());
+        }
+    };
+    telemetry_config.set_version(VERSION.to_string());
+    let _telemetry_guard = spin_telemetry::init(telemetry_config);
```
lann, line 52: Maybe I'm just forgetting how we got here but do we still need the telemetry guard? Can we set the globals instead?

lann: nm I forgot there is a global shutdown guard

crates/telemetry/src/config.rs
calebschoepp: I had a similar thought, but remembered we inject into the environment variables at `SPIN_*` so I was concerned I was invading on that namespace. Perhaps it's not that big of a deal to steal the name `OTEL_TRACING_LEVEL` from that namespace.

crates/telemetry/src/config.rs
calebschoepp: Yup, I like that more.

crates/trigger-http/src/handler.rs
calebschoepp: Yes. The trace is still properly constructed in Jaeger even when the debug spans aren't set off to Jaeger.

crates/telemetry/src/config.rs
lann: > inject into the environment variables at `SPIN_*`

Not sure what you mean here?

crates/trigger-http/src/handler.rs
lann: I'm more concerned about `tracing` losing track of the "current" span if it drops out of scope before this spawned task completes. It would be good to verify with an app that does streaming responses like `examples/wasi-http-rust-streaming-outgoing-body`.

calebschoepp:
@lann any ideas on how to test this feature?

@lann what are the steps to landing this. Any other stakeholders I should pull in for review other than you?

lann:
> @lann any ideas on how to test this feature?

It would probably be fine to do most of the e2e testing in spinkube since most users of telemetry will be there in the medium term. For Spin itself I would probably just test that things don't entirely blow up when telemetry is 1) disabled, 2) enabled without debug traces, 3) enabled with debug traces.

lann:
> > @lann what are the steps to landing this. Any other stakeholders I should pull in for review other than you?

I would suggest landing this asap after basic "things don't blow up" tests are in place. We'll want to iterate on user feedback.

crates/llm-local/src/lib.rs
```diff
@@ -32,6 +33,7 @@ pub struct LocalLlmEngine {
 
 #[async_trait]
 impl LlmEngine for LocalLlmEngine {
+    #[instrument(skip(self, prompt))]
     async fn infer(
```
rylev, line 37: I forget - by default what is the name of the span generated here? Do we want the spans for this to just say "infer"? Would be nice if it were a bit more clear from span name alone exactly what was happening. Something like "llm_generate_local_inference" and "llm_generate_local_embeddings".

calebschoepp:
> > @lann any ideas on how to test this feature?
> 
> It would probably be fine to do most of the e2e testing in spinkube since most users of telemetry will be there in the medium term. For Spin itself I would probably just test that things don't entirely blow up when telemetry is 1) disabled, 2) enabled without debug traces, 3) enabled with debug traces.

I'll obviously test those cases in Spin locally. But, I'm wondering if we have any sort of e2e testing mechanism in Spin that I could use to automate the verification of it not blowing up.

crates/llm-local/src/lib.rs
rylev: Sounds good - as I've previously said, it would be amazing if we had some sort of system for deciding how to name spans. As we add more and more of these, we're going to simply be relying on author/reviewer judgement as to what a good span name is - it would be nicer if we gave some sort of guidance.

rylev:
@calebschoepp We have [integration tests](https://github.com/fermyon/spin/blob/main/tests/integration.rs) that you might be able to cargo cult to fit your needs.

crates/telemetry/src/config.rs
lann: Ohhh; the env var provider requires a prefix of `SPIN_VARIABLE_`, not just `SPIN_`. We use lots of vars that start with `SPIN_`.

lann:
> @calebschoepp We have [integration tests](https://github.com/fermyon/spin/blob/main/tests/integration.rs) that you might be able to cargo cult to fit your needs.

Yeah you could even run jaeger as a testing-framework service and smoke test the actual span delivery.

crates/llm-local/src/lib.rs
calebschoepp: Names are updated.

To close the loop on the naming stuff: I'll flesh out the requirements for span names in my SIP.

crates/trigger-http/src/handler.rs
calebschoepp: Karthik and I spent some time playing around with the examples to make sure this was working. Most of the examples were broken, but despite that we didn't observe any issues with tracing losing track of the "current" span. Let me know if you want further investigation here.

calebschoepp:
@lann @rylev @itowlson this PR is in a place where I'm happy with it and it is ready for a final review.

crates/trigger-http/src/lib.rs
```diff
@@ -336,6 +338,22 @@ impl HttpTrigger {
...
+                            "otel.kind" = "server",
+                            "http.request.method" = %request.method(),
+                            "network.peer.address" = %addr.ip(),
+                            "network.peer.port" = %addr.port(),
+                            "network.protocol.name" = "http",
+                            "url.path" = request.uri().path(),
+                            "url.query" = request.uri().query().unwrap_or(""),
+                            "url.scheme" = request.uri().scheme_str().unwrap_or(""),
+                            "client.address" = request.headers().get("x-forwarded-for").and_then(|val| val.to_str().ok()),
+                            // TODO(Caleb): Recorded later
```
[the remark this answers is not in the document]
calebschoepp: I was picturing this as a fast follow PR. I was struggling to get this to work and didn't want to block this PR on a few span attributes.

tests/testing-framework/services/jaeger.Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM jaegertracing/all-in-one:1.55
+
```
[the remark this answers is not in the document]
calebschoepp: In my experience it seems to start fairly quickly, but I have not quantitative evidence to back that up. My assumption was that the health checks came out of the box... What would I need to do to add a health check?

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+
+/// Provides configuration for the telemetry system.
+///
+/// Largely consists of OpenTelemetry variables defined
+/// [here](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration)
+/// but also has some Spin specific configuration.
+#[derive(Debug)]
+pub struct Config {
+    /// Disables OpenTelemetry for all signals (traces and metrics).
+    pub otel_sdk_disabled: bool,
```
itowlson, line 11: I don't love the double negative here (or the "SDK") but I assume that this is forced on us by OpenTelemetry naming conventions?

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Determines the verbosity of the OpenTelemetry tracing layer.
+    ///
+    /// This is a Spin specific value. This is what allows us to have different levels of verbosity
```
itowlson, line 15: If it's Spin-specific, do we document the recognised values?  Should we reference that documentation here?  Could it be an enum?

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+    /// in the fmt layer and the otel layer. The fmt layer respects `RUST_LOG` and the otel layer
+    /// respects `SPIN_OTEL_TRACING_LEVEL`.
+    pub otel_tracing_level: String,
+
+    /// Spin version.
+    pub version: String,
```
itowlson, line 21: Is this name forced on us by OpenTelemetry, or could we rename it to something more specific such as `spin_version`?

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+        Ok(Self {
+            otel_sdk_disabled: inner.otel_sdk_disabled,
+            otel_tracing_level: inner.otel_tracing_level,
+            version,
+        })
+    }
+
+    /// Returns a boolean indicating if the OTEL layer should be enabled.
+    ///
+    /// It is considered enabled if any of the following environment variables are set:
```
itowlson, line 37: If I set one of these but also set `otel_sdk_disabled`, does the latter take precedence?

(ETA: reading elsewhere in the code, it appears yes.)

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+#[derive(Envconfig, Debug)]
+struct EnvDerivableConfig {
+    #[envconfig(from = "OTEL_SDK_DISABLED", default = "false")]
+    pub otel_sdk_disabled: bool,
+
+    #[envconfig(from = "SPIN_OTEL_TRACING_LEVEL", default = "info")]
+    pub otel_tracing_level: String,
+}
+
+// Note: Currently SPIN_OTEL_TRACING_LEVEL defaults to info and makes no distinction between spans
```
itowlson, line 63: Would the bit about defaults be more discoverable close to the `Config::otel_tracing_level` declaration?

crates/telemetry/src/detector.rs
```diff
@@ -0,0 +1,49 @@
...
+///
+/// To set service.version, it will use the version passed in new.
+#[derive(Debug)]
+pub struct SpinResourceDetector {
+    version: String,
+}
+
+impl SpinResourceDetector {
+    /// Create a new instance of SpinResourceDetector.
+    pub fn new(version: String) -> Self {
```
itowlson, line 25: Is this the Spin version?

crates/telemetry/src/detector.rs
```diff
@@ -0,0 +1,49 @@
...
+        SpinResourceDetector { version }
+    }
+}
+
+impl ResourceDetector for SpinResourceDetector {
+    fn detect(&self, _timeout: Duration) -> Resource {
+        Resource::new(vec![
+            KeyValue::new(
+                "service.name",
+                env::var(OTEL_SERVICE_NAME)
```
itowlson, line 35: This feels like a good candidate for being extracted to a variable and given a meaningful name.

crates/trigger-http/src/lib.rs
```diff
@@ -336,6 +338,22 @@ impl HttpTrigger {
                     TokioIo::new(stream),
                     service_fn(move |request| {
                         let self_ = self_.clone();
+                        let span = tracing::info_span!(
+                            "handle_http_request",
```
itowlson, line 342: Do we have a convention on where these names come from?  I've seen a few go by and not been quite sure how they relate to function names or discrete activities within a function.  It's tricky I know because something like this is a lambda and has no name of its own...

crates/trigger-http/src/lib.rs
```diff
@@ -336,6 +338,22 @@ impl HttpTrigger {
...
+                            "network.protocol.name" = "http",
+                            "url.path" = request.uri().path(),
+                            "url.query" = request.uri().query().unwrap_or(""),
+                            "url.scheme" = request.uri().scheme_str().unwrap_or(""),
+                            "client.address" = request.headers().get("x-forwarded-for").and_then(|val| val.to_str().ok()),
+                            // TODO(Caleb): Recorded later
+                            // "error.type" = Empty,
+                            // "http.response.status_code" = Empty,
+                            // "http.route" = Empty,
+                        );
```
itowlson, line 356: This 15-line code block dominates the lambda: it no longer stands out as "the service fn takes a request and handles it" but more "the service fn takes a request and does a metric butt-ton of tracing on it" - you have to read past a _lot_ of tracing fields to get to the actual behaviour.  Could this be encapsulated?

src/bin/spin.rs
```diff
@@ -42,15 +41,14 @@ async fn main() {
...
-                .add_directive("wasmtime_wasi_http=warn".parse()?)
-                .add_directive("watchexec=off".parse()?),
-        )
-        .with_ansi(std::io::stderr().is_terminal())
-        .init();
+    let telemetry_config = match spin_telemetry::config::Config::from_env(VERSION.to_string()) {
+        Ok(config) => config,
+        Err(err) => {
+            eprintln!("Failed to load telemetry config: {err}");
+            return Ok(());
```
itowlson, line 48: This exits with success if it fails to load the telemetry config: this means that scripts that rely on the exit code will assume that Spin ran okay.  It should either continue without telemetry, or return an error (which Rust will turn into exit code 1).

itowlson (changes requested):
90% of my comments are nits or sylistic fretting, and are non-blockers.  The one thing I'm concerned about is the error/exit behaviour in main.  If I've misread that (which is entirely likely!), let me know and I'll change to an approve!

Thank you for labouring through this.  It's a huge plumbing task, but super useful.

src/bin/spin.rs
itowlson: Also, to print an error explicitly, prefer `terminal::error!` unless there is a compelling reason not to - this provides consistent formatting across the Spin UI.

crates/telemetry/src/config.rs
calebschoepp: Correct, this is forced on us by OTEL [naming conventions](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration).

crates/telemetry/src/config.rs
calebschoepp: The underlying thing this maps to is `system.version` which is forced on us by OTEL. But, this we could certainly give a nicer name like `spin_version`. Will do.

crates/telemetry/src/detector.rs
calebschoepp: Yep, I'll rename.

src/bin/spin.rs
calebschoepp: This wasn't intentional to quell the error. I'll make the updates you suggested.

crates/trigger-http/src/lib.rs
calebschoepp: Not really 🤷. Ryan has already raised the great point that we should have some kind of convention around these span names. Right now the best we have is something along the lines of:

> Make sure a span name can be understood without any additional surrounding context e.g. `generate_local_llm_inference` rather than `infer`.

I'm hoping to bake this into the SIP I wrote.

crates/telemetry/src/config.rs
calebschoepp: Good call. I changed this to just directly use the underlying `tracing::Level` type that we parse it into.

src/bin/spin.rs
```diff
@@ -44,8 +44,7 @@ async fn _main() -> anyhow::Result<()> {
     let telemetry_config = match spin_telemetry::config::Config::from_env(VERSION.to_string()) {
         Ok(config) => config,
         Err(err) => {
-            eprintln!("Failed to load telemetry config: {err}");
-            return Ok(());
+            return Err(anyhow::anyhow!("Failed to load telemetry config: {err}"));
```
itowlson, line 47: A shortcut if you want it:

```suggestion
            anyhow::bail!("Failed to load telemetry config: {err}"));
```

(definitely not important)

itowlson (approved):
Yay!

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// standard. This is what allows us to have different levels of verbosity in the fmt layer and
+    /// the otel layer. The fmt layer respects `RUST_LOG` and the otel layer respects
+    /// `SPIN_OTEL_TRACING_LEVEL`.
+    ///
+    /// Note: Currently SPIN_OTEL_TRACING_LEVEL defaults to info and makes no distinction between
+    /// spans from Spin crates or external dependencies. In the future we may desire to provide a
+    /// more curated experience of what spans show by default. This curated experience could be
+    /// created by annotating every Spin span with a special attribute that can be filtered on. Then
+    /// if the user opts to set SPIN_OTEL_TRACING_LEVEL this curated filter would be turned off and
+    /// we would allow spans from any library at the level defined by the user.
```
lann, line 26: I believe we could also use the same fmt filtering that `RUST_LOG` uses: https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/filter/struct.EnvFilter.html

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /// Returns a boolean indicating if the OTEL layer should be enabled.
+    ///
+    /// It is considered enabled if any of the following environment variables are set:
+    /// - `OTEL_EXPORTER_OTLP_ENDPOINT`
+    /// - `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
+    /// - `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT`
+    ///
+    /// Note that this is overridden if OTEL_SDK_DISABLED is set to true.
```
lann, line 51:
```suggestion
    /// Note that this is overridden if OTEL_SDK_DISABLED is set and not empty.
```

crates/telemetry/src/config.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// EnvDerivableConfig represents the elements of Config that are derived from environment variables.
+#[derive(Envconfig, Debug)]
+struct EnvDerivableConfig {
+    #[envconfig(from = "OTEL_SDK_DISABLED", default = "false")]
+    pub otel_sdk_disabled: bool,
+
+    #[envconfig(from = "SPIN_OTEL_TRACING_LEVEL", default = "info")]
+    pub otel_tracing_level: Level,
+}
```
lann, line 72: If this is all we're using `envconfig` for I'd suggest dropping it and using `std::env::var` directly in `from_env`.

crates/telemetry/src/detector.rs
```diff
@@ -0,0 +1,47 @@
...
+            .map(Value::from)
+            .or_else(|| {
+                EnvResourceDetector::new()
+                    .detect(Duration::from_secs(0))
+                    .get(Key::new("service.name"))
+            })
+            .unwrap_or_else(|| "spin".into());
+        Resource::new(vec![
+            KeyValue::new("service.name", service_name),
+            KeyValue::new("service.version", self.spin_version.clone()),
```
lann, line 44: It would be pretty weird for someone to set service.name to e.g. `MyApp` and for the version to be Spin's version, i.e. "MyApp version 0.1" would effectively look like "MyApp version 2.4.0" in spans.

Maybe we should be setting [`process.runtime.name` and `.version`](https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/) for Spin instead?

crates/telemetry/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+        tracing::error!("Further OpenTelemetry errors will be logged at DEBUG level")
+    });
+    tracing::debug!(?err, "OpenTelemetry error");
+}
+
+/// An RAII implementation for connection to open telemetry services.
+///
+/// Shutdown of the open telemetry services will happen on `Drop`.
+#[must_use]
+pub struct ShutdownGuard(Option<WorkerGuard>);
```
lann, line 65: It doesn't look like the `Option<WorkerGuard>` is ever used?

crates/telemetry/src/detector.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl SpinResourceDetector {
+    /// Create a new instance of SpinResourceDetector.
+    pub fn new(spin_version: String) -> Self {
+        SpinResourceDetector { spin_version }
+    }
+}
+
+impl ResourceDetector for SpinResourceDetector {
```
lann, line 30: This seems to overlap with `EnvResourceDetector`. Is the point here just to set the default `service.name = spin`? If so, I suspect you could get the same effect with less code by manually creating a Resource in `otel_tracing_layer` and then [`merge`](https://docs.rs/opentelemetry_sdk/0.22.1/opentelemetry_sdk/struct.Resource.html#method.merge)ing in the detected resource attrs.

src/bin/spin.rs
lann: Also, would `.context("Failed to load telemetry config")` be appropriate here?

tests/testing-framework/services/jaeger.Dockerfile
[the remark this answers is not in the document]
lann: Something like this:

```
HEALTHCHECK --interval=2s CMD wget -qO - http://localhost:14269/health
```

crates/telemetry/src/config.rs
calebschoepp: We could certainly use the EnvFilter if we wanted. I opted for the LevelFilter b/c I wanted to expose less knobs to the end user. But, my whole note there is basically saying "hey eventually we'll want to not just expose LevelFilter but actually the whole EnvFilter". So given that EnvFilter is capable of acting like a LevelFilter too I think you're right that is best to just switch to using an EnvFilter.

crates/telemetry/src/detector.rs
calebschoepp: Agreed about that being weird. But, the user can also set the version if they're setting a custom name e.g. `OTEL_RESOURCE_ATTRIBUTES=service.version=78 OTEL_SERVICE_NAME=mycustomname`. I don't think it is un-reasonable to expect them to do that if they want a custom name. I also think it is important to provide the Spin version out of the box.

I think we should switch to putting this info in `process.runtime.{name, version}` like you suggest in a future PR when we start supporting automatically setting the `service.name` based off the manifest application name. Then we can also automatically set `service.version` based off the manifest application version.

crates/telemetry/src/detector.rs
calebschoepp: It also sets the version.

That's a good idea, but I think I prefer keeping it in the detector. Inlining it would remove the need for a struct definition and a constructor, but otherwise most of the code would remain. I prefer following the pattern of a detector.

lann: That's fine - and this certainly isn't a blocking comment - but if its effectively setting a default and a hard-coded version I might be tempted to argue that it isn't "detecting" anything. :slightly_smiling_face:

crates/trigger-http/src/lib.rs
calebschoepp: I'm sure it can, but I'm scared to touch code that is this critical to Spin right before I merge this. I'll make this stylistic change in a fast-follow PR when I come back to add these attributes that are added later.

lann: If you extracted it into a function then the call site information would be wrong. You could extract it into a macro instead if you want.

lann: As a further tangent, I just learned that you can use constants as the field names, such as those from [`opentelemetry-semantic-conventions`](https://docs.rs/opentelemetry-semantic-conventions/latest/opentelemetry_semantic_conventions/trace/index.html); `tracing` just requires that you wrap constants in `{}`, e.g.:

```rust
mod keep_all_these_constants_hidden_away {
  use opentelemetry_semantic_conventions::trace::*;
  ...
    span!(
      ...
      { HTTP_REQUEST_METHOD } = method,
    )
}
```

lann: (but agreed this can come in a followup, if at all)

PR #2350, opened 2024-03-12
handle connection errors on redis-trigger init

### LINKED ISSUE #2347: Redis trigger exits on connection failure without any diagnostics
When I run a Redis application, I normally see something like this:

```
ivan@hestia:~/testing$ spin up
Active Channels on redis://localhost:6379:
        redis://localhost:6379:msgs: [redis-test]
```

However, if Redis is not running then I see:

```
ivan@hestia:~/testing$ spin up
ivan@hestia:~/testing$
```

I expected it would print an error message telling me which connection(s) failed.

karthik2804:
Handles the errors returned from the futures so that the user is informed as to the reason of why the trigger did not start. 

fixes #2347

### DIFF
--- a/crates/trigger-redis/src/lib.rs
+++ b/crates/trigger-redis/src/lib.rs
@@ -99,10 +99,11 @@ impl TriggerExecutor for RedisTrigger {
             .collect();
 
         // wait for the first handle to be returned and drop the rest
-        let (_, _, rest) = futures::future::select_all(tasks).await;
+        let (result, _, rest) = futures::future::select_all(tasks).await;
+
         drop(rest);
 
-        Ok(())
+        result?
     }
 }
 


crates/trigger-redis/src/lib.rs
```diff
@@ -99,7 +99,11 @@ impl TriggerExecutor for RedisTrigger {
             .collect();
 
         // wait for the first handle to be returned and drop the rest
-        let (_, _, rest) = futures::future::select_all(tasks).await;
+        let (result, _, rest) = futures::future::select_all(tasks).await;
+        if let Err(err) = result? {
+            eprintln!("Error: {err:?}");
```
itowlson, line 104: Prefer `terminal::error!` - it will provide you with Spin-standard colourisation.

Alternatively, and perhaps simpler, return the error from `run`.  That should bubble up to the top level, so Rust will output it and produce the appropriate exit code.

itowlson, line 104: If you don't adopt the "return it from `run`" suggestion, you need to return an error or force an exit-unwind here with a non-zero exit code.

itowlson (changes requested):

crates/trigger-redis/src/lib.rs
karthik2804: I like this better! will update.

crates/trigger-redis/src/lib.rs
```diff
@@ -99,9 +99,14 @@ impl TriggerExecutor for RedisTrigger {
...
 
         // wait for the first handle to be returned and drop the rest
-        let (_, _, rest) = futures::future::select_all(tasks).await;
+        let (result, _, rest) = futures::future::select_all(tasks).await;
+
         drop(rest);
 
+        if let Err(err) = result? {
+            return Err(err);
+        }
```
itowlson, line 108: I am guessing this stanza and the next line could be replaced by something like

```
result?
```

Or maybe only this stanza and it's something like `result??;`

Just seems little point decomposing the Err case just to put it back together again.

karthik2804: Yep changed it to that.

itowlson (approved):
❤️ glad this ended up so simple!

PR #2357, opened 2024-03-13
Because I can never remember if it's `--temp` or `--tmp`

itowlson:
Every time, I type `spin up --tmp`.  EVERY TIME.

Now, finally, I will be able to get it right first time.


fibonacci1729:
SAME. THANK YOU

PR #2358, opened 2024-03-13
remove `pip3 install componentize-py` from `http_python_template_smoke_test`

dicej:
It was redundant and contradictory since the `requirements.txt` specifies to install a specific version, which is not necessarily the latest version.


dicej:
Looks like `pip3` installs `componentize-py` to `./bin/componentize-py`, which isn't in PATH, hence the CI failure.  @karthik2804 is going to look into addressing that.

karthik2804:
superseded by #2360

PR #2360, opened 2024-03-13
add ability to specify env vars for spin build in test runner

karthik2804:
Adds the ability to specify environment variables as the `http_python_template_smoke_test` needs a couple.

### DIFF
--- a/tests/testcases/mod.rs
+++ b/tests/testcases/mod.rs
@@ -1,5 +1,5 @@
 use anyhow::Context;
-use std::path::PathBuf;
+use std::{collections::HashMap, path::PathBuf};
 
 /// Run an integration test
 pub fn run_test(
@@ -161,6 +161,7 @@ pub fn http_smoke_test_template(
     template_url: Option<&str>,
     plugins: &[&str],
     prebuild_hook: impl FnOnce(&mut testing_framework::TestEnvironment<()>) -> anyhow::Result<()>,
+    build_env_vars: HashMap<String, String>,
     expected_body: &str,
 ) -> anyhow::Result<()> {
     http_smoke_test_template_with_route(
@@ -168,6 +169,7 @@ pub fn http_smoke_test_template(
         template_url,
         plugins,
         prebuild_hook,
+        build_env_vars,
         "/",
         expected_body,
     )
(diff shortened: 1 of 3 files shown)


tests/testcases/mod.rs
```diff
@@ -321,7 +327,23 @@ pub fn bootstrap_smoke_test(
...
+            let mut custom_path = value.to_owned();
+            if value.starts_with('.') {
+                let current_dir = env.path();
+                custom_path = current_dir
+                    .join(value)
+                    .to_str()
+                    .unwrap_or_default()
+                    .to_owned();
+            }
+            build.env(key, format!("{}:{}", custom_path, path));
```
dicej, line 342: I think this needs to be `"{};{}"` on Windows.

dicej: The existing expression above your changes (`format!("{path}:{}", spin_binary().parent().unwrap().display())`) should probably also be changed.  Or is this code never intended to run on Windows?

karthik2804: I think getting the integration tests to be cross-platform is something on the backlog. @rylev would be the best person for that question.

dicej (approved):
LGTM.  Thanks for doing this.

tests/testcases/mod.rs
itowlson, line 342: This works today because `build.env("PATH")` has not been modified since it was set to `path`, but feels a bit fragile.  If you want to prepend to the command PATH, it would be good to use the command PATH as the source of truth.  (I don't know if this will ever be an issue in practice though!)

tests/testcases/mod.rs
```diff
@@ -321,7 +327,23 @@ pub fn bootstrap_smoke_test(
...
     let mut build = std::process::Command::new(spin_binary());
     // Ensure `spin` is on the path
-    build.env("PATH", path).args(["build"]);
+    build.env("PATH", &path).args(["build"]);
+    build_env_vars.iter().for_each(|(key, value)| {
+        if key == "PATH" {
+            let mut custom_path = value.to_owned();
+            if value.starts_with('.') {
+                let current_dir = env.path();
+                custom_path = current_dir
```
itowlson, line 336: Consider pulling this out into a named function - it requires close reading of this code to see what's going on.  E.g.

```
let custom_path = env.absolutise_path(value);
```

itowlson:
ONCE AGAIN IVAN CLICKS THE REVIEW BUTTON MERE SECONDS AFTER KARTHIK CLICKS THE MERGE BUTTON.  How do you _do_ that?  _grin_

PR #2362, opened 2024-03-13
Improved rumqttc event loop check.

suneetnangia:


itowlson (approved):

fibonacci1729 (approved):

itowlson:
The CI failure in the plugins test is because an external web server is returning a different error from the one we expected; it has nothing to do with this PR and can be overridden (assuming there are no other failures)

itowlson:
@suneetnangia Is this intentionally still draft?  No rush, just checking that it's deliberate.

suneetnangia:
> @suneetnangia Is this intentionally still draft? No rush, just checking that it's deliberate.
We can merge now, I was trying to see if there's any better way, will raise subsequent PR as needed but for now it brings stability to the publish part.

itowlson:
@vdice Can you force merge this please?  The CI error is unrelated to the PR.

suneetnangia:
Since our last comment here, I've tested a better solution (as I alluded to) with a few thousand messages (100k or so). Update on it's way.

crates/outbound-mqtt/src/lib.rs
```diff
@@ -44,10 +44,28 @@ impl OutboundMqtt {
             })?;
             conn_opts.set_credentials(username, password);
             conn_opts.set_keep_alive(keep_alive_interval);
-            let (client, event_loop) = AsyncClient::new(conn_opts, MQTT_CHANNEL_CAP);
+            let (client, mut event_loop) = AsyncClient::new(conn_opts, MQTT_CHANNEL_CAP);
+
+            // Start the task to poll the event loop which sends the messages to the MQTT broker.
+            // Publisher client code wait for this task to complete before returning, ensuring all events are processed.
```
itowlson, line 50: I'm not quite getting how this comment applies.  I see the Tokio task being created but not how it's being awaited or errors propagated?

itowlson:
@vdice sorry, please hold off, still some changes happening after all

crates/outbound-mqtt/src/lib.rs
suneetnangia: ah that's for the client of this outbound crate i.e. Spin Rust SDK example waits for this task to complete. Can rephrase if you want me to...

itowlson: I don't understand how though.  The async block spawns a new task for the loop, but that task seems to just run until the connection fails.  It's never returned to the guest... in fact it's in `Connection::open` not even in the publish method...?

Or are you saying that this is what _allows_ `client.publish_bytes` (which _is_ called in the publish method) to return the error - the guest doesn't await this task directly, but `client.publish_bytes` internally sort of does?

suneetnangia: or completely remove it if that's better?

itowlson: Yeah I would maybe remove the reference to the publisher client code waiting for this task.  Maybe just comment that this needs to run in the background to propagate errors to the guest.

I would still like to better understand what is happening here though.  Like how that error gets to the guest.  I understood it when you polled the event loop after doing the send, but I feel a bit lost now.  I'm sirry if I'm being obtuse!

suneetnangia: Sure, let me update comment.
Having eventloop in `publish` like in my previous commit results in messages being dropped without raising error, when throughput is high i.e. 200k messages in my test. Plus it has it's own challenges i.e. we exit the eventloop on publish event but this event is not necessarily correlated to that publish call always, it can be from another parallel call going on, as eventloop is shared across client. Plus, error in one `publish` call (if 1000s are going on) can shift the publish event ack correlation for their respective calls.

Now that was the reason to try another approach (eventloop in connection):
When eventloop returns with error (e.g. due to connectivity), `client.publish_bytes` in unable to connect to it, so `client.publish_bytes` throws error, this error is then propagated to guest in `publish`. Guest can then execute `open` to reconnect, not ideal but allows us to have high throughput with connection feedback to guest. Now I was testing with with my 200K messages and it does as above but in my further testing today it's not working as expected with a smaller no of messages which is a bummer, I suspect that is because eventloop's takes time to realise connection issues to throw the error but by that time client.publish_bytes has already published a few messages. 

In main, MQTT publish currently does not work, so one option would be to use "eventloop in `publish`" approach (first) for now to get it working and find another solution which is better than both of the above? 
Apols, been juggling a few things, far from ideal.

crates/outbound-mqtt/src/lib.rs
```diff
@@ -97,16 +97,18 @@ impl v2::HostConnection for OutboundMqtt {
...
+            // We may revisit this later to manage long running connections, high throughput use cases and their issues in the connection pool.
+            loop {
+                let event = eventloop
+                    .poll()
+                    .await
+                    .map_err(|err| v2::Error::ConnectionFailed(err.to_string()))?;
+
+                if let rumqttc::Event::Outgoing(rumqttc::Outgoing::Publish(_)) = event {
+                    return Ok(());
+                }
```
karthik2804, line 110: I think we will need this part to have a bit more logic depending on the QoS of the published message. We may need to also check for `PubAck` if Qos level is 1. If QoS level is 2, then we need to wait for `PubComp`.

fibonacci1729: Would something like this work here?

```
use rumqttc::{QoS, Event, Outgoing, Incoming};

loop {
    let event = eventloop
        .poll()
        .await
        .map_err(|err| v2::Error::ConnectionFailed(err.to_string()))?;

    match (qos, event) {
        (QoS::AtMostOnce, Event::Outgoing(Outgoing::Publish(_))) |
        (QoS::AtLeastOnce, Event::Incoming(Incoming::PubAck(_))) |
        (QoS::ExactlyOnce, Event::Outgoing(Outgoing::PubComp(_))) => break,

        (_, _) => continue,
    }
}
```

suneetnangia: ...will have a look at it this weekend, test, update and create an issue to track parallel requests issue (para 1 of my [comment](https://github.com/fermyon/spin/pull/2362#discussion_r1526538543)) . Thanks.

suneetnangia: Code updated, issue #2407 created to track and fix forward. Thanks

fibonacci1729 (approved):

karthik2804 (approved):

itowlson:
Thank you @suneetnangia and enjoy your break!

PR #2363, opened 2024-03-14
Seize control of the means of producing 404s

itowlson:
The joyless suits of Big Example have changed example.com to return 500s instead of 404s, resulting in a spurious test failure.  Fortunately a plucky startup is ready to take up the reins.


karthik2804 (approved):

vdice (approved):
snuffle

PR #2364, opened 2024-03-14
check for illegal filenames when copying singular files

### LINKED ISSUE #2361: In `component.files`, mapping a single file to `/` gives an unhelpful error
The `component.files` section allows a directory or file to be mounted in the guest at a specific location, e.g. `{ source = "cat.jpg", destination = "/dog.jpg" }`.

If the destination is an illegal name for a file, such as `/`, an error occurs, but does not clearly point to the problem:

```

Caused by:
   1: Failed to load component `favicon`
   2: Failed to copy "/Users/vdice/code/github.com/benwis/benwis_spin/target/site/favicon.ico" to "/private/var/folders/nc/yfj0pylx4vzdr3s_66r129s00000gn/T/spinup-vQC7kb/assets/favicon/"
```

It should say something like "`/` is not a valid destination file name."

karthik2804:
fixes #2361 

Provide more descriptive errors when trying to copy files with an illegal destination filename. I am not sure what the best approach for it would be as Windows has a more well-defined file naming scheme. Do we want OS-specific functions to catch all the cases?

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -374,13 +374,29 @@ impl LocalLoader {
                     quoted_path(&dest_parent)
                 )
             })?;
-        crate::fs::copy(src, dest).await.with_context(|| {
-            format!(
-                "Failed to copy {} to {}",
-                quoted_path(src),
-                quoted_path(dest)
-            )
-        })?;
+
+        let failed_to_copy_error = format!(
+            "Failed to copy {} to {}",
+            quoted_path(src),
+            quoted_path(dest)
+        );
+
+        let dest_str = dest
+            .to_str()
+            .expect("Could not convert destination path to str");
+
+        if dest_str.ends_with(std::path::MAIN_SEPARATOR)
+            || dest.to_str().unwrap_or_default().ends_with('.')
+            || dest.to_str().unwrap_or_default().ends_with("..")
+        {
+            bail!(format!(
+                "{failed_to_copy_error}: Illegal destination filename"
+            ))
+        }
+
+        crate::fs::copy(src, dest)
+            .await
+            .with_context(|| failed_to_copy_error)?;
         tracing::debug!("Copied {src:?} to {dest:?}");
         Ok(())
     }


crates/loader/src/local.rs
```diff
@@ -374,13 +374,29 @@ impl LocalLoader {
...
+        let dest_str = dest
+            .to_str()
+            .expect("Could not convert destination path to str");
+
+        if dest_str.ends_with(std::path::MAIN_SEPARATOR)
+            || dest.to_str().unwrap_or_default().ends_with('.')
+            || dest.to_str().unwrap_or_default().ends_with("..")
+        {
+            bail!(format!(
+                "{failed_to_copy_error}: Illegal destination filename"
```
itowlson, line 393: If I'm reading this right, `failed_to_copy_error` will have the _host_ path, so the user will be getting told that `/tmp/gibber/gibber/gibber/` is illegal, which they will go "true but I never asked for that path, what gives."  What we want them to understand is that the _specified destination_ in the `spin.toml` is illegal.

(I may be misunderstanding though, feel free to tick me off)

crates/loader/src/local.rs
```diff
@@ -374,13 +374,29 @@ impl LocalLoader {
...
+
+        let failed_to_copy_error = format!(
+            "Failed to copy {} to {}",
+            quoted_path(src),
+            quoted_path(dest)
+        );
+
+        let dest_str = dest
+            .to_str()
+            .expect("Could not convert destination path to str");
```
itowlson, line 386: Better is probably `if let Some(dest_str) = dest.to_str() { /* do the check */ }`, or to use `dest.display()`.  Don't tear down the entire program just because some poxy path isn't stringable.

crates/loader/src/local.rs
```diff
@@ -374,13 +374,29 @@ impl LocalLoader {
...
+            || dest.to_str().unwrap_or_default().ends_with("..")
+        {
+            bail!(format!(
+                "{failed_to_copy_error}: Illegal destination filename"
+            ))
+        }
+
+        crate::fs::copy(src, dest)
+            .await
+            .with_context(|| failed_to_copy_error)?;
```
itowlson, line 399: It would be better to attempt the copy and do the check in the failure case.  Similar to the existing code except refining how we construct the error message.

(Better in the sense of "it saves an unnecessary check and makes false positives impossible.")

itowlson (changes requested):
I have a few concerns, one of which is potentially blocking but the others are more styley.

It would also be good to see:

* An example of the new error message in the PR description.  (Right now I am trying to guess the behaviour by reading the code, so am not sure if my worries are founded.)
* Tests that run on Unix and Windows.

Thanks for picking this up, sorry to be niggly about it!

itowlson:
I believe this is superseded by #2460 (sorry Karthik).  Please reopen if you think this one is still needed!

PR #2365, opened 2024-03-15
Fix build when RUSTFLAGS is set for native builds

alexcrichton:
This fixes an issue where if `RUSTFLAGS` were set then the wasm cross-compiles performed would accidentally inherit the same flags. For me locally I was configuring a different native linker which didn't support wasm, so the wasm build failed due to the custom linker being used erroneously.


build.rs
```diff
@@ -10,6 +10,10 @@ use cargo_target_dep::build_target_dep;
 const TIMER_TRIGGER_INTEGRATION_TEST: &str = "examples/spin-timer/app-example";
 
 fn main() {
+    // Don't inherit flags from out own invocation of cargo into sub-invocations
```
lann, line 13:
```suggestion
    // Don't inherit flags from our own invocation of cargo into sub-invocations
```

lann (approved):

PR #2366, opened 2024-03-15
Refactor `TriggerExecutor` to have associated types for instances

alexcrichton:
Moves the `EitherInstance*` logic into just the `spin-trigger-http` crate where the redis trigger, for example, only has to deal with components. The intention of this is to open up future customization of instances within a trigger that don't necessarily need to affect all other triggers.

### DIFF
--- a/crates/trigger-redis/src/lib.rs
+++ b/crates/trigger-redis/src/lib.rs
@@ -7,7 +7,7 @@ use futures::{future::join_all, StreamExt};
 use redis::{Client, ConnectionLike};
 use serde::{de::IgnoredAny, Deserialize, Serialize};
 use spin_common::url::remove_credentials;
-use spin_core::async_trait;
+use spin_core::{async_trait, InstancePre};
 use spin_trigger::{cli::NoArgs, TriggerAppEngine, TriggerExecutor};
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -53,6 +53,7 @@ impl TriggerExecutor for RedisTrigger {
     type RuntimeData = RuntimeData;
     type TriggerConfig = RedisTriggerConfig;
     type RunConfig = NoArgs;
+    type InstancePre = InstancePre<RuntimeData>;
 
     async fn new(engine: TriggerAppEngine<Self>) -> Result<Self> {
         let default_address: String = engine
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -291,6 +291,7 @@ pub mod help {
         type RuntimeData = ();
         type TriggerConfig = ();
         type RunConfig = NoArgs;
+        type InstancePre = spin_core::InstancePre<Self::RuntimeData>;
         async fn new(_: crate::TriggerAppEngine<Self>) -> Result<Self> {
             Ok(Self)
         }
--- a/examples/spin-timer/src/lib.rs
+++ b/examples/spin-timer/src/lib.rs
@@ -3,8 +3,8 @@ use std::collections::HashMap;
 use clap::Args;
 use serde::{Deserialize, Serialize};
 use spin_app::MetadataKey;
-use spin_core::async_trait;
-use spin_trigger::{EitherInstance, TriggerAppEngine, TriggerExecutor};
+use spin_core::{async_trait, InstancePre};
+use spin_trigger::{TriggerAppEngine, TriggerExecutor};
 
 wasmtime::component::bindgen!({
     path: ".",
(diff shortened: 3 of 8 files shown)


lann:
@itowlson Thoughts on how this might impact https://github.com/fermyon/spin/pull/2305?

lann (commented):
lgtm but @itowlson and/or @dicej probably ought to review

dicej (approved):

itowlson:
Is the proposal that every custom trigger should have to copy the stuff that is added to the timer sample?  I'm not enthusiastic about it if so.  I'd like the direction to be helping trigger authors focus on the trigger rather than to making them wrangle wasmtime internals.

@lann I don't _think_ this will affect service chaining, at least as it stands today.  The current implementation re-enters at `HttpHandlerExecutor::execute()` which doesn't seem to be affected.  I may be missing something though - that you're expressing concern makes me think I probably am.

examples/spin-timer/src/lib.rs
```diff
@@ -113,15 +118,30 @@ impl TriggerExecutor for TimerTrigger {
         }
         Ok(())
     }
+
```
itowlson, line 121: Per my comment on the issue, let's find a way that doesn't require every custom trigger to copy this code.  (That might be as simple as defaulting things in the trait, I don't know.)

itowlson (changes requested):

lann:
> that you're expressing concern makes me think I probably am.

Nope, I just knew you've been iterating on that implementation and wanted to make sure you were in the loop here.

examples/spin-timer/src/lib.rs
lann: Sadly this can't be defaulted as-is because you can't have associated type defaults on traits (https://github.com/rust-lang/rust/issues/29661).

lann: @alexcrichton Would it be reasonable to add traits e.g. `spin_core::SpinInstance(Pre)` and have these methods return `impl SpinInstance(Pre)`?

lann: I suppose the `SpinInstance` trait might be a bit gnarly there and could end up boiling down to a component-or-module enum in disguise. Depends on what exactly you're hoping to enable here.

alexcrichton: Sounds like a reasonable concern to me! I ended up creating a `TriggerInstancePre` trait for the `spin-trigger` trait before I came here and read @lann's comments. I think that this should have the desired ergonomics but I'm happy to move/rename the trait too

crates/trigger/src/cli.rs
```diff
@@ -291,6 +291,7 @@ pub mod help {
         type RuntimeData = ();
         type TriggerConfig = ();
         type RunConfig = NoArgs;
+        type InstancePre = spin_core::InstancePre<()>;
```
lann, line 294: Seems like this is the more common pattern?

```suggestion
        type InstancePre = spin_core::InstancePre<Self::RuntimeData>;
```

crates/trigger-redis/src/lib.rs
```diff
@@ -53,6 +53,7 @@ impl TriggerExecutor for RedisTrigger {
     type RuntimeData = RuntimeData;
     type TriggerConfig = RedisTriggerConfig;
     type RunConfig = NoArgs;
+    type InstancePre = InstancePre<RuntimeData>;
```
lann, line 56: Can this go away now?

examples/spin-timer/src/lib.rs
```diff
@@ -62,6 +62,8 @@ impl TriggerExecutor for TimerTrigger {
 
     type RunConfig = CliArgs;
 
+    type InstancePre = InstancePre<RuntimeData>;
```
lann, line 65: Ditto (?)

crates/trigger-redis/src/lib.rs
lann: Ohhh sorry I misunderstood how this code was working. Yeah of course you can't.

examples/spin-timer/src/lib.rs
lann: I think given the amount of cargo-culty copying required to make a custom trigger today, this is fine. The whole interaction between spin-core, spin-app, and component instantiation is pretty wonky and became wonkier with the (partial) migration to components, and I don't think this trigger architecture can or should survive long-term.

Probably more important in the near-term would be to actually document (mea culpa) this interface a bit to help people along.

itowlson: If I read this correctly, trigger authors would no longer cargo-cult the `let EitherInstance::Component...` fragment and instead cargo-cult the `type InstancePre = ...` fragment.  That feels like less cargo-culting and I'm in favour.  Thank you Alex for simplifying this!

(If I've misread it, though, @alexcrichton, please hold off so I can get my head around the actual experience.)

itowlson (approved):

examples/spin-timer/src/lib.rs
alexcrichton: No yep you've read correctly, and once Rust supports defaults for associated types we can remove the `type InstancePre = ...` as well (or at least that's my understanding of associated typ defaults)

PR #2373, opened 2024-03-18
Lift http handler type discovery up a layer

alexcrichton:
This commit lifts the discovery of what type of handler a component is, for example spin-vs-wasi-vs-wasi-snapshot, up one layer from where it currently resides. Previously this determination was made on each request and now it's made when a component first loaded and pre-instantiated instead. This would shift possible errors earlier in the system, for example.

This removes a bit of per-request work and makes the creation of the handler in `execute_wasi` a bit more straightforward with less guessing of exports for something that was already probed.

### DIFF
--- a/crates/trigger-http/src/handler.rs
+++ b/crates/trigger-http/src/handler.rs
@@ -1,7 +1,6 @@
 use std::{net::SocketAddr, str, str::FromStr};
 
 use crate::{Body, ChainedRequestHandler, HttpExecutor, HttpInstance, HttpTrigger, Store};
-use anyhow::bail;
 use anyhow::{anyhow, Context, Result};
 use futures::TryFutureExt;
 use http::{HeaderName, HeaderValue};
@@ -11,7 +10,7 @@ use outbound_http::OutboundHttpComponent;
 use spin_core::async_trait;
 use spin_core::wasi_2023_10_18::exports::wasi::http::incoming_handler::Guest as IncomingHandler2023_10_18;
 use spin_core::wasi_2023_11_10::exports::wasi::http::incoming_handler::Guest as IncomingHandler2023_11_10;
-use spin_core::Instance;
+use spin_core::{Component, Engine, Instance};
 use spin_http::body;
 use spin_http::routes::RouteMatch;
 use spin_trigger::TriggerAppEngine;
@@ -43,26 +42,21 @@ impl HttpExecutor for HttpHandlerExecutor {
         );
 
         let (instance, mut store) = engine.prepare_instance(component_id).await?;
-        let HttpInstance::Component(instance) = instance else {
+        let HttpInstance::Component(instance, ty) = instance else {
             unreachable!()
         };
 
         set_http_origin_from_request(&mut store, engine.clone(), self, &req);
 
-        let resp = match HandlerType::from_exports(instance.exports(&mut store)) {
-            Some(HandlerType::Wasi) => {
-                Self::execute_wasi(store, instance, base, route_match, req, client_addr).await?
-            }
-            Some(HandlerType::Spin) => {
+        let resp = match ty {
+            HandlerType::Spin => {
                 Self::execute_spin(store, instance, base, route_match, req, client_addr)
                     .await
                     .map_err(contextualise_err)?
             }
-            None => bail!(
-                "Expected component to either export `{WASI_HTTP_EXPORT_2023_10_18}`, \
-                 `{WASI_HTTP_EXPORT_2023_11_10}`, `{WASI_HTTP_EXPORT_0_2_0}`, \
-                 or `fermyon:spin/inbound-http` but it exported none of those"
-            ),
+            _ => {
+                Self::execute_wasi(store, instance, ty, base, route_match, req, client_addr).await?
+            }
         };
 
         tracing::info!(
@@ -157,6 +151,7 @@ impl HttpHandlerExecutor {
     async fn execute_wasi(
         mut store: Store,
         instance: Instance,
+        ty: HandlerType,
         base: &str,
         route_match: &RouteMatch,
         mut req: Request<Body>,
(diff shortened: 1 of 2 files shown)


crates/core/src/lib.rs
```diff
@@ -399,6 +399,14 @@ impl<T: OutboundWasiHttpHandler + Send + Sync> Engine<T> {
         Ok(InstancePre { inner })
     }
 
+    /// Returns the type of this component to reflect on it at runtime.
+    #[instrument(skip_all)]
```
lann, line 403: Are you intentionally wanting this to be instrumented? Only reason I can think is if `substituted_component_type` has a chance of being very slow?

crates/trigger-http/src/handler.rs
```diff
@@ -320,28 +316,36 @@ impl HttpHandlerExecutor {
...
-            return Some(HandlerType::Spin);
+    /// Determine the handler type from the exports of a component
+    pub fn from_component(ty: &ComponentType) -> Result<HandlerType> {
+        for (name, _) in ty.exports() {
+            match name {
+                WASI_HTTP_EXPORT_2023_10_18 => return Ok(HandlerType::Wasi2023_10_18),
+                WASI_HTTP_EXPORT_2023_11_10 => return Ok(HandlerType::Wasi2023_11_10),
+                WASI_HTTP_EXPORT_0_2_0 => return Ok(HandlerType::Wasi0_2),
+                "fermyon:spin/inbound-http" => return Ok(HandlerType::Spin),
+                _ => {}
```
lann, line 340: I think this isn't equivalent to the old code if a component exports more than one of these interfaces. I'm not sure whether that is an issue in practice. @dicej @rylev thoughts?

crates/core/src/lib.rs
```diff
@@ -399,6 +399,14 @@ impl<T: OutboundWasiHttpHandler + Send + Sync> Engine<T> {
...
+    pub fn component_type(&self, component: &Component) -> Result<ComponentType> {
+        // NB: With Wasmtime 20 and bytecodealliance/wasmtime#8078 this method
+        // should probably be removed to just use that instead.
+        self.linker.substituted_component_type(component)
+    }
+
     /// Creates a new [`ModuleInstancePre`] for the given [`Module`].
     #[instrument(skip_all)]
```
lann, line 411: ...though I'm not sure why this one is instrumented either... :thinking:

lann: Ah it used to acquire a lock I was probably concerned about. Could probably be removed now. :shrug:

crates/trigger-http/src/handler.rs
dicej: Looks like the old code uses the same order, so I think it would behave the same in that (unusual) case.

crates/core/src/lib.rs
alexcrichton: Oh I don't know what this attribute does, I was just copying `instantiate_pre` above. I'll got ahead and remove it if it's related to performance measuring.

crates/trigger-http/src/handler.rs
lann: If I'm reading correctly, for an app that exports both spin and wasi interfaces the old code will always return wasi while the new code will depend on the order of `exports()` (presumably the order of items in the binary).

alexcrichton: That's right yeah, I can preserve the exact same behavior as before if desired, but I figured that level of complexity probably wasn't what was intended here

crates/core/src/lib.rs
lann: Ah yeah its a `tracing` thing that wraps calls to the function in a new span. We're going to start caring more about these soon because of https://github.com/fermyon/spin/pull/2348

crates/trigger-http/src/handler.rs
lann: As long as tooling isn't "likely" to be emitting multiple of these exported interfaces I'm fine with the code as-is. I'm mostly just unsure about spin-componentize.

rylev: I think we're unlikely to run into this situation with components built using Spin provided tooling (e.g., the SDKs). However, it does seem like we'd not want to rely on order of exports to determine which will be invoked, and indeed there should be a preferred ordering of wasi 0.2, wasi release candidates, Spin specific.

alexcrichton: I've pushed up a commit to generate an error if more than one handler is found to help resolve this

alexcrichton:
I lost this a bit in the shuffle of things, but @lann or others mind if I ask for a re-review on this?

crates/core/src/lib.rs
```diff
@@ -462,6 +462,13 @@ impl<T: OutboundWasiHttpHandler + Send + Sync> Engine<T> {
         Ok(InstancePre { inner })
     }
 
+    /// Returns the type of this component to reflect on it at runtime.
+    pub fn component_type(&self, component: &Component) -> Result<ComponentType> {
+        // NB: With Wasmtime 20 and bytecodealliance/wasmtime#8078 this method
+        // should probably be removed to just use that instead.
```
rylev, line 468: Perhaps we can land #2512 and then remove this?

crates/trigger-http/src/handler.rs
```diff
@@ -188,31 +183,33 @@ impl HttpHandlerExecutor {
...
+                            .ok_or_else(|| {
+                                anyhow!("export of `{WASI_HTTP_EXPORT_2023_11_10}` not an instance")
+                            })?;
+                        Handler::Handler2023_11_10(IncomingHandler2023_11_10::new(&mut instance)?)
+                    }
+                    HandlerType::Wasi0_2 => {
+                        drop(exports);
+                        Handler::Latest(Proxy::new(&mut store, &instance)?)
+                    }
+                    HandlerType::Spin => unreachable!(),
```
rylev, line 210: nit: I prefer using a `panic` with a message just in case (through some refactoring) this becomes technically reachable.

rylev (approved):
LGTM just some nits.

crates/trigger-http/src/handler.rs
lann: `panic!("unreachable")`

PR #2376, opened 2024-03-19
Allow Vault variables provider to retry

itowlson:
Maybe it will mitigate that flaky integration test.  Please let it mitigate that flaky integration test.

### DIFF
--- a/crates/trigger/src/runtime_config/variables_provider.rs
+++ b/crates/trigger/src/runtime_config/variables_provider.rs
@@ -70,6 +70,8 @@ pub struct VaultVariablesProviderOpts {
     pub mount: String,
     #[serde(default)]
     pub prefix: Option<String>,
+    #[serde(default)]
+    pub retry_limit: Option<u8>,
 }
 
 impl VaultVariablesProviderOpts {
@@ -79,6 +81,7 @@ impl VaultVariablesProviderOpts {
             &self.token,
             &self.mount,
             self.prefix.as_deref(),
+            self.retry_limit,
         ))
     }
 }
--- a/crates/variables/src/provider/vault.rs
+++ b/crates/variables/src/provider/vault.rs
@@ -16,6 +16,7 @@ pub struct VaultProvider {
     token: String,
     mount: String,
     prefix: Option<String>,
+    retry_limit: u8, // specifically *re*-tries; e.g. 0 = try but do not retry, 1 = try then retry at most once
 }
 
 impl VaultProvider {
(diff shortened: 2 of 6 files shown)


crates/variables/src/provider/vault.rs
```diff
@@ -24,12 +25,14 @@ impl VaultProvider {
         token: impl Into<String>,
         mount: impl Into<String>,
         prefix: Option<impl Into<String>>,
+        retry_limit: Option<u8>,
```
rylev, line 28: Is there a meaningful semantic difference between `None` and `Some(0)`? Seems like we could just pass a `u8` no?

crates/variables/src/provider/vault.rs
```diff
@@ -52,12 +55,24 @@ impl Provider for VaultProvider {
...
         };
-        match kv2::read::<Secret>(&client, &self.mount, &path).await {
-            Ok(secret) => Ok(Some(secret.value)),
-            // Vault doesn't have this entry so pass along the chain
-            Err(ClientError::APIError { code: 404, .. }) => Ok(None),
-            // Other Vault error so bail rather than looking elsewhere
-            Err(e) => Err(e).context("Failed to check Vault for config"),
+
+        let mut retry_count = 0;
+        loop {
```
rylev, line 60: Might be clear to use a `for _ in 0..=self.retry_limit {`

crates/variables/src/provider/vault.rs
```diff
@@ -52,12 +55,24 @@ impl Provider for VaultProvider {
...
+            match kv2::read::<Secret>(&client, &self.mount, &path).await {
+                Ok(secret) => return Ok(Some(secret.value)),
+                // Vault doesn't have this entry so pass along the chain
+                Err(ClientError::APIError { code: 404, .. }) => return Ok(None),
+                // Other Vault error so retry, or if at retry limit bail rather than looking elsewhere
+                Err(e) => {
+                    if retry_count >= self.retry_limit {
```
rylev, line 67: I wonder if we should only retry on 5xx errors as other errors are unlikely to be recoverable.

crates/trigger/src/runtime_config/variables_provider.rs
```diff
@@ -70,6 +70,8 @@ pub struct VaultVariablesProviderOpts {
     pub mount: String,
     #[serde(default)]
     pub prefix: Option<String>,
+    #[serde(default)]
+    pub retry_limit: Option<u8>,
```
rylev, line 74: `#[serde(default)]` should allow us to just have this be a `u8`, no?

lann:
> Please let it mitigate that flaky integration test.

It looks like testing-framework supports [Docker `HEALTHCHECK`s](https://docs.docker.com/reference/dockerfile/#healthcheck). I wonder that might solve your flakiness more directly?

lann:
This seems to work:

```
HEALTHCHECK --interval=3s CMD vault status -address=http://127.0.0.1:8200
```

lann:
https://github.com/fermyon/spin/pull/2382

itowlson:
Closing in favour of #2382

PR #2377, opened 2024-03-19
More forgiving caching

### LINKED ISSUE #2339: Cache Directory is not configurable
## Intro
When trying to run the `spin` CLI on systems with (mostly) read-only filesystem it is not possible to set the used path for the cache directory.

## Example
When running on a system where only `/tmp` is writeable, doing a `spin up --log-dir "" --runtime-config-file runtime-config.toml --state-dir /tmp`* gives the following error:
```
failed to create manifests directory "/home/<user>/.cache/spin/registry/manifests"
```
*Note: The other flags are set so that `spin` doesn't try to write to other read-only parts of the file system.

## Workaround
There is a workaround though: Setting `XDG_CACHE_HOME`, which gets respected by the used `dirs` crate, to `/tmp` works. 

## Suggestion
Looking through the code it looks like everything is already prepared to set the cache dir (e.g. `Cache::new` already expects an `Option<PathBuf>`) but it doesn't get used anywhere:

`spin up` calls `spin_loader::from_file(&manifest_path, files_mount_strategy, None)` where `None` is the Option to set the cache root.

So my suggestion would be to allow setting the cache root through a `spin` specific environment variable and/or a command line flag.

## More Context
There is a thread about it in the `spin` Discord: https://discord.com/channels/926888690310053918/1215979463918096424

itowlson:
Fixes #2339.

This provides an option to override the registry/download cache directory.  (I've called it download because it's used for more than registry stuff, although that might feel a bit weird on `spin registry push`.  Open to feedback!)

This also creates the cache directory only on demand.  If a problem occurs, the user still finds out about it during startup; but if the cache is never needed, there is no need for the user to cajole Spin into using a better directory.  Note that only the write methods trigger directory creation; "xxx_file" methods already return an error if the path/file doesn't exist, and "xxx_path" methods make it clear that the caller should not assume the path exists.  Still this is an area where reviewer persnicketiness is advised.

The two chunks are separate commits in case we want to take only one of them.

### DIFF
--- a/crates/loader/src/cache.rs
+++ b/crates/loader/src/cache.rs
@@ -2,7 +2,10 @@
 
 use anyhow::{ensure, Context, Result};
 
-use std::path::{Path, PathBuf};
+use std::{
+    path::PathBuf,
+    sync::atomic::{AtomicBool, Ordering},
+};
 
 use crate::fs::{create_dir_all, write_file};
 
@@ -17,6 +20,9 @@ const DATA_DIR: &str = "data";
 pub struct Cache {
     /// Root directory for the cache instance.
     root: PathBuf,
+    /// Whether the cache directories have been checked to exist (and
+    /// created if necessary).
+    dirs_ensured_once: AtomicBool,
 }
 
 impl Cache {
@@ -29,9 +35,11 @@ impl Cache {
                 .join(CONFIG_DIR),
         };
         let root = root.join(REGISTRY_CACHE_DIR);
-        Self::ensure_dirs(&root).await?;
 
-        Ok(Self { root })
+        Ok(Self {
+            root,
+            dirs_ensured_once: AtomicBool::new(false),
+        })
     }
 
     /// The manifests directory for the current cache.
@@ -80,12 +88,14 @@ impl Cache {
 
     /// Write the contents in the cache's wasm directory.
     pub async fn write_wasm(&self, bytes: impl AsRef<[u8]>, digest: impl AsRef<str>) -> Result<()> {
+        self.ensure_dirs().await?;
         write_file(&self.wasm_path(digest), bytes.as_ref()).await?;
         Ok(())
     }
 
     /// Write the contents in the cache's data directory.
     pub async fn write_data(&self, bytes: impl AsRef<[u8]>, digest: impl AsRef<str>) -> Result<()> {
+        self.ensure_dirs().await?;
         write_file(&self.data_path(digest), bytes.as_ref()).await?;
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


itowlson:
Okay there is a place (the `verified_download` subsystem) which uses one of the `xxx_path` APIs, then writes to that path without checking that directories exist.  This could be addressed by requiring _everything_ to go through `write_wasm` (which probably means not giving out paths to the unhallowed), or by the point fix of calling `ensure_dirs` on the way into that subsystem.  For the time being, I've adopted the point fix.

crates/loader/src/cache.rs
```diff
@@ -130,12 +141,14 @@ impl Cache {
...
 
+        *self.dirs_ensured.write().await = true;
+
         Ok(())
     }
 }
 
 #[cfg(windows)]
-fn safe_name(digest: impl AsRef<str>) -> impl AsRef<Path> {
+fn safe_name(digest: impl AsRef<str>) -> impl AsRef<std::path::Path> {
```
itowlson, line 151: This change occurs because `Path` is now unused except for here.  (I could put a `cfg` on the module level `use` of it, but this seemed less intrusive.)

crates/loader/src/cache.rs
```diff
@@ -17,6 +17,7 @@ const DATA_DIR: &str = "data";
 pub struct Cache {
     /// Root directory for the cache instance.
     root: PathBuf,
+    dirs_ensured: tokio::sync::RwLock<bool>,
```
rylev, line 20: Can we use `AtomicBool` instead?

And nit: please add a comment for what this flag is used for

src/commands/registry.rs
```diff
@@ -57,6 +57,10 @@ pub struct Push {
     /// This is a string whose format is defined by the registry standard, and generally consists of <registry>/<username>/<application-name>:<version>. E.g. ghcr.io/ogghead/spin-test-app:0.1.0
     #[clap()]
     pub reference: String,
+
+    /// Cache directory for downloaded registry data.
+    #[clap(long)]
+    pub download_cache_dir: Option<PathBuf>,
```
rylev, line 63: Can we just call it `cache_dir` or do we expect have more than one on-disk cache?

crates/loader/src/cache.rs
lann: I also like `dirs_ensured_once: tokio::sync::OnceCell<()>`:

https://docs.rs/tokio/1.36.0/tokio/sync/struct.OnceCell.html

```rust
self.dirs_ensured_once.get_or_init(|| async {
    // ensure yer dirs
}).await
```

src/commands/registry.rs
itowlson: There is already a `--cache` option for the Wasmtime cache configuration file.  I wanted to distinguish this cache from that.

fibonacci1729 (approved):

Issue #168, opened 2022-03-16
GLIBC Errors when building


bketelsen:
When building on linux, I'm getting GLIBC errors unless I'm running on a very recent Ubuntu version. (> 20.04?)

Reproduction dockerfiles for wagi (works fine) and spin(GLIBC errors) at this gist:

https://gist.github.com/bketelsen/2e2c82fb15edec676ae5194130fa7a55

Spin @ b39acc384f66a9c22aad10b2a05b6541ed71383e
Wagi @ 0.8.1

Reproduced on physical machine and docker.

radu-matei:
@gmlewis in #345:

When working through the "[Taking Spin for a spin](https://spin.fermyon.dev/quickstart)" quickstart page, I get this on a Linux Mint 19.3 box:

```bash
$ spin up
spin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by spin)
spin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by spin)

$ file `which spin`
/home/glenn/tools/bin/spin: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=1f98ec953b04be93667e5222883203df722f05af, for GNU/Linux 3.2.0, with debug_info, not stripped
```

Can't `spin` be provided as a statically-linked binary executable?

radu-matei:
Currently trying to build a statically linked executable with:

```
RUSTFLAGS="-C target-feature=+crt-static" cargo build --release`
```

itowlson:
Ooh, this might come in useful for embedding scenarios too - one of the things I ran into linking with a C `main` was missing imports.  (Although I suppose then we might get conflicts between C or other host language wanting to import `glibc` and the Rust lib already having it _sigh_.)

radu-matei:
So is the command here (https://github.com/fermyon/spin/issues/168#issuecomment-1096842282) generating a statically linked binary, @gmlewis?

(Don't have access to a Linux machine right now, so can't really test this.)

If it is, then we should update the CI bit that builds this on Linux.

gmlewis:
Oh, I didn't understand you were asking me to try building it. 😁 
I had installed from the release tarball.
I'll try and get back to you when I'm back at my computer.

gmlewis:
@radu-matei - here's what I get:

```bash
[...]

PR #2378, opened 2024-03-19
use ubuntu 20.04 for PR workflow to make it consistent with release workflow

rajatjindal:
Refer to #168 for historical context of why we are using 20.04 here

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ env:
 jobs:
   lint-rust:
     name: Lint Rust
-    runs-on: "ubuntu-latest"
+    runs-on: "ubuntu-20.04"
     permissions:
       ## Allow this job to potentially cancel the running workflow (on failure)
       actions: write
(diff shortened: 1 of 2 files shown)


rylev (approved):

lann:
I'm going to bypass the failed rust lint check on this one as it is certainly unrelated to the PR.

lann:
Ugh I saw https://github.com/fermyon/spin/actions/runs/8336227041/job/22813048050 and assumed these lint failures were for the same reason but they were not.

This certainly _shouldn't_ have caused lint errors based on the CI workflow but apparently it did. Mea culpa. :facepalm:

rylev:
Should we revert for now until we figure out what the issue is?

lann:
It looks like this was caused by caching. I deleted the GH rust cache and reran the job which appears to be succeeding now: https://github.com/fermyon/spin/actions/runs/8343891983/job/22837378173

PR #2383, opened 2024-03-19
ci: Include os-release in cache key

lann:
This will hopefully prevent the cache from sharing across incompatible e.g. glibc versions.


itowlson (approved):
nice!  Thanks for the digging!

PR #2389, opened 2024-03-22
Remove dead code newly discovered by Rust 1.77.0

rylev:
With every release of Rust, we discover more dead code.


lann (approved):
:skull:

PR #2393, opened 2024-03-24
Build inline components

itowlson:
I am not sure what to do about the duplication of manifest deserialisation stuff.  It's intentional that build shouldn't break over a bad kebab etc., but I'm not loving how much this increases build's need to make "forgiving" copies of manifest code.  I'm open to ideas on a way forward!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5951,6 +5951,7 @@ version = "2.4.0-pre0"
 dependencies = [
  "anyhow",
  "futures",
+ "indexmap 1.9.3",
  "serde",
  "spin-common",
  "spin-manifest",
--- a/crates/build/Cargo.toml
+++ b/crates/build/Cargo.toml
@@ -7,6 +7,7 @@ edition = { workspace = true }
 [dependencies]
 anyhow = "1.0.57"
 futures = "0.3.21"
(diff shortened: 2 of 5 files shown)


lann:
> It's intentional that build shouldn't break over a bad kebab etc

To clarify: you're saying that you want build to work when the manifest is otherwise invalid? Is that scenario worth accommodating?

itowlson:
There is invalid and there is invalid.  It does break flow when you can't build your code because you forgot whether variables are snake or kebab case or you messed up an `allowed_outbound_host`.  "Yes but I can fix that later, I just want to get this lifetime error sorted and then I'll come back to it."  Admittedly, it breaks flow once and then it's done...  We can certainly reassess this, and I guess should: I have a feeling that we adopted the "forgiving subset" approach at a time when the loader crate did a _lot_ of validation: I think we now expose something closer to a 'raw schema' stage, so maybe that's an acceptable compromise.

lann:
> I think we now expose something closer to a 'raw schema' stage, so maybe that's an acceptable compromise.

Yeah that was part of my intention with the Great `spin-manifest` Refactor of '23. If we feel like ID casing is something we need to be flexible with here I hope we could add that flexibility to `spin-manifest` by - say - making that validation optional somehow.

PR #2395, opened 2024-03-25
feat(oci): deduplicate layers prior to push; archive if needed

vdice:
The following refactor-slash-feature brings the following:

(refactor)
- Remove extra 'layer_count' method for determining whether to employ archive layers when publishing.  Instead, assemble layers assuming we won't need to use archive layers.  After deduplicating the layers vec, if the count exceeds the max, then we re-assemble using archive layers.  Since use of archive layers is the rarer case, it seemed fitting to only apply the tax of re-traversing the locked app here, instead of always doing so.
- Add `ClientOpts` and `content_ref_inline_max_size` as first opt so that this value can be configured.  Open to alternative ways to doing the same thing (admittedly, this need came about from unit testing).
- oci-distribution API updates per rev in https://github.com/fermyon/spin/pull/2395/commits/2985f9300591525a4a893daa6fbc6f632a9a6bf7

(feature)
- Deduplicate assembled layers. For apps with multiple components and multiple static assets (think: static website), oftentimes the same component content or file content may be reused/included across multiple components.  Previously, when publishing the app (via OCI), we would potentially be publishing the same content multiple times, which is both inefficient in bandwidth and final published artifact footprint.  With the added deduplication, we ensure content is only published once and the OCI artifact is at its minimum size.

Depends on:
- [x] https://github.com/krustlet/oci-distribution/pull/121
- [x] TODO: update Cargo.toml to use rev from fermyon's fork w/ 121 merged in.

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -7,6 +7,7 @@ use anyhow::{bail, Context, Result};
 use docker_credential::DockerCredential;
 use futures_util::future;
 use futures_util::stream::{self, StreamExt, TryStreamExt};
+use itertools::Itertools;
 use oci_distribution::{
     client::ImageLayer, config::ConfigFile, manifest::OciImageManifest, secrets::RegistryAuth,
     token_cache::RegistryTokenType, Reference, RegistryOperation,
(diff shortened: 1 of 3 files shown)


crates/oci/src/client.rs
```diff
@@ -110,68 +128,29 @@ impl Client {
...
-                            "cannot push file layers for source {}",
-                            quoted_path(&source)
-                        ))?;
-                }
-            }
-            c.files = files;
-            components.push(c);
+        let mut locked_app = locked.clone();
+        let mut layers = self
+            .assemble_layers(&mut locked_app, false)
```
itowlson, line 137: Perpetual Pet Peeve: a boolean here is not informative to the reader.  Consider a two-value enum e.g. `AssemblyMode::Simple` / `AssemblyMode::Consolidate` or `Archive::Yes` / `Archive::No` or whatever you feel would be most helpful to someone reading the code at the call site.

itowlson (approved):
LGTM with one non-blocking suggestion (once the upstream OCI work has merged of course)

itowlson:
@vdice I added a "do-not-merge" label to flag the dependency on upstream OCI

crates/oci/src/client.rs
vdice: Thanks for mentioning and advising; updated in https://github.com/fermyon/spin/pull/2395/commits/3871196285580fb1bbee0b14584738c644268036

PR #2396, opened 2024-03-25
Allow modular manifest files

### LINKED ISSUE #1759: Modulizing Spin.toml files
My spin.toml is getting bigger and bigger (A couple of hundred lines). 
If I can implement some sort of 'include' feature that allows me to modularize spin.toml's content into separate TOML files, 
it would be more organized and productive.

itowlson:
Fixes #1759.

This is WIP for discussion and pondering.

The proposed UX for the modular feature is to overload the `trigger.component` field to allow it to reference a file (or directory containing a well-known file).  I considered an embedded table for this (`[[trigger.http]] component = { path = "example.toml" }`) but it felt wordy, so the current draft instead proposes an `@filename` syntax similar to what we use for the `--sqlite` flag (`[[trigger.http]] component = "@example.toml"`).  I kind of like it, but suspect it will be divisive!

There are also a couple of serious structural considerations:

1. The duplication of manifest resolution into the `build` loader, mentioned in #2393, increases yet more, which is extremely not good.  I don't have a good solution for this - I'm toying with some ideas that could defer the full parse in cases where it's not needed, but nothing looks super easy.

2. This implementation fails on the requirement of tracing errors back to the source file path.  Core manifest validation will get the file path, but if, for example, the Wasm file mentioned in the child file doesn't exist, it gives you no clue as to which file the error lies in.  We do carry some application origin info into the trigger: I'm wondering if we could attach an optional origin field to each component, but we'd not want that going into the lockfile because the lockfile also serves as the OCI thing.

Also, no, there are (almost) no tests yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5951,6 +5951,7 @@ version = "2.4.0-pre0"
 dependencies = [
  "anyhow",
  "futures",
+ "indexmap 1.9.3",
  "serde",
  "spin-common",
  "spin-manifest",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6327,6 +6327,7 @@ dependencies = [
  "indexmap 1.9.3",
  "serde",
  "serde_json",
+ "spin-common",
  "spin-serde",
  "terminal",
  "thiserror",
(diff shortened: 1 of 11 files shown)


lann:
This seems sufficiently design-laden to be worth a SIP.

PR #2397, opened 2024-03-26
Provide a way for lockfile consumers to see the actual downlevel version

itowlson:
New Spin always reports lockfiles as v1, but for transitional purposes can serialise as v0 when the app is downlevel compatible.  However, there is no current way for tools that are passing an app to a consumer to determine if the app is downlevel compatible (other than by using a downlevel version of the Spin crate to try to load it).  An example is the cloud plugin needing to validate an application because Cloud does not yet accept v1 lockfiles.  This PR provides a workaround.

### DIFF
--- a/crates/locked-app/src/locked.rs
+++ b/crates/locked-app/src/locked.rs
@@ -314,6 +314,7 @@ mod test {
 
         let reloaded = LockedApp::from_json(&json).unwrap();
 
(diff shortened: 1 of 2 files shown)


itowlson:
Ugh, this doesn't help because cloud plugin doesn't serialise the lockfile

PR #2403, opened 2024-03-28
Add recent SIPs to the index document

tschneidereit:


lann:
I wonder if we should maybe remove the index instead? I believe we _used_ to render these to the Spin docs site but I don't see that any more.

lann (approved):

tschneidereit:
Yeah, I think it wouldn't be unreasonable to do something else here, but for now we link to it from the devhub, so it should at least be up to date :)

itowlson (approved):

itowlson:
@vdice @lann or other admin, can you force merge please?  Docs never become "naturally" mergeable we skip the build workflow so required checks never pass.

PR #2404, opened 2024-03-28
ci: Run macos builds on M1

lann:


lann:
Bypassing merge requirements because they need to be updated with this change.

PR #2409, opened 2024-04-01
chore(trigger-http): bump tls-listener

vdice:
- Bumps the tls-listener crate up to 0.10.0 per https://github.com/advisories/GHSA-2qph-qpvm-2qf7

It appears we only utilize the rust_tls feature of this crate and we don't invoke the `TlsListener::new()` method susceptible to the advisory, so open to discussion on if we want to take this opportunity to bump anyways -- or close this and mark the security warning as not applicable/won't fix.

Assuming we go ahead with the bump -- any special/additional testing we want to perform?


rylev (approved):
This seems fine for us to do IMO

fibonacci1729 (approved):

PR #2410, opened 2024-04-01
Warn when sending bare 404 response

### LINKED ISSUE #1132: Nicer errors when no component route matches
Currently we just return a generic 404 if no route matches. It would be nice to give more specific feedback in the 404 response and/or console.

itowlson:
Fixes #1132

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -6,6 +6,7 @@ mod wagi;
 
 use std::{
     collections::HashMap,
+    io::IsTerminal,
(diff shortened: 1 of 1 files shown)


vdice (approved):
Looks good to me.  I see the logic is set up to warn only once on the first 404.  I had a thought that we might perhaps want to log on every 404 _if_ running in a verbose mode (perhaps log level 'trace' or some such?) but not sure if it is worth the complication and in any case, perhaps premature; seems a fine idea to get this in as-is and revisit if needed.

Issue #2146, opened 2023-12-06
Provide a request ID as a header when the http trigger is triggered


mikkelhegn:
For logging and debugging purposes it's nice to be able to pre-fix logs with a request Id. Instead of having every component implement generating a request id, it would be nice if Spin could provide one as a header in the http request send to the component.

I don't know if other trigger types could make use of the same information, or if it already exists in the runtime!?!

karthik2804:
I would like to clarify my understanding here. Is the request here to generate something like a UUID on each request and pass it through as a `spin-request-id` header??

itowlson:
I don't think it currently exists.  If we did this then yes, it would be good to make it available on all triggers, but only HTTP has the header concept: for other triggers, it would involve a change to the event payload, which is a WIT change.

A possible alternative way to do this would be to provide a new `event_id` host API, which any guest could call regardless of trigger type.  This would be additive so no change to existing signatures.  But if it were HTTP only then the header would work fine.

mikkelhegn:
I like the `event_id` idea. If the HTTP-trigger (or any trigger) could then implement it as part of a header or payload, would be up to the trigger.

rylev:
One thing to consider: we should probably provide a logging interface that it's not up to each component to log things in a uniform way. This would allow the Spin runtime and thus the end user to have more control over how logs are displayed.

mikkelhegn:
Could this be an option @rylev? https://github.com/dylibso/wasi-observe

vdice:
@calebschoepp is there any overlap with this request and current/forecasted telemetry work?

calebschoepp:
My initial impression is that there isn't a ton of overlap here with WASI Observe. In the short to medium term WASI Observe is much more focused on tracing and metrics. I'm sure long-term WASI Observe will support the OTEL logging signal but it doesn't seem like a priority right now. @endocrimes please chime in if your perception is different.

As for overlap on the Spin specific observability work I've been doing: I also think there isn't really much overlap there.

As for participating in the discussion +1 to what Ryan said for making it not the responsibility of a component to log properly. I wonder if WASI Logging has anything that could help here.

calebschoepp:
@mikkelhegn do you still feel like you have a need for this now that we have tracing?

PR #2411, opened 2024-04-02
Request ID

itowlson:
This is a possible approach to #2146 but I feel kinda queasy using an environment variable instead of an API but maybe that's just me

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6585,6 +6585,7 @@ dependencies = [
  "toml 0.5.11",
  "tracing",
  "url",
+ "uuid",
  "wasmtime",
  "wasmtime-wasi",
  "wasmtime-wasi-http",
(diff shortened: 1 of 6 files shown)


karthik2804:
Would this be needed once we have wasi-observe and otel?

itowlson:
@karthik2804 on that I was guided by https://github.com/fermyon/spin/issues/2146#issuecomment-2000217985 but for sure I do not know enough either way

radu-matei:
Could this (also?) be a Spin variable?

lann:
I think this may in fact be best served via wasi-observe. Distributed traces are typically keyed off of a trace ID which is either passed in via e.g. [W3C Trace Context](https://www.w3.org/TR/trace-context/) or generated by the server (if it is the first hop in the traced system). That trace ID could be emitted with logs and - if necessary - exposed to the guest via a wasi-observe import.

Issue #2368, opened 2024-03-17
consider removing prompt when installing plugin


michelleN:
`krew`, `kubectl`s plugin manager displays the following message instead of prompting to confirm installation of the plugin:
```
WARNING: You installed plugin "reverse-proxy" from the krew-index plugin repository.
   These plugins are not audited for security by the Krew maintainers.
   Run them at your own risk.
```

It would be nice to let people know they installed something we have not audited and skip the prompt step. I don't there is ever going to be a time when I say no and I always forget to use the `--yes` flag.

I'd love to see either of the following:
- Map environment variable to value of the `--yes` flag so that you could set false if you are more cautious and want to be prompted.
- Make default behavior more like `krew`'s warning rather than prompting to confirm installation

karthik2804:
None of our plugins are technically "audited" even the ones that come from the registry. The original design of the plugin system follows the `apt` style package managers. I would still be in favor of styling and having the user acknowledge that they are installing an explicit action.

That being said I like the idea of having an environment variable similar to how we have `SPIN_ALWAYS_BUILD` that a user can set so that they can avoid the `--yes` flag.

lann:
I'm assuming krew doesn't _automatically_ install plugins and then run them with just a warning? I think that's the main difference here.

lann:
Oh I guess if you're talking about `--yes` you mean the actual `spin plugins install` command. Yeah I think it would be fine to just display a warning there.

Separately, we might want to be a wee bit more explicit about the information displayed when we suggest installing if you run `spin kube` without the plugin already installed. Its weird that the message on explicit `spin plugins install kube`:

```
You are trying to install the `kube` plugin with Apache-2.0 license from https://github.com/spinkube/spin-plugin-kube/releases/download/v0.1.0/spin-plugin-kube-0.1.0-linux-amd64.tar.gz 
Are you sure you want to continue?
```

has more information than the prompt on `spin kube`:

```
Warning: `kube` is not a known Spin command, but there is a plugin with that name.
Would you like to install and run it now?
```

Also weird that the former defaults to "no" and the latter has no default.

I'd suggest:
- Change the `spin plugins install` prompt default to "yes" (if not removing it entirely). I feel like the current behavior _might_ have been my suggestion but *people can change their minds OK?* :sweat_smile: 
- Add more info to the "suggest installing this plugin" flow (license and URL) and have that default to "no"; accidentally installing from a suggestion like that would be no good.

iamrajiv:
[...]

PR #2412, opened 2024-04-02
Rationalise plugin install prompts

itowlson:
apropos of #2368

### DIFF
--- a/src/commands/external.rs
+++ b/src/commands/external.rs
@@ -158,28 +158,45 @@ async fn consider_install(
             }
             plugin_installer.run().await?;
         }
-        Ok(None) // No update badgering needed if we just updated/installed it!
-    } else if stderr().is_terminal() && matches_catalogue_plugin(plugin_store, plugin_name) {
-        if offer_install(plugin_name)? {
-            let plugin_installer = installer_for(plugin_name);
-            plugin_installer.run().await?;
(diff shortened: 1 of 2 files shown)


src/commands/external.rs
```diff
@@ -158,28 +158,46 @@ async fn consider_install(
...
+fn offer_install(
+    plugin: &spin_plugins::manifest::PluginManifest,
+    package: &spin_plugins::manifest::PluginPackage,
+) -> anyhow::Result<bool, anyhow::Error> {
+    terminal::warn!(
+        "`{}` is not a known Spin command, but there is a plugin with that name.",
+        plugin.name()
+    );
+    let prompt = format!(
+        "Would you like to install the '{}' with {} license from {}, and run it now?",
```
karthik2804, line 193: When using long lines on prompts, if there is a line-wrap, the confirmation will cause it to print twice. This was an issue we had with plugins earlier that we solved by splitting it into two lines with the line on the prompt being shorter.

lann, line 193: Should this read 
> Would you like to install the 'kube' **plugin** ...

?

itowlson: 🤦

lann (approved):

fibonacci1729 (approved):

PR #2418, opened 2024-04-03
Private endpoints

itowlson:
The proposed syntax is that a private endpoint still requires a `[[trigger.http]]`, but has `route = false` instead of an actual route pattern.  (This creates the slight idiosyncrasy that `route = false` is allowed but `route = true` is not.  I can live with this!)

In addition to unit tests, the front-to-back service chaining test is modified to make the `middle` component a private endpoint.  Let me know if you think further tests are needed or if they should be reorganised to be more specific.

### DIFF
--- a/crates/http/src/config.rs
+++ b/crates/http/src/config.rs
@@ -7,10 +7,30 @@ pub struct HttpTriggerConfig {
     /// Component ID to invoke
     pub component: String,
     /// HTTP route the component will be invoked for
-    pub route: String,
+    pub route: HttpTriggerRouteConfig,
     /// The HTTP executor the component requires
     #[serde(default)]
     pub executor: Option<HttpExecutorType>,
+}
+
+/// An HTTP trigger route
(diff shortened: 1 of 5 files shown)


fibonacci1729 (approved):
This looks great to me. I can live with the `route = true` weirdness.

Another possible way to express this which we can absolutely ignore that i mentioned if the hate is too strong:

```toml
[[trigger.http]]
 # public defaults to `true` to retain `route = "/..."` syntax
route = { path = "/...", public = false }
```

This has the added benefit of allowing private endpoints to express a route prefix. Admittedly i can't come up with a use-case for this. 

_Just tossing out low-blood sugar ideas._

lann (approved):

ThorstenHans:
This is great! 

Although it is already merged, I would also prefer using a property that reflects visibility as suggested by @fibonacci1729 

Some additional ideas: 

- `private = true`
- `internal = true`
- `public = false`

PR #2419, opened 2024-04-03
(Discussion): allow multi trigger to continue if a trigger exits successfully

karthik2804:
To enable running other triggers along with the `command` trigger as per https://github.com/fermyon/spin-trigger-command/issues/7

This PR just focuses on enabling the ability for multi-triggers to continue if one of them exits successfully.

### DIFF
--- a/src/commands/up.rs
+++ b/src/commands/up.rs
@@ -210,7 +210,7 @@ impl UpCommand {
 
         set_kill_on_ctrl_c(&pids)?;
 
-        let trigger_tasks = trigger_processes
+        let mut trigger_tasks = trigger_processes
             .into_iter()
             .map(|mut ch| tokio::task::spawn(async move { ch.wait().await }))
             .collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


itowlson:
I'm not sure this is the right default.  For most triggers, if the trigger is no longer accepting events, the application is not doing what you want it to do, and should exit.

We have various options here.  I thought of:

1. Do this and hope that no eventful trigger ever exits with code 0.
2. Have a manifest setting where the developer can mark a trigger as "continue after exit".
3. Add a field to the trigger launch metadata where a trigger can report itself as "expect me to terminate and continue without me".

I am wary of 1.  3 is convenient for application developers which is nice, but I don't know if it provides enough flexibility for the command use case (what if someone writes a side-car-like command that is expected to stay up for the lifetime of the application?  We can't assume init containers are the only use case for this).  2 feels like a tedious task that the app developer has to remember to do every time.  Any other ideas?

itowlson:
Looking at the original issue, it also seems to express a requirement that (in the author's particular case), the other triggers do not start _until the command trigger has successfully exited._  If that's the case, we'll need significantly greater structural changes...

karthik2804:
Those are some valid concerns, if we do not want to rely on triggers being well-behaved. I like the explicitness of option 2, it also potentially sets a way for the manifest to also define if certain triggers need to be run serially or parallelly.

lann:
https://github.com/fermyon/spin-trigger-command/issues/7#issuecomment-2035572411

karthik2804:
Closing this as it has been stale for a while and it can be reopened if it is need again in the future.

PR #2425, opened 2024-04-04
Fire and forget HTTP requests

### LINKED ISSUE #1261: "Fire and forget" option for outbound HTTP
I would like to be able to kick off a request but not wait for its response using the outbound HTTP API.

itowlson:
Fixes #1261.

This is not, repeat not, a real crack at a developer experience for this one, more a technology proof of concept.  I am not sure what the right DX is:

* How should the developer signal that they want fire-and-forget semantics?
  * For this draft, I've used a request header (`spin-detach`) that you set to say "run this in the background", but that seems terrible.
  * We could use another chaining domain such as `*.spin-background.internal` or something, preferably something.
  * Something else
  * A likely requirement for this is deterministic feature detection from the manifest.  You would not want to deploy your app onto a platform that supports "request-response" service chaining but not fire-and-forget, which the hacky header allows.
* What response should `http::send` return to the sender?  It has to return something.
  * I am currently returning a 202 with a meaningless Location and an empty body

Yes there are no tests either but it WORKED ON MY MACHINE.  Testing seems like an intriguing problem to be worked out when we agree on how to opt into the behaviour.  Also the code is getting bloaty and duplicatey and needs splitting up, I know.

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -543,6 +543,7 @@ impl HttpRuntimeData {
         use wasmtime_wasi_http::types::IncomingResponseInternal;
 
         let this = data.as_ref();
+        let run_deteched = request.request.headers().get("spin-detach").is_some();
 
         let chained_handler = this.chained_handler.clone().ok_or(wasmtime::Error::msg(
             "Internal error: internal request chaining not prepared (engine not assigned)",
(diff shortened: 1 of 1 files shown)


itowlson:
To be clear: this PR provides fire-and-forget _only_ for internal service-chained components.  #1261 does ask for a broader feature, although in principle that could be done with this by proxying through a service-chained component.  Or we could scrap this and have another crack at the broader ask.  _grin_

lann:
I think _part_ of the functionality behind https://github.com/fermyon/spin/issues/1261 is now possible with wasi-http; I believe you can return a response without stopping the guest execution. Stopping guest execution while an outbound http request is in-flight does have its uses but seems somewhat less important.

itowlson:
@lann That hasn't worked for me: you can stream a response body, yes, but the client will not see the request as complete until the Spin handler exits.  Maybe I was missing a flush or close or something although I thought I tried those.  Or maybe if I had set content-length it would have worked?  (My test case was, yes, trying to return a response to Slack and continue processing: Slack timed it out after 3 seconds.)

The other thing we've seen in the past is that once the client has received the response and disconnected, Spin (well hyper really) terminates the guest anyway.  I haven't tested that recently though.

lann:
> the client will not see the request as complete until the Spin handler exits. Maybe I was missing a flush or close or something although I thought I tried those.

You are supposed to call `finish`, at least at the component level:

https://docs.rs/wasi/0.13.0+wasi-0.2.0/wasi/http/types/struct.OutgoingBody.html#method.finish

If that isn't working its probably (?) a Spin bug.

> The other thing we've seen in the past is that once the client has received the response and disconnected, Spin (well hyper really) terminates the guest anyway. I haven't tested that recently though.

I _vaguely_ recall something like that being fixed but I also haven't tested.

itowlson:
Thanks.  I'll give that a go.

itowlson:
Yep.  We (sensibly) encapsulate `finish` in the `drop` of the unnameable outgoing body wrapper.  So doing a "premature" drop on the outgoing body sink object closes the response but allows the guest to continue running.  Thanks for the pointer!

PR #2426, opened 2024-04-04
Option to suppress plugin on-run compatibility warnings

itowlson:
Put it in your profile and never think about it again.

Until someone reports a weird bug with malformed JSON that you just can't reproduce on your local machine.  Then think about it a lot.  (Yes, this would have bitten me nastily with the launch metadata work for multi-trigger, and that's why I remain in two minds about it.)

### DIFF
--- a/crates/plugins/src/manifest.rs
+++ b/crates/plugins/src/manifest.rs
@@ -181,13 +181,16 @@ fn inner_warn_unsupported_version(
     override_compatibility_check: bool,
 ) -> Result<()> {
     if !is_version_fully_compatible(supported_on, spin_version)? {
+        let show_warnings = !suppress_compatibility_warnings();
         let version = Version::parse(spin_version)?;
         if !version.pre.is_empty() {
-            if std::io::stderr().is_terminal() {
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

vdice (commented):
Any thoughts on discoverability of this env var for users?

Too heavyweight to pair with a new flag on the `spin plugins` command suite eg `--suppress-compatibility-warnings`?  If that was in place, clap would also show the env var value, right?

Or maybe just a docs mention?

Or just _keep it as-is on a need-to-know basis only_? 🤫

itowlson:
I considered mirroring it to a flag but who is ever going to type out the flag?  This is really for "I live on Spin prereleases/canaries and I know it" people - it's completely irrelevant for stable.  But yeah maybe discoverability and/or programmatic use justify it.  Still would feel weird to see it on stable.  I dunno.

vdice:
👍 Let's get this in as-is and we can always increase awareness if requested.

vdice (approved):

kate-goldenring (approved):

PR #2427, opened 2024-04-04
Accept runtime config in either JSON or TOML

lann:
Many environments have better JSON tooling than TOML. To ease integration, accept either format for runtime config.

This chooses the format as follows:
- If the extension is `.toml`, parse as TOML
- If the extension is `.json`, parse as JSON
- If the first non-whitespace character is `{`, parse as JSON (never valid TOML)
- Otherwise, parse as TOML (compatible with current behavior)


itowlson (approved):

endocrimes:
GitHub mobile is too easy to fat-finger, sorry!

(Yay I am excited to purge a bunch of SpinKube code at some point)

fibonacci1729 (approved):

PR #2433, opened 2024-04-05
ref(*): Refactor host components to avoid returning Result<Result<T>> if they don't trap

calebschoepp:
This cleans up the code so that we can avoid extraneous `Ok(async {}.await)` shenanigans. It also has the added benefit of making `#[instrument(err)]` work out of the box because there is no nested error. If we didn't do this fix we would need to wrap every host component method returning `Result<Result<T>>` in a macro:

```rust
#[macro_export]
macro_rules! instrument_inner_err {
    ($($e:tt)*) => {{
        match async {$($e)*}.await {
            Ok(inner) => {
                match inner {
                    Ok(innest) => Ok(Ok(innest)),
                    Err(e) => {
                        tracing::event!(target:module_path!(), Level::INFO, error = %e);
                        Ok(Err(e))
                    }
                }
            }
            Err(e) => Err(e),
        }
    }};
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -127,10 +127,10 @@ hyper = { version = "1.0.0", features = ["full"] }
 reqwest = { version = "0.11", features = ["stream", "blocking"] }
 tracing = { version = "0.1", features = ["log"] }
 
-wasi-common-preview1 = { version = "18.0.1", package = "wasi-common" }
-wasmtime = "18.0.1"
-wasmtime-wasi = { version = "18.0.1", features = ["tokio"] }
-wasmtime-wasi-http = "18.0.1"
+wasi-common-preview1 = { version = "18.0.4", package = "wasi-common" }
+wasmtime = "18.0.4"
+wasmtime-wasi = { version = "18.0.4", features = ["tokio"] }
+wasmtime-wasi-http = "18.0.4"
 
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -141,3 +141,6 @@ blocks_in_conditions = "allow"
 [[bin]]
 name = "spin"
 path = "src/bin/spin.rs"
+
+[patch.crates-io]
+wasmtime-wit-bindgen = { git = "https://github.com/calebschoepp/wasmtime", branch = "hotfix" }
```
calebschoepp, line 146: Waiting for this to be backported to Spin. The code can be reviewed in the meantime.

crates/world/src/lib.rs
```diff
@@ -9,7 +9,22 @@ wasmtime::component::bindgen!({
     }
     "#,
     path: "../../wit",
-    async: true
+    async: true,
+    trappable_error_type: {
```
alexcrichton, line 13: It might be worth adding a comment above this indicating that this is a bit of a roundabout method of saying "the host implementations for these interfaces don't trap" but that's the intent of this configuration

lann (approved):

lann:
Now that its a approved I'm going to mark it a draft so it doesn't accidentally merge...

PR #2434, opened 2024-04-05
ci(release): fix dispatch conditional

vdice:
Apparently `${{ startsWith(github.ref, 'refs/tags/v') }}` always resolves to true, presumably because when wrapped in expression brackets the result is a string rather than a boolean.  Removing the expression brackets to fix... probably.


lann (approved):
woops!

PR #2435, opened 2024-04-05
ci(build/release): restore mac amd64 builds

vdice:


dicej (approved):

vdice:
Wait a sec, does `macos-latest` already resolve to `macos-14`?  Looking at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-private-repositories

Do we need to use `macos-13` now for amd64 builds?

lann:
Ha looks like they're right in the middle of migrating: https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/

lann:
We should probably stick with explicit versions for a couple of weeks at least until things calm down.

vdice:
Ok, pinned mac amd64 builds to `macos-13`.  Didn't add this config back to where we run integration tests, assuming using `macos-14` is sufficient for our needs.

PR #2437, opened 2024-04-05
Trace some more host components

calebschoepp:


rylev (approved):

crates/outbound-http/src/host_impl.rs
```diff
@@ -38,8 +39,28 @@ impl OutboundHttp {
...
+            current_span.record("http.request.method", method);
+            if let Ok(uri) = req.uri.parse::<Uri>() {
+                if let Some(authority) = uri.authority() {
+                    current_span.record("server.address", authority.host());
+                    if let Some(port) = authority.port() {
+                        current_span.record("server.port", port.as_u16());
+                    }
+                }
+            }
+
```
lann, line 63: Per other comments it would be nice to extract this at some point in the not-too-distant future.

lann (approved):

crates/outbound-http/src/host_impl.rs
lann: In this particular case I wouldn't use the attribute macro, I'd manage the span (more) manually

PR #2441, opened 2024-04-08
Trigger tracing

calebschoepp:
Note that there is no existing standard for trace context propagation in Redis. https://www.w3.org/TR/trace-context-protocols-registry/#registry

### DIFF
--- a/crates/trigger-http/src/handler.rs
+++ b/crates/trigger-http/src/handler.rs
@@ -17,7 +17,7 @@ use spin_trigger::TriggerAppEngine;
 use spin_world::v1::http_types;
 use std::sync::Arc;
 use tokio::{sync::oneshot, task};
-use tracing::{instrument, Instrument};
+use tracing::{instrument, Instrument, Level};
 use wasmtime_wasi_http::{proxy::Proxy, WasiHttpView};
 
 #[derive(Clone)]
@@ -25,7 +25,7 @@ pub struct HttpHandlerExecutor;
 
 #[async_trait]
 impl HttpExecutor for HttpHandlerExecutor {
-    #[instrument(name = "execute_wasm", skip_all, fields(otel.kind = "server"))]
+    #[instrument(name = "spin_trigger_http.execute_wasm", skip_all, err(level = Level::INFO), fields(otel.name = format!("execute_wasm_component {}", component_id)))]
     async fn execute(
         &self,
         engine: Arc<TriggerAppEngine<HttpTrigger>>,
--- a/crates/trigger-http/src/instrument.rs
+++ b/crates/trigger-http/src/instrument.rs
@@ -1,0 +1,93 @@
+use anyhow::Result;
+use http::Response;
+use tracing::Level;
+use wasmtime_wasi_http::body::HyperIncomingBody;
+
+/// Create a span for an HTTP request.
+macro_rules! http_span {
+    ($request:tt, $addr:tt) => {
+        tracing::info_span!(
+            "spin_trigger_http.handle_http_request",
+            "otel.kind" = "server",
+            "http.request.method" = %$request.method(),
+            "network.peer.address" = %$addr.ip(),
+            "network.peer.port" = %$addr.port(),
+            "network.protocol.name" = "http",
+            "url.path" = $request.uri().path(),
+            "url.query" = $request.uri().query().unwrap_or(""),
+            "url.scheme" = $request.uri().scheme_str().unwrap_or(""),
+            "client.address" = $request.headers().get("x-forwarded-for").and_then(|val| val.to_str().ok()),
+            // Recorded later
+            "error.type" = Empty,
+            "http.response.status_code" = Empty,
+            "http.route" = Empty,
+            "otel.name" = Empty,
+        )
+    };
+}
+
+pub(crate) use http_span;
+
+/// Finish setting attributes on the HTTP span.
+pub(crate) fn finalize_http_span(
+    response: Result<Response<HyperIncomingBody>>,
+    method: String,
+) -> Result<Response<HyperIncomingBody>> {
+    let span = tracing::Span::current();
+    match response {
+        Ok(response) => {
+            let matched_route = response.extensions().get::<MatchedRoute>();
+            // Set otel.name and http.route
+            if let Some(MatchedRoute { route }) = matched_route {
+                span.record("http.route", route);
+                span.record("otel.name", format!("{method} {route}"));
+            } else {
+                span.record("otel.name", method);
+            }
+
+            // Set status code
+            span.record("http.response.status_code", response.status().as_u16());
+
+            Ok(response)
+        }
+        Err(err) => {
+            instrument_error(&err);
+            span.record("http.response.status_code", 500);
+            span.record("otel.name", method);
+            Err(err)
+        }
+    }
+}
+
+/// Marks the current span as errored.
+pub(crate) fn instrument_error(err: &anyhow::Error) {
+    let span = tracing::Span::current();
+    tracing::event!(target:module_path!(), Level::INFO, error = %err);
+    span.record("error.type", format!("{:?}", err));
+}
+
+/// MatchedRoute is used as a response extension to track the route that was matched for OTel
+/// tracing purposes.
+#[derive(Clone)]
+pub struct MatchedRoute {
+    pub route: String,
+}
+
+impl MatchedRoute {
+    pub fn set_response_extension(
+        resp: &mut Response<HyperIncomingBody>,
+        route: impl Into<String>,
+    ) {
+        resp.extensions_mut().insert(MatchedRoute {
+            route: route.into(),
+        });
+    }
+
+    pub fn with_response_extension(
+        mut resp: Response<HyperIncomingBody>,
+        route: impl Into<String>,
+    ) -> Response<HyperIncomingBody> {
+        Self::set_response_extension(&mut resp, route);
+        resp
+    }
+}
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -1,6 +1,7 @@
 //! Implementation for the Spin HTTP engine.
 
 mod handler;
+mod instrument;
 mod tls;
 mod wagi;
 
@@ -25,6 +26,7 @@ use hyper::{
     Request, Response,
 };
 use hyper_util::rt::tokio::TokioIo;
+use instrument::{finalize_http_span, http_span};
 use spin_app::{AppComponent, APP_DESCRIPTION_KEY};
 use spin_core::{Engine, OutboundWasiHttpHandler};
 use spin_http::{
@@ -42,10 +44,17 @@ use tokio::{
     net::TcpListener,
     task,
 };
-use tracing::{log, Instrument};
-use wasmtime_wasi_http::{body::HyperIncomingBody as Body, WasiHttpView};
+use tracing::{field::Empty, log, Instrument};
+use wasmtime_wasi_http::{
+    body::{HyperIncomingBody as Body, HyperOutgoingBody},
+    WasiHttpView,
+};
 
-use crate::{handler::HttpHandlerExecutor, wagi::WagiHttpExecutor};
+use crate::{
+    handler::HttpHandlerExecutor,
+    instrument::{instrument_error, MatchedRoute},
+    wagi::WagiHttpExecutor,
+};
 
 pub use tls::TlsConfig;
 
@@ -240,19 +249,22 @@ impl HttpTrigger {
             req.uri()
         );
 
-        let path = req.uri().path();
+        let path = req.uri().path().to_string();
 
         // Handle well-known spin paths
         if let Some(well_known) = path.strip_prefix(spin_http::WELL_KNOWN_PREFIX) {
             return match well_known {
-                "health" => Ok(Response::new(body::full(Bytes::from_static(b"OK")))),
-                "info" => self.app_info(),
+                "health" => Ok(MatchedRoute::with_response_extension(
+                    Response::new(body::full(Bytes::from_static(b"OK"))),
+                    path,
+                )),
+                "info" => self.app_info(path),
                 _ => Self::not_found(NotFoundRouteKind::WellKnown),
             };
         }
 
         // Route to app component
-        match self.router.route(path) {
+        match self.router.route(&path) {
             Ok(component_id) => {
                 let trigger = self.component_trigger_configs.get(component_id).unwrap();
 
@@ -293,10 +305,14 @@ impl HttpTrigger {
                     }
                 };
                 match res {
-                    Ok(res) => Ok(res),
+                    Ok(res) => Ok(MatchedRoute::with_response_extension(
+                        res,
+                        raw_route.to_string(),
+                    )),
                     Err(e) => {
                         log::error!("Error processing request: {:?}", e);
-                        Self::internal_error(None)
+                        instrument_error(&e);
+                        Self::internal_error(None, raw_route.to_string())
                     }
                 }
             }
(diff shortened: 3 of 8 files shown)


crates/trigger-http/src/lib.rs
```diff
@@ -391,6 +386,40 @@ impl HttpTrigger {
...
+        self.handle(
+            request.map(|body: Incoming| {
+                body.map_err(wasmtime_wasi_http::hyper_response_error)
+                    .boxed()
+            }),
+            Scheme::HTTP,
+            addr,
+        )
+        .instrument(span)
+        .await
```
calebschoepp, line 420: This used to be wrapped in an `async move {}` which I removed. Not sure why it was there.

crates/trigger-http/src/lib.rs
```diff
@@ -244,11 +248,20 @@ impl HttpTrigger {
...
+                _ => None,
             };
+            if let Some(res) = well_known_matched {
+                current_span.record("otel.name", format!("{} {}", req.method(), path));
+                current_span.record("http.route", path);
+                current_span.record("http.response.status_code", 200);
+                return res;
+            }
+            current_span.record("otel.name", format!("{}", req.method()));
+            current_span.record("http.response.status_code", 404);
```
itowlson, line 263: I can no longer follow the flow of this stanza at all.  It's gone from a simple pattern match to big pile of tracing statements whose shape gives me no sense of the underlying logic.  (I know this is not on you and that you are trying to get it done with the tools OpenTelemetry gives you and are probably as queasy as I am!)

Is it possible that moving the tracing up a level could capture this stuff in a unified way?  The various flows here might make it difficult I realise.

crates/trigger-http/src/lib.rs
```diff
@@ -391,6 +386,40 @@ impl HttpTrigger {
         });
     }
 
+    async fn instrumented_service_fn(
+        self: Arc<Self>,
+        addr: SocketAddr,
+        request: Request<Incoming>,
+    ) -> Result<Response<HyperOutgoingBody>> {
+        let span = tracing::info_span!(
```
itowlson, line 394: Could this be moved into a `create_span(addr, request)` type function (and then we could, if desired, retain the existing structure, but with a call to that function before the `self_.handle`)?  It feels like this is mixing the concerns of "defining the instrumentation" and "handling the request".

crates/trigger-http/src/handler.rs
```diff
@@ -17,15 +17,15 @@ use spin_trigger::TriggerAppEngine;
...
+use tracing::{instrument, Instrument, Level};
 use wasmtime_wasi_http::{proxy::Proxy, WasiHttpView};
 
 #[derive(Clone)]
 pub struct HttpHandlerExecutor;
 
 #[async_trait]
 impl HttpExecutor for HttpHandlerExecutor {
-    #[instrument(name = "execute_wasm", skip_all, fields(otel.kind = "server"))]
+    #[instrument(name = "spin_trigger_http.execute_wasm", skip_all, err(level = Level::INFO), fields(otel.name = format!("execute_wasm_component {}", component_id)))]
```
lann, line 28: Micro-nit:

```suggestion
    #[instrument(name = "spin_trigger_http.execute_wasm", skip_all, err(level = Level::INFO), fields(otel.name = format!("execute_wasm_component {component_id}")))]
```

calebschoepp:
@lann @itowlson I spiked out an alternative way for us to do the instrumenting in trigger-http. I don't love it, but it feels a little less invasive.

Let me know what you think. I'm trying to sus out if I should proceed in this direction or not.

itowlson:
Thank you for taking the time to explore options here!  I do share your queasiness about the current iteration but I think you might be onto something.  What I don't yet love is the `finish_instrumenting(the_actual_response, blah, blah)` because 1) it demotes the thing, the _actual purpose of all this,_ into a mere input to the instrumentation function, and 2) it feels odd to see a `finish` function with no start!  I'm wondering if this could be tackled with something like

```rust
            let resp = match well_known {
                "health" => Ok(Response::new(body::full(Bytes::from_static(b"OK")))),
                "info" => self.app_info(),
                _ => Self::not_found(NotFoundRouteKind::WellKnown),
            };
            instrument(&resp);  // or end_current_span_with(&resp) or something
            return resp;
```

This seems to separate the "I need to capture the result in the span" from the logic of selecting the path and constructing the appropriate response.

You could even combine the last two lines into something like a `return_from_span!(resp)` macro - I have no idea if that's a good idea or a bad idea though!

I like how `http_span!` tidies away that overwhelming pile of span stuff.  Although now you have it I'd definitely consider getting rid of `instrumented_servcice_fn` and doing it all inline.  (Or just calling it `service_fn` or the like.  `instrumented` is not its purpose - I know it's front of mind for you but the _purpose_ is handling the request: the instrumentation is an implementation detail (unless it's being distinguished from another one which isn't instrumented).)

I hope that's helpful - I'm all too aware that I may be being constantly dissatisfied without knowing what I actually want other than "can you make it happen by magic"...!

crates/trigger-redis/src/spin.rs
```diff
@@ -11,6 +12,7 @@ pub struct SpinRedisExecutor;
 
 #[async_trait]
 impl RedisExecutor for SpinRedisExecutor {
+    #[instrument(name = "spin_trigger_redis.execute", skip(self, engine, payload), err(level = Level::INFO), fields(otel.name = format!("execute_wasm_component {}", component_id)))]
```
endocrimes, line 15:
```suggestion
    #[instrument(name = "spin_trigger_redis.execute_wasm", skip(self, engine, payload), err(level = Level::INFO), fields(otel.name = format!("execute_wasm_component {}", component_id)))]
```

^nit for consistency

calebschoepp:
> Thank you for taking the time to explore options here! I do share your queasiness about the current iteration but I think you might be onto something. What I don't yet love is the `finish_instrumenting(the_actual_response, blah, blah)` because 1) it demotes the thing, the _actual purpose of all this,_ into a mere input to the instrumentation function, and 2) it feels odd to see a `finish` function with no start! I'm wondering if this could be tackled with something like

I hear those concerns and agree with them.

> ```rust
>             let resp = match well_known {
>                 "health" => Ok(Response::new(body::full(Bytes::from_static(b"OK")))),
>                 "info" => self.app_info(),
>                 _ => Self::not_found(NotFoundRouteKind::WellKnown),
>             };
>             instrument(&resp);  // or end_current_span_with(&resp) or something
>             return resp;
> ```
> 
> This seems to separate the "I need to capture the result in the span" from the logic of selecting the path and constructing the appropriate response.

Unfortunately what you've proposed above wouldn't work b/c the `&resp` alone is not sufficient to get the route. I need to explicitly pass the route to my instrumentation function somehow and also tell it whether the route is high or low cardinality (this tells me if I should put it in the span name or not). Needing the route is how I've ended up wrapping all the returns.

I tried wrapping the whole handle function with a single instrument function and then passing the route via an [extension](https://docs.rs/http/latest/http/struct.Extensions.html) on the `req`, but I couldn't get that to work.

Another alternative would be to wrap the whole handle in an instrumented function like below. The downside to this is we would need to explicitly return the route from handle.

```rust
fn instrument_handle(response: Result<(Response<T>, String)>) -> Result<Response<T>> {
  // handle error case
  let (response, route) = response;
  // Do a bunch of otel stuff
  Ok(response)
}

// in service_fn
let span = http_span!(request, addr);
instrument_handle(self.handle(...).instrument(span).await)
```

What do you think about that as an alternative?

> You could even combine the last two lines into something like a `return_from_span!(resp)` macro - I have no idea if that's a good idea or a bad idea though!

I don't either 🙃 🤷 

> I like how `http_span!` tidies away that overwhelming pile of span stuff. Although now you have it I'd definitely consider getting rid of `instrumented_servcice_fn` and doing it all inline. (Or just calling it `service_fn` or the like. `instrumented` is not its purpose - I know it's front of mind for you but the _purpose_ is handling the request: the instrumentation is an implementation detail (unless it's being distinguished from another one which isn't instrumented).)

Make sense to me.

> "can you make it happen by magic"...!

🪄 
(In all seriousness though thank you for your recognition of this and it is no problem. I appreciate your tenacious hunt for the best code.)

itowlson:
> Unfortunately what you've proposed above wouldn't work b/c the &resp alone is not sufficient to get the route. I need to explicitly pass the route to my instrumentation function somehow...

I am probably being obtuse but I don't get why that would be an obstacle.  You still have that info after the match expression as far as I can tell - I was just lazy and didn't write it into my example. _grin_

```rust
        if let Some(well_known) = path.strip_prefix(spin_http::WELL_KNOWN_PREFIX) {
            let resp = match well_known {
                "health" => Ok(Response::new(body::full(Bytes::from_static(b"OK")))),
                "info" => self.app_info(),
                _ => Self::not_found(NotFoundRouteKind::WellKnown),
            };
            instrument_me_up(&resp, req.method(), path);
            return resp;
        }
```

Is it something to do with not wanting to instrument the path if it's a 404?  Could that not be handled by examining the response status in the `finish_instrumenting / instrument / whatever` function?

itowlson:
I forgot to comment on the second part.  Yes, a wrapping function that does the instrumentation after the _whole_ of `handle` makes a great deal of sense to me.  I had assumed there was a reason you weren't doing it that way - I know there is an imperial buttload of fiddly exit paths and assumed it was something to do with that - but yes, if you can put it all into a little middleware helper, then that achieves a very clean separation of concerns.

(This may be what you were aiming for with `instrumented_service_fn`, I am not sure?  If so, and if it can take all/most of the OTel stuff off handle's shoulders, that would 100% justify its existence and I would be delighted to eat my words over it!)

calebschoepp:
Okay @itowlson I took another crack at this. The code is still very rough, but I'm curious what you think about the latest approach. I like that it keeps OTEL code out of `handle` for the most part. I don't like that it requires me to return a custom `RoutedResponse` type. Unfortuantely, that was the only way I could think to get that route information out of the `handle` function.

crates/trigger-http/src/instrument.rs
```diff
@@ -24,39 +25,43 @@ macro_rules! http_span {
...
+) -> Result<Response<HyperIncomingBody>> {
     let span = tracing::Span::current();
+    match foo {
+        Ok(RoutedResponse(response, route)) => {
+            // Set otel.name and http.route
+            if let Some(route) = route {
+                span.record("http.route", &route);
+                span.record("otel.name", format!("{method} {route}"));
+            } else {
+                span.record("otel.name", method);
```
endocrimes, line 47:
```suggestion
                span.record("otel.name", "{method} unknown");
```

would be easier to filter for in trace management software

itowlson:
@calebschoepp I'm reading this but have gotten myself confused about what "route" means.  Sometimes it seems to be the request path, but other times it seems to be the matched route pattern.  Am I misreading?  Which is it intended to be?  Thanks!

calebschoepp:
> @calebschoepp I'm reading this but have gotten myself confused about what "route" means.  Sometimes it seems to be the request path, but other times it seems to be the matched route pattern.  Am I misreading?  Which is it intended to be?  Thanks!
> 

95% of the time it will be the matched route pattern. 5% of the time when well known endpoints are hit it will be a path (but really you can just think of that as the matched route).

itowlson:
@calebschoepp This feels like a viable approach, although I sympathise with your dislike of adding an extra type just to carry the instrumentation data.  If we go ahead with this, there is definitely stuff we can do to make it less intrusive, e.g. constructors for `RoutedResponse` or a `with_matched_route()` helper on Response or such like.  Thanks for all the investigating and experimenting!

calebschoepp:
> @calebschoepp This feels like a viable approach, although I sympathise with your dislike of adding an extra type just to carry the instrumentation data.  If we go ahead with this, there is definitely stuff we can do to make it less intrusive, e.g. constructors for `RoutedResponse` or a `with_matched_route()` helper on Response or such like.  Thanks for all the investigating and experimenting!

Happy to experiment.

If we're generally happy with going this direction then I think the next step would be a more thorough code review from you for ideas on how we'll clean it up. With the caveat that obviously I'll go back and full in the todo comments.

crates/trigger-http/src/instrument.rs
lann, line 1: I wonder if this would be better off in the spin-telemetry crate? It might be nice to have both inbound and outbound http telemetry there.

crates/trigger-http/src/instrument.rs
```diff
@@ -0,0 +1,69 @@
...
+            "http.route" = Empty,
+            "otel.name" = Empty,
+        )
+    };
+}
+
+pub(crate) use http_span;
+
+/// TODO
+pub struct RoutedResponse(pub Response<HyperIncomingBody>, pub Option<String>);
```
lann, line 32: I do think it should be possible (and preferable) to put the route info into a response extension.

calebschoepp: I'll give that a try. I guess I only tried putting it in the request extensions which had lifetime issues.

Are response extensions scoped to each specific response or are they global? We're going to run into issues if it's global.

lann: Extensions are scoped to the response instance. They're keyed by the extension value's [`TypeId`](https://doc.rust-lang.org/std/any/struct.TypeId.html) so you'll want a unique type e.g. `struct HttpRoute(String)`.

crates/trigger-http/src/instrument.rs
rylev: I agree on the suggestion, but slight nit:
```suggestion
                span.record("otel.name", format!("{method} <unknown route>");
```

crates/trigger-http/src/instrument.rs
rylev: If we don't go the extension route, then I would at the very least give the fields names instead of using a tuple struct.

rylev (changes requested):
Generally happy with the direction - I've requested changes so that this doesn't accidentally get merged without the various TODOs being cleaned up first.

crates/trigger-http/src/instrument.rs
lann: The value as-is matches the [OTel semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name)

> If there is no (low-cardinality) `http.route` available, HTTP server span names SHOULD be `{method}`.

calebschoepp: ^ Yep, I'm just trying to follow the OTEL semantic conventions here.

crates/trigger-http/src/instrument.rs
calebschoepp: My gut says to keep it in `trigger-http`. The outbound stuff in telemetry is used in multiple places, but the inbound stuff would only be used in `trigger-http`.

crates/trigger-http/src/instrument.rs
calebschoepp: Resolving this b/c I got the extension route working.

crates/trigger-http/src/instrument.rs
lann: Okey dokey; easy enough to move around later if desired.

crates/trigger-http/src/lib.rs
```diff
@@ -240,19 +249,25 @@ impl HttpTrigger {
...
-                "info" => self.app_info(),
+                "health" => {
+                    let mut response = Response::new(body::full(Bytes::from_static(b"OK")));
+                    response
+                        .extensions_mut()
+                        .insert(MatchedRoute { route: path });
+                    Ok(response)
+                }
+                "info" => self.app_info(path),
                 _ => Self::not_found(NotFoundRouteKind::WellKnown),
```
lann, line 265: Minor: you could probably avoid a little complexity by setting the extension out here, e.g.:

```rust
let mut resp = match well_known {
  ...
};
resp.extenstions_mut()...
return Ok(resp)
```
...though maybe that doesn't work well with the `NotFoundRouteKind::WellKnown`? :shrug:

crates/trigger-http/src/lib.rs
```diff
@@ -293,10 +308,16 @@ impl HttpTrigger {
                     }
                 };
                 match res {
-                    Ok(res) => Ok(res),
+                    Ok(mut res) => {
+                        res.extensions_mut().insert(MatchedRoute {
+                            route: raw_route.to_string(),
+                        });
```
lann, line 314: Minor: a nice pattern for this sort of thing (when you are repeating it):

```rust
impl MatchedRoute {
  fn set_response_extension(resp: &mut Response, route: impl Into<String>) { ... }
  // and / or
  fn with_response_extension(mut resp: Response ,route: impl Into<String>) -> Response { ... }
}

MatchedRoute::set_response_extension(resp, raw_route);
```

lann (approved):
A couple of minor/non-blocking suggestions; lgtm!

crates/trigger-http/src/lib.rs
calebschoepp: Correct, doesn't work b/c we don't want to set the route in the not found case.

crates/trigger-http/src/lib.rs
calebschoepp: Nice

rylev (approved):

PR #2446, opened 2024-04-10
Add Redis components to existing app

### LINKED ISSUE #2315: Enable `spin add` for Redis templates
While multiple components can be used in a single application,  we can't `spin add` a Redis component to an existing Redis trigger app or an existing HTTP trigger app:

```bash
# Start with a blank canvas; use the `http-empty` template to create a new Spin application
$ spin new -t http-empty trigger-example
Description: An new Spin application
# Change into the new Spin application directory
$ cd trigger-example
# Add a new HTTP component to the application
$ spin add -t http-rust http-trigger-example
Description: HTTP trigger example
HTTP path: /...
# Add a new Redis component to the application
$ spin add -t redis-rust redis-trigger-example
Template redis-rust doesn't support the 'add component' operation
```

At present, if creating a multi-trigger application (using the `spin new` and `spin add` commands) a user must start with a Redis trigger app and then `spin add` the HTTP trigger component. 
The user can't add `spin add` an additional Redis trigger component to this existing Redis app either.

itowlson:
Fixes #2315

### DIFF
--- a/crates/templates/src/lib.rs
+++ b/crates/templates/src/lib.rs
@@ -17,6 +17,7 @@ mod run;
 mod source;
 mod store;
 mod template;
+mod toml;
 mod writer;
 
 pub use manager::*;
--- a/crates/templates/src/reader.rs
+++ b/crates/templates/src/reader.rs
@@ -33,6 +33,48 @@ pub(crate) struct RawTemplateVariant {
     pub skip_files: Option<Vec<String>>,
     pub skip_parameters: Option<Vec<String>>,
     pub snippets: Option<HashMap<String, String>>,
+    pub conditions: Option<HashMap<String, RawConditional>>,
+}
+
+#[derive(Debug, Deserialize)]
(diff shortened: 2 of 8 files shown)


crates/templates/src/manager.rs
```diff
@@ -172,7 +172,7 @@ impl TemplateManager {
                     .file_name()
                     .map(|s| s.to_string_lossy().to_string())
                     .unwrap_or_else(|| format!("{}", source_dir.display()));
-                let message = format!("{}", e);
+                let message = format!("{:#}", e);
```
fibonacci1729, line 175: Is `{:#}` correct? If so, what does it do? I'm only familiar with it in the context of `{:#?}` (or some alternative spec of printing).

itowlson: It is valid (the alternate display representation, which for errors includes the context stack), but was left in unintentially.  Thanks for flagging.

fibonacci1729 (approved):

PR #2451, opened 2024-04-11
Bikeshed private endpoint UI

itowlson:
Per request from @ThorstenHans 

I don't have a strong feeling one way or the other.

### DIFF
--- a/crates/http/src/config.rs
+++ b/crates/http/src/config.rs
@@ -18,7 +18,15 @@ pub struct HttpTriggerConfig {
 #[serde(untagged)]
 pub enum HttpTriggerRouteConfig {
     Route(String),
-    IsRoutable(bool),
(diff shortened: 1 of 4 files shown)


lann:
I'd probably vote for `private = true` over `public = false`; `route = false` was a sneaky way to reuse the key but if we're adding an alternative key I'd default to a positive predicate.

crates/http/src/config.rs
```diff
@@ -18,7 +18,15 @@ pub struct HttpTriggerConfig {
 #[serde(untagged)]
 pub enum HttpTriggerRouteConfig {
     Route(String),
-    IsRoutable(bool),
+    Private(HttpPriveteEndpoint),
```
radu-matei, line 21: Same typo as below.

fibonacci1729 (approved):

PR #2453, opened 2024-04-12
Summarise runtime config

### LINKED ISSUE #1852: Summarise runtime config at startup
Via @mikkelhegn, broken out from #1850:

>  I still think at least an acknowledgement of what runtime config was loaded would be great. ...
>
> `Using [llm_remote] and [key_value_store.remore_redis] runtime-config from /mikkels/fancy/setup.toml`
>
> That level of detail I think would be a good balance.

(Ivan again) Printing the file name is particular attractive if the runtime config is coming from an environment variable where it's not visible on the command line.

itowlson:
Fixes #1852

### DIFF
--- a/crates/trigger/src/runtime_config.rs
+++ b/crates/trigger/src/runtime_config.rs
@@ -15,6 +15,8 @@ use serde::Deserialize;
 use spin_common::ui::quoted_path;
 use spin_sqlite::Connection;
 
+use crate::TriggerHooks;
+
 use self::{
     key_value::{KeyValueStore, KeyValueStoreOpts},
     llm::LlmComputeOpts,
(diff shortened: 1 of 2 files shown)


rylev (approved):

calebschoepp (approved):
Yay!

lann (approved):

calebschoepp:
Let's see how many approvals we can get on this PR before Ivan merges it 🚢

karthik2804 (approved):

crates/trigger/src/runtime_config.rs
```diff
@@ -259,6 +261,52 @@ fn resolve_config_path(path: &Path, config_opts: &RuntimeConfigOpts) -> Result<P
...
+        &mut self,
+        _app: &spin_app::App,
+        runtime_config: &RuntimeConfig,
+        _resolver: &Arc<spin_expressions::PreparedResolver>,
+    ) -> anyhow::Result<()> {
+        if let Some(path) = &self.runtime_config_file {
+            let mut opts = vec![];
+            for opt in runtime_config.opts_layers() {
+                for id in opt.key_value_stores.keys() {
+                    opts.push(format!("[key_value_store.{id}]"));
```
endocrimes, line 287: Should these also include types? That would be helpful for validating the right version of a config between dev/deployments

itowlson: @endocrimes Could you be specific about what you mean by "include types"?  Are you thinking e.g. whether the KV backing store is SQLite vs Redis vs Cosmos, etc.?

endocrimes: Exactly!

itowlson: Added backing type info.  It now looks like:

```
$ spin up -f ~/testing/kvmadness/ --runtime-config-file ~/testing/kvmadness/runtime-config.toml 
Logging component stdio to "/home/ivan/testing/kvmadness/.spin/logs/"
Using [key_value_store.default: spin], [key_value_store.bibblybobbly: spin] runtime config from "/home/ivan/testing/kvmadness/runtime-config.toml"
Storing default key-value data to ".spin/my_super_duper_store.db"
Storing default SQLite data to "/home/ivan/testing/kvmadness/.spin/sqlite_db.db"
```

I am not sure this is the perfect layout - it feels a bit spindly - but we can presumably iterate without causing a disaster.  But if folks hate it on sight I'm open to other ideas!

fibonacci1729 (approved):

endocrimes (approved):
This is excellent thanks @itowlson!

lann (approved):

rylev (approved):

calebschoepp (approved):

calebschoepp:
10 approvals isn't bad

Issue #2237, opened 2024-01-21
how read the current wasm module / application metadata


rgl:
how can I read the current wasm module or application metadata?

for example, to display the version?

are [application variables](https://developer.fermyon.com/spin/v2/variables) meant to be used for this? if so, is there some template syntax to get the application version, etc?

rylev:
Variables would certainly be the most appropriate way to accomplish this but there is no concept of "meta" variables that are prepopulated with certain information so for now you'd have to copy that information into a Spin.toml manifest's variables section. 

I will change this issue to track the feature request for having meta variables.

It would be great to know what some use cases are for those so that it's clearer which variables would be useful to include by default.

rgl:
Got it. Thank you! For now, I think I will shortcut this, and just include this in a separate source file before compiling the app.

Only have two use-cases, mainly for troubleshoot purposes:

1. Get the application information (e.g. its version) at runtime, so I can have it somewhere available to show in the UI/API/logs.
2. Get the host information (e.g. the name of the runtime and its version).

rgl:
I've just found out about https://github.com/bytecodealliance/wasm-tools, which contains this nugget:

> `wasm-tools metadata show` [wasm-metadata](https://crates.io/crates/wasm-metadata)	Show name and producer metadata in a component or module

It seems to be the metadata that I'm looking for. Is there a way to set it and access it from the spin sdk?

Issue #1738, opened 2023-09-06
Read access to (some) spin.toml's value from components(applications) 


seungjin:
To read a file from my (sub)component in a Rust project, 
I need to specify its path starting from the very beginning of the project.
```
let reader = maxminddb::Reader::open_readfile(
       "geoip/geodb/GeoLite2-City.mmdb"
    ).unwrap();
```
If I can read its working directory from spin.toml (component.build.workdir), I can make its file path more neutral with respect to the component's name (i.e., "{component.build.workdir}/geodb/GeoLite2-City.mmdb"). 

Additionally, allowing the application to read 'id,' 'workdir,' 'files,' 'allowed_http_hosts,' and 'route' from spin's component section would provide more flexibility and creativity for spin users.

karthik2804:
I have also run into cases where having some metadata from the `spin.toml` available to the component would be nice. The case I ran into while I was building KV/SQLite explorer apps was that it would be nice to have access to the list of DBs/stores available to the component so that they can be exposed on the UI. It can also used to provide advance warning messages on UI without having to attempt to open the store if no stores are available.

Exploring this a little further - Do we think that components must have access to all the configurations in the `spin.toml` for that component or do we think that a restricted subset of it would be sufficient?

seungjin:
I kept sending a request object to other functions to retrieve various pieces of environment information (such as project details, URLs, etc.), or I needed to store the request object in a global location. It was quite inconvenient. If I could directly retrieve some basic information from spin.toml, it would be much simpler.

Opening all configurations may seem risky, but we definitely have some very common information that can be safely shared.

PR #2454, opened 2024-04-12
SIP for guest access to application metadata

itowlson:
Apropos of #1738 and #2237 (and possibly another one that I can't find right now _grin_)

### DIFF
--- a/docs/content/sips/018-application-metadata.md
+++ b/docs/content/sips/018-application-metadata.md
@@ -1,0 +1,73 @@
+title = "SIP 018 - Application Metadata"
+template = "main"
+date = "2024-04-12T23:00:00Z"
+
+---
+
+Summary: Allow guest code to access information from the application manifest
+
+Owner(s): ivan.towlson@fermyon.com
+
+Created: Apr 12, 2024
+
+# Requirements and scope
+
+We have, over time, had a few requests for guest code to be able to read fields from the application manifest, or from the host itself. Examples include:
+
+| Field                                         | Purpose          | Notes |
+|-----------------------------------------------|------------------|-------|
+| Application name and version                  | Logging          |       |
+| Host runtime name or other identifying string | Logging          |       |
+| Host runtime version                          | Logging          |       |
+| Component id                                  | Logging (?)      |       |
+| Files mounted to the component                | Not given        | Already available by traversing the component's file system |
+| Allowed outbound hosts for the component      | Not given        | One user mentioned their app needed to scan a set of sites but I don't have details |
+| Allowed KV/SQLite stores for the component    | Explorer UI      |       |
+| Component HTTP route                          | Not given        | Already available via the `spin-raw-component-route` header |
+| The component's build directory               | Guest files mapped at `workdir` | Not clear why user mapped to the subdirectory |
+
+## Should we?
+
+We've been a bit wary of doing some of this stuff because the Wasm model is that components should define the set of capabilities they consume, and be granted only those capabilities, so there should be no need for them to be going around sniffing to find out what goodies they have access to. This concern applies specifically to the "allowed" lists (outbound hosts and KV/SQLite). Possibly this would be addressed by requiring components to declare access.
+
+Also, I'd argue we should _not_ provide access to `workdir`. This is, I believe, not currently available at runtime; and in any case it should have nothing to do with what happens at runtime - build is way distant in the rear view mirror by then. This is better handled by mapping files up a level or by passing a distinct "file root" variable.
+
+| Field                                         | Recommendation   |
+|-----------------------------------------------|------------------|
+| Application name and version                  | Unproblematic    |
+| Host runtime name or other identifying string | Unproblematic    |
+| Host runtime version                          | Unproblematic    |
+| Component id                                  | Unproblematic    |
+| Files mounted to the component                | Superfluous      |
+| Allowed outbound hosts for the component      | Caution advised  |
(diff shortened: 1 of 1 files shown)


docs/content/sips/018-application-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+If we do wish to expose this data, there are various ways we could do so:
+
+1. Provide read access to the manifest file itself. This couples application logic to the manifest format du jour. We should not do this.
+
+2. Create a new API that reads specific fields e.g. `application_meta::component_id()`. (A variant would be a single function accepting a field ID, e.g. `application_meta::get(AppMeta::ComponentId)`.) This is safe and discoverable, but does not line up with any broader conventions, and may have versioning implications if we need to add additional fields over time.
+
+3. Provide environment variables containing the information e.g. `SPIN_APP_META_COMPONENT_ID`.
+
+4. Wedge it into an existing API such as `variables` - e.g. have some special magic variable names that are implicitly mapped to these values. This limits us to strings, but provides a nice built-in permissions mechanism where a component has to indicate in its manifest what data it plans to exfiltrate.
+
```
rylev, line 60: I believe there's another possibility: provide a an access to the manifest or some subset there of that always returns optional values. Those functions would return `none` when the Spin.toml manifest does not explicitly grant access to the manifest through some new configuration (`manifest_read_access = true` or similar).

docs/content/sips/018-application-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+# User experience
+
+## Access
+
+If we do wish to expose this data, there are various ways we could do so:
+
+1. Provide read access to the manifest file itself. This couples application logic to the manifest format du jour. We should not do this.
+
+2. Create a new API that reads specific fields e.g. `application_meta::component_id()`. (A variant would be a single function accepting a field ID, e.g. `application_meta::get(AppMeta::ComponentId)`.) This is safe and discoverable, but does not line up with any broader conventions, and may have versioning implications if we need to add additional fields over time.
```
rylev, line 55: Nit: I think we should mention a possible middle ground between complete access to the manifest and access to individual fields -  we could bundle a certain subset of manifest data together and provide one new function that provides access to that data.

docs/content/sips/018-application-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+| Application name and version                  | Unproblematic    |
+| Host runtime name or other identifying string | Unproblematic    |
+| Host runtime version                          | Unproblematic    |
+| Component id                                  | Unproblematic    |
+| Files mounted to the component                | Superfluous      |
+| Allowed outbound hosts for the component      | Caution advised  |
+| Allowed KV/SQLite stores for the component    | Caution advised  |
+| Component HTTP route                          | Superfluous      |
+| The component's build directory               | Should not       |
+
```
rylev, line 46: There's also a backwards compatibility concern about how new manifest data would be exposed - this could lead to any additions to the manifest requiring changes to the Spin wit world which does not sound ideal.

rylev (commented):
Thanks for the write up! 

I agree that there are certainly issues with providing unrestricted access to all data in the manifest. It does seem like we should settle on some subset and expose that to the application.

Nit: this SIP seems pretty unsure about what it wants to actually propose. Perhaps this would be best to discuss in an issue before opening a SIP that has a more concrete proposal.

lann:
I think my hesitation with this feature request has been the lack of motivating use cases. Off the top of my head I can imagine different categories that might suggest different designs:

- Developer tools (e.g. spin-kv-explorer) to reflect on runtime config
- O11y libraries to inject e.g. app name into traces
- Terrible hacks to work around deficiencies in Spin APIs

docs/content/sips/018-application-metadata.md
```diff
@@ -0,0 +1,73 @@
...
+We have, over time, had a few requests for guest code to be able to read fields from the application manifest, or from the host itself. Examples include:
+
+| Field                                         | Purpose          | Notes |
+|-----------------------------------------------|------------------|-------|
+| Application name and version                  | Logging          |       |
+| Host runtime name or other identifying string | Logging          |       |
+| Host runtime version                          | Logging          |       |
+| Component id                                  | Logging (?)      |       |
+| Files mounted to the component                | Not given        | Already available by traversing the component's file system |
+| Allowed outbound hosts for the component      | Not given        | One user mentioned their app needed to scan a set of sites but I don't have details |
```
tpmccallum, line 24: This is one that I am interested in also.
A use case would be a Spin application that accepts jobs to check for broken links. The Spin application fetches links (broken link check jobs) to be checked from a central database. To prevent asking for jobs that the application is not willing to check, the application could send a list of links it is prepared to check when asking for jobs. But to do this the application needs to have access to the `allowed_outbound_hosts` config i.e. `["https://www.fermyon.com/", "https://developer.fermyon.com"]. The central database can filter on these hosts and just send actionable links back to the Spin app.

tpmccallum: This would save double handling a list of hosts in the config and a list of hosts that the application can access while running.

tpmccallum: Here is an example of the workflow between a Spin application that checks links and the central database.

image

lann: FWIW, I would generally expect a link checker to have `allowed_outbound_hosts = ["https://*"]`.

tpmccallum: That is a fair point :heavy_check_mark:

PR #2458, opened 2024-04-16
Refactor expressions `Resolver` to provide sync API

rylev:
For `spin-test`, we need a way to resolve expressions that isn't async which is fine because there is no need to use `Provider` types. 

This refactors the `Resolver` into two:
* `Resolver`: a sync version of resolving that is not extended by any `Provider` types.
* `ProviderResolver`: the status quo - an async version of `Resolver` that is extended by `Provider` types.

### DIFF
--- a/crates/expressions/src/lib.rs
+++ b/crates/expressions/src/lib.rs
@@ -9,6 +9,79 @@ pub use provider::Provider;
 use template::Part;
 pub use template::Template;
 
+/// A [`ProviderResolver`] that can be shared.
+pub type SharedPreparedResolver =
+    std::sync::Arc<std::sync::OnceLock<std::sync::Arc<PreparedResolver>>>;
+
+/// A [`Resolver`] which is extended by [`Provider`]s.
+#[derive(Debug, Default)]
+pub struct ProviderResolver {
+    internal: Resolver,
+    providers: Vec<Box<dyn Provider>>,
+}
+
+impl ProviderResolver {
+    /// Creates a Resolver for the given Tree.
+    pub fn new(variables: impl IntoIterator<Item = (String, Variable)>) -> Result<Self> {
+        Ok(Self {
+            internal: Resolver::new(variables)?,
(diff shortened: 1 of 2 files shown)


itowlson (changes requested):
The implementation looks fine but for me the naming is now confusing.  `Resolver` implies this is the "default" resolver that Spin code should use, whereas from the PR description it's almost more like a test mock or stub, is that right?  I'd suggest something like ProviderResolver / StaticValuesResolver, something that draws out the distinction rather than one of them being _the_ resolver (and to be honest if one of them is _the_ resolver then it seems like it's the async provider one).

Related, it would be good to draw out the distinction between PreparedResolver and NuResolver, because they feel very similar.  That might be ApplicationValuesResolver vs ComponentValuesResolver (to reflect their different interfaces), or ProviderSnapshotResolver vs TestValuesResolver (to reflect their different purposes), or something like that maybe?

rylev:
I'm not attached to any naming, so we can certainly change things! :smile:

The name `Resolver` for the new resolver was chosen because it's effectively the more standard of the two resolvers. The `ProviderResolver` is an extension of the simpler `Resolver` doing the same thing but adding the ability to resolve through configurable providers. The simpler name doesn't imply "defaultness" to me but rather being the base on which other resolvers are built. 

While the `Resolver` is being used in mocks for `spin-test`, there's nothing here that necessarily means it needs to be used in that context so I think trying to name that way feels odd. The `Resolver` is just a type that resolves values for keys. It's only different from the `ProviderResolver` in that it doesn't use providers to aid in that resolution.

The names ApplicationValuesResolver and ComponentValuesResolver don't really make sense to me as all of these resolvers are used in the context of apps and components - they're not really differentiated by that use at all. 

So, I'm happy to rename things, but the current naming makes sense to me and the suggestions don't feel like improvements, so I'm not sure where to go from here.

itowlson (approved):

PR #2459, opened 2024-04-16
ref(outbound-http): Add a small hack to improve the tracing of outbound http requests through spin core

calebschoepp:


alexcrichton (approved):

crates/core/src/lib.rs
```diff
@@ -195,7 +195,7 @@ impl<T: Send + OutboundWasiHttpHandler> WasiHttpView for Data<T> {
         &mut self.table
     }
 
-    #[instrument(name = "start_outbound_http_request", skip_all, fields(otel.kind = "client"))]
+    #[instrument(name = "spin_core.send_request", skip_all, fields(otel.kind = "client", url.full = request.request.uri().to_string(), http.request.method = request.request.method().to_string(), otel.name = request.request.method().to_string(), http.response.status_code = Empty, server.address = Empty, server.port = Empty))]
```
lann, line 198: I probably missed this all over your previous PRs, but `tracing` has a shorthand for calling `Display::to_string()`:
```suggestion
    #[instrument(name = "spin_core.send_request", skip_all, fields(otel.kind = "client", url.full = %request.request.uri(), http.request.method = %request.request.method(), otel.name = %request.request.method(), http.response.status_code = Empty, server.address = Empty, server.port = Empty))]
```

crates/trigger-http/src/lib.rs
```diff
@@ -706,7 +706,35 @@ impl OutboundWasiHttpHandler for HttpRuntimeData {
...
+                            >,
+                            anyhow::Error,
+                        > = handle.await;
+                        if let Ok(Ok(res)) = &res {
+                            tracing::Span::current()
+                                .record("http.response.status_code", res.resp.status().as_u16());
+                        }
+                        res
+                    }
+                    .instrument(tracing::Span::current()),
```
lann, line 731:
```suggestion
                    .in_current_span(),
```

crates/trigger-http/src/lib.rs
```diff
@@ -706,7 +706,35 @@ impl OutboundWasiHttpHandler for HttpRuntimeData {
...
+        // Once Wasmtime gives us the ability to do the spawn ourselves we can just call .instrument
+        // and won't have to do this workaround.
+        let response_handle = wasmtime_wasi_http::types::default_send_request(data, request)?;
+        let response = data.table().get_mut(&response_handle)?;
+        *response = match std::mem::replace(response, HostFutureIncomingResponse::Consumed) {
+            HostFutureIncomingResponse::Pending(handle) => {
+                HostFutureIncomingResponse::Pending(wasmtime_wasi::preview2::spawn(
+                    async move {
+                        let res: std::prelude::v1::Result<
+                            std::prelude::v1::Result<
```
lann, line 719: Yes, `Result` is `anyhow::Result`, but it will work anyway
```suggestion
                        let res: Result<
                            Result<
```

crates/core/src/lib.rs
```diff
@@ -205,6 +205,15 @@ impl<T: Send + OutboundWasiHttpHandler> WasiHttpView for Data<T> {
...
         Self: Sized,
     {
+        let current_span = tracing::Span::current();
+        let uri = request.request.uri();
+        if let Some(authority) = uri.authority() {
+            current_span.record("server.address", authority.host());
+            if let Some(port) = authority.port() {
+                current_span.record("server.port", port.as_u16());
+            }
+        }
```
lann, line 215: Could the two chunks of code in this PR move into the same function?

calebschoepp: Not sure what you mean by that

lann: This PR currently adds telemetry code to both this function in `spin-core` and the function in `spin-trigger-http`. Could the `http.response.status_code` code instead be up here in this function?

lann (approved):

fibonacci1729 (approved):

PR #2460, opened 2024-04-17
Check for illegal file name when copying single file

### LINKED ISSUE #2361: In `component.files`, mapping a single file to `/` gives an unhelpful error

itowlson:
Fixes #2361.

Example output:

```
$ spin up -f ./crates/loader/tests/file-errors/bad.toml 
Error: Failed to load manifest from "./crates/loader/tests/file-errors/bad.toml"

Caused by:
   0: Failed to load Spin app from "./crates/loader/tests/file-errors/bad.toml"
   1: Failed to load component `bad`
   2: Failed to copy "/home/ivan/github/spin/crates/loader/tests/file-errors/host.txt" to working path "/tmp/spinup-x8yn3O/assets/bad/": "/" is not a valid destination file name
```

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -269,7 +269,8 @@ impl LocalLoader {
             } => {
                 let src = Path::new(source);
                 let dest = dest_root.join(destination.trim_start_matches('/'));
-                self.copy_file_or_directory(src, &dest, exclude_files).await
+                self.copy_file_or_directory(src, &dest, destination, exclude_files)
+                    .await
             }
         }
     }
@@ -291,7 +292,7 @@ impl LocalLoader {
                     .await?;
             } else {
                 // "single/file.txt"
-                self.copy_single_file(&path, &dest).await?;
+                self.copy_single_file(&path, &dest, glob_or_path).await?;
             }
         } else if looks_like_glob_pattern(glob_or_path) {
             // "glob/pattern/*"
@@ -308,6 +309,7 @@ impl LocalLoader {
         &self,
         src: &Path,
         dest: &Path,
+        guest_dest: &str,
         exclude_files: &[String],
     ) -> Result<()> {
         let src_path = self.app_root.join(src);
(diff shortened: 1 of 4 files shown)


crates/loader/src/local.rs
```diff
@@ -394,17 +397,40 @@ impl LocalLoader {
...
+    ) -> anyhow::Result<T> {
+        let src_text = quoted_path(src);
+        let dest_text = quoted_path(dest);
+        let base_msg = format!("Failed to copy {src_text} to working path {dest_text}");
+
+        if let Some(io_error) = e.downcast_ref::<std::io::Error>() {
+            if Self::is_directory_like(guest_dest)
+                || io_error.kind() == std::io::ErrorKind::NotFound
+            {
+                bail!("{base_msg}: \"{guest_dest}\" is not a valid destination file name");
```
rylev, line 421: nit: could we make everything after the `base_msg` into another error (perhaps by using `with_context`) so that that information shows up on a different line when displaying the full error. As it is right now, this error message is very long.

itowlson: Good call.  I broke the message up and it now looks like:

```
$ spin up -f ./crates/loader/tests/file-errors/bad.toml 
Error: Failed to load manifest from "./crates/loader/tests/file-errors/bad.toml"

Caused by:
   0: Failed to load Spin app from "./crates/loader/tests/file-errors/bad.toml"
   1: Failed to load component `bad`
   2: Failed to copy "/home/ivan/github/spin/crates/loader/tests/file-errors/host.txt" to working path "/tmp/spinup-iloZyr/assets/bad/"
   3: "/" is not a valid destination file name
```

which draws out the underlying error much more clearly.  Thanks!

PR #2461, opened 2024-04-18
Possible mitigation for Python smoke flakes

itowlson:


itowlson:
I got it to flake... but it flaked on Vault not on Python.  Fie!

vdice:
@itowlson is this change still something we'd like to get in?

PR #2464, opened 2024-04-22
Granular route matching

### LINKED ISSUE #1923: Allow for more granular route matching in Spin
Currently, the route matching in spin can be of two forms 

```toml
# match a single route
route = "/{path}"

# match any route with the prefix
route = "/{path-prefix}/..."
```
It would be helpful in certain cases if the route matching allowed more granularity. For an example consider the case of a Bartholomew app with a proxy component. 
```toml
[[component]]
id = "bartholomew"
[component.trigger]
route = "/..."

[[component]]
id = "proxy"
[component.trigger]
route = "/spin/*"
```
With such a configuration, the `proxy` component would match the route `/spin/index` but not `/spin/v1/index`.  Thus allowing the `proxy` to handle one specific subset of the routes while `bartholomew` handles the rest. 

There is potential for another level of granularity where we could define something like 
```toml
route = "/spin/*/test"
```
Which would only match `/spin/index/test` but not `/spin/index/something-else`

Thoughts?

itowlson:
Fixes #1923.

This is WIP, pushed for squigglies if I break the integration tests.  It's not yet ready for review.


crates/trigger-http/src/lib.rs
```diff
@@ -487,48 +471,71 @@ fn strip_forbidden_headers(req: &mut Request<Body>) {
...
+
+pub(crate) fn compute_default_headers(
     uri: &Uri,
-    raw: &str,
     base: &str,
     host: &str,
+    route_match: &RouteMatch,
     client_addr: SocketAddr,
-) -> Result<Vec<(&'a [&'a str], String)>> {
+) -> Result<Vec<([Cow<'static, str>; 2], String)>> {
```
lann, line 488: I would return `String`s here. Avoiding allocations is nice and all but all these headers get rewritten a couple of times downstream anyway; probably not worth the extra complexity here.

lann (commented):
Style/nit comments; impl looks good

crates/http/src/routes.rs
```diff
@@ -3,24 +3,32 @@
...
-    pub(crate) routes: IndexMap<RoutePattern, String>,
+    /// Resolves paths to routing information - specifically component IDs
+    /// but also recording about the original route.
+    router: std::sync::Arc<routefinder::Router<RouteHandler>>,
+}
+
+#[derive(Clone, Debug)]
+struct RouteHandler {
+    component_id: String,
+    based_route: String,
```
fibonacci1729, line 22: nit: could you briefly document what these fields are?

crates/http/src/routes.rs
```diff
@@ -52,50 +69,70 @@ impl Router {
...
+
+        Ok((router, duplicates))
+    }
+
+    fn parse_route(based_route: &str) -> Result<(routefinder::RouteSpec, ParsedRoute), String> {
+        if let Some(wild_suffixed) = based_route.strip_suffix("/...") {
+            let rs = format!("{wild_suffixed}/*").try_into()?;
+            let parsed = ParsedRoute::trailing_wildcard(wild_suffixed);
+            Ok((rs, parsed))
+        } else if let Some(wild_suffixed) = based_route.strip_suffix("/*") {
```
fibonacci1729, line 120: If i'm understanding this correctly, wildcard routes can now either be specified with `...` or `*`, is that right? 

If that's the case (It's not a hill i'm willing to die on) but i'd vote for supporting just `...`.

itowlson: Yeah I don't have a strong view on this.  The sorta kinda reasoning was that if people are familiar with `routefinder` style strings, then we shouldn't make them memorise a seemingly capricious difference.  In particular, the Rust SDK `Router` uses unmodified `routefinder` strings (I believe), and it felt weird to largely accept the same strings but with an "except you have to use `/...` here and `/*` there" clause.  But I can validate this away for sure.

fibonacci1729: That sounds reasonable to me!

fibonacci1729 (approved):

lann (approved):

PR #2465, opened 2024-04-22
Setup a docker compose file that creates an o11y stack for Spin to use

calebschoepp:
~This is still a WIP, but~ I'd love feedback on the config itself and on where we want to store this config.

### DIFF
--- a/hack/o11y-stack/docker-compose.yaml
+++ b/hack/o11y-stack/docker-compose.yaml
@@ -1,0 +1,61 @@
+version: "3"
+services:
+  otel-collector:
+    image: otel/opentelemetry-collector-contrib:0.98.0
+    restart: always
+    command:
+      - "--config=/etc/otel-collector-config.yaml"
+    volumes:
+      - ./otel-collector-config.yaml:/etc/otel-collector-config.yaml
+    ports:
+      - "8888:8888" # Prometheus metrics exposed by the collector
+      - "8889:8889" # Prometheus exporter metrics
+      - "4317:4317" # OTLP gRPC receiver
+      - "4318:4318" # OTLP gRPC receiver
+    depends_on:
+      - jaeger
+      - prometheus
+      - tempo
+
+  jaeger:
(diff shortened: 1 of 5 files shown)


rylev:
Can you give me some more context on how you imagine these being used? My gut is that this should be in a different repo, but without knowing about how this will be used, it's hard to say.

calebschoepp:
> Can you give me some more context on how you imagine these being used? My gut is that this should be in a different repo, but without knowing about how this will be used, it's hard to say.

I envision some using this to easily setup an observability stack for their spin app

```shell
cd o11y
docker compose up
# Now do my Spin things
```

We could potentially put this in another repo, or hide it behind a spin plugin, or somewhere else. But, I do think we need to provide a simple and coherent story for how users of Spin can get the very wide set of o11y tools running easily.

itowlson:
Is this solving a Spin-specific observability issue or a "generic observability tools setup" issue?  That might help guide the discussion.

lann:
I don't have a strong opinion about what repo this should live in, but _if_ it's going to be in this repo I would like to put it in e.g. `contrib/o11y-stack` (or `hack/...`, as the k8s ecosystem seems to like) to indicate that it isn't _necessarily_ an actively-supported feature of Spin itself.

As an aside I think `tools/modsurfer` should be moved for the same reason.

calebschoepp:
I like the pattern of `hack/o11y-stack/*`.

calebschoepp:
> Is this solving a Spin-specific observability issue or a "generic observability tools setup" issue? That might help guide the discussion.

Good framing question @itowlson. There is nothing inherently specific to Spin here so I would say this is a "generic observability tools setup" issue.

calebschoepp:
I've moved it to `hack/o11y-stack/`. Looking for another round of reviews on if we're okay storing it there.

lann (approved):
The great thing about `hack/*` is that we can just delete it later :slightly_smiling_face:

PR #2473, opened 2024-04-25
added local_addr() to listeners to display random port numbers

### LINKED ISSUE #2471: `spin up --listen localhost:0` should log the actual bound port number
```console
$ spin up --listen localhost:0
...
Serving http://127.0.0.1:0
$ curl http://127.0.0.1:0
curl: (7) Failed to connect to 127.0.0.1 port 0 after 0 ms: Couldn't connect to server
```

Binding to `:0` causes the OS to pick a random port, which can be useful if you don't have a guaranteed free static port available. We should fix the output to show the actual selected port instead of `:0`.

garikAsplund:
Also had to move some logging info out of run and into serve and serve_tls functions since the thread is blocking. An alternative would be to spawn a new thread but this seemed to work and is the most straightforward approach.

I feel like it'd be good to add tests but I'm not exactly sure the best way to go about that. If anyone has guidance there or feedback for potential issues with this modification please let me know.

Closes #2471 

Screenshot 2024-04-25 at 12 06 43 PM

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -169,27 +169,12 @@ impl TriggerExecutor for HttpTrigger {
         let listen_addr = config.address;
         let tls = config.into_tls_config();
 
-        // Print startup messages
-        let scheme = if tls.is_some() { "https" } else { "http" };
-        let base_url = format!("{}://{:?}", scheme, listen_addr);
-        terminal::step!("\nServing", "{}", base_url);
-        log::info!("Serving {}", base_url);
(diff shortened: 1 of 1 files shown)


lann:
Thanks! I just pushed an update to dedupe a bunch of stuff I noticed while reviewing. lgtm!

lann:
I guess since I made a significant change here too I'd better get a neutral 3rd party to review as well :slightly_smiling_face:

fibonacci1729 (approved):

PR #2474, opened 2024-04-25
chore(crates): address lint errs when rust is at 1.77+

vdice:
A few lint errors block successfully running the lint and test targets locally when on a more recent version of rust than is pinned in Spin/CI.  Hoping these updates are compatible with both scenarios...

### DIFF
--- a/crates/plugins/src/badger/store.rs
+++ b/crates/plugins/src/badger/store.rs
@@ -86,7 +86,7 @@ impl BadgerRecordManager {
         let new = BadgerRecord {
             name: name.to_owned(),
             badgered_from: from.clone(),
-            badgered_to: to.iter().map(|v| <semver::Version>::clone(v)).collect(),
+            badgered_to: to.iter().cloned().map(<semver::Version>::clone).collect(),
             when: chrono::Utc::now(),
         };
 
--- a/crates/templates/src/template.rs
+++ b/crates/templates/src/template.rs
@@ -646,11 +646,14 @@ fn validate_v1_manifest(raw: &RawTemplateManifestV1) -> anyhow::Result<()> {
 mod test {
     use super::*;
 
-    struct TempFile(tempfile::TempDir, PathBuf);
+    struct TempFile {
+        _temp_dir: tempfile::TempDir,
+        path: PathBuf,
+    }
 
     impl TempFile {
         fn path(&self) -> PathBuf {
-            self.1.clone()
+            self.path.clone()
         }
     }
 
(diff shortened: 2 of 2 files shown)


crates/templates/src/template.rs
```diff
@@ -646,6 +646,7 @@ fn validate_v1_manifest(raw: &RawTemplateManifestV1) -> anyhow::Result<()> {
 mod test {
     use super::*;
 
+    #[allow(dead_code)]
```
vdice, line 649: This one I'm unsure about.  This was added to address the following error:
```
error: field `0` is never read
   --> crates/templates/src/template.rs:649:21
    |
649 |     struct TempFile(tempfile::TempDir, PathBuf);
    |            -------- ^^^^^^^^^^^^^^^^^
    |            |
    |            field in this struct
    |
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`
help: consider changing the field to be of unit type to suppress this warning while preserving the field numbering, or remove the field
    |
649 |     struct TempFile((), PathBuf);
    |                     ~~

error: could not compile `spin-templates` (lib test) due to 1 previous error
```

However, if I either update as suggested or remove the TempDir field, one of the test starts failing... perhaps due to the temp dir being dropped from memory before the assertion is made?  Better way to address this?

lann: Might be able to replace with e.g.
```rust
struct TempFile {
  path: PathBuf,
  _temp_dir: tempfile::TempDir,
}
```

vdice: Thanks @lann!  Did the trick.

fibonacci1729 (approved):

crates/plugins/src/badger/store.rs
```diff
@@ -86,7 +86,7 @@ impl BadgerRecordManager {
         let new = BadgerRecord {
             name: name.to_owned(),
             badgered_from: from.clone(),
-            badgered_to: to.iter().map(|v| <semver::Version>::clone(v)).collect(),
+            badgered_to: to.iter().cloned().map(<semver::Version>::clone).collect(),
```
rylev, line 89: Is the call to `map` still necessary?

vdice: Oh, it could prove not to be! Thanks.  I admit I just did the barebones lint-suggested thing to get it to not error.

fibonacci1729: I had the same thought yesterday but when i looked into it i started to believe that this is a false positive `map_clone` lint (=> https://github.com/rust-lang/rust-clippy/issues/12271). I.e. removing the `map` yields an type error on the resulting collection. 

A few things we could do but i'm not sure which is best:
1. Drop the `cloned` and attach `#[allow(clippy::map_clone)]` to this line
2. Keep as is and attach a note to remove the `cloned` whenever the false positive fix lands in rust proper (if it hasn't already and just requires a rust version bump 🤷‍♂️ ).

PR #2475, opened 2024-04-25
Taking a first crack at implementing metrics

calebschoepp:
This is my recommended approach for how we do metrics in Spin and supersedes #2469.


crates/telemetry/src/traces.rs
```diff
@@ -71,36 +63,3 @@ pub(crate) fn otel_tracing_layer(
...
-//
-// See https://github.com/open-telemetry/opentelemetry-rust/pull/950
-//
-// Work around that by checking OTEL_EXPORTER_OTLP_ENDPOINT, and setting
-// the endpoint url with the "/v1/traces" path ourselves. If the bug is
-// fixed in a later version, we can remove this code. But if we don't
-// remember to remove this, it won't do any harm either, as the crate will
-// just ignore the OTEL_EXPORTER_OTLP_ENDPOINT setting when the endpoint
-// is set directly with `with_endpoint`.
-fn fix_endpoint_bug() -> Option<String> {
```
calebschoepp, line 95: Turns out this wasn't doing anything.

crates/telemetry/src/env.rs
```diff
@@ -45,4 +65,18 @@ impl OtlpProtocol {
...
+    pub(crate) fn metrics_protocol_from_env() -> Self {
+        let metrics_protocol = std::env::var("OTEL_EXPORTER_OTLP_METRICS_PROTOCOL");
+        let general_protocol = std::env::var(OTEL_EXPORTER_OTLP_PROTOCOL);
+        let protocol = metrics_protocol.unwrap_or(general_protocol.unwrap_or_default());
+
+        match protocol.as_str() {
+            "grpc" => Self::Grpc,
+            "http/protobuf" => Self::HttpProtobuf,
+            "http/json" => Self::HttpJson,
+            _ => Self::HttpProtobuf,
```
lann, line 79: Recommend emitting a warning here to help people with typos.

crates/telemetry/src/lib.rs
```diff
@@ -16,9 +17,35 @@ pub use propagation::inject_trace_context;
...
+///
+/// ```no_run
+/// #[instrument(name = "span_name", Err(level = Level::INFO), fields(otel.name = "dynamically set name"))]
+/// fn func_you_want_to_trace() {}
+/// ```
+///
+/// Some notes on tracing:
+///
+/// - If you don't want the span to be collected by default emit it at a trace or debug level.
+/// - Make sure you `.instrument(tracing::Span::current())` any spawned tasks so the span context is propagated.
```
lann, line 35:
```suggestion
/// - Make sure you `.in_current_span()` any spawned tasks so the span context is propagated.
```

crates/telemetry/src/lib.rs
```diff
@@ -16,9 +17,35 @@ pub use propagation::inject_trace_context;
...
+/// - Use the otel.name attribute to dynamically set the span name.
+/// - Use the Err argument to have instrument automatically handle errors.
```
lann, line 37:
```suggestion
/// - Use the err argument to have instrument automatically handle errors.
```

crates/telemetry/src/lib.rs
```diff
@@ -16,9 +17,35 @@ pub use propagation::inject_trace_context;
...
+/// [Layer] emits [tracing] events to stderr, another sends spans to an OTel collector, and another
+/// sends metrics to an OTel collector.
 ///
-/// Configuration is pulled from the environment.
+/// Configuration for the OTel layers is pulled from the environment.
+///
+/// Examples of emitting traces from Spin:
+///
+/// ```no_run
+/// #[instrument(name = "span_name", Err(level = Level::INFO), fields(otel.name = "dynamically set name"))]
```
lann, line 28: Unless I'm mistaken...

```suggestion
/// #[instrument(name = "span_name", err(level = Level::INFO), fields(otel.name = "dynamically set name"))]
```

lann: Also, maybe leave off `level`? We probably _typically_ want errors emitted at error level.

```suggestion
/// #[instrument(name = "span_name", err, fields(otel.name = "dynamically set name"))]
```

crates/telemetry/src/lib.rs
```diff
@@ -16,9 +17,35 @@ pub use propagation::inject_trace_context;
...
+/// - Make sure you `.instrument(tracing::Span::current())` any spawned tasks so the span context is propagated.
+/// - Use the otel.name attribute to dynamically set the span name.
+/// - Use the Err argument to have instrument automatically handle errors.
+///
+/// Examples of emitting metrics from Spin:
+///
+/// ```no_run
+/// tracing::trace!(monotonic_counter.metric_foo = 1, attribute = "value");
+/// tracing::trace!(counter.metric_bar = -1, attribute = "value");
+/// tracing::trace!(histogram.another_metric = 1.1, attribute = "value");
```
lann, line 44: It might be good to add simple wrapper macros for these, e.g.:

```suggestion
/// spin_telemetry::monotonic_counter!(metric_foo = 1, attribute = "value");
/// spin_telemetry::counter!(metric_bar = -1, attribute = "value");
/// spin_telemetry::histogram!(another_metric = 1.1, attribute = "value");
```

- Slightly less magical / surprising
- Gives a place to change our mind about the implementation later.

calebschoepp: Yup I like that idea.

crates/telemetry/src/lib.rs
calebschoepp: FWIW I've been doing `level = Level::INFO` throughout the whole codebase already. My reasoning has been if we let it emit the event at the default ERROR level then it ends up getting logged to the `fmt_layer` which I figured we don't want. Just b/c some function deep in the guts of Spin is failing doesn't mean we want to log an error to the console.

If you think we should stop following this pattern then I'm happy to change the recommendation here.

lann: I guess it depends on the function 🤷

crates/telemetry/src/lib.rs
```diff
@@ -16,9 +17,35 @@ pub use propagation::inject_trace_context;
...
+///
+/// - If you don't want the span to be collected by default emit it at a trace or debug level.
+/// - Make sure you `.in_current_span()` any spawned tasks so the span context is propagated.
+/// - Use the otel.name attribute to dynamically set the span name.
+/// - Use the err argument to have instrument automatically handle errors.
+///
+/// Examples of emitting metrics from Spin:
+///
+/// ```no_run
+/// tracing::trace!(monotonic_counter.metric_foo = 1, attribute = "value");
```
itowlson, line 42: It might be useful to give more concrete examples and/or explain them.  It seems like the number is the "amount" of metric to be recorded (e.g. request count, egress bytes, etc.), is that right?  And `attribute` is actually the name under which `value` is to be associated with this amount?  E.g. it could be a `url = ...` to count requests to a given URL, or `app_name = ...` to count requests to a given app?

itowlson (commented):
I left a comment but it would be good to see a few more instances of how this would be used - it's hard to get a sense for it based on one occurrence!  I understand you don't want to invest too much in applying it when the core is not yet agreed though!

crates/telemetry/src/lib.rs
lann: > it could be a url = ... to count requests to a given URL, or app_name = ... to count requests to a given app?

Yep; "attribute" is opentelemetry parlance, sometimes called "tag" in other systems.

crates/telemetry/src/env.rs
```diff
@@ -1,18 +1,40 @@
...
-    ];
-    ENABLING_VARS
+pub(crate) fn otel_metrics_enabled() -> bool {
+    otel_enabled(&[
+        OTEL_EXPORTER_OTLP_ENDPOINT,
+        OTEL_EXPORTER_OTLP_METRICS_ENDPOINT,
+    ]) && !otel_sdk_disabled()
+}
+
+fn otel_enabled(enabling_vars: &[&str]) -> bool {
```
lann, line 36: Micro-nit: I would name this function by what it does, e.g. `any_vars_set` or something.

crates/telemetry/src/env.rs
```diff
@@ -34,15 +56,41 @@ pub(crate) enum OtlpProtocol {
 impl OtlpProtocol {
     /// Returns the protocol to be used for exporting traces as defined by the environment.
     pub(crate) fn traces_protocol_from_env() -> Self {
-        let trace_protocol = std::env::var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL");
-        let general_protocol = std::env::var("OTEL_EXPORTER_OTLP_PROTOCOL");
-        let protocol = trace_protocol.unwrap_or(general_protocol.unwrap_or_default());
+        Self::protocol_from_env(
+            std::env::var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL"),
```
lann, line 60: Micro-nit: I would make this a const too for consistency.

lann (approved):
Feel free to ignore micro-nits

PR #2479, opened 2024-04-30
Run a Wasm file as an application without a manifest

### LINKED ISSUE #2477: Proposal: "Manifest-less" Single Component Apps
Some simple apps don't require any meaningful configuration in their manifests beyond a single input component, e.g.:

- No (non-trivial) trigger config (`route = "/..."`)
- No WASI config: `files`, `environment` (could be `WASI-Virt`ed)
- No Spin-specific component config: `allowed_outbound_hosts`, `key_value_stores`, etc.

`spin up` could synthesize a manifest (or locked app, directly) for this sort of "manifest-less app", giving a similar/somewhat-compatible experience to `wasmtime serve`.

In addition to providing a nice lightweight path to running this sort of simple app, this could serve as a base for future features:

- Generate a `spin.toml`, either interactively or with TODO fill-in-the-blanks for required config
- Interactively virtualize dependencies

This would also allow us to support the proposed "[WASI OCI Design](https://docs.google.com/document/d/1bZLjDpcG22PruvSUxGH-DNL906GEV0fYIbHQH7eNzqs/edit)"

itowlson:
Fixes #2477.  Or at least it explores what a solution might look like - there are definitely other paths to consider.

For example, this goes the route of constructing an in-memory `spin.toml`, because the route from `spin.toml` to locked app to trigger is well-exercised.  But we could bypass that and construct a lockfile directly - a lot of `spin.toml` concerns are not (yet) relevant to manifestless applications.

Or we could start planning more extensively for Future Things - this POC doesn't really provide any support in terms of the OCI proposal.

One thing we discussed internally was whether to put this on `spin up` or on a new command such as `spin serve` (by analogy with `wasmtime serve`).  For the sake of this POC I put it on `spin up` and inferred "bare file or manifest" based on the file extension.  This is not intended as a proposal, just a thing to dangle the actual loader code off.

Anyway hopefully provides us with something to argue about _grin_

### DIFF
--- a/crates/loader/src/lib.rs
+++ b/crates/loader/src/lib.rs
@@ -40,6 +40,14 @@ pub async fn from_file(
     loader.load_file(path).await
 }
 
+/// Load a Spin locked app from a standalone Wasm file.
+pub async fn from_wasm_file(wasm_path: impl AsRef<Path>) -> Result<LockedApp> {
+    let app_root = std::env::current_dir()?;
+    let manifest = single_file_manifest(wasm_path)?;
+    let loader = LocalLoader::new(&app_root, FilesMountStrategy::Direct, None).await?;
+    loader.load_manifest(manifest).await
+}
+
 /// The strategy to use for mounting WASI files into a guest.
 #[derive(Debug)]
 pub enum FilesMountStrategy {
@@ -49,4 +57,37 @@ pub enum FilesMountStrategy {
     /// supports mounting full directories; mounting single files, glob
     /// patterns, and `exclude_files` are not supported.
     Direct,
+}
+
+fn single_file_manifest(
+    wasm_path: impl AsRef<Path>,
+) -> anyhow::Result<spin_manifest::schema::v2::AppManifest> {
+    use serde::Deserialize;
+
+    let wasm_path_str = wasm_path
+        .as_ref()
+        .to_str()
+        .context("Failed to stringise Wasm file path")?
+        .to_owned();
+    let app_name = wasm_path
+        .as_ref()
+        .file_stem()
+        .and_then(|s| s.to_str())
+        .unwrap_or("wasm-file")
+        .to_owned();
+
+    let manifest = toml::toml!(
+        spin_manifest_version = 2
+
+        [application]
+        name = app_name
+
+        [[trigger.http]]
+        route = "/..."
+        component = { source = wasm_path_str }
+    );
+
+    let manifest = spin_manifest::schema::v2::AppManifest::deserialize(manifest)?;
+
+    Ok(manifest)
 }
--- a/src/commands/up/app_source.rs
+++ b/src/commands/up/app_source.rs
@@ -13,6 +13,7 @@ use spin_manifest::schema::v2::AppManifest;
 pub enum AppSource {
     File(PathBuf),
     OciRegistry(String),
+    BareWasm(PathBuf),
     Unresolvable(String),
     None,
 }
@@ -31,7 +32,13 @@ impl AppSource {
 
     pub fn infer_file_source(path: impl Into<PathBuf>) -> Self {
         match spin_common::paths::resolve_manifest_file_path(path.into()) {
-            Ok(file) => Self::File(file),
+            Ok(file) => {
+                if is_wasm_file(&file) {
+                    Self::BareWasm(file)
+                } else {
+                    Self::File(file)
+                }
+            }
             Err(e) => Self::Unresolvable(e.to_string()),
         }
     }
@@ -58,11 +65,17 @@ impl AppSource {
     }
 }
 
+fn is_wasm_file(path: &Path) -> bool {
+    let extn = path.extension().and_then(std::ffi::OsStr::to_str);
+    extn.is_some_and(|e| e == "wasm" || e == "wat")
+}
+
 impl std::fmt::Display for AppSource {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::File(path) => write!(f, "local app {}", quoted_path(path)),
             Self::OciRegistry(reference) => write!(f, "remote app {reference:?}"),
+            Self::BareWasm(path) => write!(f, "Wasm file {}", quoted_path(path)),
             Self::Unresolvable(s) => write!(f, "unknown app source: {s:?}"),
             Self::None => write!(f, "<no source>"),
         }
@@ -77,6 +90,9 @@ pub enum ResolvedAppSource {
         manifest_path: PathBuf,
         manifest: AppManifest,
     },
+    BareWasm {
+        wasm_path: PathBuf,
+    },
     OciRegistry {
         locked_app: LockedApp,
     },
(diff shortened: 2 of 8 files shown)


src/commands/up/app_source.rs
```diff
@@ -31,7 +32,13 @@ impl AppSource {
 
     pub fn infer_file_source(path: impl Into<PathBuf>) -> Self {
         match spin_common::paths::resolve_manifest_file_path(path.into()) {
-            Ok(file) => Self::File(file),
+            Ok(file) => {
+                if file.extension().and_then(std::ffi::OsStr::to_str) == Some("wasm") {
```
fibonacci1729, line 36: Might be nice to support `*.wat` as well. If we want to be really precise we can use `wasmparser` to peek the special bytes to determine if the file is Wasm.

lann: > we can use `wasmparser` to peek the special bytes

...or `bytes.starts_with(b"\x00asm")` :slightly_smiling_face:

fibonacci1729: This is probably the best thing to use: https://github.com/bytecodealliance/wasm-tools/blob/755abb691d9d401dad542b8f238d19e6f3624e16/crates/wat/src/lib.rs#L228

fibonacci1729: > > we can use `wasmparser` to peek the special bytes
> 
> ...or `bytes.starts_with(b"\x00asm")` 🙂

...or https://github.com/bytecodealliance/wasm-tools/blob/755abb691d9d401dad542b8f238d19e6f3624e16/crates/wat/src/lib.rs#L228 😸

lann: I SUPPOSE

itowlson: This is absolutely world class bikeshedding, keep it up

radu-matei: To add to the bikeshedding — do we actually expect users to try this with `.wat`?

fibonacci1729: Most tools in this space accept either wat or wasm, so yes.

lann: I mean no, I don't think we expect users to use `.wat`, but also it isn't that hard to support apart from this slight inconvenience in detecting file type.

itowlson:
Folks, do we have any thoughts on whether to take this forward in its current form?  (Ignoring the detail of how we identify Wasm files.)

lann:
I remain in favor of my proposal :slightly_smiling_face:

tschneidereit:
IIUC this would not provide a way to do any IO, because that would require having anything allow-listed, right? If so, I think that's a perfectly valid starting point, but it might make sense to see if we could at least provide slightly better error messages? Such as telling folks that if they want to do outbound http, they have to use a manifest?

itowlson:
@lann Which proposal do you mean?

lann:
> @lann Which proposal do you mean?

https://github.com/fermyon/spin/issues/2477

itowlson:
@tschneidereit That would require either threading "did the app come from a loose Wasm file" all the way through Spin, or majorly reworking our error reporting architecture.

This would be a case where the Target Worlds proposal would be helpful, because we can give manifestless apps an implicit target world that doesn't include outbound HTTP, KV stores, etc. etc.  Target world validation would, I'm pretty sure, happen at a time and level where we are still threading the origin info through, so it would be easy to customise the world validation failure to talk about supplying a manifest.

If that's correct, then we could either punt this until after Target Worlds, or go ahead with this accepting that the errors will be suboptimal for now, with a view to sorting them out in the TW timeframe.

tschneidereit:
@itowlson that makes sense, and I agree that if getting to better error messages right now is too hard, then that's okay because as you say we have other things in flight that should get us there :)

itowlson:
Next discussion topic is do we put this on `spin up` or create a new `spin serve / spin run / spin something-else` command (to avoid any confusion likely to arise from flags being likely different across the manifestful and manifestless apps).

lann:
I think `spin up` is fine. "Manifestless" really means "we create a secret manifest for you", right?

src/commands/up/app_source.rs
itowlson: For now I have checked for the `.wasm` and `.wat` extensions, and isolated the "is it a Wasm file" logic into a function that we can evolve with magic numbers and wasmparsers and other such wild excesses if and when people feel strongly about it.  And by "we" I mean you Brian

lann (approved):
A few non-blocking suggestions; looks great!

crates/loader/src/lib.rs
```diff
@@ -40,6 +40,14 @@ pub async fn from_file(
     loader.load_file(path).await
 }
 
+/// Load a Spin locked app from a standalone Wasm file.
+pub async fn from_wasm_file(wasm_path: impl AsRef<Path>) -> Result<LockedApp> {
+    let app_root = std::env::current_dir()?;
```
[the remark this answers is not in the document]
lann: Actually, I believe you already have `working_dir` where this function is called; could probably just pass that through.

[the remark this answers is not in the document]
itowlson: This is the directory relative to which to resolve the Wasm file reference (and, perhaps, in future, other paths).  If we use `working_dir` then we have to copy the Wasm file first (before copying it again), and any file-related errors will show unexpected locations.

The alternative is for `app_root` to be the directory containing the Wasm file, and to override `wasm_path` to be only the name of the Wasm file (so that the relative reference resolves correctly), but I'm not sure that's going to be helpful with hypothetical future features where the user will expect paths entered on the CLI to be resolved relative to the current directory.  Of course it's hard to know because we haven't defined those features yet!

We could also do a more invasive change that threads multiple "app roots" through the loader, or redefines the app root, or absolutises everything earlier.  Or we can have a separate loader path for much longer than this PR does it, e.g. doing any copies needed then jumping straight to building the lockfile instead of leaning on the existing manifestful path.

src/commands/up/app_source.rs
```diff
@@ -13,6 +13,7 @@ use spin_manifest::schema::v2::AppManifest;
 pub enum AppSource {
     File(PathBuf),
     OciRegistry(String),
+    BareWasm(PathBuf),
```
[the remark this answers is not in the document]
itowlson: Yes, I had wondered about this.  I'm concerned it would make the docs a bit confusing, given that `--from-file` (in the current proposal) also accepts Wasm files (because my lawyers inform me that Wasm files are files within the meaning of the Act).  I suppose we could deprecate `--from-file` and define a new `--from-manifest` flag that forces the file to be interpreted as TOML, but this feels excessive.  Although maybe I'll feel differently if I understand what #2529 is about.

crates/loader/src/lib.rs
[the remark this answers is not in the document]
lann: Sounds complicated. I retract my feelings.

src/commands/up/app_source.rs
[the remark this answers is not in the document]
lann: Maybe it would be better to instead have an orthogonal option to differentiate between manifestful and manifestless sources. Anyway, that ought to be backward compatible with this so definitely not a blocker here.

PR #2483, opened 2024-05-02
Add example for Azure Key Vault application variable provider

ThorstenHans:
This PR adds an example to illustrate how one could use the Azure Key Vault application variable provider. 

As with the pre-existing Vault application variable provider, we will point to this example from the docs.


rylev (changes requested):
Our policy around examples has historically been "the more, the merrier". However, as time has gone by, we've found examples do add additional work to ensure that they stay up to date. 

I'm not opposed to merging this as is, but I'm not convinced it adds enough value to justify the maintenance and CI time costs. Would this perhaps be better located in docs somewhere? Do we want to perhaps maintain a separate repo of examples that don't need to be built every time we run CI?

cc @radu-matei @lann

ThorstenHans:
I'm fine with moving the sample to another repository. However, I don't think it's worth resisting in a dedicated repository on our GH org. 

What about moving this - and the vault sample - to [enterprise-architectures-and-patterns ](https://github.com/fermyon/enterprise-architectures-and-patterns) and refactor docs correspondingly

ThorstenHans:
I've moved the sample, so I'll close this one

Issue #2447, opened 2024-04-11
Support wasi-keyvalue


endocrimes:
wasi-keyvalue is relatively complete as an interface and the core interface (https://github.com/WebAssembly/wasi-keyvalue/blob/main/wit/store.wit) is almost identical to spin-kv (https://github.com/fermyon/spin/blob/main/wit/key-value.wit) with the primary difference being the `exists` function.

We should add support for wasi-keyvalue to get broader ecosystem compatibility - and the ability to benefit from the extended interfaces over time.

itowlson:
@endocrimes Can you expand on the difference you identified around the `exists` function please?  They look the same to me but I'm not terribly au fait with the `wasi-keyvalue` side.

(The main difference I did run into was around `list-keys` accepting a `cursor` for paging.)

endocrimes:
Past me forgot to link to a revision, but that got added since I last looked 😅 - now there's no meaningful difference 🎉

mikkelhegn:
@itowlson - We talked about this in today's Spin Up meeting. There is a release going out early next week. If there is an opporutnity to get this in, it's possible. If not, we can move it to post release in the board.

itowlson:
@mikkelhegn There is a PR out.  The first step towards a release would be to get feedback on that, specifically on the critical question of how to manage the Spin world versioning.  I'll ping the reviewers.

lann:
> the critical question of how to manage the Spin world versioning

Notably, this is basically the entire focus of WASI 0.2.1 and afaik we still haven't even completely figured out versioning for that upstream.

itowlson:
@mikkelhegn given ^^ I think it is fair to say this will not be part of 2.6

devigned:
Hey folks 👋, I'm happy to volunteer to implement the rest of the WASI key/value interface (atomics and batching) to help provide the basis for a second provider implementation for the interfaces, which will enable it to graduate to phase 3. Would the project be interested in the contribution?

itowlson:
@devigned The project would be absolutely delighted - thank you!  I've just rebased #2486 onto latest main - I don't know whether it's more convenient for you to PR onto that or to take a new branch from it and create a new PR - whatever suits you.

[...]

PR #2486, opened 2024-05-03
Implement wasi-keyvalue

itowlson:
An exploratory PR for exploring, apropos of #2447.

### DIFF
--- a/wit/deps/keyvalue-2024-05-03/world.wit
+++ b/wit/deps/keyvalue-2024-05-03/world.wit
@@ -1,0 +1,26 @@
+package wasi:keyvalue@0.2.0-draft;
+
+/// The `wasi:keyvalue/imports` world provides common APIs for interacting with key-value stores.
+/// Components targeting this world will be able to do:
+/// 
+/// 1. CRUD (create, read, update, delete) operations on key-value stores.
+/// 2. Atomic `increment` and CAS (compare-and-swap) operations.
+/// 3. Batch operations that can reduce the number of round trips to the network.
+world imports {
+	/// The `store` capability allows the component to perform eventually consistent operations on
+	/// the key-value store.
+	import store;
+
+	/// The `atomic` capability allows the component to perform atomic / `increment` and CAS
+	/// (compare-and-swap) operations.
+	import atomics;
+
+	/// The `batch` capability allows the component to perform eventually consistent batch
+	/// operations that can reduce the number of round trips to the network.
+	import batch;
+}
+
+world watch-service {
+	include imports;
+	export watcher;
+}
--- a/crates/factor-key-value/src/host.rs
+++ b/crates/factor-key-value/src/host.rs
@@ -2,6 +2,7 @@ use anyhow::{Context, Result};
 use spin_core::{async_trait, wasmtime::component::Resource};
 use spin_resource_table::Table;
 use spin_world::v2::key_value;
+use spin_world::wasi::keyvalue as wasi_keyvalue;
 use std::{collections::HashSet, sync::Arc};
 use tracing::{instrument, Level};
 
@@ -55,12 +56,21 @@ impl KeyValueDispatch {
         }
     }
 
-    pub fn get_store(&self, store: Resource<key_value::Store>) -> anyhow::Result<&Arc<dyn Store>> {
+    pub fn get_store<T: 'static>(&self, store: Resource<T>) -> anyhow::Result<&Arc<dyn Store>> {
         self.stores.get(store.rep()).context("invalid store")
     }
 
     pub fn allowed_stores(&self) -> &HashSet<String> {
         &self.allowed_stores
+    }
+
+    pub fn get_store_wasi<T: 'static>(
+        &self,
+        store: Resource<T>,
+    ) -> Result<&Arc<dyn Store>, wasi_keyvalue::store::Error> {
+        self.stores
+            .get(store.rep())
+            .ok_or(wasi_keyvalue::store::Error::NoSuchStore)
     }
 }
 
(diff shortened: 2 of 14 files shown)


itowlson:
I had to add the `wasi-keyvalue` import to the old world to get the runtime test to compile.  I assume in practice that's not what we'd aim to do though, we'd instead create a new world version and expose it only through that?

itowlson:
Do folks have any insights on how to move this forward?  It seems to work but I am not quite sure what do about the worlds stuff:

1. Should this create a rev of the Spin world?
2. The runtime tests appear use an old world: is there plumbing to let them use a current world?

itowlson:
@dicej @lann There is a request to move this forward for the impending release or the one after.  Are you the right folks to review, or should I pester someone else?

dicej:
Yup, I can take a look tomorrow morning.

dicej:
@kate-goldenring might also be interested, given her recent work on the specification.

crates/key-value/src/lib.rs
```diff
@@ -138,6 +139,100 @@ impl key_value::HostStore for KeyValueDispatch {
...
+        Error::Other(msg) => wasi_keyvalue::store::Error::Other(msg),
+    }
+}
+
+#[async_trait]
+impl wasi_keyvalue::store::Host for KeyValueDispatch {
+    async fn open(
+        &mut self,
+        identifier: String,
+    ) -> anyhow::Result<Result<Resource<wasi_keyvalue::store::Bucket>, wasi_keyvalue::store::Error>>
```
rylev, line 156: You can use the [`trappable_error_type`](https://github.com/fermyon/spin/blob/13a133fded7c48fa5d1931207619379e737e9376/crates/world/src/lib.rs#L14-L28) option to make the signatures here a bit easier to work with. See the `llm` implementation for an example use.

crates/key-value/src/lib.rs
```diff
@@ -138,6 +139,100 @@ impl key_value::HostStore for KeyValueDispatch {
...
+    }
+
+    async fn list_keys(
+        &mut self,
+        self_: Resource<Bucket>,
+        cursor: Option<u64>,
+    ) -> anyhow::Result<Result<wasi_keyvalue::store::KeyResponse, wasi_keyvalue::store::Error>>
+    {
+        if cursor.unwrap_or_default() != 0 {
+            anyhow::bail!("list_keys: cursor not supported");
```
rylev, line 222: Is the plan to eventually support the cursor?

wit/world.wit
```diff
@@ -16,6 +16,7 @@ world http-trigger-rc20231018 {
 world platform {
   include wasi:cli/imports@0.2.0;
   import wasi:http/outgoing-handler@0.2.0;
+  import wasi:keyvalue/store@0.2.0-draft;
```
rylev, line 19: We definitely need to decide how we want to handle introducing new imports. This will break any Spin runtime that does not have `wasi:keyvalue` support.

dicej (commented):
LGTM overall -- thanks for doing this!

I expect we'll need to implement cursors before we can really claim to support `wasi:keyvalue/store`, and that may be the most interesting part of this project given the variety of backends we now support (SQLite, Redis, and Azure).

wit/world.wit
lann: It seems like we should be bumping the WIT package version here right? This is still a little scary as we haven't even done it with WASI 0.2.1 yet...

itowlson: Yes, this was one of my questions: https://github.com/fermyon/spin/pull/2486#issuecomment-2094965982.  Do we need a separate conversation about how to do this?

wit/world.wit
```diff
@@ -30,6 +31,7 @@ world platform {
 world platform-rc20231018 {
   include wasi:cli/reactor@0.2.0-rc-2023-10-18;
   import wasi:http/outgoing-handler@0.2.0-rc-2023-10-18;
+  import wasi:keyvalue/store@0.2.0-draft;
```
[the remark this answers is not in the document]
itowlson: I agree!  The trouble is that the tests all (currently) use this world, so if we want tests for `wasi-keyvalue` then we will need to rev the test world, and I'm not sure of the implications of that.  https://github.com/fermyon/spin/pull/2486#issuecomment-2094965982

wit/world.wit
lann: > Do we need a separate conversation about how to do this?

Probably? I'm not sure that _anyone_ knows how best to handle this yet.

crates/key-value/src/lib.rs
itowlson: "Plan" is a strong word but, yes, it would be desirable to do on back-ends that support it.  For this initial implementation I stuck to using the existing KV implementation which unfortunately do not support cursors.

wit/deps/keyvalue-2024-05-03/world.wit
```diff
@@ -0,0 +1,26 @@
+package wasi:keyvalue@0.2.0-draft;
```
kate-goldenring, line 1: Are we planning to remove `-draft`?

kate-goldenring (commented):
LGTM. @itowlson should we mark this as ready to review?

itowlson:
@kate-goldenring I was planning to leave it as draft until we have a decision on WIT versioning. I can mark ready for review and label "do not merge" if that would be better?

wit/deps/keyvalue-2024-05-03/world.wit
itowlson: Is there a non-draft WIT yet?  The HEAD in the wasi-keyvalue GitHub still has `draft` at time of writing, but if there's a non-draft world then yeah we should definitely use that instead.

https://github.com/WebAssembly/wasi-keyvalue/blob/219ea3612a53f1bf5b2d137551b22d0268fd3c58/wit/world.wit#L1

kate-goldenring: I see. I missed that in the versioning of the upstream WIT. Thanks

kate-goldenring:
> I was planning to leave it as draft until we have a decision on WIT versioning. I can mark ready for review and label "do not merge" if that would be better?

@itowlson leaving as is is fine. Thanks for clarifying the decision criteria.

crates/key-value/src/lib.rs
itowlson: Thanks @rylev.  I gave this a try and am not sure how I feel about.  I've left it in a separate commit and would value your thoughts...

radu-matei:
Just to validate an assumption I had reading this PR: this PR adds support for the WASI KV proposal — specifically, it allows guests to use the proposal to interact with stores — and stores are still managed through the manifest `key_value_stores` field. Is that accurate?

Thanks!

itowlson:
@radu-matei Yes.  The KV stores are the exact same KV stores as you would get if you used the Spin APIs, specified and configured in exactly the same way.

radu-matei:
Is the intention to update this PR and have WASI KV be part of the Spin world?

itowlson:
@radu-matei That is the intention, but we don't yet know how to version the Spin world (at least I don't, and from previous comments this seems to be a general problem, unless the community has figured something out since June).

itowlson:
Rebased onto world3 (#2887).  @devigned if you are working on this then please note the proposed new world and new file layout.  (The last commit is how wasi-keyvalue fits in - the earlier commits are tracking the world3 proposal.)  Also please note that these are not yet final...!

PR #2491, opened 2024-05-06
Fix 1.78 lints

### LINKED ISSUE #2424: "Unknown lint: `clippy::blocks_in_conditions`" warning when clippying Spin source
The `clippy::blocks_in_conditions` lint causes an "unknown lint" warning on Rust 1.75.  Looking at the linked issue it was maybe introduced in 1.76?  The MSRV of Spin is 1.74.  Are we happy to rev that to 1.76, or are we okay to ignore the warnings (since the only reason we mention the lint is a bug that will hopefully get fixed soon)?

itowlson:
I tried installing 1.78 to see if it fixed #2424 and no it doesn't but hoo boy is it mad about the precious `clone()`s.

It also started giving deprecation warnings on `./cargo/config`, so I renamed that too.  The warning says the new name works from Cargo 1.38 so should not be a problem for us on 1.74.


itowlson:
The changes in the LLM host component were because it was giving "never constructed" lints on the `NoopLlmEngine`, and if I featured that out I had to feature out a bunch of other things that only it used, and it seemed easier to put it into a mod and feature that out.  If that fels too invasive, we could simply suppress the lint and not touch the rest of it - happy either way.

fibonacci1729 (approved):

lann:
It looks like the `clone_from` stuff is Controversial and is going to be punted out of the default checks: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202023-05-02/near/355185136

itowlson:
@lann I'd be delighted to revert it if there is consensus to do so.  I'm not a fan of the readability.

PR #2500, opened 2024-05-08
fix(docs): update release process to remove `spin-macro` consideration

kate-goldenring:
The Rust SDK is no longer a part of this repo

### DIFF
--- a/docs/content/release-process.md
+++ b/docs/content/release-process.md
@@ -34,7 +34,7 @@ To cut a major / minor release of Spin, you will need to do the following:
     git push origin v2.0.0
     ```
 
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

kate-goldenring:
I am confused why these actions are running given that we have `docs/**` as a `paths-ignore`: https://github.com/fermyon/spin/blob/b3db535c9edb72278d4db3a201f0ed214e561354/.github/workflows/build.yml#L12-L13

fibonacci1729:
Not entirely sure about that either but this has historically been the case and will require an admin to merge I believe. Do you have those perms @kate-goldenring? Otherwise I think @lann can merge since Vaughn is out.

PR #2501, opened 2024-05-09
ref(telemetry): Provide the abililty to turn off the tracing-log compat layer in tracing-subscriber dep of telemetry

calebschoepp:


calebschoepp:
I tested to make sure that `tracing::log::*` is still being picked up in the main Spin binary. This is ready for review.

lann:
What is the goal with this compat layer?

radu-matei:
ref https://github.com/spinkube/containerd-shim-spin/pull/112

calebschoepp:
> What is the goal with this compat layer?

@lann the goal of the compat layer is to keep the [`tracing-log` feature](https://docs.rs/tracing-subscriber/0.3.18/tracing_subscriber/#optional-dependencies) turned on when Spin is compiled as a standalone binary. This is so that we can still pick up any `log::*` emitted by upstream libraries. For example wasmparser uses `log::*` and not `tracing::*`.

lann (approved):

PR #2508, opened 2024-05-14
Dynamically detect support for Wasmtime's pooling allocator

### LINKED ISSUE #2119: On arm64, I get "cannot allocate memory (os error 12)" on executing spin up command
This is regarding the quickstart from the docs.
It runs on my X86 laptops just fine. But on arm64 platforms such as Raspberry Pi 3 (1 gb ram) or Arm based EC2 instances(with 2 GB ram), `spin build` works without any issues but on `spin up` I get the following error:

```
Error: failed to create stack pool mapping
caused by:
0: mmap failed to allocate 0x7d3e8888 bytes
1: Cannot allocate memory (os error 12)
```

Am I low on ram ? do I need more for Arm ?
Image attached. 
Screenshot (20)

### LINKED ISSUE #2343: Disable pooling allocator automatically in environments where it will fail
There are (at least) two environments where the Wasmtime pooling allocator is known to fail with Spin's default settings:

QEMU: https://github.com/fermyon/spin/issues/1785
Systems with <2GB (?) RAM:  https://github.com/fermyon/spin/issues/2119

Under `spin up` we should try to detect either of these environments and disable pooling (with a message) to improve UX.

alexcrichton:
This commit is intended to address #2119 and mirror bytecodealliance/wasmtime#8610. The base problem is that some systems are configured with smaller amounts of virtual memory than other systems, for example some aarch64 and riscv64 systems are shown to have only 39 bits of virtual address space rather than the 48 by default on x86_64. This means that the pooling allocator can't be initialized on these platforms since it needs more virtual memory than that.

This changes Spin to dynamically choosing whether to use the pooling allocator. It's still used by default in Wasmtime but a dynamic probe is performed to determine whether it's going to work first. While here I also added an env var to control this behavior for an escape hatch if that's needed in the future too.

Closes #2119
Closes #2343

### DIFF
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -175,7 +175,8 @@ fn use_pooling_allocator_by_default() -> bool {
         match std::env::var("SPIN_WASMTIME_POOLING") {
             Ok(s) if s == "1" => return true,
             Ok(s) if s == "0" => return false,
-            _ => {}
+            Ok(s) => panic!("SPIN_WASMTIME_POOLING={s} not supported, only 1/0 supported"),
+            Err(_) => {}
         }
 
         // If the env var isn't set then perform the dynamic runtime probe
@@ -186,12 +187,18 @@ fn use_pooling_allocator_by_default() -> bool {
         match wasmtime::Engine::new(&config) {
             Ok(engine) => {
                 let mut store = wasmtime::Store::new(&engine, ());
-                // NB: the maximum size is in wasm pages to take out the 16-bits of
-                // wasm page size here from the maximum size.
+                // NB: the maximum size is in wasm pages so take out the 16-bits
+                // of wasm page size here from the maximum size.
                 let ty = wasmtime::MemoryType::new64(0, Some(1 << (BITS_TO_TEST - 16)));
                 wasmtime::Memory::new(&mut store, ty).is_ok()
             }
-            Err(_) => false,
+            Err(_) => {
+                tracing::debug!(
+                    "unable to create an engine to test the pooling \
+                     allocator, disabling pooling allocation"
+                );
+                false
+            }
         }
     })
 }
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -15,6 +15,7 @@ mod store;
 pub mod wasi_2023_10_18;
 pub mod wasi_2023_11_10;
 
+use std::sync::OnceLock;
 use std::{path::PathBuf, time::Duration};
 
 use anyhow::Result;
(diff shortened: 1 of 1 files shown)


lann (approved):

lann:
Also fixes #2343

crates/core/src/lib.rs
```diff
@@ -142,6 +145,57 @@ impl Default for Config {
...
+fn use_pooling_allocator_by_default() -> bool {
+    static USE_POOLING: OnceLock<bool> = OnceLock::new();
+    const BITS_TO_TEST: u32 = 42;
+
+    *USE_POOLING.get_or_init(|| {
+        // Enable manual control through env vars as an escape hatch
+        match std::env::var("SPIN_WASMTIME_POOLING") {
+            Ok(s) if s == "1" => return true,
+            Ok(s) if s == "0" => return false,
+            _ => {}
```
rylev, line 178: I think we should respect `SPIN_WASMTIME_POOLING` if it's set no matter what. If it's set to "0" or "false", we can turn it off, and otherwise enable it.

crates/core/src/lib.rs
```diff
@@ -142,6 +145,57 @@ impl Default for Config {
...
+        // If the env var isn't set then perform the dynamic runtime probe
+        let mut config = wasmtime::Config::new();
+        config.wasm_memory64(true);
+        config.static_memory_maximum_size(1 << BITS_TO_TEST);
+
+        match wasmtime::Engine::new(&config) {
+            Ok(engine) => {
+                let mut store = wasmtime::Store::new(&engine, ());
+                // NB: the maximum size is in wasm pages to take out the 16-bits of
+                // wasm page size here from the maximum size.
```
rylev, line 190: I think there is a typo here?
```suggestion
                // NB: the maximum size is in wasm pages so take out the 16-bits of
                // wasm page size here from the maximum size.
```

crates/core/src/lib.rs
```diff
@@ -142,6 +145,57 @@ impl Default for Config {
...
+                let ty = wasmtime::MemoryType::new64(0, Some(1 << (BITS_TO_TEST - 16)));
+                wasmtime::Memory::new(&mut store, ty).is_ok()
+            }
+            Err(_) => false,
+        }
```
rylev, line 195: I think we'll want to log the response here perhaps at `debug` level.

crates/core/src/lib.rs
alexcrichton: Sure yeah, I opted to panic for non 0/1 values to avoid making this too too sprawling (as it's probably not going to be used that much anyway)

rylev (approved):

PR #2511, opened 2024-05-14
feat(telemetry): Add a compatibility layer that emits app logs as tracing events.

calebschoepp:
## Problem being solved

Users want to be able to emit o11y from the guest itself, but WASI Observe is quite far away.

## Solution
If Spin app logs were emitted as events on the existing traces we could get users 80% of the way there.

## Notes
This required some jank in `stdio.rs`. @fibonacci1729 and I felt the cleanest way to handle the two cases in the writer was to represent it with an enum. We can't keep using the inherit functionality b/c we want to emit this tracing event either way. We also felt that using options throughout to avoid writing to the file in the inherit case wasn't very clean.

### DIFF
--- a/crates/telemetry/src/env.rs
+++ b/crates/telemetry/src/env.rs
@@ -8,6 +8,7 @@ use opentelemetry_otlp::{
 const OTEL_SDK_DISABLED: &str = "OTEL_SDK_DISABLED";
 const OTEL_EXPORTER_OTLP_TRACES_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_TRACES_PROTOCOL";
 const OTEL_EXPORTER_OTLP_METRICS_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_METRICS_PROTOCOL";
+const SPIN_DISABLE_LOG_TO_TRACING: &str = "SPIN_DISABLE_LOG_TO_TRACING";
 
 /// Returns a boolean indicating if the OTEL tracing layer should be enabled.
 ///
@@ -35,6 +36,15 @@ pub(crate) fn otel_metrics_enabled() -> bool {
         OTEL_EXPORTER_OTLP_ENDPOINT,
         OTEL_EXPORTER_OTLP_METRICS_ENDPOINT,
     ]) && !otel_sdk_disabled()
+}
+
+/// Returns a boolean indicating if the compatibility layer that emits tracing events from
+/// applications logs should be disabled.
+///
+/// It is considered disabled if the environment variable `SPIN_DISABLED_LOG_TO_TRACING` is set and not
+/// empty. By default the features is enabled.
+pub(crate) fn spin_disable_log_to_tracing() -> bool {
+    any_vars_set(&[SPIN_DISABLE_LOG_TO_TRACING])
 }
 
 fn any_vars_set(enabling_vars: &[&str]) -> bool {
--- a/crates/telemetry/src/lib.rs
+++ b/crates/telemetry/src/lib.rs
@@ -7,6 +7,7 @@ use tracing_subscriber::{fmt, prelude::*, registry, EnvFilter, Layer};
 
 pub mod detector;
 mod env;
+pub mod log;
 pub mod metrics;
 mod propagation;
 mod traces;
@@ -51,9 +52,15 @@ pub fn init(spin_version: String) -> anyhow::Result<ShutdownGuard> {
         .with_writer(std::io::stderr)
         .with_ansi(std::io::stderr().is_terminal())
         .with_filter(
+            // Filter directives explained here https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
             EnvFilter::from_default_env()
+                // Wasmtime is too noisy
                 .add_directive("wasmtime_wasi_http=warn".parse()?)
-                .add_directive("watchexec=off".parse()?),
+                // Watchexec is too noisy
+                .add_directive("watchexec=off".parse()?)
+                // We don't want to duplicate application logs
+                .add_directive("[{app_log}]=off".parse()?)
+                .add_directive("[{app_log_non_utf8}]=off".parse()?),
         );
 
     // Even if metrics or tracing aren't enabled we're okay to turn on the global error handler
--- a/crates/telemetry/src/log.rs
+++ b/crates/telemetry/src/log.rs
@@ -1,0 +1,25 @@
+use std::{ascii::escape_default, sync::OnceLock};
+
+use crate::env;
+
+/// Takes a Spin application log and emits it as a tracing event.
+///
+/// This acts as a compatibility layer to easily get Spin app logs as events in our OTel traces.
+pub fn app_log_to_tracing_event(buf: &[u8]) {
+    static CELL: OnceLock<bool> = OnceLock::new();
+    if *CELL.get_or_init(env::spin_disable_log_to_tracing) {
+        return;
+    }
+
+    if let Ok(s) = std::str::from_utf8(buf) {
+        tracing::info!(app_log = s);
+    } else {
+        tracing::info!(
+            app_log_non_utf8 = buf
+                .iter()
+                .take(50)
+                .map(|&x| escape_default(x).to_string())
+                .collect::<String>()
+        );
+    }
+}
--- a/crates/trigger/src/stdio.rs
+++ b/crates/trigger/src/stdio.rs
@@ -56,13 +56,19 @@ impl StdioLoggingTriggerHooks {
         &self,
         component_id: &str,
         log_suffix: &str,
-        log_dir: &Path,
+        log_dir: Option<&Path>,
     ) -> Result<ComponentStdioWriter> {
         let sanitized_component_id = sanitize_filename::sanitize(component_id);
-        let log_path = log_dir.join(format!("{sanitized_component_id}_{log_suffix}.txt"));
+        let log_path = log_dir
+            .map(|log_dir| log_dir.join(format!("{sanitized_component_id}_{log_suffix}.txt",)));
+        let log_path = log_path.as_deref();
+
         let follow = self.follow_components.should_follow(component_id);
-        ComponentStdioWriter::new(&log_path, follow)
-            .with_context(|| format!("Failed to open log file {}", quoted_path(&log_path)))
+        match log_path {
+            Some(log_path) => ComponentStdioWriter::new_forward(log_path, follow)
+                .with_context(|| format!("Failed to open log file {}", quoted_path(log_path))),
+            None => ComponentStdioWriter::new_inherit(),
+        }
     }
 
     fn validate_follows(&self, app: &spin_app::App) -> anyhow::Result<()> {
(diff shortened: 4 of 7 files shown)


crates/telemetry/src/env.rs
```diff
@@ -8,6 +8,7 @@ use opentelemetry_otlp::{
 const OTEL_SDK_DISABLED: &str = "OTEL_SDK_DISABLED";
 const OTEL_EXPORTER_OTLP_TRACES_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_TRACES_PROTOCOL";
 const OTEL_EXPORTER_OTLP_METRICS_PROTOCOL: &str = "OTEL_EXPORTER_OTLP_METRICS_PROTOCOL";
+const APP_LOG_TO_TRACING_EVENT_DISABLED: &str = "APP_LOG_TO_TRACING_EVENT_DISABLED";
```
itowlson, line 11: Is this name forced on us by OTel or do we have leeway to change it?  (I ask because if it's Spin-specific then we might want to give it a `SPIN_` prefix.)

crates/telemetry/src/env.rs
```diff
@@ -37,6 +38,15 @@ pub(crate) fn otel_metrics_enabled() -> bool {
     ]) && !otel_sdk_disabled()
 }
 
+/// Returns a boolean indicating if the compatibility layer that emits tracing events from
+/// applications logs should be disabled.
```
itowlson, line 42: This feels like it is going to end with me staring at an `if` expression with my head turned 90 degrees going "so if it's _not_ disabled then it's enabled and so in the else clause it's _not_ not disabled which wait" like some unnerving game of Boolean RoboRally.  Consider flipping this to be the positive case?  I guess it depends on usage though.

crates/telemetry/src/log.rs
```diff
@@ -0,0 +1,18 @@
+use std::sync::OnceLock;
+
+use crate::env;
+
+/// Takes a Spin application log and emits it as a tracing event. This acts as a compatibility layer
+/// to easily get Spin app logs as events∆1 in our OTel traces.
+pub fn app_log_to_tracing_event(buf: &[u8]) {
+    static CELL: OnceLock<bool> = OnceLock::new();
```
itowlson, line 8: Feels like you could move this into the `app_log_to_tracing_event_disabled` function and avoid all the CELL clutter here?

crates/telemetry/src/log.rs
```diff
@@ -0,0 +1,18 @@
...
+    if *CELL.get_or_init(env::app_log_to_tracing_event_disabled) {
+        return;
+    }
+
+    if let Ok(s) = std::str::from_utf8(buf) {
+        tracing::info!(app_log = s);
+    } else {
+        tracing::info!(app_log = "Application log: <non-utf8 data>");
```
itowlson, line 16: Worth including any info that might help a log reader figure out what was going on, e.g. buffer length, initial 4 bytes, etc.?

crates/trigger/src/stdio.rs
```diff
@@ -56,13 +56,23 @@ impl StdioLoggingTriggerHooks {
         &self,
         component_id: &str,
         log_suffix: &str,
-        log_dir: &Path,
+        log_dir: Option<&Path>,
     ) -> Result<ComponentStdioWriter> {
         let sanitized_component_id = sanitize_filename::sanitize(component_id);
-        let log_path = log_dir.join(format!("{sanitized_component_id}_{log_suffix}.txt"));
+        let mut log_path: Option<PathBuf> = None;
+        if let Some(log_dir) = log_dir {
```
itowlson, line 63: Is this fancy talk for `Option::map`?

crates/trigger/src/stdio.rs
```diff
@@ -56,13 +56,23 @@ impl StdioLoggingTriggerHooks {
...
+            log_path = Some(log_dir.join(format!(
+                "{sanitized_component_id}_{log_suffix}.txt",
+                sanitized_component_id = sanitized_component_id,
```
itowlson, line 66: Why are this and the following line needed when they weren't before?

crates/trigger/src/stdio.rs
```diff
@@ -56,13 +56,23 @@ impl StdioLoggingTriggerHooks {
...
+                log_suffix = log_suffix
+            )))
+        };
+        let log_path = log_path.as_deref();
+
         let follow = self.follow_components.should_follow(component_id);
-        ComponentStdioWriter::new(&log_path, follow)
-            .with_context(|| format!("Failed to open log file {}", quoted_path(&log_path)))
+        // It is safe to unwrap the log_path b/c new will only fail if the file can't be opened in which case log_path is Some
```
itowlson, line 73: This makes some hefty assumptions about the implementation of `ComponentStdioWriter::new`.  I'm sceptical we can ensure it remains infallible in the `None` case as other people maintain it.

A possibly safer (but wordier!) approach is to split `ComponentStdioWriter::new` into two functions e.g. `ComponentStdioWriter::log_path(Path) -> (fallible)` and `ComponentStdioWriter::stdio() -> (infallible)`.  (Maybe `new_forward` and `new_inherit` already are those functions.)

crates/trigger/src/stdio.rs
```diff
@@ -167,38 +203,56 @@ impl AsyncWrite for ComponentStdioWriter {
...
-            match &this.state {
-                ComponentStdioWriterState::File => {
+            match &mut this.inner {
+                ComponentStdioWriterInner::Inherit => {
                     let written = futures::ready!(
-                        std::pin::Pin::new(&mut this.async_file).poll_write(cx, buf)
+                        std::pin::Pin::new(&mut tokio::io::stderr()).poll_write(cx, buf)
                     );
                     let written = match written {
                         Ok(e) => e,
```
itowlson, line 214: I realise this is existing code but `e`?

crates/trigger/src/stdio.rs
```diff
@@ -142,20 +161,37 @@ enum ComponentStdioWriterState {
 }
 
 impl ComponentStdioWriter {
-    pub fn new(log_path: &Path, follow: bool) -> anyhow::Result<Self> {
+    pub fn new(log_path: Option<&Path>, follow: bool) -> anyhow::Result<Self> {
+        match log_path {
+            Some(log_path) => Self::new_forward(log_path, follow),
+            None => Self::new_inherit(),
```
itowlson, line 167: Is it intentional that this does not consider `follow`?  I'm a bit fuzzy on what these two things are doing, but if `Inherit` is sending app logs to stdio then I feel like it should respect `follow`.

crates/trigger/src/stdio.rs
```diff
@@ -223,32 +283,57 @@ impl AsyncWrite for ComponentStdioWriter {
...
+
+        match &mut this.inner {
+            ComponentStdioWriterInner::Inherit => {
                 std::pin::Pin::new(&mut tokio::io::stderr()).poll_flush(cx)
             }
+            ComponentStdioWriterInner::Forward {
+                async_file, state, ..
+            } => match state {
+                ComponentStdioWriterState::File => std::pin::Pin::new(async_file).poll_shutdown(cx),
+                ComponentStdioWriterState::Follow(_) => {
```
itowlson, line 295: This is at least the second function where `Inherit` and `Forward { state = Follow }` have the same behaviour.  Should we have a helper function on `ComponentStdioWriterInner` that returns "use stderr or this handle"?  Or the destination stream so we can really unify the cases?

itowlson (commented):
The "I have no idea what I'm doing" dog has entered the chat

crates/telemetry/src/env.rs
calebschoepp: Purely from the land of my own creativity 😄. `SPIN_` prefix is a great point. Also open to workshopping the name. I struggled to achieve brevity and clarity on this complex of a concept.

crates/telemetry/src/env.rs
calebschoepp: I originally had it as the positive case b/c I too wanted to avoid Boolean RoboRally. However, where I was actually using it was `!enabled` so I figured I should just avoid double negatives.

crates/telemetry/src/log.rs
calebschoepp: Fair. I've been treating `env.rs` as purely a retrieval layer and so in my mind the caching makes sense to live here. I could be convinced otherwise.

crates/telemetry/src/log.rs
calebschoepp: Great question. IMO that probably isn't valuable to a user, but let me know if you feel otherwise. Below is an idea of what it looks like for a user.

image

crates/trigger/src/stdio.rs
fibonacci1729: Correct! Before we just called `inherit_stdio` on the store builder.

crates/telemetry/src/env.rs
itowlson: Yes, sorry, I had meant to come back to this comment after seeing where it was used.  I no longer have strong feelings about this.  (Feelings, maybe.  But not strong ones!)

crates/telemetry/src/log.rs
itowlson: The reason I was thinking about this was because it might give the reader some clue as to why they are getting this junk in the precious logs.  Like is something dumping a PNG to stdout for its own mystic reasons.  But it's an edge case for sure, and could always be retrofitted if telemetry fans demand it.

lann: I agree that we should emit _something_ about the actual data, maybe something like:

`app_log_non_utf8 = <ascii-escaped bytes>`

or for long data:

`app_log_non_utf8 = "<first 50 bytes, ascii-escaped> [truncated; total size = 1000]"`

https://doc.rust-lang.org/std/ascii/fn.escape_default.html

crates/trigger/src/stdio.rs
lann: imo combinators shouldn't be mixed with mutation. :shrug:

lann: Actually I'm skeptical of combinators any time you throw away the result.

crates/trigger/src/stdio.rs
lann: This code always writes exactly 2.71828... bytes.

crates/telemetry/src/env.rs
lann: `SPIN_DISABLE_LOG_TO_TRACING`?

crates/trigger/src/stdio.rs
itowlson: I meant it seemed like this whole stanza seemed equivalent to `let log_path = log_dir.map(...)` (sorry I should probably have put the comment on line 62) but if there is mutation of something with larger scope going on then ignore me I completely agree.

crates/trigger/src/stdio.rs
```diff
@@ -112,27 +122,36 @@ impl TriggerHooks for StdioLoggingTriggerHooks {
...
 pub struct ComponentStdioWriter {
-    sync_file: std::fs::File,
-    async_file: tokio::fs::File,
-    state: ComponentStdioWriterState,
-    follow: bool,
+    inner: ComponentStdioWriterInner,
+}
+
+enum ComponentStdioWriterInner {
+    /// TODO
```
rylev, line 146: nit: don't forget

crates/trigger/src/stdio.rs
lann: Oh maybe you are correct. I was confused by `log_path` vs `log_dir` above.

crates/trigger/src/stdio.rs
calebschoepp: Resolving this b/c follow is only relevant when we're trying to forward the content to a file

crates/trigger/src/stdio.rs
calebschoepp: I'm not really following your suggestions, could you please elaborate?

itowlson: Don't worry about it, it just seemed as if we might be in a situation where two of the three states had the same behaviour and I was just wondering if we could extract that shared behaviour.

crates/telemetry/src/lib.rs
```diff
@@ -51,9 +52,15 @@ pub fn init(spin_version: String) -> anyhow::Result<ShutdownGuard> {
         .with_writer(std::io::stderr)
         .with_ansi(std::io::stderr().is_terminal())
         .with_filter(
+            // https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html#directives
```
itowlson, line 55: I don't know what this comment is trying to tell me.  If this is just the docs link for the struct you're using then is is needed?  If it's aiming to refer to some behaviour or explanation of the directives then maybe say what that reference is about?

crates/trigger/src/stdio.rs
```diff
@@ -56,13 +56,19 @@ impl StdioLoggingTriggerHooks {
...
         component_id: &str,
         log_suffix: &str,
-        log_dir: &Path,
+        log_dir: Option<&Path>,
     ) -> Result<ComponentStdioWriter> {
         let sanitized_component_id = sanitize_filename::sanitize(component_id);
-        let log_path = log_dir.join(format!("{sanitized_component_id}_{log_suffix}.txt"));
+        let log_path = log_dir
+            .map(|log_dir| log_dir.join(format!("{sanitized_component_id}_{log_suffix}.txt",)));
+        let log_path = log_path.as_deref();
```
itowlson, line 64: Can this be tacked onto the end of the `map` (`let log_path = log_dir.map(...).as_deref()`) or do you need the intermediate value to satisfy the savage borrow checker?

itowlson (approved):
LGTM!  A couple of suggestions but they are very minor - feel free to ignore.  Caveat is that I don't feel very qualfied to review this so it would be good for someone more knowledgeable to also take a look!

crates/trigger/src/stdio.rs
lann: Not the borrow checker per se; you can't return a reference to an owned value (the result of `join`) from a closure.

The borrow on this line is only allowed because of the special "temporary lifetime extension" rule which applies to the local scope of a `let`.

lann (approved):

PR #2512, opened 2024-05-14
Update Wasmtime to 20.0.2

alexcrichton:
This commit updates the Wasmtime version used by Spin from 18.0.4 to 20.0.2. No motivation beyond keeping things up-to-date. This involved a number of minor WASI refactorings such as moving some items around and some refactorings around the `wasmtime-wasi-http` traits-and-types. Notably the chained handler was switched from `anyhow::Result` to `wasmtime_wasi_http::HttpResult` which helps make more explicit which errors are traps and which are normal error codes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  RUST_VERSION: 1.74
+  RUST_VERSION: 1.75
 
 jobs:
   dependency-review:
--- a/crates/componentize/src/abi_conformance/mod.rs
+++ b/crates/componentize/src/abi_conformance/mod.rs
@@ -36,9 +36,7 @@ use wasmtime::{
     component::{Component, InstancePre, Linker},
     Engine, Store,
 };
-use wasmtime_wasi::preview2::{
-    pipe::MemoryOutputPipe, ResourceTable, WasiCtx, WasiCtxBuilder, WasiView,
-};
+use wasmtime_wasi::{pipe::MemoryOutputPipe, ResourceTable, WasiCtx, WasiCtxBuilder, WasiView};
 
 pub use test_key_value::KeyValueReport;
 pub use test_llm::LlmReport;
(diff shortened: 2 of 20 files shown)


.github/workflows/build.yml
```diff
@@ -21,7 +21,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  RUST_VERSION: 1.74
+  RUST_VERSION: 1.75
```
itowlson, line 24: We should bump the MSRV in Cargo.toml if this is now needed.

crates/componentize/src/abi_conformance/mod.rs
```diff
@@ -189,38 +187,39 @@ pub async fn test(
...
 }
 
 pub(crate) fn create_store_with_context_and_wasi(
     engine: &Engine,
     test_config: TestConfig,
     context_builder: impl FnOnce(&mut Context),
-    wasi_builder: impl FnOnce(WasiCtxBuilder) -> WasiCtxBuilder,
-) -> Store<Context> {
+    wasi_builder: impl FnOnce(&mut WasiCtxBuilder) -> Result<()>,
+) -> Result<Store<Context>> {
```
rylev, line 213: It seems we unwrap this result in the caller. Perhaps we should just keep this infallible and unwrap (or better yet `expect`) in the closures?

crates/trigger-http/src/lib.rs
```diff
@@ -715,7 +723,9 @@ impl OutboundWasiHttpHandler for HttpRuntimeData {
                 "self".into()
             };
             eprintln!("To allow requests, add 'allowed_outbound_hosts = [\"{}\"]' to the manifest component section.", host);
-            anyhow::bail!("destination-not-allowed (error 1)")
+            return Err(HttpError::trap(anyhow::anyhow!(
```
rylev, line 726: I realize we were trapping before, but I do wonder if we should be returning `HTTP-request-denied` instead.

rylev (approved):
LGTM - some small observations, but nothing that should block merging this.

PR #2513, opened 2024-05-15
feat(oci/client.rs): add registry_from_input helper

vdice:
Adds a helper to hold parsing logic around converting the user-supplied registry server into a form compatible with the underlying oci library.  This helper also includes logic to avoid the failure mode for `spin registry login docker.io` which would previously error out with an unhelpful message.*

*Note, although the underlying oci-distribution library has a similar [helper](https://github.com/krustlet/oci-distribution/blob/main/src/reference.rs#L104-L114), it uses a qualified Reference (eg `docker.io/ubuntu:latest`) which the provided server string is not, as it only represents the registry portion.  Hence using our own logic here.

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -559,21 +559,16 @@ impl Client {
         username: impl AsRef<str>,
         password: impl AsRef<str>,
     ) -> Result<()> {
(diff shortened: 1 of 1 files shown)


itowlson (approved):
I haven't tested but it looks fine to me!

PR #2514, opened 2024-05-15
feat(telemetry): Send logs to OTel collector powered by a OpentelemetryTracingBridge layer in the tracing subscriber

calebschoepp:
This is the first way we could implement OTel logs.

Here's an example of what the logs would look like in Loki.

image

I like that this method sticks with the pattern of using tracing subscriber layers to do our work. But, I really don't like that all of this Rust tracing info pollutes the logs. Ideally the log only contains the body of the message that the user printed in their Spin app.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4507,6 +4507,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "opentelemetry-appender-tracing"
+version = "0.3.0"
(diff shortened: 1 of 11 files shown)


calebschoepp:
I haven't posted a clean implemenation of our alternative yet. But, here is what the logs would look like:

image

They're a lot cleaner b/c we are just directly setting the body to be the spin app log. None of this weirdness of the actual log being an attribute called `app_log`.

PR #2516, opened 2024-05-16
feat(telemetry): Send logs to OTel collector directly using OTel libraries

calebschoepp:
image

This is the other way we could do this compared to #2514 .


calebschoepp:
I've convinced myself that this is the right approach compared to #2514. @lann / @itowlson this is ready for a review from an implementation perspective.

crates/telemetry/src/logs.rs
```diff
@@ -0,0 +1,116 @@
...
+                )
+                .with_attribute("app_log_non_utf8", true)
+                .build(),
+        );
+    }
+}
+
+/// Takes a Spin application log and emits it as a tracing event. This acts as a compatibility layer
+/// to easily get Spin app logs as events in our OTel traces.
+fn app_log_to_tracing_event(buf: &[u8]) {
```
lann, line 57: Deja vu...didn't you just add this?

calebschoepp: It's appearing as new b/c I renamed `log.rs` to `logs.rs`

endocrimes: It looks like you didn't commit the removal of `log.rs`, which will be part of why it's showing as new 😅

calebschoepp: Ohhhhh, I'll fix that.

lann (approved):

crates/telemetry/src/logs.rs
```diff
@@ -0,0 +1,116 @@
...
+    app_log_to_otel(buf);
+    app_log_to_tracing_event(buf);
+}
+
+/// Forward the app log to OTel.
+fn app_log_to_otel(buf: &[u8]) {
+    if !otel_logs_enabled() {
+        return;
+    }
+
```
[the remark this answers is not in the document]
endocrimes: keeping the return in `app_log_to_otel` is nicer for consistency here imo (app_log_to_tracing_event does this too)

PR #2518, opened 2024-05-20
Spin Factors SIP

lann:
[Rendered](https://github.com/lann/spin/blob/factors-sip/docs/content/sips/0xx-spin-factors.md)

### DIFF
--- a/docs/content/sips/0xx-spin-factors.md
+++ b/docs/content/sips/0xx-spin-factors.md
@@ -1,0 +1,111 @@
+title = "SIP 0XX - Spin Factors"
+template = "main"
+date = "2024-05-20T12:00:00Z"
+
+---
+
+Summary: Refactor Spin Runtime Functionality with "Factors"
+
+Owner: lann.martin@fermyon.com
+
+Created: May 20, 2024
+
+## Background
+
+Spin 1.0 shipped with a mechanism for extending the runtime environment with
+loosely-coupled host functionality called "host components". As Spin has
+evolved, more and more runtime functionality has fallen outside of the scope of
+this mechanism, leading to the code for these features spreading out over the
+Spin codebase. This not only makes it hard to read and understand these features
+but also raises the bar for unfamiliar developers trying to add new features.
+
+Separately, the introduction of the SpinKube project has made it clear that
+there will be (at least) two major embeddings of the Spin runtime environment:
+the Spin CLI (`spin up`) and SpinKube (`containerd-shim-spin`). In order to
+provide better integration into the Kubernetes ecosystem, some runtime feature
+implementations will diverge between these embeddings, making the loose coupling
+of features even more important.
+
+## Proposal
+
+The basic inversion of control approach used by `HostComponent`s will be
+redesigned and expanded into a new system called Spin Factors, where independent
+feature sets are organized into individual "factors". Some of the goals of this
+new system:
+
+- Expand the set integration points that a factor can use to hook into the
+  lifecycle of an application request. These will be driven by feature
+  requirements and may include e.g.:
+  - Runtime startup
+  - Application initialization
+  - Runtime config parsing
+  - Component pre-instantiation
+  - Component instantiation
+  - Cleanup (post-execution)
+
+- Allow loosely-coupled dependencies between "factors" (features). In order to
+  simplify implementation and reasoning about these dependencies the dependency
+  graph must be acyclic.
+
+## Implementation Plan
+
+The overall implementation plan will be to:
+
+1. Introduce a new `spin-factors` crate containing the basic framework.
+
+2. Refactor existing Spin features (mostly existing `HostComponent`s) to use
+   this new framework.
+   - Framework features will be developed as needed to support this refactoring.
+
+3. Refactor `spin-core`, `spin-trigger`, etc. to use `spin-factors`.
+   - Depending on subjective evaluation at this point, possibly merge
+   `spin-core` and `spin-factors`.
+
+## Implementation Details
+
+Based on initial prototyping, the following Rust types represent the starting
+point for `spin-factors` (note that some type details are elided for clarity):
+
+```rust
+pub trait Factor {
+    // Runtime configuration
+    type RuntimeConfig;
+    // Application-wide state; may be reused between instances
+    type AppState;
+    // Builds instance state
+    type InstanceBuilder: FactorInstanceBuilder;
+
+    // Init is the runtime startup lifecycle hook.
+    //
+    // The `InitContext` type here gives the factor the ability to update global
+    // engine configuration, most notably the `Linker`. This takes the place of
+    // `HostComponent::add_to_linker`.
+    fn init(&mut self, ctx: InitContext) -> Result<()> {
+        Ok(())
+    }
+
+    // Validates application configuration and prepares AppState.
+    //
+    // Takes the place of `DynamicHostComponent::validate_app`.
+    fn configure_app(&self, ctx: ConfigureAppContext) -> Result<Self::AppState>;
+
+    // Prepares InstanceState. This can access other Factors' InstanceBuilders
+    // to implement inter-factor dependencies.
+    fn prepare(
+      &self,
+      ctx: PrepareContext,
+      builders: &mut InstanceBuilders,
+    ) --> Result<Self::InstanceBuilder>
+}
+
+pub trait FactorInstanceBuilder {
+  // Per-instance state, equivalent to `HostComponent::Data`.
+  type InstanceState;
+
+  // Builds the InstanceState.
+  //
+  // Takes the place of `HostComponent::build_data` and
+  // `DynamicHostComponent::update_data`.
+  fn build(self) -> Result<Self::InstanceState>;
+}
+```


docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,108 @@
...
+the Spin CLI (`spin up`) and SpinKube (`containerd-shim-spin`). In order to
+provide better integration into the Kubernetes ecosystem, some runtime feature
+implementations will diverge between these embeddings, making the loose coupling
+of features even more important.
+
+## Proposal
+
+The basic inversion of control approach used by `HostComponent`s will be
+redesigned and expanded into a new system called Spin Factors, where independent
+feature sets are organized into individual "factors". Some of the goals of this
```
rylev, line 33: I understand the term "factor" is intentionally vague, and we'll never really be able to come up with a concise word to describe what a factor is meant to represent. That being said, it would be massively useful if this SIP attempted to clarify what a factor is and perhaps more importantly what a factor is not.

It's hard to make a decision on this refactoring path if we're deciding on adopting a concept that isn't explicitly defined.

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,108 @@
...
+Separately, the introduction of the SpinKube project has made it clear that
+there will be (at least) two major embeddings of the Spin runtime environment:
+the Spin CLI (`spin up`) and SpinKube (`containerd-shim-spin`). In order to
+provide better integration into the Kubernetes ecosystem, some runtime feature
+implementations will diverge between these embeddings, making the loose coupling
+of features even more important.
+
+## Proposal
+
+The basic inversion of control approach used by `HostComponent`s will be
```
rylev, line 31: Nit: Some reading this who might otherwise be qualified to have an opinion on the nature of this refactoring might not know what a `HostComponent` is, and I think it would be helpful to define what they are here.

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,108 @@
...
+   `spin-core` and `spin-factors`.
+
+## Implementation Details
+
+Based on initial prototyping, the following Rust types represent the starting
+point for `spin-factors` (note that some type details are elided for clarity):
+
+```rust
+pub trait Factor {
+    // Builder represents a type that can prepare request state for the factor.
```
rylev, line 69: > prepare request state for the factor

I'm not sure what is meant by this. "request state" evokes http request state which I know is not meant here. I think perhaps being a bit more verbose and spelling out "request state" would be helpful here.

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,108 @@
...
+    type Builder: FactorBuilder<Self>;
+    // Data represents request state for the factor.
+    type Data;
```
rylev, line 72: If data === request state, why not call this `RequestState` which while vague is less generic than `Data` which might as well be called `Foo`.

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,108 @@
...
+
+    // Init is the runtime startup lifecycle hook.
+    //
+    // The `InitContext` type here gives the factor the ability to update global
+    // engine configuration, most notably the `Linker`. This takes the place of
+    // `HostComponent::add_to_linker`.
+    fn init(ctx: InitContext) -> Result<()> {
```
rylev, line 79: I would like to know more about `InitContext`. Could a rough definition of that be added to this code block?

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,108 @@
...
+
+    // Validate app is the application initialization hook.
+    //
+    // This takes the place of `DynamicHostComponent::validate_app`.
+    fn validate_app(&self, app: &App) -> Result<()> {
+        Ok(())
+    }
+}
+
+pub trait FactorBuilder<Factor> {
```
rylev, line 91: `FactorBuilder` makes it sound like this builds a `Factor` but in fact it builds a `Factor::Data` which is meant to represent "request state" - I think some adjustments to naming would be helpful.

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,108 @@
...
+    // of `DynamicHostComponent::update_data`.
+    fn prepare(ctx: PrepareContext) -> Result<Self>;
+
+    // Build is the component instantiation hook.
+    //
+    // This returns the request state for this factor. This takes the place of
+    // `HostComponent::build_data`.
+    fn build(self) -> Factor::Data;
+}
+```
```
rylev, line 108: I would really love to see a high level sketch of how this system would be used to implement an existing feature. It's hard to judge this architecture without seeing it in use.

docs/content/sips/0xx-spin-factors.md
lann: This aligns with Wasmtime's internal terminology: https://docs.rs/wasmtime/21.0.0/wasmtime/struct.Store.html#method.data

docs/content/sips/0xx-spin-factors.md
lann: `InitContext` contains "whatever is needed to serve the requirements of this lifecycle hook". The impl PR, adapted from [the prototype](https://github.com/lann/spin-factors), has [more detail](https://github.com/fermyon/spin/pull/2519/files#diff-3896b08a07b6a026f5d2aca19afdee76e166070b8402329656dd5cb3726c05bfR28-R59).

docs/content/sips/0xx-spin-factors.md
lann: The prototype PR includes a couple of examples, though they are (intentionally) more complex than average. The basic WASI factor is probably the easiest to digest: https://github.com/lann/spin-factors/blob/main/crates/wasi-factor/src/lib.rs

docs/content/sips/0xx-spin-factors.md
rylev: Hmm... by that logic we shouldn't call it a `Factor`, but rather a `Store`. I personally don't think adopting the wasmtime naming convention here helps. I consider myself fairly familiar with the wasmtime API, and I personally find aligning with that naming highly confusing.

lann: It's also what [`HostComponent` calls it](https://github.com/fermyon/spin/blob/e078f8d36e43d8ddc7f34c74d8c4d2b37550c9a6/crates/core/src/host_component.rs#L50), though that's not an argument either way in this context.

Happy to bikeshed, though "request state" is also a bad name; I've already changed this comment to reword.

rylev: `PerInstanceState` would be initial candidate into the bikeshedding arena.

lann: I did call it `InstanceState` in one prototype iteration but got sick of typing it over and over :sweat_smile:

docs/content/sips/0xx-spin-factors.md
rylev: I'm also curious about the choice of making `FactorBuilder` a generic type instead of an associated type. I would imagine that each implementation of `FactorBuilder` would have 1 and only 1 `Factor` it's associated with, no?

docs/content/sips/0xx-spin-factors.md
lann: Maybe even just `State` would be an incremental improvement?

rylev: `State` is an improvement, but I like the fact that `InstanceState` makes it clear how long the state lives for and what it associated with. `State` could sound like persisted state associated with the app, for example.

docs/content/sips/0xx-spin-factors.md
lann: I believe it was to support this: https://github.com/lann/spin-factors/blob/cef37d031bbda20e8c0894222be82b4e7d3a85e8/src/lib.rs#L187-L189

```rust
impl<T: Factor> FactorBuilder<T> for DefaultBuilder
where
    T::Data: Default,
```

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,129 @@
...
+- Allow loosely-coupled dependencies between "factors" (features). In order to
+  simplify implementation and reasoning about these dependencies the dependency
+  graph must be acyclic.
+
+## Implementation Plan
+
+The overall implementation plan will be to (in parallel):
+
+- Introduce a new `spin-factors` crate containing the basic framework.
+- Refactor existing Spin features (mostly existing `HostComponent`s) to use
```
calebschoepp, line 61: Will `spin-factors` and `HostComponents` live side by side during the migration, or is this a big bang migration?

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,129 @@
...
+   this new framework.
+   - Framework features will be developed as needed to support this refactoring.
+- Refactor `spin-core`, `spin-trigger`, etc. to use `spin-factors`.
```
calebschoepp, line 64: Wondering if this is still the intention for core and trigger

docs/content/sips/0xx-spin-factors.md
lann: Could go either way but unless there is a really compelling reason I'd prefer to go the big bang route.

docs/content/sips/0xx-spin-factors.md
lann: I haven't started on this part yet, so "yes, for now".

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,129 @@
...
+    // `Self::AppState` to be used in `prepare` below.
+    // 
+    // `ConfigureAppContext` gives access to:
+    // - The `spin_app::App`
+    // - This factors's `RuntimeConfig`
+    // - The `AppState` for any factors configured before this one
+    //
+    // These methods can also be used on their own (without `init` or `prepare`)
+    // to just validate app configuration for e.g. `spin doctor`.
+    fn configure_app(&self, ctx: ConfigureAppContext) -> Result<Self::AppState>;
```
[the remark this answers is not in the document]
rylev: Ah I see in the implementation that is (almost) like this and that's just been elided here. I think that's probably fine for this SIP.

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,129 @@
...
+
+Based on initial prototyping, the following Rust types represent the starting
+point for `spin-factors` (simplified from the actual code for clarity):
+
+```rust
+pub trait Factor {
+    // This provides a mechanism to configure this factor on a per-app basis
+    // based on "runtime config", as currently implemented by
+    // `spin_trigger::RuntimeConfig`.
+    type RuntimeConfig;
```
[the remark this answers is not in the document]
lann: Maybe? To me, "runtime config" implies "not part of the app package itself" where "app config" doe not, though at this point that might just be from staring at Spin code for too long...

docs/content/sips/0xx-spin-factors.md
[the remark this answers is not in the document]
lann: Yeah I left out a bunch of type noise here. I'll link to the source in a comment.

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,129 @@
...
+   this new framework.
+   - Framework features will be developed as needed to support this refactoring.
+- Refactor `spin-core`, `spin-trigger`, etc. to use `spin-factors`.
+   - Depending on subjective evaluation at this point, possibly merge
+   `spin-core` and `spin-factors`.
+
+## Implementation Details
+
+Based on initial prototyping, the following Rust types represent the starting
+point for `spin-factors` (simplified from the actual code for clarity):
```
lann, line 71: > simplified from the actual code for clarity

The rest of the owl: https://github.com/fermyon/spin/pull/2519/files#diff-deb13d726ac7a552e420923432af8ef603ea90b5f281de344c7a797c6b568ce3R10

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,129 @@
...
+## Proposal
+
+The basic inversion of control approach used by `HostComponent`s will be
+redesigned and expanded into a new system called Spin Factors, where independent
+feature sets are organized into individual "factors". A "factor" encapsulates a
+single logical Spin runtime feature into a reusable Rust crate, using the Spin
+Factors framework to "hook" into various points in the lifecycle of a Spin
+trigger's execution.
+
+Some of the goals of this new system:
```
kate-goldenring, line 40: Does this get us closer to being able to use RPC host components (rather than embedded Spin ones) too?

calebschoepp: @kate-goldenring what do you mean when you say RPC host components. I'm just out of the loop and curious 😄.

docs/content/sips/0xx-spin-factors.md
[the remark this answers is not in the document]
rylev: Interesting - is it important that a distinction is drawn here of where this config originates from. It seems to me that the high order piece of information we want to convey with the type name is "this is configuration that exists at the level of the app as opposed to the level of the full runtime or the wasm instance".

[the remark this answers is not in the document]
lann: > is it important that a distinction is drawn here of where this config originates from

As currently designed, yes: this is always deserialized from the data source that we call "runtime config" (`spin up --runtime-config-file`).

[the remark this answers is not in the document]
rylev: Hmm ok, I was under the impression we were trying not to tie this design 1 to 1 to having a runtime config file like mechanism (allowing for alternative ways to specify how apps should be configured dynamically), but I think I've raised enough of a point here to let you decide which path forward is best 😄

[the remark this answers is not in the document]
lann: > I was under the impression we were trying not to tie this design 1 to 1 to having a runtime config file like mechanism

The design is [a little more general](https://github.com/fermyon/spin/pull/2519/files#diff-62de7c3895d9c88dceb20fd4d8027d52eba37b078c050db5e52b5a8da832e614R24-R38) than "it has to be a file" but the concrete implementations we have today all roughly boil down to "buckets of JSON/TOML".

docs/content/sips/0xx-spin-factors.md
kate-goldenring: Host components that are not in the spin binary rather run separately and are used over RPC

calebschoepp: Now I too am curious what the answer to this question is haha

lann: For interfaces known at build time an rpc implementation could just be a normal factor. For interfaces known only at runtime it would require hooking into the (pre)instantiation step, which could be added as a factors phase or implemented outside of factors.

lann: So I think the answer to the original question is "no but it shouldn't make it harder either".

docs/content/sips/0xx-spin-factors.md
```diff
@@ -0,0 +1,129 @@
...
+
+- Introduce a new `spin-factors` crate containing the basic framework.
+- Refactor existing Spin features (mostly existing `HostComponent`s) to use
+   this new framework.
+   - Framework features will be developed as needed to support this refactoring.
+- Refactor `spin-core`, `spin-trigger`, etc. to use `spin-factors`.
+   - Depending on subjective evaluation at this point, possibly merge
+   `spin-core` and `spin-factors`.
+
+## Implementation Details
```
kate-goldenring, line 68: @lann how will factors receive runtime config information that is not part of a factor's runtime config but still applies to it's set up. Specifically, how should `state_dir` and `log_dir` be passed which are now set in [`RuntimeConfigOpts`](https://github.com/fermyon/spin/blob/66ede35186730c0ac7158d40046665273b8ef2a5/crates/trigger/src/runtime_config.rs#L254) alongside provider information.

fibonacci1729 (approved):

PR #2519, opened 2024-05-20
Spin Factors

lann:
ref #2518


crates/factor-outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        })
+    }
+}
+
+#[derive(Default)]
+pub struct AppConfig {
+    component_allowed_hosts: HashMap<String, Arc<[String]>>,
+}
+
+type AllowedHostsFuture = Shared<BoxFuture<'static, Arc<anyhow::Result<AllowedHostsConfig>>>>;
```
rylev, line 52: nit: I think I'd prefer for there to be a type alias `SharedFuture<T> = Shared<BoxFuture<'static, Arc<T>>>` so that the output of the future is still seen where it's used. I personally find completely opaque type aliases to be harder to read because I always end up looking at the definition to find out what type will actually be produced. 

For example:

```rust
pub struct InstancePreparer {
    allowed_hosts_future: SharedFuture<anyhow::Result<AllowedHostsConfig>>,
}

```

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        ctx: PrepareContext<OutboundNetworkingFactor>,
+        mut preparers: InstancePreparers<Factors>,
+    ) -> Result<Self> {
+        let hosts = ctx
+            .app_config()
+            .component_allowed_hosts
+            .get(ctx.app_component().id())
+            .cloned()
+            .context("missing component allowed hosts")?;
+        let resolver = preparers.get_mut::<VariablesFactor>()?.resolver().clone();
```
rylev, line 69: I think this would make the potential error message much clearer, but it would also make the code easier to understand:

```suggestion
        let resolver = preparers.get_mut::<VariablesFactor>().context("ADD HELPFUL MESSAGE HERE")?.resolver().clone();
```

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn new<Factors: SpinFactors>(
+        ctx: PrepareContext<OutboundNetworkingFactor>,
+        mut preparers: InstancePreparers<Factors>,
+    ) -> Result<Self> {
+        let hosts = ctx
+            .app_config()
+            .component_allowed_hosts
+            .get(ctx.app_component().id())
+            .cloned()
+            .context("missing component allowed hosts")?;
```
rylev, line 68: This error message is a bit weak. Perhaps:

```suggestion
            .with_context(|| format!("missing allowed outbound hosts configuration for component {}", ctx.app_component().id()))?;
```

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        })
+    }
+
+    fn prepare(self) -> Result<<OutboundNetworkingFactor as Factor>::InstanceState> {
+        Ok(())
+    }
+}
+
+impl InstancePreparer {
+    pub async fn resolve_allowed_hosts(&self) -> Arc<anyhow::Result<AllowedHostsConfig>> {
```
rylev, line 113: I think this is no longer used?

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+        .boxed()
+        .shared();
+        // let prepared_resolver = resolver.prepare().await?;
+        // let allowed_hosts = AllowedHostsConfig::parse(
+        //         .context("missing component allowed hosts")?,
+        //     &prepared_resolver,
+        // )?;
+
+        // Update Wasi socket allowed ports
+        let wasi_preparer = preparers.get_mut::<WasiFactor>()?;
```
rylev, line 84: So are we relying on convention that app configuration is usually handled in one particular factor and we rely on that factor reaching into other factor prepares and mutating them?

In other words, why doesn't the `WasiFactor` read from `allowed_outbound_hosts` itself?

crates/factor-wasi/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+        Ok(())
+    }
+}
+
+pub trait FilesMounter {
+    fn mount_files(&self, app_component: &AppComponent, ctx: MountFilesContext) -> Result<()>;
+}
+
+pub struct DummyFilesMounter;
```
rylev, line 76: I'm confused. Is this meant to be a temporary thing before this gets merged?

crates/factors/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+        unsafe {
+            let offset = Self::instance_preparer_offset::<T>().context("no such factor")?;
+            let ptr = preparers as *mut Self::InstancePreparers;
+            let opt = &mut *ptr.add(offset).cast::<Option<T::InstancePreparer>>();
+            Ok(opt.as_mut())
+        }
+    }
+}
+
+pub struct Getter<T, U> {
```
rylev, line 295: I think we might want to talk through some of implementation here. There's quite a bit of unsafe code here, and I'm not sure why.

crates/core/src/lib.rs
```diff
@@ -127,7 +126,7 @@ impl Default for Config {
                 // Nothing is lost from allowing the maximum size of memory for
                 // all instance as it's still limited through other the normal
                 // `StoreLimitsAsync` accounting method too.
-                .memory_pages(4 * GB / WASM_PAGE_SIZE)
+                .max_memory_size(4 * GB)
```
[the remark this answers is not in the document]
lann: This PR depends on https://github.com/bytecodealliance/wasmtime/pull/8732 which required upgrading to wasmtime "22" (main).

crates/factor-outbound-networking/src/lib.rs
lann: `SharedFuture` sounds the same as `Shared` (which is a `Future`). It could be...`SharedFutureResult`...I guess?

crates/factor-outbound-networking/src/lib.rs
lann: `get_mut` already has `.context("preparer not initialized")`, but also there are a few `anyhow::Result`s that I intend to replace with `Result<_, spin_factors::Error>` once the overall structure of things has settled down a bit.

crates/factor-outbound-networking/src/lib.rs
lann: > So are we relying on convention that app configuration is usually handled in one particular factor and we rely on that factor reaching into other factor prepares and mutating them?

In this case, yes. `allowed_outbound_hosts` supports variables, which means evaluating it requries async calls to variable provider(s) at some point. Currently spin does this eagerly at boot, which means dynamic variable changes wouldn't propagate to `allowed_outbound_hosts`. The `AllowedHostsFuture` stuff above is a first effort at resolving this issue; I expect to add some caching and/or concurrent sync in the future as well.

> In other words, why doesn't the WasiFactor read from allowed_outbound_hosts itself?

The stuff above require centralizing `allowed_outbound_hosts` resolution _somewhere_, and this is what made sense to me.

crates/factor-wasi/src/lib.rs
lann: Yes and probably for unit tests that don't care about files.

crates/factor-outbound-networking/src/lib.rs
lann: This will be used by the e.g. outbound-http factor.

crates/factor-wasi/src/lib.rs
lann: Oh also the actual FilesMounter would be defined by the runtime, though perhaps `spin up`'s mounter will live here as a standard impl.

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+use spin_factor_wasi::WasiFactor;
+use spin_factors::{
+    Factor, FactorInstancePreparer, InstancePreparers, PrepareContext, Result, SpinFactors,
+};
+use spin_outbound_networking::{AllowedHostsConfig, ALLOWED_HOSTS_KEY};
+
+pub struct OutboundNetworkingFactor;
+
+impl Factor for OutboundNetworkingFactor {
+    type AppConfig = AppConfig;
```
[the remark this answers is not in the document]
rylev: `AppState` might have my vote for the simple reason that it contrasts directly with `InstanceState`. i.e., this type represents state at the app level and can live for as long as the app lives as opposed to instance state which lives just for the single instance.

crates/factors/src/factor.rs
```diff
@@ -0,0 +1,152 @@
...
+use crate::{App, FactorInstancePreparer, Linker, ModuleLinker, RuntimeConfig, SpinFactors};
+
+pub trait Factor: Any + Sized {
+    /// Per-app configuration for this factor.
+    ///
+    /// See [`Factor::configure_app`].
+    type AppConfig: Default;
+
+    /// The [`FactorInstancePreparer`] for this factor.
+    type InstancePreparer: FactorInstancePreparer<Self>;
```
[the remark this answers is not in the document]
lann: Sketching out what that could look like:

```rust
trait Factor {
  ...

  type InstancePreparer: Default;
  fn create_preparer(&self, ...) -> Result<Self::InstancePreparer> {
    Ok(Default::default())
  }

  // Either:
  type InstanceState;
  fn prepare(&self, preparer: Self::InstancePreparer) -> Result<Self::InstanceState>;
  // Or:
  type InstanceState: Default;
  fn prepare(&self, preparer: Self::InstancePreparer) -> Result<Self::InstanceState> {
    Ok(Default::default())
  }
```

The annoying thing here is `prepare`; you either need to have no default impl _or_ you need to make `InstanceState: Default`, both of which may be annoying for different factors.

That said, I was already on the fence about how to do this so given your feedback I would probably now lean toward one of these annoying ways rather than the existing annoying way. :slightly_smiling_face:

crates/factor-wasi/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+        mut ctx: InitContext<Factors, Self>,
+    ) -> anyhow::Result<()> {
+        fn type_annotate<T, F>(f: F) -> F
+        where
+            F: Fn(&mut T) -> &mut dyn WasiView,
+        {
+            f
+        }
+        let get_data = ctx.get_data_fn();
+        let closure = type_annotate(move |data| get_data(data) as &mut dyn WasiView);
```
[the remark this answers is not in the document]
lann: This is due to upstream changes in bindgen and wasmtime-wasi: https://github.com/bytecodealliance/wasmtime/blob/fa84dd83ba0f5c45bf0174bb5f988c215d848cfe/crates/wasi/src/lib.rs#L378

crates/factors/src/spin_factors.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use anyhow::Context;
+
+use crate::Factor;
+
+// TODO(lann): Most of the unsafe shenanigans here probably aren't worth it;
+// consider replacing with e.g. `Any::downcast`.
+
+/// Implemented by `#[derive(SpinFactors)]`
+pub trait SpinFactors: Sized {
```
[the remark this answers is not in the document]
lann: `RuntimeFactors` would probably be fine; I just generally shy away from naming new things "runtime"...

crates/factors/src/runtime_config.rs
```diff
@@ -0,0 +1,94 @@
...
+        if self.unused_keys.is_empty() {
+            Ok(())
+        } else {
+            Err(self.unused_keys)
+        }
+    }
+}
+
+impl<Source: RuntimeConfigSource> RuntimeConfig for RuntimeConfigTracker<Source> {
+    fn get_config<T: DeserializeOwned>(
```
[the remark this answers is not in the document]
lann: Here is a thought:

```rust
trait Factor {
  type RuntimeConfig: FactorRuntimeConfig;
  ...
}

trait FactorRuntimeConfig: DeserializeOwned {
  const KEY: &'static str;
}

impl FactorRuntimeConfig for () {
  const KEY: &'static str = "<unused>";
}
```

crates/factors/src/runtime_config.rs
```diff
@@ -0,0 +1,94 @@
...
+    ) -> anyhow::Result<Option<T>>;
+}
+
+pub struct RuntimeConfigTracker<Source> {
+    source: Source,
+    used_keys: HashSet<String>,
+    unused_keys: HashSet<String>,
+}
+
+impl<Source: RuntimeConfigSource> RuntimeConfigTracker<Source> {
```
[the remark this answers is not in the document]
lann: I haven't really thought about this too much apart from trying to reserve `T` for `T: Factor`. :shrug:

crates/factors/src/factor.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn new(app: &'a App, app_configs: &'a Factors::AppConfigs) -> Self {
+        Self { app, app_configs }
+    }
+
+    pub fn app(&self) -> &App {
+        self.app
+    }
+
+    pub fn app_config<T: Factor>(&self) -> crate::Result<&T::AppConfig> {
+        Factors::app_config::<T>(self.app_configs).context("no such factor")
```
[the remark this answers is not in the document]
lann: Yeah this will be an explicit error variant later.

crates/factor-outbound-networking/src/lib.rs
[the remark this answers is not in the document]
lann: My only complaint about `AppState` is that it isn't mutable like `InstanceState`. That said, I expect interior mutability to be somewhat common in this type so maybe its a `mut` point :neutral_face:.

crates/factors/src/runtime_config.rs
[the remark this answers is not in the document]
rylev: You could do `F: Factor` and then use `T` and `U` for other items.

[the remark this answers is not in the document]
lann: Done.

crates/factor-outbound-networking/src/lib.rs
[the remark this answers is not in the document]
lann: Done.

crates/factor-outbound-networking/src/lib.rs
lann: Done.

crates/factors/src/factor.rs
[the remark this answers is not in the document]
lann: Done (first option)

crates/factors/src/spin_factors.rs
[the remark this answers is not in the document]
lann: Done.

crates/factors/src/factor.rs
```diff
@@ -11,7 +13,7 @@ pub trait Factor: Any + Sized {
     type AppState: Default;
 
     /// The [`FactorInstancePreparer`] for this factor.
-    type InstancePreparer: FactorInstancePreparer<Self>;
+    type InstancePreparer: Default;
```
rylev, line 16: Question: (realizing that this suggestion moves a bit back in the direction we just came from):

Would it make sense for `Factor` to only have two associated types `AppState` and `InstancePreparer`? Then the `InstancePreparer` would implement the `InstancePreparer` trait which has one associated type `InstanceState` and one method `fn prepare(self) -> anyhow::Result<Self::InstanceState>;`.

crates/factors/src/factor.rs
```diff
@@ -0,0 +1,163 @@
...
+    App, InstancePreparers, Linker, ModuleLinker, PrepareContext, RuntimeConfig, RuntimeFactors,
+};
+
+pub trait Factor: Any + Sized {
+    /// Per-app state for this factor.
+    ///
+    /// See [`Factor::configure_app`].
+    type AppState: Default;
+
+    /// The [`FactorInstancePreparer`] for this factor.
```
rylev, line 15: Many of the doc comments on this trait need to be updated to remove reference to `FactorInstancePreparer`.

crates/factors/src/factor.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// [`App`]. A runtime may - but is not required to - reuse the returned
+    /// config across multiple instances. Note that this may be called without
+    /// any call to `init` in cases where only validation is needed.
+    fn configure_app<T: RuntimeFactors>(
+        &self,
+        ctx: ConfigureAppContext<T>,
+        _runtime_config: &mut impl RuntimeConfig,
+    ) -> anyhow::Result<Self::AppState> {
+        _ = ctx;
+        Ok(Default::default())
```
rylev, line 41: Just noting that I personally think it's worth the boiler plate of having implementations that don't require app configuration or instance preparation writing out these empty methods. 

It's a more verbose but arguably more clear (the reader instantly knows that they don't do any meaningful app configuration/instance preparation without having to know about the default implementation). And it prevents `AppState` and `InstancePreparer` types to always implement `Default` even when they don't have meaningful defaults.

crates/factors/src/factor.rs
```diff
@@ -0,0 +1,163 @@
...
+            add_to_linker(linker, self.get_data)
+        } else {
+            Ok(())
+        }
+    }
+}
+
+pub struct ConfigureAppContext<'a, T: RuntimeFactors> {
+    pub(crate) app: &'a App,
+    pub(crate) app_configs: &'a T::AppState,
```
rylev, line 127: nit: should the field be renamed too?

crates/factors/src/factor.rs
lann: ...unless we bring it back :slightly_smiling_face:

crates/factors/src/factor.rs
lann: I've been trying to model how future changes to the `Factor` trait would work, but I suppose that doesn't mean we have to have default impls for _these_.

crates/factors/src/runtime_config.rs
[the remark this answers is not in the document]
lann: Done.

crates/factors/src/factor.rs
lann: Done.

crates/factors/src/factor.rs
lann: I'm not sure I'm convinced about making `configure_app` mandatory yet but I've made the change and we can see how annoying it is as we implement more factors.

lann: I'm not sure I'm convinced about making `configure_app` mandatory yet but I've made the change and we can see how annoying it is as we implement more factors.

crates/factors/src/lib.rs
lann: Some of this is now gone. Still 2 `transmute`s in the derive code. I briefly tried replacing them with `Any::downcast_ref` but ran into mysterious problems.

crates/factor-variables/src/provider_type.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    type RuntimeConfig: DeserializeOwned;
+    type Provider: Provider;
+
+    fn make_provider(&self, runtime_config: Self::RuntimeConfig) -> anyhow::Result<Self::Provider>;
+}
+
+pub(crate) type ProviderMaker = Box<dyn Fn(toml::Table) -> anyhow::Result<Box<dyn Provider>>>;
+
+pub(crate) fn provider_maker<T: VariablesProviderType>(provider_type: T) -> ProviderMaker {
```
rylev, line 18: This is tricky naming since `ProviderMaker` and `VariablesProviderType` (or what I suggest we call `ProviderBuilder)` do the same thing just that `ProviderMaker` starts from a plan toml table while `ProviderBuilder` starts from a deserialized config.

Perhaps `ProviderBuilder` and `ProviderFromToml` would be better? `ProviderFromToml` sounds more function like (which it is) and the verbosity is fine IMO for an internal implementation detail.

crates/factor-variables/src/provider_type.rs
```diff
@@ -0,0 +1,49 @@
...
+    })
+}
+
+pub struct StaticVariables;
+
+impl VariablesProviderType for StaticVariables {
+    const TYPE: &'static str = "static";
+
+    type RuntimeConfig = StaticVariablesProvider;
+    type Provider = StaticVariablesProvider;
```
rylev, line 32: I don't think you yet have an example where these two types differ. It would be nice to see that to make sure that the complexity of having two different types here is worth it (i.e., why not require the struct that represents the config also plays the role of the provider?)

crates/factors/src/prepare.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl<T: Default + 'static> FactorInstanceBuilder for DefaultInstanceBuilder<T> {
+    type InstanceState = T;
+
+    fn build(self) -> anyhow::Result<Self::InstanceState> {
+        Ok(Default::default())
+    }
+}
+
+pub trait SelfInstanceBuilder: 'static {}
```
rylev, line 23: This and `DefaultInstanceBuilder` are nice, but I'm not convinced they're worth it. It seems like they might make things harder to understand in order to save a few lines of boilerplate...

crates/factors/src/runtime_factors.rs
```diff
@@ -0,0 +1,17 @@
+use crate::Factor;
+
+// TODO(lann): Most of the unsafe shenanigans here probably aren't worth it;
+// consider replacing with e.g. `Any::downcast`.
```
rylev, line 4: Should this TODO be moved - there's no longer any `unsafe` in this file

crates/factor-variables/src/provider_type.rs
```diff
@@ -0,0 +1,49 @@
+use std::{collections::HashMap, sync::Arc};
+
+use serde::{de::DeserializeOwned, Deserialize};
+use spin_expressions::{async_trait::async_trait, Key, Provider};
+use spin_factors::anyhow;
+
+pub trait VariablesProviderType: 'static {
```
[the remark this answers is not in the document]
lann: It isn't really a Builder as the term is usually used. Could be `MakeProvider` I suppose; I've seen that a few places like [`tower`](https://docs.rs/tower/0.4.13/tower/make/index.html).

crates/factor-variables/src/provider_type.rs
lann: It will definitely be different for the vault provider.

crates/factors/src/factor.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Returns [JSON Schema](https://json-schema.org/) for this factor's
+    /// runtime config.
+    ///
+    /// Note that this represents only a fragment of an entire JSON document (a
+    /// "child instance" in JSON Schema terms), so `$schema` isn't needed.
+    ///
+    /// The default implementation returns an empty schema, which accepts any
+    /// configuration.
+    fn runtime_config_json_schema(&self) -> impl serde::Serialize {
+        HashMap::<(), ()>::new()
```
lann, line 54: This is somewhat of a placeholder for now, thinking about how to support https://github.com/fermyon/spin/issues/2529#issuecomment-2158492738

crates/factor-variables/src/provider_type.rs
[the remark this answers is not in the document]
lann: Renamed to `MakeVariablesProvider`

crates/factor-variables/src/provider_type.rs
```diff
@@ -0,0 +1,49 @@
...
+    const TYPE: &'static str;
```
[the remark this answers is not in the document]
lann: Renamed to `RUNTIME_CONFIG_TYPE`

crates/factor-variables/src/provider_type.rs
lann: Done

crates/factors/src/runtime_factors.rs
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+impl<T: RuntimeFactors, F1: Factor, F2: Factor> Clone for StateGetter2<T, F1, F2> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T: RuntimeFactors, F1: Factor, F2: Factor> Copy for StateGetter2<T, F1, F2> {}
```
lann, line 101: I ended up bringing these back. I tried less-unsafe alternatives but didn't land on anything any better.

lann: Could perhaps replace some of _my_ `unsafe` code with _someone else's_ `unsafe` code: https://crates.io/crates/field-offset

crates/factors/src/runtime_factors.rs
```diff
@@ -0,0 +1,101 @@
...
+
+// TODO: This seems fine, but then again I don't understand why `FieldOffset`'s
+// own `Sync`ness depends on `U`...
+unsafe impl<T: RuntimeFactors, F1: Factor, F2: Factor> Sync for StateGetter2<T, F1, F2> {}
```
lann, line 101: A little spooky, but less spooky than what I had before...

crates/factors-derive/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+            type AppState = #app_state_name;
+            type InstanceBuilders = #builders_name;
+            type InstanceState = #state_name;
+
+            fn app_state<F: #Factor>(app_state: &Self::AppState) -> Option<&F::AppState> {
+                let type_id = #TypeId::of::<F>();
+                #(
+                    if type_id == #TypeId::of::<#factor_types>() {
+                        unsafe {
+                            return Some(::std::mem::transmute(&app_state.#factor_names));
```
rylev, line 153: Hello! Long time user of `std::mem::transmute` here. There's a million subtle ways to get usage of `std::mem::transmute` wrong in ways that only sometimes show up or might break with future releases of the Rust compiler. From the docs:

> This makes transmute incredibly unsafe. transmute should be the absolute last resort.

I'm not convinced that we're in the situation where `transmute` is the only way to solve the issue at hand. Would it be possible to explore other solutions here?

lann: I tried using `Any::downcast_mut` in a previous iteration but couldn't get it working. That said, I _think_ this code is fundamentally doing the same thing as `Any::downcast_ref`: dynamically check the `TypeId` and then convert.

`Any::downcast_ref` converts via `as *const dyn Any as *const T` instead of `transmute`; not sure if there is any meaningful difference there.

lann: > I tried using Any::downcast_ref in a previous iteration but didn't get it working.

Apparently the trick was to "try harder": `<dyn Any>::downcast_ref` does the trick, replacing `transmute` here and in `instance_builder_mut` below.

There is still a `transmute` in `instance_state_offset`. That one is going to be trickier to avoid because it's used to implement borrow splitting; I don't have any other ideas for that one.

crates/factors/src/runtime_factors.rs
lann: I have a [fork of `field-offset`](https://github.com/Diggsey/rust-field-offset/pull/26) that fixes it, but I'm not sure that anyone is actually maintaining upstream :grimacing:.

If that PR doesn't get merged I'm not sure I even want to keep the dependency...

lann:
Closing this in favor of working on the `factors` branch.

PR #2523, opened 2024-05-22
ref(oci/client): update unpack_archive_layer to take cache; make pub

vdice:
Refactor for enabling (re-)use of the `unpack_archive_layer` logic (immediate use case: [containerd-shim-spin](https://github.com/spinkube/containerd-shim-spin/pull/121))

- Removes `unpack_archive_layer` from the oci Client struct and upates its signature to take a cache reference, so it can be utilized without needing to create or use an oci Client.

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -413,7 +413,7 @@ impl Client {
                             this.cache.write_wasm(&bytes, &layer.digest).await?;
                         }
                         ARCHIVE_MEDIATYPE => {
-                            this.unpack_archive_layer(&bytes, &layer.digest).await?;
+                            unpack_archive_layer(&this.cache, &bytes, &layer.digest).await?;
                         }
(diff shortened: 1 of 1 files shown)


itowlson:
The archive layer stuff seems like it's OCI-specific rather than cache-specific to me.  (It even has `layer` in the function name.)  I worry that this PR moves several things to slightly awkward places ~and that is my job~, such as functions used only by OCI moving into the common crate not because they are common but because moving the unpack function forces them to go somewhere else.  Plus I like the cache being a reasonably naive thing that just stores things by content address.

I'm not very au fait with the goal here, but I'd like to see the function move back to the OCI crate.  Surely it could be a free function with no dependency on the OCI Client type?

itowlson:
@vdice I like that much better (although others may disagree of course).

itowlson (approved):

PR #2524, opened 2024-05-22
Load components from a registry

itowlson:
Usage:

```
spin_manifest_version = 2

[application]
name = "wkgtest"

[[trigger.http]]
route = "/..."
component = "wkgtest"

[component.wkgtest]
source = { package = "itowlson:ccspintest", version = "0.2.0" }
```

A key outstanding question is how to specify physical registry location.  I've been using the wasm-pkg-tools default config file, but I saw @fibonacci1729's composition SIP specifying the registry as part of the reference.  I don't have strong opinions here except that the default config file is not currently ideal (because of discoverability), but that might change as component tooling consolidates anyway.

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -1,7 +1,7 @@
 use std::path::{Path, PathBuf};
 
-use anyhow::{bail, ensure, Context, Result};
-use futures::future::try_join_all;
+use anyhow::{anyhow, bail, ensure, Context, Result};
+use futures::{future::try_join_all, StreamExt};
 use reqwest::Url;
 use spin_common::{paths::parent_dir, sloth, ui::quoted_path};
 use spin_locked_app::{
@@ -13,7 +13,7 @@ use spin_locked_app::{
 };
 use spin_manifest::schema::v2::{self, AppManifest, KebabId, WasiFilesMount};
 use spin_outbound_networking::SERVICE_CHAINING_DOMAIN_SUFFIX;
-use tokio::sync::Semaphore;
+use tokio::{io::AsyncWriteExt, sync::Semaphore};
 
 use crate::{cache::Cache, FilesMountStrategy};
 
@@ -222,6 +222,14 @@ impl LocalLoader {
             v2::ComponentSource::Remote { url, digest } => {
                 self.load_http_source(&url, &digest).await?
             }
+            v2::ComponentSource::Registry {
+                registry,
+                package,
+                version,
+            } => {
+                self.load_registry_source(registry.as_ref(), &package, &version)
+                    .await?
+            }
         };
         Ok(LockedComponentSource {
             content_type: "application/wasm".into(),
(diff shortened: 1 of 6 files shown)


lann:
> how to specify physical registry location

I think it probably makes sense to be able to _optionally_ set the registry explicitly in the manifest, e.g.:
```toml
source = { package = "itowlson:ccspintest", registry = "example.com" }
```
This would override any global package name -> registry mapping but would still obey any global registry config for `example.com`.

itowlson:
Thanks @lann - I'll add that now.

crates/manifest/src/schema/common.rs
```diff
@@ -30,13 +30,21 @@ pub enum ComponentSource {
         /// `digest = `"sha256:abc123..."`
         digest: String,
     },
+    /// `{ ... }`
+    Registry {
+        /// `package = "example:component"
+        package: String,
+        /// `version = "1.2.3"
+        version: String,
```
fibonacci1729, line 38: Any reason to not use `semver::Version` here or are we keeping it fairly generic at this point in loading?

itowlson: Technically "not realising that `wasm_pkg_loader::Version` _is_ `semver::Version` is a reason so yes

itowlson:
Registry setting is added.

crates/loader/src/local.rs
```diff
@@ -251,6 +259,54 @@ impl LocalLoader {
         file_content_ref(path)
     }
 
+    async fn load_registry_source(
+        &self,
+        registry: Option<impl AsRef<str>>,
```
rylev, line 264: nit: If this isn't a public API, I don't think we need to pay the compilation price for making this generic.

```suggestion
        registry: Option<&str>,
```

rylev (approved):

fibonacci1729 (approved):

crates/manifest/src/schema/common.rs
```diff
@@ -30,13 +30,33 @@ pub enum ComponentSource {
...
+            ComponentSource::Registry {
+                registry,
+                package,
+                version,
+            } => {
+                let registry_suffix = match registry {
+                    None => "default registry".to_owned(),
+                    Some(r) => format!("registry {r:?}"),
+                };
+                write!(f, "\"{package}@{version}\" from {registry_suffix:?}")
```
lann, line 58:
```suggestion
                write!(f, "\"{package}@{version}\" from {registry_suffix}")
```

lann (approved):

crates/loader/src/local.rs
```diff
@@ -251,6 +259,57 @@ impl LocalLoader {
...
 
+    async fn load_registry_source(
+        &self,
+        registry: Option<&String>,
+        package: &str,
+        version: &semver::Version,
+    ) -> Result<ContentRef> {
+        let package: wasm_pkg_loader::PackageRef = package
+            .parse()
+            .with_context(|| format!("{package} is not a valid registry package reference"))?;
```
lann, line 270: Since I have nitted once I shall nit again:

```suggestion
            .with_context(|| format!("{package:?} is not a valid registry package reference"))?;
```

(`Debug` in error messages helps highlight surprise whitespace, etc)

crates/manifest/src/schema/common.rs
itowlson: I PUT THAT FORMATTING IN SPECIALLY FOR YOU LANN

lann: DOUBLE THE QUOTES, DOUBLE THE ENLIGHTENMENT

Issue #2520, opened 2024-05-20
Received runtime error that should be thrown at component build time


asteurer:
Received the below error when running `spin up`. I incorrectly capitalized the variables in my spin.toml file, and this caused some delays when the app succeeded after `spin build`, but failed at runtime. Can we have this error at build time, rather than runtime?
`Error: invalid variable name: "DB_URL": invalid character 'D'. Variable names may contain only lower-case letters, numbers, and underscores.`

itowlson:
We used to fully load the manifest at build time, but it threw up a lot of irrelevant “you need to fix this completely different thing before you can find out why the borrow checker is mad” errors.  It was derailing so we made it more forgiving.  That said:

1. Spin 2 introduced a better pipeline for manifest loading. Spin 1 had a monolithic loader which meant _anything_ wrong would fail the load.  So reading the whole manifest may be less derailing now.
2. We could rearrange things so that `spin build` does the build _regardless_ and also does a validation pass on the manifest. This would allow us to report validation failures at build time without them being derailing.

PR #2527, opened 2024-05-23
Try to load manifest during `spin build`

itowlson:
This partly, but only partly, addresses #2520.

With this change, `spin build` attempts to load the full manifest, falling back to the subset only if that fails, and emitting a warning for the non-build-related validation failure.  This will catch schema errors such as `allowed_outbound_hsots = ...`:

```
$ spin build
Building component self-req-test with `cargo build --target wasm32-wasi --release`
    Finished `release` profile [optimized] target(s) in 0.03s
Warning: The manifest has errors not related to the Wasm component build. Error details:
TOML parse error at line 18, column 1
   |
18 | allowed_outbound_hsots = []
   | ^^^^^^^^^^^^^^^^^^^^^^
unknown field `allowed_outbound_hsots`, expected one of `source`, `description`, `variables`, `environment`, `files`, `exclude_files`, `allowed_http_hosts`, `allowed_outbound_hosts`, `key_value_stores`, `sqlite_databases`, `ai_models`, `build`, `tool`
```

HOWEVER.

The problem in #2520 relates to variable name validation, and that does _not_ happen at manifest load, or even lockfile creation, but _in the trigger_ when it tries to resolve the variables.  Checking that in the loader would require additional validation logic in the loader, similar to how we validate kebab-ids.  Nothing stops us adding that validation logic, but it's not there now, and _for this PR_ I didn't want to extend load-time validation (with who knows what consequences) but only run the existing load-time validation at build time.  If we decide to add variable name validation at load time then `spin build` with this PR will inherit it.

### DIFF
--- a/crates/build/src/manifest.rs
+++ b/crates/build/src/manifest.rs
@@ -5,8 +5,35 @@ use std::{collections::BTreeMap, path::Path};
 use spin_manifest::{schema::v2, ManifestVersion};
 
 /// Returns a map of component IDs to [`v2::ComponentBuildConfig`]s for the
-/// given (v1 or v2) manifest path.
+/// given (v1 or v2) manifest path. If the manifest cannot be loaded, the
+/// function attempts fallback: if fallback succeeds, result is Ok but the load error
+/// is also returned via the second part of the return value tuple.
 pub async fn component_build_configs(
+    manifest_file: impl AsRef<Path>,
+) -> Result<(Vec<ComponentBuildInfo>, Option<spin_manifest::Error>)> {
+    let manifest = spin_manifest::manifest_from_file(&manifest_file);
+    match manifest {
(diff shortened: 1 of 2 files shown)


mikkelhegn (approved):
The approach in this PR looks good to me. (Not a code review :-))

crates/build/src/manifest.rs
```diff
@@ -8,6 +8,31 @@ use spin_manifest::{schema::v2, ManifestVersion};
 /// given (v1 or v2) manifest path.
 pub async fn component_build_configs(
```
lann, line 9: Perhaps rename this `component_build_configs_with_error` or something? And update the docstring?

itowlson: I'm not mad keen on that name (and can't think of a fully expressive one that isn't awkward) but have updated the doc string.

lann (approved):

PR #2539, opened 2024-05-31
chore(loader): bump wasm-pkg-tools

vdice:
To get tracing-log 'fix' (deactivation) in https://github.com/bytecodealliance/wasm-pkg-tools/pull/26

Defer to @lann on if we'd like to wait for https://github.com/bytecodealliance/wasm-pkg-tools/issues/27 as well.


lann (approved):
No reason to wait if it solves a problem for you

PR #2541, opened 2024-06-03
fix `StoreBuilder::inherit_limited_network`

dicej:
Previously, this called `WasiCtxBuilder::inherit_network`, but that had no effect since `StoreBuilder::build_with_data` later overwrites that setting by calling `WasiCtxBuilder::socket_addr_check` with a lambda that uses `StoreBuilder::net_pool` to check addresses.  In this case, `StoreBuilder::net_pool` has not had any subnets added to it, so it denies everything, which is the opposite of what we intended.

The solution is to have `StoreBuilder::inherit_limited_network` update `net_pool` to allow all IPv4 and IPv6 networks.

### DIFF
--- a/crates/core/src/store.rs
+++ b/crates/core/src/store.rs
@@ -1,10 +1,11 @@
 use anyhow::{anyhow, Result};
 use bytes::Bytes;
 use cap_primitives::net::Pool;
-use cap_std::ipnet::IpNet;
+use cap_std::ipnet::{IpNet, Ipv4Net, Ipv6Net};
 use std::{
     io::{Read, Write},
     mem,
+    net::{Ipv4Addr, Ipv6Addr},
     path::{Path, PathBuf},
     sync::{Arc, Mutex},
     time::{Duration, Instant},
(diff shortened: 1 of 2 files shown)


crates/core/src/store.rs
```diff
@@ -207,11 +208,19 @@ impl StoreBuilder {
...
             }
-            WasiCtxBuilder::Preview2(ctx) => {
+            WasiCtxBuilder::Preview2(_) => {
                 // TODO: ctx.allow_udp(false);
-                ctx.inherit_network();
             }
         });
+
+        self.net_pool.insert_ip_net_port_any(
+            IpNet::V4(Ipv4Net::new(Ipv4Addr::new(0, 0, 0, 0), 0).unwrap()),
```
rylev, line 217: I get the hesitation to do `"0.0.0.0/0".parse().unwrap()`, but I do prefer it for readability. Perhaps a comment stating that this adds all IP addresses will make this clear. It took me a second to realize this was a CIDR range and a comment would have helped.

crates/core/src/store.rs
```diff
@@ -207,11 +208,19 @@ impl StoreBuilder {
             WasiCtxBuilder::Preview1(_) => {
```
rylev, line 208: This isn't new for this PR, but the name of this method and its comment aren't really a reflection of reality. It might make sense to change this method to just `inherit_network`.

rylev (approved):

rylev:
@dicej looks like this got stuck for some reason. I'm not sure, but a rebase might fix things?

itowlson:
The DCO app got borked for a while.  It is now healthy but it seems like you have to re-push to get it to regenerate status.

Issue #1779, opened 2023-09-18
Component Imports


fibonacci1729:
For 2.0 (when we officially support preveiw2 components) allow developers to specify how to satisfy component imports in the manifest. E.g.:

```toml
[component.foo.imports]
bar = { path = "bar.wasm" }
```

fibonacci1729:
Closing in favor of #2543

PR #2543, opened 2024-06-05
Component Dependencies SIP

fibonacci1729:
This PR supersedes #1779 

[Rendered](https://github.com/fibonacci1729/spin/blob/component-dependencies-sip/docs/content/sips/020-component-dependencies.md)

### DIFF
--- a/docs/content/sips/020-component-dependencies.md
+++ b/docs/content/sips/020-component-dependencies.md
@@ -1,0 +1,145 @@
+title = "SIP 020 - Component Dependencies"
+template = "main"
+date = "2024-05-31T12:00:00Z"
+---
+Summary: This feature enables developers to specify and manage dependencies for Spin components, allowing the use of components from various languages and sources, either locally or remotely, with configurable inheritance of parent component configuration.
+
+Owner(s): [brian.hardock@fermyon.com](mailto:brian.hardock@fermyon.com)
+
+Created: May 31, 2024
+
+# Background
+This feature enables the polyglot re-use of components, allowing developers to specify how to use components written in various languages as libraries to fulfill dependencies in their Spin components.
+
+For instance, consider writing a Spin component in Go that depends on the `aws:client/s3` interface (expressed as an import in the compiled Go component, e.g., `import aws:client/s3`). To satisfy this dependency, you can use a componentized S3 client written in Rust that exports an instance of the `aws:client/s3` interface. Using the proposed syntax in this SIP, Spin developers can instruct Spin on how to compose these components together to achieve this.
+
+# Proposal
+Developers can use the following syntax to specify that they want to use a component as a dependency, either from a local disk or via a remote component registry:
+
+> ℹ️ NOTE: Throughout this document we refer to the concept of a registry. However, it is out of scope of this proposal to detail what that means. To read more about what is meant by registry see [The OpenContainers Distribution Spec](https://specs.opencontainers.org/distribution-spec/?v=v1.0.0).
+
+## Using dependencies from disk (local)
+```toml
+[component."infra-dashboard".dependencies]
+# Components can satisfy dependencies by import name (e.g., "aws:client/s3")
+# using a component from disk. It is implicitly assumed that the
+# `my_aws_client.wasm` component exports an instance of the
+# `aws:client/s3` interface.
+"aws:client/s3" = { path = "my_aws_client.wasm" }
+
+# Optionally, explicitly specify the export name to use to satisfy the `aws:client/s3` interface.
+"aws:client/s3" = { path = "my_aws_client.wasm", export = "my-s3-client" }
+
+# Without an interface name, attempt to resolve every import of the "aws:client" package.
+"aws:client" = { path = "my_aws_client.wasm" }
+```
+
+> ⚠️ NOTE: Using the package and interface forms together (e.g. `aws:client` and `aws:client/s3` respectively) within the same dependencies section will result in an error to prevent resolution ambiguities.
+
+## Using dependencies from a registry (remote)
+```toml
+[component."infra-dashboard".dependencies]
+# Equivalent to { version = "1.0.0" , package = "aws:client"}
+"aws:client" = "1.0.0"
+
+# Use the `aws:client` component package to satisfy any number of "wasi:blobstore" imports ...
+"wasi:blobstore" = { registry = "my-registry.io", version = "0.1.0", package = "aws:client" }
+```
+
+## Dependency Names
+Dependency names, as referenced above, have two forms: a plain kebab-case identifier (e.g. `my-name`) or a package pattern. The package pattern can be either the name of a package (e.g. `aws:client` or `aws:client@0.1.0`) or specify a concrete interface of a package (e.g. `aws:client/s3` or `aws:client/s3@0.1.0`). 
+
+When using the package pattern form of dependency names in a component manifest's dependency section the set of names and versions used must not conflict. Two patterns conflict if they have overlapping interfaces and semver compatible versions. In otherwords the set of patterns must be disjoint when fully mapped to the underlying imports of the dependent component.
+
+Some examples of conflicting dependency names:
+* `aws:client` _conflicts_ with `aws:client@0.1.0`
+* `aws:client` _conflicts_ with `aws:client/s3@0.1.0`
+* `aws:client/s3@0.1.0` _conflicts_ with `aws:client/s3@0.1.1`
+
+## Dependency Isolation
+Statically composing the parent Spin component and its dependencies yields a component where the parent component and dependencies are no longer isolated from each other. Each dependency will inherit the configuration of the parent component, which may be desireable in some library use-cases (e.g., allowing the `aws:client/s3` dependency to make outbound requests via `allowed_outbound_hosts`). However, making this the default behavior contradicts the explicit opt-in nature of capabilities for Spin (and the component model), so by default, component dependencies in Spin will not inherit configuration.
+
+To allow dependencies to inherit configuration from the parent Spin component, developers can set `dependencies_inherit_configuration = true`, as shown in the following example:
+
+```toml
+[component."infra-dashboard"]
+# ...
+allowed_outbound_hosts = ["https://s3.us-west-2.amazonaws.com"]
+dependencies_inherit_configuration = true
+
+[component."infra-dashboard".dependencies]
+"aws:client" = "1.0.0"
+```
+
+The above configuration allows the component used to satisfy the `aws:client` dependency to make outbound HTTP requests to an S3 bucket 
+specified by the host included in `allowed_outbound_hosts` (e.g. `https://s3.us-west-2.amazonaws.com/my-bucket/puppy.jpeg`). If `dependencies_inherit_configuration` is omitted or set to `false`, an error indicating the capability to make outbound HTTP requests is disabled, will be returned whenever the `aws:client` component attempts to make an outbound HTTP request.
+
+### Configuration inheritance
+At the time of writing this proposal, the following component-level configurations in a `Spin.toml` could be inherited by dependency components if inheritance is enabled:
+
+* `allowed_outbound_hosts` 
+* `key_value_stores`
+* `variables`
+* `ai_models`
+* `files`
+* `environment`
+
+
+### The `spin-deny-all` adapter
+To implement the default behavior (i.e. `dependencies_inherit_configuration = false`), Spin will apply what's called the `spin-deny-all` adapter to maintain isolation between components. When configuration inheritance is disabled, Spin will compose the ahead-of-time constructed deny adapter with each dependency, resulting in a new component that denies access to various capabilities (e.g. `wasi:http/outgoing-handler`, `fermyon:spin/variables`, etc.).
+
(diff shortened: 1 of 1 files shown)


docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,108 @@
...
+"aws:client" = "1.0.0"
+
+# Use the `aws:client` component package to satisfy any number of "wasi:blobstore" imports ...
+"wasi:blobstore" = { registry = "my-registry.io", version = "0.1.0", package = "aws:client" }
+```
+
+## Dependency Isolation
+Statically composing the parent Spin component and its dependencies yields a component where the parent component and dependencies are no longer isolated from eachother. Each dependency will inherit the configuration of the parent component, which may be desireable in some library use-cases (e.g., allowing the `aws:client/s3` dependency to make outbound requests via `allowed_outbound_hosts`). However, making this the default behavior contradicts the explicit opt-in nature of capabilities for Spin (and the component model), so by default, component dependencies in Spin will not inherit configuration.
+
+To allow dependencies to inherit configuration from the parent Spin component, developers can set `dependencies_inherit_configuration = true`, as shown in the following example:
```
itowlson, line 50: What gets inherited here e.g. files?  Part of the original variables plan was to allow variables to be mapped into dependencies (specifically, with different names) - is that part of this?

itowlson (commented):
Great to see this under way.  Thank you so much for driving it Brian.

docs/content/sips/0xx-component-dependencies.md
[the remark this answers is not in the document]
fibonacci1729: The way we would want to do this is to allow pointing a dependency at another component section in the Spin manifest which would allow configuring each dependency separately. This is tricky at the moment with the current state of virtualization tooling when dealing with `wasi:io`.  The best we can do in the near term to deliver value is all-or-nothing configuration inheritance plus a quick follow up to allow the more granular future option (which needs a bit of research and is why its not the plan of record here at the moment).

fibonacci1729: I'm struggling to recall, could you elaborate on the original variables plan? 

Would it be helpful to record in the SIP the set of configurations that are inherited, i.e.,:
- `allowed_outbound_hosts`
- `key_value_stores`
- `sqlite_databases`
- `variables`
- `ai_models`
- `files`
- `environment`

itowlson: The motivbation for variables was:

* I take dependencies on `alice:aws-s3` and `bob:aws-rds`.
* `alice:aws-s3` requires an AWS auth token in a variable named `aws-token`.
* `bob:aws-rds` requires an AWS auth token in a variable named `auth-for-services`. (This is why I stay away from Bob's libraries.)
* I would like to have one configuration variable that I can feed through to both - e.g. have a `aws-auth` application variable, and the Alice dep has `aws-token = "{{ aws-auth }}"` and the Bob dep has `auth-for-services = "{{ aws-auth }}"`.

Does that help answer the question?  If the answer is "you need to define both variables on the parent component and inherit" then that's fine, just seeking to clarify.

Yeah listing what gets inherited and what doesn't would be super useful!  Thanks!

fibonacci1729: Ah yeah, thanks for the refresher! Yeah that would require including both variables on the parent component and inherit within this proposal.

Will update!

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,108 @@
...
+# Components can satisfy dependencies by import name (e.g., "aws:client/s3")
+# using a component from disk. It is implicitly assumed that the
+# `my_aws_client.wasm` component exports an instance of the
+# `aws:client/s3` interface.
+"aws:client/s3" = { path = "my_aws_client.wasm" }
+
+# Optionally, explicitly specify the export name to use to satisfy the `aws:client/s3` interface.
+"aws:client/s3" = { path = "my_aws_client.wasm", export = "my-s3-client" }
+
+# Without an interface name, attempt to resolve every import of the "aws:client" package.
```
rylev, line 31: Does this include wasi imports? What happens if the provided wasm file does not include an export that satisfies an import? Do we error early or do we simply implicitly import anything that was not satisified?

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,108 @@
...
+[component."infra-dashboard"]
+# ...
+allowed_outbound_hosts = ["https://s3.us-west-2.amazonaws.com/my-bucket/puppy.jpeg"]
+dependencies_inherit_configuration = true
+
+[component."infra-dashboard".dependencies]
+"aws:client" = "1.0.0"
+```
+
+The above configuration allows the component used to satisfy the `aws:client` dependency to make outbound HTTP requests to the S3 bucket specified in `allowed_outbound_hosts`. If `dependencies_inherit_configuration` is omitted or set to `false`, an `AccessDenied` error will be returned whenever the `aws:client` component attempts to make an outbound HTTP request.
```
rylev, line 62: I was under the impression that we would (for now) simply error if `dependencies_inherit_configuration` is omitted or set to `false`. This way we only have to support inheriting capabilities but with a future compatible path towards not inheriting capabilities. 

I see below that you want to use the `spin-deny-all` adapter to achieve this. This should work as long as we're fine with having to virtualize all of `wasi:io`. I'm not fully sure about performance issues with doing full virtualization though.

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,108 @@
...
+### The `spin-deny-all` adapter
+To implement the default behavior (i.e. `dependencies_inherit_configuration = false`), Spin will apply what's called the `spin-deny-all` adapter to maintain isolation between components. When configuration inheritance is disabled, Spin will compose the ahead-of-time constructed deny adapter with each dependency, resulting in a new component that denies access to various capabilities (e.g. `wasi:http/outgoing-handler`, `fermyon:spin/variables`, etc.).
+
+The `spin-deny-all` adapter is a component that targets the following world:
+
+```
+package fermyon:spin-virt;
+
+world deny-all {
+    # WASI
```
rylev, line 73: I'm not sure this is going to be possible without virtualizing all of wasi (due to the `wasi:io` issue we've talked above before). I guess that's fine, since we'll need to do this in the future any way if we want to support component isolation, but it does significantly complicate the implementation.

fibonacci1729: Given the context Lann provided above and since we discussed out-of-band i'll resolve this thread! 

tl;dr for anyone skimming: because we aren't actually virtualizing `wasi:io` and simply passing through we avoid all the complicate details that Ryan is referring to.

docs/content/sips/0xx-component-dependencies.md
fibonacci1729: > Does this include wasi imports?

I don't have a strong technical argument for why to exclude wasi imports but i'm also tending towards being a little conservative here in the initial MVP. So maybe we should error initially if a wasi imports is listed in dependencies? We can always relax this constraint in the future.

> What happens if the provided wasm file does not include an export that satisfies an import?

This is a good question that i will update the SIP to answer once we decide what should be the expected behavior here. My current thinking (for the MVP) is that we should tend towards simply importing anything that was not satisfied.

I would like to reject and error on invalid dependencies. So if a dependency points to component that can't satisfy anything that should probably error.

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,108 @@
...
+Summary: This feature enables developers to specify and manage dependencies for Spin components, allowing the use of components from various languages and sources, either locally or remotely, with configurable inheritance of parent component configuration.
+
+Owner(s): [brian.hardock@fermyon.com](mailto:brian.hardock@fermyon.com)
+
+Created: May 31, 2024
+
+# Background
+This feature enables the polyglot re-use of components, allowing developers to specify how to use components written in various languages as libraries to fulfill dependencies in their Spin components.
+
+For instance, consider writing a Spin component in Go that depends on the `aws:client/s3` interface (expressed as an import in the compiled Go component, e.g., `import aws:client/s3`). To satisfy this dependency, you can use a componentized S3 client written in Rust that exports an instance of the `aws:client/s3` interface. Using the proposed syntax in this SIP, Spin developers can instruct Spin on how to compose these components together to achieve this.
```
itowlson, line 14: Do we allow for the possibility that the dependency has imports, e.g. callbacks?  Is that a "hie thee to `wac`" situation?

fibonacci1729: I wasn't planning on placing any restrictions on the shape of dependencies. That being said, if a dependency contains imports that can not be satisfied by Spin the component will fail to instantiate.

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,108 @@
...
+allowed_outbound_hosts = ["https://s3.us-west-2.amazonaws.com/my-bucket/puppy.jpeg"]
+dependencies_inherit_configuration = true
+
+[component."infra-dashboard".dependencies]
+"aws:client" = "1.0.0"
+```
+
+The above configuration allows the component used to satisfy the `aws:client` dependency to make outbound HTTP requests to the S3 bucket specified in `allowed_outbound_hosts`. If `dependencies_inherit_configuration` is omitted or set to `false`, an `AccessDenied` error will be returned whenever the `aws:client` component attempts to make an outbound HTTP request.
+
+### The `spin-deny-all` adapter
```
mikkelhegn, line 64: It would be great with a small note (or reference to) what an 'adapter' is. I've heard the term used in the world of components before, and I think it may be just another component, but with a particular purpose?

lann: An adapter is just a kind of component that is designed to wrap around another component and change its imports/exports in some way.

rylev: Side note: There is a slightly different (albeit related) usage of "adapter" which is the adapters used to wrap preview1 core modules in a preview 2 components (i.e., a component wrapping a module instead of the usage here of a component wrapping a component).

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,143 @@
...
+
+[component."infra-dashboard".dependencies]
+"aws:client" = "1.0.0"
+```
+
+The above configuration allows the component used to satisfy the `aws:client` dependency to make outbound HTTP requests to an S3 bucket 
+specified by the host included in `allowed_outbound_hosts` (e.g. `https://s3.us-west-2.amazonaws.com/my-bucket/puppy.jpeg`). If `dependencies_inherit_configuration` is omitted or set to `false`, an error indicating the capability to make outbound HTTP requests is disabled, will be returned whenever the `aws:client` component attempts to make an outbound HTTP request.
+
+### Configuration inheritance
+At the time of writing this proposal, the following component-level configurations in a `Spin.toml` could be inheritied by dependency components if inheritance is enabled:
```
itowlson, line 76: tiny typo
```suggestion
At the time of writing this proposal, the following component-level configurations in a `Spin.toml` could be inherited by dependency components if inheritance is enabled:
```

itowlson (approved):

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,143 @@
...
+# Optionally, explicitly specify the export name to use to satisfy the `aws:client/s3` interface.
+"aws:client/s3" = { path = "my_aws_client.wasm", export = "my-s3-client" }
+
+# Without an interface name, attempt to resolve every import of the "aws:client" package.
+"aws:client" = { path = "my_aws_client.wasm" }
+```
+
+> ⚠️ NOTE: Using the package and interface forms together (e.g. `aws:client` and `aws:client/s3` respectively) within the same dependencies section will result in an error to prevent resolution ambiguities.
+
+## Using dependencies from a registry (remote)
```
radu-matei, line 37: I assume it's not the intention of this SIP to talk more about _what_ that registry is. Do you want to explicitly point that out?

docs/content/sips/0xx-component-dependencies.md
```diff
@@ -0,0 +1,143 @@
...
+
+Some examples of conflicting dependency names:
+* `aws:client` _conflicts_ with `aws:client@0.1.0`
+* `aws:client` _conflicts_ with `aws:client/s3@0.1.0`
+* `aws:client/s3@0.1.0` _conflicts_ with `aws:client/s3@0.1.1`
+
+## Dependency Isolation
+Statically composing the parent Spin component and its dependencies yields a component where the parent component and dependencies are no longer isolated from eachother. Each dependency will inherit the configuration of the parent component, which may be desireable in some library use-cases (e.g., allowing the `aws:client/s3` dependency to make outbound requests via `allowed_outbound_hosts`). However, making this the default behavior contradicts the explicit opt-in nature of capabilities for Spin (and the component model), so by default, component dependencies in Spin will not inherit configuration.
```
radu-matei, line 58: typo: _eachother_

radu-matei (approved):
A few non blocking questions, this looks great to me, thank you!


LGTM

Issue #1692, opened 2023-08-14
Support exclusions in spin watch configuration


mikkelhegn:
In some circumstances, it might be nice to be able to "include everything, but" for the Spin watch configuration. Rather than having to explicitly include everything.

In this example: https://github.com/mikkelhegn/docusaurus-example (generated from a standard Docusaurus template), the build output is a subfolder to the `my-websites` folder, where all the source is. Which means if the whole directory is watched, it will loop at every build. To avoid the spin watch config would have to include all subdirectories and files in the root, for it to work well.

In this example, being able to tell Spin to watch everything but the `build` folder, would be a mush easier config.

kate-goldenring:
Similar to the [syntax of `.gitignore`](https://www.atlassian.com/git/tutorials/saving-changes/gitignore), I'd imaging us supporting "!build".

kate-goldenring:
Looks like we have [`ignore_patterns` in our watch Config](https://github.com/fermyon/spin/blob/main/src/watch_filter.rs#L209). @calebschoepp do we already have support for ignoring certain paths?

calebschoepp:
@kate-goldenring currently the ignored paths are hardcoded and we don't support any way for users to pass in paths that they want to ignore. I do think that supporting some sort of ignore path is a good idea. Perhaps the simplest implementation is just treating watch paths with a `!` prefix as something to be ignored like Kate mentioned.

me-diru:
Hi!

The code has gone through some refactor from the time this issue was raised. After some initial testing, I think the existing code implements @kate-goldenring and @calebschoepp suggestion of using `!` in front of the file/folder in watch files.

For example, when I included everything in the file and used `!filename.rs`, it is excluded from spin watch, similarly can exclude a folder with `!src\*`

Would like to know your thoughts on if covers the issue or anything else/edge-cases to look out

![image](https://github.com/fermyon/spin/assets/39561007/a5b0af4f-4b8e-4e76-8306-2eff3f0ac183)

Appreciate your input! :D

itowlson:
I think none of your patterns cover `spin.toml` - your wildcard is `*/*` which I would expect to cover subdirectories only.

But `spin.toml` should be watched anyway.  A component shouldn't be able to ignore away manifest changes.

itowlson:
[...]

PR #2554, opened 2024-06-13
Fix exclusions in `build.watch` getting applied to all components

itowlson:
This tidies some issues arising from the revelation that `spin watch` understands `!` syntax for exclusions after all.  Thank you @me-diru for finding these little nasties!

The first issue is that an exclusion in one component affects all other components.  This happens because we combine all component globs into one list so that we can run one instance of `watchexec`, so any exclusions that get chucked into that list and apply list-wide.

The second issue is that adding an exclusion for `spin.toml` means that components no longer rebuild on manifest changes (e.g. to be `build.command`).  Well, okay, you did ask for the bad thing.  Because of the first issue this affects _all_ components which is a bad thing you _didn't_ ask for.

The fix in this PR is, instead of combining the globs, to create a composite filterer which treats a change as triggering if _any_ component expresses an interest in it.  A  component can still use bang syntax to exclude a file, but if other components are interested in that same file then changes to that file will still cause a rebuild.  There is still only one instance of `watchexec`, but the filterer for the buildifier is now built out of per-component filterers, plus a separate filterer for the manifest (so that it doesn't need to be spliced into individual components, and therefore can't be banged away).

The composite filterer is used only for the build patterns, because assets are derived from the actual file mounts, and therefore don't have the same problem of user-provided pattern syntax.

### DIFF
--- a/src/commands/watch/filters.rs
+++ b/src/commands/watch/filters.rs
@@ -7,6 +7,7 @@ use anyhow::Context;
 use async_trait::async_trait;
 use spin_common::ui::quoted_path;
 use spin_manifest::schema::v2;
+use watchexec::filter::Filterer;
 
 #[async_trait]
 pub(crate) trait FilterFactory: Send + Sync {
@@ -15,7 +16,7 @@ pub(crate) trait FilterFactory: Send + Sync {
         manifest_file: &Path,
         manifest_dir: &Path,
         manifest: &v2::AppManifest,
-    ) -> anyhow::Result<Arc<watchexec_filterer_globset::GlobsetFilterer>>;
+    ) -> anyhow::Result<Arc<dyn Filterer>>;
 }
 
 pub(crate) struct ArtifactFilterFactory {
@@ -33,7 +34,7 @@ impl FilterFactory for ArtifactFilterFactory {
         manifest_file: &Path,
         manifest_dir: &Path,
         manifest: &v2::AppManifest,
-    ) -> anyhow::Result<Arc<watchexec_filterer_globset::GlobsetFilterer>> {
+    ) -> anyhow::Result<Arc<dyn Filterer>> {
         let manifest_glob = if self.skip_build {
             vec![manifest_path_to_watch(manifest_file)?]
         } else {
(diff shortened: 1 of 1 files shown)


src/commands/watch/filters.rs
```diff
@@ -185,3 +181,23 @@ fn standard_ignores() -> Vec<(String, Option<PathBuf>)> {
...
+        priority: watchexec::event::Priority,
+    ) -> Result<bool, watchexec::error::RuntimeError> {
+        for f in &self.filterers {
+            if f.check_event(event, priority)? {
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+}
```
calebschoepp, line 203: Nit: A comment explaining that `any(event)` triggers this filter rather than `all(event)` might be nice.

src/commands/watch/filters.rs
```diff
@@ -185,3 +181,23 @@ fn standard_ignores() -> Vec<(String, Option<PathBuf>)> {
     .map(|pat| (pat.to_owned(), None))
     .collect()
 }
+
+#[derive(Debug)]
+struct CompositeFilterer {
+    filterers: Vec<Box<dyn watchexec::filter::Filterer>>,
```
calebschoepp, line 187: Nit: Perhaps put a `Vec<Box<dyn watchexec::filter::Filterer>>` behind a type alias? Not sure if that is cleaner so I'll leave that up to you.

calebschoepp (approved):
Couple nits but otherwise LGTM. Thanks for hunting this down Ivan.

src/commands/watch/filters.rs
itowlson: Hmm, to my mind it doesn't occur enough, or have a distinct semantic concept other than "vector of filterers", to merit the indirection.  I agree Rust's insistence on box-dyn ceremony makes it harder to read than it needs to be though.  If we want to do this I think I'd alias the box-dyn and make this `Vec<AnyFilterer>`...

me-diru:
Thanks a lot, @itowlson, for the quick rescue against the negation madness! :D

I tested the fix with multiple components and tried to exclude files included under one component in another and it still worked.

Can we assume this PR adds the feature requested in #1692 as well? It seems to exclude files under the same component with a `!` operator

me-diru (approved):
Thanks a lot, @itowlson, for the quick rescue against the negation madness! :D

I tested the fix with multiple components and tried to exclude files included under one component in another and it still worked.

Can we assume this PR adds the feature requested in #1692 as well? It seems to exclude files under the same component with a `!` operator

itowlson:
@me-diru Once this merges we can flag it on #1692 for folks to test.  The reason I am not saying "yes" is that the watchexec exclusion syntax feels a bit like we lucked into whatever watchexec does rather than designing it to meet our needs, so I would rather have the folks who want the feature to confirm that it really does work the way they want it!

PR #2556, opened 2024-06-13
Spin Build Targets Check SIP

fibonacci1729:
[Rendered](https://github.com/fibonacci1729/spin/blob/65274ed4377b178a2eec7cad3179bec6a0a7caf1/docs/content/sips/0XX-spin-build-targets-check.md)

### DIFF
--- a/docs/content/sips/0XX-spin-build-targets-check.md
+++ b/docs/content/sips/0XX-spin-build-targets-check.md
@@ -1,0 +1,65 @@
+title = "SIP 0XX - Spin Build Targets Check"
+template = "main"
+date = "2024-06-13T12:00:00Z"
+---
+Summary: Extend `spin build` to validate component targets against trigger types.
+
+Owner(s): 
+    [brian.hardock@fermyon.com](mailto:brian.hardock@fermyon.com)
+    [michelle@fermyon.com](mailto:michelle@fermyon.com)
+
+Created: June 13, 2024
+
+# Background
+When associating a Spin component to a specific trigger in the Spin manifest, there is an implied world that the Spin component is targeting (e.g. the http trigger type implies `fermyon:spin/http-trigger`). However, we don't know if the Spin component actually implements that world until instantiation time (i.e. when a request / event comes in).
+
+As a visual example, if a user mistankenly associates an http trigger with a component that targeted the redis world:
+
+`examples/hello-world/spin.toml:`
+```examples/hello-world/spin.toml
+[[trigger.http]]
+route = "/hello"
+component = "hello"
+
+[component.hello]
+source = "redis-component.wasm" # this is a redis component which is wrong
+```
+
+When issuing a curl to this endpoint:
+```
+❯ curl -i http://127.0.0.1:3000/hello
+HTTP/1.1 500 Internal Server Error
+content-length: 0
+date: Tue, 04 Jun 2024 21:53:41 GMT
+```
+
+And in the output of `spin up` we see:
+```
+❯ spin up -f examples/hello-world 
+Logging component stdio to "examples/hello-world/.spin/logs/"
+
+Serving http://127.0.0.1:3000
+Available Routes:
+  hello: http://127.0.0.1:3000/hello
+    A simple component that returns hello.
+2024-06-04T21:53:16.160773Z ERROR spin_trigger_http: Error processing request: Expected component to either export `wasi:http/incoming-handler@0.2.0-rc-2023-10-18` or `fermyon:spin/inbound-http` but it exported neither    
+```
+
+# Proposal
+
+Validate that each Spin component structurally conforms to (or targets) the expected world implied by the associated trigger type at build time rather than runtime.
+
+Prior art for performing a structural target of components against a world is [`wasm-tools component targets ...`](https://github.com/bytecodealliance/wasm-tools/blob/9340ed2466a50b4dbc580b13ba18a417dee91433/src/bin/wasm-tools/component.rs#L683) subcommand. We could potentially use [wac target predicate implemtation](https://github.com/bytecodealliance/wac/blob/4c96def294e6e779c52cfc5a93e05ed4c73ee60f/crates/wac-parser/src/resolution.rs#L2708) for a cleaner more actionable user diagnostic. Effectively, spin build could invoke the targets check using the trigger implied world and the source component.
+
+An "approximate" visual for the spin build experience we could enable:
+
+```
+❯ spin build -f examples/hello-world 
+Component "hello" is not a valid http component.
+
+Error: 
+    Expected component to either export `wasi:http/incoming-handler@0.2.0-rc-2023-10-18` or `fermyon:spin/inbound-http` but it exported neither    
+```
+
+# Open question: How does this work with custom triggers?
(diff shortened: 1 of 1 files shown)


docs/content/sips/0XX-spin-build-targets-check.md
```diff
@@ -0,0 +1,65 @@
...
+```
+❯ spin build -f examples/hello-world 
+Component "hello" is not a valid http component.
+
+Error: 
+    Expected component to either export `wasi:http/incoming-handler@0.2.0-rc-2023-10-18` or `fermyon:spin/inbound-http` but it exported neither    
+```
+
+# Open question: How does this work with custom triggers?
+One possible solution is to allow trigger extensions the abilitiy to hook into build time validation via the `TriggerExecutor` trait. For each `spin build`, the `spin-cli` could construct a `build` command to invoke each trigger with allowing the custom trigger an opportunity to enforce validation on the target of each component before `up`.
```
lann, line 65: Makes sense to me. Could add a standard e.g. `--trigger-validate-app` flag.

Somewhat of a tangent, but we might want to think about how to avoid future collisions with trigger-specific flags; maybe we should reserve e.g. `--trigger-*`?

docs/content/sips/0XX-spin-build-targets-check.md
```diff
@@ -0,0 +1,65 @@
...
+Prior art for performing a structural target of components against a world is [`wasm-tools component targets ...`](https://github.com/bytecodealliance/wasm-tools/blob/9340ed2466a50b4dbc580b13ba18a417dee91433/src/bin/wasm-tools/component.rs#L683) subcommand. We could potentially use [wac target predicate implemtation](https://github.com/bytecodealliance/wac/blob/4c96def294e6e779c52cfc5a93e05ed4c73ee60f/crates/wac-parser/src/resolution.rs#L2708) for a cleaner more actionable user diagnostic. Effectively, spin build could invoke the targets check using the trigger implied world and the source component.
+
+An "approximate" visual for the spin build experience we could enable:
+
+```
+❯ spin build -f examples/hello-world 
+Component "hello" is not a valid http component.
+
+Error: 
+    Expected component to either export `wasi:http/incoming-handler@0.2.0-rc-2023-10-18` or `fermyon:spin/inbound-http` but it exported neither
```
lann, line 61: One thing this specific example brings up is that we probably don't one one monolithic "make sure it implements world W or bail" since the http trigger accepts several different worlds. I'm guessing a more flexible "validate this app" feature would work well here, combined with a library that makes it easy to check for targeting individual worlds.

docs/content/sips/0XX-spin-build-targets-check.md
fibonacci1729: I agree, `--trigger-validate-app` (or something) is a better idea. Also +1 to reserving `--trigger-*` flags.

itowlson: I'm not sure asking the trigger plugin is a good direction to go.  It will solve the immediate "does my component match my trigger" question but that's not the real problem we're trying to lay foundations for.  To refer back to the long game, suppose a user is running Spin 4.0 with v3.5 of the cron trigger, but specifies they want to the app to run on Spn 3.31 with cron 2.2, and on SpinKube 3.2 (with whatever cron trigger that includes).  Asking the local trigger is not going to help us there.  Maybe I'm YAGNI-ing though.

docs/content/sips/0XX-spin-build-targets-check.md
rylev: Some pedantry: the http trigger accepts several different guest worlds, but it implements one world (the http trigger world). In other worlds, the host world is a union of the various guest worlds it accepts. One way to think about this issue is: what world does the guest implement, and is that world a sub-type (sub-world?) of the host world?

lann: I guess in the terminology of this SIP, "HTTP apps target at least one of several worlds."

itowlson:
I was thinking a bit more about the custom triggers thing, and wanted to take a moment to align our understandings.

For me, the main value of this work is kind of a technology proof for validating against a world _once we have a world_ (or, more broadly, a set of acceptable worlds).  That's a prerequisite for target environments, but there's a whole lot of additional work in defining what an environment is and how we understand it.  Again consider the "I want to target SpinKube 3.2" scenario.  SpinKube 3.2 isn't a world: it's a mapping of trigger types to worlds.  We will need to understand from some source what that mapping is.

The custom triggers discussion seems to cross over into the "understand the environment / mapping" problem, but in a way that is very specific to the "the version of Spin I am running here and now" environment.  Yes, we want to be able to validate for that local environment for sure.  But I'm wondering if we should punt on solving the mapping problem for that "local Spin" environment until we can consider it alongside other environments.

As I say, that's just my envisioned purpose and scope for this leg of the work.  It may be (it seems likely) that other folks have a different vision and scope, so I'd like to crisp up what that is and how it fits into the eventual goal.

fibonacci1729:
> But I'm wondering if we should punt on solving the mapping problem for that "local Spin" environment until we can consider it alongside other environments.

I would be fine with us making the decision to hold off on solving this for custom triggers (for now) until we firm up our longer term goals (those of which are still not entirely clear to me at the moment).

docs/content/sips/0XX-spin-build-targets-check.md
fibonacci1729: An alternative @lann idea is to be able to query custom triggers for the worlds they support..

radu-matei: @itowlson I'm not sure I follow your argument entirely.
Let me state where *I'm* coming from with the feature and see whether we're on the same page here.

The goal of the feature is to determine at build time whether a certain component conforms to a given _versioned_ world; in your counter-example, if the user specifies a version of the world that the local trigger plugin configured does not support, I would expect the check to fail (and ideally, there would be an error that plugins should be updated)?

If is _not_ the responsibility of trigger plugins, then I'm not sure how else you would be able to validate that, given Spin itself would have no knowledge of arbitrary worlds (outside of say HTTP and whatever else might be "default").

radu-matei: (of course I should have just read the thread until the end and would have seen your detailed comment below...)

radu-matei:
I would also be aligned to solving this for the set of known worlds and not tackle custom triggers and worlds in the first go.

itowlson:
@radu-matei Thanks.  I have a technology POC that uses the `wac` libraries to validate a component or module against the HTTP target world.  The trouble is that `wac` currently only validates matching _exports_, not matching _imports_ (see https://github.com/bytecodealliance/wac/issues/127).  Now there is an alternative, to use `wasm-tools component targets`, which _does_ verify imports but... which doesn't produce useful error messages on validation failure.

Now, we can and will invest in those upstream projects to improve this - but at the moment, @fibonacci1729 is on point for that, and he is also trying to land component dependencies in Spin.  Given that, and that component dependencies will affect how target validation works, the current proposal was for him to prioritise component dependencies, then resume the upstream and target worlds stuff.

@fibonacci1729 hope this is an accurate representation of the current thinking - please correct me if not!

itowlson:
@fibonacci1729 Is this superseded by #3217?  It seems like they cover the same feature but I may be misunderstanding!

fibonacci1729:
Superseded by #3217

PR #2566, opened 2024-06-15
Add support for workload identity in the Azure CosmosDB Key/Value impl

devigned:
This PR adds the ability for the Azure CosmosDB KV store implementation to use ambient authentication (managed identity, workflow identity, azure cli). The PR should not break existing users of the `key` authentication mechanism; it should only be additive. For more information about the Azure Rust SDK identity flows, check out: https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/README.md.

Here is the associated infrastructure and sample application to demo this identity flow: https://github.com/devigned/spin-workload-id.

I'm opening this now to start getting feedback. I (or the identity SDK) also likely have a bug related to the authentication scope being passed to Entra (previously Azure Active Directory). I will sort that issue out and notify via the PR.

This work is related to https://github.com/spinkube/spin-operator/issues/252.

If you have Azure CLI installed, you can test this locally doing the following:
```bash
# Create an Azure CosmosDB
az cosmosdb create -n cosmosdb-spin-test -g test --locations regionName=southcentralus
# Create a database in our account
az cosmosdb database create -a cosmosdb-spin-test -g test -n spin
# Create a container (table) in our database
az cosmosdb container create -a cosmosdb-spin-test -g test -n  -d spin -p "/id"
# Get your subscription ID
SUB_ID=$(az account show --query "id" -o tsv)
# Get your user principal ID
ID=$(az ad user show --id your_user@your_domain.com --query "id" -o tsv)
# Add the role assignment to enable data mutations for your user
az cosmosdb sql role assignment create -a cosmosdb-spin-test -g test \
  --role-definition-name "00000000-0000-0000-0000-000000000002" \
  -s "/subscriptions/${SUB_ID}/resourceGroups/test/providers/Microsoft.DocumentDB/databaseAccounts/cosmosdb-spin-test" \
  -p ${ID}
```

Your corresponding runtime config should look like the following. Note there is no `key` used.
```
[key_value_store.default]
type = "azure_cosmos"
account = "cosmosdb-spin-test"
database = "spin"
container = "keys-and-values"
```

### DIFF
--- a/crates/key-value-azure/Cargo.toml
+++ b/crates/key-value-azure/Cargo.toml
@@ -6,7 +6,8 @@ edition = { workspace = true }
 
 [dependencies]
 anyhow = "1"
-azure_data_cosmos = "0.11.0"
+azure_data_cosmos = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
+azure_identity = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
 futures = "0.3.28"
 serde = { version = "1.0", features = ["derive"] }
 spin-key-value = { path = "../key-value" }
--- a/crates/key-value-azure/src/lib.rs
+++ b/crates/key-value-azure/src/lib.rs
@@ -15,9 +15,73 @@ pub struct KeyValueAzureCosmos {
     client: CollectionClient,
 }
 
+/// Azure Cosmos Key / Value runtime config literal options for authentication
+#[derive(Clone, Debug)]
+pub struct KeyValueAzureCosmosRuntimeConfigOptions {
+    key: String,
+}
+
+impl KeyValueAzureCosmosRuntimeConfigOptions {
+    pub fn new(key: String) -> Self {
+        Self { key }
+    }
+}
+
+/// Azure Cosmos Key / Value enumeration for the possible authentication options
+#[derive(Clone, Debug)]
+pub enum KeyValueAzureCosmosAuthOptions {
+    /// Runtime Config values indicates the account and key have been specified directly
+    RuntimeConfigValues(KeyValueAzureCosmosRuntimeConfigOptions),
+    /// Environmental indicates that the environment variables of the process should be used to
+    /// create the TokenCredential for the Cosmos client. This will use the Azure Rust SDK's
+    /// DefaultCredentialChain to derive the TokenCredential based on what environment variables
+    /// have been set.
+    ///
+    /// Service Principal with client secret:
+    /// - `AZURE_TENANT_ID`: ID of the service principal's Azure tenant.
+    /// - `AZURE_CLIENT_ID`: the service principal's client ID.
+    /// - `AZURE_CLIENT_SECRET`: one of the service principal's secrets.
+    ///
+    /// Service Principal with certificate:
+    /// - `AZURE_TENANT_ID`: ID of the service principal's Azure tenant.
+    /// - `AZURE_CLIENT_ID`: the service principal's client ID.
+    /// - `AZURE_CLIENT_CERTIFICATE_PATH`: path to a PEM or PKCS12 certificate file including the private key.
+    /// - `AZURE_CLIENT_CERTIFICATE_PASSWORD`: (optional) password for the certificate file.
+    ///
+    /// Workload Identity (Kubernetes, injected by the Workload Identity mutating webhook):
+    /// - `AZURE_TENANT_ID`: ID of the service principal's Azure tenant.
+    /// - `AZURE_CLIENT_ID`: the service principal's client ID.
+    /// - `AZURE_FEDERATED_TOKEN_FILE`: TokenFilePath is the path of a file containing a Kubernetes service account token.
+    ///
+    /// Managed Identity (User Assigned or System Assigned identities):
+    /// - `AZURE_CLIENT_ID`: (optional) if using a user assigned identity, this will be the client ID of the identity.
+    ///
+    /// Azure CLI:
+    /// - `AZURE_TENANT_ID`: (optional) use a specific tenant via the Azure CLI.
+    ///
+    /// Common across each:
+    /// - `AZURE_AUTHORITY_HOST`: (optional) the host for the identity provider. For example, for Azure public cloud the host defaults to "https://login.microsoftonline.com".
+    /// See also: https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/README.md
+    Environmental,
+}
+
 impl KeyValueAzureCosmos {
-    pub fn new(key: String, account: String, database: String, container: String) -> Result<Self> {
-        let token = AuthorizationToken::primary_from_base64(&key).map_err(log_error)?;
+    pub fn new(
+        account: String,
+        database: String,
+        container: String,
+        auth_options: KeyValueAzureCosmosAuthOptions,
+    ) -> Result<Self> {
+        let token = match auth_options {
+            KeyValueAzureCosmosAuthOptions::RuntimeConfigValues(config) => {
+                AuthorizationToken::primary_key(config.key).map_err(log_error)?
+            }
+            KeyValueAzureCosmosAuthOptions::Environmental => {
+                AuthorizationToken::from_token_credential(
+                    azure_identity::create_default_credential()?,
+                )
+            }
+        };
         let cosmos_client = CosmosClient::new(account, token);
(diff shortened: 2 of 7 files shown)


devigned:
Also to note on this PR is that the changes will also allow users to run locally using the Azure CLI to authenticate to Azure CosmosDB. This could be a convenient way of testing / developing with Azure resources without needing to deploy to Azure.

devigned:
The issue linked (az/rust#1593) is what needs to be resolved & verified prior to completing this PR.

devigned:
Ok, with https://github.com/Azure/azure-sdk-for-rust/pull/1678 this works as expected. Currently, I am pointing at my fixed branch of the Azure SDK for Rust. Do we want to delay this getting merged until the Azure Rust SDK merges the fix?

devigned:
The fix to the Azure SDK for Rust has been merged and I've updated this PR to point at that the ref of that change in main. I believe this PR is ready for review.

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -122,19 +124,33 @@ impl RedisKeyValueStoreOpts {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct AzureCosmosConfig {
```
devigned, line 126: Should this struct contain an untagged enum (https://serde.rs/enum-representations.html#untagged) rather than defaulted values?

itowlson:
@devigned I think you have some clippies (and autoformats maybe?) that are unrelated to the changes.  I believe these are already picked up in #2569 so you'll probably need to rebase.

crates/trigger/src/runtime_config/key_value.rs
itowlson: Hmm, maybe.  It seems like a user should have `environmental_auth` _or_ `key` but not both.  But I'm not sure the two cases are different enough to justify the extra complexity and the weird unhelpful errors that untagged enums give us.  I'm actually wondering, is `environmental_auth` ever needed?  That is, is the presence of `key` enough to signal "use this key instead of the one in the environment"?  If so, would it make sense to make `key` an `Option<String>` and ditch `environmental_auth` altogether?

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -122,19 +124,33 @@ impl RedisKeyValueStoreOpts {
...
+    #[serde(default)]
+    environmental_auth: bool,
 }
 
 impl AzureCosmosConfig {
     pub fn build_store(&self) -> Result<Arc<dyn StoreManager>> {
+        let auth_options = if self.key.is_empty() || self.environmental_auth {
+            KeyValueAzureCosmosAuthOptions::Environmental
+        } else {
+            if self.key.is_empty() {
```
itowlson, line 141: I am not sure how this could happen.  Wouldn't this condition always go down the "true" clause of the `if`?

crates/key-value-azure/Cargo.toml
```diff
@@ -6,7 +6,8 @@ edition = { workspace = true }
 
 [dependencies]
 anyhow = "1"
-azure_data_cosmos = "0.11.0"
+azure_data_cosmos = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
+azure_identity = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
```
radu-matei, line 10: These are not yet part of a release, right?

crates/key-value-azure/src/lib.rs
```diff
@@ -15,9 +15,48 @@ pub struct KeyValueAzureCosmos {
...
+#[derive(Clone, Debug)]
+pub enum KeyValueAzureCosmosAuthOptions {
+    /// Runtime Config values indicates the account and key have been specified directly
+    RuntimeConfigValues(KeyValueAzureCosmosRuntimeConfigOptions),
+    /// Environmental indicates that the environment variables of the process should be used to
+    /// create the TokenCredential for the Cosmos client. This will use the Azure Rust SDK's
+    /// DefaultCredentialChain to derive the TokenCredential based on what environment variables
+    /// have been set.
+    /// See also: https://github.com/Azure/azure-sdk-for-rust/blob/main/sdk/identity/README.md
+    Environmental,
```
radu-matei, line 40: It would be very helpful to explicitly state the expected environment variables that should be set for this authentication mechanism to work.

crates/trigger/src/runtime_config/key_value.rs
```diff
@@ -122,19 +124,33 @@ impl RedisKeyValueStoreOpts {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct AzureCosmosConfig {
+    #[serde(default)]
     key: String,
     account: String,
     database: String,
     container: String,
+    #[serde(default)]
+    environmental_auth: bool,
```
radu-matei, line 133: If this is set to `true`, my assumption would be that the runtime would look for the default Azure environment variables and use them.

For this to be usable in SpinKube, ahead of the Spin version update we need to update the runtime config types in the operator to support it — https://github.com/spinkube/spin-operator/blob/main/internal/runtimeconfig/runtime_config.go

radu-matei:
This is looking great, @devigned! Thanks for the contribution!

crates/key-value-azure/Cargo.toml
devigned: They are not. Also, I'm not clear on when they would be.

crates/trigger/src/runtime_config/key_value.rs
devigned: I'm not sure that is the case for Workflow Identity since those env vars are set by the Workflow Identity mutating web hook for the pod specs with the appropriate label. SpinKube would need to be able to set the labels and the service account for the pod spec, so it would work with Workload Identity. 

For other auth types (say service principal w/ secret), that would likely need to have env vars specified on the SpinApp to then be translated to pod spec env vars.

itowlson:
@devigned What's the status of this?  I believe we're keen for it if you still have bandwidth - is it just the review nits and the lints?

devigned:
Sorry, I got pulled away. I'll address the feedback and get it updated today. Thank you for your patience.

crates/trigger/src/runtime_config/key_value.rs
devigned: Yes, 🤣 that was left over from moving some code around.

crates/trigger/src/runtime_config/key_value.rs
devigned: I've removed the `environmental_auth` field and am defaulting to env auth when `key` is not provided.

crates/key-value-azure/src/lib.rs
devigned: @radu-matei please take a look at the documentation when you have a moment.

devigned:
@itowlson I believe this is ready for review. I'm going to start work on the Key Vault integration to do the same thing.

crates/oci/src/client.rs
```diff
@@ -1141,15 +1141,13 @@ mod test {
                 .expect("should have version annotation")
         );
         assert!(
-            annotations
-                .get(oci_distribution::annotations::ORG_OPENCONTAINERS_IMAGE_DESCRIPTION)
-                .is_none(),
+            !annotations
```
itowlson, line 1144: This and the next diff seem unrelated to all things Azure.  Maybe your clippy is more recent than the one in CI?  (I also noticed some format changes on the OTel attributes which suggested a toolchain having new opinions about existing code.)

itowlson (approved):
LGTM - thanks David!  I really appreciated the detailed documentation on the environment variables.  There were a couple of format/clippy changes that seem unrelated to the functional changes, possibly due to different Rust versions?  That's not a blocker though.

itowlson:
@devigned You'll need to run `make update-cargo-locks` to get CI to pass.  (This happens because new/changed dependencies in the trigger crate need to feed through into the lockfile for the custom trigger example.)

crates/oci/src/client.rs
devigned: Happy to revert the attribute changes. Let me know. Pretty sure it was just RustRover doing its auto format thing.

As for the annotations, IIRC that was a clippy error with the Rust version I am using.

itowlson: You and your fancy "up-to-date" Rust versions.  Yeah I'm not too worried.  Thanks - I'll merge once Radu has had a chance to look.

itowlson:
Integration test failure seems unrelated (it's hitting one of my PRs too).

PR #2568, opened 2024-06-17
Override local data directory via env variable

### LINKED ISSUE #2563: Support overriding the "Spin data dir" via env var
The Spin "data dir" (e.g. `~/.local/share`) is used as the root for plugin and template state. We have a request (part of #2555) to make this customizable to aid integration with other development tooling.

The data dir is constructed here: https://github.com/fermyon/spin/blob/aedc2b8ccc65f41615fa7beb3e427843dd782114/crates/common/src/data_dir.rs#L7

As suggested in the linked discussion, we could add support for an environment variable to override this algorithm. Bike-shedding on the name of that variable, I suggest `SPIN_DATA_DIR`.

In addition to adding this support, a bit of related clean-up would be nice:

- Replace [`TEST_PLUGINS_DIRECTORY`](https://github.com/fermyon/spin/blob/8534cbec177b881d1f9ae4e4008a1f8e1c712cb3/crates/plugins/src/store.rs#L28) with this new env var
- Maybe rename `default_data_dir` to...`spin_data_dir` (?); there isn't anything really "default" about it, especially with the removal of `TEST_PLUGINS_DIRECTORY`

Separately, should we do the same for "cache" (used by badger, OCI) and "config" (OCI registry auth) dirs?

cc @jandubois, @itowlson

itowlson:
Fixes #2563 

Note this specifically addresses the _data_ directory.  It does not affect the cache directory.  If that's needed then I can chuck it into this PR.

cc @jandubois

### DIFF
--- a/crates/common/src/data_dir.rs
+++ b/crates/common/src/data_dir.rs
@@ -4,7 +4,10 @@ use anyhow::{anyhow, Result};
 use std::path::{Path, PathBuf};
 
 /// Return the default data directory for Spin
-pub fn default_data_dir() -> Result<PathBuf> {
+pub fn data_dir() -> Result<PathBuf> {
+    if let Ok(data_dir) = std::env::var("SPIN_DATA_DIR") {
+        return Ok(PathBuf::from(data_dir));
+    }
     if let Some(pkg_mgr_dir) = package_manager_data_dir() {
         return Ok(pkg_mgr_dir);
     }
--- a/crates/plugins/src/store.rs
+++ b/crates/plugins/src/store.rs
@@ -1,6 +1,6 @@
 use anyhow::{Context, Result};
 use flate2::read::GzDecoder;
-use spin_common::data_dir::default_data_dir;
+use spin_common::data_dir::data_dir;
 use std::{
     ffi::OsStr,
     fs::{self, File},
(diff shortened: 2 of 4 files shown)


lann (approved):

jandubois:
> Note this specifically addresses the _data_ directory.

Thanks a lot; this is exactly what I was hoping for.

> It does not affect the cache directory. If that's needed then I can chuck it into this PR.

Sorry for the late response, as this PR has now been merged.

It is not clear to me, what the cache directory is being used for. Does the cached content depend on the version of the `spin` app that is being used. Can 2 versions of spin share the same cache directory, or would a newer `spin` executable potentially update the cache with data that is not compatible with the older version?

If there are no possible conflicts (e.g. because content is versioned), then there is less reason to make the cache directory configurable.

The remaining reason would be that doing e.g. a Rancher Desktop "factory reset" or "uninstall" could also clear out the cache directory instead of leaving it behind.

itowlson:
@jandubois The cache directory is used for content-addressed downloads, specifically for components identified by URL or registry reference and for application assets when running from an app from an OCI registry.  It should not depend on Spin version.  (If we made changes that changed the layout, that could result in re-downloading things, but never in things from other versions "crossing over" if that makes sense.)

jandubois:
@itowlson Thanks, that means it is not really necessary to have separate cache directories then.

Out of curiosity: how large does the cache grow, and is there some mechanism to expire old content?

itowlson:
No, at the moment all that is down to the user.  It would be good to improve that, but it is not yet a priority.  (I'm sure it will _become_ a priority.)

PR #2574, opened 2024-06-18
fix: add error message when file path outside root

### LINKED ISSUE #2551: Referencing files outside of app root produces bad error messages
We intentionally limit file references to be within the "app root" (i.e. the directory containing `spin.toml`). If this constraint is violated the error chain is currently pretty unhelpful:

```
   1: Failed to load component `<component-id>`
   2: prefix not found
```

I believe the source of this error is the `strip_prefix` call here: https://github.com/fermyon/spin/blob/b8e4185597615fe5559ce2c473d5d7d612206690/crates/loader/src/local.rs#L419

me-diru:
Signed-off-by: Rohit Dandamudi <rohitdandamudi.1100@gmail.com>

fixes #2551

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -416,7 +416,10 @@ impl LocalLoader {
                 continue;
             }
 
-            let app_root_path = src.strip_prefix(&self.app_root)?;
+            let Ok(app_root_path) = src.strip_prefix(&self.app_root) else {
+                bail!("{pattern} cannot be mapped because it is outside the application directory. Files must be within the application directory.");
+            };
+
             if exclude_patterns
                 .iter()
                 .any(|pattern| pattern.matches_path(app_root_path))


crates/loader/src/local.rs
```diff
@@ -416,7 +416,12 @@ impl LocalLoader {
                 continue;
             }
 
-            let app_root_path = src.strip_prefix(&self.app_root)?;
+            let app_root_path = match src.strip_prefix(&self.app_root) {
+                Ok(path) => path,
+                Err(error) => {
+                    bail!("Invalid file path outside application root, error: {error}")
```
itowlson, line 422: A couple of things on this:

1. There's probably not much point including the `StripPrefixError` in the output.  We're translating the "prefix not found" into a more meaningful error: we don't need to show how we know.
2. We need to say which file or pattern is causing the problem.  The user might have several entries in the `files` section, and would appreciate guidance to which one.  This may be a bit tricky to get in its exact form, because it looks like it's already been globbed by this point, but even the glob pattern would help them narrow down the problem.
3. Rather than saying "Invalid file path" (it may well be _valid,_ we just don't like it), consider phrasing it in terms of what the rules are and why this doesn't fall within them, e.g. `"/home/ivan/catpictures" cannot be mapped because it is outside the application directory. Files must be within the application directory.`

itowlson (changes requested):

crates/loader/src/local.rs
```diff
@@ -416,7 +416,12 @@ impl LocalLoader {
                 continue;
             }
 
-            let app_root_path = src.strip_prefix(&self.app_root)?;
+            let app_root_path = match src.strip_prefix(&self.app_root) {
```
itowlson, line 419: A more concise way to express this in current versions of Rust is to use a let-pattern-match with an `else` clause, e.g.

```rust
let Ok(app_root_path) = src.strip_prefix(&self.app_root) else {
    bail!(...);
}
```

crates/loader/src/local.rs
me-diru: Thanks a bunch for the detailed review. I have updated the commit accordingly. I have included the pattern for the user to identify where it's failing.

itowlson (approved):
LGTM!

crates/loader/src/local.rs
```diff
@@ -416,7 +416,10 @@ impl LocalLoader {
                 continue;
             }
 
-            let app_root_path = src.strip_prefix(&self.app_root)?;
+            let Ok(app_root_path) = src.strip_prefix(&self.app_root) else {
+                bail!("{pattern} cannot be mapped because it is outside the application directory. Files must be within the application directory.");
```
itowlson, line 420: Oh, I just realised - preferred style is to use `quoted_path` (from the common crate) around paths destined for the UI.  This delimits the path from the surrounding text.

me-diru: for file path
> files = [{ source = "/home/ro", destination = "/catpics" }]

spin up gives the following error

> Error: Failed to load manifest from "spin.toml"
> Caused by:
>   0: Failed to load Spin app from "spin.toml"
>   1: Failed to load component `code-pls`
>   2: /home/ro/**/* cannot be mapped because it is outside the application directory. Files must be within the application directory.

I think the file path in error message gives reasonable information for the users to figure out where it went wrong :D

PR #2588, opened 2024-06-20
Bump wasm-pkg-loader to 0.4.1

fibonacci1729:


lann (approved):

crates/manifest/Cargo.toml
```diff
@@ -14,6 +14,7 @@ thiserror = "1"
 terminal = { path = "../terminal" }
 toml = { version = "0.8.0", features = ["preserve_order"] }
 url = "2.4.1"
+wasm-pkg-loader = { workspace = true }
```
itowlson, line 17: I am distinctly wary of taking a dependency on the package loader in the manifest crate.

lann: Could use https://docs.rs/wasm-pkg-common/latest/wasm_pkg_common/ instead; the relevant types are re-exported from that by wasm-pkg-loader.

itowlson: Ooh, that looks perfect

fibonacci1729: How distinctly wary of [wasm-pkg-common](https://docs.rs/wasm-pkg-common/0.4.1/wasm_pkg_common/all.html) would you be?

fibonacci1729: Oh dangit

itowlson: 🥷

itowlson (approved):

Issue #2449, opened 2024-04-11
[governance] Formalizing Spin's status as a multi-stakeholder project


tschneidereit:
Fermyon has always seen Spin as a multi-stakeholder project, which shouldn't be inextricably tied to one organization's goals. So far, this has been implicit: we don't have a formal governance model laying out how stakeholdership works.

We think that that needs to change, because increasingly there _are_ multiple stakeholders involved, and Spin is becoming a key dependency of other projects ([1](https://github.com/spinkube/), [2](https://github.com/containerd/runwasi), [3](https://github.com/Azure/aks-spin-plugin)), and used by [other organizations](https://unit.nginx.org/news/2024/fermyon-spin-rust-sdk/). I'm opening this issue as a way to track our progress in working through the details, but also to put a stake in the ground and be explicit about the intended outcome: Spin will officially become a multi-stakeholder project in a way that we at Fermyon will not be able to take back. We have no interest in doing so, and want to ensure that everyone can fully rely on that.

-----

On a personal note, I firmly believe that what we want to achieve with Spin is too big to go it alone, and if we at Fermyon myopically tried to tightly control the project, we'd just end up suffocating much of that potential. Many of us helped create open source organizations and are actively helping run them because we share the belief that this kind of foundational work can't be done by single stakeholders alone. Based on that belief we've already moved ([1](https://github.com/bytecodealliance/componentize-py), [2](https://github.com/bytecodealliance/StarlingMonkey)) or are in [the process of moving](https://www.fermyon.com/blog/announcing-spinkube-contribution-to-cncf) key projects into explicitly open governance contexts, and look forward to doing the same with Spin!

Mossaka:
This is a fantastic step forward for Spin to be a multi-stakeholder project! 👍 

As a maintainer of the `containerd-shim-spin` project, Spin is undoubtly the most critical part of what we build and I am all in on making sure Spin continue to thrive under a new governence model. I'd love to get more involved in shaping the governence model and looking forward to working with you all!

michelleN:
Since this issue was opened, we have created formal governance for the Spin project and have been accepted into the CNCF as a sandbox project. I'll close this issue for now but please re-open if there are gaps we still need to close.

PR #2593, opened 2024-06-21
add SIP for governance.md

michelleN:
This SIP includes a proposal for a `Governance.md` document which describes governance for the Spin project. This goal of this initial version of governance is to most closely mirror what we currently do, formally define and open maintainer-ship, and provide a process for iterating on governance further. It is not the goal of this proposal to fully resolve all governance matters but it does kick off the process.

related to #2449

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -47,7 +47,7 @@ The default decision making process is objection-free consensus. In other words
 
 Decision making scenarios should be promoted appropriately by the maintainer overseeing the issue:
 
-- tag maintainers in relevant issues / SIPs
+- tag maintainers in relevant issues and SIPs
 - send out mailing list reminders
 - communicate updates in meetings
 
@@ -82,27 +82,11 @@ The Spin Governance Committee will select a chair to set agendas, call meetings,
 
 ### Decision Making
 
-The default decision making process is objection-free consensus except for changes to the Governance of the project which requires a supermajority agreement of the committee. The process for making governance related changes is described below.
+The default decision making process is objection-free consensus. All governance related decisions require supermajority quorum. All substantial changes require a SIP. A substantial change is one that changes the project governance itself, instead of editorial changes providing clarifications, small bugfixes, etc.
 
 Decision making scenarios should be promoted appropriately by the member overseeing the issue.
 
-In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort,
-a member can call for a supermajority vote from the committee.
-
-### Changes to Governance
-
-The process for changes to the GOVERNANCE.md file is as follows:
-
-1. Decide whether the change is substantial in nature. A substantial change is one that changes the project governance itself, instead of editorial changes providing clarifications, small bugfixes, etc
-2. Then,
-    - For substantial changes:
-        i. Open a new SIP for the changes
-        ii. Follow the [normal SIP process](https://github.com/fermyon/spin/blob/main/docs/content/sips/index.md), requiring a supermajority for acceptance
-    - For editorial changes:
-        i. Open a PR with the changes to GOVERNANCE.md
-        ii. Land the PR after getting at least two positive reviews, and waiting for at least 3 workdays across all time zones
-
-_Note: PRs landed without SIPs can be challenged as substantial for up to 2 weeks afterwards, at which point they have to be backed out, and the proposed changes have to go through the SIP process instead._
+In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort, a committee member can call for a supermajority vote form the committee. If another member seconds the vote, the vote must take place.
 
 ## Glossary
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -37,24 +37,9 @@ Changes to project maintainers use the following:
 
 - A project maintainer may step down by emailing the mailing list. When a project maintainer steps down, they become an emeritus maintainer.
 - Project maintainers MUST remain active on the project. If they are unresponsive for > 3 months, they will lose project maintainer-ship, unless the remaining project maintainers of the given project and the Spin Governance Committee agree to extend the period to be greater than 3 months.
-- New maintainers must be nominated by existing maintainers. Maintainers are to discuss and agree in a private setting adding a new maintainer. Once a descision has been made, a maintainer may be added to the project via a pull request to the relevant MAINTAINERS.md file.
+- New maintainers MUST be nominated by existing maintainers. Maintainers are to discuss and agree in a private setting adding a new maintainer. Once a descision has been made, a maintainer may be added to the project via a pull request to the relevant MAINTAINERS.md file.
 - A maintainer may be removed for a [code of conduct](CODE_OF_CONDUCT.md) violation by the Spin Governance Committee. Code of conduct violations may be submitted to any member(s) on the Spin Governance Committee by email. See email information on MAINTAINERS.md.
 - When a project has no active maintainers, the maintainers of the [fermyon/spin Github repo](https://github.com/fermyon/spin) become responsible for it, and may archive the project, or find new maintainers
-
-### Decision Making
-
-The default decision making process is objection-free consensus. In other words, a decision is made when all decision makers have had time to consider the decision and do not raise any objections. Silence on any consensus decision is equivalent to non-objection. Explicit agreement may be stated at will.
-
-Decision making scenarios should be promoted appropriately by the maintainer overseeing the issue:
-
-- tag maintainers in relevant issues and SIPs
-- send out mailing list reminders
-- communicate updates in meetings
-
-In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort,
-a maintainer can call for a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote from the project maintainers for a repo on a decision. If quorum cannot be met for a decision, all members of the Spin Governance Committee are added to the relevant vote.
-
-If a decision impacts multiple repositories or requires a coordinated effort across multiple repositories and project maintainers are unable to reach a decision on their own for the relevant projects, a maintainer can call for a decision from the Spin Governance Committee.
 
 ## Spin Governance Committee
 
@@ -80,13 +65,19 @@ The Spin Governance Committee have to match the following criteria:
 
 The Spin Governance Committee will select a chair to set agendas, call meetings, and oversee the decision making process.
 
-### Decision Making
+## Decision Making
 
-The default decision making process is objection-free consensus. All governance related decisions require supermajority quorum. All substantial changes to governance require a SIP. A substantial change is one that changes the project governance itself, instead of editorial changes providing clarifications, small bugfixes, etc.
+The default decision making process is objection-free consensus. In other words, a decision is made when all decision makers have had time to consider the decision and do not raise any objections. Silence on any consensus decision is equivalent to non-objection. Explicit agreement may be stated at will.
 
-Decision making scenarios should be promoted appropriately by the member overseeing the issue.
+Decision making scenarios MUST be promoted appropriately by the maintainer or committee member overseeing the issue. All substantial changes in any part of the Spin project including for governance related changes require a SIP.
 
-In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort, a committee member can call for a supermajority vote form the committee. If another member seconds the vote, the vote must take place.
+All substantial changes to governance require a supermajority quorum on the governance committee.
+
+In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort between project maintainers on a project level decision, a project maintainer can call for a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote from the project maintainers for a repo on a decision. If quorum cannot be met for a decision, all members of the Spin Governance Committee are added to the relevant vote.
+
+If a decision impacts multiple repositories or requires a coordinated effort across multiple repositories and project maintainers are unable to reach a decision on their own for the relevant projects, a maintainer can call for a decision from the Spin Governance Committee.
+
+In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort on the governance committee, a committee member can call for a supermajority vote form the committee. If another member seconds the vote, the vote MUST take place.
 
 ## Glossary
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -82,7 +82,7 @@ The Spin Governance Committee will select a chair to set agendas, call meetings,
 
 ### Decision Making
 
-The default decision making process is objection-free consensus. All governance related decisions require supermajority quorum. All substantial changes require a SIP. A substantial change is one that changes the project governance itself, instead of editorial changes providing clarifications, small bugfixes, etc.
+The default decision making process is objection-free consensus. All governance related decisions require supermajority quorum. All substantial changes to governance require a SIP. A substantial change is one that changes the project governance itself, instead of editorial changes providing clarifications, small bugfixes, etc.
 
 Decision making scenarios should be promoted appropriately by the member overseeing the issue.
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,0 +1,111 @@
+## Overview
+
+This document describes the governance of the Spin project. 
+
+The Spin project consists of several codebases with different release cycles. These codebases include:
+
+- Core Spin:
+    - [Spin](https://github.com/fermyon/spin)
+- SDKs:
+    - [Spin Python SDK](https://github.com/fermyon/spin-python-sdk)
+    - [Spin JavaScript SDK](https://github.com/fermyon/spin-js-sdk)
+    - [Spin Rust SDK](https://github.com/fermyon/spin-rust-sdk)
+    - [Spin .NET SDK](https://github.com/fermyon/spin-dotnet-sdk)
+    - [Spin Nim SDK](https://github.com/fermyon/spin-nim-sdk)
+- Plugins:
+    - [Spin Plugins Index Repository](https://github.com/fermyon/spin-plugins)
+    - [Fermyon Platform Plugin](https://github.com/fermyon/platform-plugin)
+    - [Spin Test Plugin](https://github.com/fermyon/spin-test)
+    - [Cloud Plugin](https://github.com/fermyon/cloud-plugin)
+    - [Spin Cloud GPU Plugin](https://github.com/fermyon/spin-cloud-gpu)
+- Triggers:
+    - [Spin Command Trigger](https://github.com/fermyon/spin-trigger-command)
+    - [Spin SQS Trigger](https://github.com/fermyon/spin-trigger-sqs)
+    - [Spin Cron Trigger](https://github.com/fermyon/spin-trigger-cron)
+
+Each repository is subject to the same overall governance model, but has different teams of people (“maintainers”) with permissions and access to the repository. This is meant to increase diversity of maintainers in the Spin project and also increases the velocity of code changes. Major changes and features to the project including additions to the repository list above are to be proposed through the [Spin Improvement Proposal](docs/content/sips/index.md) process.
+
+## Spin Project Maintainers
+
+Project maintainers are responsible for activities surrounding the development and release cycle of a specific repository they own. Technical decisions for code resides with the project maintainers, unless there is a decision related to cross maintainer groups that cannot be resolved by those groups. Those cases can be escalated to the Spin Governance Committee (discussed below).
+
+Project maintainers do not need to be software developers. No explicit role is placed upon them, and they can be anyone appropriate for the work being produced.
+
+Project maintainers and emeritus maintainers for each project should be outlined in the MAINTAINERS.md file in the corresponding GitHub repository.
+
+Changes to project maintainers use the following:
+
+- A project maintainer may step down by emailing the mailing list. When a project maintainer steps down, they become an emeritus maintainer.
+- Project maintainers MUST remain active on the project. If they are unresponsive for > 3 months, they will lose project maintainer-ship, unless the remaining project maintainers of the given project and the Spin Governance Committee agree to extend the period to be greater than 3 months.
+- New maintainers must be nominated by existing maintainers. Maintainers are to discuss and agree in a private setting adding a new maintainer. Once a descision has been made, a maintainer may be added to the project via a pull request to the relevant MAINTAINERS.md file.
+- A maintainer may be removed for a [code of conduct](CODE_OF_CONDUCT.md) violation by the Spin Governance Committee. Code of conduct violations may be submitted to any member(s) on the Spin Governance Committee by email. See email information on MAINTAINERS.md.
+- When a project has no active maintainers, the maintainers of the [fermyon/spin Github repo](https://github.com/fermyon/spin) become responsible for it, and may archive the project, or find new maintainers
+
+### Decision Making
+
+The default decision making process is objection-free consensus. In other words, a decision is made when all decision makers have had time to consider the decision and do not raise any objections. Silence on any consensus decision is equivalent to non-objection. Explicit agreement may be stated at will.
+
+Decision making scenarios should be promoted appropriately by the maintainer overseeing the issue:
+
+- tag maintainers in relevant issues / SIPs
+- send out mailing list reminders
+- communicate updates in meetings
+
+In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort,
+a maintainer can call for a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote from the project maintainers for a repo on a decision. If quorum cannot be met for a decision, all members of the Spin Governance Committee are added to the relevant vote.
+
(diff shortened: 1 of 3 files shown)


radu-matei (approved):
Content looks great, thank you!

Do you intend that the markdown block from the SIP be used historically to describe the initial intention of the SIP?

(trying to understand if it would make more sense to have the separate governance.md file as part of this PR instead — not blocking, though).


LGTM

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,127 @@
...
+
+Created: June 18, 2024
+
+## Background
+
+Spin is a project now used by many organizations and individuals across the software ecosystem. In order for the Spin community to continue to rely on, build on, and integrate with Spin, there should be a public collective understanding of how the Spin project is governed and how to iterate on governance.
+
+## Proposal
+
+This SIP is a proposal to add the following in a Governance.md file to the root of the Spin repository. This governance structure aims to describe transparently how we've been doing things so far. It formally opens up the project for maintainers from multiple organizations. It also includes some future directives for working toward a proper multi stakeholder governed project.
```
rylev, line 19: I think we should expand this a bit to talk about how this governance model is explicitly temporary and meant as a bootstrapping process. In the governance document we say the following:

> In addition to the responsibilities listed above, [the Spin Governance Committee] is also responsible for bootstrapping a multi-stakeholder steering committee of 5-7 people to govern the project. The group is responsible for creating a steering committee Charter and shaping and executing the processes around selecting committee members. Once a steering committee is in place, the Spin Governance Committee will be dismantled and replaced with the Spin Steering Committee. The Spin Steering Committee will be responsible for bootstrapping a Code of Conduct Committee.

I think we're burying the lead by not stating the explicit goal of progressing the status quo to a new governance structure. Something like this could be added here in the introduction:

> The current structure outlined in the Governance.md document is established as a temporary bootstrapping phase on the road to a more permanent, multi-stakeholder model. The Spin Governance Committee established in this document should eventually be dismantled and replaced with a smaller, more well defined group only tasked with governing the Spin project as a whole (a "Steering Committee"). This group will in turn establish other committees for handling more specific governance concerns (e.g., Project moderation and CoC enforcement).

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,127 @@
...
+- Making decisions about project licensing, contribution requirements, and other legal matters
+- Resolving escalated project decisions when responsible project maintainers are blocked
+- Managing the Spin brand
+- Managing access to Spin assets such as source repositories, hosting, project calendars
+- Handling code of conduct violations for any repository under the Spin project
+- Deciding what sub-groups and repositories are part of the Spin project
+- Overseeing the resolution and disclosure of security issues
+- Managing financial decisions related to the project
+
+In addition to the responsibilities listed above, this group is also responsible for bootstrapping a multi-stakeholder steering committee of 5-7 people to govern the project. The group is responsible for creating a steering committee Charter and shaping and executing the processes around selecting committee members. Once a steering committee is in place, the Spin Governance Committee will be dismantled and replaced with the Spin Steering Committee. The Spin Steering Committee will be responsible for bootstrapping a Code of Conduct Committee.
```
rylev, line 93: It'd be nice to be a wee bit less perspective as to exactly what the steering committee has to do and instead nudge it in the right direction while still explicitly giving it the freedom to do things as it sees fit. 

```suggestion
In addition to the responsibilities listed above, this group is also responsible for bootstrapping a multi-stakeholder steering committee of 5-7 people to govern the project. The group is responsible for creating a steering committee Charter and shaping and executing the processes around selecting committee members. Once a steering committee is in place, the Spin Governance Committee will be dismantled and replaced with the Spin Steering Committee. The Spin Steering Committee should then establish additional governance structures as it sees fit (e.g., a Code of Conduct Committee for project moderation).
```

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,127 @@
...
+
+The Spin Governance Committee have to match the following criteria:
+
+- Spin Governance Committee MUST remain active on the project. If they are unresponsive for > 3 months, they will lose membership, unless a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) of the other governance members agrees to extend the period to be greater than 3 months
```
rylev, line 97: I believe this is what was meant (right now it sort of reads like we're talking about the committee as a whole instead of requirements placed on the individual members). 

```suggestion
- Spin Governance Committee members MUST remain active on the project. If they are unresponsive for > 3 months, they will lose membership, unless a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) of the other governance members agrees to extend the period to be greater than 3 months
```

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,127 @@
...
+
+The Spin Governance Committee will select a chair to set agendas, call meetings, and oversee voting.
```
rylev, line 99: Nit: voting is only a part of the decision making process:
```suggestion
The Spin Governance Committee will select a chair to set agendas, call meetings, and oversee the decision making process.
```

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,127 @@
...
+Decision making scenarios should be promoted appropriately by the member overseeing the issue.
+
+When a consensus cannot be found, a member can call for a supermajority vote from the committee.
+
+### Changes to Governance.md
+
+The process for changes to the Governance.md file is as follows:
+
+1) Open an issue
+2) Open a pull request to this SIP describing changes and the Governance.md file with proposed changes.
```
rylev, line 114: It feels strange that future changes would not require a SIP. I could understand small clarifications that don't obviously materially change the intent of the governance.md doc being done through PR, but IMO any changes of substance should go through the SIP process.

tschneidereit (approved):
This looks great! I left some comments, in particular around decision-making, but nothing really fundamental

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,127 @@
...
+
+Changes to project maintainers use the following:
+
+- A project maintainer may step down by emailing the mailing list. When a project maintainer steps down, they become an emeritus maintainer.
+- Project maintainers MUST remain active on the project. If they are unresponsive for > 3 months, they will lose project maintainer-ship, unless a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) (two-thirds majority) of the project maintainers of the given project agrees to extend the period to be greater than 3 months. If two-thirds majority is not possible, the Spin Governance Committee members will be included in the group of people voting on this decision in addition to the project maintainers of the project.
+- New maintainers can be added to a project by a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote of the project’s existing maintainers. New maintainers must be nominated by existing maintainers. If quorum is not met, the Spin Govenance Committee members will be included in this vote.
```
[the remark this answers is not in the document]
michelleN: I think my only reason for this is that voting and specifically voting in private allows current maintainers to raise concerns more readily.

docs/content/sips/019-governance.md
michelleN: That makes sense to me. Thanks @tschneidereit. Updated with this process.

docs/content/sips/019-governance.md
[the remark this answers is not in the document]
rylev: The point about keeping the conversation private makes sense though we should explicitly note that in the text. However, I don't see objection-free consensus and privacy as being mutually exclusive. Objection-free consensus is about ensuring that there isn't a culture of decision making where there are winners and losers - if possible everyone should feel at worst that their concerns were heard and not blocking objections exist.

GOVERNANCE.md
```diff
@@ -0,0 +1,114 @@
...
+    - https://github.com/fermyon/spin-cloud-gpu
+    - https://github.com/fermyon/spin-plugins
+    - https://github.com/fermyon/platform-plugin
+- Triggers:
+    - https://github.com/fermyon/spin-trigger-command
+    - https://github.com/fermyon/spin-trigger-sqs
+    - https://github.com/fermyon/spin-trigger-cron
+    
+
+Each repository is subject to the same overall governance model, but has different teams of people (“maintainers”) with permissions and access to the repository. This is meant to increase diversity of maintainers in the Spin project and also increases the velocity of code changes. Major changes and features to the project are to be proposed through the [Spin Improvement Proposal](docs/content/sips/index.md) process.
```
rylev, line 26: It feels right to me to be very explicit about this.

```suggestion
Each repository is subject to the same overall governance model, but has different teams of people (“maintainers”) with permissions and access to the repository. This is meant to increase diversity of maintainers in the Spin project and also increases the velocity of code changes. Major changes and features to the project (including changes to project governance) are to be proposed through the [Spin Improvement Proposal](docs/content/sips/index.md) process.
```

GOVERNANCE.md
```diff
@@ -0,0 +1,114 @@
...
+        ii. Follow the [normal SIP process](https://github.com/fermyon/spin/blob/main/docs/content/sips/index.md), requiring a supermajority for acceptance
+    - For editorial changes:
+        i. Open a PR with the changes to GOVERNANCE.md
+        ii. Land the PR after getting at least two positive reviews, and waiting for at least 3 workdays across all time zones
+
+_Note: PRs landed without SIPs can be challenged as substantial for up to 2 weeks afterwards, at which point they have to be backed out, and the proposed changes have to go through the SIP process instead._
+
+1) Open an issue
+2) Open a pull request to this SIP describing changes and the GOVERNANCE.md file with proposed changes. 
+3) Spin Govenance Committee will review pull request. Pull request may be merged when a supermajority of Spin Governance Committee has approved the pull request.
```
rylev, line 107: I think this needs to be removed, right?

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Background
+
+Spin is a project now used by many organizations and individuals across the software ecosystem. In order for the Spin community to continue to rely on, build on, and integrate with Spin, there should be a public collective understanding of how the Spin project is governed and how to iterate on governance.
+
+## Proposal
+
+This SIP is a proposal to add the following in a Governance.md file to the root of the Spin repository. This governance structure aims to describe transparently how we've been doing things so far. It formally opens up the project for maintainers from multiple organizations. It also includes some future directives for working toward a proper multi stakeholder governed project.
+
+```Governance.md
```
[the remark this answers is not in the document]
rylev: I think this can be removed now that it's in the dedicated governance.md file, right?

rylev (commented):
I'm pretty happy with how this is turning out - thanks for your great work! The last thing I'd need before giving my personal ✅ is some resolution to the voting vs consensus question. I would personally like to always see consensus as the preferred method for any and all decisions, and voting only used in truly exceptional circumstances where consensus cannot be reached.

GOVERNANCE.md
```diff
@@ -0,0 +1,114 @@
...
+- Project maintainers MUST remain active on the project. If they are unresponsive for > 3 months, they will lose project maintainer-ship, unless a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) (two-thirds majority) of the project maintainers of the given project agrees to extend the period to be greater than 3 months. If two-thirds majority is not possible, the Spin Governance Committee members will be included in the group of people voting on this decision in addition to the project maintainers of the project.
+- New maintainers can be added to a project by a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote of the project’s existing maintainers. New maintainers must be nominated by existing maintainers. If quorum is not met, the Spin Govenance Committee members will be included in this vote.
+- A maintainer may be removed for a [code of conduct](../../../CODE_OF_CONDUCT.md) violation via a supermajority vote of the Spin Governance Committee. Code of conduct violations may be submitted to any member(s) on the Spin Governance Committee by email. See email information on MAINTAINERS.md.
```
adamreese, line 41:
```suggestion
- A maintainer may be removed for a [code of conduct](CODE_OF_CONDUCT.md) violation via a supermajority vote of the Spin Governance Committee. Code of conduct violations may be submitted to any member(s) on the Spin Governance Committee by email. See email information on MAINTAINERS.md.
```

michelleN: updated location of GOVERNANCE.md so that should cover this issue. Thanks!

GOVERNANCE.md
michelleN: updated

michelleN:
@radu-matei @rylev I've updated the PR to include the GOVERNANCE.md file. I've also updated the SIP so it no longer contains the the governance document rather describes some of the important aspects we've discussed around governance that are reflected in the document.

@rylev I've removed most references to voting and opted for consensus across the board.

GOVERNANCE.md
```diff
@@ -0,0 +1,111 @@
...
+Decision making scenarios should be promoted appropriately by the maintainer overseeing the issue:
+
+- tag maintainers in relevant issues / SIPs
+- send out mailing list reminders
+- communicate updates in meetings
+
+In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort,
+a maintainer can call for a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote from the project maintainers for a repo on a decision. If quorum cannot be met for a decision, all members of the Spin Governance Committee are added to the relevant vote.
+
+If a decision impacts multiple repositories or requires a coordinated effort across multiple repositories and project maintainers are unable to reach a decision on their own for the relevant projects, a maintainer can call for a supermajority vote from the Spin Governance Committee.
```
rylev, line 57: This seems to imply that if a decision needs to be escalated from project maintainers to the Spin Governance Commitee, that the decision making process is automatically super majority vote which seems wrong to me. I would imagine the Spin Governance Committee would just use the default decision making process (i.e., objection free consensus) and only vote in extreme cases where consensus cannot be reached.


```suggestion
If a decision impacts multiple repositories or requires a coordinated effort across multiple repositories and project maintainers are unable to reach a decision on their own for the relevant projects, a maintainer can call for the Spin Governance Committee to make the decision.
```

rylev (changes requested):
@michelleN I'm a bit confused - I was assuming that the SIP would simply reference GOVERNANCE.md for much of the content. In other words, I had assumed the SIP would say more or less "see this commit of the GOVERNANCE.md file for how the SIP is proposing governance will work", and the SIP only would only provide additional information/context that isn't appropriate for the GOVERNANCE.md file.

However, it seems that the content of GOVERNANCE.md is duplicated across the SIP and GOVERNANCE.md. Unfortunately, it's not a 1 to 1 copy and there appears to be contradictions between the two documents. For example, on decision making, the GOVERNANCE.md file provides an exception to when objection free consensus is used (for changes to governance), while the SIP does not. 

Let me know if I'm missing something here, but as it stands, it makes reviewing difficult as the proposal is split, copied, and slightly modified between two places.

michelleN:
@rylev I can see how that might be confusing. I've updated the SIP so that it no longer contains the additional sections. Let me know if you have other suggestions.

docs/content/sips/019-governance.md
```diff
@@ -0,0 +1,21 @@
...
+
+Created: June 18, 2024
+
+## Background
+
+Spin is a project now used by many organizations and individuals across the software ecosystem. In order for the Spin community to continue to rely on, build on, and integrate with Spin, there should be a public collective understanding of how the Spin project is governed and how to iterate on governance.
+
+## Proposal
+
+This SIP is a proposal to add a GOVERNANCE.md file to the root of the Spin repository. The governance structure described in the corresponding GOVERNANCE.md file aims to describe transparently a process close to how we've been doing things so far in the project and how to make changes. It formally opens up the project for maintainers from multiple organizations, defines maintainership roles, describes decision making processes and more.
```
rylev, line 19: We should try to remember to link to the current version of `GOVERNANCE.md` so that it's easy in the future to see the exact version of the governance.md that was agreed to by this SIP (without having to crawl GitHub). Of course, we can't link to the exact version until we're sure it's final.

GOVERNANCE.md
rylev: @michelleN this was marked as resolved but nothing changed in the text. Could I get an explanation of why my concern isn't relevant here?

GOVERNANCE.md
```diff
@@ -0,0 +1,111 @@
...
+
+The Spin Governance Committee have to match the following criteria:
+
+- Spin Governance Committee members MUST remain active on the project. If they are unresponsive for > 3 months, they will lose membership, unless the remaining members of the committee agree to extend the period to be greater than 3 months
+
+The Spin Governance Committee will select a chair to set agendas, call meetings, and oversee the decision making process.
+
+### Decision Making
+
+The default decision making process is objection-free consensus except for changes to the Governance of the project which requires a supermajority agreement of the committee. The process for making governance related changes is described below.
```
rylev, line 85: Still not sure why changes to the Governance of the project use a vote instead of the normal decision making process. If there's a good reason for this can we document it here?

michelleN: I kept this in here because I would want to ensure a supermajority of the governance committee has explicitly reviewed and thought through any changes to the governance document. I'm open to changing it but I'd like to hear some more thoughts on the matter.

cc/ @endocrimes @tschneidereit @radu-matei @adamreese

michelleN:
>@michelleN this was marked as resolved but nothing changed in the text. Could I get an explanation of why my concern isn't relevant here??

I was not overlooking a concern. I had taken your suggestion but I think I must have fumbled a rebase. I appreciate your input.

GOVERNANCE.md
endocrimes: > My straw-man proposal would be for non-governance decisions to require a simple majority quorum and for governance decisions to require super-majority quorum. I think this would address the concern while still favoring consensus over "winner takes all" voting.

That seems like a reasonable proposal actually - I really dislike unqualified objection-free consensus for a whole bunch of reasons, so even if it's the goal, big +1 to defining requirements that give a clear path to forward progress.

michelleN: >My straw-man proposal would be for non-governance decisions to require a simple majority quorum and for governance decisions to require super-majority quorum. I think this would address the concern while still favoring consensus over "winner takes all" voting.

@rylev could you describe the process around requiring quorum to make changes to governance? Maybe compare that with how we've currently outlined the process for the changes to governance currently?

rylev: The top-level concern I have is the use of a "winner takes all" voting process for governance changes as opposed to the consensus seeking process used for all other decisions. The concern you had was around ensuring that a sufficient number of voices take place in the decision making process, since a strict reading of the consensus process could lead to situations where, for example, only 2 people out of N number of members of the Spin Governance Committee actually give input and work towards consensus (since silence == consensus). This is a very legitimate concern. We want to ensure that decisions not only represent some sort of consensus but that consensus is actively shared among a large enough percentage of the Spin Governance Committee. In other words, it's not enough that Bob and Alice agree while everyone else was on vacation - if not enough people participate in building consensus, then we should not consider that actual consensus.

This is where the idea of quorum comes in. We no longer say that consensus among *any* subset of the Spin Governance Committee is sufficient but rather that consensus must *actively* be reached by a certain percentage of the Spin Governance Committee.

The specific change here is to no longer strictly say silence == consensus. In order for decision to be made, some % of Spin Governance Committee members must actively give their consensus by stating they have no blocking objections.

How does this differ from voting? Quorum is about participation percentage where as voting is about a decision making threshold. In other words, for a decision to be made by consensus with 50% quorum requirement, 50% of potential decision makers need to actively take place in the process, but 100% of those who participate need to have no objections. This differs from a simple majority vote where as long as 50% of those who participate agree, it doesn't matter if other participants vehemently object, the vote still passes. Consensus seeks to address concerns and ensure that at the very least everyone can live with the decision.

I'd like to also make clear one other distinction between voting and consensus that isn't explicit in the explanation above. Consensus asks "can you live with this decision?" while voting asks "do you actively support this decision?". In voting a decision is made by those who actively agree with a decision while in consensus a decision is made by everyone at the very least not blocking a decision. Consensus ensures everyone's opinions are heard whereas in voting once it's clear enough people agree, there is zero incentive to listen to those who don't actively agree.

So, the original comment I left is all about consistency around the decision making process. I propose we use the same decision making process for all decisions in the project including changes to governance: consensus. The only difference between governance related decisions and other decisions is the quorum requirement: i.e., how many people must actively state they have no objections. In my straw-man proposal the quorum figure is 50% for regular decisions and 66.66% for governance decisions. Again, 100% of those who participate though must have no objections.  

Sorry for the longwinded answer. I hope this clarifies things, but let me know if things are still not clear.

michelleN: Thanks thats a really helpful answer.

michelleN:
I've updated this doc most recently to include:
- all governance related changes require supermajority quorum (rather than the voting structure outlined previously)
- a call for a vote in the governance committee (in the extreme case) requires a second for the vote to take place

GOVERNANCE.md
```diff
@@ -0,0 +1,95 @@
...
+
+The Spin Governance Committee have to match the following criteria:
+
+- Spin Governance Committee members MUST remain active on the project. If they are unresponsive for > 3 months, they will lose membership, unless the remaining members of the committee agree to extend the period to be greater than 3 months
+
+The Spin Governance Committee will select a chair to set agendas, call meetings, and oversee the decision making process.
+
+### Decision Making
+
+The default decision making process is objection-free consensus. All governance related decisions require supermajority quorum. All substantial changes require a SIP. A substantial change is one that changes the project governance itself, instead of editorial changes providing clarifications, small bugfixes, etc.
```
rylev, line 85: > All substantial changes require a SIP. A substantial change is one that changes the project governance itself, instead of editorial changes providing clarifications, small bugfixes, etc.

This doesn't read quite right to me. As it stands now, it's not clear to me whether "All substantial changes require a SIP." is saying all substantial changes to *governance* require SIP or all substantial changes to anything in the Project require a SIP. 

I can see this stating either one (or both). Whatever one we mean, can we rephrase this to be less ambiguous?

GOVERNANCE.md
```diff
@@ -0,0 +1,95 @@
...
+- Project maintainers MUST remain active on the project. If they are unresponsive for > 3 months, they will lose project maintainer-ship, unless the remaining project maintainers of the given project and the Spin Governance Committee agree to extend the period to be greater than 3 months.
+- New maintainers must be nominated by existing maintainers. Maintainers are to discuss and agree in a private setting adding a new maintainer. Once a descision has been made, a maintainer may be added to the project via a pull request to the relevant MAINTAINERS.md file.
+- A maintainer may be removed for a [code of conduct](CODE_OF_CONDUCT.md) violation by the Spin Governance Committee. Code of conduct violations may be submitted to any member(s) on the Spin Governance Committee by email. See email information on MAINTAINERS.md.
+- When a project has no active maintainers, the maintainers of the [fermyon/spin Github repo](https://github.com/fermyon/spin) become responsible for it, and may archive the project, or find new maintainers
+
+### Decision Making
```
rylev, line 44: I probably sound like a broken record so apologies, but it seems the decision making process for the Project maintainers differs from the decision making process of the Spin governance committee in the following ways:

* Silence in Project maintainers decision making is equivalent to silence whereas for Spin governance committee it's not stated what silence means.
* When decisions require a SIP is not specified for Project maintainers but it is for the Spin governance committee
* quorum rules are set for some decisions for the Spin governance committee but not for Project maintainers.

It's not clear to me why these differences in decision making process exist. Perhaps we can only have one section on decision making in the document, and if there is an explicit difference between Spin governance committee and Project maintainers we just call it out there? Having two difference sections on decision making makes it very hard to understand which differences are intentional and which is just due to the sections being worded slightly differently.

michelleN: Does this read better to you (collapsed decision making section)?

```
## Decision Making

The default decision making process is objection-free consensus. In other words, a decision is made when all decision makers have had time to consider the decision and do not raise any objections. Silence on any consensus decision is equivalent to non-objection. Explicit agreement may be stated at will.

Decision making scenarios should be promoted appropriately by the maintainer or committee member overseeing the issue. All substantial changes in any part of the project and for governance related changes require a SIP.

All substantial changes to governance require a supermajority quorum on the governance committee.

In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort between project maintainers on a project level decision, a project maintainer can call for a [supermajority](https://en.wikipedia.org/wiki/Supermajority#Two-thirds_vote) vote from the project maintainers for a repo on a decision. If quorum cannot be met for a decision, all members of the Spin Governance Committee are added to the relevant vote.

If a decision impacts multiple repositories or requires a coordinated effort across multiple repositories and project maintainers are unable to reach a decision on their own for the relevant projects, a maintainer can call for a decision from the Spin Governance Committee.

In the extreme case that objection-free consensus cannot be reached after a reasonable amount of time and effort on the governance committee, a committee member can call for a supermajority vote form the committee. If another member seconds the vote, the vote must take place.
```

rylev: Looks really good! I just have one nit (chaning a "should" to a "must"):

> Decision making scenarios must be promoted appropriately by the maintainer[...]

michelleN: updated

rylev (approved):
Awesome work! Thanks so much for listening to all the feedback I had 😄

GOVERNANCE.md
endocrimes: Chiming in with a that change LGTM too

michelleN:
updated with your notes @kapilt https://github.com/fermyon/spin/pull/2745

PR #2596, opened 2024-06-24
add support for client certs

rajatjindal:
this PR attempts to add support for client certs based auth. As a summary following are the changes done in this PR:

- copies the `default_send_request_handler` function from `wasmtime::wasi_http` crate to spin and modifies it to support client-cert-auth and custom-root-ca.
- adds `client_tls` as a new `runtime_config` option 
- parses the runtime config and store it indexed by `component-id -> host:port` (TBD: do we consider default 443 port here if none provided)
- uses the client tls config when making outbound request

Known Limitation:

As of now, it will only work when the Outbound request is made from with-in an http-trigger. The other triggers (e.g. redis-trigger) uses a different code path and that has not been changed in this PR).

Example Runtime config:

```
[[client_tls]]
hosts = ["localhost:6551"]
component_ids = ["app"]
custom_root_ca_file = "certs/custom_root_ca"
cert_chain_file = "certs/cert_chain"
private_key_file = "certs/private_key"
```

I will attempt to add some runtime tests for these changes, but I wanted to submit the PR to be able to collect some initial feedback. 

Thank you


crates/trigger-http/src/lib.rs
```diff
@@ -583,6 +586,8 @@ struct ChainedRequestHandler {
 pub struct HttpRuntimeData {
     origin: Option<String>,
     chained_handler: Option<ChainedRequestHandler>,
+    /// If provided, these options used for client cert auth
```
rylev, line 589: nit: can you say in the comment what the `String` in the hashmap is for:

```suggestion
    /// Optional mapping of authority and TLS options 
```

crates/trigger-http/src/lib.rs
```diff
@@ -1000,3 +1007,212 @@ mod tests {
         assert!(req.headers().get("Host").is_some());
     }
 }
+
+/// Translate a [`hyper::Error`] to a wasi-http `ErrorCode` in the context of a request.
+pub fn hyper_request_error(err: hyper::Error) -> ErrorCode {
```
rylev, line 1012: nit: can you move `hyper_request_error` and `dns_error` down to the bottom of the file so that `default_send_request_handler` is further up in the file?

crates/trigger-http/src/tls.rs
```diff
@@ -19,28 +19,118 @@ impl TlsConfig {
...
+    .collect::<io::Result<Vec<rustls_pki_types::CertificateDer<'static>>>>()
 }
 
-// Loads private key from file.
-fn load_keys(path: impl AsRef<Path>) -> io::Result<Vec<rustls::PrivateKey>> {
-    pkcs8_private_keys(&mut io::BufReader::new(fs::File::open(path)?))
-        .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid key"))
-        .map(|mut keys| keys.drain(..).map(rustls::PrivateKey).collect())
+// load_keys parse and return the first private key from the provided file
+pub fn load_keys(path: impl AsRef<Path>) -> io::Result<rustls_pki_types::PrivateKeyDer<'static>> {
```
rylev, line 50: nit: should this be `load_key` now that it's only loading one key?

crates/trigger/src/runtime_config.rs
```diff
@@ -182,6 +183,29 @@ impl RuntimeConfig {
...
+            HashMap::new();
+
+        for opt_layer in self.opts_layers() {
+            for opts in &opt_layer.client_tls_opts {
+                let parsed = parse_client_tls_opts(opts).context("parsing client tls options")?;
+                for component_id in &opts.component_ids {
+                    let mut hostmap: HashMap<String, ParsedClientTlsOpts> = HashMap::new();
+                    for host in &opts.hosts {
+                        hostmap.insert(host.to_owned(), parsed.clone());
+                    }
```
rylev, line 199:
```suggestion
                    let mut hostmap: HashMap<String, ParsedClientTlsOpts> = 
                    opts.hosts.cloned().map(|host| (host, parsed.clone()).collect();
```

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -0,0 +1,39 @@
...
+    let mut custom_root_ca_cursor = Cursor::new(contents);
+
+    Ok(rustls_pemfile::certs(&mut custom_root_ca_cursor)
+        .into_iter()
+        .map(|certs| certs.unwrap())
+        .collect())
+}
+
+// load_keys parse and return the first private key from the provided file
+pub fn load_keys(
```
rylev, line 31:
```suggestion
pub fn load_key(
```

rylev (commented):
From a code quality perspective, this looks good to me (just a few nits). I don't have much opinion on whether this is the best way to expose SSL configuration options, but it looks reasonable to me.

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::io;
+use std::io::Cursor;
+use std::{fs, path::Path};
+
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "snake_case", tag = "type")]
+pub struct ClientTlsOpts {
+    pub component_ids: Vec<String>,
+    pub hosts: Vec<String>,
+    pub custom_root_ca_file: Option<String>,
```
lann, line 12: Maybe `additional_root_ca_file` to clarify that it doesn't replace the default roots?

crates/trigger-http/src/lib.rs
```diff
@@ -1000,3 +1007,212 @@ mod tests {
...
+                .with_no_client_auth();
+        }
+    };
+
+    if let Some(custom_root_ca) = &client_tls_opts_for_host.custom_root_ca {
+        for cer in custom_root_ca {
+            match root_cert_store.add(cer.to_owned()) {
+                Ok(_) => {}
+                Err(e) => {
+                    tracing::warn!("failed to add custom cert to root_cert_store. error: {}", e)
```
lann, line 1200: Is there a use case where we don't want this to be a fatal error?

crates/trigger-http/src/lib.rs
```diff
@@ -1000,3 +1007,212 @@ mod tests {
...
+    })
+}
+
+fn get_client_tls_config_for_authority(
+    authority: &String,
+    client_tls_opts: Option<HashMap<String, ParsedClientTlsOpts>>,
+) -> rustls::ClientConfig {
+    // derived from https://github.com/tokio-rs/tls/blob/master/tokio-rustls/examples/client/src/main.rs
+    let mut root_cert_store = rustls::RootCertStore {
+        roots: webpki_roots::TLS_SERVER_ROOTS.into(),
```
endocrimes, line 1174: If you're providing your own certs, you probably don't want to also trust standard roots? (or they should at least be disable-able)

crates/trigger/src/runtime_config.rs
```diff
@@ -515,3 +542,41 @@ mod tests {
         }
     }
 }
+
+// parsed client tls options
+#[derive(Debug, Clone)]
+pub struct ParsedClientTlsOpts {
+    pub components: Vec<String>,
+    pub hosts: Vec<String>,
```
endocrimes, line 550: Tests and documentation for what happens when multiple TLS blocks contain overlapping hosts would be pretty useful

crates/trigger-http/src/lib.rs
lann: This impl is copied from wasmtime-wasi-http upstream.

crates/trigger-http/src/lib.rs
```diff
@@ -1000,3 +1007,212 @@ mod tests {
...
+                {
+                    dns_error("address not available".to_string(), 0)
+                } else {
+                    types::ErrorCode::ConnectionRefused
+                }
+            }
+        })?;
+
+    let (mut sender, worker) = if use_tls {
+        #[cfg(any(target_arch = "riscv64", target_arch = "s390x"))]
```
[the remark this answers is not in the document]
lann: This impl is copied from wasmtime-wasi-http upstream.

crates/trigger-http/src/lib.rs
lann: The least surprising behavior would probably be:

1. Default to just standard roots
2. If ca file is given, disable standard roots
3. Opt-in to combining the two above

Maybe something like:

```toml
ca_roots_file = "..."
ca_roots_webpki = true # defaults to true iff ca_roots_file not specified
```

crates/trigger-http/src/lib.rs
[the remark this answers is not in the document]
endocrimes: ah i guess they haven't fixed it since upstream got fixed

crates/trigger-http/src/lib.rs
bacongobbler: rustls has two implementations of adding certificates to a certificate store:

1. Adding a short list of known certs: https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.add
2. Adding a long list of certs, such as an OS root cert store: https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.add_parsable_certificates

The former is useful when adding a small set of certs that are known to the user, and the caller would want to know if there was an issue adding those certificates to the store.

The latter is more useful when you are adding a large list of certificates to your store. This is because large collections of root certificates often include ancient or syntactically invalid certificates. `add_parsable_certificates` adds as many certificates to the store as possible in a best-effort fashion.

The intended use case for this feature leans more towards the former than the latter.

crates/trigger/src/runtime_config.rs
```diff
@@ -216,6 +240,9 @@ pub struct RuntimeConfigOpts {
 
     #[serde(skip)]
     pub file_path: Option<PathBuf>,
+
+    #[serde(rename = "client_tls", default)]
+    pub client_tls_opts: Vec<ClientTlsOpts>,
```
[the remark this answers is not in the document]
bacongobbler: To be clear, this is intended for cases where the client runtime cannot load client certificates. The Spin Javascript SDK v2 uses the browser's `fetch` API is a clear example of this. The browser API cannot load certificates presented by the client - it needs the user to add those certificates into the root certificate store in order for that endpoint to be trusted.

I am hopeful that [StarlingMonkey](https://github.com/bytecodealliance/StarlingMonkey) or Spin's SDK will eventually provide ways to load TLS certs via gusts code similar to [Deno](https://github.com/denoland/deno/issues/3444), but we're not at that point yet.

(and yes, this is far from ideal)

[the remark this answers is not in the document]
lann: I think ideally we could define a standard API in terms of wasi:http, possibly using proposed [request metadata](https://github.com/WebAssembly/wasi-http/issues/4).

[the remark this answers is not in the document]
bacongobbler: I just filed https://github.com/fermyon/spin-js-sdk/issues/249 to track this feature request.

crates/trigger-http/src/lib.rs
rajatjindal: done.

crates/trigger-http/src/tls.rs
rajatjindal: done.

crates/trigger/src/runtime_config/client_tls.rs
rajatjindal: done

crates/trigger-http/src/lib.rs
```diff
@@ -1000,3 +1007,212 @@ mod tests {
...
+pub fn dns_error(rcode: String, info_code: u16) -> ErrorCode {
+    ErrorCode::DnsError(wasmtime_wasi_http::bindings::http::types::DnsErrorPayload {
+        rcode: Some(rcode),
+        info_code: Some(info_code),
+    })
+}
+
+/// This is a fork of wasmtime_wasi_http::default_send_request_handler function.
+/// This fork provides the ability to configure client cert auth for mTLS
+pub async fn default_send_request_handler(
```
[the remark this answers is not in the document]
rajatjindal: done

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub cert_chain_file: Option<String>,
+    pub private_key_file: Option<String>,
```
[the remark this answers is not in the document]
rajatjindal: 👍  done

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -22,7 +25,6 @@ pub fn load_certs(
     let mut custom_root_ca_cursor = Cursor::new(contents);
 
     Ok(rustls_pemfile::certs(&mut custom_root_ca_cursor)
-        .into_iter()
         .map(|certs| certs.unwrap())
```
lann, line 28: I see you didn't write this, but if you change the other `unwrap`s to return errors it would be nice to fix this one too.

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -0,0 +1,41 @@
...
+
+// load_certs parse and return the certs from the provided file
+pub fn load_certs(
+    path: impl AsRef<Path>,
+) -> anyhow::Result<Vec<rustls_pki_types::CertificateDer<'static>>> {
+    let contents = fs::read_to_string(path).expect("Should have been able to read the file");
+    let mut custom_root_ca_cursor = Cursor::new(contents);
+
+    Ok(rustls_pemfile::certs(&mut custom_root_ca_cursor)
+        .map(|certs| certs.unwrap())
```
lann, line 28: As below, this should return an error rather than unwrap.

crates/trigger-http/src/lib.rs
```diff
@@ -1000,3 +1006,216 @@ mod tests {
...
+    }
+
+    match (
+        &client_tls_opts_for_host.cert_chain,
+        &client_tls_opts_for_host.private_key,
+    ) {
+        (Some(cert_chain), Some(private_key)) => rustls::ClientConfig::builder()
+            .with_root_certificates(root_cert_store)
+            .with_client_auth_cert(cert_chain.to_owned(), private_key.clone_key())
+            .unwrap(),
```
lann, line 1195: Return error rather than unwrap.

crates/trigger-http/src/lib.rs
rajatjindal: great. done.

crates/trigger-http/src/lib.rs
rajatjindal: added a comment `Optional mapping of authority and TLS options for the current component`

crates/trigger/src/runtime_config.rs
rajatjindal: done

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -0,0 +1,39 @@
...
+use rustls_pemfile::private_key;
+use std::io;
+use std::io::Cursor;
+use std::{fs, path::Path};
+
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "snake_case", tag = "type")]
+pub struct ClientTlsOpts {
+    pub component_ids: Vec<String>,
+    pub hosts: Vec<String>,
```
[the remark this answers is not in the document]
rajatjindal: done. thank you

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -0,0 +1,39 @@
+use anyhow::Context;
+use rustls_pemfile::private_key;
+use std::io;
+use std::io::Cursor;
+use std::{fs, path::Path};
+
+#[derive(Debug, serde::Deserialize)]
+#[serde(rename_all = "snake_case", tag = "type")]
+pub struct ClientTlsOpts {
+    pub component_ids: Vec<String>,
```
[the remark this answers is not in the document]
rajatjindal: 👍  I now added them as KebabId in ClientTlsOpts, but in ParsedClientTlsOpts I am storing them as string. (makes it easy to call map.get(component_id_from_other_struct: &str)

rajatjindal:
I've addressed most of the review comments (except a few unwrap comments, and I do plan to address them before merge). if there are any additional comments, kindly let me know. thanks

I will also continue to add more testcases to it until we are ready to merge the PR.

crates/trigger/src/runtime_config.rs
lann: Technically we should allow multiple certs to be configured; the tls negotiation can select from multiple.

crates/trigger-http/src/tls.rs
```diff
@@ -19,28 +19,118 @@ impl TlsConfig {
...
+    private_key(&mut io::BufReader::new(fs::File::open(path).map_err(
+        |err| {
+            io::Error::new(
+                io::ErrorKind::InvalidInput,
+                format!("failed to read private key file {:?}", err),
+            )
+        },
+    )?))
+    .map_err(|_| io::Error::new(io::ErrorKind::InvalidInput, "invalid private key"))
+    .map(|keys| keys.unwrap())
```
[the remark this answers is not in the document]
rajatjindal: done.

crates/trigger/src/runtime_config/client_tls.rs
rajatjindal: done

crates/trigger/src/runtime_config/client_tls.rs
```diff
@@ -0,0 +1,39 @@
...
+
+// load_keys parse and return the first private key from the provided file
+pub fn load_keys(
+    path: impl AsRef<Path>,
+) -> anyhow::Result<rustls_pki_types::PrivateKeyDer<'static>> {
+    private_key(&mut io::BufReader::new(
+        fs::File::open(path).context("loading private key")?,
+    ))
+    .map_err(|_| anyhow::anyhow!("invalid input"))
+    .map(|keys| keys.unwrap())
```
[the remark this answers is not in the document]
rajatjindal: done, thank you

crates/trigger/src/runtime_config/client_tls.rs
rajatjindal: updated the logic as per suggestion in a different comment (load default certs true by default unless custom ca provided, in which case it is disable by default. It can be forced to be enabled by explicit configuration).

crates/trigger-http/src/lib.rs
rajatjindal: 👍  this is done.

crates/trigger-http/src/lib.rs
rajatjindal: 👍  done

crates/trigger/src/runtime_config.rs
rajatjindal: in that case maybe we keep it simple and let the last block win (the current behavior) or should we fail it for now?

rajatjindal:
I think I've addressed most of the review feedback. I can spend some more time to add more tests, but otherwise it looks ok for re-review. 

thanks

crates/trigger/src/runtime_config.rs
lann: I might suggest first-wins here since that is how variable provider configs already work.

rajatjindal: 👍 , I have now changed the logic to make first entry win.

crates/trigger/src/lib.rs
```diff
@@ -425,6 +438,13 @@ impl<Executor: TriggerExecutor> TriggerAppEngine<Executor> {
         })
     }
 
+    pub fn get_client_tls_opts(
+        &self,
+        component_id: &str,
+    ) -> Option<HashMap<Authority, ParsedClientTlsOpts>> {
+        self.client_tls_opts.get(&component_id.to_string()).cloned()
+    }
```
rylev, line 446: Nit: does this have to clone?

```suggestion
    pub fn get_client_tls_opts(
        &self,
        component_id: &str,
    ) -> Option<&HashMap<Authority, ParsedClientTlsOpts>> {
        self.client_tls_opts.get(&component_id.to_string())
    }
```

crates/trigger/src/runtime_config.rs
```diff
@@ -182,6 +184,58 @@ impl RuntimeConfig {
...
+        let mut components_map: HashMap<String, HashMap<Authority, ParsedClientTlsOpts>> =
+            HashMap::new();
+
+        // if available, use the existing client tls opts value for a given component-id and host-authority
+        // to ensure first-one wins incase of duplicate options
+        fn use_existing_if_available(
+            existing_opts: Option<&HashMap<Authority, ParsedClientTlsOpts>>,
+            host: Authority,
+            newopts: ParsedClientTlsOpts,
+        ) -> Option<(Authority, ParsedClientTlsOpts)> {
```
rylev, line 201: It looks like this function never returns `None`. Can this just return `(Authority, ParsedClientTlsOpts)` instead of the option?

crates/trigger/src/runtime_config.rs
```diff
@@ -182,6 +184,58 @@ impl RuntimeConfig {
...
+            match existing_opts {
+                None => Some((host, newopts.clone())),
```
rylev, line 203:
```suggestion
                None => Some((host, newopts)),
```

crates/trigger/src/runtime_config.rs
```diff
@@ -182,6 +184,58 @@ impl RuntimeConfig {
...
+                Some(opts) => match opts.get(&host) {
+                    Some(existing_opts_for_component_and_host) => {
+                        Some((host, existing_opts_for_component_and_host.to_owned()))
+                    }
+                    None => Some((host, newopts.clone())),
```
rylev, line 208: These clones aren't necessary
```suggestion
                    None => Some((host, newopts)),
```

crates/trigger/src/runtime_config.rs
```diff
@@ -182,6 +184,58 @@ impl RuntimeConfig {
...
+                        .into_iter()
+                        .filter_map(|host| {
+                            use_existing_if_available(
+                                existing_opts_for_component,
+                                host,
+                                parsed.clone(),
+                            )
+                        })
+                        .collect::<HashMap<Authority, ParsedClientTlsOpts>>();
+                    components_map.insert(component_id.to_string().to_owned(), hostmap);
```
rylev, line 231:
```suggestion
                    components_map.insert(component_id.to_string(), hostmap);
```

crates/trigger/src/runtime_config.rs
```diff
@@ -182,6 +184,58 @@ impl RuntimeConfig {
...
+                    None => Some((host, newopts.clone())),
+                },
+            }
+        }
+
+        for opt_layer in self.opts_layers() {
+            for opts in &opt_layer.client_tls_opts {
+                let parsed = parse_client_tls_opts(opts).context("parsing client tls options")?;
+                for component_id in &opts.component_ids {
+                    let existing_opts_for_component = components_map.get(&component_id.to_string());
```
rylev, line 217:
```suggestion
                    let existing_opts_for_component = components_map.get(component_id.as_ref());
```

rylev (approved):
Looks good - I have some nits, but you can choose whether to address them or not.

rajatjindal:
> Looks good - I have some nits, but you can choose whether to address them or not.

I have addressed most of the nits you suggested.

rylev:
@rajatjindal feel free to merge this whenever you're ready

PR #2610, opened 2024-07-01
Allow sqlite migrations for non-default databases.

### LINKED ISSUE #2607: SQLITE error after update from 2.5.1 to 2.6.0
The application was created with spin 2.5.1 and was runnable with spin up.
The following steps were taken:

```toml
spin_manifest_version = 2

[application]
name = "TEST-PROCESSOR"
version = "0.1.0"
authors = ["*********"]
description = "*************"

[[trigger.http]]
route = "/api/vehicle/..."
component = "vehicle"

[[trigger.http]]
route = "/api/updates/..."
component = "update-status"

[component.vehicle]
source = "vehicle/target/wasm32-wasi/release/vehicle.wasm"
allowed_outbound_hosts = []
sqlite_databases = ["vehicle"]

[component.vehicle.build]
workdir = "vehicle"

[component.update-status]
source = "update_status/target/wasm32-wasi/release/update_status.wasm"
allowed_outbound_hosts = []
sqlite_databases = ["update-status"]

[component.update-status.build]
workdir = "update_status"
```

Building and Starting with
```bash
spin build --up --sqlite @migration_vehicle.sql --sqlite @migration_update_status.sql
```

In the .spin folder, an sqlite_db.db is created.
The migration data was also correctly created.

After updating via brew to spin 2.6.0,
the application can no longer be started.

```bash
spin up
```

  Error: One or more components use SQLite databases which are not defined.
  See https://developer.fermyon.com/spin/dynamic-configuration#sqlite-storage-runtime-configuration
  Details:
  - Component vehicle uses database 'vehicle'
  - Component update-status uses database 'update-status'

As a test, the .spin/sqlite_db.db was deleted
and called with the command

```bash
spin build --up --sqlite @migration_vehicle.sql --sqlite @migration_update_status.sql
```
The sqlite_db.db is correctly created, but the application does not start, same error.

Let me know if you need any further adjustments!

* Spin version (`2.6.0`)
* Installed plugin versions:
** loud 0.9.0 [installed]
** js2wasm 0.6.1 [installed]
** py2wasm 0.3.2 [installed]

rylev:
Fixes #2607

The database will be selected by the name of the migration file. This makes the name of the migration file significant, whereas previously it was not. If the name of the file and a known database don't line up, `spin up` will fail with an error message.

### DIFF
--- a/crates/trigger/src/runtime_config/sqlite.rs
+++ b/crates/trigger/src/runtime_config/sqlite.rs
@@ -51,11 +51,16 @@ async fn execute_statements(
     }
 
     for m in statements {
-        if let Some(file) = m.strip_prefix('@') {
-            let database = file.strip_suffix(".sql").unwrap_or(file);
+        if let Some(config) = m.strip_prefix('@') {
+            let config = config.trim();
+            let (file, database) = match config.split_once(':') {
+                Some((file, database)) if database.trim().is_empty() => (file.trim(), "default"),
+                Some((file, database)) => (file.trim(), database.trim()),
+                None => (config, "default"),
+            };
             let database = databases.get(database).with_context(|| {
                 format!(
-                    "based on the sql file '{file}' a registered database named '{database}' was expected but not found. The registered databases are '{:?}'", databases.keys()
+                    "based on the '@{config}' a registered database named '{database}' was expected but not found. The registered databases are '{:?}'", databases.keys()
                 )
             })?;
             let sql = std::fs::read_to_string(file).with_context(|| {
--- a/crates/trigger/src/runtime_config/sqlite.rs
+++ b/crates/trigger/src/runtime_config/sqlite.rs
@@ -49,24 +49,30 @@ async fn execute_statements(
     if statements.is_empty() {
         return Ok(());
     }
-    let Some(default) = databases.get("default") else {
-        debug_assert!(
-            false,
-            "the 'default' sqlite database should always be available but for some reason was not"
-        );
-        return Ok(());
-    };
 
     for m in statements {
         if let Some(file) = m.strip_prefix('@') {
+            let database = file.strip_suffix(".sql").unwrap_or(file);
+            let database = databases.get(database).with_context(|| {
+                format!(
+                    "based on the sql file '{file}' a registered database named '{database}' was expected but not found. The registered databases are '{:?}'", databases.keys()
+                )
+            })?;
             let sql = std::fs::read_to_string(file).with_context(|| {
                 format!("could not read file '{file}' containing sql statements")
             })?;
-            default
+            database
                 .execute_batch(&sql)
                 .await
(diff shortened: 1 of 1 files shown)


itowlson (changes requested):
Given we have docs that use examples like `spin up --sqlite @migrations.sql`, and that migrations may come as a sequence of files (a la `migrations001.sql`, `migrations002.sql`, etc.), I'm reluctant to adopt this UI.  Let's instead look for a more structured and specific syntax, such as `--sqlite custom-db:@migrations.sql` or `--sqlite @migrations.sql>custom-db` (the latter would presumably make shells mad but just trying to spark ideas).

itowlson:
`@file.sql:database-name` works for me.  Thanks!

rylev:
@itowlson ok done!

crates/trigger/src/runtime_config/sqlite.rs
```diff
@@ -49,24 +49,35 @@ async fn execute_statements(
...
+            database
                 .execute_batch(&sql)
                 .await
                 .with_context(|| format!("failed to execute sql from file '{file}'"))?;
         } else {
+            let Some(default) = databases.get("default") else {
+                debug_assert!(
+            false,
+            "the 'default' sqlite database should always be available but for some reason was not"
+        );
```
itowlson, line 78: It may be GH preview silliness but lines 76-78 seem confusingly outdented?

crates/trigger/src/runtime_config/sqlite.rs
```diff
@@ -49,24 +49,35 @@ async fn execute_statements(
...
-        debug_assert!(
-            false,
-            "the 'default' sqlite database should always be available but for some reason was not"
-        );
-        return Ok(());
-    };
 
     for m in statements {
-        if let Some(file) = m.strip_prefix('@') {
+        if let Some(config) = m.strip_prefix('@') {
```
itowlson, line 54: For me the argument parser is now complicated enough to merit pulling out into a function (ideally with a few simple tests).  I know some folks have different tolerances for this though, especially as you'd probably need to add an enum for the "file vs literal SQL" distinction.  Definitely not a blocker!

crates/trigger/src/runtime_config/sqlite.rs
```diff
@@ -49,24 +49,35 @@ async fn execute_statements(
...
+            let config = config.trim();
+            let (file, database) = match config.split_once(':') {
+                Some((file, database)) if database.trim().is_empty() => (file.trim(), "default"),
```
itowlson, line 57: This would be the `@file.sql:` case?  That feels a bit like a syntax error to me but I don't feel strongly - not a blocker.

itowlson (approved):
Couple of non-blocking suggestions, but before merging could you double check that the indents around lines 76-78 are right in the real code?  I'm assuming a GH artifact but good to be sure!  Otherwise LGTM - thanks!

Issue #2259, opened 2024-01-30
`spin registry push` adding OCI metadata


radu-matei:
#2236 added optional support for pushing metadata in the OCI artifact (thank you, @rgl!)

Should we consider automatically adding the Spin application description, authors (other metadata) in the manifest?

vdice:
This seems like a good issue to track automatically generating and adding annotations to the OCI manifest.  Recently discussed in https://github.com/fermyon/spin/issues/2612#issuecomment-2204029049.

@radu-matei do you envision using the [pre-defined annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys) for such auto-generated metadata?  Or do you have ideas on alternate (or additional) keys?

PR #2618, opened 2024-07-09
Infer predefined annotations when pushing to registry

### LINKED ISSUE #2612: Follow OCI container image annotation standards when generating containers
As a part of generating a docker container when running something like: `spin registry push ttl.sh/something:1h`, we get very basic information about the container.

```
➜ docker manifest inspect ttl.sh/something:1h
{
    "config": {
        "digest": "sha256:668d372ccc29ffe6f0000bbddbff67d97c2812dbd48b5338ae8532ab6b7c4e59",
        "mediaType": "application/vnd.oci.image.config.v1+json",
        "size": 209
    },
    "layers": [
        {
            "digest": "sha256:2bcd094bfd433033c641264d94d0afbd0adabc666d69433d2f766e0efbe5e598",
            "mediaType": "application/vnd.wasm.content.layer.v1+wasm",
            "size": 36276905
        },
        {
            "digest": "sha256:3f05015b248403ca99a9bf1861202d107885ae300f5ff165708415603a39c248",
            "mediaType": "application/vnd.fermyon.spin.application.v1+config",
            "size": 596
        }
    ],
    "schemaVersion": 2
}
```

But there is an OCI standard for [pre-defined annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys).

It would be great if spin could generate most of these annotations out of the box, if possible, guess them from context. But make it possible to set all values through  the`spin.toml` file.

itowlson:
Fixes #2612 (or at least part of it).

I'm not sure the best way to test this in CI, but trying it locally it seems promising:

```
# Manifest for testing
ivan@hestia:~/testing$ cat v2vanillareallyv2/spin.toml
spin_manifest_version = 2

[application]
name = "v2vanillareallyv2"
version = "0.1.0"
authors = ["itowlson <ivan.towlson@fermyon.com>"]
description = ""
# moar TOML

# If the user doesn't set any of the predefined annotations
ivan@hestia:~/testing$ spin registry push localhost:5000/vanilla:v0.1 -f v2vanillareallyv2/ --insecure
Pushing app to the Registry..
Pushed with digest sha256:9110a84446a12c3479978316e2343e60bda6d5375ff39fb0ef0f77759d7d1011

# ...then they are inferred (formatting added for legibility)
ivan@hestia:~/testing$ docker buildx imagetools inspect localhost:5000/vanilla:v0.1 --raw
{"annotations":{
  "org.opencontainers.image.authors":"itowlson <ivan.towlson@fermyon.com>",
  "org.opencontainers.image.created":"2024-07-09T01:33:37.290049791+00:00",
  "org.opencontainers.image.title":"v2vanillareallyv2",
  "org.opencontainers.image.version":"0.1.0"
},
# moar JSON
}

# But if the user specifies, say, the `authors` annotation...
ivan@hestia:~/testing$ spin registry push localhost:5000/vanilla:v0.2 -f v2vanillareallyv2/ --insecure --annotation org.opencontainers.image.authors="THE GREAT OZ"
Pushing app to the Registry..
Pushed with digest sha256:63420c721dbb51261c0c3af18cded3b6ef81d8fc65c820be535500f27cb4285e

# ...then we do not infer it but respect the one we are given (formatting added for legibility)
ivan@hestia:~/testing$ docker buildx imagetools inspect localhost:5000/vanilla:v0.2 --raw
{"annotations":{
  "org.opencontainers.image.authors":"THE GREAT OZ",
  "org.opencontainers.image.created":"2024-07-09T01:36:01.402920121+00:00",
  "org.opencontainers.image.title":"v2vanillareallyv2",
  "org.opencontainers.image.version":"0.1.0"
},
# moar JSON
}
```

The OCI `authors` annotation is a string, whereas Spin allows an array: the PR comma-space-concatenates them, e.g. `"The Great Oz, Dorothy Gale"`.  OCI describes the field as a freeform string but if there's a common community convention then I can adopt that.

This PR does not provide a way for a user to suppress inference but the infrastructure is there in case we want to add an option, or for programmatic use (e.g. reproducibility).

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -84,6 +84,17 @@ pub struct ClientOpts {
     pub content_ref_inline_max_size: usize,
 }
 
+/// Controls whether predefined annotations are generated when pushing an application.
+/// If an explicit annotation has the same name as a predefined one, the explicit
+/// one takes precedence.
+#[derive(Debug, PartialEq)]
+pub enum InferPredefinedAnnotations {
+    /// Infer annotations for created, authors, version, name and description.
+    All,
+    /// Do not generate any annotations; use only explicitly supplied annotations.
+    None,
+}
+
 impl Client {
     /// Create a new instance of an OCI client for distributing Spin applications.
     pub async fn new(insecure: bool, cache_root: Option<PathBuf>) -> Result<Self> {
@@ -107,6 +118,7 @@ impl Client {
         manifest_path: &Path,
         reference: impl AsRef<str>,
         annotations: Option<BTreeMap<String, String>>,
+        infer_annotations: InferPredefinedAnnotations,
     ) -> Result<Option<String>> {
         let reference: Reference = reference
             .as_ref()
@@ -125,7 +137,7 @@ impl Client {
         )
         .await?;
 
-        self.push_locked_core(locked, auth, reference, annotations)
+        self.push_locked_core(locked, auth, reference, annotations, infer_annotations)
             .await
     }
 
@@ -136,6 +148,7 @@ impl Client {
         locked: LockedApp,
         reference: impl AsRef<str>,
         annotations: Option<BTreeMap<String, String>>,
+        infer_annotations: InferPredefinedAnnotations,
     ) -> Result<Option<String>> {
         let reference: Reference = reference
             .as_ref()
@@ -143,7 +156,7 @@ impl Client {
             .with_context(|| format!("cannot parse reference {}", reference.as_ref()))?;
         let auth = Self::auth(&reference).await?;
 
-        self.push_locked_core(locked, auth, reference, annotations)
+        self.push_locked_core(locked, auth, reference, annotations, infer_annotations)
             .await
     }
 
@@ -155,6 +168,7 @@ impl Client {
         auth: RegistryAuth,
         reference: Reference,
         annotations: Option<BTreeMap<String, String>>,
+        infer_annotations: InferPredefinedAnnotations,
     ) -> Result<Option<String>> {
         let mut locked_app = locked.clone();
         let mut layers = self
@@ -173,6 +187,8 @@ impl Client {
                 .await
                 .context("could not assemble archive layers for locked application")?;
         }
+
+        let annotations = all_annotations(&locked_app, annotations, infer_annotations);
 
         // Push layer for locked spin application config
         let locked_config_layer = ImageLayer::new(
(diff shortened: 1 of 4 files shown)


itowlson:
Also linked #2259

crates/oci/src/client.rs
```diff
@@ -136,14 +148,15 @@ impl Client {
         locked: LockedApp,
         reference: impl AsRef<str>,
         annotations: Option<BTreeMap<String, String>>,
+        predefined_annotations: PredefinedAnnotations,
```
radu-matei, line 151: The fact that `annotations` is an option of a map and `predefined_annotations` is an enum is somewhat confusing here; how much of this behavior is an external consumer of this API supposed to understand?

src/commands/registry.rs
```diff
@@ -86,7 +86,14 @@ impl Push {
 
         let _spinner = create_dotted_spinner(2000, "Pushing app to the Registry".to_owned());
 
-        let digest = client.push(&app_file, &self.reference, annotations).await?;
+        let digest = client
+            .push(
+                &app_file,
+                &self.reference,
+                annotations,
+                PredefinedAnnotations::Infer,
```
radu-matei, line 94: So the behavior here is: always infer, and if the user provided any annotation explicitly, it takes priority over the inferred value??

crates/oci/src/client.rs
```diff
@@ -688,6 +704,68 @@ fn registry_from_input(server: impl AsRef<str>) -> String {
...
+
+    let authors = locked_app
+        .get_metadata(APP_AUTHORS_KEY)
+        .unwrap_or_default()
+        .unwrap_or_default();
+    if !authors.is_empty() {
+        let authors = authors.join(", ");
+        add_inferred(
+            &mut current,
+            "org.opencontainers.image.authors",
```
vdice, line 731: We should be able to utilize the [upstream definitions](https://github.com/oras-project/rust-oci-client/blob/main/src/annotations.rs) for these keys, via our current fermyon/oci-distribution dependency.

crates/oci/src/client.rs
```diff
@@ -84,6 +84,17 @@ pub struct ClientOpts {
     pub content_ref_inline_max_size: usize,
 }
 
+/// Controls whether predefined annotations are generated when pushing an application.
+/// If an explicit annotation has the same name as a predefined one, the explicit
+/// one takes precedence.
+#[derive(Debug, PartialEq)]
+pub enum PredefinedAnnotations {
```
vdice, line 91: I wonder if adding this, or something similar, into ClientOpts might be an option?  Then we'd be able to access it from the client when we need it and wouldn't have to pass it through functions as an additional arg.

crates/oci/src/client.rs
```diff
@@ -688,6 +704,68 @@ fn registry_from_input(server: impl AsRef<str>) -> String {
     }
 }
 
+fn infer_annotations(
```
vdice, line 707: I think if we covered this method in a unit test that'd take the pressure off adding higher-level `spin registry push` coverage.

vdice (commented):
Nice, thanks Ivan!

vdice:
> OCI describes the field as a freeform string but if there's a common community convention then I can adopt that.

Not able to find a convention after a brief search.  I think the comma-separated string as you've initially added seems a sane way to go for now.

src/commands/registry.rs
itowlson: If the user provides _one of the annotations that would be inferred_ explicitly, then the explicit value takes precedence - see the comment on the enum.

crates/oci/src/client.rs
itowlson: EGAD

I didn't even think to look!  Thanks @vdice

crates/oci/src/client.rs
itowlson: Folding into ClientOpts definitely feels worth investigation.  That would probably allow us to get rid of the enum and use a boolean into the bargain.  I'll take a look.

itowlson: @vdice I had an explore of the ClientOpts approach.  It ended up feeling a bit awkward that the explicit annotations were set up in one place and the inferred ones in another.  So I am going to back that out.  Appreciate the idea though!

crates/oci/src/client.rs
itowlson: @radu-matei I am gazing at this and struggling to think of a better design.  I don't want to put it on consumers to construct the predefined annotations, so it shouldn't be a map.  So I am having difficulty getting away from the "I want the predefined annotations or I don't" flag.  I would value suggestions.  Or is it as simple as "we should rename the flag to `infer_predefined_annotations` or something, so it sounds like an action instead of a set" perhaps?

itowlson: @radu-matei I tried some renaming - if that doesn't improve matters then I'd value suggestions - thanks!

crates/oci/src/client.rs
itowlson: @vdice I've added unit tests - I think I have all the cases!

vdice (approved):
🎸👨‍🎤 🔈 LGTM+11!

PR #2620, opened 2024-07-09
factors: Make instance state fields again

lann:


rylev (approved):
This makes a lot of sense and is a definite improvement!

PR #2621, opened 2024-07-09
factors: Add spin-factors-test crate

lann:
This provides a TestEnvironment type to aid factor authors in writing tests.

In order to enable TestEnvironment, move all of the generated RuntimeFactors methcods into the trait.

Exercise TestEnvironment in a spin-factor-variables test.

### DIFF
--- a/crates/factors-test/src/lib.rs
+++ b/crates/factors-test/src/lib.rs
@@ -1,0 +1,69 @@
+use spin_app::locked::LockedApp;
+use spin_factors::{
+    anyhow::{self, Context},
+    serde::de::DeserializeOwned,
+    wasmtime::{Config, Engine},
+    App, Linker, RuntimeConfigSource, RuntimeFactors,
+};
+use spin_loader::FilesMountStrategy;
+
+pub use toml::toml;
+
+#[derive(Default)]
+pub struct TestEnvironment {
+    pub manifest: toml::Table,
+    pub runtime_config: toml::Table,
+}
+
+impl TestEnvironment {
+    /// Starting from a new _uninitialized_ [`RuntimeFactors`], run through the
+    /// [`Factor`]s' lifecycle(s) to build a [`RuntimeFactors::InstanceState`].
+    pub async fn build_instance_state<T: RuntimeFactors>(
+        &mut self,
+        mut factors: T,
+    ) -> anyhow::Result<T::InstanceState> {
+        let mut linker = Self::new_linker::<T>();
+        factors.init(&mut linker)?;
+
+        let locked_app = self.build_locked_app().await?;
+        let app = App::inert(locked_app);
+        let runtime_config = TomlRuntimeConfig(&self.runtime_config);
+        let configured_app = factors.configure_app(app, runtime_config)?;
+
+        let component = configured_app
+            .app()
+            .components()
+            .next()
+            .context("no components")?;
+        factors.build_store_data(&configured_app, component.id())
+    }
+
+    pub fn new_linker<T: RuntimeFactors>() -> Linker<T> {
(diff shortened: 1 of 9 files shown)


crates/factors-test/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        let engine = Engine::new(Config::new().async_support(true)).expect("engine");
+        Linker::<T>::new(&engine)
+    }
+
+    pub async fn build_locked_app(&self) -> anyhow::Result<LockedApp> {
+        let toml_str = toml::to_string(&self.manifest).expect("serialize toml");
+        let dir = tempfile::tempdir().expect("tempdir");
+        let path = dir.path().join("spin.toml");
+        std::fs::write(&path, toml_str).expect("write spin.toml");
+        spin_loader::from_file(&path, FilesMountStrategy::Direct, None).await
```
rylev, line 51: Is there a reason `spin_loader` only allow for loading from disk? This isn't the only place where we're writing some in-memory representation of the manifest to a temp directory only to then load it.

crates/factors-test/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    pub fn new_linker<T: RuntimeFactors>() -> Linker<T> {
+        let engine = Engine::new(Config::new().async_support(true)).expect("engine");
+        Linker::<T>::new(&engine)
+    }
+
+    pub async fn build_locked_app(&self) -> anyhow::Result<LockedApp> {
+        let toml_str = toml::to_string(&self.manifest).expect("serialize toml");
+        let dir = tempfile::tempdir().expect("tempdir");
```
rylev, line 48: What's the reason for panicking here instead of returning an error? In any case, can we make these error messages a bit clearer (e.g., "failed to serialize TestEnvironment manifest to a string")

crates/factors-test/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        let component = configured_app
+            .app()
+            .components()
+            .next()
+            .context("no components")?;
+        factors.build_store_data(&configured_app, component.id())
+    }
+
+    pub fn new_linker<T: RuntimeFactors>() -> Linker<T> {
```
rylev, line 41: Does this need to be `pub`?

crates/factors-test/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        let engine = Engine::new(Config::new().async_support(true)).expect("engine");
+        Linker::<T>::new(&engine)
+    }
+
+    pub async fn build_locked_app(&self) -> anyhow::Result<LockedApp> {
```
rylev, line 46: Same here, does this need to be public?

rylev (approved):
Looks good! Just have a few nits

crates/factors-test/src/lib.rs
lann: No fundamental reason; it would require some refactoring to handle the cases where the disk is actually read to e.g. expand `files` patterns.

crates/factors-test/src/lib.rs
lann: The first iteration of these methods didn't return `Result`s; I just didn't clean everything up when I changed it.

crates/factors-test/src/lib.rs
lann: These are just available if a test wants to customize some of what `build_instance_state` does.

crates/factors-test/src/lib.rs
lann: See above.

PR #2626, opened 2024-07-10
factors: Improve tests

lann:
The `spin-factors-derive` `expander` feature is very useful for debugging macro output but breaks rust-analyzer due to https://github.com/rust-lang/rust-analyzer/issues/17040. Fix with some annoying build script boilerplate.

Also build out the WASI factor a bit, including a test.


crates/factor-variables/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    println!("cargo::rerun-if-changed=build.rs");
+    // Enable spin-factors-derive to emit expanded macro output.
+    let out_dir = std::env::var("OUT_DIR").unwrap();
+    println!("cargo:rustc-env=SPIN_FACTORS_DERIVE_EXPAND_DIR={out_dir}");
+}
```
calebschoepp, line 6: Do we have to do this in every factor? Is the only benefit so that you can `CTRL`+`CLICK` rather than run expand macro from the command palette of VS code?

crates/factor-wasi/tests/factor_test.rs
```diff
@@ -0,0 +1,37 @@
+use spin_factor_wasi::{DummyFilesMounter, WasiFactor};
```
calebschoepp, line 1: Nit: What do you think about the file name `wasi_test.rs` so we don't have 13 million `factor_test.rs` files? Just a thought.

crates/factor-variables/build.rs
lann: The main benefit is that error messages pointing into generated code are actually useful.

crates/factor-wasi/tests/factor_test.rs
lann: 🤷

crates/factor-variables/build.rs
lann: Now that if you omit the build script you just get normal macro behavior.

rylev (approved):

crates/factor-variables/build.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    println!("cargo::rerun-if-changed=build.rs");
```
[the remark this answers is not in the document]
lann: `cargo`-culted from https://github.com/bytecodealliance/wasmtime/blob/0e5048e20a24f6da19b3794eeb94111f052dc414/crates/component-macro/build.rs#L2

From the cargo docs:
> If the build script inherently does not need to re-run under any circumstance, then emitting cargo::rerun-if-changed=build.rs is a simple way to prevent it from being re-run (otherwise, the default if no rerun-if instructions are emitted is to scan the entire package directory for changes).

PR #2627, opened 2024-07-11
Make yes the default for generating new project in non-emtpy directory

rylev:
If the --accept-defaults is passed then the user won't be prompted.

I discovered this when investigating the issue with https://github.com/fermyon/spin/pull/2615 which was failing because we could not create a project in a non-empty directory. I was very surprised to see that `--accept-defaults` did nothing.

### DIFF
--- a/crates/templates/src/interaction.rs
+++ b/crates/templates/src/interaction.rs
@@ -102,7 +102,7 @@ impl InteractionStrategy for Silent {
 }
 
 pub(crate) fn confirm(text: &str) -> std::io::Result<bool> {
-    Confirm::new().with_prompt(text).interact()
+    Confirm::new().with_prompt(text).default(true).interact()
 }
 
 pub(crate) fn prompt_parameter(parameter: &TemplateParameter) -> Option<String> {
--- a/crates/templates/src/run.rs
+++ b/crates/templates/src/run.rs
@@ -93,12 +93,14 @@ impl Run {
         // TODO: rationalise `path` and `dir`
         let to = self.generation_target_dir();
 
-        match interaction.allow_generate_into(&to) {
-            Cancellable::Cancelled => return Ok(None),
-            Cancellable::Ok(_) => (),
-            Cancellable::Err(e) => return Err(e),
-        };
-
+        // If the user isn't accepting default, asking for permission to generate into a non-empty directory
+        if !self.options.accept_defaults {
+            match interaction.allow_generate_into(&to) {
+                Cancellable::Cancelled => return Ok(None),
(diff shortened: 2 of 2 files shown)


itowlson:
Hmm, I thought `--accept-defaults` applied only to parameters, not to the overwrite question.  I thought we had a `--overwrite` CLI flag for that but apparently not.

itowlson:
Hmm, when I try generating into a non-empty directory, the user _is_ prompted.  Maybe I'm misunderstanding your first sentence.

itowlson:
I am anxious that this makes a destructive option the default.  A couple of thoughts for an alternate way of meeting the need:

1. Detect if stderr is a terminal, and suppress the overwrite prompt if not, inferring a suitable default
1. Have an explicit flag e.g. `--overwrite`/`--no-overwrite` to override the default overwrite behaviour

(Broadly the template system should be more aware of the "stderr is not a terminal" case.  We can switch into programmatically using the `interaction::Silent` strategy, but should do so automatically in non-terminal environments.  This will give us proper errors instead of hanging if we are in a non-terminal environment and we do not have values or defaults for everything we need.  But that's out of scope for this PR for sure!)

rylev:
> Hmm, when I try generating into a non-empty directory, the user _is_ prompted. Maybe I'm misunderstanding your first sentence.

Sorry, the first sentence in the description is describing the *new* behavior. 

As for which path I think is correct, I don't have a strong opinion here, just that we should allow some way of not prompting the user.

I assume you don't have strong feelings on what the right path forward is, and neither do I. To avoid letting this just die here, perhaps doing terminal detection is right compromise?

itowlson:
I woke up at 4am and had a stronger opinion... I see people using `--accept-defaults` to blow past the HTTP path etc., and I don't want to have those people accidentally overwrite existing code - I want them to either be made aware (prompt) or to knowingly opt in (off-by-default flag).

I also realised I should share some context.  Generally the template system was never really designed for the use case of "a CI system instantiating templates" - I always assumed that template instantiation was always part of an interactive process because you would never create an app_except to edit it!  There is a non-interactive mode for unit testing the template system itself, but that isn't exposed via the CLI for the reason just mentioned.

So my proposal is:

* Terminal detection in `spin add/new` to invoke silent mode if non-interactive
* Default behaviour in silent mode remains "fail if directory already exists"
* New option to allow generating into an existing directory
  * Surfaced in the CLI,
  * Effective in both interactive and silent modes

Thanks for being patient with my rambling on this, and for highlighting it!

rylev:
Closing in favor of https://github.com/fermyon/spin/pull/2643

PR #2630, opened 2024-07-11
Add a llm-factors

rylev:
Adds an LLM factors.

I ran out of time to test the host implementation. Feel free to keep this open, and I'll add that tomorrow.

### DIFF
--- a/crates/factor-llm/src/lib.rs
+++ b/crates/factor-llm/src/lib.rs
@@ -1,0 +1,114 @@
+mod host;
+
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use spin_factors::{
+    ConfigureAppContext, Factor, InstanceBuilders, PrepareContext, RuntimeFactors,
+    SelfInstanceBuilder,
+};
+use spin_locked_app::MetadataKey;
+use spin_world::v1::llm::{self as v1};
+use spin_world::v2::llm::{self as v2};
+
+pub const ALLOWED_MODELS_KEY: MetadataKey<Vec<String>> = MetadataKey::new("ai_models");
+
+pub struct LlmFactor {
+    create_engine: Box<dyn Fn() -> Box<dyn LlmEngine> + Send + Sync>,
+}
+
+impl LlmFactor {
+    pub fn new<F>(create_engine: F) -> Self
+    where
+        F: Fn() -> Box<dyn LlmEngine> + Send + Sync + 'static,
+    {
+        Self {
+            create_engine: Box::new(create_engine),
+        }
+    }
+}
+
+impl Factor for LlmFactor {
+    type RuntimeConfig = ();
+    type AppState = AppState;
+    type InstanceBuilder = LlmDispatch;
+
+    fn init<T: RuntimeFactors>(
+        &mut self,
+        mut ctx: spin_factors::InitContext<T, Self>,
+    ) -> anyhow::Result<()> {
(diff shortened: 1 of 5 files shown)


crates/factor-llm/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+        Self {
+            create_engine: Box::new(create_engine),
+        }
+    }
+}
+
+impl Factor for LlmFactor {
+    type RuntimeConfig = ();
+    type AppState = AppState;
+    type InstanceBuilder = LlmDispatch;
```
calebschoepp, line 35: I will say I was confused for quite awhile reading this code what `LlmDispatch` was. I expected it to be called `InstanceState` like we do in some of the other factors. Maybe it would be nice to consistently name this? Maybe I just need to read the code better? Idk, 🤷 pretty ambivalent about what you choose to do here, but thought I would point out I struggled.

crates/factor-llm/tests/factor.rs
```diff
@@ -0,0 +1,65 @@
...
+use spin_world::v1::llm::{self as v1};
+use spin_world::v2::llm::{self as v2, Host};
+
+#[derive(RuntimeFactors)]
+struct TestFactors {
+    llm: LlmFactor,
+}
+
+#[tokio::test]
+async fn static_provider_works() -> anyhow::Result<()> {
```
calebschoepp, line 15: Function name doesn't make sense.

calebschoepp (commented):
If I understand correctly `factor-llm` replaces `llm` but the `llm-local` and `llm-remote-http` crates will remain? Please confirm.

crates/factor-llm/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+    ) -> anyhow::Result<Self::InstanceBuilder> {
+        let allowed_models = ctx
+            .app_state()
+            .component_allowed_models
+            .get(ctx.app_component().id())
+            .cloned()
+            .unwrap_or_default();
+
+        Ok(LlmDispatch {
+            engine: (self.create_engine)(),
```
lann, line 83: Is `create_engine` supposed to be cheap? This gets called unconditionally whether a component uses llm or not.

crates/factor-llm/src/lib.rs
rylev: I'm fine with renaming it to `InstanceState` - `LlmDispatch` was just used in the old `llm` implementation.

crates/factor-llm/src/lib.rs
rylev: It's relatively cheap for both implementations (some cloning of some data structures happening), but there's no I/O happening in either.

rylev:
@calebschoepp that is correct - the `llm-local` and `llm-remote-http` crates are the backing implementations of `LlmEngine`.

lann (approved):

calebschoepp:
@rylev maybe you could add this to the factors smoketest. @lann is that smoketest intended to stick around or is it redundant once we have things integrated.

lann:
> @rylev maybe you could add this to the factors smoketest. @lann is that smoketest intended to stick around or is it redundant once we have things integrated.

The factors smoke test is for testing factors-derive. Individual factors should either be "unit" tested in their own crates or as part of integration/e2e/conformance tests I think.

rylev:
@calebschoepp's question does make me wonder when we want to start actually putting these factors together into a `Factors` that can be run from Spin CLI. How hard would it be to add some sort of environment variable that swaps out the current `HostComponent` backed implementations for the new `Factors` backed implementation?

PR #2631, opened 2024-07-11
factors: Support old WASI -rc interfaces

lann:
Also add a basic OutboundHttpFactor test.

### DIFF
--- a/crates/factor-outbound-http/src/lib.rs
+++ b/crates/factor-outbound-http/src/lib.rs
@@ -1,5 +1,7 @@
 mod spin;
 mod wasi;
+mod wasi_2023_10_18;
+mod wasi_2023_11_10;
 
 use spin_factor_outbound_networking::{OutboundAllowedHosts, OutboundNetworkingFactor};
 use spin_factors::{
--- a/crates/factor-outbound-http/src/wasi_2023_10_18.rs
+++ b/crates/factor-outbound-http/src/wasi_2023_10_18.rs
@@ -1,0 +1,630 @@
+use anyhow::Result;
+use spin_factors::GetFactorState;
+use wasmtime::component::{Linker, Resource};
+use wasmtime_wasi_http::{WasiHttpImpl, WasiHttpView};
+
+mod latest {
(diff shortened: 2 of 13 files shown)


crates/factor-outbound-http/src/wasi_2023_10_18.rs
lann, line 1: All these `wasi_2023_*` files are lightly adapted from the matching spin-core impls.

crates/factor-outbound-http/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 mod spin;
 mod wasi;
+mod wasi_2023_10_18;
+mod wasi_2023_11_10;
```
rylev, line 4: Why do we need these in this crate as opposed to using the definitions in `spin_world`?

rylev (commented):
I'm not sure I fully understand the change here (i.e., why the bindings adapters need to be copied over). Could you provide more context on that?

crates/factor-outbound-http/src/lib.rs
lann: These adapters need customized bindgen options to be compatible with the base `wasmtime-wasi{,-http}` impls.

(The _real_ answer is that I just blindly adapted @alexcrichton's code from `spin-core`. :sweat_smile:)

crates/factor-outbound-http/tests/factor_test.rs
```diff
@@ -0,0 +1,59 @@
...
+    let req = Request::get("https://denied.test").body(Default::default())?;
+    let res = wasi_http.send_request(req, test_request_config());
+    let Err(err) = res else {
+        bail!("expected Err, got Ok");
+    };
+    assert!(matches!(err.downcast()?, ErrorCode::HttpRequestDenied));
+    Ok(())
+}
+
+fn test_request_config() -> OutgoingRequestConfig {
```
rylev, line 52: FYI: this could be a `const` instead of a function

rylev (approved):

PR #2632, opened 2024-07-11
Outbound pg factor

calebschoepp:


crates/factor-outbound-pg/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+authors = { workspace = true }
+edition = { workspace = true }
+
+[dependencies]
+anyhow = "1.0"
+native-tls = "0.2.11"
+postgres-native-tls = "0.5.0"
+spin-core = { path = "../core" }
+spin-factor-outbound-networking = { path = "../factor-outbound-networking" }
+spin-factor-variables = { path = "../factor-variables" }
```
rylev, line 13: I'm not sure we need variables at all (see my other comment), but if we do, and they're only be used in tests, can we move this to `dev-dependencies`?

crates/factor-outbound-pg/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+spin-factor-wasi = { path = "../factor-wasi" }
+spin-factors = { path = "../factors" }
+spin-world = { path = "../world" }
+table = { path = "../table" }
+tokio = { version = "1", features = ["rt-multi-thread"] }
+tokio-postgres = { version = "0.7.7" }
```
rylev, line 19:
```suggestion
tokio-postgres = "0.7.7"
```

crates/factor-outbound-pg/src/host.rs
```diff
@@ -0,0 +1,423 @@
...
+                    // The port we use is either:
+                    // * The port at the same index as the host
+                    // * The first port if there is only one port
+                    let port =
+                        ports
+                            .get(i)
+                            .or_else(|| if ports.len() == 1 { ports.get(1) } else { None });
+                    let port_str = port.map(|p| format!(":{}", p)).unwrap_or_default();
+                    let url = format!("{address}{port_str}");
+                    // TODO: Should I be unwrapping this?
```
rylev, line 55: Not sure if this is copied from the current implementation, but it does seem like `is_address_allowed` should be returning an `anyhow::Result<bool>` so that this unwrap isn't necessary.

crates/factor-outbound-pg/src/host.rs
```diff
@@ -0,0 +1,423 @@
...
+            .push(
+                build_client(address)
+                    .await
+                    .map_err(|e| v2::Error::ConnectionFailed(format!("{e:?}")))?,
+            )
+            .map_err(|_| v2::Error::ConnectionFailed("too many connections".into()))
+            .map(Resource::new_own)
+    }
+
+    async fn get_client(&mut self, connection: Resource<Connection>) -> Result<&Client, v2::Error> {
```
rylev, line 32: This isn't blocking merging this PR since this is how things work in the current implementation, but it would be really nice if the host implementation here didn't assume a `tokio_postgres::Client`. 

Instead we could have the connection pool keep track of `Box<dyn PgClient>` where `PgClient` in this case is some trait that abstracts over the underlying postgres calls. 

When running this library as part of the wider Spin CLI, the `PgClient` impl would be backed by a `tokio_postgres::Client` but in test it would be backed by some dummy implementation.

This is what we're doing in the `llm` implementation so that tests don't need to run an actual LLM.  

Thoughts?

crates/factor-outbound-pg/tests/factor_test.rs
```diff
@@ -0,0 +1,50 @@
...
+use spin_factor_variables::{StaticVariables, VariablesFactor};
+use spin_factor_wasi::{DummyFilesMounter, WasiFactor};
+use spin_factors::{anyhow, RuntimeFactors};
+use spin_factors_test::{toml, TestEnvironment};
+use spin_world::v2::postgres::HostConnection;
+
+#[derive(RuntimeFactors)]
+struct TestFactors {
+    wasi: WasiFactor,
+    variables: VariablesFactor,
```
rylev, line 13: Why do we include this here? Does `OutboundNetworkingFactor` unconditionally depend on `VariablesFactor`? I would have assumed that it would only depend on it if a variable was used but perhaps not?

crates/factor-outbound-pg/tests/factor_test.rs
```diff
@@ -0,0 +1,50 @@
...
+        .await;
+    let Err(err) = res else {
+        bail!("expected Err, got Ok");
+    };
+    assert!(matches!(err, ErrorConnectionFailed)); // TODO: Fix this assert
+
+    Ok(())
+}
+
+// TODO: Write more tests
```
rylev, line 50: Nit: such a todo isn't really meaningful in my opinion and should just be removed.

crates/factor-outbound-pg/tests/factor_test.rs
```diff
@@ -0,0 +1,50 @@
...
+    let mut state = env.build_instance_state(factors).await?;
+
+    let res = state
+        .pg
+        .open("postgres://postgres.test:5432/test".to_string())
+        .await;
+    let Err(err) = res else {
+        bail!("expected Err, got Ok");
+    };
+    assert!(matches!(err, ErrorConnectionFailed)); // TODO: Fix this assert
```
rylev, line 45: If you plan on leaving this TODO past this PR please provide more information on what needs to be "fixed" here.

crates/factors/tests/smoke.rs
```diff
@@ -20,6 +21,7 @@ struct Factors {
     outbound_networking: OutboundNetworkingFactor,
     outbound_http: OutboundHttpFactor,
     key_value: KeyValueFactor,
+    outbound_pg: OutboundPgFactor,
```
rylev, line 24: @lann I think we should decide here and now if adding new factors to the smoke test is necessary. If the point of the smoke test is to make sure `#[derive(Factors)]` works, I think we've already accomplished that, and this change shouldn't be necessary.

lann: Agreed - I don't think we need to keep updating this test.

crates/factor-outbound-pg/tests/factor_test.rs
lann: Yeah, the dependency is unconditional. Making it optional would require some refactoring in `AllowedHostsConfig` to detect whether `allowed_outbound_hosts` needs variables, and more generally might make factor dependencies more difficult to reason about.

crates/factor-outbound-pg/src/host.rs
lann: :+1: It could also be a defaulted generic param on the whole factor (i.e. `OutboundPgFactor<Client = tokio_postgres::Client>`.

crates/factor-outbound-pg/tests/factor_test.rs
calebschoepp: Oh yeah sorry, I'm not done with this PR so this is a TODO for myself to finish up this PR. I marked it draft, but I guess it wasn't super clear b/c I requested reviews.

Agreed that in the abstract this isn't a helpful PR.

crates/factor-outbound-pg/tests/factor_test.rs
calebschoepp: Same as above.

crates/factor-outbound-pg/src/host.rs
calebschoepp: Filed an issue to do this work in a followup PR #2649

calebschoepp:
@lann @rylev this is ready for another round of reviews.

lann (approved):

rylev (approved):

crates/factor-outbound-pg/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[dependencies]
+anyhow = "1.0"
+native-tls = "0.2.11"
+postgres-native-tls = "0.5.0"
+spin-core = { path = "../core" }
+spin-factor-outbound-networking = { path = "../factor-outbound-networking" }
+spin-factors = { path = "../factors" }
+spin-world = { path = "../world" }
+table = { path = "../table" }
+tokio = { version = "1", features = ["rt-multi-thread"] }
```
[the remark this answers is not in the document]
calebschoepp: We use it for the spawn

PR #2634, opened 2024-07-11
Better error if component from registry has regrettable version

itowlson:
If you have a component sourced from a registry, the version has to be semver.  The previous error message was bad.

Before:

```
ivan@hestia:~/testing/wkgtest$ spin up
Error: TOML parse error at line 15, column 10
   |
15 | source = { registry = "registrytest-vfztdiyy.fermyon.app", package = "component:ccspintest", version = "canary" }
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
data did not match any variant of untagged enum ComponentSource
```

After:

```
ivan@hestia:~/testing/wkgtest$ spin up

Caused by:
   1: Failed to load component `wkgtest`
   2: Failed to load Wasm source "component:ccspintest@canary" from registry Registry(registrytest-vfztdiyy.fermyon.app)
   3: Component wkgtest specifies an invalid semantic version ("canary") for its package version
   4: unexpected character 'c' while parsing major version number
```

### DIFF
--- a/crates/manifest/src/schema/common.rs
+++ b/crates/manifest/src/schema/common.rs
@@ -39,7 +39,7 @@ pub enum ComponentSource {
         /// `package = "example:component"`
         package: PackageRef,
         /// `version = "1.2.3"`
-        version: semver::Version,
+        version: String,
     },
 }
 
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -156,7 +156,7 @@ impl LocalLoader {
             .take();
 
         let source = self
-            .load_component_source(component.source.clone())
+            .load_component_source(id, component.source.clone())
             .await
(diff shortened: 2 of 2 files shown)


rylev (approved):

crates/manifest/src/schema/common.rs
```diff
@@ -39,7 +39,7 @@ pub enum ComponentSource {
         /// `package = "example:component"`
         package: PackageRef,
         /// `version = "1.2.3"`
-        version: semver::Version,
+        version: String,
```
lann, line 42: Another option here would be a custom `deserialize_with`; the thing I appreciate about using `Version` here is that you get nice location information in errors.

PR #2636, opened 2024-07-12
Add some more documentation to factors

rylev:
First pass just giving a bit more context on the various pieces of the factors system. I think this will mainly help for hovering over types in IDEs to remind users what things are. 

Besides documentation one code change is made: `RuntimeFactors::build_store_data` => `RuntimeFactors::build_instance_state`. I know that the value returned from this method is what gets passed to `wasmtime::Store::new` which refers to it as `data`, but given the rest of the factors framework refers to this as instance state (which is much more descriptive than wasmtime's nomenclature), I think this is a better name.

### DIFF
--- a/crates/factors/src/factor.rs
+++ b/crates/factors/src/factor.rs
@@ -6,15 +6,27 @@ use crate::{
     RuntimeFactors,
 };
 
+/// A contained (i.e., "factored") piece of runtime functionality.
 pub trait Factor: Any + Sized {
+    /// The particular runtime configuration relevant to this factor.
+    ///
+    /// Runtime configuration allows for user provided customization of the
+    /// factor's behavior on a per app basis.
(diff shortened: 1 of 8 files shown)


crates/factors/src/factor.rs
```diff
@@ -7,15 +7,20 @@ use crate::{
     InstanceBuilders, Linker, PrepareContext, RuntimeConfigSource, RuntimeFactors,
 };
 
+/// A contained (i.e., "factored") piece of runtime functionality.
 pub trait Factor: Any + Sized {
```
lann, line 11: I wrote better docs for this trait [in the SIP](https://github.com/fermyon/spin/pull/2518/files#diff-3ea0e4a724fed06ab763f915e6cdcbbabf8c84bc63e90882cd6da1515b55a768R74) which could be adapted.

lann (approved):

PR #2639, opened 2024-07-12
factors: Add Error type

lann:
Adds a custom `Error` enum.

One effect here is more-helpful errors when inter-factor dependencies aren't met:

```

Error: spin_factor_outbound_http::OutboundHttpFactor prepare failed: no such factor: spin_factor_outbound_networking::OutboundNetworkingFactor

### DIFF
--- a/crates/factors/src/lib.rs
+++ b/crates/factors/src/lib.rs
@@ -22,10 +22,55 @@ pub type Linker<T> = wasmtime::component::Linker<<T as RuntimeFactors>::Instance
 pub type App = spin_app::App<'static, spin_app::InertLoader>;
 pub type AppComponent<'a> = spin_app::AppComponent<'a, spin_app::InertLoader>;
 
-// TODO: Add a real Error type
-pub type Result<T> = wasmtime::Result<T>;
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: crates/factors/src/lib.rs [hunk 1]
```diff
@@ -22,10 +22,55 @@ pub type Linker<T> = wasmtime::component::Linker<<T as RuntimeFactors>::Instance
...
+    #[error("no such factor: {0}")]
+    NoSuchFactor(&'static str),
+    #[error("{factor} requested already-consumed key {key:?}")]
+    RuntimeConfigReusedKey { factor: &'static str, key: String },
+    #[error("runtime config error: {0}")]
+    RuntimeConfigSource(#[source] anyhow::Error),
+    #[error("unused runtime config key(s): {}", keys.join(", "))]
+    RuntimeConfigUnusedKeys { keys: Vec<String> },
+    #[error("{factor} {method} failed: {source}")]
+    RuntimeFactorError {
```
lann, line 40: It might ultimately be better to split this out into 4 separate errors e.g. `FactorInitError`, `FactorPrepareError`.

calebschoepp (approved):

crates/factors/src/lib.rs
rylev: I would certainly be in favor of this!

PR #2643, opened 2024-07-15
Improve non-interactive behaviour of `spin new`

itowlson:
An alternative approach to the problem identified in #2627.  With this PR:

* In a non-interactive environment (not a terminal), `spin new` and `spin add` will:
  * Not prompt for parameters (must be supplied via `-value`, `--values-file`, or `--accept-defaults`
  * Not prompt about generating into a non-empty directory (will fail and error out by default)
  * The implementation uses the established `interaction::Silent` strategy, which already gets exercised via the unit tests
* `spin new` and `spin add` now support a `--allow-overwrite` flag which will allow generating into a non-empty directory.  In a terminal, this skips the prompt; in a non-terminal, it changes the behaviour.

The `--allow-overwrite` flag has a `--allow-overwrites` alias because I kept typing the latter when testing _despite having defined the singular form not two minutes before_ that is why.

The PR has unit tests for the non-interactive case; obviously it's not very practical to includes tests for the interactive case but it seemed to work when I tried it!

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -782,6 +782,7 @@ mod tests {
             values,
             accept_defaults: false,
             no_vcs: false,
+            allow_overwrite: false,
         };
 
         template.run(options).silent().await.unwrap();
@@ -816,6 +817,7 @@ mod tests {
             values,
             accept_defaults: true,
             no_vcs: false,
+            allow_overwrite: false,
         };
 
         template.run(options).silent().await.unwrap();
@@ -882,6 +884,7 @@ mod tests {
                 values,
                 accept_defaults: false,
                 no_vcs: false,
+                allow_overwrite: false,
             };
 
             template.run(options).silent().await.unwrap();
@@ -910,6 +913,7 @@ mod tests {
                 values,
                 accept_defaults: false,
                 no_vcs: false,
+                allow_overwrite: false,
             };
 
             template.run(options).silent().await.unwrap();
@@ -935,6 +939,7 @@ mod tests {
                 values,
                 accept_defaults: false,
                 no_vcs: false,
+                allow_overwrite: false,
             };
 
             template.run(options).silent().await.unwrap();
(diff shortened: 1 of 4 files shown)


crates/templates/src/manager.rs
```diff
@@ -1387,6 +1403,104 @@ mod tests {
...
+            .expect_err("generate into existing dir should have failed");
+
+        assert!(tokio::fs::read_to_string(&manifest_path)
+            .await
+            .unwrap()
+            .contains("cookies"));
+    }
+
+    #[tokio::test]
+    async fn can_generate_over_existing_files_if_so_configured() {
```
rylev, line 1456: nit: there's a lot of set up for these tests that is indentical. Would it make sense to test both scenarios in the same test perhaps by doing the "cookies" test, then deleting the spin.toml, and trying again?

src/commands/new.rs
```diff
@@ -66,6 +67,15 @@ pub struct TemplateNewCommandCore {
     /// An optional argument that allows to skip creating .gitignore
     #[clap(long = "no-vcs", takes_value = false)]
     pub no_vcs: bool,
+
+    /// If the output directory already contains files, generate the new files into
+    /// it without confirming, overwriting any existing files with the same names.
+    #[clap(
+        long = "allow-overwrite",
```
rylev, line 74: I'm not 100% sold on the name as it can somewhat beg the question of *what* we are allowing the overwrite of, but I can't think of a better alternative so I'm fine with it as is.

rylev (approved):
Looks good! Have some small nits and thoughts but nothing blocking.

crates/templates/src/manager.rs
itowlson: Yes, there's a lot of shared test setup across the template tests.  I'd prefer to keep the individual tests specific, but as a later pass we should refactor the entire test suite to reduce duplication/copy-paste and make the "what is specific about this setup and about the expected results" pop a lot more.  So I am going to punt on this for now but I do totally accept your point!

PR #2646, opened 2024-07-15
factor: Add key value Redis factor

kate-goldenring:
We could also add these changes to `key-value-redis` crate and later rename that crate to `factor-key-value-redis`. @lann which would you prefer?

Also, adds support for configuring the default KV store.

### DIFF
--- a/crates/factor-key-value/src/lib.rs
+++ b/crates/factor-key-value/src/lib.rs
@@ -20,11 +20,19 @@ use store::{store_from_toml_fn, StoreFromToml};
 
 pub use store::MakeKeyValueStore;
 
-#[derive(Default)]
 pub struct KeyValueFactor {
     store_types: HashMap<&'static str, StoreFromToml>,
+    default_store_type: &'static str,
 }
 
+impl Default for KeyValueFactor {
+    fn default() -> KeyValueFactor {
+        KeyValueFactor {
+            store_types: HashMap::default(),
+            default_store_type: "spin",
+        }
+    }
+}
 impl KeyValueFactor {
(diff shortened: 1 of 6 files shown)


crates/factor-key-value/src/lib.rs
```diff
@@ -20,11 +20,19 @@ use store::{store_from_toml_fn, StoreFromToml};
 
 pub use store::MakeKeyValueStore;
 
-#[derive(Default)]
 pub struct KeyValueFactor {
     store_types: HashMap<&'static str, StoreFromToml>,
+    default_store_type: &'static str,
```
lann, line 25: Just had a chat with @rylev about how to handle this in the sqlite factor; his implementation is very different from this one which I think is fine for now but we should try to converge on a single pattern "pretty soon".

lann (approved):

lann:
> We could also add these changes to key-value-redis crate and later rename that crate to factor-key-value-redis. @lann which would you prefer?

I actually think what you have here is ideal for review. A separate refactor PR that just shuffles code around is easier to deal with later.

crates/factor-key-value/src/lib.rs
rylev: I'm obviously biased since I wrote it, but I do like the other implementation for being completely Spin CLI agnostic. A Spin CLI specific approach can be bundled into the crate for convenience, but the main implementation has no specific knowledge of how Spin CLI handles things. 

Unless there's objections, I can open a PR that moves this implementation to mirror that approach so we can discuss whether we think it's the right way to go.

kate-goldenring: @rylev that sounds good to me. I like how your approach is agnostic and doesn't define what "string" is default.

PR #2648, opened 2024-07-15
Factor sqlite

rylev:
Adds a factor for SQLite

### DIFF
--- a/crates/factor-sqlite/src/runtime_config.rs
+++ b/crates/factor-sqlite/src/runtime_config.rs
@@ -24,11 +24,14 @@ pub struct StoreConfig {
 }
 
 /// Resolves some piece of runtime configuration to a connection pool
-pub trait RuntimeConfigResolver {
-    /// Get a connection pool for a given runtime configuration type and the raw configuration.
+pub trait RuntimeConfigResolver: Send + Sync {
+    /// Get a connection pool for a given database kind and the raw configuration.
+    ///
+    /// `database_kind` is equivalent to the `type` field in the
+    /// `[sqlite_database.$databasename]` runtime configuration table.
     fn get_pool(
         &self,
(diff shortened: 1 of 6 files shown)


crates/factor-sqlite/src/runtime_config.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[serde(flatten)]
+    pub config: toml::Table,
+}
+
+/// Resolves some piece of runtime configuration to a connection pool
+pub trait RuntimeConfigResolver {
+    /// Get a connection pool for a given runtime configuration type and the raw configuration.
+    fn get_pool(
+        &self,
+        r#type: &str,
```
lann, line 31: Maybe `store_type` here?

crates/factor-sqlite/src/runtime_config.rs
```diff
@@ -0,0 +1,39 @@
...
+#[derive(Deserialize)]
+pub struct StoreConfig {
+    #[serde(rename = "type")]
+    pub type_: String,
+    #[serde(flatten)]
+    pub config: toml::Table,
+}
+
+/// Resolves some piece of runtime configuration to a connection pool
+pub trait RuntimeConfigResolver {
```
lann, line 27: Would it make sense for this crate to provide an impl that "does what spin does today"?

rylev: I think so - though I think it might make sense to put it behind a feature flag since the implementation isn't trivial, and if a user of this crate doesn't want it we shouldn't hurt their compile times. 

I'd like to do that in a separate PR though since I think this will open up new questions, and I'd rather not have to worry about the base implementation becoming out of date.

lann (approved):

PR #2651, opened 2024-07-15
Abstract pg client

### LINKED ISSUE #2649: Abstract pg client type in factor-outbound-pg
This is followup work to #2632

calebschoepp:
Closes #2649

### DIFF
--- a/crates/factor-outbound-pg/src/client.rs
+++ b/crates/factor-outbound-pg/src/client.rs
@@ -1,0 +1,284 @@
+use anyhow::{anyhow, Result};
+use native_tls::TlsConnector;
+use postgres_native_tls::MakeTlsConnector;
+use spin_world::async_trait;
+use spin_world::v2::postgres::{self as v2};
+use spin_world::v2::rdbms_types::{Column, DbDataType, DbValue, ParameterValue, RowSet};
+use tokio_postgres::types::Type;
+use tokio_postgres::{config::SslMode, types::ToSql, Row};
+use tokio_postgres::{Client as TokioClient, NoTls, Socket};
+
+#[async_trait]
+pub trait Client {
+    async fn build_client(address: &str) -> Result<Self>
+    where
+        Self: Sized;
+
+    async fn execute(
+        &self,
+        statement: String,
+        params: Vec<ParameterValue>,
+    ) -> Result<u64, v2::Error>;
+
+    async fn query(
+        &self,
+        statement: String,
+        params: Vec<ParameterValue>,
+    ) -> Result<RowSet, v2::Error>;
+}
+
+#[async_trait]
+impl Client for TokioClient {
+    async fn build_client(address: &str) -> Result<Self>
+    where
+        Self: Sized,
+    {
+        let config = address.parse::<tokio_postgres::Config>()?;
+
+        tracing::debug!("Build new connection: {}", address);
+
+        if config.get_ssl_mode() == SslMode::Disable {
+            let (client, connection) = config.connect(NoTls).await?;
+            spawn_connection(connection);
+            Ok(client)
+        } else {
+            let builder = TlsConnector::builder();
+            let connector = MakeTlsConnector::new(builder.build()?);
+            let (client, connection) = config.connect(connector).await?;
+            spawn_connection(connection);
+            Ok(client)
+        }
+    }
+
+    async fn execute(
+        &self,
+        statement: String,
+        params: Vec<ParameterValue>,
+    ) -> Result<u64, v2::Error> {
+        let params: Vec<&(dyn ToSql + Sync)> = params
+            .iter()
+            .map(to_sql_parameter)
+            .collect::<Result<Vec<_>>>()
+            .map_err(|e| v2::Error::ValueConversionFailed(format!("{:?}", e)))?;
+
+        self.execute(&statement, params.as_slice())
+            .await
+            .map_err(|e| v2::Error::QueryFailed(format!("{:?}", e)))
+    }
+
+    async fn query(
+        &self,
+        statement: String,
+        params: Vec<ParameterValue>,
+    ) -> Result<RowSet, v2::Error> {
+        let params: Vec<&(dyn ToSql + Sync)> = params
+            .iter()
+            .map(to_sql_parameter)
+            .collect::<Result<Vec<_>>>()
+            .map_err(|e| v2::Error::BadParameter(format!("{:?}", e)))?;
+
+        let results = self
+            .query(&statement, params.as_slice())
+            .await
+            .map_err(|e| v2::Error::QueryFailed(format!("{:?}", e)))?;
+
+        if results.is_empty() {
+            return Ok(RowSet {
+                columns: vec![],
+                rows: vec![],
+            });
+        }
+
+        let columns = infer_columns(&results[0]);
+        let rows = results
+            .iter()
+            .map(convert_row)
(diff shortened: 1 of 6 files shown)


crates/factor-outbound-pg/src/client.rs
```diff
@@ -0,0 +1,75 @@
+use native_tls::TlsConnector;
+use postgres_native_tls::MakeTlsConnector;
+use spin_world::async_trait;
+use tokio_postgres::{config::SslMode, types::ToSql, Error, Row, ToStatement};
+use tokio_postgres::{Client as PgClient, NoTls, Socket};
```
rylev, line 5: Nit

```suggestion
use tokio_postgres::{Client as TokioClient, NoTls, Socket};
```

I prefer `TokioClient` as a name since presumably all clients are pg clients, no?

crates/factor-outbound-pg/src/client.rs
```diff
@@ -0,0 +1,75 @@
...
+        } else {
+            async move {
+                let builder = TlsConnector::builder();
+                let connector = MakeTlsConnector::new(builder.build()?);
+                let (client, connection) = config.connect(connector).await?;
+                spawn(connection);
+                Ok(client)
+            }
+            .await
+        }
```
rylev, line 48: Nit: this removes a bit of redundant boiler plate and makes the code a bit easier to read IMO.
```suggestion
        Ok(async move {
            if config.get_ssl_mode() == SslMode::Disable {
                let (client, connection) = config.connect(NoTls).await?;
                spawn(connection);
                client
            } else {
                let builder = TlsConnector::builder();
                let connector = MakeTlsConnector::new(builder.build()?);
                let (client, connection) = config.connect(connector).await?;
                spawn(connection);
                client
            }
        }.await)
```

crates/factor-outbound-pg/src/client.rs
```diff
@@ -0,0 +1,75 @@
...
+pub trait Client {
+    async fn build_client(address: &str) -> anyhow::Result<Self>
+    where
+        Self: Sized;
+
+    async fn execute<T>(&self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<u64, Error>
+    where
+        T: ?Sized + ToStatement + Sync + Send;
+
+    async fn query<T>(&self, query: &T, params: &[&(dyn ToSql + Sync)]) -> Result<Vec<Row>, Error>
```
rylev, line 17: Seems strange that we're forcing all implementors to return a `Row` type that is defined in `tokio_postgres` (along with other uses of `tokio_postgres` types). I would have assumed we would define this crate in terms of the wit types. This would mean conversion from implementation specific types to wit types would be contained inside the `Client` impl which makes sense to me.

crates/factor-outbound-pg/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+            .allowed_hosts();
+        Ok(InstanceState {
+            allowed_hosts,
+            connections: Default::default(),
+        })
+    }
+}
+
+impl<C> Default for OutboundPgFactor<C> {
+    fn default() -> Self {
```
rylev, line 53: Nit: I like having a `new` static function in addition to `Default`. The `new` function would just call `Self::default`. I like it, because it shows up more prominently in the docs and is therefore IMO more discoverable.

crates/factor-outbound-pg/tests/factor_test.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    state
+        .pg
+        .query(connection, "SELECT * FROM test".to_string(), vec![])
+        .await?;
+
+    Ok(())
+}
+
+pub struct MockClient {}
```
rylev, line 118: Is the plan to eventually expand this mock implementation to track how it's been called and let the tests assert that the implementation has been called properly?

crates/factor-outbound-pg/src/host.rs
```diff
@@ -0,0 +1,379 @@
...
+                    // The port we use is either:
+                    // * The port at the same index as the host
+                    // * The first port if there is only one port
+                    let port =
+                        ports
+                            .get(i)
+                            .or_else(|| if ports.len() == 1 { ports.get(1) } else { None });
+                    let port_str = port.map(|p| format!(":{}", p)).unwrap_or_default();
+                    let url = format!("{address}{port_str}");
+                    // TODO: Should I be unwrapping this?
```
rylev, line 53:
```suggestion
```

crates/factor-outbound-pg/tests/factor_test.rs
calebschoepp: Yep

crates/factor-outbound-pg/src/client.rs
calebschoepp: I think you're probably right. I spent some time looking at how I would do this and I was stumped by one thing though. The `v1` and `v2` interfaces have a slightly different shape (return different errors and take different params). How should I reconcile that in a single client interface?

rylev: Simply by mapping the params and return values. We do this in SQLite: https://github.com/fermyon/spin/blob/14f8a5897f17c7cfc053d2b9f59e1843b6c9a72e/crates/factor-sqlite/src/host.rs#L108-L116

calebschoepp:
This is ready for review. Feel free to take it over and merge it while I'm gone on vacation @lann @rylev

crates/factor-outbound-pg/tests/factor_test.rs
```diff
@@ -1,39 +1,51 @@
...
     wasi: WasiFactor,
     variables: VariablesFactor,
     networking: OutboundNetworkingFactor,
-    pg: OutboundPgFactor,
+    pg: OutboundPgFactor<MockClient>,
+}
+
+fn factors() -> Result<TestFactors> {
+    let mut f = TestFactors {
+        wasi: WasiFactor::new(DummyFilesMounter),
```
lann, line 25: FYI this shouldn't be necessary any more; `OutboundNetworkingFactor`'s dependency on it is now optional.

crates/factor-outbound-pg/src/client.rs
```diff
@@ -0,0 +1,287 @@
...
+                Ok(client)
+            } else {
+                let builder = TlsConnector::builder();
+                let connector = MakeTlsConnector::new(builder.build()?);
+                let (client, connection) = config.connect(connector).await?;
+                spawn(connection);
+                Ok(client)
+            }
+        }
+        .await
```
lann, line 53: Is this `async { ... }.await` wrapper necessary? :thinking:

crates/factor-outbound-pg/src/client.rs
```diff
@@ -0,0 +1,287 @@
...
+            .iter()
+            .map(convert_row)
+            .collect::<Result<Vec<_>, _>>()
+            .map_err(|e| v2::Error::QueryFailed(format!("{:?}", e)))?;
+
+        Ok(RowSet { columns, rows })
+    }
+}
+
+fn spawn<T>(connection: tokio_postgres::Connection<Socket, T>)
```
lann, line 106: I was confused for a minute thinking this was just `use tokio::spawn`; might help to give it a distinct name:

```suggestion
fn spawn_connection<T>(connection: tokio_postgres::Connection<Socket, T>)
```

lann (approved):
Minor notes

PR #2652, opened 2024-07-16
Add runtime config impl that Spin uses to factor-sqlite

rylev:
Provide an optional default implementation for sqlite's handling of runtime config that mirrors how it works in Spin CLI.


crates/factor-sqlite/src/lib.rs
```diff
@@ -114,13 +114,34 @@ pub struct AppState {
...
-/// A store of connections for all accessible databases for an application
+/// A pool of connections for a particular SQLite database
 #[async_trait]
 pub trait ConnectionPool: Send + Sync {
     /// Get a `Connection` from the pool
     async fn get_connection(&self) -> Result<Arc<dyn Connection + 'static>, v2::Error>;
 }
 
+/// A simple [`ConnectionPool`] that always creates a new connection.
+pub struct SimpleConnectionPool(
```
lann, line 125: Is this wrapper needed? Could you directly `impl<F: Fn() -> ...> ConnectionPool for F`?

crates/factor-sqlite/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,237 @@
...
+    state_dir: PathBuf,
+    base_path: Option<PathBuf>,
+}
+
+impl SpinSqliteRuntimeConfig {
+    /// Create a new `SpinSqliteRuntimeConfig`
+    ///
+    /// This takes as arguments:
+    /// * the state directory path (i.e., the path to the `.spin` file). This
+    /// is used to derive the default path a local SQLite database file.
```
lann, line 31:
```suggestion
    /// * the state directory path (i.e., the path to the `.spin` directory). This
    /// is used to derive the default path for a local SQLite database file.
```

crates/factor-sqlite/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,237 @@
...
+    /// * the base path from which relative paths referenced in configuration are resolved
+    /// (this should most likely be the path to the runtime-config file). If
+    /// `None`, the current working directory is used.
```
lann, line 34: > If `None`, the current working directory is used.

As a pattern, I was thinking maybe this should bail on relative paths. :shrug:

crates/factor-sqlite/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,237 @@
...
+    base_path: Option<PathBuf>,
+}
+
+impl SpinSqliteRuntimeConfig {
+    /// Create a new `SpinSqliteRuntimeConfig`
+    ///
+    /// This takes as arguments:
+    /// * the state directory path (i.e., the path to the `.spin` file). This
+    /// is used to derive the default path a local SQLite database file.
+    /// * the base path from which relative paths referenced in configuration are resolved
```
lann, line 32: I was surprised to see `.parent()` below and had to come back and reread this; I think it would be less surprising if the caller was responsible for `.parent` instead.

crates/factor-sqlite/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,237 @@
...
+        let factory = move || {
+            let connection = spin_sqlite_inproc::InProcConnection::new(location.clone())?;
+            Ok(Arc::new(connection) as _)
+        };
+        Ok(SimpleConnectionPool::new(factory))
+    }
+}
+
+/// Resolve a relative path against an optional base path.
+fn resolve_relative_path(path: &Path, base_path: Option<&Path>) -> anyhow::Result<PathBuf> {
```
kate-goldenring, line 182: This could probably moved to a shared place for factors as it is a replacement for `resolve_config_path` which both KV and SQLite use

crates/factor-sqlite/src/lib.rs
rylev: Yea, I find it easier to read with the simple wrapper struct, but I don't feel strongly either way.

crates/factor-sqlite/src/runtime_config/spin.rs
rylev: This is a good point. I think the question is _where_. It's a highly generic yet super specific piece of code... I think until it becomes obvious where this should live, duplicating isn't the worst thing in the world.

rylev:
@lann @kate-goldenring I ended up simplifying path handling quite a bit and forcing the caller to get their paths in order: https://github.com/fermyon/spin/pull/2652/commits/717c743ef55ac4abeb708f1226b93cf8919884b4

Let me know what you think.

lann (approved):

kate-goldenring (approved):
LGTM. I can work on moving KV closer to this approach.

PR #2653, opened 2024-07-16
factors: Enhance OutboundNetworkingFactor

lann:
- Make dependency on WasiFactor optional
- Fix wasi socket addr config and add test
- Add support for explicit 'tcp' and 'udp' schemes in allowed_outbound_hosts for wasi sockets

### DIFF
--- a/crates/factor-outbound-networking/src/lib.rs
+++ b/crates/factor-outbound-networking/src/lib.rs
@@ -5,10 +5,10 @@ use futures_util::{
     FutureExt,
 };
 use spin_factor_variables::VariablesFactor;
-use spin_factor_wasi::WasiFactor;
+use spin_factor_wasi::{SocketAddrUse, WasiFactor};
 use spin_factors::{
     anyhow::{self, Context},
-    ConfigureAppContext, Factor, FactorInstanceBuilder, InstanceBuilders, PrepareContext,
+    ConfigureAppContext, Error, Factor, FactorInstanceBuilder, InstanceBuilders, PrepareContext,
     RuntimeFactors,
 };
 use spin_outbound_networking::{AllowedHostsConfig, ALLOWED_HOSTS_KEY};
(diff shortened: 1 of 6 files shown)


crates/factor-outbound-networking/src/lib.rs
```diff
@@ -67,31 +67,36 @@ impl Factor for OutboundNetworkingFactor {
...
+                    async move {
+                        match hosts_future.await {
+                            Ok(allowed_hosts) => {
+                                // TODO: validate against existing spin-core behavior
+                                let scheme = match addr_use {
+                                    SocketAddrUse::TcpBind => return false,
+                                    SocketAddrUse::TcpConnect => "tcp",
+                                    SocketAddrUse::UdpBind | SocketAddrUse::UdpConnect | SocketAddrUse::UdpOutgoingDatagram => "udp",
+                                };
+                                spin_outbound_networking::check_url(&addr.to_string(),scheme, &allowed_hosts)
```
lann, line 86: This is a significant change from the way outbound sockets are checked today. It skips the address "Pool" and checks against the allowed hosts directly. This required adding an actual scheme for the checks, so it made sense to me to support "tcp" and "udp" schemes to allow filtering by those protocols.

crates/factor-outbound-networking/tests/factor_test.rs
```diff
@@ -0,0 +1,55 @@
...
+            network
+                .check_socket_addr(not_allowed.parse().unwrap(), SocketAddrUse::TcpConnect)
+                .await
+                .unwrap_err()
+                .kind(),
+            std::io::ErrorKind::PermissionDenied
+        );
+    }
+    Ok(())
+}
```
rylev, line 55: Can you add a test confirming that not having a `wasi-factor` works?

lann: Done

rylev (approved):

PR #2656, opened 2024-07-16
factors: handle default kv store resolution

kate-goldenring:
Inspired by SQLite implementation https://github.com/fermyon/spin/pull/2648

### DIFF
--- a/crates/factor-key-value/src/lib.rs
+++ b/crates/factor-key-value/src/lib.rs
@@ -1,5 +1,5 @@
 mod runtime_config;
-pub mod spin_cli_resolver;
+pub mod delegating_resolver;
 mod store;
 
 use std::{
@@ -14,8 +14,8 @@ use spin_factors::{
     PrepareContext, RuntimeFactors,
 };
 use spin_key_value::{
-    CachingStoreManager, DelegatingStoreManager, DefaultManagerGetter, KeyValueDispatch, StoreManager,
-    KEY_VALUE_STORES_KEY,
+    CachingStoreManager, DefaultManagerGetter, DelegatingStoreManager, KeyValueDispatch,
+    StoreManager, KEY_VALUE_STORES_KEY,
 };
 pub use store::MakeKeyValueStore;
 
@@ -70,7 +70,7 @@ impl Factor for KeyValueFactor {
         }
         let resolver_clone = self.runtime_config_resolver.clone();
         let default_fn: DefaultManagerGetter =
-            Arc::new(move |label| resolver_clone.default(label));
+            Arc::new(move |label| resolver_clone.default_store(label));
 
         let delegating_manager = DelegatingStoreManager::new(store_managers, default_fn);
         let caching_manager = CachingStoreManager::new(delegating_manager);
@@ -89,7 +89,7 @@ impl Factor for KeyValueFactor {
                 // TODO: port nicer errors from KeyValueComponent (via error type?)
                 ensure!(
                     store_manager_manager.is_defined(label)
-                        || self.runtime_config_resolver.default(label).is_some(),
+                        || self.runtime_config_resolver.default_store(label).is_some(),
                     "unknown key_value_stores label {label:?} for component {component_id:?}"
                 );
             }
--- a/crates/factor-key-value/src/lib.rs
+++ b/crates/factor-key-value/src/lib.rs
@@ -1,5 +1,5 @@
-mod runtime_config;
 pub mod delegating_resolver;
+mod runtime_config;
 mod store;
 
 use std::{
--- a/crates/factor-key-value/src/lib.rs
+++ b/crates/factor-key-value/src/lib.rs
@@ -1,3 +1,5 @@
+mod runtime_config;
+pub mod spin_cli_resolver;
 mod store;
 
 use std::{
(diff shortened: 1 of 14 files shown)


crates/factor-key-value/src/spin_cli_resolver.rs
```diff
@@ -0,0 +1,47 @@
+use crate::runtime_config::RuntimeConfigResolver;
+use crate::store::{store_from_toml_fn, MakeKeyValueStore, StoreFromToml};
+use spin_key_value::StoreManager;
+use std::{collections::HashMap, sync::Arc};
+
+#[derive(Default)]
+pub struct SpinCliRuntimeConfigResolver {
```
lann, line 7: I think with a tweak (below) this could be a general purpose "delegating resolver" e.g. `DelegatingRuntimeConfigResolver`

crates/factor-key-value/src/spin_cli_resolver.rs
```diff
@@ -0,0 +1,47 @@
...
+        store_from_toml(config)
+    }
+
+    fn default(&self, label: &str) -> Option<Arc<dyn StoreManager>> {
+        if label == "default" {
+            self.get_store("spin", toml::value::Table::new()).ok()
+        } else {
+            None
+        }
+    }
```
lann, line 46: To make this not-`spin up`-specific, maybe we could make this configurable, with a new method on the factor, e.g. `kv_factor.default_store("spin", toml::value::Table::new())`. Note also that the `Table` there will actually need a `path` for the default store derived by the runtime anyway.

crates/factor-key-value/src/spin_cli_resolver.rs
kate-goldenring: I agree that would be nice to make it generalized with that change too. Only other consideration is how to set the default DB directory. It may make sense to also set that in the resolver

crates/factor-key-value/src/spin_cli_resolver.rs
kate-goldenring: The tricky thing here is that we want to abstract away what the default label is too. For Spin it happens to be "default". Furthermore, a runtime may want to establish multiple defaults, say "cache" is always the local sqlite impl.

kate-goldenring: @lann what do you think of this 7f7ebc33a63f9393670daadfd96bb034394497f3

crates/factor-key-value/src/lib.rs
```diff
@@ -112,7 +88,8 @@ impl Factor for KeyValueFactor {
             for label in &key_value_stores {
                 // TODO: port nicer errors from KeyValueComponent (via error type?)
                 ensure!(
-                    store_manager.is_defined(label),
+                    store_manager_manager.is_defined(label)
+                        || self.runtime_config_resolver.default(label).is_some(),
```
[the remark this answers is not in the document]
kate-goldenring: We could add a function called `default_exists` or `validate_default` that just does an existence check if we have concerns about performance affects here

crates/factor-key-value/src/spin_cli_resolver.rs
kate-goldenring: I am now trying to figure out the best way to pass the default store opts when we make a store: https://github.com/kate-goldenring/spin/blob/41a5796603c80617eebed2eac7a12d8f475bfaae/crates/trigger/src/runtime_config/key_value.rs#L91-L92

kate-goldenring: Think I got it all situated: 318e1660113d45cbcae8b202a47e174b14a938d9

kate-goldenring:
@lann @rylev I'll add doc comments and tests in a subsequent PR after merging this pending your review @lann

crates/factor-key-value-sqlite/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "spin-factor-key-value-sqlite"
```
rylev, line 2: nit: I'm fine with the crate name (even if it is a bit long), but I'm not sure I like that the directory inside the `crates` dir starts with `factor` even though there is not factor impl in it. I guess eventually this will just merge with `spin-key-value-sqlite` and we can use that name, right?

crates/factor-key-value-sqlite/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+use serde::{Deserialize, Serialize};
+use spin_factor_key_value::MakeKeyValueStore;
+use spin_key_value_sqlite::{DatabaseLocation, KeyValueSqlite};
+
+pub struct SpinKeyValueStore {
+    base_path: PathBuf,
+}
+
+impl SpinKeyValueStore {
+    pub fn new(base_path: Option<PathBuf>) -> anyhow::Result<Self> {
```
rylev, line 16: Let's make sure to doc this stuff as well (e.g., that if `base_path` is `None`, current_dir will be used.

rylev (approved):

crates/factor-key-value-sqlite/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+use anyhow::Context;
+use serde::{Deserialize, Serialize};
+use spin_factor_key_value::MakeKeyValueStore;
+use spin_key_value_sqlite::{DatabaseLocation, KeyValueSqlite};
+
+pub struct SpinKeyValueStore {
+    base_path: PathBuf,
+}
+
+impl SpinKeyValueStore {
```
lann, line 15: Its confusing that the crate is `-sqlite` but the store is `Spin`. Given the runtime config `type = "spin"` maybe we can live with `spin-factor-key-value-spin`?

kate-goldenring: Good call. Just adjusted it.

lann (approved):

crates/factor-key-value-sqlite/Cargo.toml
itowlson: Shouldn't we keep the `spin-` prefix to allow for eventual publishing on crates.io?

PR #2657, opened 2024-07-17
Deduplicate template tests

itowlson:
The template installation and execution tests had grown a very large amount of repeated code, which made it hard to see what was different, and therefore any given test was verifying.

This PR doesn't entirely fix that - in particular, there's a lot of file path clutter - but it makes a start on some common patterns.  It's offered as an incremental improvement rather than as a perfect end state!

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -468,21 +468,64 @@ mod tests {
         PathBuf::from(crate_dir).join("tests")
     }
 
+    // A template manager in a temp dir.  This exists to ensure the lifetime
+    // of the TempDir
+    struct TempManager {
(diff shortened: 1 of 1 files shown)


crates/templates/src/manager.rs
```diff
@@ -468,21 +468,104 @@ mod tests {
...
+
+    impl std::ops::Deref for TempManager {
+        type Target = TemplateManager;
+
+        fn deref(&self) -> &Self::Target {
+            &self.manager
+        }
+    }
+
+    fn run_options(
```
rylev, line 521: I don't really find this function self explanatory. Perhaps some small docs can make this clearer should others have any confusion like I did.

rylev (approved):
Looks good!

PR #2660, opened 2024-07-18
fix: change the func impl to reflect trait

me-diru:
quick fix to the trait implementation.

### DIFF
--- a/crates/factor-outbound-pg/src/lib.rs
+++ b/crates/factor-outbound-pg/src/lib.rs
@@ -14,7 +14,7 @@ impl Factor for OutboundPgFactor {
     type AppState = ();
     type InstanceBuilder = InstanceState;
 
-    fn init<T: RuntimeFactors>(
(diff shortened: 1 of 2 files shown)


crates/factor-outbound-pg/src/lib.rs
```diff
@@ -6,6 +6,7 @@ use spin_factors::{
     SelfInstanceBuilder,
 };
 use tokio_postgres::Client;
+use std::marker::Send;
```
lann, line 9: This is part of the automatic rust "prelude".

```suggestion
```

lann (approved):

crates/factor-outbound-pg/src/lib.rs
me-diru: Ahh! Will have to change for the `factor_outbound_redis` one too 😅 . might as well add to this PR!

PR #2661, opened 2024-07-18
factors: Pare down spin-core and spin-app

lann:
These features are moving as part of the migration to factors.

Notably, the spin-core integration test now shows roughly how factors will be integrated with core in spin-trigger.


crates/factors/src/lib.rs
```diff
@@ -17,9 +18,6 @@ pub use crate::{
 };
 
 // Temporary wrappers while refactoring
```
rylev, line 20: I think this is no longer needed, right? The `Result` type is likely to stay I would think

crates/factors/src/prepare.rs
```diff
@@ -1,6 +1,8 @@
 use std::any::Any;
 
-use crate::{AppComponent, Error, Factor, RuntimeFactors};
+use spin_app::AppComponent;
```
rylev, line 3: Would it make sense to `pub use` this at the `factors` crate level like we're doing for `spin_app::App`?

crates/core/src/host_component.rs
rylev, line 1: The `HostComponent` is dead! Long live the `Factor`!

crates/core/tests/integration_test.rs
```diff
@@ -131,83 +76,82 @@ async fn test_set_deadline_violated() {
...
+        "components": [{
+            "id": "test-component",
+            "source": {
+                "content_type": "application/wasm",
+                "content": {},
+            },
+        }]
+    }))?;
+    let app = App::new("test-app", locked);
+    let configured_app = factors.configure_app(app, ())?;
```
rylev, line 142: 😍

rylev (approved):
This gets my vote for most satisfying PR of the year.

PR #2662, opened 2024-07-19
Summarise plugins list

### LINKED ISSUE #2638: Hide old plugin versions in `spin plugins list`
Most people aren't going to care about old versions of plugins. The full list could be hidden behind an e.g. `--all` flag, and the default behavior could show one line per plugin with the existing `[installed]` tag and perhaps an indication that an update is available, e.g. `foo 1.0.0 [installed; update available: 2.0.0]`.

itowlson:
Fixes #2638, but:

* I kept the existing default behaviour and put this on a flag, because it's a breaking change and I vaguely think some things consume the existing.output.  We can switch the default behavious in v3.
* After implementing the behaviour proposed in #2638, I found it awkward because the number after the plugin meant different things (installed or latest) depending on context.  So this PR instead proposes a summary where the number by the plugin is _always_ the latest version, and if you have a different version installed, then that's shown in the `[installed]` bracket.  E.g.

```
$ spin plugins list --summary
check-for-update 0.1.0 [installed]
cloud-gpu 0.1.0
js2wasm 0.6.1 [installed]
kube 0.1.1
py2wasm 0.3.2 [installed version: 0.3.0]
```

Note for reviewers: The underlying struct for plugin listing gets a little weird as a result of this, because it can now represent either a single version of a plugin, or an entire plugin (accompanied with latest/installed versions).  I considered splitting it out into two different structs, but that would have made the code rather more complicated.  Possibly it could be an enum but it didn't feel worth the candle.  Definitely open to feedback on this!

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -437,6 +437,7 @@ fn list_installed_plugins() -> Result<Vec<PluginDescriptor>> {
             installed: true,
             compatibility: PluginCompatibility::for_current(&m),
             manifest: m,
+            installed_version: None,
         })
         .collect();
     Ok(descriptors)
@@ -458,6 +459,7 @@ async fn list_catalogue_plugins() -> Result<Vec<PluginDescriptor>> {
             installed: m.is_installed_in(store),
             compatibility: PluginCompatibility::for_current(&m),
             manifest: m,
+            installed_version: None,
         })
         .collect();
     Ok(descriptors)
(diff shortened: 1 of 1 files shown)


src/commands/plugins.rs
```diff
@@ -469,13 +471,78 @@ async fn list_catalogue_and_installed_plugins() -> Result<Vec<PluginDescriptor>>
...
+            latest.installed_version = Some(installed.version);
+            vec![latest]
+        })
+        .collect()
+}
+
+fn latest(mut plugins: Vec<PluginDescriptor>) -> (Option<PluginDescriptor>, Vec<PluginDescriptor>) {
+    let Ok(versions) = plugins
+        .iter()
+        .map(|pd| semver::Version::parse(&pd.version))
```
rylev, line 513: I'd hate to make this more complicated, but perhaps we want to treat unparseable versions as not possibly being the latest but still try with any versions that do successfully parse.

src/commands/plugins.rs
```diff
@@ -469,13 +471,78 @@ async fn list_catalogue_and_installed_plugins() -> Result<Vec<PluginDescriptor>>
...
+        .collect::<Result<Vec<_>, _>>()
+    else {
+        return (None, plugins);
+    };
+    let Some(latest_ver) = versions.iter().max() else {
```
rylev, line 518: Would it be possible to do `versions.iter().enumerate().max()` to get the index of the max item. This index should match the index of the of the plugin in the plugins collection which means you can avoid turning the version into a string and iterating over the plugins collection again.

rylev (approved):
This is fine since it's relatively self contained, but I do find it to be slightly hard to follow since we have a bunch of state that mutates as we go along. I would personally find a more verbose solution easier to read, but I won't block this PR from landing if you disagree.

PR #2663, opened 2024-07-19
Run conformance tests as part of runtime tests

rylev:
This gets rid of many of the runtime tests that test what the conformance tests test. I imagine eventually all runtime tests will go away in favor of conformance tests unless we start seeing some runtime behavior that we deem Spin CLI specific.


tests/testing-framework/Cargo.toml
```diff
@@ -13,7 +13,7 @@ nix = "0.26.1"
 regex = "1.10.2"
 reqwest = { workspace = true }
 temp-dir = "0.1.11"
-test-environment = { git = "https://github.com/fermyon/conformance-tests", branch = "main" }
+test-environment = { git = "https://github.com/fermyon/conformance-tests", rev = "2bb3eb83467a9e9595fa0fd11d7a49fa0300b113" }
```
lann, line 16: Should conformance-tests move into this repo?

lann (approved):

tests/testing-framework/Cargo.toml
rylev: The point of moving them out of this repo was to facilitate reuse across multiple runtimes, so I don't think so.

lann: We reuse other spin repo code in other runtimes... :shrug:

I'm just always a little wary of having test code separated from the (main) thing its testing.

rylev: As the main person working on this, I can see it has very much helped me keeping these separate. If others find it difficult to contribute because of this separation, we certainly should reconsider my decision, but until there's a concrete reason to move this closer, I'd strongly prefer to continue to keep them apart.

PR #2664, opened 2024-07-19
factors: Add basic CI

lann:


crates/factor-variables/build.rs
```diff
@@ -1,5 +1,5 @@
 fn main() {
-    println!("cargo::rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=build.rs");
```
rylev, line 2: If this line necessary? I believe I commented before that build.rs files automatically build if they're modified, but I'm not sure what the response to that was.

crates/key-value/src/host_component.rs
```diff
@@ -56,7 +56,8 @@ impl HostComponent for KeyValueComponent {
     }
 }
 
-impl DynamicHostComponent for KeyValueComponent {
+// impl DynamicHostComponent for KeyValueComponent {
```
rylev, line 59: Any reason to keep this line around?

rylev (approved):

crates/factor-variables/build.rs
rylev: Ah - I had somehow completely forgotten that that's the case...

lann:
I gave up on getting spin-core tests passing in CI. YOU WIN THIS TIME GITHUB ACTIONS

PR #2665, opened 2024-07-19
Remove toml assumption

rylev:
Currently many factors assume that their runtime configs are serialized as toml. This hurts reusability as this assumption might not be true for many runtimes. 

This PR removes this assumption (for sqlite and kv) instead making the implementations generic on their runtime configurations. While this does add a generic to the code, the underlying factor code already was generic on runtime config - it just made the assumption that that generic config was serialized as toml. Now it knows nothing about the config other than it implements `DeserializeOwned`.

You can notice that the shape of how runtime config is serialized is moved to Spin CLI specific code while the core Factor impl does not care at all about this.

*Note: eventually I would like to remove the assumption that runtime config is serialized at all, but that's a bigger change that requires more controversial decisions. This is still a step towards that direction while still being a good decision IMO even if we keep deserialization assumptions.*

### DIFF
--- a/crates/factor-key-value/src/delegating_resolver.rs
+++ b/crates/factor-key-value/src/delegating_resolver.rs
@@ -1,5 +1,6 @@
 use crate::runtime_config::RuntimeConfigResolver;
 use crate::store::{store_from_toml_fn, MakeKeyValueStore, StoreFromToml};
+use serde::Deserialize;
 use spin_key_value::StoreManager;
 use std::{collections::HashMap, sync::Arc};
 
@@ -9,20 +10,13 @@ pub struct DelegatingRuntimeConfigResolver {
     defaults: HashMap<&'static str, StoreConfig>,
 }
 
-type StoreConfig = (&'static str, toml::value::Table);
-
 impl DelegatingRuntimeConfigResolver {
     pub fn new() -> Self {
         Self::default()
     }
 
-    pub fn add_default_store(
-        &mut self,
-        label: &'static str,
-        store_kind: &'static str,
-        config: toml::value::Table,
-    ) {
-        self.defaults.insert(label, (store_kind, config));
+    pub fn add_default_store(&mut self, label: &'static str, config: StoreConfig) {
+        self.defaults.insert(label, config);
     }
 }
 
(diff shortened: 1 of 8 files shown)


crates/factor-key-value/src/delegating_resolver.rs
rylev, line 1: @kate-goldenring this implementation is somewhat generic but still makes assumptions that are likely specific to Spin CLI (e.g.,toml). The core factor implementation is now very generic and makes very little assumptions about the embedding runtime. It would be nice if we kept generic code and runtime specific code separate for clarity.

Right now in sqlite, we have the `runtime_config` module which contains generic runtime_config related code, but then there is a `spin` submodule which implements the Spin CLI specific runtime config scheme.

In my head, eventually the `factor-sqlite` crate would only contain the highly generic, runtime agnostic code and there would be another crate `factor-sqlite-spin-cli` (or similar) that takes a dependency on `factor-sqlite` and does all of the Spin CLI specific stuff. 

Anyway, wanted to flag this as a thought I'm having so we can works towards a solution in the medium term.

crates/factor-key-value/src/runtime_config.rs
```diff
@@ -1,38 +1,23 @@
...
-    /// Get a store manager for a given store kind and the raw configuration.
-    ///
-    /// `store_kind` is equivalent to the `type` field in the
-    /// `[key_value_store.$storename]` runtime configuration table.
-    fn get_store(
-        &self,
-        store_kind: &str,
-        config: toml::Table,
-    ) -> anyhow::Result<Arc<dyn StoreManager>>;
+pub trait RuntimeConfigResolver<C>: Send + Sync {
```
lann, line 18: Could `C` be an associated type on this trait instead? e.g. `type StoreConfig: DeserializeOwned`?

rylev: Oh yea, that's probably better.

crates/factor-key-value/src/lib.rs
```diff
@@ -19,22 +20,22 @@ use spin_key_value::{
 };
 pub use store::MakeKeyValueStore;
 
-pub struct KeyValueFactor {
-    runtime_config_resolver: Arc<dyn runtime_config::RuntimeConfigResolver>,
+pub struct KeyValueFactor<C> {
+    runtime_config_resolver: Arc<dyn runtime_config::RuntimeConfigResolver<Config = C>>,
```
lann, line 24: If this is going to need a generic it seems like it might as well use that for the `RuntimeConfigResolver` instead. Then you could `type RuntimeConfig = RuntimeConfig<T::Config>`.

lann: Seems to work out: https://github.com/fermyon/spin/commit/03281141b4446b4fa58ccd69a03ebb10811f04c0

rylev: This seems like a worse UX though. Instead of providing the runtime config type (which should be fairly obvious), the user of the API has to supply this resolver type which requires the user to think more deeply about what that type actually is.

lann: 🤔 doesn't the use already have instantiate the resolver in order to instantiate the factor?

rylev: You aren't wrong.......

crates/factor-key-value/src/delegating_resolver.rs
kate-goldenring: Following that model, this file really is the Spin CLI implementation, so we could move it to `runtime_config/spin.rs`. Other runtimes may also choose to use this resolver as well, but it is not a given

crates/factor-key-value/src/lib.rs
rylev: Ok I changed to being generic over the resolver.

crates/factors/tests/smoke.rs
```diff
@@ -42,8 +43,10 @@ async fn smoke_test_works() -> anyhow::Result<()> {
         SpinKeyValueRuntimeConfig::default(Some(PathBuf::from("tests/smoke-app/.spin")));
     key_value_resolver.add_default_store(
         "default",
-        SpinKeyValueStore::RUNTIME_CONFIG_TYPE,
-        toml::value::Table::try_from(default_config)?,
+        StoreConfig {
+            type_: SpinKeyValueStore::RUNTIME_CONFIG_TYPE.to_string(),
+            config: toml::value::Table::try_from(default_config)?,
+        },
```
lann, line 49: Would probably benefit from a `StoreConfig::new` :shrug:

lann (approved):

crates/factors/tests/smoke.rs
rylev: Good point. I'll do this in a follow up.

PR #2666, opened 2024-07-19
factors: key value tests, doc comments, and Azure factor

kate-goldenring:
Adds inline docs, tests, and the final KV Azure factor

### DIFF
--- a/crates/factor-key-value-azure/Cargo.toml
+++ b/crates/factor-key-value-azure/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "spin-factor-key-value-azure"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+anyhow = "1.0"
+serde = { version = "1.0", features = ["rc"] }
+spin-factor-key-value = { path = "../factor-key-value" }
+# TODO: merge with this crate
+spin-key-value-azure = { path = "../key-value-azure" }
+
+[lints]
+workspace = true
--- a/crates/factor-key-value-spin/src/lib.rs
+++ b/crates/factor-key-value-spin/src/lib.rs
@@ -16,13 +16,8 @@ pub struct SpinKeyValueStore {
 
 impl SpinKeyValueStore {
     /// Create a new SpinKeyValueStore with the given base path.
-    /// If the base path is None, the current directory is used.
-    pub fn new(base_path: Option<PathBuf>) -> anyhow::Result<Self> {
-        let base_path = match base_path {
-            Some(path) => path,
-            None => std::env::current_dir().context("failed to get current directory")?,
-        };
-        Ok(Self { base_path })
+    pub fn new(base_path: PathBuf) -> Self {
+        Self { base_path }
     }
 }
 
@@ -38,10 +33,11 @@ impl SpinKeyValueRuntimeConfig {
     const DEFAULT_SPIN_STORE_FILENAME: &'static str = "sqlite_key_value.db";
 
     /// Create a new runtime configuration with the given state directory.
-    /// If the state directory is None, the database is in-memory.
-    /// If the state directory is Some, the database is stored in a file in the state directory.
-    pub fn default(state_dir: Option<PathBuf>) -> Self {
-        let path = state_dir.map(|dir| dir.join(Self::DEFAULT_SPIN_STORE_FILENAME));
+    ///
+    /// If the database directory is None, the database is in-memory.
+    /// If the database directory is Some, the database is stored in a file in the state directory.
+    pub fn default(default_database_dir: Option<PathBuf>) -> Self {
+        let path = default_database_dir.map(|dir| dir.join(Self::DEFAULT_SPIN_STORE_FILENAME));
         Self { path }
     }
 }
--- a/crates/factor-key-value-spin/src/lib.rs
+++ b/crates/factor-key-value-spin/src/lib.rs
@@ -8,11 +8,15 @@ use serde::{Deserialize, Serialize};
 use spin_factor_key_value::MakeKeyValueStore;
 use spin_key_value_sqlite::{DatabaseLocation, KeyValueSqlite};
 
+/// A key-value store that uses SQLite as the backend.
 pub struct SpinKeyValueStore {
+    /// The base path or directory for the SQLite database file.
     base_path: PathBuf,
 }
 
 impl SpinKeyValueStore {
+    /// Create a new SpinKeyValueStore with the given base path.
+    /// If the base path is None, the current directory is used.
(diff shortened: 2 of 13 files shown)


kate-goldenring:
@rylev I can rebase and add in the generic substitution for the Azure factor too once this is in: https://github.com/fermyon/spin/pull/2665/files/bbdf0aa5cc8dc22159aa6ba3b0c250d1552184c8#diff-c336678c804222fc652b2ea241955544fdc5b24f852c533fa02ec722b063c304

crates/factor-key-value-azure/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "spin-factor-key-value-azure"
```
rylev, line 2: I know that naming is a bit up in the air especially since we have a bunch of crates that will eventually go away, but I wonder if we already have an idea if this naming convention is worth keeping.

Imaging a world where the transition to factors is complete, I would personally like to see the following naming conventions:
* For the factor themselves: `factor-$import` (e.g., `factor-llm` or `factor-key-value`). Even in a world where factors are not new, I think naming the crates with `factor` makes sense so that we can quickly see where all of the factors are defined. This is always a reason to start with `factor_` so that all of the factors are bundled together.
  * An issue with this naming however, is that it's not clear in a global namespace like on crates.io what the crated does...
* For Spin CLI specific implementations that wrap the factors: `$import-$impl_specific_description` (e.g., `key-value-azure` or `sqlite-local-database`). Keeping the kind of import first, means all related implementations will be bundled (e.g., everything key-value related will be bundled).
  * The thing I'm unsure about is how to address the need to specify that this is what is used for Spin CLI. The issue is that "spin" is used for much more than the Spin CLI. I can also imagine these being used by some other runtime (e.g., perhaps SpinKube will use some of these). I'm not sure if there's a terse way to label these crates to concisely show where they're used.
  * Of course, the issue listed above for the factor crates is also an issue here: publishing these on crates.io is difficult where they need to be unique and reasonably clear when viewed outside the context of this repo.

crates/factor-key-value-spin/src/lib.rs
```diff
@@ -8,11 +8,15 @@ use serde::{Deserialize, Serialize};
...
+/// A key-value store that uses SQLite as the backend.
 pub struct SpinKeyValueStore {
+    /// The base path or directory for the SQLite database file.
     base_path: PathBuf,
 }
 
 impl SpinKeyValueStore {
+    /// Create a new SpinKeyValueStore with the given base path.
+    /// If the base path is None, the current directory is used.
     pub fn new(base_path: Option<PathBuf>) -> anyhow::Result<Self> {
```
rylev, line 20: I do wonder if we should just take the path (instead of making it optional) and make the caller decide whether they want to use the current directory 🤷

crates/factor-key-value-spin/src/lib.rs
```diff
@@ -22,14 +26,20 @@ impl SpinKeyValueStore {
...
+    /// The path to the SQLite database file.
     path: Option<PathBuf>,
 }
 
 impl SpinKeyValueRuntimeConfig {
+    /// The default filename for the SQLite database.
     const DEFAULT_SPIN_STORE_FILENAME: &'static str = "sqlite_key_value.db";
 
+    /// Create a new runtime configuration with the given state directory.
+    /// If the state directory is None, the database is in-memory.
```
rylev, line 41: Nit: For all of the docs added here, can you add a space between the first sentence and the rest of the doc comment. That tends to format better with `rustdoc`.

crates/factor-key-value-spin/src/lib.rs
```diff
@@ -22,14 +26,20 @@ impl SpinKeyValueStore {
...
+    /// If the state directory is Some, the database is stored in a file in the state directory.
     pub fn default(state_dir: Option<PathBuf>) -> Self {
```
rylev, line 43: Instead of calling this `state_dir`, perhaps we can call it `default_database_dir`? I think this is a bit more description of what this is actually being used for. The fact that in Spin CLI it's the state directory is something that the caller knows but that this implementation doesn't necessarily need to care about. We do [this for sqlite](https://github.com/fermyon/spin/blob/7c6faed98a55a8b5ef6a73f3014c8158a39fe251/crates/factor-sqlite/src/runtime_config/spin.rs#L37).

crates/factor-key-value/src/lib.rs
```diff
@@ -19,11 +20,14 @@ use spin_key_value::{
...
 pub use store::MakeKeyValueStore;
 
+/// A factor that provides key-value storage.
 pub struct KeyValueFactor {
+    /// Resolves runtime configuration into store managers.
     runtime_config_resolver: Arc<dyn runtime_config::RuntimeConfigResolver>,
 }
 
 impl KeyValueFactor {
+    /// Create a new KeyValueFactor.
```
rylev, line 30: I think the important thing to document here is the `runtime_config_resolver` argument. Something like:

```suggestion
    /// Create a new KeyValueFactor.
    ///
    /// The `runtime_config_resolver` is used to resolve runtime configuration into store managers.
```

crates/factor-key-value/tests/test.rs
```diff
@@ -0,0 +1,214 @@
...
+    let mut env = TestEnvironment::default_manifest_extend(toml! {
+        [component.test-component]
+        source = "does-not-exist.wasm"
+        key_value_stores = labels_clone
+    });
+    if let Some(runtime_config) = runtime_config {
+        env.runtime_config.extend(runtime_config);
+    }
+    let state = env.build_instance_state(factors).await?;
+    // String::new("foo").as
```
rylev, line 72:
```suggestion
```

crates/factor-key-value/tests/test.rs
```diff
@@ -0,0 +1,214 @@
...
+        vec![SpinKeyValueStore::new(None)?],
+        vec!["custom"]
+    )
+    .await
+    .is_err());
+    Ok(())
+}
+
+#[tokio::test]
+async fn multiple_custom_key_value_fails() -> anyhow::Result<()> {
```
rylev, line 182: I guess this doesn't actually fail? It seems like we just ignore the second configuration. Is that what we want?

rylev (approved):
Looks like a clear improvement. I have some points for discussion, but I don't think we really need to block on merging this.

crates/factor-key-value-spin/src/lib.rs
lann: "space" -> "blank line" :+1:

crates/factor-key-value/tests/test.rs
kate-goldenring: This test led me to update the impl to check for existence of a store manager for a label before adding the label and manager. I updated this test name to be `multiple_custom_key_value_uses_first_store`. We could also error here

PR #2668, opened 2024-07-22
Break circular dependency between `http` and `testing` crates

itowlson:
I'm not sure quite what was going on here (well, _that's_ always a reassuring start to a PR) but updating to latest `main` made rust-analyser weird and unresponsive and unwilling to load updated dependencies.  There were a lot of errors in the log about a circular dependency.  I couldn't see where the http->testing dependency was used, and when I broke it, nothing _seemed_ to fail, and rust-analyser seemed a lot happier.  So here it is - if I'm missing something, please let me know (if CI doesn't tell me first!).

(The change revved a bunch of unrelated stuff in `Cargo.lock`.  I'm not sure if that's anything we need to worry about but just so you know.)


itowlson:
The first take on this PR failed because something brought in a new version of `rustls` which in turn brought in an AWS crate (!), which refused to compile in CI because it objected to the `cc` compiler.  I am trying pinning `conformance-test` to a SHA instead of `main` (and, for now, have chosen a SHA before the latest commit, to see if that fixes it; it does seem to remove the AWS dependency from the lockfile but we shall see).  Apologies @rylev for possibly fouling up what you are trying to do with the conformance-test integration (but of course I will not commit any change of this nature without your say-so!).

itowlson:
Revved it to the latest commit on conformance-tests and somehow that _didn't_ seem to bring back the AWS dependency so heaven knows where that came from.  Maybe the main pointer was picking up some crazy local cache thing while generating the lockfile.  Anyway pushed with that later rev and WE SHALL SEE.

itowlson:
Okay I am not sure what is going on with that AWS library (`aws-lc-sys`) that refuses to compile on `cc`.  It's _not_ associated with conformance-tests, because I can bring conformance-tests to the latest commit and it's grand.  _But_ as soon as run `cargo update`, the library appears.  It's a dependency of `rustls` 0.23.11, which gets roped in by `reqwest` 0.12.5.  (Everything else we have is happy with rustls 0.22, which doesn't have the problem dependency.)

The cc build issue was raised in a now-closed issue with the rustls maintainers, whose reponse was basically "sucks to be you."  So it's not gonna be fixed, and we can't avoid running `cargo update` forever!  So the fix is really to stop CI using `cc` - it compiled locally for me on WSL so evidently there is some different default between my Rust and CI Rust.

@rylev After determining that AWS was unrelated to conformance-tests, I tried resetting the conformance-tests dependency back to main, but once again rust-analyser started complaining with the errors:

image

So I am leaving this PR with the pinned SHA.  But happy to take your guidance on this.  Sorry for all the faff.

rylev (approved):
Sorry for the hassle here! I'd actually like to get away from using branch in the Cargo.toml and instead pin to specific revs. I've done this in my branch here: https://github.com/fermyon/spin/pull/2663, so I think this is totally fine.

rylev:
I'm going to merge this (the CI failure is a known issue and unrelated to the change here).

PR #2670, opened 2024-07-22
Add environment variables provider

rylev:
Adds a `EnvVariables` provider and moves the static provider to its own submodule. If this looks good, the next step is to implement the vault and azure extensions.

### DIFF
--- a/crates/factor-variables/src/provider.rs
+++ b/crates/factor-variables/src/provider.rs
@@ -1,7 +1,11 @@
-use std::{collections::HashMap, sync::Arc};
+mod env;
+mod statik;
 
-use serde::{de::DeserializeOwned, Deserialize};
-use spin_expressions::{async_trait::async_trait, Key, Provider};
+pub use env::EnvVariables;
+pub use statik::StaticVariables;
+
+use serde::de::DeserializeOwned;
+use spin_expressions::Provider;
 use spin_factors::anyhow;
 
 pub trait MakeVariablesProvider: 'static {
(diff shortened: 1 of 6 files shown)


crates/factor-variables/src/provider.rs
```diff
@@ -1,7 +1,11 @@
...
-use serde::{de::DeserializeOwned, Deserialize};
-use spin_expressions::{async_trait::async_trait, Key, Provider};
+pub use env::EnvVariables;
+pub use statik::StaticVariables;
+
+use serde::de::DeserializeOwned;
+use spin_expressions::Provider;
 use spin_factors::anyhow;
 
 pub trait MakeVariablesProvider: 'static {
```
lann, line 11: Do you think we should switch to the runtime resolver pattern here?

crates/factor-variables/src/provider/env.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Creates a new EnvProvider.
+    ///
+    /// * `prefix` - The string prefix to use to distinguish an environment variable that should be used.
+    /// If not set, the default prefix is used.
+    /// * `env_fetcher` - The function to use to fetch an environment variable.
+    /// * `dotenv_path` - The path to the .env file to load environment variables from. If not set,
+    /// no .env file is loaded.
+    pub fn new(
+        prefix: Option<impl Into<String>>,
+        env_fetcher: impl Fn(&str) -> Result<String, VarError> + Send + Sync + 'static,
```
lann, line 68: Do we have a concrete need for this? Seems a bit marginal vs implementing an entirely new provider...

crates/factor-variables/src/provider.rs
rylev: That's the next PR 😄

crates/factor-variables/src/provider/env.rs
rylev: It makes testing not rely on the environment it's run in. If it complicates too many things in your opinion, we can go back to just assuming `std::env::variable`.

crates/factor-variables/src/provider.rs
rylev: https://github.com/fermyon/spin/compare/factors...untoml-variables

crates/factor-variables/src/provider/env.rs
```diff
@@ -0,0 +1,215 @@
...
+use spin_world::async_trait;
+use tracing::{instrument, Level};
+
+use crate::MakeVariablesProvider;
+
+/// Creator of a environment variables provider.
+pub struct EnvVariables;
+
+impl MakeVariablesProvider for EnvVariables {
+    const RUNTIME_CONFIG_TYPE: &'static str = "static";
```
rylev, line 20: Whoops. Need to change this.

crates/factor-variables/src/provider/env.rs
lann: Maybe a separate `with_env_fetcher`? :shrug:

lann (approved):

lann:
I don't know why all these other checks ran. Feel free to ignore them and merge.

PR #2673, opened 2024-07-22
Implementation of Component Dependencies

fibonacci1729:
This PR implements the [Component Dependencies SIP](https://github.com/fermyon/spin/pull/2543).

See [here](https://github.com/fibonacci1729/spin-animal-facts) for a simple example.

There are a few remaining items to address, namely better error messages and some lingering documentation TODOs. Additionally i'd like to put a few examples together.


itowlson:
@fibonacci1729 I know this is extremely WIP and things are still moving, but is the move of the `spin-serde` crate into `spin-common` intended as permanent?  If so, what's the reasoning for that?

fibonacci1729:
@itowlson Permanent! The reasoning was just to simplify the code organization a bit. The `DependencyName` type introduced into `spin-common` would introduce a dependency on `spin-serde` and that felt annoying.

crates/common/src/dependencies.rs
```diff
@@ -0,0 +1,171 @@
...
+            package,
+            version,
+            interface,
+        })
+    }
+}
+
+impl DependencyName {
+    /// Returns true if the dependency name matches the provided import name string.
+    pub fn matches_import(&self, import_name: &str) -> bool {
```
rylev, line 105: Couldn't we try to parse `import_name` and do `DependencyName` to `DependencyName` equality check then?

crates/common/src/dependencies.rs
```diff
@@ -0,0 +1,171 @@
+//! Types for working with component dependencies.
+
+use crate::serde::KebabId;
+use serde::{Deserialize, Serialize};
+use std::str::FromStr;
+use wasm_pkg_common::package::PackageRef;
+
+/// Dependency name, `foo:bar/baz@0.1.0`, `foo:bar/baz`, `foo:bar@0.1.0`, `foo:bar`
```
rylev, line 8:
```suggestion
/// Name of an import dependency. 
///
/// For example, `foo:bar/baz@0.1.0`, `foo:bar/baz`, `foo:bar@0.1.0`, `foo:bar`
```

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+            for import_name in import_keys.iter() {
+                if dependency_name.matches_import(import_name) {
+                    match imports.swap_remove(import_name) {
+                        Some(_item) => {
+                            matched.push(import_name.clone());
+                            break;
+                        }
+                        None => {
+                            anyhow::bail!(
+                                "import '{import_name}' of component '{component_id}' is already used by another dependency",
```
rylev, line 109: I think we'll want to workshop these errors a bit as I think they come across a bit more technical than we may want.

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        component: &LockedComponent,
+    ) -> anyhow::Result<IndexMap<String, DependencyInfo>> {
+        let mut expanded = IndexMap::new();
+
+        let mut imports = self.graph.types()[world_id].imports.clone();
+
+        // TODO: collect duplicates and return
+
+        for (dependency_name, dependency) in component.dependencies.iter() {
+            let import_keys = imports.keys().cloned().collect::<Vec<_>>();
```
rylev, line 96: Shouldn't we be collecting this outside the iteration of dependencies? Otherwise we reset the list of imports on each iteration.

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        dependency_name: DependencyName,
+        dependency: &LockedComponentDependency,
+    ) -> anyhow::Result<DependencyInfo> {
+        let mut dependency_source = self.loader.load_component(&dependency.source).await?;
+
+        let package_name = dependency_name.package.to_string();
+
+        match dependency.inherit.as_ref() {
+            Some(InheritConfiguration::Some(_configurations)) => {
+                unimplemented!("granular configuration inheritance is not yet supported");
```
rylev, line 222: Nit: shouldn't this be an `anyhow!` error?

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+            loader,
+        }
+    }
+
+    async fn prepare_dependencies(
+        &mut self,
+        world_id: WorldId,
+        component: &LockedComponent,
+    ) -> anyhow::Result<IndexMap<String, DependencyInfo>> {
+        let mut expanded = IndexMap::new();
```
rylev, line 89: I don't get the choice of naming this `expanded`? Can you take a bit about that?

rylev (commented):
Looking like a really great start! 🎉

crates/common/src/dependencies.rs
lann: This isn't a general match between two dependency names, it is specifically checking if a (possibly wildcard) dependency matches a specific import name.

crates/compose/src/lib.rs
lann: This method used to be called `expand_dependencies` until I suggested that @fibonacci1729 change it :sweat_smile:

crates/compose/src/lib.rs
lann: While pairing on this @fibonacci1729 said to me "I'll let people with strong opinions about error messages comment on the PR".

lann: It might be nice to frame these as "conflicts" between pairs of dependencies. We also discussed collecting all conflicts before bailing to give more comprehensive conflict reports.

crates/compose/src/lib.rs
fibonacci1729: This code is technically never executed because `Some(InheritConfiguration::Some(_))` is never constructed. It's a variant i anticipate to use soon after landing this to support granular inheritance. If anything i think this should be an `unreachable`.

crates/compose/src/lib.rs
fibonacci1729: We reset the imports here because we are consuming them from `imports` as we match against dependency names. Since after each match, the set of imports is potentially smaller so we don't want to walk to full set of import keys on each iteration. This is just a slight optimization to keep the complexity best-case O(m*log(n)), worst-case O(n*m).

crates/compose/src/lib.rs
fibonacci1729: Yeah this is definitely an area where i welcome feedback from our wordsmiths.

crates/common/src/dependencies.rs
fibonacci1729: @rylev FWIW that's where i started at but it's more confusing because of the subtle nuance @lann mentioned.

crates/compose/src/lib.rs
rylev: Ah ok makes sense. Can we document that with a comment?

crates/manifest/src/schema/v2.rs
```diff
@@ -145,11 +174,116 @@ pub struct Component {
...
+
+    fn try_from(value: Map<DependencyName, ComponentDependency>) -> Result<Self, Self::Error> {
+        let dependencies = ComponentDependencies { inner: value };
+        dependencies.validate()?;
+        Ok(dependencies)
+    }
+}
+
+impl ComponentDependencies {
+    fn validate(&self) -> anyhow::Result<()> {
```
itowlson, line 204: This feels like a lot of logic to have in a schema - this puts us into "we can't even deserialise the thing to reason about it" territory.  For some reason I thought we had a separate "validate" step but maybe not.  And maybe this is no different from the way we fail the read for kebab-id violations.  Anyway, just sharing it as a reservation...!

crates/manifest/src/schema/v2.rs
```diff
@@ -145,11 +174,116 @@ pub struct Component {
...
+        self.ensure_package_names_no_export()?;
+        self.ensure_disjoint()?;
+        Ok(())
+    }
+
+    /// This method ensures that dependency names in the package form (e.g. "foo:bar" or "foo:bar@0.1.0") do not have an export specified.
```
itowlson, line 210:
```suggestion
    /// Whole-package imports (e.g. "foo:bar" or "foo:bar@0.1.0") cannot map to specific exported interfaces.
```

(I struggled with the current wording because it seemed like "do not have an export specified" was referring to the dependency name and I was going "well duh."  Not sure my suggestion is quite right but hopefully inspires something better!)

crates/manifest/src/schema/v2.rs
```diff
@@ -95,6 +98,32 @@ impl TryFrom<toml::Value> for ComponentSpec {
     }
 }
 
+/// Component dependency
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(untagged, deny_unknown_fields)]
```
itowlson, line 103: Untagged enums are super brittle with strict formats like VersionReq, PackageRef or KebabId.  Like if someone typoes `"a:b" = { package = "c:d", version = "canary" }` they'll get some error about "data did not match any variant of untagged enum ComponentDependency."  (Cf. https://github.com/fermyon/spin/pull/2634.)  I'd strongly encourage being more forgiving at deserialisation time and post-processing into a stricter form when you know which variant you're in!

crates/trigger/src/loader.rs
```diff
@@ -204,13 +204,43 @@ fn check_uncomponentizable_module_deprecation(module_path: &Path) {
     }
 }
 
+struct TriggerComponentLoader;
```
itowlson, line 207: Are there any other component loaders?  Could we call this e.g. ComponentLoader or DefaultComponentLoader?  It doesn't seem to have anything particular to do with triggers.

itowlson (commented):
A few random and not very organised comments - I hope it's not too early for them, feel free to ignore if it is!  I'm very much enjoying playing with the new feature.  Pretty slick!

crates/trigger/src/loader.rs
```diff
@@ -87,14 +87,16 @@ impl Loader for TriggerLoader {
     async fn load_component(
         &self,
         engine: &spin_core::wasmtime::Engine,
-        source: &LockedComponentSource,
+        component: &AppComponent,
```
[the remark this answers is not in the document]
fibonacci1729: Yeah! That's a great suggestion. I'll take a look!

crates/trigger/src/loader.rs
itowlson: Further to the ComponentLoader comment: while labouring in the target environment mines, I realised I was going to need to call the loader/composer as part of validation.  This is needed because we validate that the _fully resolved composition,_ rather than the raw component source, targets the right world.  But I would like the validator not to depend on the `trigger` crate.

It may be that the solution for that is for the validator to implement its own ComponentLoader, and that's fine.  It's not like there's a huge amount of logic in there after all!  And we can certainly address that as a refactoring in he validation work.  Mostly just thinking out loud!

crates/compose/src/lib.rs
radu-matei: +1 on the user facing error here being more descriptive.

crates/common/src/dependencies.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        Ok(Self {
+            package,
+            version,
+            interface,
+        })
+    }
+}
+
+impl DependencyName {
```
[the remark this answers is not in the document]
fibonacci1729: Reverted back to `spin-serde` and made `matches_import` a freestanding function of `spin-compose`

crates/manifest/src/schema/v2.rs
```diff
@@ -145,11 +174,116 @@ pub struct Component {
     /// Settings for custom tools or plugins. Spin ignores this field.
     #[serde(default, skip_serializing_if = "Map::is_empty")]
     pub tool: Map<String, toml::Table>,
+    /// If true, allow dependencies to inherit configuration.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub dependencies_inherit_configuration: Option<bool>,
```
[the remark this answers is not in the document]
fibonacci1729: Updated, PTAL.

crates/locked-app/src/locked.rs
```diff
@@ -221,6 +223,29 @@ pub struct LockedComponent {
...
+    /// Which configurations to inherit from parent
+    pub inherit: Option<InheritConfiguration>,
+}
+
+/// InheritConfiguration specifies which configurations to inherit from parent.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum InheritConfiguration {
+    /// Dependencies will inherit all configurations from parent.
+    All,
+    /// Dependencies will inherit only the specified configurations from parent (must contain atleast 1 option).
```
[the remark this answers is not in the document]
fibonacci1729: Updated to use `InheritConfiguration::Some([])` to indicate `deny-all`, PTAL.

crates/trigger/src/loader.rs
fibonacci1729: Renamed to `ComponentLoader`.

crates/manifest/src/schema/v2.rs
fibonacci1729: Updated the wording to make it more explicit that the right-hand-side of a `DependencyName` cannot have an export specified. LMKWYT

crates/trigger/src/loader.rs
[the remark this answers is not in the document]
fibonacci1729: Updated to use `LockedComponent` here and uncommented the test, thanks!

crates/compose/src/lib.rs
fibonacci1729: Updated, PTAL

crates/manifest/src/schema/v2.rs
fibonacci1729: Do you have reservation about this code living in `v2.rs` or just about from where `validate` is called? I agree that it makes sense to call `validate` from the loader for the reasons you stated.

crates/manifest/src/schema/v2.rs
fibonacci1729: I agree. Just to confirm, should `Registry` and `PackageRef` be specified as `String`s here (and in that PR)?

fibonacci1729: That PR only seems to cover `version`...

fibonacci1729: Updated to relax the constraints in the untagged enum `ComponentDependency`. PTAL (+@ the error messages when parsing in loader), thanks!

crates/manifest/src/schema/v2.rs
fibonacci1729: Updated to defer validating dependency sections until locking (just after normalization). PTAL, thanks!

crates/manifest/src/schema/v2.rs
itowlson: Yeah the linked PR was just an example of where it bit us most recently.  Anything that can fail to parse is iffy in an untagged enum.

(And since _anything,_ can fail to parse, even ints and strings... but that is another fret for another day.)

crates/trigger/src/loader.rs
```diff
@@ -112,14 +113,12 @@ impl Loader for TriggerLoader {
...
-                        "failed to read component source from disk at path {}",
-                        quoted_path(&path)
-                    )
-                })?;
-                let component = spin_componentize::componentize_if_necessary(&bytes)?;
-                spin_core::Component::new(engine, component.as_ref())
+                let component_loader = ComponentLoader;
+                let composed = spin_compose::compose(&component_loader, component)
+                    .await
+                    .with_context(|| format!("composing component {:?}", component.id))?;
```
rylev, line 119:
```suggestion
                    .with_context(|| format!("failed to compose component {:?}", component.id))?;
```

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+
+            // Ensure that export_ty is a subtype of import_ty
+            checker.is_subtype(
+                *export_ty,
+                self.graph.types(),
+                *import_ty,
+                self.graph.types(),
+            ).with_context(|| {
+                format!(
+                    "dependency '{dependency_name}' export '{export_name}' of type '{export_ty:?}' is not a subtype of import '{import_name}' of type '{import_ty:?}'",
```
rylev, line 200: `export_ty` and `import_ty` are meaningless in debug form (e.g., they look like `Instance(InterfaceId(Id { idx: 31 }))`). 

We might be better off completely removing the type information as it might be more confusing than helpful. For example, do we expect users to know what an instance is? 

If we keep the type information we should at the very least do:

```rust
                let export_ty = export_ty.desc(&self.graph.types());
                let import_ty = import_ty.desc(&self.graph.types());
```

But I would do:

```suggestion
                    "dependency '{dependency_name}' exports '{export_name}' which is not compatible with import '{import_name}'",
```

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+            import_name.ends_with(&version.to_string())
+        } else {
+            // Otherwise, the dependency name has no version and matches any version.
+            true
+        }
+    } else {
+        // Interface wildcard (i.e. versionless).
+        if dependency_name.version.is_none() {
+            // We have an interface wildcard (i.e. versionless) where package/interface must match.
+            import_name.starts_with(&dependency_name.to_string())
```
rylev, line 364: I think this isn't quite right. This makes dependency matching work sort of like a wildcard I believe. For example, in the example app, the following works without problem:

```
"component:animal-facts/animal" = { path = "target/wasm32-wasi/release/animal_facts_dep.wasm" }
```

Notice the dependency is `"component:animal-facts/animal"` and not the expected `"component:animal-facts/animal-facts"`

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+                        previous_dependency_name = previous_info.manifest_name,
+                    );
+                }
+            }
+        }
+
+        Ok(prepared)
+    }
+
+    async fn build_instantiation_arguments(
```
rylev, line 147: These functions could very much use a bit of explanation of what they are used for.

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,406 @@
...
+
+        for (dependency_name, dependency) in component.dependencies.iter() {
+            // As we match dependency names against import names, we remove the
+            // matched import name from the imports map to prevent duplicate
+            // matches. Collecting the import keys here is an optimization to
+            // avoid walking the full set of import keys on each iteration.
+            let import_keys = imports.keys().cloned().collect::<Vec<_>>();
+
+            let mut matched = Vec::new();
+
```
rylev, line 103: I think we also want to keep track of the imports of the dependencies. The imports themselves should not have any dependencies. Right now, we just float those dependencies through and you end up with a linker error.

crates/compose/src/lib.rs
fibonacci1729: Great catch!

crates/compose/src/lib.rs
fibonacci1729: I don't think it's quite that easy. There are some instances (in-fact likely the common case) where we want to float dependencies of dependencies through. This is where a target worlds check would be applied subsequently to composition to ensure that composition conforms to what Spin knows how to handle.

crates/compose/src/lib.rs
fibonacci1729: Updated, PTAL.

crates/compose/src/lib.rs
fibonacci1729: Updated and simplified, PTAL.

crates/compose/src/lib.rs
fibonacci1729: The latest commit switches to using `thiserror` to return a typed error. I'm sure the descriptions could be improved so let me know if you have ideas.

crates/compose/src/lib.rs
rylev: Sounds good - I think we'll want to eventually work on improving how linker errors get bubbled to users anyway, and if we improve that, this concern goes away.

crates/compose/deny-all/src/wasi.rs
```diff
@@ -0,0 +1,341 @@
...
+
+impl http::outgoing_handler::Guest for crate::Component {
+    fn handle(
+        _request: http::outgoing_handler::OutgoingRequest,
+        _options: Option<http::outgoing_handler::RequestOptions>,
+    ) -> Result<
+        http::outgoing_handler::FutureIncomingResponse,
+        http::outgoing_handler::ErrorCode,
+    > {
+        Err(http::outgoing_handler::ErrorCode::InternalError(Some("Access denied".to_string())))
```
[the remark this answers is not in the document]
fibonacci1729: Good idea.

crates/manifest/src/schema/v2.rs
```diff
@@ -95,6 +110,32 @@ impl TryFrom<toml::Value> for ComponentSpec {
...
+    /// `... = ">= 0.1.0"`
+    Version(String),
+    /// `... = { version = "0.1.0", registry = "registry.io", ...}`
+    Package {
+        /// Package version requirement
+        version: String,
+        /// Optional registry spec
+        registry: Option<String>,
+        /// Optional package name `foo:bar`
+        package: Option<String>,
```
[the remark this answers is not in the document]
fibonacci1729: The `package` is inferred from the the dependency name. This is similar to the `package` field in `Cargo.toml`

crates/compose/deny-all/src/fermyon.rs
```diff
@@ -0,0 +1,217 @@
...
+    fn open(_address: String) -> Result<mysql::Connection, mysql::Error> {
+        Err(mysql::Error::Other("Access Denied".to_string()))
+    }
+
+    fn query(
+        &self,
+        _statement: String,
+        _params: Vec<mysql::ParameterValue>,
+    ) -> Result<mysql::RowSet, mysql::Error> {
+        unimplemented!()
```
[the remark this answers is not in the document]
fibonacci1729: Updated to `unreachable!` but didn't add a message.

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,512 @@
...
+            package: package.to_string(),
+            interface: interface.to_string(),
+            version,
+        })
+    }
+}
+
+/// Returns true if the dependency name matches the provided import name string.
+fn matches_import(dependency_name: &DependencyName, import_name: &str) -> anyhow::Result<bool> {
+    let import_name = import_name.parse::<ImportName>()?;
```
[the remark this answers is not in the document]
fibonacci1729: This actually made me realize that i completely forgot to handle the case where import names are plain names. To handle this I had to introduce a new variant of `DependencyName::Plain(KebabId)` that _only_ matches identical import names.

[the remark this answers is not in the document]
fibonacci1729: @rylev PTAL when you get a chance. Updated the code to handle dependents and dependencies with root level plain name import & exports.

radu-matei:
Trying to run a component that doesn't have any dependencies using this branch results in an error.

For the following component configuration:

```toml
[component.hello]
source = "bin/Release/net9.0/wasi-wasm/publish/App.wasm"
[component.hello.build]
command = "dotnet publish App.csproj"
```

This is the error:

```
$ spin up
Error: Failed to instantiate component 'hello'

Caused by:
   0: failed to compose component "hello"
   1: an error occurred when preparing dependencies
   2: failed to read component source from disk at path "/Users/radu/projects/src/github.com/fermyon/spin-dotnet-sdk/samples/http-hello/bin/Release/net9.0/wasi-wasm/publish/App.wasm"
```

fibonacci1729:
I've been thinking about this over the last week and i'm not really sold that `.../*` is really all that more intuitive over what we have. Personally having multiple syntaxes to imply variabilitiy (e.g. `.../*` and no version to imply any-version) is a bit confusing. As a dev, if i saw `foo:bar/*` to mean "any interface" i would intuitively think that i would need to say `foo:bar/*@*` to imply any interface and version. @rylev i know this isn't a super strong argument against but i think i'd like to land this with the syntax as-is unless you really feel strongly.

crates/compose/src/lib.rs
```diff
@@ -0,0 +1,546 @@
...
+        Package::from_bytes(dependency_name, None, dependency_source, graph.types_mut())?;
+
+    let dependency_id = graph.register_package(dependency_package)?;
+
+    let deny_adapter_package = Package::from_bytes(
+        "spin-virt-deny-all-adapter",
+        None,
+        SPIN_VIRT_DENY_ALL_ADAPTER_BYTES,
+        graph.types_mut(),
+    )?;
```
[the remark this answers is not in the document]
lann: Is this actually expensive?

crates/manifest/src/schema/v2.rs
```diff
@@ -145,6 +187,145 @@ pub struct Component {
...
+                        export.is_none(),
+                        "using an export to satisfy the package dependency `{dependency_name}` is not currently permitted",
+                    );
+                }
+            }
+        }
+        Ok(())
+    }
+
+    // TODO: document this invariant
```
itowlson, line 266: DOCUMENT IT BRIAN

crates/manifest/src/schema/v2.rs
```diff
@@ -145,6 +187,145 @@ pub struct Component {
...
+            }
+        }
+
+        if let (Some(this), Some(other)) = (this.interface.as_ref(), other.interface.as_ref()) {
+            if this != other {
+                return Ok(());
+            }
+        }
+
+        anyhow::bail!("'{this}' dependency conflicts with '{other}'")
```
itowlson, line 303: double plus mega super duper nit: and now single quotes

crates/manifest/src/schema/v2.rs
```diff
@@ -145,6 +187,145 @@ pub struct Component {
...
+    }
+
+    fn normalize_version(mut version: semver::Version) -> semver::Version {
```
itowlson, line 306: Can you clarify the operation of this function?  I am not sure what "normalising" means in this context but it seems like e.g. 2.2.1 would go to 2.0.0, 0.2.1 would go to 0.2.0, etc.  Am I reading that right?  Is it like "keep the highest nonzero part" kind of thing?  What is the intent - something to do with what is permitted for disjointness purposes?  Could we hint at that in the function name or in a comment?

crates/manifest/src/schema/v2.rs
```diff
@@ -145,6 +187,145 @@ pub struct Component {
...
+    ) -> anyhow::Result<()> {
+        assert_eq!(this.package, other.package);
+
+        if let (Some(this), Some(other)) = (this.version.clone(), other.version.clone()) {
+            if Self::normalize_version(this) != Self::normalize_version(other) {
+                return Ok(());
+            }
+        }
+
+        if let (Some(this), Some(other)) = (this.interface.as_ref(), other.interface.as_ref()) {
```
itowlson, line 297: I know it is legal but I'm finding the repeated re-use of `this` and `other` to refer to different kinds of things quite distracting.  Consider e.g. `this_ver`, `this_itf`.

crates/manifest/src/schema/v2.rs
```diff
@@ -404,4 +587,147 @@ mod tests {
...
+            ("foo:bar/baz@0.1.0", "foo:bar/bub@0.1.0"),
+            ("foo:bar@0.1.0", "foo:bar/bub@0.2.0"),
+            ("foo:bar@1.0.0", "foo:bar@2.0.0"),
+            ("foo:bar@0.1.0", "foo:bar@1.0.0"),
+            ("foo:bar/baz", "foo:bar/bub"),
+            ("foo:bar/baz@0.1.0-alpha", "foo:bar/baz@0.1.0-beta"),
+        ] {
+            let a: DependencyPackageName = a.parse().unwrap();
+            let b: DependencyPackageName = b.parse().unwrap();
+            ComponentDependencies::check_disjoint(&a, &b).unwrap(); // TODO: should method on depednecny name
```
itowlson, line 605: Maybe do an `expect` with the two strings so it's easier to see which one has failed?  Or have the test return `Result` and just `?` that sucker

crates/manifest/src/schema/v2.rs
```diff
@@ -404,4 +587,147 @@ mod tests {
...
+                "{a} should conflict with {b}",
+            );
+        }
+    }
+
+    #[test]
+    fn test_validate_dependencies() {
+        // Specifying a dependency name as a plain-name without a package is an error
+        assert!(ComponentDependencies::deserialize(toml! {
+            "foo-bar" = "0.1.0"
```
itowlson, line 628: This might be a case where `foo` and `bar` are unhelpful identifiers.  I've seen so many `foo:bar` today that it took me a _lot_ of staring at this to realise this was hyphenated rather than coloned, and hence why it was bad.

E.g. `my-name`, `my-package:my-name`, `ns:name/itf`, etc.

crates/serde/src/dependencies.rs
```diff
@@ -0,0 +1,179 @@
...
+use serde::{Deserialize, Serialize};
+use std::str::FromStr;
+use wasm_pkg_common::package::PackageRef;
+
+/// Name of an import package dependency.
+///
+/// For example: `foo:bar/baz@0.1.0`, `foo:bar/baz`, `foo:bar@0.1.0`, `foo:bar`.
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, Hash)]
+#[serde(into = "String", try_from = "String")]
+pub struct DependencyPackageName {
```
itowlson, line 14: At the risk of opening a huge can of worms, is `Package` correct here?  It seems like part of it is the package, but it might also contain an interface.

crates/serde/src/dependencies.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+// TODO: replace with derive once wasm-pkg-common is released
+impl Ord for DependencyName {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self, other) {
+            (DependencyName::Plain(a), DependencyName::Plain(b)) => a.cmp(b),
+            (DependencyName::Package(a), DependencyName::Package(b)) => {
+                let big_ole_tup = (
```
itowlson, line 105: BIG OLE TUP

No action required.  I just like saying it.

crates/compose/deny-all/src/fermyon.rs
```diff
@@ -0,0 +1,218 @@
...
+}
+
+impl mqtt::GuestConnection for crate::Component {
+    fn open(
+        _address: String,
+        _username: String,
+        _password: String,
+        _keep_alive_interval_in_secs: u64,
+    ) -> Result<mqtt::Connection, mqtt::Error> {
+        Err(mqtt::Error::Other(format_deny_error("fermyon:spin/mqtt")))
```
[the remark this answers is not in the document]
fibonacci1729: I'm not sure i agree. To me there is a useful distinction between the capability being disabled via virtualization and a misconfiguration of the otherwise enabled capability.

crates/serde/src/dependencies.rs
fibonacci1729: Short of `DependencyPackageOrInterfaceName` i don't have any better suggestions. Do you have any?

crates/loader/src/local.rs
```diff
@@ -70,6 +72,8 @@ impl LocalLoader {
     pub(crate) async fn load_manifest(&self, mut manifest: AppManifest) -> Result<LockedApp> {
         spin_manifest::normalize::normalize_manifest(&mut manifest);
 
+        manifest.validate_depdendencies()?;
```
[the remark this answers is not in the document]
fibonacci1729: DEPDENDS!

crates/compose/deny-all/src/fermyon.rs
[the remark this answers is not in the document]
itowlson: Yes, but that distinction is useful to the operator, emitted via logs.  I'm not sure the component should be told.  But probably I am overly paranoid about what sort of things could lead to "Towlson in disclosure vulnerability shocker" on the HN front page...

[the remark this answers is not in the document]
fibonacci1729: I think it _could_ be useful to the developer to provide fallback behavior and such. If i can't distinguish between "disabled" and "misconfigured" in code then there's no way for me to, e.g. gracefully fallback to using the local-randomly-generated-animal-fact-generator if outbound-http is disabled.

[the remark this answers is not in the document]
itowlson: I think we will have to agree to disagree.  It's not a blocker for sure!

crates/serde/src/dependencies.rs
itowlson: Would `DependencyName` work?  Not a blocker either way.

fibonacci1729: Unfortunately `DependencyName` is used below to mean either a plain name or this package form.

itowlson (approved):
Thanks @fibonacci1729 for making this happen!

itowlson:
I did have one further (extremely non-blocking) question on the deny-all adapter.  The adapter Wasm file is checked in.  But I am nut sure how it gets built.  Is this a "build locally and check in the binary" thing?  If so it would be good to investigate (as a follow-up) how to move it to a controlled, trusted build environment, but I am not sure how that would play nicely with needing to have it checked in and being included as a blob.

crates/compose/src/lib.rs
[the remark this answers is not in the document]
rylev: In the grand scheme of things, not really. It's certainly not trivial, but it doesn't do any I/O so it's probably fine.

rylev:
> I've been thinking about this over the last week and i'm not really sold that `.../*` is really all that more intuitive over what we have.

@fibonacci1729 I'm not tied to my proposed syntax, but I do maintain that the current syntax is not ideal. The fact that `"org:package/interface" means a specific interface while`"org:package"` means any interface seems inconsistent, and if you had asked me to guess what the syntax would be, I would have guessed that. What's more, the current syntax doesn't seem to be very extensible. For example, how would we imagine a syntax for specifying a list of interfaces?

I'm not tied to any particular solution, but I feel somewhat strongly that the existing syntax is the wrong way go. 

In the interest of making some sort of alternative for discussion, how about:

```
# A specific interface
"org:package" = { interface = "my-interface", path = "path/to/component.wasm" }
# All interfaces is the default
"org:package" = {  path = "path/to/component.wasm" }
```

In the above syntax, the key is *always* the package.

lann:
> how would we imagine a syntax for specifying a list of interfaces?

```toml
"org:package" = { exports = ["my-interface"], ... }
# equivalent to
"org:package/my-interface" = { ... }
```

fibonacci1729:
@rylev could say more about why `org:package/interface` meaning a specific interface while `org:package` meaning a set of interfaces seems inconsistent? If you're concerned about enabling a more granular control over which interfaces can be matched what @lann suggested is what we've been planning for in a follow-up interation. Currently the code enforces a constraint to prevent this where it checks that if a dependency name is in the package pattern form the dependency object (right-hand-side) cannot specify an export. In the future it would seem to make sense to allow saying `org:package = { ..., exports = ["a", "b"] }` to mean that for all compatible imports part of the `org:package` package use `a` and `b` to satisfy them. 

I explicitly punted this syntax down the road to be a little conservative up front.

itowlson:
Regarding the UI, my mental model was that the left hand side was "the thing that is being satisfied" and the right hand side was "the thing that satisfies it".  I appreciate that Ryan's `"org:package" = { interface = "..." }` idea was intended to spark new thinking, but for me the specific detail of putting "the interface that is being satisfied" on the right hand side feels a bit awkward.  But maybe my mental model is out of sync with the realities of components?  It wouldn't be the first time!

(And if Lann's comment is how we plan to do multiple interfaces then Ryan's idea seems consistent with it, although perhaps I am misunderstanding.)

PR #2675, opened 2024-07-23
Retry running conformance tests as part of runtime tests

rylev:
Brings back https://github.com/fermyon/spin/pull/2663

This time we skip tests that require Docker when docker is not installed.

The first commit is equivalent to https://github.com/fermyon/spin/pull/2663 so the only new code is in the subsequent commit.


lann:
I think we should maybe require explicitly opting out of docker tests (via e.g. ugly env var) rather than detecting whether docker is installed.

rylev:
I don't have a strong opinion here. What are your reasons for this? Allowing people to filter out docker based tests locally even when they have docker installed?

lann:
> I don't have a strong opinion here. What are your reasons for this? Allowing people to filter out docker based tests locally even when they have docker installed?

Just "don't make it too easy to skip tests accidentally"

lann (approved):

PR #2678, opened 2024-07-23
Enable environmental authentication for Azure Key Vault var provider

devigned:
This PR is related to #2566 in that it provides similar authentication behaviors for Azure Key Vault var provider. This change should be additive and not cause any breakage to existing applications.

It would probably be wise to follow on with some integration tests to guard against future breakage. I'm not clear on the project stance on spinning up cloud resources for testing, but if help is wanted, I'd be delighted to lend a hand.

To create the infra and test this, simply follow https://developer.fermyon.com/spin/v2/dynamic-configuration#azure-key-vault-application-variable-provider-example. You can skip over creating the service principal and instead use your current identity provided through Azure CLI to authenticate.

Runtime config should look like the following with `${your_vault_name]` replaced:
```toml
[[config_provider]]
type = "azure_key_vault"
vault_url = "https://${your_vault_name}.vault.azure.net/"
```

On a side note, fetching secrets from Key Vault upon each request is pretty slow. Might want to consider some form of memoization / caching for a period of time.

### DIFF
--- a/crates/trigger/src/runtime_config/variables_provider.rs
+++ b/crates/trigger/src/runtime_config/variables_provider.rs
@@ -1,6 +1,10 @@
 use std::path::PathBuf;
 
+use anyhow::{anyhow, Result};
 use serde::Deserialize;
+use spin_variables::provider::azure_key_vault::{
+    AzureKeyVaultAuthOptions, AzureKeyVaultRuntimeConfigOptions,
+};
 use spin_variables::provider::{
     azure_key_vault::{AzureAuthorityHost, AzureKeyVaultProvider},
     env::EnvProvider,
@@ -27,7 +31,7 @@ impl VariablesProviderOpts {
         ))
     }
 
-    pub fn build_provider(&self) -> VariablesProvider {
+    pub fn build_provider(&self) -> Result<VariablesProvider> {
         match self {
             Self::Env(opts) => opts.build_provider(),
             Self::Vault(opts) => opts.build_provider(),
@@ -60,11 +64,11 @@ impl EnvVariablesProviderOpts {
         }
     }
 
-    pub fn build_provider(&self) -> VariablesProvider {
-        Box::new(EnvProvider::new(
+    pub fn build_provider(&self) -> Result<VariablesProvider> {
+        Ok(Box::new(EnvProvider::new(
             self.prefix.clone(),
             self.dotenv_path.clone(),
-        ))
+        )))
     }
 }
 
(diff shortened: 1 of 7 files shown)


crates/variables/src/provider/azure_key_vault.rs
```diff
@@ -2,54 +2,113 @@ use std::sync::Arc;
...
         Self {
-            client_id: client_id.into(),
-            client_secret: client_secret.into(),
-            tenant_id: tenant_id.into(),
-            vault_url: vault_url.into(),
-            authority_host: authority_host.into(),
+            client_id,
+            client_secret,
+            tenant_id,
+            authority_host: authority_host.map_or(AzureAuthorityHost::default(), |ah| ah),
```
itowlson, line 32: If I'm reading this right, it would be more idiomatic to use `.unwrap_or_else(|| AzureAuthorityHost::default()` rather than passing things through the identity function.  But I might be missing something!

crates/trigger/src/runtime_config/variables_provider.rs
```diff
@@ -79,35 +83,52 @@ pub struct VaultVariablesProviderOpts {
...
-            &self.client_secret,
-            &self.tenant_id,
+    pub fn build_provider(&self) -> Result<VariablesProvider> {
+        let auth_config_runtime_vars = [&self.client_id, &self.tenant_id, &self.client_secret];
+        let any_some = auth_config_runtime_vars.iter().any(|&var| var.is_some());
+        let any_none = auth_config_runtime_vars.iter().any(|&var| var.is_none());
+
+        if any_none && any_some {
+            // some of the service principal auth options were specified, but not enough to authenticate.
+            return Err(anyhow!("Azure Key Vault provider requires each client_id, client_secret, and tenant_id to be provided. If none are provided, the provider will attempt to authenticate using ambient authentication (e.g. env vars, Azure CLI, Managed Identity, Workload Identity"));
```
itowlson, line 114: I had a go at wordsmithing this to focus more on "what is the problem and what is the remedy."  I'm not super happy with the result but sharing it in the hope that it will inspire you to come up with the perfect wording...

```suggestion
            return Err(anyhow!("The current runtime config specifies some but not all of the Azure KeyVault 'client_id', 'client_secret', and 'tenant_id' values. Provide the missing values to authenticate to Azure KeyVault with the given service principal, or remove all these values to authenticate using ambient authentication (e.g. env vars, Azure CLI, Managed Identity, Workload Identity)."));
```

itowlson (approved):
I had a couple of non-blocking suggestions / questions, but otherwise LGTM.  Thanks!

(And you're right about the caching.  That's a cross-cutting concern though and definitely on the radar!)

crates/trigger/src/runtime_config/variables_provider.rs
devigned: Yours is an improvement on what I had. I think it is clear, but might benefit from a link to auth docs when they are authored. I don't think any brief error message is going to provide the clarity needed to understand all of the options implied.

rylev:
@itowlson is this waiting on anything else to merge?

squillace:
someone's vacation, I imagine. :-P

itowlson:
@squillace UNCANNY

PR #2679, opened 2024-07-23
Update Wasmtime to 23.0.1

alexcrichton:
This commit updates Wasmtime from 22.0.0 to 23.0.1. This involved a number of changes around how the auto-generated bindings around worlds work and the handling of exports. The string lookups currently performed at http-handler-time as part of the http trigger are now performed at load time instead which should shave some cycles off execution of handler for each request.

This currently has a downside though of the spin-timer example has gotten more complicated with a custom impl of the `TriggerInstancePre` trait. Removing that will require upstream changes in Wasmtime by adding some more methods to generated structures.

### DIFF
--- a/.github/actions/spin-ci-dependencies/action.yml
+++ b/.github/actions/spin-ci-dependencies/action.yml
@@ -8,7 +8,7 @@ inputs:
     type: bool
   rust-version:
     description: 'Rust version to setup'
-    default: '1.76'
+    default: '1.77'
     required: false
     type: string
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  RUST_VERSION: 1.76
+  RUST_VERSION: 1.77
 
 jobs:
   dependency-review:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ on:
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 env:
-  RUST_VERSION: 1.76
+  RUST_VERSION: 1.77
 
 jobs:
   build-and-sign:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1594,21 +1594,32 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.109.0"
+version = "0.110.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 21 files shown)


itowlson:
@alexcrichton Could you say more about what is different in the timer's TriggerInstancePre compared to the built-ins?  Is this something all custom triggers will need to take on?  Can it be pushed up into the trigger crate so that custom triggers don't have to care?

lann:
...also I'm in the process of killing trigger-specific InstancePres on the factors branch...

itowlson:
HUZZAH

rylev (approved):

alexcrichton:
@itowlson previously the spin-timer example would use:

```rust
let instance = SpinTimer::new(&mut store, &instance)?;
```

to go from the `Instance` of the default settings to the `SpinTimer` wrapper. This `SpinTimer` wrapper was auto-generated by `wasmtime::component::bindgen!`. In Wasmtime 22.0.0 [this method in the generated bindings](https://docs.rs/wasmtime/22.0.0/wasmtime/component/bindgen_examples/_2_world_exports/struct.HelloWorld.html#method.new) (that's an example, but the `SpinTimer` struct would be similar) has now been changed to [these bindings instead](https://docs.rs/wasmtime/latest/wasmtime/component/bindgen_examples/_2_world_exports/struct.HelloWorld.html#method.new) in Wasmtime 23.0.0. The previous constructor has been removed and the replacement isn't sufficient for the previous use case here. 

That's the underlying reason motivating this change, and the best "fix" that I could come up with is a custom trait implementation of `TriggerInstancePre`. I realize though that that isn't great, especially from an example perspective.

>  Is this something all custom triggers will need to take on? 

Most likely yes because the custom worlds don't have the same constructor as before. 

> Can it be pushed up into the trigger crate so that custom triggers don't have to care?

Unfortunately no because this needs to be implemented for the types generated by the `bindgen!` macro which the trigger crate can't bundle implementations for.

---

I'm happy to defer the Wasmtime 23 upgrade if this isn't something Spin wants to take on though. I can work on adding the same constructor as before, or similar, to `bindgen!` to get released with Wasmtime 24. If the factors branch removes most of this though the point may also become moot as well.

rylev:
@alexcrichton it appears you need to sign your commits for this to be mergable.

alexcrichton:
Oh good point, but I'm also happy to continue to make changes as needed (and/or redirect this to the factors branch if that's the best way to go)

rylev:
@lann what are your thoughts on landing this on `main` vs the factors branch?

lann:
Makes sense to me, though some of the updated code has changed significantly in the branch.

alexcrichton:
I'll take the route of https://github.com/bytecodealliance/wasmtime/pull/9175 and will reopen this for Wasmtime 25 rebased with factors when that's out.

PR #2680, opened 2024-07-25
Changes need to make Rust and Clippy happy on Rust 1.80

rylev:
This includes a bump to the time crate version which used to compile but no longer does due to an addition of a trait impl to the std library. This was decided to be acceptable by the Rust project, and the fix is to bump the version of the `time` crate.

The rest of the changes are just small cosmetic changes to make the linters happy.

### DIFF
--- a/crates/core/tests/integration_test.rs
+++ b/crates/core/tests/integration_test.rs
@@ -160,7 +160,6 @@ async fn test_host_component_data_update() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-#[cfg(not(tarpaulin))]
 async fn test_panic() {
     let err = run_core_wasi_test(["panic"], |_| {}).await.unwrap_err();
     let trap = err.downcast::<Trap>().expect("trap");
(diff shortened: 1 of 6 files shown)


crates/core/tests/integration_test.rs
```diff
@@ -160,7 +160,6 @@ async fn test_host_component_data_update() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-#[cfg(not(tarpaulin))]
```
lann, line 163: What was this one about? IIRC this test really did break with tarpaulin...

lann: Ah: https://github.com/xd009642/tarpaulin?tab=readme-ov-file#ignoring-code-in-files

> Unfortunately, due to the unexpected cfg warnings cargo now emits you will likely want to add the recommended lints to your Cargo.toml, or utilise any existing build scripts.

lann: This seems to be what it means by "utilise any existing build scripts":

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-check-cfg

Anyway, doesn't need to be handled in this PR.

PR #2684, opened 2024-07-26
Change implementation of `incoming-request.authority`

rylev:
The guest will now see `incoming-request.authority` as whatever is set in the incoming request's `Host` header.

Previously we were relying on the definition of `incoming-request.authority` not being set by user input because outbound requests to the special `self` host would be ultimately directed towards whatever was in `incoming-request.authority`. Now, `self` requests are specifically defined (with scheme and authority) in the host. The implementation currently sets the `self` definition to be the same scheme as the incoming request and the authority is set to the listener address of the Spin web server.

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -280,7 +280,7 @@ impl HttpTrigger {
                 let executor = trigger.executor.as_ref().unwrap_or(&HttpExecutorType::Http);
                 // Set the definition of outbound requests to `self` to be equal to
                 // the incoming request's scheme and the bound listening address.
-                let self_definition = SelfRequestDefinition {
+                let self_origin = SelfRequestOrigin {
                     scheme,
                     authority: server_addr.to_string(),
                 };
@@ -294,7 +294,7 @@ impl HttpTrigger {
                                 &route_match,
                                 req,
                                 client_addr,
-                                self_definition,
+                                self_origin,
                             )
                             .await
                     }
@@ -309,7 +309,7 @@ impl HttpTrigger {
                                 &route_match,
                                 req,
                                 client_addr,
-                                self_definition,
+                                self_origin,
                             )
                             .await
                     }
@@ -493,18 +493,18 @@ fn parse_listen_addr(addr: &str) -> anyhow::Result<SocketAddr> {
 /// The incoming request's scheme and authority
 ///
 /// The incoming request's URI is relative to the server, so we need to set the scheme and authority.
-/// The `Host` header is used to set the authority. This function will error if not `Host` header is
+/// The `Host` header is used to set the authority. This function will error if no `Host` header is
 /// present or if it is not parseable as an `Authority`.
 fn set_req_uri(req: &mut Request<Body>, scheme: Scheme) -> Result<()> {
     let uri = req.uri().clone();
     let mut parts = uri.into_parts();
     let headers = req.headers();
-    let authority = headers
+    let host_header = headers
         .get(HOST)
         .context("missing Host header")?
         .to_str()
         .context("Host header is not valid UTF-8")?;
-    let authority = authority
+    let authority = host_header
         .parse()
         .context("Host header contains an invalid authority")?;
     parts.scheme = Some(scheme);
@@ -600,13 +600,13 @@ pub(crate) trait HttpExecutor: Clone + Send + Sync + 'static {
         route_match: &RouteMatch,
         req: Request<Body>,
         client_addr: SocketAddr,
-        self_definition: SelfRequestDefinition,
+        self_origin: SelfRequestOrigin,
     ) -> Result<Response<Body>>;
 }
 
-/// The definition of the `self` host for outbound requests.
+/// The origin of the `self` host for outbound requests.
 #[derive(Clone)]
-pub struct SelfRequestDefinition {
+pub struct SelfRequestOrigin {
     scheme: Scheme,
     authority: String,
 }
@@ -615,7 +615,7 @@ pub struct SelfRequestDefinition {
 struct ChainedRequestHandler {
     engine: Arc<TriggerAppEngine<HttpTrigger>>,
     executor: HttpHandlerExecutor,
-    self_definition: SelfRequestDefinition,
+    self_origin: SelfRequestOrigin,
 }
 
 #[derive(Default)]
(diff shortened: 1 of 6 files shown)


rylev:
FYI: The CI failure is expected as this breaks one of the conformance tests. Once we're happy with the changes here, I will change the conformance test to match this behavior.

crates/trigger-http/src/lib.rs
```diff
@@ -475,11 +492,21 @@ fn parse_listen_addr(addr: &str) -> anyhow::Result<SocketAddr> {
...
     let uri = req.uri().clone();
     let mut parts = uri.into_parts();
-    let authority = format!("{}:{}", addr.ip(), addr.port()).parse().unwrap();
+    let headers = req.headers();
+    let authority = headers
+        .get(HOST)
+        .context("missing Host header")?
+        .to_str()
+        .context("Host header is not valid UTF-8")?;
+    let authority = authority
```
tschneidereit, line 507:
```suggestion
    let host_header = headers
        .get(HOST)
        .context("missing Host header")?
        .to_str()
        .context("Host header is not valid UTF-8")?;
    let authority = host_header
```

tschneidereit: It took me a couple reads to understand the change in meaning for `authority`

crates/trigger-http/src/lib.rs
```diff
@@ -475,11 +492,21 @@ fn parse_listen_addr(addr: &str) -> anyhow::Result<SocketAddr> {
 
 /// The incoming request's scheme and authority
 ///
-/// The incoming request's URI is relative to the server, so we need to set the scheme and authority
-fn set_req_uri(req: &mut Request<Body>, scheme: Scheme, addr: SocketAddr) -> Result<()> {
+/// The incoming request's URI is relative to the server, so we need to set the scheme and authority.
+/// The `Host` header is used to set the authority. This function will error if not `Host` header is
```
tschneidereit, line 496:
```suggestion
/// The `Host` header is used to set the authority. This function will error if no `Host` header is
```

crates/trigger-http/src/wagi.rs
```diff
@@ -28,6 +28,7 @@ impl HttpExecutor for WagiHttpExecutor {
         route_match: &RouteMatch,
         req: Request<Body>,
         client_addr: SocketAddr,
+        _self_authority: SelfRequestDefinition,
```
tschneidereit, line 31:
```suggestion
        _self_definition: SelfRequestDefinition,
```

crates/trigger-http/src/lib.rs
```diff
@@ -573,13 +600,22 @@ pub(crate) trait HttpExecutor: Clone + Send + Sync + 'static {
         route_match: &RouteMatch,
         req: Request<Body>,
         client_addr: SocketAddr,
+        self_definition: SelfRequestDefinition,
     ) -> Result<Response<Body>>;
 }
 
+/// The definition of the `self` host for outbound requests.
+#[derive(Clone)]
+pub struct SelfRequestDefinition {
```
tschneidereit, line 609: Could this perhaps be `SelfRequestOrigin`? "Definition" is pretty vague and I doubt future readers would have a reasonable intuition for what it stands for.

The `origin` concept of course technically includes the port, too; one way to address that would be to simply include it. But I think it'd also be fine to just add a comment saying that for our purposes that part is irrelevant, so left out.

tschneidereit (approved):
This looks good to me. I left a few in-line comments and suggestions, but they're all minor naming quibbles.

crates/trigger-http/src/wagi.rs
lann: Another alternative:

```suggestion
        _self_base_uri: Uri,
```

crates/trigger-http/src/lib.rs
lann: `Authority` does include port, either explicitly or implicitly via scheme. This struct could also alternatively wrap (or be replaced by) a `Uri`.

rylev: I renamed to `SelfRequestOrigin` and kept its fields (`URI` contains a lot of other stuff we don't need and I think it makes the code harder to read).

tests/testing-framework/src/runtimes/in_process_spin.rs
```diff
@@ -40,12 +40,16 @@ impl InProcessSpin {
...
-                // TODO(rylev): convert headers and body as well
-                .body(spin_http::body::empty())
-                .unwrap();
+                .uri(req.path);
+
+            for (key, value) in req.headers {
+                builder = builder.header(*key, *value);
+            }
+
+            // TODO(rylev): convert body as well
```
lann, line 51: A `TODO` IN _YOUR_ CODE?!? :slightly_smiling_face:

crates/trigger-http/src/lib.rs
```diff
@@ -475,11 +492,21 @@ fn parse_listen_addr(addr: &str) -> anyhow::Result<SocketAddr> {
 
 /// The incoming request's scheme and authority
 ///
-/// The incoming request's URI is relative to the server, so we need to set the scheme and authority
-fn set_req_uri(req: &mut Request<Body>, scheme: Scheme, addr: SocketAddr) -> Result<()> {
+/// The incoming request's URI is relative to the server, so we need to set the scheme and authority.
+/// The `Host` header is used to set the authority. This function will error if no `Host` header is
```
lann, line 496: > The `Host` header is used to set the authority.

Should this also check the `req.uri().authority()`? I think it _might_ be most correct to use either one, verifying that they are identical if both are present.

lann: I'm pretty sure that's right, based on https://github.com/hyperium/hyper/issues/1612

radu-matei:
Does this need another set of reviews, or is it good to merge?
Thanks!

rylev:
We can land this whenever we feel comfortable with the breaking change.

One thought I had was that we'll need to bring these changes into the factors break. Instead of landing this in `main`, we could target the factors work so that this lands whenever factors does. Thoughts @radu-matei @lann?

radu-matei:
Given we know it's the right change, bringing it in together with the rest of the breaking changes from the factors work seems reasonable to me.

PR #2686, opened 2024-07-26
factors: Update runtime config and tests

lann:
I swear this started out as a focused commit...

Anyway: this is a random grab-bag of changes to simplify and homogenize factors runtime config and tests:

- Simplify runtime config handling in `RuntimeFactors::configure_app`
- Add `#[derive(Default)]` to derived `RuntimeFactors`' runtime configs; replace `()` as an "empty source" with `Default::default()`
- Change `TestEnvironment` to use a builder-style interface
- Simplify or inline some test helpers
- Homogenize factors tests a bit, including:
  - Rename all factor test files to `factor_test.rs` (commence bikeshedding!)
  - Remove some now-unnecessary `WasiFactor` deps


crates/factor-variables/src/spin_cli/env.rs
```diff
@@ -52,10 +52,12 @@ pub struct EnvVariablesConfig {
 
 const DEFAULT_ENV_PREFIX: &str = "SPIN_VARIABLE";
 
+type EnvFetcherFn = Box<dyn Fn(&str) -> Result<String, VarError> + Send + Sync>;
+
 /// A config Provider that uses environment variables.
 pub struct EnvVariablesProvider {
     prefix: Option<String>,
-    env_fetcher: Box<dyn Fn(&str) -> Result<String, VarError> + Send + Sync>,
+    env_fetcher: EnvFetcherFn,
```
lann, line 60: Clippy

crates/factors-test/src/lib.rs
```diff
@@ -27,60 +40,54 @@ impl Default for TestEnvironment {
...
+    /// Extends the manifest with the given TOML.
     ///
     /// The default manifest includes boilerplate like the
     /// `spin_manifest_version` and `[application]` section, so you typically
     /// need to pass only a `[component.test-component]` section.
-    pub fn default_manifest_extend(manifest_merge: toml::Table) -> Self {
-        let mut env = Self::default();
-        env.manifest.extend(manifest_merge);
-        env
+    pub fn extend_manifest(mut self, manifest_merge: toml::Table) -> Self {
```
rylev, line 55: This seems to be mimicking a builder pattern. Normally this would be a separate struct with a `build` method, but I think in this case, just mutating the `TestEnvironment` is fine. Just wanted to note it here as a place for potential future change.

crates/factor-sqlite/tests/factor_test.rs
```diff
@@ -62,26 +62,21 @@ async fn no_error_when_database_is_configured() -> anyhow::Result<()> {
...
         [sqlite_database.foo]
         type = "spin"
     };
     let sqlite_config = SpinSqliteRuntimeConfig::new("/".into(), "/".into());
-    if let Err(e) = env
-        .build_instance_state(
-            factors,
-            TomlRuntimeSource::new(&runtime_config, sqlite_config),
-        )
+    let env = TestEnvironment::new(factors)
```
rylev, line 70:
```suggestion
    let _ = TestEnvironment::new(factors)
```
I'm surprised this doesn't lead to a warning?

rylev (approved):
Looks great! I'm going to merge this just so we don't run into any conflicts down the line.

crates/factor-sqlite/tests/factor_test.rs
lann: `env` is used at 77

PR #2687, opened 2024-07-29
Non generic `KeyValueFactor`

rylev:
Removes the generic from the `KeyValueFactor` and separates out runtime config handling to its own Spin CLI specific module.

I suggest looking at the tests to get an idea of what changes have happened to the public API.


crates/factor-key-value/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,150 @@
...
+impl RuntimeConfigResolver {
+    /// Create a new RuntimeConfigResolver.
+    pub fn new() -> Self {
+        <Self as Default>::default()
+    }
+
+    /// Adds a default store configuration for a label.
+    ///
+    /// Users must ensure that the store type for `config` has been registered with
+    /// the resolver using [`Self::register_store_type`].
```
lann, line 64: Could probably require that this come after `register_store_type` and actually check that the type has been registered.

crates/factor-key-value/tests/factor_test.rs
```diff
@@ -224,7 +227,16 @@ async fn multiple_custom_key_value_uses_first_store() -> anyhow::Result<()> {
     Ok(())
 }
 
-struct TomlConfig(Option<toml::Table>);
+struct TomlConfig {
+    resolver: Arc<RuntimeConfigResolver>,
```
lann, line 231: This is quite awkward; I think we can probably do better by making the `Factor` available in `get_runtime_config`.

lann (approved):
I think this is revealing an opportunity to improve the factors interface but it gets the job done.

crates/factor-key-value/src/lib.rs
```diff
@@ -18,27 +14,28 @@ use spin_key_value::{
...
+    default_label_resolver: Arc<dyn DefaultLabelResolver>,
 }
 
-impl<R> KeyValueFactor<R> {
+impl KeyValueFactor {
     /// Create a new KeyValueFactor.  
     ///  
-    /// The `runtime_config_resolver` is used to resolve runtime configuration into store managers.
-    pub fn new(runtime_config_resolver: R) -> Self {
+    /// The `default_label_resolver` is used to resolver store managers for labels that
```
kate-goldenring, line 28:
```suggestion
    /// The `default_label_resolver` is used to resolve store managers for labels that
```

crates/factor-key-value/src/lib.rs
```diff
@@ -52,23 +49,10 @@ impl<R: RuntimeConfigResolver + 'static> Factor for KeyValueFactor<R> {
...
     ) -> anyhow::Result<Self::AppState> {
-        // Build store manager from runtime config
-        let mut store_managers: HashMap<String, Arc<dyn StoreManager>> = HashMap::new();
-        if let Some(runtime_config) = ctx.take_runtime_config() {
-            for (store_label, config) in runtime_config.store_configs {
-                if let std::collections::hash_map::Entry::Vacant(e) =
-                    store_managers.entry(store_label)
-                {
-                    // Only add manager for labels that are not already configured. Runtime config
-                    // takes top-down precedence.
```
kate-goldenring, line 63: I see the test asserting this is passing but not where this logic is ported to. Can you point it out to me?

kate-goldenring (approved):

crates/factor-key-value/src/lib.rs
rylev: This is just verifying that we only ever have one store manager per label right? I don't see a test that asserts this. Can you point that out to me?

crates/factor-key-value/src/runtime_config.rs
```diff
@@ -1,28 +1,21 @@
...
 /// Runtime configuration for all key value stores.
-#[derive(Deserialize)]
-#[serde(transparent)]
-pub struct RuntimeConfig<C> {
-    /// Map of store names to store configurations.
-    pub store_configs: HashMap<String, C>,
+#[derive(Default)]
+pub struct RuntimeConfig {
+    /// Map of store names to store managers.
+    pub store_managers: HashMap<String, Arc<dyn StoreManager>>,
```
lann, line 11: Hide this behind an e.g. `add_store_manager` method?

crates/factor-key-value/src/lib.rs
kate-goldenring: Not exactly. You should be able to list multiple managers for the same label but the top one will be used (the variables provider works like this too). The `multiple_custom_key_value_uses_first_store` test checks this.

crates/factor-key-value/tests/factor_test.rs
rylev: Lann and I talked about this, and came to the tentative conclusion that while having shared state between the factor and the runtime config source is maybe not 100% ideal, it works well in this case. We'll continue to keep an eye on this area and continue to adjust as we get more comfortable with the consequences of the current design.

crates/factor-key-value/src/lib.rs
rylev: Ah I think this is just a consequence of deserializing the toml. When toml is deserialized as a HashMap it just takes the first key (in this case the label `custom`) and ignores repeats. So none of the code we write explicitly does this, it happens automatically when we deserialize.

PR #2689, opened 2024-07-29
Factors: Vault variable provider

rylev:
Adds a variable provider for vault.


rylev:
@lann ping - I'm not sure why the other CI jobs ran.

lann (approved):

PR #2692, opened 2024-07-30
Add `spin-factors-executor`

lann:
This crate provides a `FactorsExecutor` type which extracts a large chunk of the logic currently managed by `spin_trigger::TriggerExecutor`. This represents `spin up`'s particular approach to preinitializing apps and will be the primary interface for triggers to instantiate components.

Also replaced some `AsMut` bounds with new traits to avoid coherence problems. I should have done this initially but I was lazy and thought I could get away with it. :neutral_face:

### DIFF
--- a/crates/factors-executor/src/lib.rs
+++ b/crates/factors-executor/src/lib.rs
@@ -1,0 +1,223 @@
+use std::collections::HashMap;
+
+use anyhow::Context;
+use spin_app::{App, AppComponent};
+use spin_core::Component;
+use spin_factors::{AsInstanceState, ConfiguredApp, RuntimeFactors, RuntimeFactorsInstanceState};
+
+/// A FactorsExecutor manages execution of a Spin app.
+///
+/// `Factors` is the executor's [`RuntimeFactors`]. `ExecutorInstanceState`
+/// holds any other per-instance state needed by the caller.
+pub struct FactorsExecutor<T: RuntimeFactors, U = ()> {
+    factors: T,
+    core_engine: spin_core::Engine<InstanceState<T::InstanceState, U>>,
+    configured_app: ConfiguredApp<T>,
+    // Maps component IDs -> InstancePres
+    component_instance_pres: HashMap<String, InstancePre<T, U>>,
+}
+
+type InstancePre<T, U> =
+    spin_core::InstancePre<InstanceState<<T as RuntimeFactors>::InstanceState, U>>;
+
+impl<T: RuntimeFactors, U: Send + 'static> FactorsExecutor<T, U> {
+    /// Constructs a new executor.
+    pub fn new(
+        core_config: &spin_core::Config,
+        mut factors: T,
+        app: App,
+        mut component_loader: impl ComponentLoader,
+        runtime_config: T::RuntimeConfig,
+    ) -> anyhow::Result<Self> {
+        let core_engine = {
+            let mut builder =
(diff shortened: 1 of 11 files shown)


lann:
woops, didn't even try to run the tests

crates/factors-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+    core_engine: spin_core::Engine<InstanceState<Factors::InstanceState, ExecutorInstanceState>>,
+    configured_app: ConfiguredApp<Factors>,
+    component_instance_pres: HashMap<String, InstancePre<Factors, ExecutorInstanceState>>,
+}
+
+/// A ComponentLoader is responsible for loading Wasmtime [`Component`]s.
+pub trait ComponentLoader {
+    /// Loads a [`Component`] for the given [`AppComponent`].
+    fn load_component(&mut self, component: &AppComponent) -> anyhow::Result<Component>;
+}
```
rylev, line 23: Nit: can we move this somewhere down below. I think it's more readable to have the `FactorsExecutor` impl block next to the struct definition.

crates/factors-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// A FactorsExecutor manages execution of a Spin app.
+///
+/// `Factors` is the executor's [`RuntimeFactors`]. `ExecutorInstanceState`
+/// holds any other per-instance state needed by the caller.
+pub struct FactorsExecutor<Factors: RuntimeFactors, ExecutorInstanceState = ()> {
+    factors: Factors,
+    core_engine: spin_core::Engine<InstanceState<Factors::InstanceState, ExecutorInstanceState>>,
+    configured_app: ConfiguredApp<Factors>,
+    component_instance_pres: HashMap<String, InstancePre<Factors, ExecutorInstanceState>>,
```
rylev, line 16: Nit: can you document what `component_instance_pres` is? At the very least it would be nice to know what the keys are without having to read the code.

crates/factors-executor/src/lib.rs
```diff
@@ -0,0 +1,167 @@
...
+pub trait ComponentLoader {
+    /// Loads a [`Component`] for the given [`AppComponent`].
+    fn load_component(&mut self, component: &AppComponent) -> anyhow::Result<Component>;
+}
+
+type InstancePre<Factors, ExecutorInstanceState> = spin_core::InstancePre<
+    InstanceState<<Factors as RuntimeFactors>::InstanceState, ExecutorInstanceState>,
+>;
+
+impl<Factors: RuntimeFactors, ExecutorInstanceState: Send + 'static>
```
rylev, line 29: Nit: I very much prefer short generic names like `F`. I believe we've had a discussion about this before and have differing opinions, but I personally find the more verbose names trick me into thinking that there is a concrete type called `Factors` instead of reminding me that we're dealing with a generic.

rylev (approved):
Just some nits, but otherwise looks good!

crates/factors-executor/src/lib.rs
lann: I mostly find it confusing when _writing_ code but yeah I'm fine shortening these.

crates/factors-executor/src/lib.rs
lann: done

crates/factors-executor/src/lib.rs
lann: done

Issue #2690, opened 2024-07-29
Spin 2.7 release 


me-diru:
A decision was made on the July 22nd Spin community meeting to have Spin 2.7 release
https://docs.google.com/document/d/1EG392gb8Eg-1ZEPDy18pgFZvMMrdAEybpCSufFXoe00/edit#bookmark=id.mqvnq7lc0rxb

Release process
https://github.com/fermyon/spin/blob/main/docs/content/release-process.md

- [x]  Release engineer assigned @me-diru 
- [x]  Release manager assigned @vdice 

Issues/Commits to track
- [x]  #2683 

Not sure about blockers
Referenced #2570

me-diru:
cc: @vdice, maybe you can be the release manager?

itowlson:
Can this be closed or are there outstanding tasks?

vdice:
Thanks @itowlson we can close this.  The last remaining work item is docs-related (https://github.com/fermyon/developer/pull/1358) but the release has indeed been cut.

PR #2695, opened 2024-07-30
Bump versions for v2.7 release

me-diru:
related to #2690

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1504,7 +1504,7 @@ dependencies = [
 
 [[package]]
 name = "conformance"
-version = "2.7.0-pre0"
+version = "2.7.0"
 dependencies = [
  "anyhow",
  "conformance-tests",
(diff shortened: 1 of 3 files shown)


vdice (commented):
For the version bumps on main, we'd want to bump to the prerelease for the next anticipated release.  In this case, `v2.8.0-pre0`.

me-diru:
thanks! Should we go for `v3.0.0-pre0` instead, as we are expecting Spin `3.0` as our next release.

itowlson:
I would avoid making the assumption that 3.0 will be next.  I'd stick to 2.8-pre for now, and rev to 3.0 once we commit to "counting down to 3.0".

itowlson:
Basically rev to 3.0-pre once we start landing breaking changes.  Unless we expect to start doing so near-immediately, in which case :shipit:

me-diru:
I think we already wanted to go for 3.0 after 2.6, however, 2.7 was the last anticipated version. I think spin-factors and other work will lead to breaking changes that will require a major release of `3.0`. @lann might give better input on that end.

PR #2697, opened 2024-08-01
Run v0.1.0 of the conformance tests

rylev:
This allows us to evolve the conformance tests without immediately breaking Spin CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1515,7 +1515,7 @@ dependencies = [
 [[package]]
 name = "conformance-tests"
 version = "0.1.0"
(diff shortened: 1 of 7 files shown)


lann (approved):
I wonder if it would make sense for the conformance tests versions to follow the Spin version? Maybe just the major/minor numbers; the patch number could be decoupled.

rylev:
I think that might become difficult to manage for a few reasons:
* We will likely want to bump the major version of the tests when we introduce a new schema of test description (i.e., the json file that describes the test)
* Other runtimes will evolve separately and at different paces, and I'm not sure we want the version of Spin CLI to be the thing that ties them together.

I personally think, at least for now, it makes sense for the tests to evolve as they do and not tie them to any external versioning scheme. In the future, we can revisit this question when the rate of change of the tests slows and we have a better picture of how versioning across the entire Spin ecosystem might work.

PR #2698, opened 2024-08-01
update node version for integration test

karthik2804:


tests/integration.rs
```diff
@@ -617,10 +617,10 @@ Caused by:
             "http-ts",
             Some("https://github.com/fermyon/spin-js-sdk"),
             None,
-            &["js2wasm"],
+            &[],
             prebuild,
             HashMap::default(),
-            "Hello from TS-SDK",
+            "hello universe",
```
rylev, line 623: Just curious. Why was this changed?

rylev: Bit of a shame - I think the previous versions were better, but this isn't the place to discuss that I suppose.

rylev (approved):

Issue #2677, opened 2024-07-23
Spin Docker image


vdice:
Do we want to also publish a Docker image wrapping the spin binary on each release? (eg `ghcr.io/fermyon/spin:v2.6.0`)

One recent use case is in the [SpinKube project](https://github.com/spinkube): having the ability to run a workload (Spin app) via such an image for certain scenarios (when, say, the standard containerd shim isn't used).

However, perhaps there are other scenarios where users would like to run Spin in container form?

If Yes, we'd like to publish, we have a few options to consider:

1. A lightweight (as much as possible) image (eg distroless) containing a statically-linked Spin binary (would need to be linux amd64 or arm64 currently).  No default plugins/templates/etc included.  Image size around 150MB-ish (static binary itself looks to be 110MB)
2. A heavierweight image containing distro, glibc, etc, containing a dynamically-linked Spin binary.  This could be installed via Spin's standard install script and include default plugins/templates/etc.  More in the 500-700MB range (haven't looked super in-depth at min size)
3. Other (somewhere in between?)

itowlson:
I like this.  A lightweight image would presumably also be able to work as a base for users to create customised images with the plugins etc. required for their use case, is that right?  I guess that would require a distroful image though?

Your mention of templates makes me wonder about targets for this work though.  Do you envisage this as primarily about running Spin workloads, or do you want it to enable the Spin developer loop?  I'm not sufficiently au fait with Docker to know how friendly the "new > build > up > commit" loop would be in the latter case.

calebschoepp:
@vdice chime in if you disagree: I envisage this image being purely focused on running Spin workloads not being used for the developer loop. Templates is really a bit of a red herring, the more important thing gained from the install script is plugins which would allow you to run custom triggers (although I suppose there aren't any custom triggers in the default plugins anyways).

Yep using the lightweight image as a base makes sense to me. The tricky part is making sure that dynamic linking works for the plugins since the lightweight base image uses a static binary.

Overall I'm a +1 for putting the image here instead of SpinKube.

One other thing worth noting is that I'd love for this to generate images for all versions of Spin (greater than some reasonable cutoff point like 2.5 🤷 ).

calebschoepp:
Your proposal makes sense to me. I'm fine with the naming scheme you proposed.

Would the midweight image still expose an entrypoint of Spin?

vdice:
> Would the midweight image still expose an entrypoint of Spin?

[...]

PR #2700, opened 2024-08-01
feat(docker): add spin cli Dockerfiles; add build/push to release.yml

vdice:
Adds Dockerfiles/CI for producing images wrapping the spin CLI on releases.  These may be generally useful but were recently motivated by utilization in SpinKube.  See https://github.com/fermyon/spin/issues/2677 for further motivation/discussion.

On a Spin release, the default image, eg `ghcr.io/fermyon/spin:v2.8.0`, has a Debian base (currently `bookworm-slim`) and includes the standard, dynamically-linked Spin binary, with a few utilities to enable further extension if desired (eg installing plugins).  A smaller, distroless image is also produced, eg `ghcr.io/fermyon/spin:v2.8.0-distroless`, which bundles a statically-linked Spin binary for purposes when, say, only `spin up --from ...` is needed and further extensibility isn't required.

Here I've opted to pull in binaries built in the previous build steps, as opposed to attempting to (re)-run the build in the container itself.  This cuts out potential headaches of debugging failed Docker builds and repurposes the same artifacts that are installed/run directly (sans container).

Closes https://github.com/fermyon/spin/issues/2677

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -380,3 +380,80 @@ jobs:
           repository: fermyon/homebrew-tap
           event-type: spin-release
           client-payload: '{"version": "${{ github.ref_name }}"}'
+
+  docker:
+    runs-on: "ubuntu-20.04"
+    needs: [build-and-sign, build-spin-static]
+    permissions:
+      contents: read
+      packages: write
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
+    strategy:
+      matrix:
+        config:
+          - { dockerfile: "Dockerfile", tag-suffix: "" }
+          - { dockerfile: "distroless.Dockerfile", tag-suffix: "-distroless" }
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup version info
+        id: version
+        run: |
+          if [[ "${{ startsWith(github.ref, 'refs/tags/v') }}" == "true" ]]; then
+            echo "version=${{ github.ref_name }}" >> $GITHUB_OUTPUT
+          else
+            echo "version=canary" >> $GITHUB_OUTPUT
+          fi
+
+      - name: download release assets
+        uses: actions/download-artifact@v3
+        with:
+          name: spin
+
+      - name: extract binaries
+        shell: bash
+        run: |
+          if [[ "${{ matrix.config.tag-suffix }}" == "-distroless" ]]; then
+            static="-static"
+          fi
+          tar xvf spin-${{ steps.version.outputs.version }}${static}-linux-amd64.tar.gz
+          mv spin spin${static}-linux-amd64
+          tar xvf spin-${{ steps.version.outputs.version }}${static}-linux-aarch64.tar.gz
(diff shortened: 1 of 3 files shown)


.github/Dockerfile.distroless
```diff
@@ -0,0 +1,7 @@
+FROM gcr.io/distroless/static-debian12
```
calebschoepp, line 1: Nit: It's kind of annoying that we don't get syntax highlighting here b/c we don't have a `.Dockerfile` postfix. I'm assuming you chose this name ordering for a reason though. So if that is the case then 🤷 I guess we just deal with no syntax highlighting.

calebschoepp (commented):
Would these images be published already for me to test?

.github/Dockerfile.distroless
vdice: No specific reason for ordering; changed to `distroless.Dockerfile` and it looks like syntax highlighting is back!

vdice:
> Would these images be published already for me to test?

Yes: `ghcr.io/vdice/spin:v2.7.0-vdice-test` and `ghcr.io/vdice/spin:v2.7.0-vdice-test-distroless`

calebschoepp (approved):
I tried both of the images and they both worked on my draft spintainer PR. LGTM!

.github/workflows/release.yml
```diff
@@ -380,3 +380,80 @@ jobs:
           repository: fermyon/homebrew-tap
           event-type: spin-release
           client-payload: '{"version": "${{ github.ref_name }}"}'
+
+  docker:
+    runs-on: "ubuntu-20.04"
```
rajatjindal, line 385: possibly use 22.04?

.github/workflows/release.yml
```diff
@@ -380,3 +380,80 @@ jobs:
...
+      contents: read
+      packages: write
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: ${{ github.repository }}
+    strategy:
+      matrix:
+        config:
+          - { dockerfile: "Dockerfile", tag-suffix: "" }
+          - { dockerfile: "distroless.Dockerfile", tag-suffix: "-distroless" }
```
rajatjindal, line 397: Hi @vdice, just a curious question: does `distroless` means `static` implicitly?

vdice: @rajatjindal Yes; well both.  It uses a distroless base image and includes the static Spin binary.  Open to alternative names... let me know.

.github/workflows/release.yml
vdice: I propose relegating to a follow-up; this is the version used elsewhere in the workflow so it might be nice to update all at the same time.

rajatjindal: Yes 👍🏻 that makes sense. Thank you

.github/workflows/release.yml
rajatjindal: I was wondering if it is worth adding static to the name, if distroless implies static by default, then we can omit it. But otherwise it would be nice to call it out explicitly.

vdice: Indeed, I do think "distroless implies static by default".

radu-matei (approved):

PR #2701, opened 2024-08-01
factors: Split `HttpHandlerExecutor` into `SpinHttpExecutor` and `WasiHttpExecutor`

lann:
Also break out common headers logic into `mod headers`.


fibonacci1729 (approved):
Looks good!

PR #2703, opened 2024-08-01
factor-outbound-networking: Add support for client TLS

lann:
Port the `[[client_tls]]` runtime config logic from `spin-trigger` to `spin-factor-outbound-networking`.

This rewords the keys used in that runtime config section, which is a breaking change but I think its less confusing and worth it given that ~no one is using it yet and it will be released with 3.0.

This also switch from using `Authority` to a plain string for hostnames. Using `Authority` was my idea but I realize now that it makes things awkward when trying to deal with port numbers, so I bailed out to just plain `host: String`, though still using `Authority`'s parser to validate that string.

The `rustls::ClientConfig` derived from this runtime config is exposed to dependent factors via `InstanceBuilder::component_tls_configs`.


lann:
@rajatjindal FYI

lann:
@rylev I'll be particularly interested in your thoughts on the RuntimeConfig approach.

crates/factor-outbound-networking/testdata/invalid-cert.pem
lann, line 1: These are copied from `spin-trigger-http` tests.

crates/factor-outbound-networking/src/runtime_config.rs
```diff
@@ -0,0 +1,250 @@
...
+                "client TLS 'hosts' list may not be empty"
+            );
+            let client_config = Arc::new(tls_config.to_client_config()?);
+            for component in &tls_config.components {
+                let host_configs = component_host_client_configs
+                    .entry(component.clone())
+                    .or_default();
+                for host in &tls_config.hosts {
+                    validate_host(host)?;
+                    // First matching (component, host) pair wins
```
rylev, line 40: This comment is confusing to me. I think this means that if we've seen this component, host pair before, we ignore this new configuration.

In any case, I think we should probably error rather than silently ignore.

crates/factor-outbound-networking/src/runtime_config.rs
```diff
@@ -0,0 +1,250 @@
...
+                    Arc::get_mut(host_configs)
+                        .unwrap()
+                        .entry(host.clone())
+                        .or_insert_with(|| client_config.clone());
+                }
+            }
+        }
+
+        let default_client_config = Arc::new(TlsConfig::default().to_client_config()?);
```
rylev, line 49: Nit: I'm not sure that a `Default` impl for `TlsConfig makes a ton of sense considering we treat empty lists of components or hosts as errors elsewhere. 

Instead I think it makes more sense to just have a function for getting a default client config. I recognize this would require refactoring `to_client_config` and you might not want to bother, but I do think it makes the code easier to understand.

crates/factor-outbound-networking/src/runtime_config.rs
```diff
@@ -0,0 +1,250 @@
...
+    // Validate hostname
+    http::uri::Authority::from_str(host).with_context(|| format!("invalid TLS 'host' {host:?}"))?;
+    if host.contains(':') {
+        anyhow::bail!("invalid TLS 'host' {host:?}; ports not currently supported");
+    }
+    Ok(())
+}
+
+#[derive(Clone)]
+pub struct ComponentTlsConfigs {
```
rylev, line 89:
```suggestion
/// TLS configurations for a specific component.
#[derive(Clone)]
pub struct ComponentTlsConfigs {
```

crates/factor-outbound-networking/src/runtime_config.rs
```diff
@@ -0,0 +1,250 @@
...
+        for tls_config in tls_configs {
+            ensure!(
+                !tls_config.components.is_empty(),
+                "client TLS 'components' list may not be empty"
+            );
+            ensure!(
+                !tls_config.hosts.is_empty(),
+                "client TLS 'hosts' list may not be empty"
+            );
+            let client_config = Arc::new(tls_config.to_client_config()?);
```
rylev, line 33: The errors coming from `to_client_config` will likely be extremely specific to the `ClientConfig` creation process. At the very least, we should add `.context` call here to let the caller know there was an error when trying to create a tls configuration.

crates/factor-outbound-networking/src/runtime_config.rs
```diff
@@ -0,0 +1,250 @@
...
+        }])?;
+        let client_config = runtime_config.get_client_config("test-component", "test-host");
+        // Check that we didn't just get the default
+        let default_config = runtime_config.get_client_config("other_component", "test-host");
+        assert!(!Arc::ptr_eq(&client_config, &default_config));
+        Ok(())
+    }
+
+    #[test]
+    fn test_maximal_config() -> anyhow::Result<()> {
```
rylev, line 189: Nit: I'm not sure this a maximal config considering `use_webpki_roots` is set to false?

```suggestion
    fn test_custom_client_cert() -> anyhow::Result<()> {
```

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+};
+
+use super::{validate_host, TlsConfig};
+
+pub struct SpinTlsRuntimeConfig {
+    runtime_config_dir: PathBuf,
+}
+
+impl SpinTlsRuntimeConfig {
+    pub fn new(runtime_config_dir: impl Into<PathBuf>) -> Self {
```
rylev, line 17: Let's document `runtime_config_dir`. This is the directory that that relative paths in the runtime config toml will be resolved against, correct? Perhaps we can name a document that so this is clear without needing to review the code.

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+    ) -> anyhow::Result<Option<Vec<TlsConfig>>> {
+        let Some(array) = table.get("client_tls") else {
+            return Ok(None);
+        };
+        let toml_configs: Vec<RuntimeConfigToml> = array.clone().try_into()?;
+
+        let tls_configs = toml_configs
+            .into_iter()
+            .map(|toml_config| self.load_tls_config(toml_config))
+            .collect::<anyhow::Result<Vec<_>>>()?;
```
rylev, line 58: I think we want  a `context` here so that the `load_tls_config` errors are shown to the user in the context of parsing the runtime config toml file. Right now that might not be clear to the user.

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+            !hosts.is_empty(),
+            "[[client_tls]] 'hosts' list may not be empty"
+        );
+
+        let components = component_ids.into_iter().map(Into::into).collect();
+
+        let hosts = hosts
+            .iter()
+            .map(|host| {
+                let host: &str = host;
```
rylev, line 85: I believe this isn't needed?

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+        let path = self.runtime_config_dir.join(path);
+        rustls_pemfile::certs(&mut io::BufReader::new(fs::File::open(path).map_err(
+            |err| {
+                io::Error::new(
+                    io::ErrorKind::InvalidInput,
+                    format!("failed to read cert file {:?}", err),
+                )
+            },
+        )?))
+        .collect::<io::Result<Vec<rustls_pki_types::CertificateDer<'static>>>>()
```
rylev, line 136:
```suggestion
        .collect()
```

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    // Parse a private key from the provided file
+    fn load_key(
+        &self,
+        path: impl AsRef<Path>,
+    ) -> anyhow::Result<rustls_pki_types::PrivateKeyDer<'static>> {
+        let path = self.runtime_config_dir.join(path);
+        rustls_pemfile::private_key(&mut io::BufReader::new(
+            fs::File::open(path).context("loading private key")?,
```
rylev, line 146:
```suggestion
            fs::File::open(path).with_context(|| format!("failed to load private key from '{}'", path.display()))?,
```

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+        ))
+        .map_err(|_| anyhow::anyhow!("invalid input"))
```
rylev, line 148: The error here can probably be improved.

```suggestion
        .with_context(|_| format!("failed to get private key from '{}'", path.display()))
```

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+        .transpose()
+        .ok_or_else(|| {
+            io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "private key file contains no private keys",
```
rylev, line 153:
```suggestion
                format!("private key file '{}' contains no private keys", path.display()),
```

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -0,0 +1,273 @@
...
+    ca_use_webpki_roots: Option<bool>,
+    ca_roots_file: Option<PathBuf>,
+    client_cert_file: Option<PathBuf>,
+    client_private_key_file: Option<PathBuf>,
+}
+
+fn deserialize_hosts<'de, D: Deserializer<'de>>(deserializer: D) -> Result<Vec<String>, D::Error> {
+    let hosts = Vec::<String>::deserialize(deserializer)?;
+    for host in &hosts {
+        validate_host(host).map_err(serde::de::Error::custom)?;
```
rylev, line 174: If we're going to validate the host later, is there really a need for it here?

rylev (approved):
I've got lots of nits, but this looks good overall 🎉

crates/factor-outbound-networking/src/runtime_config.rs
```diff
@@ -0,0 +1,250 @@
...
+        let component_config = self.get_component_tls_configs(component_id);
+        component_config.get_client_config(host).clone()
+    }
+}
+
+pub(crate) fn validate_host(host: &str) -> anyhow::Result<()> {
+    // Validate hostname
+    http::uri::Authority::from_str(host).with_context(|| format!("invalid TLS 'host' {host:?}"))?;
+    if host.contains(':') {
+        anyhow::bail!("invalid TLS 'host' {host:?}; ports not currently supported");
```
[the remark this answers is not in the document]
lann: The behavior didn't seem obvious enough to me to roll it into a refactor.

crates/factor-outbound-networking/src/runtime_config.rs
lann: It would probably make sense to reject two sections if their `components`/`hosts` _all_ match, but it might be convenient to allow overlap.

crates/factor-outbound-networking/src/runtime_config.rs
lann: It's just to enable the `TlsConfig { ..., ..Default::default() }` idiom :shrug:

crates/factor-outbound-networking/src/runtime_config.rs
lann: It is "maximally different from the defaults" :slightly_smiling_face:

crates/factor-outbound-networking/src/runtime_config/spin.rs
lann: Validating here gives nice TOML errors.

PR #2704, opened 2024-08-02
Factors outbound http

lann:
Pending #2703, and probably some tests (?)


crates/factor-outbound-http/src/lib.rs
```diff
@@ -38,19 +51,90 @@ impl Factor for OutboundHttpFactor {
...
+    /// If this method returns `Some(...)`, the inner result will be returned as
+    /// the result of the request, bypassing the default handler.
+    fn intercept(&self, intercepted: Intercepted)
+        -> Option<HttpResult<HostFutureIncomingResponse>>;
+}
+
+pub struct Intercepted<'a> {
+    pub request: &'a mut Request,
+    pub config: &'a mut OutgoingRequestConfig,
+}
```
lann, line 140: I thought I was going to need more fields on `Intercepted`. Could probably flatten them into `intercept` args.

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -61,10 +70,261 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
...
+
+async fn send_request_impl(
+    mut request: Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
+    mut config: wasmtime_wasi_http::types::OutgoingRequestConfig,
+    allowed_hosts: OutboundAllowedHosts,
+    tls_client_config: Arc<ClientConfig>,
+) -> anyhow::Result<Result<IncomingResponse, ErrorCode>> {
+    let allowed_hosts = allowed_hosts.resolve().await?;
+
+    let is_relative_url = request.uri().authority().is_none();
```
lann, line 115: This skips a lot of logic from the current impl but afaict `Authority` enforces everything we care about.

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -61,10 +70,261 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
...
+
+        let path_and_query = request.uri().path_and_query().cloned();
+        *request.uri_mut() = origin.into_uri(path_and_query);
+    } else {
+        let outbound_url = OutboundUrl::parse(request.uri().to_string(), "https")
+            .map_err(|_| ErrorCode::HttpRequestUriInvalid)?;
+        if !allowed_hosts.allows(&outbound_url) {
+            return handle_not_allowed(request.uri(), false);
+        }
+    }
```
lann, line 139: This is significantly reorganized but hopefully easier to follow.

crates/trigger-http2/src/lib.rs
```diff
@@ -296,7 +297,6 @@ mod tests {
 
         let default_headers = compute_default_headers(req.uri(), host, &route_match, client_addr)?;
 
-        // TODO: we currently replace the scheme with HTTP. When TLS is supported, this should be fixed.
```
lann, line 299: TODID

crates/trigger-http2/src/outbound_http.rs
```diff
@@ -0,0 +1,64 @@
...
+            let resp_fut = async move {
+                match server
+                    .handle_trigger_route(req, route_match, CHAINED_CLIENT_ADDR)
+                    .await
+                {
+                    Ok(resp) => Ok(Ok(IncomingResponse {
+                        resp,
+                        between_bytes_timeout,
+                        worker: None,
+                    })),
```
lann, line 50: I don't really understand whats going on here (copied from spin-trigger-http) but it compiles!

crates/trigger-http2/src/outbound_http.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl spin_factor_outbound_http::OutboundHttpInterceptor for OutboundHttpInterceptor {
+    fn intercept(
+        &self,
+        intercepted: spin_factor_outbound_http::Intercepted,
+    ) -> Option<wasmtime_wasi_http::HttpResult<spin_factor_outbound_http::HostFutureIncomingResponse>>
+    {
+        let uri = intercepted.request.uri();
+
+        // Handle service chaining requests
```
lann, line 34: @itowlson Porting service chaining to factors

crates/trigger-http2/src/server.rs
```diff
@@ -119,71 +122,91 @@ impl HttpServer {
...
-                match res {
-                    Ok(res) => Ok(MatchedRoute::with_response_extension(
-                        res,
-                        route_match.raw_route(),
-                    )),
-                    Err(err) => {
-                        tracing::error!("Error processing request: {err:?}");
-                        instrument_error(&err);
-                        Self::internal_error(None, route_match.raw_route())
+        let component_id = route_match.component_id();
```
lann, line 146: This part of the diff is mostly indentation.

crates/trigger-http2/src/server.rs
```diff
@@ -119,71 +122,91 @@ impl HttpServer {
...
+    ) -> anyhow::Result<Response<Body>> {
         let app_id = self
             .trigger_app
             .app()
             .get_metadata(APP_NAME_KEY)?
             .unwrap_or_else(|| "<unnamed>".into());
 
-        // Route to app component
-        match self.router.route(&path) {
-            Ok(route_match) => {
```
lann, line 130: This diff chunk is mostly whitespace (dedent).

crates/factor-outbound-http/src/lib.rs
```diff
@@ -38,19 +51,90 @@ impl Factor for OutboundHttpFactor {
...
+
+impl InstanceState {
+    pub fn set_request_interceptor(
+        &mut self,
+        interceptor: impl OutboundHttpInterceptor + 'static,
+    ) -> anyhow::Result<()> {
+        if self.request_interceptor.is_some() {
+            anyhow::bail!("set_request_interceptor can only be called once");
+        }
```
rylev, line 80: Are we sure we don't want to just silently overwrite the interceptor?

crates/factor-outbound-http/src/lib.rs
```diff
@@ -38,19 +51,90 @@ impl Factor for OutboundHttpFactor {
...
+    fn use_tls(&self) -> bool {
+        self.scheme == Scheme::HTTPS
+    }
+
+    fn host_header(&self) -> HeaderValue {
+        HeaderValue::from_str(self.authority.as_str()).unwrap()
+    }
+}
+
+pub trait OutboundHttpInterceptor: Send + Sync {
```
rylev, line 125: Nit: some docs on the trait itself (if for nothing else than to help with IDE hover overlays).

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -61,10 +70,261 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
...
+            let intercepted = Intercepted {
+                request: &mut request,
+                config: &mut config,
+            };
+            if let Some(res) = interceptor.intercept(intercepted) {
+                return res;
+            }
+        }
+
+        let host = request.uri().host().unwrap_or_default();
```
rylev, line 86: Are we asking for issues if `host` is `""`? I guess this _should_ return the default tls config, and everything will just work, but it does seem a bit round about, and I worry things might get confusing with future code changes.

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -61,10 +70,261 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
...
+        current_span.record("server.address", authority.host());
+        if let Some(port) = authority.port() {
+            current_span.record("server.port", port.as_u16());
+        }
+    }
+
+    Ok(send_request_handler(request, config, tls_client_config).await)
+}
+
+// TODO(factors): Move to some callback on spin-factor-outbound-networking (?)
```
rylev, line 152: I'd like to figure this out sooner rather than later since we're requiring a fairly Spin CLI specific logic (e.g., the use of `terminal::warn`) on every user of this factor. It would be nice to ensure we can handle this outside of the factor.

crates/factor-outbound-http/src/lib.rs
```diff
@@ -38,19 +51,90 @@ impl Factor for OutboundHttpFactor {
...
+    /// Intercept an outgoing HTTP request.
+    ///
+    /// If this method returns `None`, the (possibly updated) request and config
+    /// will be passed on to the default outgoing request handler.
+    ///
+    /// If this method returns `Some(...)`, the inner result will be returned as
+    /// the result of the request, bypassing the default handler.
+    fn intercept(&self, intercepted: Intercepted)
```
rylev, line 133: Instead of this returning an `Option` and sneaking the origin through the request extensions, I wonder if instead we could make `OutboundHttpInterceptor` generic and have:

```rust
pub trait OutboundHttpInterceptor<T>: Send + Sync {
    fn intercept(&self, intercepted: Intercepted) -> Intercepted<T>;
}

enum InterceptionResult<T> {
    Intercepted(HttpResult<HostFutureIncomingResponse>),
    NotIntercepted(T)
}
```

Then in `impl` the origin is stored in `InterceptionResult::NotIntercepted` which eliminates the implicit storing of the information in the request.

There's no need to block on this idea, we can also try it out after we merge this.

crates/factor-outbound-http/tests/factor_test.rs
```diff
@@ -34,11 +35,12 @@ async fn disallowed_host_fails() -> anyhow::Result<()> {
...
 
     let req = Request::get("https://denied.test").body(Default::default())?;
-    let res = wasi_http.send_request(req, test_request_config());
-    let Err(err) = res else {
-        bail!("expected Err, got Ok");
+    let mut future_resp = wasi_http.send_request(req, test_request_config())?;
+    future_resp.ready().await;
+    match future_resp.unwrap_ready() {
+        Ok(_) => bail!("expected Err, got Ok"),
+        Err(err) => assert!(matches!(err.downcast()?, ErrorCode::HttpRequestDenied)),
```
rylev, line 42: Nit:
```
future_resp.unwrap_ready().unwrap_err()
```

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -157,11 +157,19 @@ impl FactorInstanceBuilder for InstanceBuilder {
...
 // TODO: Refactor w/ spin-outbound-networking crate to simplify
+#[derive(Clone)]
 pub struct OutboundAllowedHosts {
     allowed_hosts_future: SharedFutureResult<AllowedHostsConfig>,
 }
 
 impl OutboundAllowedHosts {
+    pub async fn resolve(&self) -> anyhow::Result<Arc<AllowedHostsConfig>> {
+        self.allowed_hosts_future.clone().await.map_err(|err| {
+            // TODO: better way to handle this?
```
rylev, line 168: I know this was just moved, but why doesn't `?` work?

crates/trigger-http2/src/outbound_http.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl OutboundHttpInterceptor {
+    pub fn new(server: Arc<HttpServer>, origin: SelfRequestOrigin) -> Self {
+        Self { server, origin }
+    }
+}
+
+const CHAINED_CLIENT_ADDR: SocketAddr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 0);
+
+impl spin_factor_outbound_http::OutboundHttpInterceptor for OutboundHttpInterceptor {
```
rylev, line 26: I'm surprised to see this remain in the trigger. I had been thinking that factors would allow us to move this logic (which doesn't seem to be trigger specific) closer to the outbound-http handler code.

rylev (changes requested):
Mostly looks pretty good. The only thing I'd like to figure out before we merge is the question of where the request interception impl should live. I was surprised to see it in the trigger crate when the logic appears to not be trigger dependent.

crates/trigger-http2/src/outbound_http.rs
rylev: Looking more deeply into this, I think I understand why at least *some* of this has to be in the trigger. The interceptor is meant as a way for the trigger to inject custom logic on each request. In this case, essentially the HTTP trigger's interceptor is responsible for the following things:
1. Establishing the `SelfRequestOrigin` so that self requests can be routed properly.
  1a. The factor has an implicit assumption that somewhere someone is adding a `SelfRequestOrigin` to the request extensions. If no one does, requests to `self` will fail. 
2. Figuring out whether the request is a a service chained request. If it is a service chaining request, then routing that back through the HTTP trigger. 

Both seem to be very heavily tied to the specific HTTP trigger though I'm realizing now, I'm not sure we ever really rationalized these features outside of the HTTP trigger. What does a self request or a service chained request mean when the app is invoked outside of the HTTP trigger? 

As the code is currently written, self requests seem *indirectly* to not be a feature outside of the HTTP trigger. If the factor is invoked outside of an HTTP trigger and no origin is set on the request extensions, then the outgoing request will fail. If a trigger wants to opt into supporting self requests, they have to do so in a very indirect way (by setting the `SelfRequestOrigin` on the request extensions) - this is certainly less than ideal... The outbound-http factor code is very much aware of this feature, but exposes it to embedders in a very indirect way. 

Service chaining, on the other hand, seems to be explicitly an HTTP trigger only feature. If "spin.internal" is used in an outbound request in any other trigger, that request will be routed to "spin.internal" as if it were any other DNS based authority (I guess let's hope that the user is running Spin on a network where `spin.internal` doesn't route to anything malicious...). The outbound-http factor knows nothing about this feature, and it is implemented purely within the HTTP trigger.

lann: I explored a couple of different approaches here but decided to follow the existing semantics for pretty much the reasons you mention:

- "origin" is inherently somewhat trigger-specific; we _could_ have a universal e.g. runtime config option to set/override it for any trigger type, but that seemed out of scope for a refactor

- service chaining could be implemented for other trigger types but would require a large chunk of the http trigger be included in the outbound http factor which again seems out of scope

crates/factor-outbound-http/src/lib.rs
lann: Pretty sure. Since this is used with dependency inversion the end developer could be surprised when composing two factors that call this if one of them silently loses its interceptor.

The real solution would be to allow multiple interceptors but that is a little complicated in rust so I punted.

crates/factor-outbound-http/src/lib.rs
lann: The enum result type is probably a good idea regardless, but I personally think the extension is fine here 🤷

crates/factor-outbound-http/src/wasi.rs
lann: Well, `host` could also be present-but-empty I think? Or do your recent changes make that impossible?

crates/trigger-http2/src/outbound_http.rs
rylev: All of that sounds good to me - I have another comment about how I don't love how we're feeding the origin information back into the factor, but that's an implementation detail we can change after we merge this PR if we want.

crates/factor-outbound-http/src/wasi.rs
rylev: I'm actually not really sure. I *think* from a logic point of view this is probably fine. Then again, just because something works doesn't mean it's the right implementation. I personally would likely feel better with a more verbose and explicit implementation, but I won't die on this hill 😄.

crates/factor-outbound-http/src/lib.rs
rylev: It's a balance - the reason I'm not a fan of extension approach is that it's a very roundabout public API. In order to unlock support for self, users of this factor need to know that they have to somehow add a `SelfRequestOrigin` to the request extensions. It works, and we certainly can make sure this documented somewhere, but it's not a particularly ergonomic or intuitive API. 

I get that this library is unlikely to be used by a large number of people, so if we can't be super ergonomic/easy to understand, that's not the end of the world, but for my future self's sake, I'd love for us to try to make this more straight forward.

lann: Added return enum `InterceptOutcome`

crates/factor-outbound-http/tests/factor_test.rs
lann: The `Ok` side doesn't impl `Debug`.

crates/factor-outbound-networking/src/lib.rs
lann: `err` is an `Arc<Error>` here

lann:
Addressed some of the feedback and added a couple of basic smoke tests.

rylev (approved):

crates/runtime-config/src/lib.rs
lann, line 1: @rylev added this

PR #2705, opened 2024-08-05
connect to sqlite remote db before each query

karthik2804:
This PR attempts to fix the timeout issues occur when connecting to a `libsql` remote database due to the stream expiring. 

I am not sure if this is the right approach. I based it on https://github.com/tursodatabase/libsql/issues/985

ref: https://discord.com/channels/926888690310053918/1269064908645859330/1269376861671002155

### DIFF
--- a/crates/sqlite-libsql/src/lib.rs
+++ b/crates/sqlite-libsql/src/lib.rs
@@ -1,16 +1,15 @@
 use spin_world::v2::sqlite::{self, RowResult};
 use tracing::{instrument, Level};
 
-#[derive(Clone)]
 pub struct LibsqlClient {
-    inner: libsql::Connection,
+    inner: libsql::Database,
 }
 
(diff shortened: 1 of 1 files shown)


karthik2804:
@lann I believe it is created only once and not on every request.

karthik2804:
Since the fix for this should land in the `factors`, we can leave this as a draft unless we are doing a `v2.8` at which point we can take a look at this fix.

rylev (approved):
A better approach would be to ensure that new connections are created each time `Connection::open` is called from the guest, but considering this is fixed in Factors, I think this is a fine approach should we need to land a fix before Factors lands.

rylev:
@karthik2804 I believe this can be closed now that factors has landed, right?

karthik2804:
Yes @rylev this can be now closed!

PR #2715, opened 2024-08-14
[Factors] Key-Value in Trigger2

rylev:
*Builds on #2711 so marking as draft until that merges*

This moves runtime config handling for Spin CLI to its own crate.


crates/runtime-config/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+        key_value::StoreConfig {
+            type_: spin_factor_key_value_spin::SpinKeyValueStore::RUNTIME_CONFIG_TYPE.to_owned(),
+            config: toml::toml! {
+                path = DEFAULT_SPIN_STORE_FILENAME
+            },
+        },
+    );
+
+    key_value
+}
```
lann, line 127: Of course it doesn't need to be done in this PR but we'll almost certainly break this out into an e.g. `key_value` module with the next factor added, yeah?

crates/runtime-config/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+const DEFAULT_SPIN_STORE_FILENAME: &str = "sqlite_key_value.db";
+
+/// The key-value runtime configuration resolver used by the trigger.
+///
+/// Takes a base path for the local store.
+pub fn key_value_resolver(local_store_base_path: PathBuf) -> key_value::RuntimeConfigResolver {
+    let mut key_value = key_value::RuntimeConfigResolver::new();
+
+    // Register the supported store types.
+    // Unwraps are safe because the store types are known to not overlap.
```
lann, line 102: micro-nit: avoid the "s" word...

```suggestion
    // Unwraps are fine because the store types are known to not overlap.
```

lann (approved):

crates/runtime-config/src/lib.rs
rylev: Yea most likely.

crates/runtime-config/src/lib.rs
rylev: 👍 I'll change this in the next PR to avoid having to run CI again.

PR #2716, opened 2024-08-15
Set initial key-values

rylev:
*Depends on https://github.com/fermyon/spin/pull/2715*

Shows how we can handle initial configuration of factors by handling the initial key-value pairs for the default key-value store.

### DIFF
--- a/crates/trigger2/src/cli.rs
+++ b/crates/trigger2/src/cli.rs
@@ -11,7 +11,7 @@ use spin_common::{arg_parser::parse_kv, sloth};
 use spin_factors_executor::{ComponentLoader, FactorsExecutor};
 use spin_runtime_config::ResolvedRuntimeConfig;
 
-use crate::factors::TriggerFactors;
+use crate::factors::{TriggerFactors, TriggerFactorsRuntimeConfig};
 use crate::stdio::{FollowComponents, StdioLoggingExecutorHooks};
 use crate::Trigger;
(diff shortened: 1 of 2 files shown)


crates/trigger2/src/cli.rs
```diff
@@ -194,11 +194,18 @@ impl<T: Trigger> FactorsTriggerCommand<T> {
...
+                    runtime_config_path,
+                    self.state_dir.as_deref(),
+                )?
             }
             None => ResolvedRuntimeConfig::default(),
         };
 
+        runtime_config
+            .set_initial_key_values(&self.key_values)
+            .await?;
```
lann, line 207: May be out of scope for this PR, but I think we should probably support something like:

```toml
[key_value_store.default]
type = "none"
```

...which would then suggest doing something like:

```suggestion
        if !self.key_values.is_empty() {
            runtime_config
                .set_initial_key_values(&self.key_values)
                .await?;
        }
```

lann (approved):

crates/trigger2/src/cli.rs
rylev: Is this point of such a feature to ensure that a default store never gets created? If the `key-value` interface gets used with the default store is it expected to just error?

lann: Yep. Could be helpful if you know your target environment won't have a KV store.

PR #2718, opened 2024-08-15
remove *.md docs from gh build workflow

michelleN:
removes all recently added governance related docs from build as well.


radu-matei (approved):

radu-matei:
Kicked off a re-run of the integration test action, which is mysteriously failing. Given it's definitely not related to the changes here, we can merge this whenever I think.

rylev (approved):

PR #2722, opened 2024-08-19
Add outbound MQTT factor

karthik2804:
This PR adds the outbound-mqtt factor along with the associated tests.

### DIFF
--- a/crates/factor-outbound-mqtt/Cargo.toml
+++ b/crates/factor-outbound-mqtt/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "spin-factor-outbound-mqtt"
+version = { workspace = true }
+authors = { workspace = true }
+edition = { workspace = true }
+
+[dependencies]
+anyhow = "1.0"
+rumqttc = { version = "0.24", features = ["url"] }
+spin-factor-outbound-networking = { path = "../factor-outbound-networking" }
+spin-factors = { path = "../factors" }
+spin-core = { path = "../core" }
+spin-world = { path = "../world" }
+tracing = { workspace = true }
+table = { path = "../table" }
+tokio = { version = "1.0", features = ["sync"] }
+
+[dev-dependencies]
+spin-factor-variables = { path = "../factor-variables" }
+spin-factors-test = { path = "../factors-test" }
+tokio = { version = "1", features = ["macros", "rt"] }
+
+[lints]
+workspace = true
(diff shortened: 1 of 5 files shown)


crates/factor-outbound-mqtt/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition = { workspace = true }
+
+[dependencies]
+anyhow = "1.0"
+rumqttc = { version = "0.24", features = ["url"] }
+spin-factor-outbound-networking = { path = "../factor-outbound-networking" }
+spin-factors = { path = "../factors" }
+spin-core = { path = "../core" }
+spin-world = { path = "../world" }
+tracing = { workspace = true }
```
rylev, line 14: nit: all the dependencies are in alphabetical order except for this one 🚨

crates/factor-outbound-mqtt/src/host.rs
```diff
@@ -0,0 +1,131 @@
+use std::{sync::Arc, time::Duration};
+
+use anyhow::Result;
+use spin_core::{async_trait, wasmtime::component::Resource};
+use spin_factor_outbound_networking::OutboundAllowedHosts;
+use spin_world::v2::mqtt::{self as v2, Connection, Error, Qos};
+use tracing::{instrument, Level};
+
+pub type CreateClient = Arc<
```
rylev, line 9: Looking at the other factors, we've normally been handling this pattern with custom trait trait objects rather than closure trait objects. So instead of this type alias we would do something like:


```rust
trait ClientCreator {
  fn create(
    address: String,
    username: String,
    password: String,
    keep_alive_interval: Duration,
  )
}
// ...

impl InstanceState {
    pub fn new(allowed_hosts: OutboundAllowedHosts, create_client: Box<dyn ClientCreator>) { 
// ...
```

crates/factor-outbound-mqtt/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+        Ok(InstanceState::new(
+            allowed_hosts,
+            self.create_client.clone(),
+        ))
+    }
+}
+
+impl SelfInstanceBuilder for InstanceState {}
+
+pub struct NetworkedMqttClient {
```
rylev, line 67: Let's document that this is as a concrete implementation of `MqttClient`

rylev (approved):
I think this is a good start. I won't block on the change from the closure trait object to using a custom trait instead, but I do think it's something we probably should do eventually.

crates/factor-outbound-mqtt/Cargo.toml
karthik2804: fixed

lann (approved):

lann:
I rebased...

PR #2723, opened 2024-08-19
[Factors] Change how `incoming-request.authority` is set.

rylev:
The moral equivalent of https://github.com/fermyon/spin/pull/2684 for factors.

The second commit fixes a remaining issue: `incoming-request.authority` won't otherwise be set in service chained requests. This is because authority was set a layer above where service chained requests hook into the http trigger stack. The setting of `authority` has been moved into `HttpServer::handle_trigger_route` from the `HttpServer::handle` method. I believe this will work fine as the only thing the `handle` function does in addition to calling `handle_trigger_route` is handling the well known routes, 404s, and `extract_trace_context` all of which shouldn't need the scheme and authority.

### DIFF
--- a/crates/trigger-http2/src/lib.rs
+++ b/crates/trigger-http2/src/lib.rs
@@ -67,7 +67,8 @@ pub(crate) type InstanceState = ();
 
 /// The Spin HTTP trigger.
 pub struct HttpTrigger {
-    listen_addr: SocketAddr,
+    /// The address the server will listen on.
+    addr_to_bind: SocketAddr,
     tls_config: Option<TlsConfig>,
     router: Router,
     // Component ID -> component trigger config
@@ -112,7 +113,7 @@ impl Trigger for HttpTrigger {
         );
 
         Ok(Self {
-            listen_addr: cli_args.address,
+            addr_to_bind: cli_args.address,
             tls_config: cli_args.into_tls_config(),
             router,
             component_trigger_configs,
@@ -121,7 +122,7 @@ impl Trigger for HttpTrigger {
 
     async fn run(self, trigger_app: TriggerApp) -> anyhow::Result<()> {
         let Self {
-            listen_addr,
+            addr_to_bind: listen_addr,
             tls_config,
             router,
             component_trigger_configs,
@@ -131,6 +132,12 @@ impl Trigger for HttpTrigger {
             .await
             .with_context(|| format!("Unable to listen on {listen_addr}"))?;
 
+        // Get the address the server is actually listening on
+        // We can't use `self.listen_addr` because it might not
+        // be fully resolved (e.g, port 0).
+        let listen_addr = listener
+            .local_addr()
+            .context("failed to retrieve address server is listening on")?;
         let server = Arc::new(HttpServer::new(
             listen_addr,
             trigger_app,
--- a/crates/trigger-http2/src/lib.rs
+++ b/crates/trigger-http2/src/lib.rs
@@ -68,7 +68,10 @@ pub(crate) type InstanceState = ();
 /// The Spin HTTP trigger.
 pub struct HttpTrigger {
     /// The address the server will listen on.
-    addr_to_bind: SocketAddr,
+    ///
+    /// Note that this might not be the actual socket address that ends up being bound to.
+    /// If the port is set to 0, the actual address will be determined by the OS.
+    listen_addr: SocketAddr,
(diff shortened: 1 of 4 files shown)


crates/trigger-http2/src/lib.rs
```diff
@@ -67,7 +67,8 @@ pub(crate) type InstanceState = ();
 
 /// The Spin HTTP trigger.
 pub struct HttpTrigger {
-    listen_addr: SocketAddr,
+    /// The address the server will listen on.
+    addr_to_bind: SocketAddr,
```
lann, line 71: I wouldn't change this; the CLI flag is `--listen`...

rylev: I hear ya. I wanted to draw attention to the fact that this might not be the actual `SocketAddr` the socket is listening on (in the case of a 0 port), but perhaps that's not worth the potential confusion in the other direction.

crates/trigger-http2/src/server.rs
```diff
@@ -88,28 +90,32 @@ impl HttpServer {
...
+    /// This method handles well known paths and routes requests to the handler when the router
+    /// matches the requests path.
     async fn handle(
         self: &Arc<Self>,
         mut req: Request<Body>,
-        scheme: Scheme,
+        server_scheme: Scheme,
         client_addr: SocketAddr,
     ) -> anyhow::Result<Response<Body>> {
-        set_req_uri(&mut req, scheme, self.listen_addr)?;
```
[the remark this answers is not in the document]
rylev: Putting it in `instrumented_service_fn` moves the logic up the call stack where we need to move it down the call stack. 

The call graph is: `instrumented_service_fn` -> `handle` -> `handle_trigger_route`. The handler for service chained calls calls `handle_trigger_route` which is why we moved `set_req_uri` to that function - the service chained request also needs its uri manipulated. If we were to move that logic up to `instrumented_service_fn`, the issue this change is trying to address would persist.

crates/trigger-http2/src/lib.rs
rylev: By the way, the CLI arg is actually `--address`... But I'll still change it back.

lann: Unfortunately it is a trick:

```rust
#[clap(long = "listen", ...)]
pub address: SocketAddr,
```

crates/trigger-http2/src/server.rs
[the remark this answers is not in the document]
lann: Ah gotcha. Perhaps it would make sense to split the scheme and host updates? :thinking:

crates/trigger-http2/src/outbound_http.rs
```diff
@@ -42,7 +43,7 @@ impl spin_factor_outbound_http::OutboundHttpInterceptor for OutboundHttpIntercep
             let server = self.server.clone();
             let resp_fut = async move {
                 match server
-                    .handle_trigger_route(req, route_match, CHAINED_CLIENT_ADDR)
+                    .handle_trigger_route(req, route_match, Scheme::HTTP, CHAINED_CLIENT_ADDR)
                     .await
                 {
                     Ok(resp) => Ok(Ok(IncomingResponse {
```
lann, line 49: I wonder if we should be setting the `SelfRequestOrigin` unconditionally in this interceptor. It seems like `self` requests ought to work in service-chained handlers? :thinking:

crates/trigger-http2/src/server.rs
```diff
@@ -322,11 +335,30 @@ impl HttpServer {
...
+        .context("'Host' header contains an invalid authority")?;
+    // Ensure that if `req.authority` is set, it matches what was in the `Host` header
+    // https://github.com/hyperium/hyper/issues/1612
+    if let Some(a) = parts.authority.as_ref() {
+        if a != &authority {
+            return Err(anyhow::anyhow!(
+                "authority in 'Host' header does not match authority in URI"
+            ));
+        }
+    }
```
lann, line 361: Sorry, I wasn't very clear. I believe it is valid to have just the `host` header, just the uri `authority`, or both set, where in the "both" case they must match.

Egads this seems complicated: https://www.rfc-editor.org/rfc/rfc9113.html#section-8.3.1-2.3.3

lann (approved):
Given that the HTTP trigger doesn't currently support http/2 I think the current impl is OK to merge.

PR #2724, opened 2024-08-19
[Factors] Azure Key Vault Variables

rylev:
Moves over https://github.com/fermyon/spin/pull/2678 to Factors (and does some clean up on the way. 

cc @devigned


crates/factor-variables/src/spin_cli/azure_key_vault.rs
```diff
@@ -0,0 +1,157 @@
...
+pub struct AzureKeyVaultVariablesConfig {
+    pub vault_url: String,
+    pub client_id: Option<String>,
+    pub client_secret: Option<String>,
+    pub tenant_id: Option<String>,
+    pub authority_host: Option<AzureAuthorityHost>,
+}
+
+#[derive(Debug, Copy, Clone, Deserialize, Default)]
+pub enum AzureAuthorityHost {
```
lann, line 27: This is probably out of scope for _this_ PR but this seems like an odd enum to maintain here...

crates/factor-variables/src/spin_cli/azure_key_vault.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl From<AzureAuthorityHost> for Url {
+    fn from(value: AzureAuthorityHost) -> Self {
+        let url = match value {
+            AzureAuthorityHost::AzureChina => "https://login.chinacloudapi.cn/",
+            AzureAuthorityHost::AzureGovernment => "https://login.microsoftonline.us/",
+            AzureAuthorityHost::AzureGermany => "https://login.microsoftonline.de/",
+            AzureAuthorityHost::AzurePublicCloud => "https://login.microsoftonline.com/",
```
lann, line 153: https://docs.rs/azure_core/0.20.0/azure_core/authority_hosts/index.html :neutral_face:

lann (approved):
Just a couple of notes about the existing code :shrug:

PR #2725, opened 2024-08-19
factors: Get `spin up` working

lann:
This is a minimal patch to get `spin up` working:

- Comment out references to the not-yet-ported redis trigger
- Replace `TodoComponentLoader` with `SimpleComponentLoader`; some functionality still missing


rylev (approved):
🎉

PR #2727, opened 2024-08-19
Land the outbound-mysql factor

calebschoepp:


calebschoepp:
Tagging @rylev and/or @lann for review

crates/factor-outbound-mysql/src/client.rs
```diff
@@ -0,0 +1,308 @@
...
+        (ColumnType::MYSQL_TYPE_STRING, true) => DbDataType::Binary, // BINARY type
+        (ColumnType::MYSQL_TYPE_STRING, _) => DbDataType::Str,
+        (ColumnType::MYSQL_TYPE_VAR_STRING, true) => DbDataType::Binary, // VARBINARY type
+        (ColumnType::MYSQL_TYPE_VAR_STRING, _) => DbDataType::Str,
+        (_, _) => DbDataType::Other,
+    }
+}
+
+fn convert_numeric_type(column: &mysql_async::Column) -> DbDataType {
+    match (column.column_type(), is_signed(column)) {
```
lann, line 151: This is somewhat controversial but I think tiny single-call-site functions should (usually) be inline.

If I need to understand exactly how this code works then I have to scroll down to read the function body, and if I don't need to understand then its just added noise. :shrug: 

```suggestion
    let is_signed = ...;
    match (column.column_type(), is_signed) {
```

crates/factor-outbound-mysql/src/client.rs
```diff
@@ -0,0 +1,308 @@
...
+use mysql_async::{from_value_opt, Conn as MysqlClient, Opts, OptsBuilder, SslOpts};
+use spin_core::async_trait;
+use spin_world::v2::mysql::{self as v2};
+use spin_world::v2::rdbms_types::{
+    self as v2_types, Column, DbDataType, DbValue, ParameterValue, RowSet,
+};
+use url::Url;
+
+#[async_trait]
+pub trait Client {
```
lann, line 15: For these sort of single-consumer traits you can save yourself some type noise by adding "supertrait" constraints:

```suggestion
pub trait Client: Send + Sync + 'static {
```

lann (approved):
Nothing too important

PR #2729, opened 2024-08-20
Allow multiple commands in a component build section

itowlson:
So we don't need to resort to `&&` kludges which thwart our Windows users.

(This inspired by e.g. the Leptos template and the Moonbit SDK.)

Syntax:

```
# multiple
[component.motivational.build]
command = [
    "cargo build --target wasm32-wasi --release",
    "echo HUZZAH"
]

# existing syntax still works
[component.sensible.build]
```

### DIFF
--- a/crates/manifest/src/schema/common.rs
+++ b/crates/manifest/src/schema/common.rs
@@ -83,13 +83,34 @@ pub enum WasiFilesMount {
 #[serde(deny_unknown_fields)]
 pub struct ComponentBuildConfig {
     /// `command = "cargo build"`
-    pub command: String,
+    pub command: Commands,
     /// `workdir = "components/main"
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub workdir: Option<String>,
     /// watch = ["src/**/*.rs"]
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub watch: Vec<String>,
+}
+
+impl ComponentBuildConfig {
+    /// The commands to execute for the build
+    pub fn commands(&self) -> impl Iterator<Item = &String> {
+        let as_vec = match &self.command {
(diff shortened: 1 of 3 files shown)


radu-matei (approved):
Thank you for this!


LGTM

crates/manifest/src/schema/common.rs
```diff
@@ -92,6 +92,27 @@ pub struct ComponentBuildConfig {
...
 }
 
+impl ComponentBuildConfig {
+    /// The commands to execute for the build
+    pub fn commands(&self) -> impl Iterator<Item = &String> {
+        let as_vec = match &self.command {
+            Commands::Single(cmd) => vec![cmd],
+            Commands::Multiple(cmds) => cmds.iter().collect(), // I know, but I haven't been able to find a less worse way to line up the types between the two arms
+        };
+        as_vec.into_iter()
```
rylev, line 102: I don't think the comment adds anything here. I think we should delete it. Also, you can just clone instead of iter and collecting. 

```suggestion
        let as_vec = match &self.command {
            Commands::Single(cmd) => vec![cmd],
            Commands::Multiple(cmds) => cmds.clone(),
        };
        as_vec.into_iter()
```

rylev (approved):

crates/manifest/src/schema/common.rs
itowlson: Cloning doesn't typecheck (`Vec<String>` vs `Vec<&String>`).  I can clone `cmd` in the single case too, if you like?

itowlson: @rylev Actually, since you kindly approved despite this, I'm going to merge with the code as is (I've deleted the comment as you suggested) and we can simplify it in a later PR - hope that's okay.

PR #2732, opened 2024-08-20
[Factors] in-process runtime and conformance tests

rylev:
This allows us to run runtime and conformance tests using the "in-process runtime" (as opposed to the Spin CLI spawned process runtime). 

This requires a few refactorings that I think push us more towards other goals as well (namely spin-trigger being generic over the `RuntimeFactors` type). 

The actual code for instantiating and triggering an HTTP trigger without running an actual HTTP server looks like this:

```rust
    let app = spin_app::App::new("my-app", locked_app);
    let trigger = HttpTrigger::new(&app, "127.0.0.1:80".parse().unwrap(), None)?;
    let mut builder = TriggerAppBuilder::new(trigger, PathBuf::from("."));
    let trigger_app = builder.build(app, TriggerAppOptions::default()).await?;
    let server = builder.trigger.into_server(trigger_app)?;
    server
        .handle(
              req,
              http::uri::Scheme::HTTP,
              (std::net::Ipv4Addr::LOCALHOST, 7000).into(),
         )
        .await?;
```


crates/trigger-http2/src/server.rs
```diff
@@ -333,6 +288,111 @@ impl HttpServer {
     }
 }
 
+/// Handles a routed HTTP trigger request.
+pub struct RequestHandler {
```
lann, line 292: Could you say more about this change? It isn't clear how it relates to the conformance tests.

rylev: To be fair this was from an evolutionary dead end. I do think it helps make things a bit easier to follow (at the expense of being slightly more verbose) by only giving the `OutboundHttpInterceptor` what it needs and nothing more. It encapsulates just the component handling and leaves out all the other HTTP sever stuff to the encapsulating `HttpServer` types. 

I can revert it if you want me to though.

lann: Its mostly just really hard to review the diff :sweat_smile:

rylev: Sorry! If you aren't able to review, I'll revert that change tomorrow.

rylev: I've reverted that change so hopefully the diff is a bit easier to read. Unfortunately, there's one spot where git's diffing still gets a bit confused.

lann (approved):

crates/trigger-http2/src/server.rs
lann: Thanks!

PR #2736, opened 2024-08-21
Use `spin.alt` domain for service chaining

### LINKED ISSUE #2731: Consider switching service chaining domain from `spin.internal` to `spin.alt`
Looking at https://www.rfc-editor.org/rfc/rfc9476.html I think `.alt` might be more appropriate:

- `.alt` ["should not be resolved using the DNS protocol"](https://www.rfc-editor.org/rfc/rfc9476.html#section-1-1)
- `.internal` is to be used for internal networks _with_ DNS

itowlson:
Fixes #2731

### DIFF
--- a/crates/locked-app/src/locked.rs
+++ b/crates/locked-app/src/locked.rs
@@ -15,7 +15,8 @@ use crate::{
 pub type LockedMap<T> = std::collections::BTreeMap<String, T>;
 
 /// If present and required in `host_requirements`, the host must support
-/// local service chaining (*.spin.internal) or reject the app.
+/// local service chaining (via *.spin.alt, and, for backward compatibility,
+/// *.spin.internal) or reject the app.
 pub const SERVICE_CHAINING_KEY: &str = "local_service_chaining";
 
 /// Indicates that a host feature is optional. This is the default and is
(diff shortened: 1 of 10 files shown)


crates/locked-app/src/locked.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
 pub type LockedMap<T> = std::collections::BTreeMap<String, T>;
 
 /// If present and required in `host_requirements`, the host must support
-/// local service chaining (*.spin.internal) or reject the app.
+/// local service chaining (*.spin.alt/.internal) or reject the app.
```
radu-matei, line 18: nit: Only reading the doc comment it's somewhat unclear whether both are supported.

rylev:
Just going to note that we'll want to bring this change into the Factors work as well. We're close to merging that into main which will get rid of the possibility of things landing in main but not in factors.

crates/locked-app/src/locked.rs
itowlson: Ugh, that also makes me think that this becomes _another_ compatibility flag because an app that uses `spin.alt` won't work in an environment that supports only `spin.internal`.

lann: Given that we're doing this with a major release I think it should be fine to just note how it changed for 3.0.

itowlson: That rather flies in the face of being able to validate that an application runs in _other_ environments though, which was the whole point of adding the host requirements to the lockfile in the first place.

lann: I suppose the principled thing would be to add a e.g. `local_service_chaining_alt` host requirement.

PR #2737, opened 2024-08-21
factors: Update outbound networking

lann:
Fell down this rabbit hole while implementing OutboundHttpFactor's fermyon:spin/http interface:

- Re-add instrumentation to wasi outbound http impl
- Introduce OutboundNetworkingFactor "disallowed host callback" to handle Spin-CLI-specific terminal output
  - Rearrange a bunch of allowed outbound host checking code to dedupe this callback logic
- Refactor `SelfRequestOrigin` to be a more straightforward InstanceState setter; the extension method didn't work well for the spin interface anyway

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -106,6 +106,7 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
                     request,
                     config,
                     self.state.allowed_hosts.clone(),
+                    self.state.self_request_origin.clone(),
                     tls_client_config,
                 )
                 .in_current_span(),
@@ -118,6 +119,7 @@ async fn send_request_impl(
     mut request: Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
     mut config: wasmtime_wasi_http::types::OutgoingRequestConfig,
     outbound_allowed_hosts: OutboundAllowedHosts,
+    self_request_origin: Option<SelfRequestOrigin>,
     tls_client_config: Arc<ClientConfig>,
 ) -> anyhow::Result<Result<IncomingResponse, ErrorCode>> {
     if request.uri().authority().is_some() {
@@ -137,10 +139,7 @@ async fn send_request_impl(
             return Ok(Err(ErrorCode::HttpRequestDenied));
         }
 
-        let origin = request
-            .extensions()
-            .get::<SelfRequestOrigin>()
-            .cloned()
+        let origin = self_request_origin
             .context("cannot send relative outbound request; no 'origin' set by host")?;
 
         config.use_tls = origin.use_tls();
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -1,13 +1,13 @@
 use std::{error::Error, sync::Arc};
 
 use anyhow::Context;
-use http::{header::HOST, uri::Authority, Request, Uri};
+use http::{header::HOST, Request};
 use http_body_util::BodyExt;
 use rustls::ClientConfig;
-use spin_factor_outbound_networking::{OutboundAllowedHosts, OutboundUrl};
+use spin_factor_outbound_networking::OutboundAllowedHosts;
 use spin_factors::{wasmtime::component::ResourceTable, RuntimeFactorsInstanceState};
 use tokio::{net::TcpStream, time::timeout};
-use tracing::Instrument;
+use tracing::{field::Empty, instrument, Instrument};
 use wasmtime_wasi_http::{
     bindings::http::types::ErrorCode,
     body::HyperOutgoingBody,
@@ -68,6 +68,19 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
         self.table
     }
 
+    #[instrument(
+        name = "spin_outbound_http.send_request",
+        skip_all,
+        fields(
+            otel.kind = "client",
+            url.full = %request.uri(),
+            http.request.method = %request.method(),
+            otel.name = %request.method(),
+            http.response.status_code = Empty,
+            server.address = Empty,
+            server.port = Empty,
+        ),
+    )]
     fn send_request(
         &mut self,
         mut request: Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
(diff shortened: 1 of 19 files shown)


crates/outbound-networking/src/lib.rs
```diff
@@ -8,31 +8,6 @@ pub const ALLOWED_HOSTS_KEY: MetadataKey<Vec<String>> = MetadataKey::new("allowe
...
-        let msg = if let Some(port) = port {
-            format!("`allowed_outbound_hosts = [\"{scheme}://{host}:{port}\"]`")
-        } else {
-            format!("`allowed_outbound_hosts = [\"{scheme}://{host}:$PORT\"]` (where $PORT is the correct port number)")
-        };
-        eprintln!("To allow requests, add {msg} to the manifest component section.");
-    }
-    is_allowed
-}
-
```
lann, line 35: Moved to `OutboundAllowedHosts::check_url` above

crates/outbound-networking/src/lib.rs
```diff
@@ -8,31 +8,6 @@ pub const ALLOWED_HOSTS_KEY: MetadataKey<Vec<String>> = MetadataKey::new("allowe
...
-    if !is_allowed {
-        terminal::warn!("A component tried to make a request to non-allowed url '{url}'.");
-        let (scheme, host, port) = (url.scheme, url.host, url.port);
-        let msg = if let Some(port) = port {
-            format!("`allowed_outbound_hosts = [\"{scheme}://{host}:{port}\"]`")
-        } else {
-            format!("`allowed_outbound_hosts = [\"{scheme}://{host}:$PORT\"]` (where $PORT is the correct port number)")
-        };
-        eprintln!("To allow requests, add {msg} to the manifest component section.");
```
lann, line 31: This language is merged with the outbound HTTP warning language in `disallowed_host_callback` below. This is probably a bit of a regression around the "where $PORT is the correct port number" stuff, but I feel like we could probably do better there anyway.

crates/trigger-http2/src/outbound_http.rs
```diff
@@ -58,7 +57,6 @@ impl spin_factor_outbound_http::OutboundHttpInterceptor for OutboundHttpIntercep
             let resp = HostFutureIncomingResponse::pending(wasmtime_wasi::runtime::spawn(resp_fut));
             InterceptOutcome::Complete(Ok(resp))
         } else {
-            request.extensions_mut().insert(self.origin.clone());
```
lann, line 61: @rylev It turns out this was a bad idea. WHO COULD HAVE GUESSED

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -104,21 +118,28 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
...
-    let is_relative_url = request.uri().authority().is_none();
-    if is_relative_url {
+    if request.uri().authority().is_some() {
+        // Absolute URI
+        let is_allowed = outbound_allowed_hosts
+            .check_url(&request.uri().to_string(), "https")
+            .await
+            .map_err(|_| ErrorCode::HttpRequestUriInvalid)?;
+        if !is_allowed {
+            return Ok(Err(ErrorCode::HttpRequestDenied));
```
rylev, line 132: ~I think we're missing a call to `outbound_allowed_hosts.report_disallowed_host` here, no?~

I see later on that `report_disallowed_host` gets called automatically in `check_url`. Doing it in the `check_url` call but not in `allows_relative_url` is confusing. I think we should encapsulate the call to `report_disallowed_host` in the same equivalent places for both relative and non-relative requests.

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -104,21 +118,28 @@ impl<'a> WasiHttpView for WasiHttpImplInner<'a> {
...
         if !allowed_hosts.allows_relative_url(&["http", "https"]) {
-            return Ok(handle_not_allowed(request.uri(), true));
+            outbound_allowed_hosts.report_disallowed_host("http", "self");
+            return Ok(Err(ErrorCode::HttpRequestDenied));
         }
 
-        let origin = request
-            .extensions()
-            .get::<SelfRequestOrigin>()
-            .cloned()
```
rylev, line 121: 😍

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -127,12 +148,6 @@ async fn send_request_impl(
...
         *request.uri_mut() = origin.into_uri(path_and_query);
-    } else {
-        let outbound_url = OutboundUrl::parse(request.uri().to_string(), "https")
-            .map_err(|_| ErrorCode::HttpRequestUriInvalid)?;
-        if !allowed_hosts.allows(&outbound_url) {
-            return Ok(handle_not_allowed(request.uri(), false));
-        }
     }
 
     if let Some(authority) = request.uri().authority() {
```
rylev, line 153: Nit: I know this was here before, but it seems strange we have two different blocks for when `request.uri().authority()` is `Some` instead of just combining them.

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -22,7 +22,24 @@ pub use runtime_config::ComponentTlsConfigs;
 
 pub type SharedFutureResult<T> = Shared<BoxFuture<'static, Result<Arc<T>, Arc<anyhow::Error>>>>;
 
-pub struct OutboundNetworkingFactor;
+#[derive(Default)]
+pub struct OutboundNetworkingFactor {
+    disallowed_host_callback: Option<DisallowedHostCallback>,
+}
+
+pub type DisallowedHostCallback = fn(scheme: &str, authority: &str);
```
rylev, line 30: I know it's more typing, but should we only limit this to function pointers instead of any `impl Fn(&str, &str)`?

rylev (approved):

crates/factor-outbound-http/src/wasi.rs
lann: Hmm...authority _should_ always be set here. I'll change it to an error.

crates/factor-outbound-networking/src/lib.rs
lann: Ugh but its _so much more typing_ :neutral_face:

crates/factor-outbound-http/src/wasi.rs
lann: I reworked this significantly; see the latest commit.

PR #2738, opened 2024-08-21
Switch back to the old sync version of HostOutputStream.

rylev:
Previously only one write would succeed before something locked up. We're still investigating what exactly the issue is, but this goes back to the implementation currently used in `main`.

### DIFF
--- a/crates/factor-wasi/src/lib.rs
+++ b/crates/factor-wasi/src/lib.rs
@@ -1,18 +1,23 @@
+mod io;
 pub mod spin;
 mod wasi_2023_10_18;
 mod wasi_2023_11_10;
 
-use std::{future::Future, net::SocketAddr, path::Path};
+use std::{
+    future::Future,
+    io::{Read, Write},
+    net::SocketAddr,
+    path::Path,
+};
 
+use io::{PipeReadStream, PipedWriteStream};
 use spin_factors::{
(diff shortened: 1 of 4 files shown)


dicej (approved):
Might want to consider also bringing `PipeStdinStream` back and using it in `stdin_pipe`.

crates/factor-wasi/src/lib.rs
```diff
@@ -190,11 +191,8 @@ impl InstanceBuilder {
     }
 
     /// Sets the WASI `stdout` descriptor to the given [`AsyncWrite`]r.
-    pub fn stdout_pipe(&mut self, w: impl AsyncWrite + Send + Unpin + 'static) {
-        self.stdout(AsyncStdoutStream::new(AsyncWriteStream::new(
-            1024 * 1024,
-            w,
-        )));
+    pub fn stdout_pipe(&mut self, w: impl Write + Send + Sync + Unpin + 'static) {
```
lann, line 194: Maybe rename this to `stdout_sync`? :shrug:

rylev: Does `pipe` refer to the asyncness? I thought `pipe` fits just as much as it did before. I don't have strong feelings though either way.

lann: I don't know. It's all pipes in the end isn't it?

PR #2740, opened 2024-08-21
factors: Implement spin outbound http

lann:
In the interest of avoiding rebases I tacked on a couple of tangential changes:

- A couple of miscellaneous changes to the wasi outbound http impl
- Fixing some integration tests, mostly removing `base`


crates/factor-outbound-http/src/lib.rs
```diff
@@ -81,6 +82,8 @@ pub struct InstanceState {
     component_tls_configs: ComponentTlsConfigs,
     self_request_origin: Option<SelfRequestOrigin>,
     request_interceptor: Option<Box<dyn OutboundHttpInterceptor>>,
+    // Connection-pooling client for 'fermyon:spin/http' interface
```
rylev, line 85: Nit: make this visible to IDEs 

```suggestion
    /// Connection-pooling client for 'fermyon:spin/http' interface
```

rylev (approved):

PR #2742, opened 2024-08-22
[Factors] Integrate llm factor

rylev:
Integrates the LLM factor into trigger2. 

I don't really love how the `use_gpu` configuration item (which for now is hardcoded as true) is being weaved through out the code, but I'd rather go back in a follow up PR to try to clean up how we handle configuration in some holistic way.

### DIFF
--- a/crates/factor-llm/Cargo.toml
+++ b/crates/factor-llm/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "factor-llm"
+name = "spin-factor-llm"
 version.workspace = true
 authors.workspace = true
 edition.workspace = true
@@ -11,10 +11,16 @@ rust-version.workspace = true
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
+serde = "1.0"
 spin-factors = { path = "../factors" }
+spin-llm-local = { path = "../llm-local" }
+spin-llm-remote-http = { path = "../llm-remote-http" }
 spin-locked-app = { path = "../locked-app" }
 spin-world = { path = "../world" }
 tracing = { workspace = true }
+tokio = { version = "1", features = ["sync"] }
+toml = "0.8"
+url = "2"
 
 [dev-dependencies]
 spin-factors-test = { path = "../factors-test" }
--- a/crates/factor-llm/src/lib.rs
+++ b/crates/factor-llm/src/lib.rs
@@ -1,4 +1,5 @@
 mod host;
+pub mod spin;
 
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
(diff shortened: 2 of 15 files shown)


crates/llm-local/src/lib.rs
```diff
@@ -107,18 +106,13 @@ impl LlmEngine for LocalLlmEngine {
...
-        let mut engine = Self {
+    pub fn new(registry: PathBuf, use_gpu: bool) -> Self {
+        Self {
             registry,
             use_gpu,
             inferencing_models: Default::default(),
             embeddings_models: Default::default(),
-        };
-
-        let _ = engine.inferencing_model("llama2-chat".into()).await;
```
rylev, line 118: I had to delete this because initialization is now synchronous  - we prefer to initialize factors as fast as possible. This was trying to bubble up missing model errors to initialization time. They now move to first usage.

lann: If we feel like retaining this validation is worthwhile we could make both `LlmFactor::new` and `TriggerFactors::new` async and fallible and check there.

rylev: Unfortunately, the engine constructed in `LlmFactor::new` but rather in `Factor::configure_app` since the engines are configured by the runtime config. We would have to make `Factor::configure_app` async which I think we're less willing to do?

lann (approved):

crates/factor-llm/src/lib.rs
```diff
@@ -11,26 +12,28 @@ use spin_factors::{
...
 use spin_world::v1::llm::{self as v1};
 use spin_world::v2::llm::{self as v2};
+use tokio::sync::Mutex;
 
 pub const ALLOWED_MODELS_KEY: MetadataKey<Vec<String>> = MetadataKey::new("ai_models");
 
+/// The factor for LLMs.
 pub struct LlmFactor {
-    create_engine: Box<dyn Fn() -> Box<dyn LlmEngine> + Send + Sync>,
+    default_engine_creator: Box<dyn LlmEngineCreator>,
```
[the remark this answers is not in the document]
rylev: Yea, I'd personally like to move away from generics and towards using dynamic dispatch like we do here.

crates/llm-local/src/lib.rs
lann: I suppose we could also do whatever checks we like in `spin-runtime-config`... :thinking:

crates/factor-llm/Cargo.toml
```diff
@@ -11,10 +11,16 @@ rust-version.workspace = true
 [dependencies]
 anyhow = "1.0"
 async-trait = "0.1"
+serde = "1.0"
 spin-factors = { path = "../factors" }
+spin-llm-local = { path = "../llm-local" }
```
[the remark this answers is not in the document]
rylev: I'm going to do some small refactorings in a follow up that will make this easier to bring back.

PR #2744, opened 2024-08-22
factors: Reorganize some http trigger code

lann:
The actual goal was to fix an integration test (which this also does) but it ultimately turned into a reorg:

- Move header tests closer to the code they test
- Move some server init code into `Server::new` to avoid passing interdependent params


crates/trigger-http2/src/server.rs
```diff
@@ -58,9 +58,42 @@ impl HttpServer {
         listen_addr: SocketAddr,
         tls_config: Option<TlsConfig>,
         trigger_app: TriggerApp,
-        router: Router,
-        component_trigger_configs: HashMap<String, HttpTriggerConfig>,
     ) -> anyhow::Result<Self> {
+        // This needs to be a vec before building the router to handle duplicate routes
```
lann, line 62: This was the actual bug fix

crates/trigger-http2/src/server.rs
```diff
@@ -58,9 +58,42 @@ impl HttpServer {
...
+        let component_trigger_configs = Vec::from_iter(
```
lann, line 63: Changing `HashMap` -> `Vec` was the only substantive logic change in this PR.

rylev (approved):

PR #2750, opened 2024-08-22
Factors redis trigger

lann:


crates/trigger-redis/src/lib.rs
```diff
@@ -1,203 +1,226 @@
...
+use futures::{StreamExt, TryFutureExt};
+use redis::{Client, Msg};
+use serde::Deserialize;
+use spin_factor_variables::VariablesFactor;
+use spin_trigger2::{cli::NoCliArgs, App, Trigger, TriggerApp};
+use spin_world::exports::fermyon::spin::inbound_redis;
 use tracing::{instrument, Level};
 
-use crate::spin::SpinRedisExecutor;
+pub struct RedisTrigger;
```
rylev, line 12: Nit: can we do the `_priv: ()` trick here?

crates/trigger-redis/src/lib.rs
```diff
@@ -1,203 +1,226 @@
...
-                .or_default()
-                .push(config.component.clone());
-        }
-        Ok(Self {
-            engine: Arc::new(engine),
-            server_channels,
-        })
+    type CliArgs = NoCliArgs;
+
+    type InstanceState = ();
```
rylev, line 38: Random thought: I like how `type CliArgs = NoCliArgs;` reads much better than `type InstanceState = ();`. Perhaps we should consider a `NoInstanceState` type which could be used for empty instance state instead of `()`.

crates/trigger-redis/src/lib.rs
```diff
@@ -1,203 +1,226 @@
...
         address: String,
+        trigger_app: Arc<TriggerApp<RedisTrigger>>,
         channel_components: ChannelComponents,
-    ) -> Result<()> {
-        tracing::info!("Connecting to Redis server at {}", address);
-        let mut client = Client::open(address.to_string())?;
-        let mut pubsub = client
-            .get_async_connection()
+    ) -> anyhow::Result<Self> {
+        let client = Client::open(address)?;
```
rylev, line 132: Is there a reason to open the client connection here instead of in `run_listener`?

rylev (approved):
Looks great! I hope the rebase is easy.

rylev:
@lann I rebased this and am going to merge since we have some other changes that move lots of files around, and it'll be easier on everyone if we just get this one in.

crates/trigger-redis/src/lib.rs
lann: fwiw `NoCliArgs` is only necessary because of the `type CliArgs: clap::Args` bound.

crates/trigger-redis/src/lib.rs
lann: `open` is a lie - its just parsing the `address`

PR #2755, opened 2024-08-26
llm-factor: migrate to `candle`

### LINKED ISSUE #2735: Migrate from `rustformes/llm` to `Candle`
Spin relies on [`rustformers/llm`](https://github.com/rustformers/llm). The library has now been archived and no longer actively maintained. The usage of that library has also constrained us to use only models in the `ggml` format which is not popular anymore and most models are not distributed in that format. 

I suggest that we migrate to [`candle`](github.com/huggingface/candle/) which will allow us to be able to use models in the much more popular `safetensors` format. This migration will also unlock the ability to use newer models like LLama3(.1). 

The current implementation supports a few models as listed as [features](https://github.com/rustformers/llm/blob/b11ffb14fde039f068c808df8dcbd1257a01b8db/crates/llm/Cargo.toml#L37). I think we can begin with only supporting Llama models for inferencing first. We can then optionally add support for other models as desired and feature-gate it to keep the binary size in check. 

I would love to hear your thoughts!

karthik2804:
This PR replaces the dependency on `rustformers/llm` to `huggingface/candle`. This allows us to run newer models like Llama 3(.1). This now requires the models to be of the `safetensors` format. 

This PR also removes the concept of well-known models. This ensures a consistent directory structure for all models. The rationale is that, with this change, the only group of models initially supported is the Llama family. 

Closes #2735

### DIFF
--- a/crates/llm-local/Cargo.toml
+++ b/crates/llm-local/Cargo.toml
@@ -5,31 +5,29 @@ authors = { workspace = true }
 edition = { workspace = true }
 
 [dependencies]
-anyhow = { workspace = true }
-candle = { git = "https://github.com/huggingface/candle", rev = "b80348d22f8f0dadb6cc4101bde031d5de69a9a5", package = "candle-core" }
-candle-nn = { git = "https://github.com/huggingface/candle", rev = "b80348d22f8f0dadb6cc4101bde031d5de69a9a5" }
-chrono = "0.4"
-llm = { git = "https://github.com/rustformers/llm", rev = "2f6ffd4435799ceaa1d1bcb5a8790e5b3e0c5663", features = [
-  "tokenizers-remote",
-  "llama",
-], default-features = false }
-lru = "0.12"
+anyhow = "1.0"
+candle = { git = "https://github.com/huggingface/candle", rev = "e3261216b157a7305c18ccdd766b6e2a41afe483", package = "candle-core" }
+candle-nn = { git = "https://github.com/huggingface/candle", rev = "e3261216b157a7305c18ccdd766b6e2a41afe483" }
+candle-transformers = { git = "https://github.com/huggingface/candle", rev = "e3261216b157a7305c18ccdd766b6e2a41afe483" }
+chrono = "0.4.26"
+lru = "0.9.0"
 num_cpus = "1"
 rand = { workspace = true }
 safetensors = "0.3.3"
 serde = { workspace = true }
+serde_json = "1.0.125"
 spin-common = { path = "../common" }
 spin-core = { path = "../core" }
 spin-world = { path = "../world" }
 terminal = { path = "../terminal" }
-tokenizers = "0.13.4"
-tokio = { version = "1", features = ["macros", "sync"] }
+tokenizers = "0.19.1"
+tokio = { version = "1.32.0", features = ["macros", "sync"] }
 tracing = { workspace = true }
 
 [features]
 default = []
-metal = ["llm/metal"]
-cublas = ["llm/cublas"]
+metal = ["candle/metal", "candle-nn/metal", "candle-transformers/metal"]
+cublas = ["candle/cuda", "candle-nn/cuda", "candle-transformers/cuda"]
 
 [lints]
 workspace = true
--- a/crates/llm-local/src/bert.rs
+++ b/crates/llm-local/src/bert.rs
@@ -4,7 +4,7 @@
 ///
 /// TODO: Remove this file when a new release of Candle makes it obsolete.
 use anyhow::{bail, Result};
-use candle::{DType, Tensor};
+use candle::{DType, Module, Tensor};
 use candle_nn::{Embedding, VarBuilder};
 use serde::Deserialize;
 
--- a/crates/llm-local/src/lib.rs
+++ b/crates/llm-local/src/lib.rs
@@ -1,33 +1,57 @@
 mod bert;
+mod llama;
 
 use anyhow::Context;
 use bert::{BertModel, Config};
-use candle::DType;
+use candle::{safetensors::load_buffer, DType};
 use candle_nn::VarBuilder;
-use llm::{
-    InferenceFeedback, InferenceParameters, InferenceResponse, InferenceSessionConfig, Model,
-    ModelArchitecture, ModelKVMemoryType, ModelParameters,
-};
-use rand::SeedableRng;
 use spin_common::ui::quoted_path;
+use spin_core::async_trait;
 use spin_world::v2::llm::{self as wasi_llm};
 use std::{
-    collections::hash_map::Entry,
-    collections::HashMap,
-    convert::Infallible,
+    collections::{hash_map::Entry, HashMap},
     path::{Path, PathBuf},
-    sync::{Arc, Mutex},
+    str::FromStr,
+    sync::Arc,
 };
 use tokenizers::PaddingParams;
 
 const MODEL_ALL_MINILM_L6_V2: &str = "all-minilm-l6-v2";
+type ModelName = String;
 
 #[derive(Clone)]
 pub struct LocalLlmEngine {
     registry: PathBuf,
-    use_gpu: bool,
-    inferencing_models: HashMap<(String, bool), Arc<dyn llm::Model>>,
+    inferencing_models: HashMap<ModelName, Arc<dyn InferencingModel>>,
     embeddings_models: HashMap<String, Arc<(tokenizers::Tokenizer, BertModel)>>,
+}
+
+#[derive(Debug)]
+enum InferencingModelArch {
+    Llama,
+}
+
+impl FromStr for InferencingModelArch {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "llama" => Ok(InferencingModelArch::Llama),
+            _ => Err(()),
+        }
+    }
+}
+
+/// A model that is prepared and cached after loading.
+///
+/// This trait does not specify anything about if the results are cached.
+#[async_trait]
+trait InferencingModel: Send + Sync {
+    async fn infer(
+        &self,
+        prompt: String,
+        params: wasi_llm::InferencingParams,
+    ) -> anyhow::Result<wasi_llm::InferencingResult>;
 }
 
 impl LocalLlmEngine {
@@ -38,56 +62,11 @@ impl LocalLlmEngine {
         params: wasi_llm::InferencingParams,
     ) -> Result<wasi_llm::InferencingResult, wasi_llm::Error> {
         let model = self.inferencing_model(model).await?;
-        let cfg = InferenceSessionConfig {
-            memory_k_type: ModelKVMemoryType::Float16,
-            memory_v_type: ModelKVMemoryType::Float16,
-            n_batch: 8,
-            n_threads: num_cpus::get(),
-        };
 
-        let mut session = Model::start_session(model.as_ref(), cfg);
-        let inference_params = InferenceParameters {
-            sampler: generate_sampler(params),
-        };
-        let mut rng = rand::rngs::StdRng::from_entropy();
-        let mut text = String::new();
-
-        #[cfg(debug_assertions)]
-        {
-            terminal::warn!(
-                "\
-                This is a debug build - running inference might be prohibitively slow\n\
-                You may want to consider switching to the release build"
-            )
-        }
-        let res = session.infer::<Infallible>(
-            model.as_ref(),
-            &mut rng,
-            &llm::InferenceRequest {
-                prompt: prompt.as_str().into(),
-                parameters: &inference_params,
-                play_back_previous_tokens: false,
-                maximum_token_count: Some(params.max_tokens as usize),
-            },
-            &mut Default::default(),
-            |r| {
-                match r {
-                    InferenceResponse::InferredToken(t) => text.push_str(&t),
-                    InferenceResponse::EotToken => return Ok(InferenceFeedback::Halt),
-                    _ => {}
-                };
-                Ok(InferenceFeedback::Continue)
-            },
-        );
-        let stats = res.map_err(|e| {
-            wasi_llm::Error::RuntimeError(format!("Error occurred during inferencing: {e}"))
-        })?;
-        let usage = wasi_llm::InferencingUsage {
-            prompt_token_count: stats.prompt_tokens as u32,
-            generated_token_count: (stats.predict_tokens - stats.prompt_tokens) as u32,
-        };
-        let response = wasi_llm::InferencingResult { text, usage };
-        Ok(response)
+        model
+            .infer(prompt, params)
+            .await
+            .map_err(|e| wasi_llm::Error::RuntimeError(e.to_string()))
     }
 
     pub async fn generate_embeddings(
@@ -103,10 +82,9 @@ impl LocalLlmEngine {
 }
 
 impl LocalLlmEngine {
-    pub fn new(registry: PathBuf, use_gpu: bool) -> Self {
+    pub fn new(registry: PathBuf) -> Self {
         Self {
             registry,
-            use_gpu,
             inferencing_models: Default::default(),
             embeddings_models: Default::default(),
         }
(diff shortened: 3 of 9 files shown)


radu-matei:
Given this is a breaking change, I'd suggest adding the 3.0 label.

karthik2804:
@radu-matei I do not believe I can add labels in this repository.

crates/llm-local/Cargo.toml
```diff
@@ -6,30 +6,32 @@ edition = { workspace = true }
 
 [dependencies]
 anyhow = "1.0"
-candle = { git = "https://github.com/huggingface/candle", rev = "b80348d22f8f0dadb6cc4101bde031d5de69a9a5", package = "candle-core" }
-candle-nn = { git = "https://github.com/huggingface/candle", rev = "b80348d22f8f0dadb6cc4101bde031d5de69a9a5" }
+candle = { git = "https://github.com/huggingface/candle", package = "candle-core", features = [
```
rylev, line 9: Why are we still using the git dependency instead of crates.io?

crates/llm-local/src/bert.rs
```diff
@@ -4,7 +4,7 @@
 ///
 /// TODO: Remove this file when a new release of Candle makes it obsolete.
```
rylev, line 5: Has this still not happened?

crates/llm-local/src/llama.rs
```diff
@@ -0,0 +1,168 @@
...
+    } else {
+        Ok(Device::Cpu)
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct LlamaModels {
+    model: Arc<Llama>,
+    config: Config,
+    cache: Box<Cache>,
```
rylev, line 34: Do we want to clone the cache every time we clone `LlamaModels`?

crates/llm-local/Cargo.toml
karthik2804: it is. due to use using `llama::LlamaEosToks` which is not part of the last release available on crates.io.

crates/llm-local/src/llama.rs
karthik2804: removed the usage of `Box` on cache

crates/llm-local/src/bert.rs
karthik2804: I would prefer to leave the embeddings part unchanged in this PR if that is acceptable.

crates/llm-local/Cargo.toml
rylev: Can we keep a `rev` here so that we're not relying on Cargo.lock for the right version?

crates/llm-local/src/llama.rs
```diff
@@ -0,0 +1,168 @@
...
+impl CachedInferencingModel for LlamaModels {
+    async fn infer(
+        &self,
+        prompt: String,
+        params: wasi_llm::InferencingParams,
+    ) -> anyhow::Result<wasi_llm::InferencingResult> {
+        let model = Arc::clone(&self.model);
+        let config = &self.config;
+        let tokenizer = self.tokenizer.clone();
+        let mut cache = self.cache.clone();
```
rylev, line 80: Just conforming that we want a new `cache` everytime we run `infer`? I'm not sure how cache is supposed to be used here.

crates/llm-local/Cargo.toml
karthik2804: Added a `rev` tag.

crates/llm-local/src/llama.rs
karthik2804: As I understand it, we need a new cache for each `infer` call as the cache must contain only relevant data to each call.

crates/llm-local/src/lib.rs
```diff
@@ -381,75 +329,10 @@ fn load_tokenizer(tokenizer_file: &Path) -> anyhow::Result<tokenizers::Tokenizer
 }
 
 fn load_model(model_file: &Path) -> anyhow::Result<BertModel> {
-    let buffer = std::fs::read(model_file)
-        .with_context(|| format!("Failed to read model file {}", quoted_path(model_file)))?;
-    let weights = safetensors::SafeTensors::deserialize(&buffer)?;
-    let vb = VarBuilder::from_safetensors(vec![weights], DType::F32, &candle::Device::Cpu);
+    // TODO: Check if there is a safe way to load the model from the file
```
karthik2804, line 332: The safe method seems to load only a single safe tensor file.

karthik2804:
The test failure does not seem to be related?

crates/llm-local/src/llama.rs
```diff
@@ -0,0 +1,168 @@
...
+        let config: LlamaConfig = serde_json::from_slice(&fs::read(config_path)?)?;
+
+        // TODO: flash attention is supposed to minimize memory read and writes - Do we want to turn it on
+        let config = config.into_config(false);
+        let cache = llama::Cache::new(true, dtype, &config, &device)?;
+
+        let safetensor_files = load_safetensors(model_dir, MODEL_SAFETENSORS_INDEX_FILE)?;
+
+        // TODO: Check if there is a safe way to load the model from the file
+        let vb = unsafe { VarBuilder::from_mmaped_safetensors(&safetensor_files, dtype, &device)? };
```
rylev, line 57: I don't believe we can consider this safe to call. From the docs:

> All file-backed memory map constructors are marked unsafe because of the potential for Undefined Behavior (UB) using the map if the underlying file is subsequently modified, in or out of process. Applications must consider the risk and take appropriate precautions when using file-backed maps. Solutions such as file permissions, locks or process-private (e.g. unlinked) files exist but are platform specific and limited.

It is certainly possible for the files to be modified out of band causing undefined behavior. It seems like `VarBuilder::from_buffered_safetensors` should do the trick at the expense of more memory pressure. 

I think we want to solve this issue before merging.

crates/llm-local/src/lib.rs
rylev: See my other comment below.

crates/llm-local/src/lib.rs
```diff
@@ -164,73 +144,36 @@ impl LocalLlmEngine {
     async fn inferencing_model(
         &mut self,
         model: wasi_llm::InferencingModel,
-    ) -> Result<Arc<dyn Model>, wasi_llm::Error> {
-        let use_gpu = self.use_gpu;
-        let progress_fn = |_| {};
-        let model = match self.inferencing_models.entry((model.clone(), use_gpu)) {
+    ) -> Result<Arc<dyn CachedInferencingModel>, wasi_llm::Error> {
+        // let use_gpu = self.use_gpu;
```
rylev, line 148: What's happening here? Are we completely abandoning this option. If so, let's remove it all the way up the stack.

crates/llm-local/src/lib.rs
```diff
@@ -1,93 +1,73 @@
...
+}
+
 impl LocalLlmEngine {
     pub async fn infer(
         &mut self,
         model: wasi_llm::InferencingModel,
         prompt: String,
         params: wasi_llm::InferencingParams,
     ) -> Result<wasi_llm::InferencingResult, wasi_llm::Error> {
+        // return self.inference(model).await;
```
rylev, line 64: Delete?

crates/llm-local/src/lib.rs
```diff
@@ -1,93 +1,73 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "llama" => Ok(InferencingModelArch::Llama),
+            _ => Err(()),
+        }
+    }
+}
+
+#[async_trait]
+trait CachedInferencingModel: Send + Sync {
```
rylev, line 49: Can we document this trait? What about it makes it `Cached`? Are implementors required to cache results or does it just happen that the current implementors do?

rylev (changes requested):

crates/llm-local/src/lib.rs
karthik2804: Yes, we do not need it anymore.

karthik2804: will remove it all the way

crates/llm-local/src/llama.rs
karthik2804: I agree that we want to solve this but I am not sure if `from_buffered_safetensors` is a drop in replacement as it seems to take one binary buffer for the safetensors, I am currently unsure on how to go from multiple file to a single buffer. I will look into this.

crates/llm-local/src/utils.rs
```diff
@@ -0,0 +1,25 @@
...
+pub fn load_safetensors(model_dir: &Path, json_file: &str) -> Result<Vec<std::path::PathBuf>> {
+    let json_file = model_dir.join(json_file);
+    let json_file = std::fs::File::open(json_file)?;
+    let json: serde_json::Value =
+        serde_json::from_reader(&json_file).map_err(candle::Error::wrap)?;
+    let weight_map = match json.get("weight_map") {
+        None => candle::bail!("no weight map in {json_file:?}"),
+        Some(serde_json::Value::Object(map)) => map,
+        Some(_) => candle::bail!("weight map in {json_file:?} is not a map"),
+    };
```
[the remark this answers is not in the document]
karthik2804: I reverted this change and the other because it was leading to some off error, where the returned vector was a duplicate of the same thing repeated several times which meant the same files were being loaded over and over which led to consuming large amounts of memory.

crates/llm-local/src/token_output_stream.rs
```diff
@@ -0,0 +1,68 @@
+/// Implementation for TokenOutputStream Code is borrowed from
+/// https://github.com/huggingface/candle/blob/main/candle-examples/src/token_output_stream.rs
+/// (Commit SHA 4fd00b890036ef67391a9cc03f896247d0a75711)
+///
+/// This is a wrapper around a tokenizer to ensure that tokens can be returned to the user in a
```
rylev, line 5: Nit: typically doc comments start with a description of the type and then add other notes (such as where the code may be borrowed from) at the end. This seems to be the other way around.

crates/llm-local/src/token_output_stream.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn new(tokenizer: tokenizers::Tokenizer) -> Self {
+        Self {
+            tokenizer,
+            tokens: Vec::new(),
+            prev_index: 0,
+            current_index: 0,
+        }
+    }
+
+    fn decode(&self, tokens: &[u32]) -> anyhow::Result<String> {
```
rylev, line 24: Nit: I'd personally prefer private methods to come after public ones.

crates/llm-local/src/token_output_stream.rs
```diff
@@ -0,0 +1,68 @@
...
+            let (_, text) = text.split_at(prev_text.len());
+            self.prev_index = self.current_index;
+            self.current_index = self.tokens.len();
+            Ok(Some(text.to_string()))
+        } else {
+            Ok(None)
+        }
+    }
+
+    pub fn decode_rest(&self) -> anyhow::Result<Option<String>> {
```
rylev, line 53: Any docs for this method? What does this do?

crates/llm-local/src/token_output_stream.rs
```diff
@@ -0,0 +1,68 @@
...
+            prev_index: 0,
+            current_index: 0,
+        }
+    }
+
+    fn decode(&self, tokens: &[u32]) -> anyhow::Result<String> {
+        match self.tokenizer.decode(tokens, true) {
+            Ok(str) => Ok(str),
+            Err(err) => anyhow::bail!("cannot decode: {err}"),
+        }
```
rylev, line 28:
```suggestion
       self.tokenizer.decode(tokens, true).context("failed to decode token stream")
```

crates/llm-local/src/token_output_stream.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub fn next_token(&mut self, token: u32) -> anyhow::Result<Option<String>> {
+        let prev_text = if self.tokens.is_empty() {
+            String::new()
+        } else {
+            let tokens = &self.tokens[self.prev_index..self.current_index];
+            self.decode(tokens)?
+        };
+        self.tokens.push(token);
+        let text = self.decode(&self.tokens[self.prev_index..])?;
+        if text.len() > prev_text.len() && text.chars().last().unwrap().is_alphanumeric() {
```
rylev, line 43: I don't fully understand what this check is supposed to be doing. Why do we care about the length of the next text vs the previous, and why do we care whether the last character is alphanumeric?

crates/llm-local/src/token_output_stream.rs
```diff
@@ -0,0 +1,68 @@
...
+    fn decode(&self, tokens: &[u32]) -> anyhow::Result<String> {
+        match self.tokenizer.decode(tokens, true) {
+            Ok(str) => Ok(str),
+            Err(err) => anyhow::bail!("cannot decode: {err}"),
+        }
+    }
+
+    /// Processes the next token in the sequence, decodes the current token stream,
+    /// and returns any newly decoded text.
+    /// https://github.com/huggingface/text-generation-inference/blob/5ba53d44a18983a4de32d122f4cb46f4a17d9ef6/server/text_generation_server/models/model.py#L68
```
rylev, line 33:
```suggestion
    /// Processes the next token in the sequence, decodes the current token stream,
    /// and returns any newly decoded text.
    ///
    /// Based on the following code: <https://github.com/huggingface/text-generation-inference/blob/5ba53d44a18983a4de32d122f4cb46f4a17d9ef6/server/text_generation_server/models/model.py#L68>
```

crates/llm-local/src/llama.rs
```diff
@@ -0,0 +1,209 @@
...
+            usage: InferencingUsage {
+                prompt_token_count: tokens.len() as u32,
+                generated_token_count: tokens_generated,
+            },
+        })
+    }
+}
+
+///  Loads a list of SafeTensors file paths from a given model directory and
+///  path to the model index JSON file.
```
rylev, line 186: The docs don't seem right. The path that gets loaded is the model_dir + the json_file. This makes it sound like `json_file` is an independent path which it is not.

crates/llm-local/src/llama.rs
```diff
@@ -0,0 +1,209 @@
...
+                Some(llama::LlamaEosToks::Multiple(ref eos_ids))
+                    if eos_ids.contains(&next_token) =>
+                {
+                    break;
+                }
+                _ => (),
+            }
+            // Decode the token and add it to the output.
+            if let Some(t) = tokenizer.next_token(next_token)? {
+                print!("{}", t);
```
rylev, line 165: Do we want to print this?

karthik2804: That was used for some debugging I will remove it.

crates/llm-local/src/llama.rs
karthik2804: updating to mention that it is relative to model directory.

crates/llm-local/src/token_output_stream.rs
karthik2804: The length check is to see if we have any new tokens. The alphanumeric check is supposed to be to check if we have a valid token to decode. That is what I gather from the python function the docs link to 

https://github.com/huggingface/text-generation-inference/blob/5ba53d44a18983a4de32d122f4cb46f4a17d9ef6/server/text_generation_server/models/model.py#L68

crates/llm-local/src/token_output_stream.rs
karthik2804: It does look like I cannot do this because `tokenizer.decode` returns a  `Result<String, Box<dyn Error + Send + Sync>>` which does not seem to be suitable to use context on(?)

crates/llm-local/src/lib.rs
```diff
@@ -1,35 +1,60 @@
...
+        match s {
+            "llama" => Ok(InferencingModelArch::Llama),
+            _ => Err(()),
+        }
+    }
+}
+
+/// `CachedInferencingModel` implies that the model is prepared and cached after
+/// loading, allowing faster future requests by avoiding repeated file reads
+/// and decoding. This trait does not specify anything about if the results are cached.
```
rylev, line 48: Formatting nit - typically you want a single line descriptor, a blank line, and then further information:

```suggestion
/// A model that is prepared and cached after loading.
///
/// This trait does not specify anything about if the results are cached.
```

crates/llm-local/src/lib.rs
rylev: I'm fine with keeping the name, but I personally find the name `CachedInferencingModel` confusing when implementors aren't required to cache anything. `InferencingModel` seems like a more appropriate name.

crates/llm-local/src/lib.rs
```diff
@@ -381,75 +327,11 @@ fn load_tokenizer(tokenizer_file: &Path) -> anyhow::Result<tokenizers::Tokenizer
 }
 
 fn load_model(model_file: &Path) -> anyhow::Result<BertModel> {
-    let buffer = std::fs::read(model_file)
-        .with_context(|| format!("Failed to read model file {}", quoted_path(model_file)))?;
-    let weights = safetensors::SafeTensors::deserialize(&buffer)?;
-    let vb = VarBuilder::from_safetensors(vec![weights], DType::F32, &candle::Device::Cpu);
+    let device = &candle::Device::Cpu;
+    // TODO: Check if there is a safe way to load the model from the file
```
rylev, line 331: I imagine this TODO can be removed now?

crates/llm-local/src/token_output_stream.rs
rylev: The python code is dealing with unfinished utf-8 byte sequences which is not possible at this point in the Rust code. Rust `char`s are guaranteed to be valid utf-8. The check for alphanumeric chars is checking that the character is `A-Z | a-z | 0-9` which does seem to be what we want.

The `Tokenizer::decode` function returns `String`s so I'm guessing somehow the `tokenizer` crate is taking care of byte sequences that aren't valid utf-8?

crates/llm-local/src/token_output_stream.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// additional part as `Some(String)`. Otherwise, returns `None`.
+    pub fn decode_rest(&self) -> anyhow::Result<Option<String>> {
+        let prev_text = if self.tokens.is_empty() {
+            String::new()
+        } else {
+            let tokens = &self.tokens[self.prev_index..self.current_index];
+            self.decode(tokens)?
+        };
+        let text = self.decode(&self.tokens[self.prev_index..])?;
+        if text.len() > prev_text.len() {
```
rylev, line 61: We're decoding the previous text just to check its length against the new text. I still don't fully understand why that's necessary, and it seems fairly wasteful. Can you explain why this check is necessary?

karthik2804: That is not a part I fully understand either. It is a lift and shift from the [`candle` samples](https://github.com/huggingface/candle/blob/main/candle-examples/src/token_output_stream.rs)

crates/llm-local/src/token_output_stream.rs
karthik2804: Here is the relevant rust version from where this is borrowed.
https://github.com/huggingface/candle/blob/6eea45a761fc1636b5e8012d02bdaa93321652ca/candle-examples/src/token_output_stream.rs#L43

crates/llm-local/src/token_output_stream.rs
radu-matei: Not sure I follow the part with "just to check". It's also decoding and returning the actual text (conditionally, if there is new text generated).

rylev: Right, but we're returning part of `text` (not `prev_text`). The only information from `prev_text` we use is it's byte length.

karthik2804: Okay I have spent a little more time staring at this, My understanding now is that adding a token does not mean we have a new character and also addition of a token might introduce a non alphanumeric character because we currently are in the middle of a bunch of tokens that make an actual word. We can definitely look at just storing the length of the prev text to avoid the extra `decode`. 

alternatively one other thing we could do is drop the `token_output_stream` on this PR and just decode once the entire content is generated since we do not stream the responses back either way. This approach is strictly no worse than what we have today.

rylev: YAGNI I'd be in favor of dropping the streaming for now - since we're not actually using it, it's simply more maintenance overhead.

karthik2804: I have pushed a commit removing it. If everything looks good, I will squash the commits.

rylev (approved):
🎉

PR #2756, opened 2024-08-26
Fix path handling in sqlite and kv

rylev:
The follow changes were made:
* The "spin" key-value store handler's take an optional `base_path`. This path is used to determine the location for stores. If it is `None`, an in-memory database will always be used (regardless of if a path is set in the runtime config).
  * If the runtime-config doesn't provide a path, then an in-memory database will be used (regardless of whether a base path is set).
* SQLite databases work differently. An optional base directory path is provided to the `RuntimeConfigResolver` - if it is set the default db will be created at that path - otherwise it will be in-memory. A second *non*-optional path is provided for the base path of SQLite databases with a path set in the runtime config. This will always be the parent directory of the database if a path is provided in the runtime-config. 
  * In other words, there is no way to force all SQLite databases to be in-memory regardless of runtime config (unlike the mechanism above for key-value which can force all key-values to be in-memory, regardless of runtime-config). Should we also make this optional and provide the CLI with some configuration option?
* Lots of tests were changed as they were simply not testing what they said they were. For example, some tests simply asserted that the temporary directory created at the beginning of test existed. Such tests would always pass as temporary directories are created immediately and not lazily.

### DIFF
--- a/crates/factor-key-value-spin/src/lib.rs
+++ b/crates/factor-key-value-spin/src/lib.rs
@@ -17,15 +17,49 @@ pub struct SpinKeyValueStore {
 impl SpinKeyValueStore {
     /// Create a new SpinKeyValueStore with the given base path.
     ///
-    /// If the database directory is None, the database will always be in-memory.
-    /// If it's `Some`, the database will be stored at the combined `base_path` and
-    /// the `path` specified in the runtime configuration.
+    /// If `base_path` is `Some`, the database will be stored at the combined
+    /// `base_path` and the `path` specified in the runtime configuration. Otherwise,
+    /// only if the `path` in the runtime config is an absolute path will it be used as is.
+    /// In all other cases, an in-memory database will be used.
     pub fn new(base_path: Option<PathBuf>) -> Self {
         Self { base_path }
     }
 }
 
-/// Runtime configuration for the SQLite key-value store.
+impl MakeKeyValueStore for SpinKeyValueStore {
+    const RUNTIME_CONFIG_TYPE: &'static str = "spin";
+
+    type RuntimeConfig = SpinKeyValueRuntimeConfig;
+
+    type StoreManager = KeyValueSqlite;
+
+    fn make_store(
+        &self,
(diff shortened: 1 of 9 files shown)


crates/factor-key-value-spin/src/lib.rs
```diff
@@ -11,12 +11,16 @@ use spin_key_value_sqlite::{DatabaseLocation, KeyValueSqlite};
...
     /// The base path or directory for the SQLite database file.
-    base_path: PathBuf,
+    base_path: Option<PathBuf>,
 }
 
 impl SpinKeyValueStore {
     /// Create a new SpinKeyValueStore with the given base path.
-    pub fn new(base_path: PathBuf) -> Self {
+    ///
+    /// If the database directory is None, the database will always be in-memory.
```
lann, line 20: I think this isn't quite right. If the directory is None and there is an _absolute_ path I would expect that to work fine. This might be important for SpinKube.

When the directory is None and the path is _relative_ I'm not sure what the current behavior is but it seems like it would be least surprising for that to be a fatal error.

rylev: I've changed it so that absolute paths are still allowed when `--state-dir` is unset. However, this does mean a runtime can never fully force in-memory databases.

rylev: Ok I'll error for now - probably easier in the future to make it less restrictive than it is to make it more.

lann: > a runtime can never fully force in-memory databases

If I'm reading `main` correctly it looks like you can omit `path` to force an in-memory database.

rylev: Yes, but that's in the runtime config not in the actual running of the runtime engine. Are we assuming that whoever is controlling the runtime is the same as whoever is setting the runtime config? We're requiring coordination between whoever is setting the runtime config and whoever is actually running the runtime binary.

rylev:
Closing in favor of #2757

PR #2759, opened 2024-08-26
Build fixes

rylev:
* Make Clippy happy
* Remove the smoke test
* Remove factors specific CI


crates/componentize/src/module_info.rs
lann, line 1: I'm having trouble getting my local clippy to be angry about this :thinking:

rylev: Which version of Rust are you running?

rylev: In any case, I think the `||` is nicer 🤷

lann (approved):

crates/componentize/src/module_info.rs
lann: 1.80

PR #2760, opened 2024-08-26
Replace tempdir with tempfile

lann:
`tempdir` is abandoned

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -312,7 +312,7 @@ impl<T: Trigger> TriggerAppBuilder<T> {
         let use_gpu = true;
         let state_dir = match options.state_dir {
             // Make sure `--state-dir=""` unsets the state dir
-            Some(s) if s.trim().is_empty() => None,
+            Some("") => None,
             Some(s) => Some(PathBuf::from(s)),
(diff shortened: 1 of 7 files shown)


crates/trigger/src/cli.rs
```diff
@@ -312,7 +312,7 @@ impl<T: Trigger> TriggerAppBuilder<T> {
         let use_gpu = true;
         let state_dir = match options.state_dir {
             // Make sure `--state-dir=""` unsets the state dir
-            Some(s) if s.trim().is_empty() => None,
+            Some("") => None,
```
lann, line 315: Sneaky lint fix...

kate-goldenring (approved):
LGTM! Thanks!

crates/trigger/src/cli.rs
rylev: This isn't equivalent, no? `Some("  ")` would match the first but not the second.

lann: Ah sorry I blindly obeyed clippy. That said, `--state-dir " "` seems more likely to be an accident...

lann: Though it now occurs to me that its perfectly valid to create a directory named `" "`...

PR #2761, opened 2024-08-26
Initial key values

calebschoepp:
We don't want to unnecessarily open the default store. This adds a trace that the user doesn't really want to see in OTel and is also just generally unnecessary.


lann (approved):

lann:
Rebasing should fix CI

PR #2764, opened 2024-08-26
Fix more factors tests

lann:
- Configure `rustls` to use `ring` instead of new `aws-lc-rs` core which was causing CI build problems for Ubuntu and Windows 
- Switch `TomlResolver::state_dir` from `std::fs::canonicalize` to `std::path::absolute` which doesn't require the path to exist
- Side-quest: Fix `integration_tests::bad_build_test` when `RUST_LOG` is set by parent shell


lann:
@rylev FYI I'm YOLOing this in to fix the build

PR #2741, opened 2024-08-21
Add counter telemetry(WIP)

me-diru:
Fixes #2564 

I think this captures the metrics of LLM and Key-Value store

PR #2765, opened 2024-08-26
factors: Fix toml key tracking

lann:
@rylev I don't know if this is what you had in mind; just trying to unblock https://github.com/fermyon/spin/pull/2741


lann:
@me-diru @calebschoepp FYI

rylev (approved):

PR #2772, opened 2024-08-27
Relax buggy wasi-sdk heuristic

lann:
Unfortunately there are some toolchains that don't emit enough information for this heuristic to detect that they are positively safe, so we'll need to default to allow.


rylev (approved):
Unfortunate, but such is life.

PR #2775, opened 2024-08-27
feat(wasi-observe): WASI Observe factor (tracing impl)

calebschoepp:
This is what a wasi-observe factor would look like backed by [tracing](https://docs.rs/tracing/0.1.40/tracing/). This will likely be superseded by an implementation backed by [opentelemetry](https://docs.rs/opentelemetry/0.24.0/opentelemetry/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2330,6 +2330,25 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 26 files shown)


calebschoepp:
We're definitely not going down this path.

PR #2776, opened 2024-08-28
Use Swift 6.0 for tests

lann:
Currently just testing CI here...

Swift 5.x uses an old wasi-sdk which _might_ make it susceptible to https://github.com/WebAssembly/wasi-libc/pull/377 when used with WAGI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -196,10 +196,9 @@ jobs:
         with:
           node-version: '20.x'
       - name: Install Swift
-        run: |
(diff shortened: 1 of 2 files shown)


rylev:
Looks like the Swift installation wasn't quite right:

```
stdout:
Building component test with `swiftc -target wasm32-unknown-wasi main.swift -o main.wasm`

stderr:
error: missing external dependency '/opt/hostedtoolcache/swiftwasm/wasm-6.0-SNAPSHOT-2024-07-20-a/x86_64/usr/lib/swift/wasi/static-executable-args.lnk'
```

PR #2780, opened 2024-08-28
Run sqlite statements

rylev:
This works, but I don't love how it turned out. `ResolvedRuntimeConfig` knows nothing about the underlying runtime config type that is parsed from the runtime config source. This means that only the caller knows that the parsed runtime config has access to sqlite specific runtime config (and thus the list of labels -> connection creators). So we have to pass that information back down from the caller into the callee.

It's perhaps not 100% elegant, but it works. Perhaps we can think of a better way to structure this so that this round about data passing isn't necessary.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -68,16 +68,6 @@ error: the `wasm32-wasi` target is not installed
     std::fs::create_dir_all("target/test-programs").unwrap();
 
     build_wasm_test_program("core-wasi-test.wasm", "crates/core/tests/core-wasi-test");
-    // build_wasm_test_program("redis-rust.wasm", "crates/trigger-redis/tests/rust");
-    // build_wasm_test_program(
-    //     "spin-http-benchmark.wasm",
-    //     "crates/trigger-http/benches/spin-http-benchmark",
-    // );
-    // build_wasm_test_program(
-    //     "wagi-benchmark.wasm",
-    //     "crates/trigger-http/benches/wagi-benchmark",
-    // );
-    // build_wasm_test_program("timer_app_example.wasm", "examples/spin-timer/app-example");
 
     cargo_build(TIMER_TRIGGER_INTEGRATION_TEST);
 }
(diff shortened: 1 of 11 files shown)


lann:
Could this be done in a `ExecutorHooks::configure_app`? That has access to `AppState`, and `SqliteFactor`'s `AppState` should be able to execute without this separate logic.

build.rs
```diff
@@ -68,16 +68,6 @@ error: the `wasm32-wasi` target is not installed
     std::fs::create_dir_all("target/test-programs").unwrap();
 
     build_wasm_test_program("core-wasi-test.wasm", "crates/core/tests/core-wasi-test");
-    // build_wasm_test_program("redis-rust.wasm", "crates/trigger-redis/tests/rust");
```
rylev, line 71: I hope it's ok to sneak this in.

rylev (commented):
@lann I moved to using an `ExecutorHooks`. I like how this works now. One issue is that the work I need to do is async while `ExecutorHooks` is sync. I'm just spawning the task on the current tokio runtime, but I do wonder if we want to consider making `ExecutorHooks` async.

lann (commented):
> consider making ExecutorHooks async.

I think that would probably be fine. I don't expect ExecutorHooks to be broadly reusable, so I'm less worried about slowing app load there than in a Factor.

lann (approved):

rylev:
@lann  - I made `ExecutorHooks` an async trait. It required constraining a few `AppState`'s to be `Sync`, but I think that's fine?

lann:
> It required constraining a few AppState's to be Sync, but I think that's fine?

We should probably have a blanket constraint on `AppState`s to be `Send + Sync` anyway; I'm pretty sure it is implicitly required already.

lann:
Also notable: the one impl of `ConnectionCreator` isn't actually async...

rylev:
I made `ConnectionCreator` non-async, and I added some tests. I think this should be good to merge now (assuming CI passes).

PR #2786, opened 2024-08-29
Add SqliteDefaultStoreSummaryHook

lann:
Also:

- Add `AppState: Sync` constraints to factors traits
- Stop `KeyValueDefaultStoreSummaryHook` printing for unused default store

### DIFF
--- a/crates/factor-sqlite/src/lib.rs
+++ b/crates/factor-sqlite/src/lib.rs
@@ -169,12 +169,16 @@ impl AppState {
     /// Get a connection for a given database label.
     ///
     /// Returns `None` if there is no connection creator for the given label.
-    pub async fn get_connection(
-        &self,
-        label: &str,
-    ) -> Option<Result<Box<dyn Connection>, v2::Error>> {
+    pub fn get_connection(&self, label: &str) -> Option<Result<Box<dyn Connection>, v2::Error>> {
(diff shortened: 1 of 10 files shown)


crates/factor-sqlite/src/lib.rs
```diff
@@ -205,4 +209,11 @@ pub trait Connection: Send + Sync {
     ) -> Result<v2::QueryResult, v2::Error>;
 
     async fn execute_batch(&self, statements: &str) -> anyhow::Result<()>;
+
+    /// A human-readable summary of the connection's configuration
+    ///
+    /// Example: "libSQL at libsql://example.com"
+    fn summary(&self) -> Option<String> {
```
rylev, line 216: What's the thinking around returning `Option` here and defaulting this implementation? Seems like we're just asking for new implementors to forget to implement this method leading to subtle bugs. Why not force implementors to implement this?

rylev (approved):
Just one question, but we don't have to block merging this.

crates/factor-sqlite/src/lib.rs
lann: My thinking is that this is really only applicable to `spin up`, so if some other runtime has its own store impls it probably doesn't care about this method.

lann: also the consequence for omitting this is small, at least as currently used

PR #2787, opened 2024-08-29
feat(wasi-observe): WASI Observe factor implementation

calebschoepp:
- Adds a new WASI Observe factor.
- Adds a suite of integration tests for the factor.
- Makes all other factors take an optional dependency on the `ObserveFactor` to handle proper span nesting of other factors.

### DIFF
--- a/crates/factor-observe/src/lib.rs
+++ b/crates/factor-observe/src/lib.rs
@@ -146,7 +146,7 @@ impl ObserveContext {
         };
 
         // If there are no active guest spans then there is nothing to do
-        let Some(current_span_id) = state.active_spans.last() else {
+        let Some(active_span) = state.active_spans.last() else {
             return;
         };
 
@@ -166,7 +166,7 @@ impl ObserveContext {
         // Now reparent the current span to the last active guest span
         let span_context = state
             .guest_spans
-            .get(*current_span_id)
+            .get(*active_span)
             .unwrap()
             .inner
             .span_context()
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1426,7 +1426,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
@@ -1469,7 +1469,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
@@ -1539,7 +1539,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
@@ -1604,7 +1604,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
@@ -1665,7 +1665,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
@@ -1727,7 +1727,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
@@ -1793,7 +1793,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
@@ -1851,7 +1851,7 @@ mod otel_integration_tests {
             },
             ServicesConfig::none(),
             |env| {
-                env.set_env_var("OTEL_EXPORTER_OTLP_ENDPOINT", collector_endpoint);
+                env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", collector_endpoint);
                 env.set_env_var("OTEL_EXPORTER_OTLP_TRACES_PROTOCOL", "grpc");
                 env.set_env_var("OTEL_BSP_SCHEDULE_DELAY", "5");
                 Ok(())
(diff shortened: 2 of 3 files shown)


crates/trigger-http/src/wasi.rs
```diff
@@ -98,27 +98,27 @@ impl HttpExecutor for WasiHttpExecutor {
...
+        // TODO(Caleb): Verify that using tracing::Span::current() does not cause regressions
         let handle = task::spawn(
             async move {
                 let result = match handler {
                     Handler::Latest(proxy) => {
                         proxy
                             .wasi_http_incoming_handler()
                             .call_handle(&mut store, request, response)
-                            .instrument(span)
+                            .instrument(tracing::Span::current())
```
lann, line 109:
```suggestion
                            .in_current_span()
```

:shrug:

crates/factor-observe/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+        // TODO: Move this duplicate logic into its own impl
+        let state = self.state.read().unwrap();
+        if state.active_spans.is_empty() {
+            return;
+        }
+
+        let parent_context = Context::new().with_remote_span_context(
+            state
+                .guest_spans
+                .get(*state.active_spans.last().unwrap().1)
```
lann, line 102: Slightly nicer imo:

```suggestion
        let Some(current_span_id) = state.active_spans.values().last() else {
            return;
        }

        let parent_context = Context::new().with_remote_span_context(
            state
                .guest_spans
                .get(*current_span_id)
```

crates/factor-observe/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub struct ObserveContext {
+    pub(crate) state: Arc<RwLock<State>>,
+}
+
+impl ObserveContext {
+    /// TODO comment
+    pub fn oh_dear_i_better_get_renamed(&self) {
+        // TODO: Move this duplicate logic into its own impl
+        let state = self.state.read().unwrap();
```
lann, line 94: Warning: trying to dedupe code that accesses a lock may be a path of pain...

crates/trigger-http/src/wasi.rs
calebschoepp: Actually I wonder if these are actually doing anything? There is a `in_current_span` on the async block already.

crates/factor-key-value/src/lib.rs
```diff
@@ -88,17 +89,23 @@ impl Factor for KeyValueFactor {
...
         let allowed_stores = app_state
             .component_allowed_stores
             .get(ctx.app_component().id())
             .expect("component should be in component_stores")
             .clone();
+        let observe_context = match ctx.instance_builder::<ObserveFactor>() {
+            Ok(factor) => Some(factor.get_observe_context()),
+            Err(spin_factors::Error::NoSuchFactor(_)) => None,
+            Err(e) => return Err(e.into()),
+        };
```
lann, line 104: Another option to consider here which would reduce the amount of boilerplate in each dependent factor:

```rust
pub struct ObserveContext {
    pub(crate) state: Option<Arc<RwLock<State>>>,
}

impl ObserveContext {
  pub fn from_prepare_context<T: RuntimeFactors, F: Factor>(
    prepare_context: &mut PrepareContext<T, F>
  ) -> ObserveContext {
    let state = // same match expr as above, but getting `Option<Arc<State>>`
    ObserveContext { state }
  }
}
```

Then `reparent_tracing_span` could handle `self.state == None` internally.

tests/integration.rs
```diff
@@ -153,72 +155,438 @@ mod integration_tests {
...
+                    );
+
+                    Ok(())
+                },
+            )
+        })
+        .await??;
+
+        Ok(())
+    }
```
calebschoepp, line 584: @lann I'm in a bit of a sync/async pickle and need some help.

This test is async so that we can start the fake opentelemetry collector. But then we execute `run_test_inited` in a sync context b/c otherwise it fails. The problem is that I need to[^1] depend on a later version of the fake opentelemetry collector which makes the `exported_spans()` method async. Is there a way I can have an async context within the sync context within the async context? Is there a way I can simplify this?

[^1]: This was necessary to get access to the event fields https://github.com/davidB/tracing-opentelemetry-instrumentation-sdk/pull/160.

Cargo.toml
```diff
@@ -66,7 +66,7 @@ terminal = { path = "crates/terminal" }
 
 subprocess = "0.2.9"
 tempfile = "3.8.0"
-tokio = { version = "1.23", features = ["full"] }
+tokio = { version = "1.40", features = ["full"] }
```
calebschoepp, line 69: This was necessary to make `fake-opentelemetry-collector` happy which relies on an up to date version of `tokio`.

crates/world/src/conversions.rs
```diff
@@ -218,3 +218,179 @@ mod llm {
...
+                tracer::TraceFlags::SAMPLED
+            } else {
+                tracer::TraceFlags::empty()
+            }
+        }
+    }
+
+    impl From<tracer::SpanContext> for opentelemetry::trace::SpanContext {
+        fn from(sc: tracer::SpanContext) -> Self {
+            // TODO(Reviewer): Should this be try_from instead an propagate this error out of the WIT?
```
calebschoepp, line 272: Some of these conversions end up being fallible b/c we're parsing things that must conform to a format. Right now I handle this by just quelling the error and silently omitting the invalid data. This doesn't feel like a great user experience (even if it is documented in the WIT interface).

An alternative would be to make this `try_from` and then expose the error out through the WIT interface. This would make change:
- `start: func(name: string, options: option<start-options>) -> span;` to `start: func(name: string, options: option<start-options>) -> result<span>;`
- `add-link: func(link: link);` to `add-link: func(link: link) -> result<()>;`

I don't love the idea of exposing errors out of these functions just for the incorrect parsing of a span context. I'm not sure what the right choice is here.

tests/integration.rs
```diff
@@ -1463,3 +1397,573 @@ route = "/..."
         Ok(())
     }
 }
+
+// TODO(Reviewer): How can I move this to a new file? I wasn't able to get the imports to work out.
```
calebschoepp, line 1401: I couldn't figure out how to put this module in another file, but still import things from the testcases module.

wit/deps/observe/world.wit
```diff
@@ -0,0 +1,8 @@
+package wasi:observe@0.2.0-draft;
+
+world imports {
+    import tracer;
+}
+
+// TODO(Reviewer): Should we make this an experimental wasi package or a Spin package
```
calebschoepp, line 7: I would suggest that we land WASI Observe as a release candidate `wasi:observe@0.2.2-rc2024-09-17`. I think it is the right choice to put it in the WASI namespace b/c WASI Observe is a [phase 1 proposal](https://github.com/WebAssembly/wasi-observe/pull/15).

If we're feeling conservative we could add this to a new `http-trigger-experimental` world where we make less support guarantees.

```wit
package fermyon:spin@2.0.0;

/// The full world of a guest targeting an http-trigger
world http-trigger {
  include platform;
  export wasi:http/incoming-handler@0.2.0;
}

/// Like `http-trigger`, but using WASI 0.2.0-rc-2023-10-18
world http-trigger-rc20231018 {
  include platform-rc20231018;
  export wasi:http/incoming-handler@0.2.0-rc-2023-10-18;
}

/// Like `http-trigger`, but experimental and with no support guarantees
world http-trigger-experimental {
  include platform-experimental;
  export wasi:http/incoming-handler@0.2.0;
}

world platform {
  include wasi:cli/imports@0.2.0;
  import wasi:http/outgoing-handler@0.2.0;
  import llm;
  import redis;
  import mqtt;
  import postgres;
  import mysql;
  import sqlite;
  import key-value;
  import variables;
}

/// Like `platform`, but using WASI 0.2.0-rc-2023-10-18
world platform-rc20231018 {
  include wasi:cli/reactor@0.2.0-rc-2023-10-18;
  import wasi:http/outgoing-handler@0.2.0-rc-2023-10-18;
  import llm;
  import redis;
  import mqtt;
  import postgres;
  import mysql;
  import sqlite;
  import key-value;
  import variables;
}

/// Like `platform`, but experimental and with no support guarantees
world platform-experimental {
  include platform;
  include wasi:observe/imports@0.2.2-rc2024-09-17;
}
```

wit/deps/observe/world.wit
```diff
@@ -0,0 +1,8 @@
...
+// TODO(Reviewer): Would adding a metrics interface to this in a future version be a breaking change?
```
calebschoepp, line 8: @dicej does adding a new interface to a wold constitute a breaking change?

wit/world.wit
```diff
@@ -24,6 +24,7 @@ world platform {
   import sqlite;
   import key-value;
   import variables;
+  include wasi:observe/imports@0.2.0-draft;
```
itowlson, line 27: Is the plan to modify the 2.0 world, or is this a placeholder for dev work?  This is the issue holding up wasi-keyvalue so very interested in the answer..!

calebschoepp:
TODO item for Caleb: It seems that this is not working with the Rust `spin_sdk``v3.0.1`.

wit/deps/observe/world.wit
tschneidereit: It does not, no: additions to interfaces are semver-compatible. For now, that is not true for additions to:
- function parameters
- struct fields
- tuple entries
- variant and enum arms

wit/deps/observe/world.wit
tschneidereit: (This is probably a topic deserving of a SIP)

I think we should use WIT's [feature gates](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md#feature-gates) for this, like so:

```wit
world http-trigger {
  include platform;
  export wasi:http/incoming-handler@0.2.0;
  @unstable(feature = wasi-observe)
  include wasi:observe@0.2.0-draft-2024-09-20/imports;
}
```

Additionally, unstable features should require runtime opt-in with something like a CLI flag along the lines of `--enable-unstable-wit-features=wasi-observe,...`

tschneidereit: Yes: sloppiness. It absolutely should be in `platform`

wit/deps/observe/world.wit
rylev: Just to clarify - it is not a breaking change from the perspective of the runtime, but it is from the perspective of the guest. In other words, if you add an import to a world, you can run old guests on an new runtime, but you cannot run new guests on an old runtime.

itowlson:
@calebschoepp Not sure where you're at with this but a heads up that we now have a minty fresh new world for you to put the wasi:observe import into - see e.g. #2869 for an example

calebschoepp:
> @calebschoepp Not sure where you're at with this but a heads up that we now have a minty fresh new world for you to put the wasi:observe import into - see e.g. #2869 for an example

Thanks for leading the way on this @itowlson

itowlson:
@calebschoepp Just checking on status - is this something you're still planning to come back to, and if so is it close to completion - or is it parked for the foreseeable future?  (I know you have a bunch of other stuff on, and completely understand that you've not had time to work on it - my intent is to ask about plans, not to nag!)

calebschoepp:
> @calebschoepp Just checking on status - is this something you're still planning to come back to, and if so is it close to completion - or is it parked for the foreseeable future? (I know you have a bunch of other stuff on, and completely understand that you've not had time to work on it - my intent is to ask about plans, not to nag!)

Thanks for checking in @itowlson. I definitely got very busy with other things and sort of just parked this — not sure when I'll have time to come back and finish this. But, I definitely want to land it at some point.

I would say the implementation is 80% of the way there. In fact this is more blocked by reaching consensus upstream in WASI Observe on what we want the API to be. Also there is some desire upstream to rename it from WASI Observe to WASI OTel b/c that is more accurate to what we're attempting to do with this WIT.

Does that give you enough detail? (I recognize I'm basically just saying idk when this will get done 🤣)

itowlson:
Yes, that's great context - thanks!  "80% done until they remodel the API into a French bistro" gives me a solid sense of where this is at and the scope of the uncertainties around it!

calebschoepp:
Superseded by https://github.com/spinframework/spin/pull/3033

PR #2789, opened 2024-08-29
Update templates for mandatory componentization

### LINKED ISSUE #2774: Swift and Grain Templates Tests Not Working
After the Factors work the Swift and Grain templates have stopped working:

## Swift

When the `http_swift_template_smoke_test` test is run, the following error appears:

```
Error: preparing wasm "/tmp/t68d4-a/main.wasm"

Caused by:
      This Wasm module appears to have been compiled with wasi-sdk version <19 which contains a critical memory safety bug. For more information, see: https://github.com/fermyon/spin/issues/2552
```

The Swift toolchain uses a fairly old version of wasi-sdk. It is possible that a newer version of the Swift toolchain would work, but as of yet we've not been able to get things working.

## Grain

When the `http_grain_template_smoke_test` test is run, the following error appears:

```
Error: Expected status 200 for / but got 500
Body: 'Exactly one of 'location' or 'content-type' must be specified'
Stderr: '2024-08-27T15:11:00.204342Z ERROR spin_http::wagi: HTTP 500 error error=Exactly one of 'location' or 'content-type' must be specified
```

Most likely the wagi module is not giving any output and thus there are no headers to parse leading to the error. Why this would be happy is not clear at the moment. 

For now, we will be turning off both tests while we continue to investigate.

lann:
The Swift template is no longer usable due to the deprecation of running uncomponentized modules and issues with componentizing: https://github.com/fermyon/spin/issues/2774

The Grain template has been updated based on upstream advice: https://github.com/grain-lang/grain/issues/2153

Closes #2774


itowlson:
There are some tests in the `templates` crate that use the installed templates and make assertions by counting.  The const for the expected total count is called `TPLS_IN_THIS`. https://github.com/fermyon/spin/blob/09de9c38f6c9ca280020cd9afd5eb655651b8766/crates/templates/src/manager.rs#L569

itowlson (approved):

PR #2791, opened 2024-08-30
Fix the path for default key-value stores

rylev:
We were erroneously concatenating the base path on the default config.


lann:
If I'm following the code correctly it also looks like we're passing `state_dir` in as the `SpinKeyValueStore` `base_path` which I don't think is right? That should be the directory containing the runtime config toml.

We could really use some tests here...

rylev:
@lann I tested with the latest version of `spin`, and the base path for key-value stores is `.spin` so I believe the current behavior is correct.

lann:
```console
$ cat runtime_config.toml
[key_value_store.default]
type = "spin"
path = "./foo.db"

$ spin-270 --version
$ spin-270 up --runtime-config runtime_config.toml
[...]
Storing default key-value data to "./foo.db"

$ spin-pr-2791 --version
spin 2.8.0-pre0 (65c62d8b 2024-08-30)
$ spin-pr-2791 up --runtime-config runtime_config.toml
[...]
Storing default key-value data to ".spin/sqlite_key_value.db".
```

rylev:
@lann ah ok - I was indeed testing with a non-default db. 

So it seems that for Spin 2.0 the `path` in the runtime config has a different meaning depending on whether it's the default store or a custom store. If default, the path is relative to the runtime config file, and if it's a custom store, it's relative to the `.spin` directory. 

That seems... less than ideal. One reason I was convinced there wasn't a bug is because the behavior of always being relative to `.spin` seemed correct to me. I could buy the argument that `path` should always be relative to the runtime config file, but it seems obviously wrong to treat custom stores and default stores differently.

lann:
The 2.x behavior is:

- If there is no explicit `[key_value_store.default]` then the default path is `<state_dir>/sqlite_key_value.db`
- If there is an explicit `[key_value_store.default]` then the path is resolved relative to the runtime config file, just like for any other store

Its just that default implicit `path` that is special.

rylev:
@lann I believe I have addressed the issue. Fundamentally, the `KeyValueSqlite` `StoreManager` needs to be able to handle relative paths.

lann (approved):

PR #2792, opened 2024-08-30
factors: Remove InstanceFactors

lann:
This was the design used in an early iteration of `Factor::prepare` but I screwed up the lifetimes and it made `PrepareContext` hard to use so I split out the separate `InstanceFactors`. I realized my mistake today and fixed it here.

### DIFF
--- a/crates/factors/src/prepare.rs
+++ b/crates/factors/src/prepare.rs
@@ -39,42 +39,34 @@ impl<T: SelfInstanceBuilder> FactorInstanceBuilder for T {
 /// A PrepareContext is passed to [`Factor::prepare`].
 ///
 /// This gives the factor access to app state and the app component.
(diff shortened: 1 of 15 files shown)


crates/factors/src/prepare.rs
```diff
@@ -39,51 +39,43 @@ impl<T: SelfInstanceBuilder> FactorInstanceBuilder for T {
...
         Self {
             app_state,
             app_component,
+            instance_builders,
         }
     }
 
     /// Get the app state related to the factor.
-    pub fn app_state(&self) -> &F::AppState {
+    pub fn app_state(&self) -> &'a F::AppState {
```
lann, line 63: This is what I screwed up before.

rylev (approved):
Awesome stuff!

PR #2798, opened 2024-09-04
Prevent needlessly opening the default store

calebschoepp:
This is creating an extra span which is blocking my o11y work.

### DIFF
--- a/crates/runtime-config/src/lib.rs
+++ b/crates/runtime-config/src/lib.rs
@@ -3,7 +3,7 @@ use std::path::{Path, PathBuf};
 use anyhow::Context as _;
 use spin_common::ui::quoted_path;
 use spin_factor_key_value::runtime_config::spin::{self as key_value};
-use spin_factor_key_value::{DefaultLabelResolver as _, KeyValueFactor};
+use spin_factor_key_value::KeyValueFactor;
 use spin_factor_llm::{spin as llm, LlmFactor};
 use spin_factor_outbound_http::OutboundHttpFactor;
 use spin_factor_outbound_mqtt::OutboundMqttFactor;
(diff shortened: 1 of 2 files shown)


crates/runtime-config/src/lib.rs
```diff
@@ -156,33 +156,6 @@ where
...
-            .await
-            .expect("trigger was misconfigured and lacks a default store");
-        for (key, value) in iter {
-            store
-                .set(key, value.as_bytes())
-                .await
-                .context("failed to set key-value pair")?;
-        }
-        Ok(())
-    }
```
calebschoepp, line 184: @lann @rylev I'm not 100% confident this can be removed. It compiles without it and I can't find references to it. But, still not super sure.

lann: I'm 90% sure this is a refactoring artifact.

crates/trigger/src/cli/initial_kv_setter.rs
```diff
@@ -23,6 +23,9 @@ impl<F: RuntimeFactors, U> ExecutorHooks<F, U> for InitialKvSetterHook {
         &mut self,
         configured_app: &spin_factors::ConfiguredApp<F>,
     ) -> anyhow::Result<()> {
+        if self.kv_pairs.is_empty() {
+            return Ok(());
+        }
         let Some(kv) = configured_app.app_state::<KeyValueFactor>().ok() else {
             return Ok(());
         };
```
lann, line 31: It seems like this should actually be an error now... :shrug:

lann (approved):

crates/trigger/src/cli/initial_kv_setter.rs
lann: Maybe some context e.g. "you tried to set some kv but no kv!"

calebschoepp: Done

crates/runtime-config/src/lib.rs
rylev: I'm 100% sure this is just an artifact 😄

crates/trigger/src/cli/initial_kv_setter.rs
```diff
@@ -23,9 +23,12 @@ impl<F: RuntimeFactors, U> ExecutorHooks<F, U> for InitialKvSetterHook {
...
         configured_app: &spin_factors::ConfiguredApp<F>,
     ) -> anyhow::Result<()> {
-        let Some(kv) = configured_app.app_state::<KeyValueFactor>().ok() else {
+        if self.kv_pairs.is_empty() {
             return Ok(());
-        };
+        }
+        let kv = configured_app
+            .app_state::<KeyValueFactor>()
+            .context("attempted to set initial kv pairs but there is no kv")?;
```
rylev, line 31: This would only ever happen if the runtime is misconfigured so I think it's better to provide a more technical error message since only runtime embedders should see this (and end users should never encounter this).

```suggestion
            .context("attempted to set initial kv pairs but the key-value factor was not configured")?;
```

rylev (approved):

PR #2800, opened 2024-09-04
Move kv and llm instrumentation up a level of abstraction

calebschoepp:
This is to make it so we don't need to pass the observe context all the way down the stack.

### DIFF
--- a/crates/factor-key-value/src/host.rs
+++ b/crates/factor-key-value/src/host.rs
@@ -5,6 +5,7 @@ use spin_locked_app::MetadataKey;
 use spin_world::v2::key_value;
 use std::{collections::HashSet, sync::Arc};
 use table::Table;
+use tracing::{instrument, Level};
 
 pub const KEY_VALUE_STORES_KEY: MetadataKey<Vec<String>> = MetadataKey::new("key_value_stores");
 
@@ -79,6 +80,7 @@ impl key_value::Host for KeyValueDispatch {}
 
 #[async_trait]
 impl key_value::HostStore for KeyValueDispatch {
+    #[instrument(name = "spin_key_value.open", skip(self), err(level = Level::INFO), fields(otel.kind = "client", kv.backend=self.manager.summary(&name).unwrap_or("unknown".to_string())))]
     async fn open(&mut self, name: String) -> Result<Result<Resource<key_value::Store>, Error>> {
         Ok(async {
             if self.allowed_stores.contains(&name) {
@@ -94,6 +96,7 @@ impl key_value::HostStore for KeyValueDispatch {
         .await)
     }
 
+    #[instrument(name = "spin_key_value.get", skip(self, store), err(level = Level::INFO), fields(otel.kind = "client"))]
     async fn get(
         &mut self,
         store: Resource<key_value::Store>,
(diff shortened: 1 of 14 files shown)


calebschoepp:
@lann marking you for review.

crates/llm-remote-http/src/lib.rs
```diff
@@ -6,12 +6,11 @@ use reqwest::{
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use spin_world::v2::llm::{self as wasi_llm};
-use tracing::{instrument, Level};
 
 #[derive(Clone)]
 pub struct RemoteHttpLlmEngine {
     auth_token: String,
-    url: Url,
+    pub url: Url,
```
rylev, line 13: Nit: might be better to add a `fn url()` getter method instead since we don't really want people to write to this field.

crates/factor-key-value/src/host.rs
```diff
@@ -79,6 +80,7 @@ impl key_value::Host for KeyValueDispatch {}
 
 #[async_trait]
 impl key_value::HostStore for KeyValueDispatch {
+    #[instrument(name = "spin_key_value.open", skip(self), err(level = Level::INFO), fields(otel.kind = "client", kv.backend=self.manager.summary(&name).unwrap_or_default()))]
```
rylev, line 83: If the summary returns `None` do we really `kv.backend=""` or should it be something like `kv.backend="unknown"`?

crates/factor-llm/src/spin.rs
```diff
@@ -161,5 +169,9 @@ mod noop {
                 "Local LLM operations are not supported in this version of Spin.".into(),
             ))
         }
+
+        fn summary(&self) -> Option<String> {
+            None
```
rylev, line 174: Why not?

```suggestion
            Some("noop model".to_owned())
```

rylev (approved):
Overall looks good. I just had a few non-blocking questions.

crates/llm-remote-http/src/lib.rs
calebschoepp: Great idea. It did feel a little awkward just making this pub.

crates/factor-key-value/src/host.rs
calebschoepp: Yep, I like that better.

crates/factor-key-value/src/host.rs
```diff
@@ -79,6 +80,7 @@ impl key_value::Host for KeyValueDispatch {}
 
 #[async_trait]
 impl key_value::HostStore for KeyValueDispatch {
+    #[instrument(name = "spin_key_value.open", skip(self), err(level = Level::INFO), fields(otel.kind = "client", kv.backend=self.manager.summary(&name).unwrap_or("unknown".to_string())))]
```
lann, line 83: I'm not sure that we should use `summary` here. It could have pretty high cardinality (and/or be overly verbose) for some stores like sqlite in a tempdir. Maybe we should add an e.g. `Store::type_name` method?

lann: Alternatively just omit `kv.backend` entirely and see if its actually a problem in practice :shrug:

lann: The more I think about this the less I think its actually a bad idea. Feel free to ignore.

PR #2808, opened 2024-09-06
Give outbound http ability to ban private ips

rylev:
We still allow private IPs in Spin CLI, but we might add the ability in the future to turn this off. Other runtimes might also want to take advantage of this feature.

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -1,8 +1,9 @@
-use std::{error::Error, sync::Arc};
+use std::{error::Error, net::IpAddr, sync::Arc};
 
 use anyhow::Context;
 use http::{header::HOST, Request};
 use http_body_util::BodyExt;
+use ip_network::IpNetwork;
 use rustls::ClientConfig;
 use spin_factor_outbound_networking::OutboundAllowedHosts;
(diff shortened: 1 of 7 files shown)


crates/factor-outbound-http/src/wasi.rs
```diff
@@ -204,7 +208,21 @@ async fn send_request_handler(
...
+        socket_addrs.retain(|addr| !is_private_ip(addr.ip()));
+        if socket_addrs.is_empty() {
+            return Err(ErrorCode::DestinationIpProhibited);
+        }
+    }
+
+    let tcp_stream = timeout(connect_timeout, TcpStream::connect(socket_addrs.as_slice()))
         .await
         .map_err(|_| ErrorCode::ConnectionTimeout)?
         .map_err(|err| match err.kind() {
```
lann, line 228: Looks like we could remove the `.starts_with("failed to lookup address information")` silliness below now.

lann (approved):
It would be good to have a test for this.

rylev:
@lann I added a test. I don't love that the test relies on the actual system DNS resolver and is subject to being messed up if the local system is not what we expect (e.g., someone setting `localhost` to resolve to a non-local IP), but hopefully we don't run into issues.

lann (approved):

PR #2809, opened 2024-09-06
Make OutboundHttpInterceptor async

lann:


rylev (approved):

lann:
@rylev I haven't finished implementing interception for the spin interface but I did update `OutboundHttpInterceptor` to remove some details that were too specific to the wasi interface. Please check that the latest commit works for you.

rylev (approved):

PR #2810, opened 2024-09-06
Move instruments up a level in factor-variables

calebschoepp:


crates/factor-variables/src/spin_cli/azure_key_vault.rs
```diff
@@ -134,7 +133,6 @@ impl AzureKeyVaultProvider {
 
 #[async_trait]
 impl Provider for AzureKeyVaultProvider {
-    #[instrument(name = "spin_variables.get_from_azure_key_vault", skip(self), err(level = Level::INFO), fields(otel.kind = "client"))]
```
rylev, line 137: Does it make sense to completely remove these? We're losing the information of which providers were called.

rylev (changes requested):

crates/factor-variables/src/spin_cli/azure_key_vault.rs
lann: Maybe the top level could set e.g. `spin.variables.provider = Empty` and then these impls could just record that field?

calebschoepp: Are we guaranteed that only one provider will be called per variable?

lann: Though you are guaranteed that a provider that actually provides a variable will be the last called...

lann (approved):

rylev (approved):

PR #2817, opened 2024-09-10
Support interceptor in spin outbound http

lann:
- Update interceptor interface:
  - Move into new module
  - Add new `InterceptRequest` wrapper type to handle differences between spin and wasi requests
  - Make `intercept` request param owned to be returned in `InterceptOutcome::Continue` branch, replacing awkward `std::mem::take`

### DIFF
--- a/crates/factor-outbound-http/src/intercept.rs
+++ b/crates/factor-outbound-http/src/intercept.rs
@@ -1,0 +1,107 @@
+use http::{Request, Response};
+use http_body_util::{BodyExt, Full};
+use spin_world::async_trait;
+use wasmtime_wasi_http::{body::HyperOutgoingBody, HttpResult};
+
+pub type HyperBody = HyperOutgoingBody;
+
+/// An outbound HTTP request interceptor to be used with
+/// [`InstanceState::set_request_interceptor`].
+#[async_trait]
+pub trait OutboundHttpInterceptor: Send + Sync {
+    /// Intercept an outgoing HTTP request.
+    ///
+    /// If this method returns [`InterceptedResponse::Continue`], the (possibly
+    /// updated) request will be passed on to the default outgoing request
+    /// handler.
+    ///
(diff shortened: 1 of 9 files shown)


crates/factor-outbound-http/src/intercept.rs
```diff
@@ -0,0 +1,107 @@
+use http::{Request, Response};
+use http_body_util::{BodyExt, Full};
+use spin_world::async_trait;
+use wasmtime_wasi_http::{body::HyperOutgoingBody, HttpResult};
+
+pub type HyperBody = HyperOutgoingBody;
```
rylev, line 6: This seems a bit unnecessary? I don't see how `HyperBody` is clearer than `HyperOutgoingBody`

crates/factor-outbound-http/src/intercept.rs
```diff
@@ -0,0 +1,107 @@
...
+    body: InterceptBody,
+}
+
+enum InterceptBody {
+    Hyper(HyperBody),
+    Vec(Vec<u8>),
+}
+
+impl InterceptRequest {
+    pub fn into_hyper_request(self) -> Request<HyperBody> {
```
rylev, line 50: Can this also be `pub(crate)`?

crates/factor-outbound-http/src/intercept.rs
```diff
@@ -0,0 +1,107 @@
...
+
+pub type HyperBody = HyperOutgoingBody;
+
+/// An outbound HTTP request interceptor to be used with
+/// [`InstanceState::set_request_interceptor`].
+#[async_trait]
+pub trait OutboundHttpInterceptor: Send + Sync {
+    /// Intercept an outgoing HTTP request.
+    ///
+    /// If this method returns [`InterceptedResponse::Continue`], the (possibly
```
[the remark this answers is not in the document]
lann: You can still update it, just as owned-`mut` rather than `&mut`. :shrug:

PR #2818, opened 2024-09-10
Reimplement unsafe-aot-compilation feature

lann:


fibonacci1729 (approved):

crates/trigger/src/loader.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.aot_compilation_enabled = true;
+    }
+
+    #[cfg(feature = "unsafe-aot-compilation")]
+    fn load_precompiled_component(
+        &self,
+        engine: &wasmtime::Engine,
+        path: &std::path::Path,
+    ) -> anyhow::Result<Component> {
+        assert!(self.aot_compilation_enabled);
```
rylev, line 48: Is there a reason we're panicking here instead of using `anyhow::ensure!`?

lann: This is just a belt-and-suspenders regression assertion; this condition is unreachable with the current code.

PR #2821, opened 2024-09-12
Adding the `spin plugins inspect` command

ThorstenHans:
Allowing Spin users to print detailed information of plugins, before installing them on their machine

## Usage

```bash
spin plugins inspect otel
otel: 0.1.2 (License: Apache-2.0)
https://github.com/fermyon/otel-plugin

A plugin to make working with OTel in Spin easy
```

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -41,7 +41,7 @@ pub enum PluginCommands {
     Update,
 
     /// Print information about a plugin.
-    Inspect(Inspect),
+    Show(Show),
 }
 
 impl PluginCommands {
@@ -53,7 +53,7 @@ impl PluginCommands {
             PluginCommands::Uninstall(cmd) => cmd.run().await,
             PluginCommands::Upgrade(cmd) => cmd.run().await,
             PluginCommands::Update => update().await,
-            PluginCommands::Inspect(cmd) => cmd.run().await,
+            PluginCommands::Show(cmd) => cmd.run().await,
         }
     }
 }
@@ -421,12 +421,12 @@ impl Upgrade {
 }
 
 #[derive(Parser, Debug)]
-pub struct Inspect {
+pub struct Show {
     /// Name of Spin plugin.
     pub name: String,
 }
 
-impl Inspect {
+impl Show {
     pub async fn run(self) -> Result<()> {
         let manager = PluginManager::try_default()?;
         let manifest = manager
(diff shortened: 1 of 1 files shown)


src/commands/plugins.rs
```diff
@@ -416,6 +420,38 @@ impl Upgrade {
...
+
+        println!(
+            "{}: {} (License: {})\n{}\n\n{}",
+            manifest.name(),
+            manifest.version(),
+            manifest.license(),
+            manifest
+                .homepage_url()
+                .map(|u| u.to_string())
+                .unwrap_or_default(),
```
itowlson, line 448: Nit: This will leave an empty line if there's no homepage.  You could workaround this by moving the `\n` into the `map` a la `manifest.homepage_url().map(|u| format!("{u}\n")).unwrap_or_default()` I think.  It's not very important though.

src/commands/plugins.rs
```diff
@@ -39,6 +39,9 @@ pub enum PluginCommands {
 
     /// Fetch the latest Spin plugins from the spin-plugins repository.
     Update,
+
+    /// Print information about a plugin.
+    Inspect(Inspect),
```
itowlson, line 44: `inspect` feels like it oversells this command - I always assume I'm going to get detailed, even internal, information about a thing when I inspect it.  How would you feel about `info` or `show` instead?  Too vague?

itowlson (commented):
This seems okay except for bikeshedding the name, but I do wonder if the stated purpose would be better served by adding the info into the existing list and search commands - by the time I know enough about a plugin to choose it as a subject for this command, I probably already have a decent sense of what it does.  I don't have any objection to this going in, just a bit curious about how we see it fitting into a user journey.

src/commands/plugins.rs
ThorstenHans: I'm cool with renaming it to `info` or `show`. I was just using same vocabulary as `brew` does.

Will change it to `show`

ThorstenHans:
> I do wonder if the stated purpose would be better served by adding the info into the existing list and search commands

IMO, we should not update the output of list and/or search commands. Especially the description field could lead to messing up the output if not truncated. If we truncate the description, it would lead to users invoking `spin plugins show` either way.

src/commands/plugins.rs
ThorstenHans: updated the output accordingly

itowlson (approved):

PR #2822, opened 2024-09-12
Improve the logs and docs in outbound networking factor

rylev:
Adds some docs and logs, and removes some redundant logs.

### DIFF
--- a/crates/factor-outbound-networking/src/lib.rs
+++ b/crates/factor-outbound-networking/src/lib.rs
@@ -111,17 +111,20 @@ impl Factor for OutboundNetworkingFactor {
                 wasi_builder.outbound_socket_addr_check(move |addr, addr_use| {
                     let allowed_hosts = allowed_hosts.clone();
                     async move {
-                        // TODO: validate against existing spin-core behavior
                         let scheme = match addr_use {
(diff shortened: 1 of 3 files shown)


crates/factor-outbound-networking/src/lib.rs
```diff
@@ -111,17 +111,20 @@ impl Factor for OutboundNetworkingFactor {
...
                             SocketAddrUse::TcpBind => return false,
                             SocketAddrUse::TcpConnect => "tcp",
-                            SocketAddrUse::UdpBind | SocketAddrUse::UdpConnect | SocketAddrUse::UdpOutgoingDatagram => "udp",
+                            SocketAddrUse::UdpBind
+                            | SocketAddrUse::UdpConnect
+                            | SocketAddrUse::UdpOutgoingDatagram => "udp",
                         };
-                        allowed_hosts.check_url(&addr.to_string(), scheme).await.unwrap_or_else(|err| {
-                            // TODO: should this trap (somehow)?
-                            tracing::error!(%err, "allowed_outbound_hosts variable resolution failed");
```
rylev, line 122: We log the same thing inside of `check_url` so this ends up as a duplicate log.

PR #2823, opened 2024-09-12
Don't require a `SocketAddr` to make a `SelfRequestOrigin`

rylev:
This was forcing consumers who only had a DNS authority to resolve to an IP address ahead of time instead of just passing the authority along.

### DIFF
--- a/crates/factor-outbound-http/src/lib.rs
+++ b/crates/factor-outbound-http/src/lib.rs
@@ -4,7 +4,7 @@ mod wasi;
 pub mod wasi_2023_10_18;
 pub mod wasi_2023_11_10;
 
-use std::{net::SocketAddr, sync::Arc};
+use std::sync::Arc;
 
 use anyhow::Context;
 use http::{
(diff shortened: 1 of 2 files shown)


crates/factor-outbound-http/src/lib.rs
```diff
@@ -138,13 +138,12 @@ pub struct SelfRequestOrigin {
...
-    pub fn create(scheme: Scheme, addr: &SocketAddr) -> anyhow::Result<Self> {
+    pub fn create(scheme: Scheme, auth: &str) -> anyhow::Result<Self> {
         Ok(SelfRequestOrigin {
             scheme,
-            authority: addr
-                .to_string()
+            authority: auth
                 .parse()
-                .with_context(|| format!("address '{addr}' is not a valid authority"))?,
+                .with_context(|| format!("address '{auth}' is not a valid authority"))?,
```
lann, line 146: > `auth`

This abbreviation threw me off for a second :sweat_smile:

lann (approved):

PR #2826, opened 2024-09-12
Add experimental `spin up --component` flag to run a subset of app components

### LINKED ISSUE #2820: Enable spinning up only a subset of an application's components
## Summary
Say I have an app with 4 components but only want to run 2, I should be able to do the equivalent of `spin up --component "foo"  --component "bar"`

## Background

Spin unifies serverless application development. Developers can define distinct functions as components and group them all together into one fully-featured Spin application. Each of these components has unique requirements, such as access to a database or LLM. Not all of these requirements are always available to the developer or in production. For these reasons, developers should be able to run a subset of components of a Spin application.

## Proposed DevX
```sh
# Run all components 
spin up
# Run the `hello-rust` and `hello-go` components
spin up --component "hello-rust" --component "hello-go"
```

kate-goldenring:
fixes #2820

Say I have an app with 4 components but only want to run 2, I can do the equivalent of `spin up --component-id "foo"  --component-id "bar"`

Approach: Modifies the locked app to remove undesired components and triggers before the locked app is loaded. Takes somewhat hacky approach of creating a temporary `App` struct to pull out information mapping components to their triggers.


src/commands/up.rs
```diff
@@ -646,6 +668,76 @@ fn trigger_command_for_resolved_app_source(
...
+fn retain_components(locked_app: &mut LockedApp, component_trigger_id_map: ComponentTriggerMap) {
+    let components = component_trigger_id_map.keys().collect::<HashSet<_>>();
+    let triggers = component_trigger_id_map.values().collect::<HashSet<_>>();
+    locked_app.components.retain(|c| components.contains(&c.id));
+    locked_app.triggers.retain(|t| triggers.contains(&t.id));
+}
+
+// Given a list of components and App, returns a map of component ID to trigger ID.
+// Also validates that the components exist in the app and have valid service chaining.
+fn component_trigger_map(
```
kate-goldenring, line 682: We should consider moving this and `retain_components` to a crate that other runtimes can consume if they would like to do component filtering

src/commands/up.rs
```diff
@@ -164,9 +173,9 @@ impl UpCommand {
             .context("Could not canonicalize working directory")?;
 
         let resolved_app_source = self.resolve_app_source(&app_source, &working_dir).await?;
-
-        let trigger_cmds = trigger_command_for_resolved_app_source(&resolved_app_source)
-            .with_context(|| format!("Couldn't find trigger executor for {app_source}"))?;
+        let mut trigger_cmds =
+            trigger_command_for_resolved_app_source(resolved_app_source.trigger_types()?)
+                .with_context(|| format!("Couldn't find trigger executor for {app_source}"))?;
```
lann, line 178: Rather than doing this (potentially) twice, maybe this can be moved inside `if self.help` below and the second call to `trigger_command_for_resolved_app_source` below can move outside of that `if` block?

src/commands/up.rs
```diff
@@ -194,6 +203,23 @@ impl UpCommand {
             .load_resolved_app_source(resolved_app_source, &working_dir)
             .await?;
 
+        if let Some(components) = &self.components {
+            // Create a temporary app to access parsed component and trigger information
+            let tmp_app = spin_app::App::new("tmp", locked_app.clone());
+            let component_trigger_id_map = component_trigger_map(&tmp_app, components)?;
+            retain_components(&mut locked_app, component_trigger_id_map);
```
lann, line 210: Can these top two lines move into `retain_components`? i.e.

```suggestion
            retain_components(&mut locked_app, components);
```

src/commands/up.rs
```diff
@@ -35,6 +36,9 @@ const MULTI_TRIGGER_START_OFFSET: tokio::time::Duration = tokio::time::Duration:
 // any exited" check.
 const MULTI_TRIGGER_LET_ALL_START: tokio::time::Duration = tokio::time::Duration::from_millis(500);
 
+// Map of component ID to trigger ID
+type ComponentTriggerMap = HashMap<String, String>;
```
lann, line 40: This really should be e.g. `HashMap<String, HashSet<String>>`; one component can have multiple triggers pointing at it.

src/commands/up.rs
```diff
@@ -646,6 +668,76 @@ fn trigger_command_for_resolved_app_source(
...
+                    None
+                }
+            }
+            Err(_) => None,
+        })
+        .collect();
+
+    // Allow only service chaining among retained components
+    // All wildcard service chaining is disallowed
+    for (c, _) in &component_triggers {
```
lann, line 702: :shrug: 
```suggestion
    for c in component_triggers.keys() {
```

src/commands/up.rs
```diff
@@ -646,6 +668,76 @@ fn trigger_command_for_resolved_app_source(
...
+        let allowed_hosts = allowed_hosts(c)?;
+        allowed_hosts.iter().try_for_each(|host| {
+            let uri = host.parse::<http::Uri>().unwrap();
+            if let Some(local_component) = parse_service_chaining_target(&uri) {
+                if !components.contains(&local_component) {
+                    if local_component == "*" {
+                        bail!("Wildcard service chaining not allowed in retained components");
```
lann, line 709: I think this could be wordsmithed to be a little more clear; maybe something like:

```
Component selected with '--component {id}' cannot use wildcard service chaining:
  allowed_outbound_hosts = ["http://*.spin.internal"]
```

src/commands/up.rs
```diff
@@ -646,6 +668,76 @@ fn trigger_command_for_resolved_app_source(
...
+                    }
+                    bail!(
+                        "Service chaining to non-retained component: {}",
```
lann, line 712: Maybe:

```
Component selected with '--component {id}' cannot use service chaining to unselected component:
  allowed_outbound_hosts = ["http://{other}.spin.internal"]
```

src/commands/up.rs
kate-goldenring: Good catch. Thank you!

src/commands/up.rs
kate-goldenring: This is much clearer. Thank you!

src/commands/up.rs
kate-goldenring: At this point it is a vector not a HashMap. I agree that this looks a little wonky, but we need the full `AppComponent` object at this point so it is too early to convert it to a `ComponentTriggerMap`.

lann: yeah nbd

src/commands/up.rs
itowlson: That message seems less clear to me: it is no longer clear to me that the 'selection' is the _cause_ of not being able to use the wildcard service chaining.  I read it as "this component, which is included in the selection, can't use wildcard service chaining."

lann: Yeah its definitely a little awkward; alternative wordsmithing welcome!

lann: I think the problem with the original wording is that "wildcard service chaining" and "retained components" are not necessarily familiar concepts.

kate-goldenring: >  I read it as "this component, which is included in the selection, can't use wildcard service chaining."

I think this is an accurate reading. No retained component is allowed to use wildcard service chaining. They can only use precise/targeted service chaining

itowlson: @kate-goldenring Ironically, I poorly worded my comment about the wording.  What I read was that "this component can't use wildcard service chaining (for some reason)" (and Spin needs to tell you about this because this component is selected).  Your original wording made it clear that the problem is that wildcard service chaining isn't compatible with selection.

How about something like:

"You cannot use the `*.spin.internal` wildcard host in combination with the `--component` flag. Change the `allowed_outbound_hosts` for component 'isometric-darwin' to list specific components for service chaining, or remove the `--component` flag."

(This is wordy for sure, and may be too specific to CLI syntax, I'm just offering it as a sighting shot.)

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+            not_found.push(c);
+        }
+    });
+    ensure!(
+        not_found.is_empty(),
+        "Specified components not found in application: {:?}",
+        not_found
+    );
+
+    Ok(component_trigger_map)
```
itowlson, line 748: I got quite lost in this function I'm afraid.  Conceptually the idea seems to be "remove all the components that aren't in the list, then remove all the triggers whose components are no longer present."  (Or, equivalently, "create two remove lists, one of all the components that aren't in the list, and one of all the triggers whose components aren't in the list.")  I didn't quite get where the mapping came in, or why the service chaining validation and UI sanity checking were intermingled.  If this is the best way to do it then fair enough, it just seemed like it could be simpler but I could well be missing a lot of nuance or edge cases!

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
         .collect()
 }
 
+// Scrubs the locked app to only contain the given list of components
+// Introspects the LockedApp to find and selectively retain the triggers that correspond to those components
+fn retain_components(locked_app: &mut LockedApp, components: &[String]) -> Result<()> {
+    // Create a temporary app to access parsed component and trigger information
+    let tmp_app = spin_app::App::new("tmp", locked_app.clone());
```
itowlson, line 674: I feel like `up` should not have to create an `App` just to look at the triggers and components.  It seems like this should be possible to get from the `LockedApp` but maybe not (I know the loose typing can introduce a bit of faff there...).  Or maybe App and LockedApp have largely converged at this point - I see the factors work has removed the dependency from spin-app to spin-core which was always my concern in the past - maybe App is just a helpful wrapper around LockedApp now?

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
         .collect()
 }
 
+// Scrubs the locked app to only contain the given list of components
+// Introspects the LockedApp to find and selectively retain the triggers that correspond to those components
+fn retain_components(locked_app: &mut LockedApp, components: &[String]) -> Result<()> {
```
itowlson, line 672: Am I right in remembering that Wasm files have to exist in order to build a LockedApp?  That is, suppose I did a `spin up` where I excluded `component-xyz`: would it still require me to have `component_xyz.wasm`?

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+            Err(_) => None,
+        })
+        .collect();
+
+    // Allow only service chaining among retained components
+    // All wildcard service chaining is disallowed
+    for (c, _) in &component_triggers {
+        let allowed_hosts = allowed_hosts(c)?;
+        allowed_hosts.iter().try_for_each(|host| {
+            let uri = host.parse::<http::Uri>().unwrap();
```
itowlson, line 711: I think you might have a problem here with templated URIs?  I believe they're resolved during trigger startup, unless the factors work has changed that.  Might merit a case in the unit tests.

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+                        "Component selected with '--component {}' cannot use service chaining to unselected component:
+  allowed_outbound_hosts = [\"http://{}.spin.internal\"]",
+                        c.id(), local_component
+                    );
+                }
+            }
+            Ok(())
+        })?;
+    }
+    let mut component_trigger_map: ComponentTriggerMap = HashMap::new();
```
itowlson, line 728:
```suggestion
    let mut component_trigger_map = ComponentTriggerMap::new();
```

itowlson (commented):
I think these are mostly nits or "is this the best way" kibbitzing - I do think it's worth adding a test case for templated hosts though.  I love that you were able to bring this off so simply!

src/commands/up.rs
lann: We can't (currently) resolve templates here, so I guess there are two questions:

- Will this code panic if it tries to resolve a template? (seems likely to me)

- What _should_ happen with templates here? I would probably suggest doing nothing. Using templates for service chaining seems like an uncommon scenario and the consequence here is pretty low: runtime error instead of this nicer startup validation.

itowlson: Yes, the current code will panic on an unsubstituted template.

I agree that ignoring the error is likely best.  Service chaining via a templated URL is one of those hazy features.  It _happens_ to work in the current CLI implementation, and I don't believe we complain if someone does it, but there are no guarantees around it.  So it seems reasonable for a failed URL parse to be interpreted as "it's not a service chaining URL" which means it is of no further interest.

If we wanted more belt and braces we could validate during trigger load (after template substitution) that all non-wildcard service chaining URLs pointed to components that exist in the app.  Which might not be a bad plan anyway, and can be done outside the scope of this PR.

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -42,6 +42,12 @@ impl OutboundNetworkingFactor {
     }
 }
 
+pub fn allowed_hosts(component: &AppComponent<'_>) -> anyhow::Result<Vec<String>> {
```
rylev, line 45: Nit: can you move this function down? Having a helper function like this between two `impl` blocks makes the code harder to read IMO. 

Also, can we just add a doc comment for what this function does to make IDE users lives nicer?

src/commands/up.rs
```diff
@@ -35,6 +36,9 @@ const MULTI_TRIGGER_START_OFFSET: tokio::time::Duration = tokio::time::Duration:
 // any exited" check.
 const MULTI_TRIGGER_LET_ALL_START: tokio::time::Duration = tokio::time::Duration::from_millis(500);
 
+// Map of component ID to trigger ID
```
rylev, line 39:
```suggestion
/// Map of component ID to trigger IDs
```

src/commands/up.rs
```diff
@@ -630,11 +654,8 @@ fn trigger_command(trigger_type: &str) -> Vec<String> {
     vec!["trigger".to_owned(), trigger_type.to_owned()]
 }
 
-fn trigger_command_for_resolved_app_source(
-    resolved: &ResolvedAppSource,
-) -> Result<Vec<Vec<String>>> {
-    let trigger_type = resolved.trigger_types()?;
-    trigger_type
+fn trigger_command_for_resolved_app_source(trigger_types: Vec<&str>) -> Result<Vec<Vec<String>>> {
```
rylev, line 657: Is the name of this function still correct?

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+fn retain_components(locked_app: &mut LockedApp, components: &[String]) -> Result<()> {
+    // Create a temporary app to access parsed component and trigger information
+    let tmp_app = spin_app::App::new("tmp", locked_app.clone());
+    let component_trigger_id_map = component_trigger_map(&tmp_app, components)?;
+    let components = component_trigger_id_map.keys().collect::<HashSet<_>>();
+    let triggers = component_trigger_id_map
+        .values()
+        .flatten()
+        .collect::<HashSet<_>>();
+    locked_app.components.retain(|c| components.contains(&c.id));
```
rylev, line 681: Why collect `components` into a hashset instead of using [`component_trigger_id_map.contains_key`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.contains_key)?

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
         .collect()
 }
 
+// Scrubs the locked app to only contain the given list of components
+// Introspects the LockedApp to find and selectively retain the triggers that correspond to those components
```
rylev, line 671: All of these functions should have doc comments `///` instead of plain comments.

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+        .triggers()
+        .filter_map(|t| match t.component() {
+            Ok(comp) => {
+                if components.contains(&comp.id().to_string()) {
+                    Some((comp, t.id().to_string()))
+                } else {
+                    None
+                }
+            }
+            Err(_) => None,
```
rylev, line 702:
```suggestion
            Ok(comp) if components.contains(&comp.id().to_string()) => {
                    Some((comp, t.id().to_string()))
               }
               _ => None,
```

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+// Also validates that the components exist in the app and have valid service chaining.
+fn component_trigger_map(
+    app: &spin_app::App,
+    components: &[String],
+) -> Result<ComponentTriggerMap> {
+    let component_triggers: Vec<(AppComponent<'_>, String)> = app
+        .triggers()
+        .filter_map(|t| match t.component() {
+            Ok(comp) => {
+                if components.contains(&comp.id().to_string()) {
```
rylev, line 696: Doesn't `id()` return `&str` making the `to_string()` unnecessary?

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+                    Some((comp, t.id().to_string()))
+                } else {
+                    None
+                }
+            }
+            Err(_) => None,
+        })
+        .collect();
```
rylev, line 704: Looks like you're just iterating after this which makes the call to `collect` here unnecessary.

tests/integration.rs
```diff
@@ -348,12 +348,7 @@ mod integration_tests {
             },
         )?;
 
-        let expected = r#"Error: Couldn't find trigger executor for local app "spin.toml"
-
-Caused by:
-      no triggers in app
-"#;
-
+        let expected = "Error: No triggers in app\n";
```
rylev, line 351: This error might be a bit too terse. I'm not sure what a better wording would be though...

src/commands/up.rs
```diff
@@ -189,10 +197,27 @@ impl UpCommand {
...
         let mut locked_app = self
             .load_resolved_app_source(resolved_app_source, &working_dir)
-            .await?;
+            .await
+            .context("Failed to load application")?;
+        if let Some(components) = &self.components {
+            retain_components(&mut locked_app, components)?;
+        }
+
+        // Remove duplicate triggers
```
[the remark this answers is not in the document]
kate-goldenring: I am commenting on how I am collecting the triggers in a hashset in order to remove duplicates. May be unnecessary and distracting and best to remove.

src/commands/up.rs
kate-goldenring: I am pretty sure the wasm files do not have to exist; rather it builds the lockfile based on the manifest and where it says the file should be. For example, in this test that passes, the wasm file doesn't exist: https://github.com/kate-goldenring/spin/blob/67af2d3e2fbdce7c9113fa3c0fa74e4ff8151ee2/src/commands/up.rs#L778-L779

src/commands/up.rs
kate-goldenring: I am collecting so that I can iterate by reference to check service chaining but still later assemble to component trigger map without the iterator having been absorbed already. I am sure there is a better way

src/commands/up.rs
kate-goldenring: Good call out. This was legacy from a couple iterations ago.

src/commands/up.rs
kate-goldenring: I agree that it takes on two purposes: 
1. Get a mapping of component ID to trigger ID
2. Ensure that component retention request is valid

I clumped it together in order to reduce iterations but i think if i make the component trigger map hold the `AppComponent` then i could make step 2 a subsequent step

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
         .collect()
 }
 
+// Scrubs the locked app to only contain the given list of components
+// Introspects the LockedApp to find and selectively retain the triggers that correspond to those components
+fn retain_components(locked_app: &mut LockedApp, components: &[String]) -> Result<()> {
+    // Create a temporary app to access parsed component and trigger information
+    let tmp_app = spin_app::App::new("tmp", locked_app.clone());
+    let component_trigger_id_map = component_trigger_map(&tmp_app, components)?;
```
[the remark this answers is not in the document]
kate-goldenring: This was a good call out. I cleaned up the flow to not need the map

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+                }
+            }
+            Err(_) => None,
+        })
+        .collect();
+
+    // Allow only service chaining among retained components
+    // All wildcard service chaining is disallowed
+    for (c, _) in &component_triggers {
+        let allowed_hosts = allowed_hosts(c)?;
```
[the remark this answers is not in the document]
kate-goldenring: Moved each of the checks (service chaining and component exists) into their own function

src/commands/up.rs
```diff
@@ -646,6 +667,87 @@ fn trigger_command_for_resolved_app_source(
...
+    locked_app.components.retain(|c| components.contains(&c.id));
+    locked_app.triggers.retain(|t| triggers.contains(&t.id));
+    Ok(())
+}
+
+// Given a list of components and App, returns a map of component ID to trigger ID.
+// Also validates that the components exist in the app and have valid service chaining.
+fn component_trigger_map(
+    app: &spin_app::App,
+    components: &[String],
```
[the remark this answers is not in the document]
kate-goldenring: Added "retain" to the argument name where used

src/commands/up.rs
kate-goldenring: Reorganized the code to make this cleaner and clearer. Let me know what you think

src/commands/up.rs
kate-goldenring: Good call out here. I agree that we should ignore URLs that cannot be parsed for now and +1 to validating non-wildcard service chaining URLs are to existing components.

src/commands/up.rs
kate-goldenring: Nice! What do you call this kind of expression resolution?

src/commands/up.rs
kate-goldenring: Cleaned this up!

src/commands/up.rs
kate-goldenring: Updated

src/commands/up.rs
itowlson: @kate-goldenring I've usually seen them called "match guards" if that's the bit you're referring to? https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#extra-conditionals-with-match-guards

kate-goldenring:
@itowlson I added a test to validate templated hosts are ignored/allowed

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -145,6 +141,13 @@ impl Factor for OutboundNetworkingFactor {
     }
 }
 
+/// Returns the allowed hosts for a component.
```
itowlson, line 144: maybe just to be absolutely explicit:

```suggestion
/// Returns the allowed host strings for a component.
/// This does not resolve templated hosts.
```

or is that too much?

src/commands/up.rs
```diff
@@ -646,6 +663,81 @@ fn trigger_command_for_resolved_app_source(
...
+                            bail!("Component selected with '--component {}' cannot use wildcard service chaining: allowed_outbound_hosts = [\"http://*.spin.internal\"]", component.id());
+                        }
+                        bail!(
+                            "Component selected with '--component {}' cannot use service chaining to unselected component: allowed_outbound_hosts = [\"http://{}.spin.internal\"]",
+                            component.id(), local_component
+                        );
+                    }
+                }
+            }
+                Ok(())
```
itowlson, line 715: I'm not sure if this is a GitHub diff artefact but something seems odd in the indenting here - but if it looks sensible in the real code then ignore me!

src/commands/up.rs
```diff
@@ -646,6 +663,81 @@ fn trigger_command_for_resolved_app_source(
...
+) -> Result<()> {
+    app
+        .triggers().try_for_each(|t| {
+        if let Ok(component) = t.component() {
+            if retained_components.contains(&component.id().to_string()) {
+            let allowed_hosts = allowed_hosts(&component).context("failed to get allowed hosts")?;
+            allowed_hosts.iter().try_for_each(|host| {
+                // Templated URLs are not resolved at this point, so ignore unresolvable URIs
+                if let Ok(uri) = host.parse::<http::Uri>() {
+                if let Some(local_component) = parse_service_chaining_target(&uri) {
```
itowlson, line 703:
```suggestion
                if let Some(chaining_target) = parse_service_chaining_target(&uri) {
```

to help differentiate which component is in which role

itowlson (approved):
Thanks for all the changes @kate-goldenring - this reads really clearly to me!  I left some minor suggestions but LGTM

kate-goldenring:
@itowlson @rylev thoughts on moving `retain_components` to the spin-locked crate or spin-app crate? The shim will want to use it. It uses spin-locked types and functions from spin-app imports so regardless would want to export it from there. I could see it being a method on a `LockedApp`

kate-goldenring:
@itowlson what are your thoughts on:
```sh
spin up --component "foo" --component "bar"
```
vs 
```sh
spin up --components "foo,bar,bap"
```

I think i prefer the latter though it is more prone to parsing errors

itowlson:
@kate-goldenring The comma separated form _should_ be safe to parse, because component IDs can't contain tricksy characters, but even as I write that I hear my ghost gloating "famous last words."  I do have reservations about it requiring knowledge (what is the separator), though, and we use the "multiple occurrences" form for most other things.

My suggestion would be to go with multiple occurrences for now, and see if it is used by human users (as opposed to deployment scripts) often enough for people to complain about the verbosity - if so we can add support for a CSV form as well - how about that?

itowlson:
Re moving it to `spin-app` or `spin-locked-app`, it feels awkward dragging `http` and `factor-outbound-networking` dependencies into those crates.  They feel like largely less-opinionated and more-opinionated schema crates right now.  It kind of feels closer to a `spin-loader` behaviour, but the loader is specific to local now, so I'm not sure what a good home would be - maybe the awkwardness is the least worst option.

itowlson:
@kate-goldenring I just remembered we do have precedent for this in `spin build -c`, which uses multiple occurrences, which reinforces my preference for sticking with that.

I did notice that the `build` flag is named `-c/--component-id` instead of `--component` - I wonder if we should settle on one preferred option, and alias for back compat if necessary.  Although that in turn makes me unsure how we would parse `spin build --up --component admin` if we used the same flag name!  What do you reckon?

src/commands/up.rs
michelleN: +1 chaining_target reads nice

src/commands/up.rs
```diff
@@ -113,6 +114,11 @@ pub struct UpCommand {
     #[clap(long, takes_value = false, env = ALWAYS_BUILD_ENV)]
     pub build: bool,
 
+    /// Specific component to run. Can specify multiple. If omitted, all
+    /// components are run.
+    #[clap(hide = true, long = "component")]
```
michelleN, line 119: Is `hide=true` here because this is experimental?

radu-matei:
+1 to @itowlson's comment about the `spin build` flag that controls the same behaviour being named `--component-id`.

radu-matei (approved):
Tested this with applications with several triggers and components and the behaviour LGTM!

Thanks!

src/commands/up.rs
kate-goldenring: Yeah. I am not sure if there is another way we've marked features as experimental in the past. Should i update the comment to say it is experimental too?

crates/factor-outbound-networking/src/lib.rs
kate-goldenring: Always a fan of more verbosity. Changing

src/commands/up.rs
kate-goldenring: I am surprised the formatter didn't update the indentation of this block. I agree it could use some help

kate-goldenring:
> I did notice that the build flag is named -c/--component-id instead of --component - I wonder if we should settle on one preferred option, and alias for back compat if necessary. 

@itowlson thank you for noticing this. I think we should similarly name the flag `--component-id` and then we can always add the alias for component later.

> Although that in turn makes me unsure how we would parse spin build --up --component admin if we used the same flag name! What do you reckon?

In that case, i think we should ideally apply the flag to both commands. I am not sure if that is possible though

kate-goldenring:
@itowlson ff7feef commit brings in changes to enable running `spin build --up --component-id foo` which will cause only the foo component to be built and run. What do you think?

src/commands/up.rs
michelleN: yea not sure what the precedent is but I think a comment would be helpful.

itowlson:
@kate-goldenring I am honestly not sure what the right experience is with `spin build --component-id foo --up`.  It feels like two separate concerns are colliding: minimising rebuilds, and running subsets.  Looking back at when the feature was introduced (https://github.com/fermyon/spin/pull/1515) it seems the original consumer of `build --component-id` was `spin doctor` - so I don't really know if it's something people actually use manually.  But if they are then they might be surprised that they now only get a subset of their application.

I guess whatever we choose will be surprising to some people, but they always have the get-out clause of `spin build --whatever && spin up --whatever`.

Perhaps for Spin 3 we should enable `spin build --up -- --up-arg-1 --up-arg-2` and deprecate inlining the up-args into the build-args, with a view to phasing that out in Spin 4.  That's out of scope for this PR for sure though!

Sorry for the long and indecisive ramble...!

kate-goldenring:
@lann @rylev I think i have addressed your comments. This is ready for a final review

michelleN (approved):

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -1,8 +1,6 @@
 mod config;
 pub mod runtime_config;
 
-use std::{collections::HashMap, sync::Arc};
```
rylev, line 4: Nit: I don't think it's a hard rule, but we tend to keep `std` uses separate from external crates (which are both separate from uses local to the crate). It's not blocking, but I would consider reverting this.

src/commands/up.rs
```diff
@@ -646,6 +662,81 @@ fn trigger_command_for_resolved_app_source(
...
+
+/// Validates that all service chaining of an app will be satisfied by the retained components.
+/// All wildcard service chaining is disallowed
+fn validate_retained_components_service_chaining(
+    app: &spin_app::App,
+    retained_components: &[String],
+) -> Result<()> {
+    app
+        .triggers().try_for_each(|t| {
+        if let Ok(component) = t.component() {
```
rylev, line 696: Nit: You can reduce some nesting by using `let else`:

```rust
let Ok(component) = t.component() else  { return Ok(()) };
```

src/commands/up.rs
```diff
@@ -189,10 +193,25 @@ impl UpCommand {
...
             app_source.build().await?;
         }
-
         let mut locked_app = self
             .load_resolved_app_source(resolved_app_source, &working_dir)
-            .await?;
+            .await
+            .context("Failed to load application")?;
+        if !self.components.is_empty() {
+            retain_components(&mut locked_app, &self.components)?;
```
rylev, line 201: I would add some additional context to this function and some of the subfunctions with `.context()`. Right now an error from this function might simply read "failed to get allowed hosts" which is highly confusing without the context of the code.

src/commands/up.rs
```diff
@@ -646,6 +662,81 @@ fn trigger_command_for_resolved_app_source(
...
+                                "Component selected with '--component {}' cannot use service chaining to unselected component: allowed_outbound_hosts = [\"http://{}.spin.internal\"]",
+                                component.id(), chaining_target
+                            );
+                        }
+                    }
+                }
+                Ok(())
+            })?;
+        }}
+        Ok(()) as Result<(), anyhow::Error>
```
rylev, line 717:
```suggestion
        anyhow::Ok(())
```

src/commands/up.rs
```diff
@@ -646,6 +662,81 @@ fn trigger_command_for_resolved_app_source(
...
+fn validate_retained_components_service_chaining(
+    app: &spin_app::App,
+    retained_components: &[String],
+) -> Result<()> {
+    app
+        .triggers().try_for_each(|t| {
+        if let Ok(component) = t.component() {
+            if retained_components.contains(&component.id().to_string()) {
+            let allowed_hosts = allowed_outbound_hosts(&component).context("failed to get allowed hosts")?;
+            allowed_hosts.iter().try_for_each(|host| {
```
rylev, line 699: Nit: `for_each` and `try_for_each` aren't super common in my experience. This code would be a bit more readable IMO if `for in` were used instead.

src/commands/up.rs
```diff
@@ -646,6 +662,81 @@ fn trigger_command_for_resolved_app_source(
...
+                // Templated URLs are not resolved at this point, so ignore unresolvable URIs
```
rylev, line 700: Does this mean we might allow some components that we shouldn't?

src/commands/up.rs
```diff
@@ -646,6 +662,81 @@ fn trigger_command_for_resolved_app_source(
...
+/// Validates that all components specified to be retained actually exist in the app
+fn validate_retained_components_exist(
+    app: &spin_app::App,
+    retained_components: &[String],
+) -> Result<()> {
+    let app_components = app
+        .components()
+        .map(|c| c.id().to_string())
+        .collect::<HashSet<_>>();
+    retained_components.iter().try_for_each(|c| {
```
rylev, line 732: Same note as above on the use of `try_for_each`

rylev (changes requested):
I mostly have nits (feel free to ignore them if you like), but I am unsure about whether the service chaining check is correct.

crates/factor-outbound-networking/src/lib.rs
kate-goldenring: I'm not sure i follow. Are you saying to not use these dependencies, not use them in the root of the crate or change the formatting somehow? Other factors use std libs so i am assuming that is not it: https://github.com/kate-goldenring/spin/blob/6d29d49d5b2d9aad5b13f4c95602a2c4a77b16e8/crates/factor-key-value/src/util.rs#L5-L6

src/commands/up.rs
kate-goldenring: It means that we may spin up something that would later fail to execute. For example, say you have a spin app with 3 components, foo, bar and baz. You run `spin up --component-id "foo" --component-id "baz"` and foo is  configured with `allowed_outbound_hosts = [ "https://{{ myvar }}.spin.internal"]`. We know we want to retain foo and baz but we also want to check to make sure neither do internal service chaining to bar so that we can catch that error and fail the run. However, we cannot parse that host, so we cannot whether it is service chaining and even if we could we couldn't determine what component this is referencing, so we continue. 

No extra components are run but components may be run that cannot service chain. We discuss this a bit in this thread https://github.com/fermyon/spin/pull/2826#discussion_r1757766131

crates/factor-outbound-networking/src/lib.rs
rylev: Sorry - this is all about formatting of the `use` statements. Typically `use` statements are grouped into three groups:
* `std` lib uses (e.g., std::fmt::Display)
* external crate uses (e.g., `tokio::task::spawn`) 
* uses local to the crate (e.g., `crate::foo`)

These three groups are then separated by an empty line. 

There are a million exceptions, and it's not really important, so I only bring it up since you moved this line for seemingly no other reason than aesthetics. Feel free to ignore my comment 😄

src/commands/up.rs
rylev: Ah! Sorry I was reading the logic backwards. Sounds good to me!

One nit that might make this less confusing to readers in the future. We could add a comment outlining that we're doing a _best effort_ lookup of components that are allowed to be accessed through service chaining, and we try to error _early_ if a component tries to chain to another component that is not retained.

src/commands/up.rs
```diff
@@ -646,6 +662,81 @@ fn trigger_command_for_resolved_app_source(
...
+    retained_components: &[String],
+) -> Result<()> {
+    app
+        .triggers().try_for_each(|t| {
+        if let Ok(component) = t.component() {
+            if retained_components.contains(&component.id().to_string()) {
+            let allowed_hosts = allowed_outbound_hosts(&component).context("failed to get allowed hosts")?;
+            allowed_hosts.iter().try_for_each(|host| {
+                // Templated URLs are not resolved at this point, so ignore unresolvable URIs
+                if let Ok(uri) = host.parse::<http::Uri>() {
```
rylev, line 701: Allowed host configs are very often not valid Uris but are still statically resolvable. For example, `*://{foo.spin.internal, bar.spin.internal}`. This check wouldn't run because the above cannot be parsed as an `http::Uri`. 

You might want to consider `AllowedHostConfig::parse` instead. This handles all of the interpolation syntax that allowed host configs are allowed to have. You can then add a new method like `AllowedHostConfig::service_chaining_target` that returns a `Vec<String>` with all of the components that that `AllowedHostConfig` targets.

rylev (approved):
I still have some suggestions for improvements, but I don't want to block merging this any more!

src/commands/up.rs
kate-goldenring: I didn't know you could chain to multiple targets in one host `*://{foo.spin.internal, bar.spin.internal}`. Should we update the docs on this: https://developer.fermyon.com/spin/v2/http-outbound#local-service-chaining?

src/commands/up.rs
kate-goldenring: Good call out. I elaborated in the comment to make the best effort clearer.

PR #2834, opened 2024-09-13
Upgrade OTel

calebschoepp:


crates/telemetry/src/logs.rs
```diff
@@ -106,7 +101,9 @@ pub(crate) fn init_otel_logging_backend(spin_version: String) -> anyhow::Result<
         )
         .build();
 
-    global::set_logger_provider(provider);
+    LOGGER
+        .set(provider.logger("spin"))
+        .expect("should set logger");
```
rylev, line 106: Do we really want to panic here?

crates/telemetry/src/logs.rs
```diff
@@ -30,20 +29,16 @@ fn app_log_to_otel(buf: &[u8]) {
         return;
     }
 
-    let logger = global::logger_provider().logger("spin");
+    let logger = LOGGER.get().unwrap();
```
rylev, line 32: Is panicking here the right way to go? Or would it make sense to return early if the logger isn't set (and perhaps log a warning)?

calebschoepp: If we `.expect()` when we set the logger than I think we should unwrap here.

Otherwise if we allow ourselves to fail to set a logger than I think we should do what you said here and return early. The annoying thing then is we should probably do the thing where you log at warn once about this failing and then at trace the rest of the time b/c this is going to be very noisy.

crates/telemetry/src/logs.rs
calebschoepp: My assumption is that this should never fail because this is the only place we're setting the logger. I also assume that if this does fail then we don't want to continue b/c something weird is going on and we're not going to have an OTel logger.

What we do here affects how we handle the other place you left a comment.

Personally unless there is a clear reason setting a `OnceLock` would fail, I'd prefer to panic here.

crates/telemetry/src/logs.rs
lann: I agree this probably shouldn't panic. I am ambivalent as to whether it needs to log instead given that this will log to tracing perfectly happily even if this fails.

crates/telemetry/src/logs.rs
rylev: It panics if it's called more than once not only if it's called concurrently. Or are we checking just before this that the value isn't already set?

calebschoepp:
FYI this is blocking me from marking #2787 as ready to review. I need to rebase it on these changes first.

crates/telemetry/src/logs.rs
lann: > It panics if it's called more than once not only if it's called concurrently.

Oh sure enough; I misread the docs.

lann: Given :point_up: I'd suggest just ignoring a failure here rather than panicking :shrug:

calebschoepp: Fixed

lann (approved):

PR #2836, opened 2024-09-16
Fix `watch` with workdir not working on Windows

### LINKED ISSUE #2835: `spin watch` not watching on Windows
Using the same test repo as in https://github.com/fermyon/spin/issues/2829, if I run `spin watch` and then edit a file under `apps/admin/src`, the admin build command does not run.

itowlson:
Fixes #2835.

(cced to Ryan because a Certified Rust Wizard might know a better way of handling this, because mine sure feels kludgy)

### DIFF
--- a/src/commands/watch/filters.rs
+++ b/src/commands/watch/filters.rs
@@ -125,7 +125,7 @@ fn create_source_globs(cid: &str, c: &v2::Component) -> Option<Vec<String>> {
             build
                 .watch
                 .iter()
-                .filter_map(|w| Path::new(workdir).join(w).to_str().map(String::from))
+                .map(|w| concatenate_glob_friendly(workdir, w))
                 .collect()
         })
         .unwrap_or_else(|| build.watch.clone());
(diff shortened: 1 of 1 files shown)


src/commands/watch/filters.rs
```diff
@@ -137,6 +137,18 @@ fn create_source_globs(cid: &str, c: &v2::Component) -> Option<Vec<String>> {
     }
 }
 
+fn concatenate_glob_friendly(workdir: &str, watch_glob: &str) -> String {
+    // Using Path::join on Windows correctly joins with a backslash. But the watchexec glob
```
rylev, line 141: Nit: I'd rather information like this live as a doc comment on the function itself. That makes things like hover in IDEs work so that readers are forced to always come to the function definition should they want to learn more.

src/commands/watch/filters.rs
```diff
@@ -137,6 +137,18 @@ fn create_source_globs(cid: &str, c: &v2::Component) -> Option<Vec<String>> {
...
+    // engines hates backslashes. So we must always use forward slashes in glob expressions,
+    // and in workdirs, and for concatenating the two. (We could use Path::join on Unix but
+    // the signature ends up a bit different because to_str returns an Option, so...)
+    if workdir.ends_with('/') {
```
rylev, line 145: What about if `watch_glob` starts with `'/'`?

itowlson: Watching absolute paths is kind of an anti-pattern (and watching absolute paths _with a workdir_ feels even weirder) but okay I guess we can try to address it.  But then we ought to support them on Windows too, which means dealing with `D:\Git\Something\External\**\*.thingy` (or, hopefully, the forward-slashed equivalent) and somehow turning that into a forward-slashable watchexec expression.  I'll cross my fingers and hope that `Path::is_absolute` will magic my problems away!

calebschoepp (approved):

fibonacci1729 (approved):

PR #2838, opened 2024-09-16
Fix issues with publishing or consuming registries on Windows

### LINKED ISSUE #2829: Static assests like CSS and JS are not coming properly for SPIN application after deploying to cloud
Do any of the following links help with your problem?

[Application Prerequisites](https://developer.fermyon.com/spin/build#setting-up-for-spin-build)
Spin Build and Spin Up is Working
[Troubleshooting - Rust](https://developer.fermyon.com/spin/rust-components#troubleshooting)
NA
[Cloud FAQ](https://developer.fermyon.com/cloud/faq)
NA

If you still see a bug or problem, please let us know:

I have setup demo project here https://github.com/kunjee17/spin_cloud_test . Basically everything works without any issue. Even spin build and spin up also works. But as they say `it works on my machine` . When I deployed the application https://pi-health.fermyon.app/ html works but it looses css and js files. I did try to trouble shoot it, but unlike normal server I don't if that is possible to get into WASM to see if all files are there. More frustrating is that it works locally.  Do let me know if any other thing is required from my end. 

* Spin version (`spin --version`) spin 2.7.0 (a111517 2024-07-30)
```
cloud 0.9.1 [installed]
js2wasm 0.6.1 [installed]
```

itowlson:
Fixes #2829.  This PR forces all asset paths in the registry lockfile to be forward-slash format, which works on both Windows and Unix.

This also fixes an issue found during testing, where running from a local registry failed with a cryptic error, because the port separator is an illegal character in a Windows file name.

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -360,6 +360,10 @@ impl Client {
             }
             // Can unwrap because we got to 'entry' from walking 'source'
             let rel_path = entry.path().strip_prefix(source).unwrap();
+            // Paths must be in portable (forward slash) format in the registry,
+            // so that they can be placed correctly on any host system
+            let rel_path = portable_path(rel_path);
+
             tracing::trace!("Adding new layer for asset {rel_path:?}");
             // Construct and push layer, adding its digest to the locked component files Vec
             let layer = Self::data_layer(entry.path(), DATA_MEDIATYPE.to_string()).await?;
@@ -367,7 +371,7 @@ impl Client {
             let content_inline = content.inline.is_some();
             files.push(ContentPath {
                 content,
-                path: rel_path.into(),
+                path: rel_path,
             });
             // As a workaround for OCI implementations that don't support very small blobs,
             // don't push very small content that has been inlined into the manifest:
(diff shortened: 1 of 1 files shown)


itowlson:
To clarify: this should fix the original issue _when incorporated into `cloud-plugin`,_ although I haven't tested it in the cloud scenario yet.

rylev (approved):

crates/oci/src/client.rs
```diff
@@ -782,6 +786,41 @@ fn add_inferred(map: &mut BTreeMap<String, String>, key: &str, value: Option<Str
...
+#[cfg(not(target_os = "windows"))]
+fn portable_path(rel_path: &Path) -> PathBuf {
+    rel_path.into()
+}
+
+/// Takes a string intended for use as part of a path and makes it
+/// compatible with the local filesystem.
+#[cfg(target_os = "windows")]
+fn fs_safe_segment(segment: &str) -> impl AsRef<Path> {
+    segment.replace(':', "_")
```
lann, line 813: I think it would be fine to replace this unconditionally :shrug:

vdice (approved):

kate-goldenring (approved):

PR #2839, opened 2024-09-17
Pass loader into TriggerAppBuilder

rylev:
This allows users of the `TriggerAppBuilder` (like the containerd-shim) to pass in whatever loader they want.

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -14,10 +14,9 @@ use spin_common::sloth;
 use spin_common::ui::quoted_path;
 use spin_common::url::parse_file_url;
 use spin_factors::RuntimeFactors;
-use spin_factors_executor::FactorsExecutor;
+use spin_factors_executor::{ComponentLoader, FactorsExecutor};
 
-use crate::loader::ComponentLoader;
-use crate::{Trigger, TriggerApp};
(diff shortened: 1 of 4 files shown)


crates/trigger/src/cli.rs
```diff
@@ -220,7 +219,19 @@ impl<T: Trigger<B::Factors>, B: RuntimeFactorsBuilder> FactorsTriggerCommand<T,
...
-        let run_fut = builder.run(app, common_options, self.builder_args).await?;
+
+        // Create a component loader and enable AOT loading if the feature is enabled.
+        #[allow(unused_mut)]
+        let mut loader = ComponentLoaderImpl::new();
+        #[cfg(feature = "unsafe-aot-compilation")]
+        // Safety: This is safe because the feature is enabled.
+        unsafe {
+            loader.enable_loading_aot_compiled_components()
+        };
```
lann, line 230: I'm not sure that we want to enable this based just on the feature. At least, that isn't how it has worked previously.

rylev: Whoops - I mistakenly remembered this being turned on in the CLI when compiled with the feature, but I suppose this function was only used by those consuming `spin-trigger` as a library. I'll remove this.

lann (approved):

PR #2840, opened 2024-09-17
Simplify default database resolution in sqlite

rylev:
This greatly simplifies how labels are resolved to connection creators in `factor-sqlite`.

Instead of treating default labels as something special, the factor loses all knowledge of "default" labels. The label to connection creator mapping resides entirely inside of the factor's runtime config. The sqlite factor's runtime config was previously responsible for label to connection creator mapping for all database labels except for the default one. Now it handles the default one as well.

This removes the very notion of a "default" labeled database from the factor and instead moves that to a higher level concern (how the runtime configures each factors runtime config). 

The tests for the sqlite factor have also be massively simplified as they no longer use the Spin CLI's notion of runtime config toml and instead just supply the runtime config directly. This makes it much easier to understand the what the test is actually testing. 

If you're happy with this approach, I'd like to also do the same thing for key-value.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7448,7 +7448,6 @@ dependencies = [
  "spin-factors",
  "spin-factors-test",
  "spin-locked-app",
- "spin-sqlite",
  "spin-world",
  "table",
  "tokio",
--- a/crates/factor-sqlite/Cargo.toml
+++ b/crates/factor-sqlite/Cargo.toml
@@ -21,7 +21,6 @@ tracing = { workspace = true }
 
 [dev-dependencies]
 spin-factors-test = { path = "../factors-test" }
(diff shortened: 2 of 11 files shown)


itowlson:
@rylev Just to confirm I'm reading the code right: this doesn't change the user experience, right?  It's solely moving the implementation of that user experience to a different part of the code.

lann (approved):

rylev:
@itowlson correct if by "user" you mean "user of the Spin CLI" - this is purely a library change so it will change the experience for those using Spin as a library building their own Spin compliant runtime.

itowlson:
@rylev Yes, sorry, I should have been clearer.  Thanks for clarifying - sounds good!

PR #2841, opened 2024-09-18
Fix warning from redis crate.

rylev:
You can read about the warning here: https://github.com/redis-rs/redis-rs/issues/1228

Essentially, the functions in question are subject to a subtle change that will be coming in Rust 2024. To get ahead of this change, a warning is now produced in Rust 1.81. The fix is small even if it's unfortunately a bit awkward.


itowlson (approved):
Egad!  Nice find @rylev.  Do you think the assignment to unit merits a comment/link?  Definitely not a blocker though.

radu-matei (approved):

PR #2842, opened 2024-09-18
manifest: fix serialization of component dependency section

karthik2804:
This PR fixes the serialization of the `AppManifest` in regards to the component dependencies section. Previously using `try_from` meant that during serialization, the dependencies section had an extra nested `inner` field, leading to an invalid manifest where `spin up` will error out due to an unknown `inner` field.

### DIFF
--- a/crates/manifest/src/schema/v2.rs
+++ b/crates/manifest/src/schema/v2.rs
@@ -203,14 +203,6 @@ pub struct ComponentDependencies {
     pub inner: Map<DependencyName, ComponentDependency>,
 }
 
-impl TryFrom<Map<DependencyName, ComponentDependency>> for ComponentDependencies {
-    type Error = anyhow::Error;
-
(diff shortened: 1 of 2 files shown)


lann (approved):

crates/manifest/src/schema/v2.rs
```diff
@@ -197,7 +197,7 @@ pub struct Component {
 
 /// Component dependencies
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
-#[serde(try_from = "Map<DependencyName, ComponentDependency>")]
```
lann, line 200: Is the matching `TryFrom` still used? Might as well delete it if not.

karthik2804: I thought it was used in a test somewhere but apparently not. I have deleted it.

fibonacci1729 (approved):

PR #2843, opened 2024-09-18
Key Value re-*factor*ing

rylev:
Similar in scope to #2840. This is just a refactoring and so only impacts users of the key-value factor library.

* Removes the idea of a "default" store manager instead relying on runtime config handling to add a store manager for the default label.
* The "default" store manager is now added directly after runtime config is deserialized from toml.
* The key-value factor tests are reimagined
  * The tests now try to verify the actual behavior of the factor and not of the runtime config
  * Runtime configuration is now tested in the runtime config crate.


crates/runtime-config/src/lib.rs
```diff
@@ -448,32 +438,57 @@ fn sqlite_config_resolver(
...
 mod tests {
+    use std::{collections::HashMap, sync::Arc};
+
     use spin_factors::RuntimeFactors;
+    use spin_factors_test::TestEnvironment;
 
     use super::*;
 
+    /// Define a test factor with the given field and factor type.
+    macro_rules! define_test_factor {
```
lann, line 449: OOO FANCY

lann (approved):

PR #2845, opened 2024-09-19
Enable running some commands from subdirectories

### LINKED ISSUE #2828: Running some Spin commands from project subdirectories
It would be nice to be able to run commands like `spin build` and `spin up` from project sub-directories similar to how `git` works.

itowlson:
Fixes #2828.

@karthik2804 you asked me a couple of days ago about something related to this, so I am not sure if you were already working on it - there was nothing in GitHub and I saw you were working on something much cooler - if you did have something under way then feel free to close this.

The experience for running from a subdirectory looks like this:

image

(Exception: `spin doctor` where it integrates into an existing "this is the file you're partying on" notification.)

That output is not currently centralised - I thought about making the `spin_common` function print it, but `spin up` can't use that function anyway, and it would have required `spin_common` to depend on `terminal`, so I passed on that for now.  But we can do it if people prefer.  I certainly don't much love the internal API (loose booleans, oh the self-loathing) but thought I'd get feedback before I determined how to clean it up.

Also, the way I did it resulted in showing the absolute path (`/home/ivan/whatever/spin.toml`) rather than the relative path (`../spin.toml`).  The UX here seemed a little tricky - for one or two levels, the relative path is probably more helpful, but for more than that it just turns into a blur (../../../../../spin.toml).  For now I went with the absolute path because I had it to hand, but realistically the commonest case is one level down, maybe two, so there's a strong case for doing it the other way.  (Indeed, I'm thinking the terror and confusion of ../../../../../spin.toml would actually make it more noticeable that you were hitting some crazy random manifest you didn't intend.)

This PR adds the "run from subdirectory" behaviour to:
* build
* doctor
* registry push
* up
* watch

We _could_ add it to `spin add` but I think it might get confusing where the new component's directory got created.  We do already allow `spin add -f ../..` though, so maybe I'm being overly cautious.  Maybe not behaving like other commands is worse from a potential surprise POV.

### DIFF
--- a/crates/common/src/paths.rs
+++ b/crates/common/src/paths.rs
@@ -8,6 +8,19 @@ use crate::ui::quoted_path;
 /// The name given to the default manifest file.
 pub const DEFAULT_MANIFEST_FILE: &str = "spin.toml";
 
+/// Attempts to find a manifest. If a path is provided, that path is resolved
+/// using `resolve_manifest_file_path`; otherwise, a directory search is carried out
+/// using `search_upwards_for_manifest`. If a manifest is found, a boolean is
+/// also returned indicating if it was the default: this can be used to
+/// notify the user that a non-default manifest is being used.
+pub fn find_manifest_file_path(provided_path: Option<impl AsRef<Path>>) -> Result<(PathBuf, bool)> {
+    match provided_path {
+        Some(provided_path) => resolve_manifest_file_path(provided_path).map(|p| (p, true)),
+        None => search_upwards_for_manifest()
+            .ok_or_else(|| anyhow!("\"{}\" not found", DEFAULT_MANIFEST_FILE)),
+    }
+}
+
 /// Resolves a manifest path provided by a user, which may be a file or
 /// directory, to a path to a manifest file.
 pub fn resolve_manifest_file_path(provided_path: impl AsRef<Path>) -> Result<PathBuf> {
@@ -33,6 +46,30 @@ pub fn resolve_manifest_file_path(provided_path: impl AsRef<Path>) -> Result<Pat
             Ok(true) => anyhow!("Path {pd} is neither a file nor a directory"),
         };
         Err(err)
+    }
+}
+
+/// Starting from the current directory, searches upward through
+/// the directory tree for a manifest (that is, a file with the default
+/// manifest name `spin.toml`). If found, the path to the manifest
+/// is returned, with a boolean flag indicating if the found path was
+/// the default (i.e. `spin.toml` in the current directory).
+/// If no matching file is found, the function returns None.
+pub fn search_upwards_for_manifest() -> Option<(PathBuf, bool)> {
+    let mut inferred_dir = std::env::current_dir().unwrap();
+    let mut is_default = true;
+
+    loop {
+        let candidate = inferred_dir.join(DEFAULT_MANIFEST_FILE);
+
+        if candidate.is_file() {
+            return Some((candidate, is_default));
+        }
+
+        is_default = false;
+        let parent = inferred_dir.parent()?;
+
+        inferred_dir = parent.to_owned();
     }
 }
 
--- a/crates/common/src/paths.rs
+++ b/crates/common/src/paths.rs
@@ -10,12 +10,16 @@ pub const DEFAULT_MANIFEST_FILE: &str = "spin.toml";
 
 /// Attempts to find a manifest. If a path is provided, that path is resolved
 /// using `resolve_manifest_file_path`; otherwise, a directory search is carried out
-/// using `search_upwards_for_manifest`. If a manifest is found, a boolean is
-/// also returned indicating if it was the default: this can be used to
-/// notify the user that a non-default manifest is being used.
-pub fn find_manifest_file_path(provided_path: Option<impl AsRef<Path>>) -> Result<(PathBuf, bool)> {
+/// using `search_upwards_for_manifest`. If we had to search, and a manifest is found,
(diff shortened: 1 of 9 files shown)


crates/common/src/paths.rs
```diff
@@ -36,6 +49,30 @@ pub fn resolve_manifest_file_path(provided_path: impl AsRef<Path>) -> Result<Pat
...
+
+        if candidate.is_file() {
+            return Some((candidate, is_default));
+        }
+
+        is_default = false;
+        let parent = inferred_dir.parent()?;
+
+        inferred_dir = parent.to_owned();
+    }
```
karthik2804, line 73: Do we want to restrict the level of parent directories that we look for? or would that be unexpected behavior.

lann: The one heuristic I would maybe consider would be to stop the search if you see a `.git` directory. I can imagine edge cases where traversing up out of a repository could cause some pain. :shrug:

src/commands/build.rs
```diff
@@ -37,7 +37,16 @@ pub struct BuildCommand {
 
 impl BuildCommand {
     pub async fn run(self) -> Result<()> {
-        let manifest_file = spin_common::paths::resolve_manifest_file_path(&self.app_source)?;
+        let (manifest_file, is_default) =
+            spin_common::paths::find_manifest_file_path(self.app_source.as_ref())?;
+        if !is_default {
+            terminal::einfo!(
+                "Using 'spin.toml' from parent directory:",
```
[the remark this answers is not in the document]
itowlson: @lann I expected some high-quality bikeshedding on this one and I gotta say you have lived up to those expectations and more.

itowlson:
I HOPE YOU ARE HAPPY LANN 

image

lann:
SO HAPPY

crates/common/src/paths.rs
itowlson: Okay, it now stops if it sees a `.git` directory, and I also changed the loop to top out at 20 iterations - this shouldn't stop normal use but it does provide belt and braces in case something weird is going on.  The 20 is arbitrary and can be increased or decreased if people have opinions.  "If" people have opinions, who am I kidding

crates/common/src/paths.rs
```diff
@@ -8,6 +8,23 @@ use crate::ui::quoted_path;
 /// The name given to the default manifest file.
 pub const DEFAULT_MANIFEST_FILE: &str = "spin.toml";
 
+/// Attempts to find a manifest. If a path is provided, that path is resolved
+/// using `resolve_manifest_file_path`; otherwise, a directory search is carried out
+/// using `search_upwards_for_manifest`. If we had to search for a manifest is found,
```
karthik2804, line 13:
```suggestion
/// using `search_upwards_for_manifest`. If we had to search for a manifest and it is found,
```

crates/common/src/paths.rs
```diff
@@ -36,6 +53,41 @@ pub fn resolve_manifest_file_path(provided_path: impl AsRef<Path>) -> Result<Pat
...
+/// The search is abandoned if it reaches the root directory, or the
+/// root of a Git repository, without finding a 'spin.toml'.
+pub fn search_upwards_for_manifest() -> Option<(PathBuf, usize)> {
+    let candidate = PathBuf::from(DEFAULT_MANIFEST_FILE);
+
+    if candidate.is_file() {
+        return Some((candidate, 0));
+    }
+
+    for distance in 1..20 {
```
karthik2804, line 72: 20 seems a little bit high but no strong opinions. Maybe this is something we want to be configurable to users using env vars?

karthik2804 (approved):
Couple of non blocking suggestions - LGTM!

Thanks!

crates/common/src/paths.rs
```diff
@@ -36,6 +53,41 @@ pub fn resolve_manifest_file_path(provided_path: impl AsRef<Path>) -> Result<Pat
...
+        let inferred_dir = PathBuf::from("../".repeat(distance));
```
lann, line 73: FYI: https://doc.rust-lang.org/std/path/struct.Path.html#method.ancestors

e.g. `for candidate in Path::new(".").ancestors().take(20)`

crates/common/src/paths.rs
itowlson: Configuring it seems wildly excessive even by the standards of great-great-great-grandparent string code.  This was randomly picked to be high enough that nobody will ever run into it but low enough that if we hit some kind of crazy infinite loop case then it will exit before taking too long.  But happy to reduce it.  I originally had it at 10.  What am I bid?

(Price Is Right audience "higher! lower!" sound FX)

crates/common/src/paths.rs
itowlson: I don't think this does what I think you're suggesting.  In my tests `Path::new(".").ancestors()` produced `.` then the empty string, then stops.  Are you seeing something different?  Maybe depends on OS?

lann: Oh sure enough. I glanced at the example in its docs, saw `..` and mindlessly thought "that looks about right!"...

crates/common/src/paths.rs
itowlson: @karthik2804 I'm going to punt on this since it's non-blocking - happy to tweak as a follow-up.

itowlson:
oh wait no.  I can't merge because Karthik isn't an approved reviewer.  Aaaugh!

fibonacci1729 (approved):

PR #2848, opened 2024-09-19
Remove non-ascii header values instead of failing

lann:


crates/factor-outbound-http/src/spin.rs
```diff
@@ -233,3 +205,20 @@ async fn response_from_reqwest(res: reqwest::Response) -> Result<Response, HttpE
...
+
+fn headers_from_map(map: &http::HeaderMap) -> Vec<(String, String)> {
+    map.iter()
+        .filter_map(|(key, val)| {
+            Some((
+                key.to_string(),
+                val.to_str()
+                    .ok()
+                    .or_else(|| {
+                        tracing::warn!("Non-ascii response header value for {key}");
```
rylev, line 217: Would there be any use in printing the escaped version of the value?

lann: Yes but I'm trying to cut down on logging potentially sensitive data :grimacing:

PR #2851, opened 2024-09-19
Update to clap 4

rylev:
On the heels of #2847, this updates to Clap 4 which gets rid of some duplicate dependencies in the dependency tree. I did some light testing, but I'm not sure the best way to verify that this is functionally equivalent to the previous version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1292,6 +1292,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim 0.11.1",
+ "terminal_size",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


itowlson:
A couple of things I noticed while testing:

* The `spin --help` commands are no longer sorted alphabetically.  It would be good to restore the sorting.

* There are obvious formatting differences in help output, some of which are neutral but some of which make it harder to read.  For example, here is a fragment of `spin up --help` with clap 3:

image

and with clap 4:

image

The way clap 4 leaves line wrapping to the terminal, especially combined with the removal of colour highlighting, seems like a step backward.  I don't know if clap 4 provides us with any knobs to control this behaviour though.

* Okay this is super duper weird but when I do `spin up --follow foo` it now panics with "thread 'main' panicked at /home/ivan/.cargo/registry/src/index.crates.io-6f17d22bba15001f/clap_builder-4.5.2/src/builder/debug_asserts.rs:143:13:
Command http: Argument or group 'tls-key' specified in 'requires*' for 'tls_cert' does not exist".  This doesn't happen with Spin canary.  `--disable-wasmtime-cache` , `--listen`, and presumably other trigger options, have the same problem.  Heaven knows if there are any other bear traps like this, I only found it by accident when testing "multiple occurrence" flags...!

Issue #2816, opened 2024-09-09
OTEL Traces for Database interactions should only optionally include values


endocrimes:
Database queries often contain potentially sensitive data (PII, credentials/api tokens, ...), currently, Spin will always send those values to OTel (example from a `sqlite` database):

<img width="333" alt="image" src="https://github.com/user-attachments/assets/14d46d75-f005-4cc4-b752-77566c2ab10b">

This would be a compliance or security issue in some contexts, so ideally we should only do this if a developer _opts in_ for further debugging support (in local/dev envs). Or never emit those values (and instead just include types or similar).

PR #2857, opened 2024-09-23
Telemetry fixes

calebschoepp:
## Summary

- Fixes bug that slipped in with #2834.
- Improves a comment.
- Removes potentially sensitive values from OTel traces.

## Questions

- Do we want to remove DB addresses from OTel traces b/c they might contain credentials inline?
- Do we want to remove key values everywhere from OTel traces? This seems valuable enough to me that I think we should just keep it and document it.

Fixe #2816

### DIFF
--- a/crates/factor-outbound-mysql/src/host.rs
+++ b/crates/factor-outbound-mysql/src/host.rs
@@ -5,6 +5,7 @@ use spin_world::v1::mysql as v1;
 use spin_world::v2::mysql::{self as v2, Connection};
 use spin_world::v2::rdbms_types as v2_types;
 use spin_world::v2::rdbms_types::ParameterValue;
+use tracing::field::Empty;
 use tracing::{instrument, Level};
 
 use crate::client::Client;
@@ -38,8 +39,10 @@ impl<C: Client> v2::Host for InstanceState<C> {}
 
 #[async_trait]
 impl<C: Client> v2::HostConnection for InstanceState<C> {
-    #[instrument(name = "spin_outbound_mysql.open_connection", skip(self), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql"))]
+    #[instrument(name = "spin_outbound_mysql.open", skip(self, address), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", db.address = Empty, server.port = Empty, db.namespace = Empty))]
     async fn open(&mut self, address: String) -> Result<Resource<Connection>, v2::Error> {
+        spin_factor_outbound_networking::record_address_fields(&address);
+
         if !self
             .is_address_allowed(&address)
             .await
--- a/crates/factor-outbound-mysql/src/host.rs
+++ b/crates/factor-outbound-mysql/src/host.rs
@@ -52,7 +52,7 @@ impl<C: Client> v2::HostConnection for InstanceState<C> {
         self.open_connection(&address).await
     }
 
-    #[instrument(name = "spin_outbound_mysql.execute", skip(self, connection), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", otel.name = statement))]
+    #[instrument(name = "spin_outbound_mysql.execute", skip(self, connection, params), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", otel.name = statement))]
     async fn execute(
         &mut self,
         connection: Resource<Connection>,
@@ -66,7 +66,7 @@ impl<C: Client> v2::HostConnection for InstanceState<C> {
             .await?)
     }
 
-    #[instrument(name = "spin_outbound_mysql.query", skip(self, connection), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", otel.name = statement))]
+    #[instrument(name = "spin_outbound_mysql.query", skip(self, connection, params), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", otel.name = statement))]
     async fn query(
         &mut self,
         connection: Resource<Connection>,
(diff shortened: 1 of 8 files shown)


rylev (approved):
Seems like an improvement. As for your questions:

* re: DB addresses - this seems like a smart idea to remove
* key-value: I think we definitely want to remove values - I can see the argument both for keeping and removing keys. Perhaps we keep them for now, and see what people think over time?

lann:
> Do we want to remove key values everywhere from OTel traces? This seems valuable enough to me that I think we should just keep it and document it.

We could perhaps drop it by default but allow opting-in by level, e.g.
```rust
#[instrument(name = "spin_key_value.get", skip(self, store, key), err(level = Level::INFO), fields(otel.kind = "client", key = Empty))]
...
// I believe this would allow e.g. RUST_LOG=spin_factor_key_value=debug
if tracing::enabled!(Level::DEBUG) {
  span.record("key", key);
}
```

calebschoepp:
> > Do we want to remove key values everywhere from OTel traces? This seems valuable enough to me that I think we should just keep it and document it.
> 
> We could perhaps drop it by default but allow opting-in by level, e.g.
> 
> ```rust
> #[instrument(..., fields(key = Empty))]
> ...
> // I believe this would allow e.g. RUST_LOG=spin_factor_key_value=debug
> if tracing::enabled!(Level::DEBUG) {
>   span.record("key", key);
> }
> ```

I tried this and it didn't work.

crates/factor-outbound-mysql/src/host.rs
```diff
@@ -38,8 +40,15 @@ impl<C: Client> v2::Host for InstanceState<C> {}
 
 #[async_trait]
 impl<C: Client> v2::HostConnection for InstanceState<C> {
-    #[instrument(name = "spin_outbound_mysql.open_connection", skip(self), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql"))]
+    #[instrument(name = "spin_outbound_mysql.open", skip(self, address), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", db.address = Empty, server.port = Empty, db.namespace = Empty))]
     async fn open(&mut self, address: String) -> Result<Resource<Connection>, v2::Error> {
+        if let Ok(url) = Url::parse(&address) {
+            let span = tracing::Span::current();
+            span.record("db.address", url.host_str().unwrap_or_default());
```
rylev, line 47: Should this be `db.host` instead of `db.address` or is there some other external reason to label this as a address?

crates/factor-outbound-mysql/src/host.rs
```diff
@@ -38,8 +40,15 @@ impl<C: Client> v2::Host for InstanceState<C> {}
...
+            span.record("server.port", url.port().unwrap_or_default());
```
rylev, line 48: I'm also curious about the choice around "server.port" as opposed to perhaps "db.port"

crates/factor-outbound-mysql/src/host.rs
```diff
@@ -38,8 +40,15 @@ impl<C: Client> v2::Host for InstanceState<C> {}
 
 #[async_trait]
 impl<C: Client> v2::HostConnection for InstanceState<C> {
-    #[instrument(name = "spin_outbound_mysql.open_connection", skip(self), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql"))]
+    #[instrument(name = "spin_outbound_mysql.open", skip(self, address), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", db.address = Empty, server.port = Empty, db.namespace = Empty))]
     async fn open(&mut self, address: String) -> Result<Resource<Connection>, v2::Error> {
+        if let Ok(url) = Url::parse(&address) {
```
rylev, line 45: I wish we could capture this logging functionality in one place instead of copying it. In particular I'm most interested in capturing a comment about how we specifically _don't_ want to log the full address due to it potentially containing sensitive data. 

How would this look if we moved this to a function in `spin-factor-outbound-networking`?

rylev (approved):
I still have questions, but nothing that should block if you feel this is ready to merge.

crates/factor-outbound-mysql/src/host.rs
calebschoepp: Following the guidelines [here](https://opentelemetry.io/docs/specs/semconv/database/sql/).

crates/factor-outbound-mysql/src/host.rs
calebschoepp: Following the guidelines [here](https://opentelemetry.io/docs/specs/semconv/database/sql/).

crates/factor-outbound-mysql/src/host.rs
calebschoepp: I've addressed this (pun intended)

calebschoepp:
Ready for one more review and then for someone to merge it if they think it's good to go.

rylev:
@calebschoepp looks like the failure in CI is legit:

```
error: expected statement after outer attribute
 --> crates/factor-outbound-networking/src/lib.rs:258:1
  |
3 | #[instrument(fields(db.address = Empty, server.port = Empty, db.namespace = Empty))]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

calebschoepp:
> @calebschoepp looks like the failure in CI is legit:
> 
> ```
> error: expected statement after outer attribute
>  --> crates/factor-outbound-networking/src/lib.rs:258:1
>   |
> 3 | #[instrument(fields(db.address = Empty, server.port = Empty, db.namespace = Empty))]
>   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```

Fixed

rylev (approved):

PR #2862, opened 2024-09-24
Use official async-tar

### LINKED ISSUE #2846: Replace async-tar dependency
We're currently relying on [a fork of async-tar](https://github.com/vdice/async-tar) - it seems because of previous Windows build failures. It would be best to replace this dependency with a more maintained version. Ideally we would replace `async-tar` entirely as it is currently the only dependency bringing in `async-std` which itself has ~5 dependencies we don't otherwise use. It might make sense to switch to the sync `tar` crate and use `tokio::sync::spawn_blocking` where appropriate.

itowlson:
Fixes #2846.

This built locally for me on Windows.  I'm pushing this to see whether it builds in CI, and if the artefacts are usable.

@vdice The dependency is used only (I believe) in the archive layers stuff (https://github.com/fermyon/spin/pull/1847).  Do we have a test case I can use to hit that?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -455,8 +455,9 @@ dependencies = [
 
 [[package]]
 name = "async-tar"
-version = "0.4.2"
-source = "git+https://github.com/vdice/async-tar?rev=71e037f9652971e7a55b412a8e47a37b06f9c29d#71e037f9652971e7a55b412a8e47a37b06f9c29d"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a42f905d4f623faf634bbd1e001e84e0efc24694afa64be9ad239bf6ca49e1f8"
 dependencies = [
  "async-std",
  "filetime",
--- a/crates/oci/Cargo.toml
+++ b/crates/oci/Cargo.toml
@@ -7,8 +7,7 @@ edition = { workspace = true }
 [dependencies]
 anyhow = { workspace = true }
 async-compression = { version = "0.4", features = ["gzip", "tokio"] }
-# Fork with nested async-std dependency bumped to satisfy Windows build; branch/revision is protected
-async-tar = { git = "https://github.com/vdice/async-tar", rev = "71e037f9652971e7a55b412a8e47a37b06f9c29d" }
+async-tar = "0.5"
 base64 = "0.22"
 chrono = "0.4"
 # Fork with updated auth to support ACR login


vdice:
> @vdice The dependency is used only (I believe) in the archive layers stuff (https://github.com/fermyon/spin/pull/1847). Do we have a test case I can use to hit that?

I thought I did add one but it turns out I was thinking of an e2e test in the [containerd-shim-spin repo](https://github.com/spinkube/containerd-shim-spin/blob/6f9162914f07b1ccda538e30b98f2e8011228f7a/scripts/up.sh#L32-L36).

So, the quick way to test is to `export SPIN_OCI_ARCHIVE_LAYERS=1` which will force use of archive layers and then go through the flow of `spin registry push <ref>` and `spin up -f <ref>` to ensure the published/pulled app works as expected.

vdice:
PS One other testing note: I believe the app does need at least one static asset with a size over a [certain minimum](https://github.com/fermyon/spin/blob/a22119c97a865201dcd3ea7ed266135d22c1b36e/crates/oci/src/client.rs#L52), or else the content is inlined in the oci manifest.

itowlson:
@vdice Okay, I set that environment variable, and did a push from Windows of a test application with one small asset (about a dozen bytes).  I was able to spin up from the registry on both Windows and Linux and verify that the application could read the asset.  I also downloaded the registry image with a simple OCI client and printed out the layers' media types:

```
the image has 3 layers
application/vnd.wasm.content.layer.v1+wasm
application/vnd.wasm.content.bundle.v1.tar+gzip
application/vnd.fermyon.spin.application.v1+config
```

If I remove the environment variable and re-push, the `tar+gzip` layer disappears, so I'm assuming that my test did indeed exercise the code path and it is doing a reasonable thing (it certainly isn't crashing or anything).

I realise this is a very limited test, but I am going to mark this as ready to review - if there are further tests you'd like me to run then feel free to put the mockers on it.

Thank you ever so much for your guidance on this!

rylev (approved):
LGTM, but I will note it doesn't fully address the issues laid out in #2846, namely the use of `async-tar` in general which brings in `async-std` and associated dependencies. 

I'd be happy with merging this PR as is but keeping #2846 open so we can discuss whether removing `async-tar` entirely (and using `tar` along with `spawn_blocking`) would be a better way to go.

vdice (approved):
LGTM!  All apps I tested with work great.

itowlson:
@rylev I can't find a way to make GitHub unlink the issue but I hear you.  I will merge this then reopen your issue.

PR #2863, opened 2024-09-25
Move `HandlerType` to `spin-http`

rylev:
This is useful functionality that other runtimes can use to determine which handler export they should call. In the future, we probably want to move calling of the right export completely into `spin-http`, but that can be done as a follow up.

### DIFF
--- a/crates/factors-executor/src/lib.rs
+++ b/crates/factors-executor/src/lib.rs
@@ -213,7 +213,7 @@ impl<'a, T: RuntimeFactors, U> FactorsInstanceBuilder<'a, T, U> {
     }
 
     /// Returns the underlying wasmtime engine for the instance.
-    pub fn engine(&self) -> &spin_core::wasmtime::Engine {
+    pub fn wasmtime_engine(&self) -> &spin_core::WasmtimeEngine {
         self.instance_pre.engine()
     }
 
--- a/crates/factors-executor/src/lib.rs
+++ b/crates/factors-executor/src/lib.rs
@@ -211,6 +211,16 @@ impl<'a, T: RuntimeFactors, U> FactorsInstanceBuilder<'a, T, U> {
     pub fn factor_builder<F: Factor>(&mut self) -> Option<&mut F::InstanceBuilder> {
         self.factor_builders().for_factor::<F>()
     }
+
+    /// Returns the underlying wasmtime engine for the instance.
+    pub fn engine(&self) -> &spin_core::wasmtime::Engine {
+        self.instance_pre.engine()
+    }
+
+    /// Returns the compiled component for the instance.
+    pub fn component(&self) -> &Component {
+        self.instance_pre.component()
+    }
 }
 
 impl<'a, T: RuntimeFactors, U: Send> FactorsInstanceBuilder<'a, T, U> {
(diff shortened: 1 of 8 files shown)


crates/factors-executor/src/lib.rs
```diff
@@ -211,6 +211,16 @@ impl<'a, T: RuntimeFactors, U> FactorsInstanceBuilder<'a, T, U> {
     pub fn factor_builder<F: Factor>(&mut self) -> Option<&mut F::InstanceBuilder> {
         self.factor_builders().for_factor::<F>()
     }
+
+    /// Returns the underlying wasmtime engine for the instance.
+    pub fn engine(&self) -> &spin_core::wasmtime::Engine {
```
lann, line 216: Slightly confusing because in this same crate `FactorsExecutorApp::engine()` returns a `spin_core::Engine`. Maybe:

```suggestion
    pub fn wasmtime_engine(&self) -> &spin_core::wasmtime::Engine {
```

lann (approved):

crates/factors-executor/src/lib.rs
rylev: I agree. The change seems like a fine one to make. While we're here though - a larger put perhaps cleaner change might be to rename `spin_core::Engine` to `spin_core::WasmRuntime`. This might be preferrable since even with your suggested change, a user still has to know how to disambiguate when a type signature includes an unqualified `Engine`.

lann: or maybe `SpinEngine`? "Runtime" is also awfully overloaded...

lann: yet another alternative would be to re-export `wasmtime::Engine` as `spin_core::WasmtimeEngine`

crates/http/Cargo.toml
```diff
@@ -9,13 +9,14 @@ anyhow = { workspace = true }
...
 hyper = { workspace = true }
-indexmap = "1"
+indexmap = "2"
 percent-encoding = "2"
 routefinder = "0.5.4"
 serde = { workspace = true }
 spin-app = { path = "../app", optional = true }
 spin-locked-app = { path = "../locked-app" }
 tracing = { workspace = true }
+wasmtime = { workspace = true }
```
itowlson, line 19: This should, I think, be under the `runtime` feature.  We have consumers of this crate who use it for things like route parsing (e.g. the cloud plugin), but do not want to drag the whole of wasmtime along with it.

(I'm happy if you prefer to reorganise things to avoid the feature thing, but I really really want a way for cloud-plugin to get what it needs without including wasmtime.  Whether that is features or separate crates, I don't care, but it makes a big difference!)

itowlson: Ah, I just tried updating cloud-plugin and it looks like some of the other recebt changes have also brought in wasmtime dependency.  I'll get a PR out for all of it together.

itowlson: Okay, I did some more investigation and I'm gonna punt on it for now.  The wasmtime dependency makes 1 min of difference to a clean cloud-plugin build, but not enough difference to an incremental build to be worth the candle, and we are churning cloud-plugin dependencies a lot more slowly than when it drove me to extremes.  Something to maybe think about how we can streamline in a more sustainable way in future, but not worth fretting about for now.

Sorry for bending your ear @rylev!

PR #2865, opened 2024-09-25
Update rdbms-types to support Date / Tmestamp

tyler-harpool:
Purpose of this PR is to add datetime and timestamp support.

### DIFF
--- a/wit/rdbms-types.wit
+++ b/wit/rdbms-types.wit
@@ -23,6 +23,10 @@ interface rdbms-types {
       floating64,
       str,
       binary,
+      date,
+      time,
+      datetime,
+      timestamp,
       other,
   }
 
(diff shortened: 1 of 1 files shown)


rylev (changes requested):
Thanks for the PR! Adding this will constitute a breaking change so we'll need to make sure we do accept this at the right time. We're figuring out what changes to land in Spin 3.0 so this will definitely be a good candidate for discussion. In the meantime, we appreciate your patience while the discussion happens.

itowlson:
Note this will also require updating the conversion routines in the MySQL and Postgres factors.

tyler-harpool:
Would love to handle this, but bandwidth is limited as I'm on holiday/PTO. I'm happy to punt as this would be a challenge on my side. Happy to do anything to assist.

itowlson:
I'm closing this in favour of #2890 - @tyler-harpool is aware.  Thank you again Tyler for getting this process under way!

PR #2886, opened 2024-10-14
reduce environment variable provider precedence

kate-goldenring:
Inverts precedence of variables providers giving the environment variable provider least precedence

fixes https://github.com/orgs/fermyon/projects/64/views/1?pane=issue&itemId=79983920&issue=fermyon%7Cspin%7C2584

### DIFF
--- a/crates/variables/src/lib.rs
+++ b/crates/variables/src/lib.rs
@@ -19,20 +19,22 @@ use spin_factor_variables::runtime_config::RuntimeConfig;
 /// Resolves a runtime configuration for the variables factor from a TOML table.
 pub fn runtime_config_from_toml(table: &impl GetTomlValue) -> anyhow::Result<RuntimeConfig> {
     // Always include the environment variable provider.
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

lann (approved):

rylev (approved):
This is fine from a technical point of view. As for whether this is the right behavior or not, I'm not sure. I'm curious if @itowlson still has concerns about this.

kate-goldenring:
I agree that I would like @itowlson's thoughts before merging. I understand that env vars seem like they should take precedence on the CLI but seem less relevant when orchestrating Spin apps with SpinKube. Maybe we should make this configurable by the runtime when it loads the variables factor.

itowlson:
Yes, I still have concerns, but by my count the majority of people feel this is the right thing to do.  And my concern is for a specific scenario (the override on the command line one) that, at a technical level, is impossible to differentiate from other scenarios where I _do_ agree this is the right thing to do.  So I am keeping mousy quiet.

What?  By my standards, 60 words _is_ mousy quiet.

PR #2892, opened 2024-10-21
fix: move component filtering methods to crates

kate-goldenring:
Moves all the component filtering helpers to crates that are accessible to other runtime implementors to prevent code duplication. Also, enables runtimes to pass in extra validating functions.

Thank you @itowlson for pairing on this! 

Most of the code is the same. I did change the error message on `validate_service_chaining_for_components` to not be specific to CLI input.

### DIFF
--- a/crates/app/Cargo.toml
+++ b/crates/app/Cargo.toml
@@ -5,6 +5,12 @@ authors = { workspace = true }
 edition = { workspace = true }
 
 [dependencies]
+anyhow = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 spin-locked-app = { path = "../locked-app" }
+
+[dev-dependencies]
+toml = { workspace = true }
+spin-factors-test = { path = "../factors-test" }
+tokio = { workspace = true }
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -31,7 +31,7 @@ pub const OCI_IMAGE_DIGEST_KEY: MetadataKey = MetadataKey::new("oci_image_digest
 
 /// Validation function type for ensuring that applications meet requirements
 /// even with components filtered out.
-pub type ValidatorFn = dyn Fn(&App, &[String]) -> anyhow::Result<()>;
+pub type ValidatorFn = dyn Fn(&App, &[&str]) -> anyhow::Result<()>;
 
 /// An `App` holds loaded configuration for a Spin application.
 #[derive(Debug, Clone)]
@@ -171,7 +171,7 @@ impl App {
     /// Introspects the LockedApp to find and selectively retain the triggers that correspond to those components
     fn retain_components(
         mut self,
-        retained_components: &[String],
+        retained_components: &[&str],
         validators: &[&ValidatorFn],
     ) -> Result<LockedApp> {
         self.validate_retained_components_exist(retained_components)?;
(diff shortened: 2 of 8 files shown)


crates/app/src/lib.rs
```diff
@@ -27,6 +29,8 @@ pub const APP_DESCRIPTION_KEY: MetadataKey = MetadataKey::new("description");
 /// MetadataKey for extracting the OCI image digest.
 pub const OCI_IMAGE_DIGEST_KEY: MetadataKey = MetadataKey::new("oci_image_digest");
 
+type ValidatorFn = dyn Fn(&App, &[String]) -> anyhow::Result<()>;
```
itowlson, line 32: As this is used in a public interface, I wonder if it needs a `pub` or a doc comment or something?  I'm not sure what the expectation is with type aliases (and clearly the compiler doesn't mind).

crates/factor-outbound-networking/src/config.rs
```diff
@@ -34,6 +34,46 @@ pub fn allowed_outbound_hosts(component: &AppComponent) -> anyhow::Result<Vec<St
...
 
+/// Validates that all service chaining of an app will be satisfied by the
+/// supplied subset of components.
+///
+/// This does a best effort look up of components that are
+/// allowed to be accessed through service chaining and will error early if a
+/// component is configured to to chain to another component that is not
+/// retained. All wildcard service chaining is disallowed and all templated URLs
+/// are ignored.
+pub fn validate_service_chaining_for_components(
```
itowlson, line 45: It might be desirable to give this a more specific name but the doc comment covers it well and the only names I can think of are a bit verbose (e.g. `validate_retained_components_include_service_chaining_dependencies`!).  Not a blocker, just if inspiration strikes!

crates/app/Cargo.toml
```diff
@@ -5,6 +5,13 @@ authors = { workspace = true }
...
 
 [dependencies]
+anyhow = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 spin-locked-app = { path = "../locked-app" }
+
+[dev-dependencies]
+toml = { workspace = true }
+tempfile = { workspace = true }
```
itowlson, line 15: Is this used?

itowlson (approved):
LGTM, couple of questions/checks but nothing important!

crates/factor-outbound-networking/src/config.rs
kate-goldenring: Even though this is used by Spin when retaining only a set of components, technically this could be used elsewhere to establish that a set of components have satisfied service chaining. Maybe a runtime implementor would use this for app splitting ... but i like the vague name because technically the functionality could be used in another objective

crates/app/Cargo.toml
kate-goldenring: It isn't now that i pulled in the factors tests. Removing!

crates/factors-test/src/lib.rs
```diff
@@ -91,10 +91,14 @@ impl<T: RuntimeFactors> TestEnvironment<T> {
...
-        let toml_str = toml::to_string(&self.manifest).context("failed serializing manifest")?;
-        let dir = tempfile::tempdir().context("failed creating tempdir")?;
-        let path = dir.path().join("spin.toml");
-        std::fs::write(&path, toml_str).context("failed writing manifest")?;
-        spin_loader::from_file(&path, FilesMountStrategy::Direct, None).await
+        build_locked_app(&self.manifest).await
     }
 }
+
+pub async fn build_locked_app(manifest: &toml::Table) -> anyhow::Result<LockedApp> {
```
rylev, line 98: Should this be a `TryFrom<toml::Toml> for LockedApp` impl?

rylev (approved):
Looks good - I only have a few nits and suggestions, but feel free to ignore them.

PR #2893, opened 2024-10-23
Include component dependencies in OCI artifact

### LINKED ISSUE #2891: Raise error when trying to push app with local dependencies
When running `spin registry push ...` or `spin deploy` a check should happen on the locked app to be published that yields an error if any component in the app has a dependency referencing a wasm component located on disk.

E.g.:

```toml
[component.spin-animal-facts.dependencies]
"component:animal-facts/animal-facts" = { path = "animal_facts_dep.wasm" }
```

Alternatively, we could package and publish the dependencies transitively but that seems a little too magical.

itowlson:
Fixes #2891.

@fibonacci1729 @vdice this is rather "works on my machine" at the moment; do you have any thoughts on how to rassle it into a test harness?

### DIFF
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -266,6 +266,24 @@ impl Client {
 
             layers.push(layer);
 
+            let mut deps = BTreeMap::default();
+            for (dep_name, mut dep) in c.dependencies {
+                let source = dep
+                    .source
+                    .content
+                    .source
+                    .context("dependency loaded from disk should contain a file source")?;
+                let source = parse_file_url(source.as_str())?;
+
+                let layer = Self::wasm_layer(&source).await?;
+
+                dep.source.content = self.content_ref_for_layer(&layer);
+                deps.insert(dep_name, dep);
+
+                layers.push(layer);
+            }
+            c.dependencies = deps;
+
             let mut files = Vec::new();
             for f in c.files {
                 let source = f
@@ -292,6 +310,7 @@ impl Client {
                 }
             }
             c.files = files;
+
             components.push(c);
         }
         locked.components = components;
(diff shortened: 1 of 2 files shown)


crates/oci/src/client.rs
```diff
@@ -266,6 +266,24 @@ impl Client {
 
             layers.push(layer);
 
+            let mut deps = BTreeMap::default();
+            for (dep_name, mut dep) in c.dependencies {
```
fibonacci1729, line 270: Is this always assuming that dependencies are located on disk?

itowlson: Hmm, by this stage everything not on disk _should_ have been resolved to file paths in the cache - that's what happens for primary components coming from the network.  But you're right, dependencies travel their own road, so I need to verify that assumption.  Thanks for the catch!

itowlson: Yeah I think we're good.  It's fiddly to test for sure given that a lurking remote reference would presumably still get resolved without much deleting, but this makes me pretty confident:

```toml
[component.spin-comp-deps-test-two]
source = "spin_comp_deps_test.wasm"
[component.spin-comp-deps-test-two.dependencies]
"vscode:example/types" = { package = "vscode:example", version = "0.0.2", registry = "registrytest-vfztdiyy.fermyon.app" }
```

```json
// lockfile
    {
      "id": "spin-comp-deps-test-two",
      "source": {
        "content_type": "application/wasm",
        "source": "file:///home/ivan/testing/target-env-test/spin_comp_deps_test.wasm"
      },
      "dependencies": {
        "vscode:example/types": {
          "source": {
            "content_type": "application/wasm",
            "source": "file:///home/ivan/.cache/spin/registry/wasm/sha256:fcc77e6020ee40ffcc90667f60605adb3bc050b93e86e1086d1ff307bf7e80ef"
          },
          "export": null
        }
      }
    }
```

```json
// OCI config
    {
      "id": "spin-comp-deps-test-two",
      "source": {
        "content_type": "application/wasm",
        "digest": "sha256:7acaaaab00c5c08e5563656eca8a52e7d718c3da775d298b2b222876f76677ec"
      },
      "dependencies": {
        "vscode:example/types": {
          "source": {
            "content_type": "application/wasm",
            "digest": "sha256:fcc77e6020ee40ffcc90667f60605adb3bc050b93e86e1086d1ff307bf7e80ef"
          },
          "export": null
        }
      }
    }
```

fibonacci1729 (approved):
This looks good to me! I'm also not sure about how to add a test harness or how best to test this. Nonetheless, manually tested so it at least works on 2 machines! 🪅

vdice (approved):
Re: testing, we did have a unit test around assemble_layers which is [now commented-out](https://github.com/fermyon/spin/blob/main/crates/oci/src/client.rs#L866-L1111) from some iteration of factors re-factoring.  Adding a case for component deps there could've made sense -- but we'd need to uncomment/update that test first.

Meanwhile, I don't yet see equivalent unit tests for loader.

Do we think we're at a state where we also might want some sort of integration test with a local registry, etc?

All of which is to say, if @fibonacci1729 agrees, I'd say we're fine to get this in and then address some tech/testing debt in follow-up(s).

itowlson:
@vdice Integration tests with a local registry sound like a good longer terms plan.  It gets gnarly for e2e though because a real test requires you to reference a component from Registry A, push the application to Registry B, and then _turn off Registry A..._

In the meantime I've fixed the existing `assemble_layers` test and added a case for a dependency (not very visible because of the comment diff, sorry).  Thanks for highlighting this.

vdice:
> In the meantime I've fixed the existing assemble_layers test and added a case for a dependency

🎉 Thank you!

PR #2895, opened 2024-10-24
WASI Key Value 0.2.0-draft2 support

devigned:
This pull request adds support for the `wasi-keyvalue@0.2.0-draft2`. 

I'm opening this as a draft to get some initial feedback as I implement the remaining providers. Currently, the only provider implemented is the SQLite provider.

fixes #2486


crates/key-value-spin/src/store.rs
```diff
@@ -156,6 +157,140 @@ impl Store for SqliteStore {
...
+                .map_err(log_error)?
+                .execute((":name", &self.name, ":keys", ptr))
+                .map_err(log_error)
+                .map(drop)
+        })
+    }
+
+    // The assumption with increment is that if the value for the key does not exist, it will be
+    // assumed to be zero. In the case that we are unable to unmarshal the value into an i64 an error will be returned.
+    async fn increment(&self, key: String, delta: i64) -> Result<i64, Error> {
```
itowlson, line 223: This is an interesting design niggle - we have not previously defined an encoding for integers, and I don't think `wasi-keyvalue` defines one either.  All the "set" APIs are byte arrays.  So how do I set an initial value to which I can later apply `increment`?  Hopefully this won't be an issue in practice, but might be worth considering in the wasi-keyvalue specification.

crates/key-value-spin/src/store.rs
```diff
@@ -156,6 +157,140 @@ impl Store for SqliteStore {
...
+    async fn current(&self) -> Result<Option<Vec<u8>>, Error> {
+        Ok(self.value.clone())
+    }
+
+    async fn swap(&self, value: Vec<u8>) -> Result<(), Error> {
+        task::block_in_place(|| {
+            self.connection
+                .lock()
+                .unwrap()
+                .prepare_cached("UPDATE spin_key_value SET value=$3 WHERE store=$1 and key=$2")
```
itowlson, line 287: I think this omits the "compare" part of the CAS operation?  You might need to add a clause to the WHERE to require `value` = `self.value` (hmm, except self.value is an option).  And then check the affected row count.

The specification requires a "value has changed" CAS failure to return a `cas-failed` error containing a _new_ `cas` resource set to the current value.  Perhaps that's best done at the top level rather than repeating it in each provider, I am not sure - but the current top-level doesn't include it I think.

crates/key-value-spin/src/store.rs
```diff
@@ -156,6 +157,140 @@ impl Store for SqliteStore {
...
+        })
+    }
+
+    async fn set_many(&self, key_values: Vec<(String, Vec<u8>)>) -> Result<(), Error> {
+        task::block_in_place(|| {
+            let mut binding = self.connection.lock().unwrap();
+            let tx = binding.transaction().map_err(log_error)?;
+            for kv in key_values {
+                tx.prepare_cached(
+                    "INSERT INTO spin_key_value (store, key, value) VALUES ($1, $2, $3)
```
itowlson, line 193: You're right, this is pretty unlovely, but I have no other ideas.  SQLite does provide for bulk inserts, but I don't think they let you do an update where the row already existed.  Hopefully it will not be too bad because in-process...!

crates/factor-key-value/src/host.rs
```diff
@@ -141,6 +178,201 @@ impl key_value::HostStore for KeyValueDispatch {
...
+        &mut self,
+        cas_res: Resource<wasi_keyvalue::atomics::Cas>,
+        value: Vec<u8>,
+    ) -> Result<std::result::Result<(), wasi_keyvalue::atomics::CasError>> {
+        let cas = self
+            .get_cas(cas_res)
+            .map_err(|e| wasi_keyvalue::atomics::CasError::StoreError(wasi_keyvalue::atomics::Error::Other(e.to_string())))?;
+        Ok(cas.swap(value)
+            .await
+            .map_err(|e| wasi_keyvalue::atomics::CasError::StoreError(wasi_keyvalue::atomics::Error::Other(e.to_string()))))
```
itowlson, line 372: See comment below - in some cases this needs to be `CasError::CasFailed`

itowlson (commented):
Thanks for working on this David.  Fingers crossed that the SQLite was the worst of it.  I left a few comments but it's brilliant how it's coming along!

crates/key-value-spin/src/store.rs
devigned: Nice catch! I'm just going with little endian encoding of i64s. I think the guidance on using increment right now is to start with a key/value that doesn't exist and "increment" it to your value of choice. Then use increment or get from there on out.

I do believe this is something that we'll need to consider for future interface revisions.

crates/key-value-spin/src/store.rs
devigned: I have fixed up the CAS failures and added tests to the sqlite implementation to verify the behavior.

crates/factor-key-value/src/host.rs
devigned: Indeed. I added that behavior when I added the tests for sqlite.

crates/factor-key-value/src/lib.rs
```diff
@@ -40,6 +41,7 @@ impl Factor for KeyValueFactor {
     fn init<T: Send + 'static>(&mut self, mut ctx: InitContext<T, Self>) -> anyhow::Result<()> {
         ctx.link_bindings(spin_world::v1::key_value::add_to_linker)?;
         ctx.link_bindings(spin_world::v2::key_value::add_to_linker)?;
+        ctx.link_bindings(spin_world::wasi::keyvalue::store::add_to_linker)?;
```
itowlson, line 44: Do we need to link the atomics and batch interfaces too?  In wasi-blobstore, I seemed to run into a problem if I didn't link the additional interfaces.

crates/key-value-redis/Cargo.toml
```diff
@@ -12,6 +12,7 @@ spin-core = { path = "../core" }
 spin-factor-key-value = { path = "../factor-key-value" }
 tokio = { workspace = true }
 url = { workspace = true }
+log = "0.4.22"
```
itowlson, line 15: I think we use `tracing` rather than `log` in the runtime.  `log` is sometimes used in tests.

crates/key-value-redis/src/store.rs
```diff
@@ -25,10 +26,11 @@ impl KeyValueRedis {
 #[async_trait]
 impl StoreManager for KeyValueRedis {
     async fn get(&self, _name: &str) -> Result<Arc<dyn Store>, Error> {
+        let client = Client::open(self.database_url.clone()).map_err(log_error)?;
```
itowlson, line 29: This seems to create a `Client` even if `self.connection` already contains a connection?  On every single `StoreManager::get`?

(Oh, this is so extra connections can be created for CAS operations.  I still kinda feel we should be able to reuse the client we created for the original connection, but maybe I'm off the mark there.)

devigned:
@michelleN, please take a look when you have a minute. Thank you!

crates/key-value-redis/src/store.rs
```diff
@@ -98,4 +108,110 @@ impl Store for RedisStore {
...
+    }
+
+    async fn increment(&self, key: String, delta: i64) -> Result<i64, Error> {
+        self.connection
+            .lock()
+            .await
+            .incr(key, delta)
+            .await
+            .map_err(log_error)
+    }
```
itowlson, line 146: Heh, I look at these implementations compared to the others, and it becomes _very_ clear what the `atomics` and `batch` interfaces were modelled on...!

crates/factor-key-value/src/host.rs
```diff
@@ -141,12 +177,235 @@ impl key_value::HostStore for KeyValueDispatch {
...
+    ) -> Result<std::result::Result<(), CasError>> {
+        let cas_rep = cas_res.rep();
+        let cas = self
+            .get_cas(Resource::<Bucket>::new_own(cas_rep))
+            .map_err(|e| CasError::StoreError(atomics::Error::Other(e.to_string())))?;
+
+        match cas.swap(value).await {
+            Ok(_) => Ok(Ok(())),
+            Err(err) => {
+                if err.to_string().contains("CAS_ERROR") {
```
itowlson, line 383: I'm sure you're more conscious than me that this doesn't feel ideal!  But time runs short and we don't want to have to rework the error system, so I guess we go with it for now and re-assess later.  Could you add a "TODO" or "FIXME" or something like that please?

tests/test-components/components/wasi-key-value/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+        ensure_ok!(store.set("qux", b"yay"));
+        ensure_matches!(keys(&store.list_keys(None)), Ok(c) if c.len() == 2 && c.contains(&"bar".into()) && c.contains(&"qux".into()));
+
+        // Delete everything
+        ensure_ok!(store.delete("bar"));
+        ensure_ok!(store.delete("bar"));
+        ensure_ok!(store.delete("qux"));
+        ensure_matches!(store.exists("bar"), Ok(false));
+        ensure_matches!(store.get("qux"), Ok(None));
+        ensure_matches!(keys(&store.list_keys(None)), Ok(&[]));
```
itowlson, line 43: Can we add some atomics and batch tests here, please, even if they're only basic smoke tests for now?  I know you have tests in the implementations, but just something that verifies that the APIs work _from the guest_ even if we can't practically test their atomicity or batchitude.  (Particularly CAS because resources.)

itowlson (commented):
This looks really good, David, and I'm so grateful to you for taking it on.  I have some comments which I don't think are blockers but would appreciate if you have time to address them.  If not then no worries, I can work on them after the RC.

crates/factor-key-value/src/host.rs
```diff
@@ -141,12 +177,235 @@ impl key_value::HostStore for KeyValueDispatch {
...
+    }
+
+    async fn list_keys(
+        &mut self,
+        self_: Resource<Bucket>,
+        cursor: Option<String>,
+    ) -> Result<wasi_keyvalue::store::KeyResponse, wasi_keyvalue::store::Error> {
+        match cursor {
+            Some(_) => Err(wasi_keyvalue::store::Error::Other(
+                "list_keys: cursor not supported".to_owned(),
```
rylev, line 261: Is there ever a plan to support a cursor?

crates/factor-key-value/src/host.rs
```diff
@@ -141,12 +177,235 @@ impl key_value::HostStore for KeyValueDispatch {
...
+impl wasi_keyvalue::atomics::Host for KeyValueDispatch {
+    #[instrument(name = "spin_key_value.increment", skip(self, bucket, key, delta), err(level = Level::INFO), fields(otel.kind = "client"))]
+    async fn increment(
+        &mut self,
+        bucket: Resource<wasi_keyvalue::atomics::Bucket>,
+        key: String,
+        delta: i64,
+    ) -> Result<i64, wasi_keyvalue::store::Error> {
+        let store = self.get_store_wasi(bucket)?;
+        store.increment(key, delta).await.map_err(to_wasi_err)
```
rylev, line 366: Is `Store::increment` documented as requiring to be atomic?

crates/factor-key-value/src/lib.rs
rylev, line 44: Just a note: We probably don't want to force all embedders of this factor to support wasi key_value. We can leave it as is for now, but we'll probably want to come up with some sort of feature system that allows embedders to choose which interfaces get linked in.

crates/factor-key-value/src/lib.rs
```diff
@@ -131,6 +133,14 @@ impl AppState {
     }
 }
 
+#[async_trait]
```
rylev, line 136: Can we document this trait? What guarantees are implementers expected to uphold?

crates/factor-key-value/src/util.rs
```diff
@@ -237,4 +237,68 @@ impl Store for CachingStore {
...
+        }
+
+        Ok(found)
+    }
+
+    async fn set_many(&self, key_values: Vec<(String, Vec<u8>)>) -> anyhow::Result<(), Error> {
+        let mut state = self.state.lock().await;
+
+        for (key, value) in key_values.clone() {
+            state.cache.put(key.to_owned(), Some(value));
```
rylev, line 271:
```suggestion
            state.cache.put(key, Some(value));
```

crates/factor-key-value/src/util.rs
```diff
@@ -237,4 +237,68 @@ impl Store for CachingStore {
...
+        }
+
+        self.inner.set_many(key_values).await
+    }
+
+    async fn delete_many(&self, keys: Vec<String>) -> anyhow::Result<(), Error> {
+        let mut state = self.state.lock().await;
+
+        for key in keys.clone() {
+            state.cache.put(key.to_owned(), None);
```
rylev, line 281:
```suggestion
            state.cache.put(key, None);
```

crates/key-value-spin/src/store.rs
endocrimes: thankfully single-transaction-many-inserts is equally performant to single-transaction-bulk-insert - and this is easier to read (SQLite bulk inserts require the data source to be a `SELECT`, which is... unintuitive), so I wouldn't personally be too worried about perf here

crates/factor-key-value/src/host.rs
devigned: In what sense? If I give a delta and then ask for it to be applied it will be without regard to the underlying value. The only issue with atomicity would be in the case where one fetches and increments while another has mutated the underlying value. In that case, there should be isolation from each mutation. In the case of Redis, this is handled with the `INCR` instruction. In other cases, it should be done within a transaction or optimistically.

I guest the short answer is that increment should be serializable and consistent.

crates/factor-key-value/src/host.rs
devigned: Yeah, this was GROSS! I'd was hoping someone would have a better idea than `contains`.

crates/factor-key-value/src/host.rs
devigned: This is for data stores that support paging. You could imagine a case where too many keys are in the store and the client would like to fetch them a page at a time. 

Personally, I don't like this. I think listing all keys, possibly across partitions, is a really bad idea.

itowlson: This is an artefact of the original Spin key-value interface not supporting a cursor.  There's no technical reason we couldn't implement one in the WASI implementation (and have the Spin interface sit over that, rather than the other way round).  It's just down to time and effort.

crates/factor-key-value/src/lib.rs
itowlson: As Ryan says, this seems like a system-wide consideration and principle which we should approach from a system-wide perspective.  For example it would apply equally to "embedders shouldn't be forced to support the v1 WITs."  I agree with Michelle, let's capture this but let's not ad-hoc it into each PR.

crates/factor-key-value/src/host.rs
itowlson: @devigned You control the `Cas` trait, I think?  You could give the trait `swap` method a custom error type (such as an enum) that allows for a distinguished "CAS error" case.  (It's internal so can leave as a TODO though!)

crates/key-value-azure/src/store.rs
```diff
@@ -151,6 +160,150 @@ impl Store for AzureCosmosStore {
...
+        let operations = vec![Operation::incr("/value", delta).map_err(log_error)?];
+        let _ = self
+            .client
+            .document_client(key.clone(), &key.as_str())
+            .map_err(log_error)?
+            .patch_document(operations)
+            .await
+            .map_err(log_error)?;
+        let pair = self.get_pair(key.as_ref()).await?;
+        match pair {
```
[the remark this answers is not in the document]
devigned: I'm going to dig a little deeper. It wasn't clear initially that patch supported that.

crates/key-value-azure/src/store.rs
```diff
@@ -151,6 +160,150 @@ impl Store for AzureCosmosStore {
...
+                    }
+                    None => None,
+                }
+            }
+            None => None,
+        };
+
+        match current_value {
+            Some((value, etag)) => {
+                *self.etag.lock().unwrap() = Some(etag);
```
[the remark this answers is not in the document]
devigned: I've added doc comments about etag and its usage in CAS.

crates/key-value-azure/src/store.rs
```diff
@@ -151,6 +160,150 @@ impl Store for AzureCosmosStore {
...
+            .query_cross_partition(true)
+            .max_item_count(1)
+            .into_stream::<Pair>();
+
+        let current_value: Option<(Vec<u8>, String)> = match stream.next().await {
+            Some(r) => {
+                let r = r.map_err(log_error)?;
+                match r.results.first() {
+                    Some((item, attr)) => {
+                        Some((item.clone().value, attr.clone().unwrap().etag().to_string()))
```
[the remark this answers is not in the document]
devigned: It should, but I removed the unwrap anyway.

crates/factor-key-value/src/lib.rs
devigned: I added the linker bindings for batch and atomic.

crates/factor-key-value/src/host.rs
devigned: You are totally right, @itowlson! I'm not sure why that didn't occur to me. I've fixed it by introducing a new error enum.

crates/factor-key-value/src/lib.rs
devigned: added!

crates/key-value-redis/src/store.rs
devigned: Fixed! That was not intended.

crates/key-value-azure/src/store.rs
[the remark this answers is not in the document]
itowlson: Looks like the REST API returns the original document and the Rust SDK throws it all away except the document attributes.

itowlson (approved):
I'm not sure what we can do about the race condition in Cosmos.  The rest looks good to me but I'll leave this with you and Ryan to make a decision on the Cosmos woe.  Thanks again for all your hard work on this!

endocrimes (approved):
Impls seem sound to me - I'd take the cosmos race as a future improvement if the rust sdk makes it difficult (because it's already a net improvement over our current interfaces)

crates/key-value-azure/src/store.rs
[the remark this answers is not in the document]
rylev: @itowlson I think this would be the only one of me comments where we might consider blocking the PR on. What are your thoughts on whether the race here is worth blocking the PR?

devigned:
There's a test failure, but I think it might be unrelated to the PR since it is `outbound_mqtt`. Not saying it wasn't me, just that it probably isn't me. If it was, I'll gladly fix it. Thank you so much for the reviews, folks. I really appreciate y'all.

crates/key-value-azure/src/store.rs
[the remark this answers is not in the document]
itowlson: @rylev My preference is to merge as is.  We can document it as a known issue in the RC, continue investigating between now and release, and assuming no joy, we can either:

* Keep documenting it as a known issue in the Cosmos back end; or
* Change it to _always_ error with a "not supported on this back end" message

before release.

itowlson:
It's not you.  MQTT is a known flake.  Thanks for battling through this!

rylev (approved):

PR #2901, opened 2024-11-04
Ignore MQTT test flake

itowlson:
I _think_ this is right.  Although if it's wrong then I bet it's the one time the flake _doesn't_ whomp us.

### DIFF
--- a/tests/runtime.rs
+++ b/tests/runtime.rs
@@ -28,7 +28,7 @@ mod runtime_tests {
 
     #[test]
     fn conformance_tests() -> anyhow::Result<()> {
-        let config = conformance_tests::Config::new("canary");
+        let config = conformance_tests::Config::new("canary").ignore("outbound-mqtt");
         let conclusion = conformance_tests::run_tests(config, move |test| {
             conformance::run_test(test, &spin_binary())
         })?;


itowlson:
Didn't work!  I'll keep investigating

tests/runtime.rs
```diff
@@ -28,7 +28,7 @@ mod runtime_tests {
 
     #[test]
     fn conformance_tests() -> anyhow::Result<()> {
-        let config = conformance_tests::Config::new("canary");
+        let config = conformance_tests::Config::new("canary").ignore("outbounbd-mqtt");
```
vdice, line 31:
```suggestion
        let config = conformance_tests::Config::new("canary").ignore("outbound-mqtt");
```
Perhaps the typo is preventing it from working?

itowlson:
BECAUSE I MADE A STUPID TYPO THAT IS WHY

tests/runtime.rs
itowlson: I think we realised this at the exact same moment!  Sorry for wasting your time - and thanks!

itowlson:
Of course now it hits two _different_ flakes...!

vdice:
> Of course now it hits two different flakes...!

😂 😭

vdice (approved):

rylev:
Do we know why this is flaky? Are we tracking the flakiness anywhere?

itowlson:
We don't know, and yes, I should raise an issue.  We can look into the Postgres tests while we're at it!

PR #2906, opened 2024-11-05
Fix `@` in URL path confusing the parser

### LINKED ISSUE #2905: incorrect outbound http validation blocks request if url contains '@'
I am trying to debug an issue with one of my spin app, and it seems like spin incorrectly blocks the outbound request if the url contains '@'

here is how my outbound http config looks like in spin.toml:

`allowed_outbound_hosts = ["https://api.github.com", "https://github.com"]`

and 

the url I am trying to make request to is : `https://api.github.com/repos/rajatjindal/spin-plugins/contents/manifests/kube/kube@0.2.0.json?ref=heads%2Fkube-v0.2.0`

On trying to make this request, following error is thrown:

```
2024-11-05T02:13:16.254329Z ERROR spin_runtime_factors: Outbound network destination not allowed: https://0.2.0.json
Warning: A component tried to make an outbound network connection to disallowed destination 'https://0.2.0.json'.
To allow this request, add 'allowed_outbound_hosts = ["https://0.2.0.json"]' to the manifest component section.
Handler returned an error: Error::UnexpectedError("ErrorCode::HttpRequestDenied")
```

The source code for this sample can be found at: https://github.com/rajatjindal/spin-outbound-http-issue/

itowlson:
Fixes #2905.

I've added some tests to cover the behaviour.  However, it will get confused in a different way if the username-password segment contains a forward slash (`ivan:sec/ret@example.com`).  I'm not sure if that's a concern?  @rylev you may know a better way to tackle this in which case feel free to close this!

### DIFF
--- a/crates/factor-outbound-networking/src/config.rs
+++ b/crates/factor-outbound-networking/src/config.rs
@@ -443,12 +443,19 @@ impl OutboundUrl {
         // we can always url encode the authority even if it is already encoded.
         if let Some(at) = url.find('@') {
             let scheme_end = url.find("://").map(|e| e + 3).unwrap_or(0);
-            let userinfo = &url[scheme_end..at];
+            let path_start = url[scheme_end..]
+                .find('/') // This can calculate the wrong index if the username or password contains a '/'
+                .map(|e| e + scheme_end)
+                .unwrap_or(usize::MAX);
 
-            let encoded = urlencoding::encode(userinfo);
-            let prefix = &url[..scheme_end];
-            let suffix = &url[scheme_end + userinfo.len()..];
-            url = format!("{prefix}{encoded}{suffix}");
+            if at < path_start {
+                let userinfo = &url[scheme_end..at];
+
+                let encoded = urlencoding::encode(userinfo);
+                let prefix = &url[..scheme_end];
+                let suffix = &url[scheme_end + userinfo.len()..];
+                url = format!("{prefix}{encoded}{suffix}");
+            }
         }
 
         let parsed = match url::Url::parse(&url) {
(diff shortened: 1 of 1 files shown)


crates/factor-outbound-networking/src/config.rs
```diff
@@ -443,12 +443,19 @@ impl OutboundUrl {
...
         if let Some(at) = url.find('@') {
             let scheme_end = url.find("://").map(|e| e + 3).unwrap_or(0);
-            let userinfo = &url[scheme_end..at];
-
-            let encoded = urlencoding::encode(userinfo);
-            let prefix = &url[..scheme_end];
-            let suffix = &url[scheme_end + userinfo.len()..];
-            url = format!("{prefix}{encoded}{suffix}");
+            let path_start = url[scheme_end..]
+                .find('/') // This will still be foxed if the username or password contains a '/'
```
rylev, line 447: Nit: not sure what you mean by "be foxed" here. Can your reword this?

rylev (approved):
This seems like an improvement though we probably want to at some point throw away this code and use some sort of off-the-shelf URL parser.

crates/factor-outbound-networking/src/config.rs
```diff
@@ -562,6 +569,25 @@ mod test {
...
+    fn outbound_url_handles_at_in_paths() {
+        let url = "https://example.com/file@0.1.0.json";
+        let url = OutboundUrl::parse(url, "https").expect("should have parsed url");
+        assert_eq!("example.com", url.host);
+
+        let url = "https://user:password@example.com/file@0.1.0.json";
+        let url = OutboundUrl::parse(url, "https").expect("should have parsed url");
+        assert_eq!("example.com", url.host);
+
+        let url = "https://user:pass#word@example.com/file@0.1.0.json";
```
lann, line 582: This isn't a valid url, fwiw (can't have an unescaped `#`)...

itowlson: @lann Hmm, this is the example that Ryan was trying to address though (https://github.com/fermyon/spin/pull/2116).  If this can't happen then why do we need all the `@` shenanigans?

lann: WELL WHO REVIEWED THAT PR...oh...

I have no explanation :shrug:

lann: Ah sure, this was mostly in the context of "throw away this code and use some sort of off-the-shelf URL parser".

PR #2929, opened 2024-11-19
CI now needs wasm32-wasip1

itowlson:
PR checks for https://github.com/fermyon/spin/pull/2927 didn't catch this.  But it's now needed for smoke testing the templates.

### DIFF
--- a/.github/actions/spin-ci-dependencies/action.yml
+++ b/.github/actions/spin-ci-dependencies/action.yml
@@ -95,7 +95,7 @@ runs:
         rustup default ${{ inputs.rust-version }}
 
     - name: "Install Wasm Rust target"
-      run: rustup target add wasm32-wasi && rustup target add wasm32-wasip1 &&rustup target add wasm32-unknown-unknown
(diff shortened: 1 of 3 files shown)


fibonacci1729 (approved):

radu-matei:
Not sure if CI failure is transient, re-kicked off the integration test.

radu-matei (approved):

rylev:
Looks like one more spot needs to be updated: 

https://github.com/fermyon/spin/blob/78f3e762c9fee752c451f5e4a7da91b373feb378/.github/workflows/build.yml#L198-L199

itowlson:
@lann Double echo is one of the classic flakes, just no longer as frequent as some of the others.

Thanks for fixing up that last one (and thanks @rylev for finding it), and for the nice tidying of the installs!

vanvuongngo:
I think also the templates need this wasm32-wasip1 target, e.g. the http- and redis-trigger templates for rust:
`command = "cargo build --target wasm32-wasip1 --release"`

PR #2931, opened 2024-11-19
Fix instrumentation name for delete_many

ogghead:


lann (approved):

lann:
Thanks - we have a build issue due to a new Rust version that we're trying to resolve in #2929 that should fix the build here.

Issue #2928, opened 2024-11-19
More forgiving `wasi:http` version detection


itowlson:
We detect a couple of supported RCs of `wasi:http`, plus releases `0.2.0` and `0.2.1`.  However, the `wasi` crate (https://docs.rs/wasi/latest/wasi/) has already moved on to 0.2.2, and will presumably move on to 0.2.3 etc. in other minor versions.  This means users of the `wasi` crate in Spin apps have to pin to an exact patch version instead of the 0.13 minor version (see https://github.com/fermyon/leptos-spin/pull/48#discussion_r1847541217).

What would be the practicalities of enabling all 0.2 versions of `wasi:http`, present and future?

cc @alexcrichton

itowlson:
Closed via #2935.

PR #2935, opened 2024-11-26
Relax wasi 0.2 constraint

ogghead:
* Addresses #2928 by relaxing wasi constraint
* Tested with a Leptos Spin app, verified that this allows using `wasi` crate version 0.13 without locking patch version

### DIFF
--- a/crates/http/src/trigger.rs
+++ b/crates/http/src/trigger.rs
@@ -27,10 +27,8 @@ pub enum HandlerType {
 pub const WASI_HTTP_EXPORT_2023_10_18: &str = "wasi:http/incoming-handler@0.2.0-rc-2023-10-18";
 /// The `incoming-handler` export for `wasi:http` version rc-2023-11-10
 pub const WASI_HTTP_EXPORT_2023_11_10: &str = "wasi:http/incoming-handler@0.2.0-rc-2023-11-10";
-/// The `incoming-handler` export for `wasi:http` version 0.2.0
-pub const WASI_HTTP_EXPORT_0_2_0: &str = "wasi:http/incoming-handler@0.2.0";
-/// The `incoming-handler` export for `wasi:http` version 0.2.1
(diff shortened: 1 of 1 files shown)


crates/http/src/trigger.rs
```diff
@@ -27,10 +27,8 @@ pub enum HandlerType {
 pub const WASI_HTTP_EXPORT_2023_10_18: &str = "wasi:http/incoming-handler@0.2.0-rc-2023-10-18";
 /// The `incoming-handler` export for `wasi:http` version rc-2023-11-10
 pub const WASI_HTTP_EXPORT_2023_11_10: &str = "wasi:http/incoming-handler@0.2.0-rc-2023-11-10";
-/// The `incoming-handler` export for `wasi:http` version 0.2.0
-pub const WASI_HTTP_EXPORT_0_2_0: &str = "wasi:http/incoming-handler@0.2.0";
-/// The `incoming-handler` export for `wasi:http` version 0.2.1
-pub const WASI_HTTP_EXPORT_0_2_1: &str = "wasi:http/incoming-handler@0.2.1";
+/// The `incoming-handler` export prefix for all `wasi:http` 0.2 versions
+pub const WASI_HTTP_EXPORT_0_2: &str = "wasi:http/incoming-handler@0.2";
```
itowlson, line 31:
```suggestion
pub const WASI_HTTP_EXPORT_0_2_PREFIX: &str = "wasi:http/incoming-handler@0.2";
```

itowlson (commented):
Hah, I _just this moment_ thought, "eh looks like Darwin doesn't have bandwidth, I'll pick it up."  My apologies for doubting you!

One "matter of taste" suggestion but otherwise looks good.  Thanks!

itowlson (approved):
Ugh that was meant to be an approve

crates/http/src/trigger.rs
ogghead: Hah! Great minds think alike eh? I was deep in the leptos_wasi world for a couple days there but glad to be digging into Spin again and that this was quick to update :)

PR #2937, opened 2024-11-27
Update snippet for the Staic Fileserver template so 'spin add' also use with 0.3.0

mikkelhegn:


lann:
I believe we can update the `spin/templates/v3.0` tag in-place for template fixes like this.

mikkelhegn:
> I believe we can update the `spin/templates/v3.0` tag in-place for template fixes like this.

@lann Is this done, or still something that's needed to be done? And also I have no idea how to do it :-)

itowlson:
I have bumped the tag.

PR #2948, opened 2024-12-05
ref(*): replace wasm32-wasi with wasm32-wasip1

vdice:
- Replaces the remaining uses of the wasm32-wasi target with wasm32-wasip1 per https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html#renaming-wasm32-wasi-to-wasm32-wasip1
  - This includes corresponding 'spin doctor' logic, build.rs, examples and other updates


itowlson (approved):
Thanks for undertaking this thrill ride of a PR!

PR #2950, opened 2024-12-11
Fix C and Zig templates

itowlson:
We uh apparently got a bit uh overzealous in #2948.

Unfortunately, the integration tests tend not to catch template errors; I think they are testing for if the runtime breaks the templates in main, rather than testing the templates in the PR?


itowlson:
The failed integration tests are the existing borked templates being tested from main, and a known flake.

rylev (approved):

vdice (approved):
Whoops, thank you @itowlson!

> really-really-really-wanna-zig-to-pass-integration-tests

I'd say it might've almost been worth the build failures...

PR #2953, opened 2024-12-13
Azure Cosmos multiple stores per container

### LINKED ISSUE #2951: [Azure Cosmos Key-Value] Add support for multiple stores in a single container
## The problem 

Currently the Azure Cosmos implementation of key-value assumes a 1 to 1 relationship between a key-value store and CosmosDB container. While this makes sense at first glance, it might be too limiting for many use cases. CosmosDB has [a limit on the number of database/container combinations](https://learn.microsoft.com/en-us/azure/cosmos-db/concepts-limits) one can have: 500 per account. This means that the current implementation will only support 500 stores. 

## A high level solution

Alternatively, the implementation could be changed to store many different stores in a single container. This could potentially mean that for multi-tenant use cases of Spin, all key-value pairs would be stored in a single container. Single tenant users who know they will never exceed more than 500 stores can continue using a single container per store or switch to multiple stores per container if they wish to.

## Details 

In order to model multiple stores in a single container, we'll need to add additional information to each key/value pair. Instead of only storing the key and value as the current implementation does, we'll also want to store the store name. Since store names are not globally unique, we also want to allow storing some additional globally unique application identifier. For single tenant users this can be left blank or hardcoded to any value, but for multi-tenant users who care about supporting multiple apps, they can store some globally unique app identifier. 

cc @radu-matei and @devigned who have both worked on the current CosmosDB  implementation for their thoughts.

rylev:
Fixes #2951 

This adds support in the Azure Cosmos key-value implementation for storing multiple key-value stores in a single Azure Cosmos container. 

**IMPORTANT**: This currently does not change the Spin CLI experience at all. Users of Spin CLI continue to have the 1 to 1 relationship between containers and key-value stores. This only allows Spin runtime embedders the ability to have many stores per containers. We can consider changing the default behavior for the Spin CLI in the future, but this might need to be at a major version change. 

For embedders who provide an `app_id` to the `KeyValueAzureCosmos`, stores will be created in the supplied container and each key/value pair will include a partition_key field in the form of "$app_id/$store_id".

cc @devigned

### DIFF
--- a/crates/key-value-azure/src/store.rs
+++ b/crates/key-value-azure/src/store.rs
@@ -1,6 +1,5 @@
 use anyhow::Result;
 use azure_data_cosmos::prelude::Operation;
-use azure_data_cosmos::resources::collection::PartitionKey;
 use azure_data_cosmos::{
     prelude::{AuthorizationToken, CollectionClient, CosmosClient, Query},
     CosmosEntity,
@@ -13,6 +12,11 @@ use std::sync::{Arc, Mutex};
 
 pub struct KeyValueAzureCosmos {
     client: CollectionClient,
+    /// An optional app id
+    ///
+    /// If provided, the store will handle multiple stores per container using a
+    /// partition key of `/$app_id/$store_name`, otherwise there will be one container
+    /// per store, and the partition key will be `/id`.
     app_id: Option<String>,
 }
 
@@ -97,7 +101,7 @@ impl StoreManager for KeyValueAzureCosmos {
     async fn get(&self, name: &str) -> Result<Arc<dyn Store>, Error> {
         Ok(Arc::new(AzureCosmosStore {
             client: self.client.clone(),
-            partition_key: self.app_id.as_ref().map(|i| format!("{i}/{name}")),
+            store_id: self.app_id.as_ref().map(|i| format!("{i}/{name}")),
         }))
     }
 
@@ -117,10 +121,10 @@ impl StoreManager for KeyValueAzureCosmos {
 #[derive(Clone)]
 struct AzureCosmosStore {
     client: CollectionClient,
-    /// An optional partition key to use for all operations.
+    /// An optional store id to use as a partition key for all operations.
     ///
-    /// If the partition key is not set, the store will use `/id` as the partition key.
-    partition_key: Option<String>,
+    /// If the store id not set, the store will use `/id` as the partition key.
+    store_id: Option<String>,
 }
 
 #[async_trait]
@@ -134,7 +138,7 @@ impl Store for AzureCosmosStore {
         let pair = Pair {
             id: key.to_string(),
             value: value.to_vec(),
-            partition_key: self.partition_key.clone(),
+            store_id: self.store_id.clone(),
         };
         self.client
             .create_document(pair)
@@ -148,7 +152,7 @@ impl Store for AzureCosmosStore {
         if self.exists(key).await? {
             let document_client = self
                 .client
-                .document_client(key, &self.partition_key)
+                .document_client(key, &self.store_id)
                 .map_err(log_error)?;
             document_client.delete_document().await.map_err(log_error)?;
         }
@@ -201,7 +205,7 @@ impl Store for AzureCosmosStore {
         let operations = vec![Operation::incr("/value", delta).map_err(log_error)?];
         let _ = self
             .client
-            .document_client(key.clone(), &self.partition_key)
+            .document_client(key.clone(), &self.store_id)
             .map_err(log_error)?
             .patch_document(operations)
             .await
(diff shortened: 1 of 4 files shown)


crates/key-value-azure/src/store.rs
```diff
@@ -86,15 +88,16 @@ impl KeyValueAzureCosmos {
...
     }
 }
 
 #[async_trait]
 impl StoreManager for KeyValueAzureCosmos {
-    async fn get(&self, _name: &str) -> Result<Arc<dyn Store>, Error> {
+    async fn get(&self, name: &str) -> Result<Arc<dyn Store>, Error> {
         Ok(Arc::new(AzureCosmosStore {
             client: self.client.clone(),
+            partition_key: self.app_id.as_ref().map(|i| format!("{i}/{name}")),
```
calebschoepp, line 100: If `app_id` is None will this incorrectly leave a leading forward slash in the partition key?

calebschoepp: I see below that it seems like the leading slash is okay, but I'll leave the question here to double check.

crates/key-value-azure/src/store.rs
```diff
@@ -227,21 +228,43 @@ impl Store for AzureCosmosStore {
...
+    }
+
+    fn append_partition_key(&self, query: &mut String) {
+        if let Some(pk) = &self.partition_key {
+            query.push_str(" AND c.partition_key='");
+            query.push_str(pk);
+            query.push('\'')
+        }
+    }
+}
```
calebschoepp, line 258: Why are these redefined here?

calebschoepp (approved):
I have very little familiarity with this code, but reading through it it seems plausibly correct!

crates/key-value-azure/src/store.rs
rylev: If `app_id` is `None` then the `partition_key` will also be `None` which will cause the default of "/id" to be used.

crates/key-value-azure/src/store.rs
rylev: Laziness. I'll try to factor out the common impl.

crates/key-value-azure/src/store.rs
```diff
@@ -318,55 +342,90 @@ impl AzureCosmosStore {
...
+        append_partition_key_condition(query, self.partition_key.as_deref());
+    }
+}
+
+/// Appends an option partition key condition to the query.
+fn append_partition_key_condition(query: &mut String, partition_key: Option<&str>) {
+    if let Some(pk) = partition_key {
+        query.push_str(" AND c.partition_key='");
+        query.push_str(pk);
+        query.push('\'')
```
kate-goldenring, line 411: What is the purpose of this trailing `\`?

kate-goldenring (approved):
LGTM. Love that the enables users to get more out of one cosmos container. It would be nice to add some unit tests, maybe with something like `mockall` to assert queries and partitions are correctly set.

crates/key-value-azure/src/store.rs
```diff
@@ -114,13 +117,10 @@ impl StoreManager for KeyValueAzureCosmos {
...
-struct CompareAndSwap {
-    key: String,
-    client: CollectionClient,
-    bucket_rep: u32,
-    etag: Mutex<Option<String>>,
+    /// An optional partition key to use for all operations.
+    ///
+    /// If the partition key is not set, the store will use `/id` as the partition key.
+    partition_key: Option<String>,
 }
```
kate-goldenring, line 124: Looks like `get` and `get_many` still need to be updated to be partition specific. `get_pair` still queries across all partitions:
```rust
    async fn get_pair(&self, key: &str) -> Result<Option<Pair>, Error> {
        let query = self
            .client
            .query_documents(Query::new(self.get_query(key)))
            .query_cross_partition(true)
            .max_item_count(1);
```

kate-goldenring (changes requested):
I am testing this locally and gets don't seem to be working as expected despite the query being formulated as expected:
```sh
SELECT * FROM c WHERE c.id='wow' AND c.partition_key='foo/default'
```

(I hard coded app id into spin to test this out for a default store)

kate-goldenring (approved):
I realized what my issue was. The implementation assumes that the container "Partition Key" is set to /partition_key. You mention this in the description but i skipped over it accidentally. Maybe naming the partition key something else such as "store_identifier" would help.

crates/key-value-azure/src/store.rs
rylev: We're pushing single quote (`'`) on to the end of the query (to close the single quote that comes right after the `=`). The `\` escapes the `'` character since characters are always themselves enclosed in `'`'.

crates/key-value-azure/src/store.rs
rylev: I'm not sure I understand your question. Can you reformulate it?

rylev:
I've switched from the `partition_key` partition key to `store_id`. Hopefully things are a bit clearer. I'm going to merge now to unblock some other work, and I'll add tests in a new PR.

PR #2957, opened 2024-12-15
Install templates from remote tarball

### LINKED ISSUE #2955: Install templates from a remote tarball
Broken out from #2555 @jandubois: 

> **There is no way to install templates from a remote tarball**
>
> Since we [sc. Rancher Desktop] can't rely on git being installed, we need to install using `--dir` from a local directory because there is no option to install from a remote tarball like this:
>
> `spin templates install --from https://github.com/fermyon/spin/archive/refs/heads/main.tar.gz`
>
> That means we need to download the file ourselves, unpack it into a temporary directory, install from there, and then remove the temporary directory again.
>
> Since we are doing this from a shell script, it means we need curl or wget, neither of which are guaranteed to be installed.

itowlson:
Fixes #2955.

```
$ rm -rf ~/.local/share/spin/templates/
$ spin templates install --tar https://github.com/fermyon/spin/archive/refs/tags/v3.0.0.tar.gz
Copying remote template source
Installing template http-go...
Installing template redis-go...
Installing template http-grain...
Installing template http-empty...
Installing template redis-rust...
Installing template static-fileserver...
Installing template http-rust...
Installing template http-zig...
Installing template redirect...
Installing template http-c...
Installing template http-php...
Installed 11 template(s)

+------------------------------------------------------------------------+
| Name                Description                                        |
+========================================================================+
| http-c              HTTP request handler using C and the Zig toolchain |
| http-empty          HTTP application with no components                |
| http-go             HTTP request handler using (Tiny)Go                |
| http-grain          HTTP request handler using Grain                   |
| http-php            HTTP request handler using PHP                     |
| http-rust           HTTP request handler using Rust                    |
| http-zig            HTTP request handler using Zig                     |
| redirect            Redirects a HTTP route                             |
| redis-go            Redis message handler using (Tiny)Go               |
| redis-rust          Redis message handler using Rust                   |
| static-fileserver   Serves static files from an asset directory        |
+------------------------------------------------------------------------+
```

### DIFF
--- a/crates/templates/src/source.rs
+++ b/crates/templates/src/source.rs
@@ -25,6 +25,12 @@ pub enum TemplateSource {
     /// Templates much be in a `/templates` directory under the specified
     /// root.
     File(PathBuf),
+    /// Install from a remote tarball.
+    ///
+    /// Templates should be in a `/templates` directory under the root of the tarball.
+    /// The implementation also allows for there to be a single root directory containing
+    /// the `templates` directory - this makes it compatible with GitHub release tarballs.
+    RemoteTar(Url),
 }
 
 /// Settings for installing templates from a Git repository.
@@ -72,6 +78,9 @@ impl TemplateSource {
                     None
                 }
             }
+            Self::RemoteTar(url) => Some(crate::reader::RawInstalledFrom::RemoteTar {
+                url: url.to_string(),
+            }),
         }
     }
 
@@ -96,6 +105,7 @@ impl TemplateSource {
         match self {
             Self::Git(git_source) => clone_local(git_source).await,
             Self::File(path) => check_local(path).await,
+            Self::RemoteTar(url) => download_untar_local(url).await,
         }
     }
 
@@ -103,6 +113,7 @@ impl TemplateSource {
         match self {
             Self::Git { .. } => true,
             Self::File(_) => false,
+            Self::RemoteTar(_) => true,
         }
     }
 }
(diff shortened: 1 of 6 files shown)


crates/templates/src/source.rs
```diff
@@ -192,6 +203,84 @@ async fn check_local(path: &Path) -> anyhow::Result<LocalTemplateSource> {
...
+
+    Ok(LocalTemplateSource {
+        root: templates_root,
+        _temp_dir: Some(temp_dir),
+    })
+}
+
+fn bypass_gh_added_root(unpack_dir: PathBuf) -> PathBuf {
+    // GitHub adds a prefix directory to release tarballs (e.g. spin-v3.0.0/...).
+    // We try to locate the repo root within the unpacked tarball.
```
rylev, line 236: Nit: any reason to not make this a doc comment on the function so that this surfaces more easily when code spelunking?

src/commands/templates.rs
```diff
@@ -119,15 +131,19 @@ impl Install {
...
-            (None, Some(dir)) => {
+            (None, Some(dir), None) => {
                 let abs_dir = dir.absolutize().map(|d| d.to_path_buf());
                 TemplateSource::File(abs_dir.unwrap_or_else(|_| dir.clone()))
             }
+            (None, None, Some(tar_url)) => {
+                let url = url::Url::parse(tar_url).context("Invalid URL for remote tar")?;
+                TemplateSource::RemoteTar(url)
+            }
             _ => anyhow::bail!("Exactly one of `git` and `dir` sources must be specified"),
```
rylev, line 147: This error should be updated to reflect the third option

crates/templates/src/source.rs
```diff
@@ -192,6 +203,84 @@ async fn check_local(path: &Path) -> anyhow::Result<LocalTemplateSource> {
     }
 }
 
+async fn download_untar_local(url: &Url) -> anyhow::Result<LocalTemplateSource> {
```
rylev, line 206: Would this make sense as a function on `LocalTemplateSource`? Like `LocalTemplateSource::from_remote_tarball`? Then you can leave a doc comment outlining that it downloads the tarball to a temp dir.

rylev (approved):
I have some nits that I'd love to see addressed, but nothing is blocking.

crates/templates/src/source.rs
itowlson: The other "pull down a remote source" functions are free, so I've followed the same pattern for now.  We can change them all to methods in a separate refactoring pass if we prefer that.  In the meantime I've added the doc comment.

PR #2964, opened 2024-12-16
Work around breaking change in `cross`

itowlson:
`cross` no longer works on Rust 1.79 so we need to move to 1.81.


itowlson:
The Redis tests fall foul of the new "this assumes ! fallback type is ()" lint.  Upgrading to latest Redis crate didn't help.  I guess maybe we need to pick a concrete type of something but I'm proposing to punt on it for now in order to unblock PRs, and circle back to it in our copious spare time.

lann (approved):

PR #2970, opened 2024-12-19
Templates: add a Rust template for non-trigger components

tschneidereit:
This new template makes it easy to create a component that can be used as a dependency in another component. Until now, that required fully manually setting up a component and adding it to the Spin manifest, etc.

### DIFF
--- a/templates/component-rust/content/Cargo.toml.tmpl
+++ b/templates/component-rust/content/Cargo.toml.tmpl
@@ -1,0 +1,16 @@
+[package]
+name = "{{project-name | kebab_case}}"
+authors = ["{{authors}}"]
+description = "{{project-description}}"
+version = "0.1.0"
+rust-version = "1.78"
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+anyhow = "1"
+wit-bindgen = "0.36.0"
+
+[workspace]
--- a/templates/component-rust/content/spin.toml
+++ b/templates/component-rust/content/spin.toml
@@ -1,0 +1,13 @@
+spin_manifest_version = 2
+
+[application]
+name = "{{project-name | kebab_case}}"
+version = "0.1.0"
+authors = ["{{authors}}"]
+description = "{{project-description}}"
+
+[component.{{project-name | kebab_case}}]
+source = "target/wasm32-wasip2/release/{{project-name | snake_case}}.wasm"
+[component.{{project-name | kebab_case}}.build]
+command = "cargo build --target wasm32-wasip2 --release"
+watch = ["src/**/*.rs", "Cargo.toml"]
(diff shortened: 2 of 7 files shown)


templates/component-rust/metadata/snippets/component.txt
```diff
@@ -0,0 +1,6 @@
+[component.{{project-name | kebab_case}}]
+source = "{{ output-path }}/target/wasm32-wasip2/release/{{project-name | snake_case}}.wasm"
```
itowlson, line 2: We use wasip1 everywhere else - is p2 stable enough that we should be moving to it?

templates/component-rust/content/spin.toml
```diff
@@ -0,0 +1,13 @@
+spin_manifest_version = 2
```
itowlson, line 1: It feels extremely awkward to create a `spin.toml` for a library component.  Why does a dependency need a Spin manifest file?  It can't be referenced by a trigger.  It can't be referenced by its component ID (although we could allow that of course).  Most of its settings are misleading e.g. setting allowed outbound hosts looks like it grants access but actually does nothing.  Is this just to opt it into getting included `spin build`?

I feel like the ask here is for the manifest to provide for building and referencing library components but I am not sold that mixing them in with "application" components is the way to go right now: perhaps we should step back and ask how we should represent library components in the manifest, then create a template as part of that?

templates/component-rust/metadata/spin-template.toml
```diff
@@ -0,0 +1,12 @@
+manifest_version = "1"
+id = "component-rust"
+description = "Pure Rust component to be used as a dependency for other component"
```
itowlson, line 3: Remove "pure" e.g.

```suggestion
description = "Library (dependency) component using Rust"
```

or something like that?

templates/component-rust/content/Cargo.toml.tmpl
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "{{project-name | kebab_case}}"
+authors = ["{{authors}}"]
+description = "{{project-description}}"
+version = "0.1.0"
+rust-version = "1.78"
```
itowlson, line 6: What is the status of the wasm32-wasip2 target in 1.78?

templates/component-rust/metadata/snippets/component.txt
tschneidereit: It's stable, and allows us to build components without using `cargo component`. So yes, I do think it makes sense to move to it where feasible. I also don't think that we need to be particularly principled about it, and that it's fine to do this in an ad-hoc fashion :)

templates/component-rust/content/spin.toml
tschneidereit: Agreed! I didn't know that it's possible to create templates that don't do this. I'll remove the manifest.

tschneidereit:
@itowlson thank you for the comments! I addressed the smaller ones in a new push as much as I could. I tried to make `new_application.supported = false` work, but the template still shows up in the list for `spin new`. Is there some other magic incantation I'm missing?

itowlson:
I'll have a look, that might be a bug with me not checking new/add compat until after template selection.

PR #2971, opened 2024-12-19
Fix deny-all adapter for socket imports

### LINKED ISSUE #2969: Better error messages when component dependencies are not allowed to inherit configuration
Using Spin 3.1.0.

My Spin component has a dependency on a component which need access to an inference model. I did not configure `dependencies_inherit_configuration = true` for the Spin component, so the dependency would not get access to the model. The error message I was getting was not helping me track down the problem:

```
> spin up
Storing default key-value data to ".spin/sqlite_key_value.db".
Error: failed to resolve dependencies for component "challenge1"

Caused by:
   0: an error occurred when preparing dependencies
   1: encoding produced a component that failed validation
   2: type mismatch for import `wasi:sockets/udp-create-socket@0.2.0`
type mismatch in instance export `udp-socket`
resource types are not the same (ResourceId { globally_unique_id: 3, contextually_unique_id: 21 } vs. ResourceId { globally_unique_id: 3, contextually_unique_id: 102 }) (at offset 0x227e1ee)
```

The dependent component was built in Python, using componentize-py.

fibonacci1729:
Fixes #2969


lann (commented):
lgtm but since we paired on the fix another set of eyes would be good

itowlson (approved):
LGTM but since I am as ignorant as apples... _grin_

Should we do a 3.1 patch for this?

fibonacci1729:
Yes a patch makes sense. I can't get to it today but I can cut a release tomorrow morning.

PR #2976, opened 2025-01-09
fix #2974 -- use redis::aio::ConnectionManager

### LINKED ISSUE #2974: io error: broken pipe when accessing redis-backed KV store
I have an application ([tfenster/verified-bluesky](https://github.com/tfenster/verified-bluesky)) that worked fine on your cloud, but due to the limit of 1024 entries to the KV store, I moved it to Azure (Azure Kubernetes Service, Azure Cache for Redis). Since then, it works for some time (didn't exactly check, but seems like a few minutes), but then calls to the KV store fail with "io error: broken pipe". I have narrowed it down to the following piece of code as the easiest repro, but other places accessing the store also fail with the same error:

```
package main

import (
	"encoding/json"
	"fmt"
	"net/http"
	"strings"

	spinhttp "github.com/fermyon/spin/sdk/go/v2/http"
	"github.com/fermyon/spin/sdk/go/v2/kv"
)

func init() {
	spinhttp.Handle(func(w http.ResponseWriter, r *http.Request) {
		switch r.Method {
		case http.MethodGet:
			fmt.Println("Getting stats")
			store, err := kv.OpenStore("default")
			if err != nil {
				http.Error(w, "Error opening kv store "+err.Error(), http.StatusInternalServerError)
				return
			}
			defer store.Close()

			fmt.Println("Getting keys")
			keys, err := store.GetKeys()
			if err != nil {
				http.Error(w, "Error getting keys from kv store "+err.Error(), http.StatusInternalServerError)
				return
			}
...
		}
	})
}

func main() {}
```

The call to `kv.OpenStore` still seems to work, but `store.GetKeys()` fails and I get this response:

```
HTTP/1.1 500 Internal Server Error
Date: Sun, 05 Jan 2025 08:23:35 GMT
Content-Type: text/plain; charset=utf-8
Content-Length: 55
Connection: close
x-content-type-options: nosniff
Strict-Transport-Security: max-age=31536000; includeSubDomains

Error getting keys from kv store io error: broken pipe
```

For full context, see [main.go](https://github.com/tfenster/verified-bluesky/blob/main/stats/main.go). The logs only show this

```
Getting stats
Getting keys
```

The setup of the redis backend and creation of the runtime config looks like this

```
az redis create -n $redisName -g $rgAndClusterName -l $location --sku Basic --vm-size c0 --redis-configuration '{ "maxmemory-policy": "noeviction" }'
$redisKeys = $(az redis list-keys -n $redisName -g $rgAndClusterName) | ConvertFrom-Json

@"
[key_value_store.default]
type = "redis"
url = "rediss://:$($redisKeys.primaryKey)@$redisName.redis.cache.windows.net:6380"
"@ | Out-File ./runtime-config-redis.toml
```

I then bring that to this secret (content obviously redacted)

```
apiVersion: v1
kind: Secret
metadata:
  name: verified-bluesky-runtime-config
type: Opaque
data:
  runtime-config.toml: ...
```

Deploying the spin app looks like this [spinapp.yaml](https://github.com/tfenster/verified-bluesky/blob/main/deployment/spinapp.yaml) and the latest build to create and push the OCI artifact is [here](https://github.com/tfenster/verified-bluesky/actions/runs/12617707923/job/35160266826).

Let me know if I can share anything else

Not sure if that is relevant, but this is the output of the commands you asked for in the template, taken from my dev environment:
* Spin version (`spin --version`): 
```
vscode ➜ /workspaces/verified-bluesky/deployment (main) $ spin --version
spin 3.1.1 (aa919ce 2024-12-20)
```
```
vscode ➜ /workspaces/verified-bluesky/deployment (main) $ spin plugins list --installed
cloud 0.10.0 [installed]
kube 0.3.1 [installed]
```

fibonacci1729:
Fixes #2974

This replaces the use of `redis::aio::MultiplexedConnection` with `redis::aio::ConnectionManager` which automatically attempts to retry connecting upon error. See [this](https://docs.rs/redis/latest/redis/aio/struct.ConnectionManager.html#behavior) for behavior of `ConnectionManager`. To handle *not* passing on the first error to the user as described by "When a command sent to the server fails with an error that represents a “connection dropped” condition, that error will be passed on to the user..." relevant implementations of `Store` now implement `after_open()` which will to try ping the connection to trigger the automatic retry.

### DIFF
--- a/crates/key-value-redis/src/store.rs
+++ b/crates/key-value-redis/src/store.rs
@@ -1,15 +1,14 @@
 use anyhow::{Context, Result};
-use redis::{aio::MultiplexedConnection, parse_redis_url, AsyncCommands, Client, RedisError};
+use redis::{aio::ConnectionManager, parse_redis_url, AsyncCommands, Client, RedisError};
 use spin_core::async_trait;
 use spin_factor_key_value::{log_error, Cas, Error, Store, StoreManager, SwapError};
-use std::ops::DerefMut;
 use std::sync::Arc;
-use tokio::sync::{Mutex, OnceCell};
+use tokio::sync::OnceCell;
 use url::Url;
 
 pub struct KeyValueRedis {
     database_url: Url,
-    connection: OnceCell<Arc<Mutex<MultiplexedConnection>>>,
+    connection: OnceCell<ConnectionManager>,
 }
 
 impl KeyValueRedis {
@@ -30,10 +29,8 @@ impl StoreManager for KeyValueRedis {
             .connection
             .get_or_try_init(|| async {
                 Client::open(self.database_url.clone())?
-                    .get_multiplexed_async_connection()
+                    .get_connection_manager()
                     .await
-                    .map(Mutex::new)
-                    .map(Arc::new)
             })
             .await
             .map_err(log_error)?;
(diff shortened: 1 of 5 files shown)


dicej (approved):
LGTM.  It's a little weird that we have to clone the `ConnectionManager` to call any `redis::Commands` methods, but I guess it makes more sense than adding a layer of `Arc<Mutex<...>>` on top of it.

crates/key-value-redis/src/store.rs
```diff
@@ -55,90 +52,69 @@ impl StoreManager for KeyValueRedis {
...
-    connection: Arc<Mutex<MultiplexedConnection>>,
+    connection: ConnectionManager,
     bucket_rep: u32,
 }
 
 #[async_trait]
 impl Store for RedisStore {
+    async fn after_open(&self) -> Result<(), Error> {
+        if let Err(_error) = self.connection.clone().ping::<()>().await {
+            //If IO error happens, ConnectionManager will start reconneciton in background
```
itowlson, line 69:
```suggestion
            //If IO error happens, ConnectionManager will start reconnection in background
```

crates/key-value-redis/src/store.rs
```diff
@@ -55,90 +52,69 @@ impl StoreManager for KeyValueRedis {
...
+            //so we do not take any action and just pray re-connection will be successful.
```
itowlson, line 70: Boy the enterprise adopters are gonna love that "just pray"

Is it worth emitting an info or trace log that the connection expired and we've pinged for a reconnect?

crates/key-value-redis/src/store.rs
```diff
@@ -55,90 +52,69 @@ impl StoreManager for KeyValueRedis {
...
     key: String,
-    connection: Arc<Mutex<MultiplexedConnection>>,
+    connection: ConnectionManager,
     bucket_rep: u32,
 }
 
 #[async_trait]
 impl Store for RedisStore {
+    async fn after_open(&self) -> Result<(), Error> {
+        if let Err(_error) = self.connection.clone().ping::<()>().await {
```
itowlson, line 68: Is there a reason for doing this at store open time instead of just-in-time if a command fails?  The "just pray" thing makes me worry we have a race condition if we don't allow recovery during a command...?

fibonacci1729: If we do it per command, then there is a round-trip cost incurred for each request. Doing it this way would at least trigger the `ConnectionManager` to kickoff an automatic reconnect before any commands are issued.

crates/key-value-redis/src/store.rs
fibonacci1729: I was going to and noticed that `key-value-redis` didn't log anywhere :trollface: 

Happy to bring log or tracing into this crate instead of gobbling the error.

itowlson: Nah no worries, leave it for now

crates/key-value-redis/src/store.rs
itowlson: Well I was thinking do it if a command _failed_ rather than on _every_ command, so it would only incur the ping cost if actually needed rather than on every open, and that would also handle the situation where great Zeus on high Olympus spurned the praying.  (Goats.  You need more goats.)  But I'm probably imagining situations that never happen in practice - very happy to get this into users' hands as is.

fibonacci1729: I hear you. It seems like we'd be swapping one potential race for another or at least have to implement a command retry mechanism on top of each request to ache8ve that which seems like alot more code 🫣

If you're ok with it, I think I'd like to land this as-is and revisit if weirdness persists.

itowlson: :shipit: :shipit: :shipit:

itowlson (approved):
did I mention :shipit:

Issue #2184, opened 2023-12-15
Logging fails on TinyGo Spin Apps


kate-goldenring:
Problem:
TinyGo Spin apps that use `log` gets the following runtime error from Wasmtime:

```sh
$ spin build --up
Building component go-app with `tinygo build -target=wasi -gc=leaking -no-debug -o main.wasm main.go`
Finished building all Spin components
Logging component stdio to ".spin/logs/"

Serving http://127.0.0.1:3000
Available Routes:
  go-app: http://127.0.0.1:3000 (wildcard)
panic: runtime error: nil pointer dereference
2023-12-15T23:25:33.689230Z ERROR spin_trigger_http: Error processing request: error while executing at wasm backtrace:
    0: 0xeb05 - <unknown>!runtime.runtimePanicAt.llvm.3901647063377195771
    1: 0x33a77 - <unknown>!byn$mgfn-shared$runtime.lookupPanic.llvm.3901647063377195771
    2: 0x574a - <unknown>!runtime.nilPanic.llvm.3901647063377195771
    3: 0xcfe7 - <unknown>!(*sync.Mutex).Lock
    4: 0x32ca5 - <unknown>!spin_http_handle_http_request
    5: 0x33870 - <unknown>!__wasm_export_spin_http_handle_http_request
    6: 0x441b8 - wit-component:adapter:wasi_snapshot_preview1!fermyon:spin/inbound-http#handle-request

Caused by:
    wasm trap: wasm `unreachable` instruction executed    
```
## Example app:
```go
package main

import (
	"fmt"
	"log"
	"net/http"

	spinhttp "github.com/fermyon/spin/sdk/go/v2/http"
)

func init() {
	spinhttp.Handle(func(w http.ResponseWriter, r *http.Request) {
		log.Printf("THIS WILL CAUSE AN ERROR")

		fmt.Fprintln(w, "== THIS WILL NOT BE REACHED ==")
	})
}

func main() {}
```

* Spin version (`spin --version`)
 `spin 2.0.1 (1d72f1c 2023-11-10)`

rajatjindal:
possible similar issues here: 

https://github.com/fermyon/spin/issues/1259 and https://github.com/tinygo-org/tinygo/issues/3798

initializing a logger instance and using that for logging fixes the issue:

```
package main

import (
	"fmt"
	"log"
	"net/http"
	"os"

	spinhttp "github.com/fermyon/spin/sdk/go/v2/http"
)

func init() {
[...]

PR #2977, opened 2025-01-12
Update TinyGo templates to address static init problems

itowlson:
E.g. this should fix static logging objects not being ready, os.Environ(), etc.

See https://github.com/fermyon/spin/issues/2184#issuecomment-2585955161.

Marking as draft initially to verify that I haven't muddled up the changes and that I've updated CI correctly.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -9,12 +9,12 @@ FROM mcr.microsoft.com/vscode/devcontainers/rust:0-${VARIANT}
 #     && apt-get -y install --no-install-recommends <your-package-list-here>
 
 # Go installation, see https://go.dev/doc/install
-ARG GO_URL="https://go.dev/dl/go1.20.1.linux-amd64.tar.gz"
+ARG GO_URL="https://go.dev/dl/go1.22.0.linux-amd64.tar.gz"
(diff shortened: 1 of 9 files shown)


itowlson:
It was Go 1.21 that introduced wasip1 but that had a lot of patches so I've bumped templates and CI to 1.22.  If we'd prefer to be more conservative and use 1.21 instead then I can do that.

itowlson:
CI failed in `cabi_realloc`.  It works locally for me, but a possible difference is that I'm on Go 1.23 rather than 1.22.  Trying updating Go in CI and will see how it goes.

itowlson:
Hmm, it seems like the smoke test is trying to test the template in the _repo_ not the template in the _PR._  Which seems like it makes the CI failure even stranger.  Maybe the existing template doesn't work with tinygo 0.35 because of the `wasi` build target?

itowlson:
Yes, running the _existing_ Go template with TinyGo 0.35 / Go 1.23 reproduces the CI error.

I don't know how to validate the new template in CI so this may have to involve people accepting a Works On My Machine.

I'll roll CI back to Go 1.22, and also remove the stub `main` that's no longer needed.

dicej (approved):

PR #2981, opened 2025-01-14
fix(templates/http-go): Remove scheduler flag from http-go build command

adamreese:
This removes the `-scheduler=none` flag from the build command in the http-go template. Including this flag prevents Spin applications using `http.Transport` from compiling.

```
❯ spin build
Building component http-roundtrip-test with `tinygo build -target=wasip1 -gc=leaking -scheduler=none -buildmode=c-shared -no-debug -o main.wasm .`
/opt/homebrew/Cellar/tinygo/0.35.0/src/net/http/transfer.go:213:2: attempted to start a goroutine without a scheduler
Error: Build command for component http-roundtrip-test failed with status Exited(1)
```

### DIFF
--- a/crates/componentize/src/lib.rs
+++ b/crates/componentize/src/lib.rs
@@ -440,8 +440,9 @@ mod tests {
         let mut cmd = process::Command::new("tinygo");
         cmd.arg("build")
             .current_dir("tests/go-case")
-            .arg("-target=wasi")
+            .arg("-target=wasip1")
             .arg("-gc=leaking")
+            .arg("-buildmode=c-shared")
             .arg("-no-debug")
             .arg("-o")
             .arg(out_dir.join("go_case.wasm"))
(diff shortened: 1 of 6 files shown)


itowlson:
@deadprogram I blindly copied `-scheduler=none` from your sample build command in https://github.com/fermyon/spin/issues/2184#issuecomment-2585955161 - would you have any concerns with us removing it?

deadprogram:
> would you have any concerns with us removing it?

Not at all! Please remove for compatibility.

I am not using any Go routines in my specific use cases, so I was able to remove even more size from the module that way.

itowlson (approved):

itowlson (approved):

itowlson (approved):
Man, thanks for finding all this stuff.  Sorry for leaving it in this state for you.

adamreese:
It took some artisanal greping to find them all

itowlson:
@fermyon/administrators Please force merge.  I believe the JS CI failures are unrelated, the double-echo is a known flake, and the Go one is because it's using the templates from main, which this PR fixes.  Thanks!

Issue #2952, opened 2024-12-13
Key-Value Swallows Write Errors When Backing Impl Fails


rylev:
The `CachingStore` completely swallows errors when the inner implementation fails on writes.

https://github.com/fermyon/spin/blob/1f2269b5dd074fe953c1a8e46caf363d713fc984/crates/factor-key-value/src/util.rs#L190

This can lead to writes appearing like they succeed (even being reflected by reads within the same process) but not actually working.

lann:
FWIW, it looks like this was an intentional tradeoff: https://github.com/fermyon/spin/blob/1f2269b5dd074fe953c1a8e46caf363d713fc984/crates/factor-key-value/src/util.rs#L75-L80

lann:
After some further investigation, it looks like some of the newer KV implementations (Azure, AWS) are using lazy connections that cannot fail on `open`, which makes this problem much worse.

dicej:
I'm the one who implemented `CachingStore` and wrote the comment @lann referenced above.  My intention was that the Spin implementation would support the weakest consistency and durability model allowed by the interface in key-value.wit so as to help the application developer incorporate that model into _their_ design and implementation early on and give them confidence that their app would be portable to _any_ KV implementation.

Unfortunately, the consistency and durability model was apparently never documented in [key-value](https://github.com/fermyon/spin/blob/main/wit/deps/spin%402.0.0/key-value.wit), and that's definitely a problem.  I'll open a PR to address that.

FWIW, the `wasi-keyvalue` docs [do address this](https://github.com/WebAssembly/wasi-keyvalue/blob/main/wit/store.wit#L11-L25), although those docs could also be improved by adding an explicit discussion of durability as well as consistency.

dicej:
> I assume that these docs are not only talking about subsequent reads in the same component instance but also across component instances. Let me know if this assumption is incorrect.

My reading is different, but I think that's because the docs aren't clear enough about the terms "client", "guest", and "caller" in the paragraph following the one you quoted.  I've put up [a PR](https://github.com/WebAssembly/wasi-keyvalue/pull/56) that reflects my interpretation; I'm happy to refine and/or rewrite it based on feedback.

[...]

PR #2995, opened 2025-01-27
remove `CachingStoreManager` from `factor-key-value`

dicej:
The semantic (non-)guarantees for wasi-keyvalue are still [under discussion](https://github.com/WebAssembly/wasi-keyvalue/pull/56), but meanwhile the behavior of Spin's write-behind cache has caused [some headaches](https://github.com/fermyon/spin/issues/2952), so I'm removing it until we have more clarity on what's allowed and what's disallowed by the proposed standard.

The original motivation behind `CachingStoreManager` was to reflect the anticipated behavior of an eventually-consistent, low-latency, cloud-based distributed store and, per [Hyrum's Law](https://www.hyrumslaw.com/) help app developers avoid depending on the behavior of a local, centralized store which would not match that of a distributed store.  However, the write-behind caching approach interacts poorly with the lazy connection establishment which some `StoreManager` implementations use, leading writes to apparently succeed even when the connection fails.

Subsequent discussion regarding the above issue arrived at a consensus that we should not consider a write to have succeeded until and unless we've successfully connected to and received a write confirmation from at least one replica in a distributed system.  I.e. rather than the replication factor (RF) = 0 we've been effectively providing up to this point, we should provide RF=1. The latter still provides low-latency performance when the nearest replica is reasonably close, but improves upon RF=0 in that it shifts responsibility for the write from Spin to the backing store prior to returning "success" to the application.

Note that RF=1 (and indeed anything less than RF=ALL) cannot guarantee that the write will be seen immediately (or, in the extreme case of an unrecoverable failure, at all) by readers connected to other replicas.  Applications requiring a stronger consistency model should use an ACID-style backing store rather than an eventually consistent one.


lann:
How hard would it be to change the cache from write-behind to write-through?

dicej:
> How hard would it be to change the cache from write-behind to write-through?

Probably not hard.  The question is: do we want caching at all, and if so, how much?

We can always bring it back if/when there's a clear need, but for now the simplest thing seems to be to just remove it.  I also wonder if specific key-value implementations might want to enforce their own caching rules.

lann (approved):
Fair enough. I guess if you're performance-sensitive enough for this to matter then you probably ought to avoid immediately reading your writes anyway :shrug:

itowlson (approved):

Issue #2988, opened 2025-01-23
Feature Request: Add flag to registry push that pushes composed components


fibonacci1729:
With the introduction of component dependencies pre-existing runtimes are unable to reconcile the new Spin locked app format (e.g. SpinKube). This flag would indicate to `spin registry push` to compose Spin components with their dependencies (before publishing to a registry) resulting in an artifact that contains the dependency-erased components. The obvious downside of this approach is that component layers for applications can't be deduplicated in OCI or across the wire; however this flag would assist in running applications that use dependencies in these runtimes.

cc/ @kate-goldenring @michelleN

PR #2996, opened 2025-01-27
feat(push): allow pushing composed components to registry

fibonacci1729:
~WIP for now to flesh out the concept.~

Implements #2988

~Still working to add a test but the meat-and-potatoes is ready to serve.~

### DIFF
--- a/crates/compose/src/lib.rs
+++ b/crates/compose/src/lib.rs
@@ -2,6 +2,7 @@ use anyhow::Context;
 use indexmap::IndexMap;
 use semver::Version;
 use spin_app::locked::{self, InheritConfiguration, LockedComponent, LockedComponentDependency};
+use spin_common::{ui::quoted_path, url::parse_file_url};
 use spin_serde::{DependencyName, KebabId};
 use std::collections::BTreeMap;
 use thiserror::Error;
@@ -40,6 +41,36 @@ pub trait ComponentSourceLoader {
         &self,
         source: &locked::LockedComponentSource,
     ) -> anyhow::Result<Vec<u8>>;
+}
+
+/// A ComponentSourceLoader that loads component sources from the filesystem.
+pub struct ComponentSourceLoaderFs;
+
+#[async_trait::async_trait]
+impl ComponentSourceLoader for ComponentSourceLoaderFs {
+    async fn load_component_source(
+        &self,
+        source: &locked::LockedComponentSource,
+    ) -> anyhow::Result<Vec<u8>> {
+        let source = source
+            .content
+            .source
+            .as_ref()
+            .context("LockedComponentSource missing source field")?;
+
+        let path = parse_file_url(source)?;
+
+        let bytes: Vec<u8> = tokio::fs::read(&path).await.with_context(|| {
+            format!(
+                "failed to read component source from disk at path {}",
+                quoted_path(&path)
+            )
+        })?;
+
+        let component = spin_componentize::componentize_if_necessary(&bytes)?;
+
+        Ok(component.into())
+    }
 }
 
 /// Represents an error that can occur when composing dependencies.
--- a/crates/oci/Cargo.toml
+++ b/crates/oci/Cargo.toml
@@ -22,6 +22,7 @@ reqwest = "0.12"
 serde = { workspace = true }
 serde_json = { workspace = true }
 spin-common = { path = "../common" }
+spin-compose = { path = "../compose" }
 spin-loader = { path = "../loader" }
 spin-locked-app = { path = "../locked-app" }
 tempfile = { workspace = true }
@@ -29,3 +30,8 @@ tokio = { workspace = true, features = ["fs"] }
 tokio-util = { version = "0.7", features = ["compat"] }
 tracing = { workspace = true }
 walkdir = "2"
+
+[dev-dependencies]
+wasm-encoder = { workspace = true }
+wit-component = { workspace = true, features = ["dummy-module"] }
+wit-parser = { workspace = true }
--- a/crates/oci/src/client.rs
+++ b/crates/oci/src/client.rs
@@ -16,9 +16,10 @@ use reqwest::Url;
 use spin_common::sha256;
 use spin_common::ui::quoted_path;
 use spin_common::url::parse_file_url;
+use spin_compose::ComponentSourceLoaderFs;
 use spin_loader::cache::Cache;
 use spin_loader::FilesMountStrategy;
-use spin_locked_app::locked::{ContentPath, ContentRef, LockedApp};
+use spin_locked_app::locked::{ContentPath, ContentRef, LockedApp, LockedComponent};
 use tokio::fs;
 use walkdir::WalkDir;
 
@@ -58,6 +59,7 @@ const DEFAULT_CONTENT_REF_INLINE_MAX_SIZE: usize = 128;
 const DEFAULT_TOKEN_EXPIRATION_SECS: usize = 300;
 
 /// Mode of assembly of a Spin application into an OCI image
+#[derive(Copy, Clone)]
 enum AssemblyMode {
     /// Assemble the application as one layer per component and one layer for
     /// every static asset included with a given component
@@ -65,6 +67,15 @@ enum AssemblyMode {
     /// Assemble the application as one layer per component and one compressed
     /// archive layer containing all static assets included with a given component
     Archive,
+}
+
+/// Indicates whether to compose the components of a Spin application when pushing an image.
+#[derive(Copy, Clone)]
+pub enum ComposeMode {
+    /// Compose components before pushing the image.
+    All,
+    /// Skip composing components before pushing the image.
+    Skip,
 }
 
 /// Client for interacting with an OCI registry for Spin applications.
@@ -119,6 +130,7 @@ impl Client {
         reference: impl AsRef<str>,
         annotations: Option<BTreeMap<String, String>>,
         infer_annotations: InferPredefinedAnnotations,
+        compose_mode: ComposeMode,
     ) -> Result<Option<String>> {
         let reference: Reference = reference
             .as_ref()
@@ -137,8 +149,15 @@ impl Client {
         )
         .await?;
 
-        self.push_locked_core(locked, auth, reference, annotations, infer_annotations)
-            .await
+        self.push_locked_core(
+            locked,
+            auth,
+            reference,
+            annotations,
+            infer_annotations,
+            compose_mode,
+        )
+        .await
     }
 
     /// Push a Spin application to an OCI registry and return the digest (or None
@@ -149,6 +168,7 @@ impl Client {
         reference: impl AsRef<str>,
         annotations: Option<BTreeMap<String, String>>,
         infer_annotations: InferPredefinedAnnotations,
+        compose_mode: ComposeMode,
     ) -> Result<Option<String>> {
         let reference: Reference = reference
             .as_ref()
@@ -156,8 +176,15 @@ impl Client {
             .with_context(|| format!("cannot parse reference {}", reference.as_ref()))?;
         let auth = Self::auth(&reference).await?;
 
-        self.push_locked_core(locked, auth, reference, annotations, infer_annotations)
-            .await
+        self.push_locked_core(
+            locked,
+            auth,
+            reference,
+            annotations,
+            infer_annotations,
+            compose_mode,
+        )
+        .await
     }
 
     /// Push a Spin application to an OCI registry and return the digest (or None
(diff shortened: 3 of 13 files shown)


src/commands/registry.rs
```diff
@@ -49,6 +49,10 @@ pub struct Push {
     )]
     pub insecure: bool,
 
+    /// Skip composing the application's components before pushing it.
+    #[clap(long = "skip-compose")]
```
fibonacci1729, line 53: Support toggling this with an environment variable

src/commands/registry.rs
```diff
@@ -49,6 +49,10 @@ pub struct Push {
     )]
     pub insecure: bool,
 
+    /// Skip composing the application's components before pushing it.
+    #[clap(long = "skip-compose", env = PUSH_ALWAYS_SKIP_COMPOSE)]
```
itowlson, line 53: I'm not sure if an environment variable fits well with how this would be used.  I don't think we have a clear philosophy on that though... happy to keep this if you think it's appropriate.

fibonacci1729: Yeah i'm not particularly bound to this; someone mentioned that they would like to disable it by tossing an environment variable in their shell.

crates/oci/src/client.rs
```diff
@@ -246,43 +264,57 @@ impl Client {
...
-            // Update the module source with the content ref of the layer.
-            c.source.content = self.content_ref_for_layer(&layer);
-
-            layers.push(layer);
-
-            let mut deps = BTreeMap::default();
-            for (dep_name, mut dep) in c.dependencies {
-                let source = dep
+            if !skip_compose {
+                let composed = spin_compose::compose(&ComponentSourceLoader, &c)
```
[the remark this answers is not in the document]
fibonacci1729: Yeah i considered that but ended up abandoning it for "less code"... Happy to revisit if this is too awkward.

crates/oci/src/client.rs
```diff
@@ -119,6 +120,7 @@ impl Client {
         reference: impl AsRef<str>,
         annotations: Option<BTreeMap<String, String>>,
         infer_annotations: InferPredefinedAnnotations,
+        skip_compose: bool,
```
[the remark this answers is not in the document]
fibonacci1729: Updated to use `ComposeMode::{All, Skip}`

src/commands/registry.rs
fibonacci1729: Updated and removed.

fibonacci1729:
cc/ @itowlson a simple test is now included.

fibonacci1729:
@itowlson friendly ping that this is ready for another review!

crates/oci/src/client.rs
```diff
@@ -1108,6 +1218,32 @@ mod test {
...
+                        "source": {
+                            "content_type": "application/wasm",
+                            "source": format!("file://{}", working_dir.path().join("dep_a.wasm").to_str().unwrap()),
+                            "digest": "digest",
+                        },
+                        "export": null,
+                    }
+                }
+                }]),
+                expected_layer_count: 1,
```
itowlson, line 1244: Does this validate the changes to the lockfile (e.g. component source is now the composed component, component now has no deps)?  I am not sure how practical that is.

crates/oci/src/client.rs
```diff
@@ -669,6 +712,32 @@ impl Client {
     }
 }
 
+struct ComponentSourceLoader;
+
+#[async_trait]
+impl spin_compose::ComponentSourceLoader for ComponentSourceLoader {
```
itowlson, line 718: This seems to duplicate the `trigger` crate ComponentSourceLoader.  Can we unify them?

crates/oci/src/client.rs
```diff
@@ -246,43 +273,59 @@ impl Client {
...
         let mut components = Vec::new();
         for mut c in locked.clone().components {
-            // Add the wasm module for the component as layers.
-            let source = c
-                .clone()
-                .source
-                .content
-                .source
-                .context("component loaded from disk should contain a file source")?;
+            match compose_mode {
```
itowlson, line 281: This stanza is very long and deeply nested, and by the time I got to the end of it I had kind of lost track of where I was or how it fitted into the flow.  (E.g. does line 328 end the for loop or only the match?  I need to scroll a whole screen up and squint at braces to tell.)  It would be really good to see if we could break the loop body out into two or three helper functions e.g. 

```
for ... {
  match compose_mode {
    All => composify(),
    Skip => dont_composify(),
  };
  assemble_layers_for_one();
}
```

kind of thing.  Or whatever you feel the appropriate cohesive units are.

crates/oci/Cargo.toml
```diff
@@ -22,10 +23,17 @@ reqwest = "0.12"
 serde = { workspace = true }
 serde_json = { workspace = true }
 spin-common = { path = "../common" }
+spin-componentize = { path = "../componentize" }
+spin-compose = { path = "../compose" }
```
itowlson, line 27: I apologise for harping on this but it breaks my heart that the OCI crate has to know about this stuff (but I understand that you've looked into this and it's the least worst way)

itowlson (commented):
First off, sorry for taking so long to get to this.

I don't really feel capable of reviewing this beyond "I affirm that Brian is not sneaking in a Bitcoin miner," but from what I can tell it seems sound, and I assume you've tested it against suitable hosts which is more reassuring than any reading of mine could be!

I did have some comments on the style and the testing - they're not blockers, but I'd like to see if they're easy to address before we merge.  (And please ignore the melodramatic last comment - we've already covered that but I'm struggling to reconcile myself to the outcome!)

crates/oci/Cargo.toml
fibonacci1729: I managed to eliminate the `spin-componentize` dependency by unifying the loader from trigger and here into a `ComponentSourceLoaderFs` that lives in `spin-compose`.

crates/oci/src/client.rs
fibonacci1729: Unified the loaders into a `spin_compose::ComponentSourceLoaderFs`

crates/oci/src/client.rs
fibonacci1729: It does not; i have manually verified that. If you think having that coverage here makes sense I can see about factoring out the "compose" tests to asserts against the final locked manifest layer.

itowlson: Happy to punt on that!

crates/oci/src/client.rs
```diff
@@ -246,16 +274,43 @@ impl Client {
         &mut self,
         locked: &mut LockedApp,
         assembly_mode: AssemblyMode,
+        compose_mode: ComposeMode,
     ) -> Result<Vec<ImageLayer>> {
-        let mut layers = Vec::new();
```
fibonacci1729, line 250: @itowlson Calling attention to this diff because it's minimized by default. Let me know if this is easier to follow.

itowlson (approved):
Thanks for those changes - this feels much more readable now (at least to me!).  Appreciate your patience with my niggling.  If you're still comfortable with it then good to go!  :shipit:

fibonacci1729:
Thanks for the review! I think i'll try to get 1 or 2 more 👀 just to spread awareness.

src/commands/registry.rs
```diff
@@ -49,6 +49,10 @@ pub struct Push {
     )]
     pub insecure: bool,
 
+    /// Skip composing the application's components before pushing it.
+    #[clap(long = "skip-compose")]
+    pub skip_compose: bool,
```
kate-goldenring, line 54: Clarifying that now this will default to true right so this is a breaking change to how users expect artifacts to be produced by default? Run runtimes should still be able to handle the composed artifacts so it shouldn't break runtimes that are tied to this CLI

crates/compose/src/lib.rs
```diff
@@ -42,6 +43,36 @@ pub trait ComponentSourceLoader {
     ) -> anyhow::Result<Vec<u8>>;
 }
 
+/// A ComponentSourceLoader that loads component sources from the filesystem.
+pub struct ComponentSourceLoaderFs;
+
+#[async_trait::async_trait]
+impl ComponentSourceLoader for ComponentSourceLoaderFs {
```
kate-goldenring, line 50: Why did this structure move to compose from loader?

kate-goldenring (approved):

src/commands/registry.rs
fibonacci1729: It defaults to false but yes the rest of your comment is correct! A follow up to this would be to set this to true for all the plugins that push to the registry.

crates/compose/src/lib.rs
fibonacci1729: To share the loader between the trigger crate and the oci; it's a fairly generic impl that isn't tied to anything trigger specific.

src/commands/registry.rs
kate-goldenring: if it defaults to false then there is no change. I thought the idea was for the default to unlock success for SpinKube (the shim)

fibonacci1729: Sorry! Yeah you're correct: the default is `skip-compose = false` which will always push compositions by default. My bad, I was responding literally to the value being false not the behavior 😅

kate-goldenring: So spinkube users will need to pass the `--skip-compose` flag? Will we switch the default for the next major release?

fibonacci1729: No SpinKube users won't need to do anything; the default of `spin registry push ...` will push composed components.

itowlson: Maybe we should call it `compose` or `precompose` and avoid the double negative, because I know what the defaults are and what they're for, but every time I try to untangle it I confuse myself...!  Semi kidding, but also can see a lot of people getting tangled in the grammar of this.

(Although I also find myself getting tangled in the grammar of the positive option, so, eh.  `--precompose-mode compatible` vs `--precompose-mode optimised` anyone...  _runs away_)

fibonacci1729: Switched to `--compose=true` as the default. PTAL

src/commands/registry.rs
```diff
@@ -49,6 +49,10 @@ pub struct Push {
     )]
     pub insecure: bool,
 
+    /// Specifies to compose components before pushing the application.
```
itowlson, line 52: Maybe help folks along with e.g.

```suggestion
    /// Compose component dependencies before pushing the application.
    ///
    /// The default is to compose before pushing, which maximises compatibility with
    /// different Spin runtime hosts. Turning composition off can optimise
    /// bandwidth for shared dependencies, but makes the pushed image incompatible
    /// with hosts that cannot carry out composition themselves.
```

Wordier I know but maybe helpful?

src/commands/registry.rs
kate-goldenring: I know this has been a back and forth but i think a `--skip-compose` flag makes more sense than `--compose=false` as an arg. I was just wrapping my head around the default but my head settled and i think your original intuition was best

fibonacci1729: Can you explain your reasoning for why it makes more sense? I tend to agree that the double negative was an annoying extra cognitive cycle.

kate-goldenring: Isn't the CLI experience the following:
```sh
# pushes OCI artifact with compositions
spin registry push
# push an artifact without composition
spin registry push --skip-compose
```

The non-negative, don't you have to set the value of the flag too:
```sh
# pushes OCI artifact with compositions
spin registry push
# push an artifact without composition
spin registry push --compose=false
```

fibonacci1729: Yes that's a fair point. I'm not too hung up on this because I don't anticipate regular developers to use this incantation as it will be covered by whatever deployment plugin they're using.

fibonacci1729: I'll switch it back to `--skip-compose` unless @itowlson is willing to die on this hill? 😆

itowlson: No, I am happy to stay alive in the valley arguing with myself about which hill to go up.

I do agree that the "without composition" case is something that interactive users should use only with the greatest of caution, so I have no problems with making it look graunchy.  But hopefully the negative form looks graunchy enough!

fibonacci1729: Oh now i'm leaning towards keeping it because the graunchiest dissuasion argument is quite compelling...

itowlson: Oh sorry, bad phrasing on my part, I wasn't trying to argue for the positive form in order to dissuade.  It seems like Kate has stronger feelings than either of us and I'm happy to go with what she recommends.

fibonacci1729: 😆 i think i'll keep it as is; thanks for the debate!

fibonacci1729:
Ran `make test` locally and everything seems green; Also fixed a few lints that popped up in `tests/`

Issue #930, opened 2022-12-04
JSON schema for `spin.toml`?


radu-matei:
I would really like to get some documentation and completion support when writing a `spin.toml` file, and apparently, the [Even better TOML](https://marketplace.visualstudio.com/items?itemName=tamasfe.even-better-toml#completion-and-validation-with-json-schema) VS Code extension lets you do that:

![](https://raw.githubusercontent.com/tamasfe/taplo/master/editors/vscode/images/schema.gif)

There is a configuration key a user would need to add to get schema support, but I think this would be a really good improvement to the user experience for writing `spin.toml` files.

itowlson:
related: https://taplo.tamasfe.dev/configuration/developing-schemas.html#visual-studio-code-extensions (they did this for us so we should get around to making use of it!)

itowlson:
and indeed https://github.com/fermyon/spin-vscode/issues/2

radu-matei:
Ooh, thanks for the schema, @cardoso!

@itowlson — is this something we can use for the VS Code extension?

cardoso:
@radu-matei yep, just some superficial tweaks to the manifest models to fix some gaps in the hints and it should be good.

You can use the link @itowlson posted above: https://taplo.tamasfe.dev/configuration/developing-schemas.html#publishing

I remember testing that a few months ago and it worked pretty well just by adding the contribution.

I'll see if I can push what I have locally. Would this be better as a build script or a command in the CLI?

radu-matei:
I think a build script would work well here.

Thoughts, @itowlson ?

lann:
It would be pretty straightforward to manually maintain a schema that covers the trigger-agnostic + common trigger (i.e. HTTP) types. It would also be possible to build a system into spin that would generate schema semi-automatically from code but Spin's trigger and factor systems make that significantly less straightforward than manual maintenance.

itowlson:
I recall generating and installing a schema for Even Better TOML / Taplo / whatever it's called, but being dissatisfied with the results when I installed it into VS Code (to the point of frustration).  Things have moved on since then, though, so I can take another look.

cardoso:
@itowlson I had the same experience back when I used schemars to generate the json schema (https://github.com/fermyon/spin/issues/930#issuecomment-1523765443). It was unusable due to a top-level `anyOf` in the json schema. If you removed that it worked pretty well.

Here's the diff:
[...]

PR #2997, opened 2025-01-28
Generate schema for v2 manifest

itowlson:
Attempts #930.

This is for discussion right now.  Some questions:

* Does this provide enough value to invest further time making it wholesome?  E.g. trigger properties are not autosuggested (but they are validated).  You can test the current output by installing Even Better TOML or Taplo and adding the `#:schema https://gist.githubusercontent.com/itowlson/080922335ea080a185b77a17b9f1eb03/raw/3c24973fbeec3f3c78a1adc34284c235eba6df23/manifest.v2.triggy.json` at the top of a spin.toml file.
* Which triggers should we bake in schemas for?  I've done HTTP and Redis, but should we do cron, command, MQTT, SQS, etc. too?  (Currently they are not validated but do not cause an error either.)
* How do we want to expose generation?  For now I have it as a `spin schema` command but there's no reason anyone except people updating the manifest format should ever need to run it.
* How do we want to publish/distribute the schema?  E.g. asset URL in a GH release, S3 bucket, VS Code extension, BitTorrent...  Even Better TOML is happy with HTTP URLs via the `#:schema` directive, which we could add to templates.

If we decide to press ahead, the main remaining work is:

* Agree design and review implementation
* Update the descriptions/doc comments (which are currently examples rather than explanations)
* Implement publishing / distribution.

Docs are probably a few hours, and I can do them; distribution would depend on how we want to do it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,6 +34,7 @@ jobs:
               extraArgs: "--features openssl/vendored",
               target: "",
               targetDir: "target/release",
+              generateManifestSchema: true,
             }
           - {
               os: "ubuntu-22.04",
@@ -132,6 +133,10 @@ jobs:
             --output-signature spin.sig \
             ${{ matrix.config.targetDir }}/spin${{ matrix.config.extension }}
 
+      - name: Generate manifest schema
+        if: matrix.config.generateManifestSchema
+        run: ${{ matrix.config.targetDir }}/spin${{ matrix.config.extension }} maintenance generate-manifest-schema -o manifest.schema.json
+
       - name: package release assets
         if: runner.os != 'Windows'
         shell: bash
@@ -167,6 +172,13 @@ jobs:
         with:
           name: spin-${{ env.RUNNER_OS }}-${{ matrix.config.arch }}
           path: _dist/spin-${{ env.RELEASE_VERSION }}-${{ env.RUNNER_OS }}-${{ matrix.config.arch }}.zip
+      
+      - name: Upload schema as GitHub artifact
+        if: matrix.config.generateManifestSchema
+        uses: actions/upload-artifact@v4
+        with:
+          name: manifest.schema.json
+          path: manifest.schema.json
 
   checksums:
     name: generate release checksums
@@ -206,10 +218,17 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: download release assets
+      - name: download binaries for release assets
         uses: actions/download-artifact@v4
         with:
           pattern: spin-*
+          path: _dist
+          merge-multiple: true
+
+      - name: download schema for release assets
+        uses: actions/download-artifact@v4
+        with:
+          pattern: manifest.schema.json
           path: _dist
           merge-multiple: true
 
--- a/crates/manifest/Cargo.toml
+++ b/crates/manifest/Cargo.toml
@@ -7,6 +7,7 @@ edition = { workspace = true }
 [dependencies]
 anyhow = { workspace = true }
 indexmap = { workspace = true, features = ["serde"] }
+schemars = { version = "0.8.21", features = ["indexmap2", "semver"] }
 semver = { workspace = true, features = ["serde"] }
 serde = { workspace = true }
 spin-serde = { path = "../serde" }
--- a/crates/manifest/src/schema/v2.rs
+++ b/crates/manifest/src/schema/v2.rs
@@ -1,18 +1,21 @@
 use anyhow::Context;
+use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use spin_serde::{DependencyName, DependencyPackageName, FixedVersion, LowerSnakeId};
 pub use spin_serde::{KebabId, SnakeId};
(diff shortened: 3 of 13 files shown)


karthik2804:
I like adding them to templates, can we add multiple directives and do they need to be at the top?

lann:
Given that schema is linked to Spin version I suggest adding it as a release artifact, with the latest then naturally available under the canary pre-release e.g. https://github.com/fermyon/spin/releases/download/canary/maniest-schema-or-whatever.json

itowlson:
@karthik2804 I don't know about multiple directives - I'm not sure what it would mean to specify multiple schemas for a file - can you elaborate?

@lann agree, GH release asset seems easiest if we are comfortable with that

karthik2804:
I was wondering if we could have multiple schemes injected by each of the template for the trigger such that the entire thing can be validated.

itowlson:
I don't believe so.  I think you'd need something dynamic to inject a field and schema reference for each trigger in the manifest.  Or you'd need to select out a range and apply a subschema to that.  (You can't attempt to validate the whole document against the subschemata.)  At that point, as far as I know, you're drifting beyond what JSON Schema can do and into a full-blown language server.

karthik2804:
Looking at their docs, there i no support for multiple definitions in the same doc https://taplo.tamasfe.dev/configuration/directives.html#the-schema-directive

itowlson:
Rebased and updated to be a subcommand of the `maintenance` command we introduced for generating the CLI reference.

itowlson:
Added release actions to include the schema JSON as part of GH releases.  Tested by running in my fork (albeit a subset) - https://github.com/itowlson/spin/actions/runs/14369671054, https://github.com/itowlson/spin/releases/tag/canary) and that seemed okay.

.github/workflows/release.yml
```diff
@@ -132,6 +132,10 @@ jobs:
             --output-signature spin.sig \
             ${{ matrix.config.targetDir }}/spin${{ matrix.config.extension }}
 
+      - name: Generate manifest schema
+        if: matrix.config.os == 'ubuntu-22.04' && matrix.config.arch == 'amd64'
```
lann, line 136: It might be nice to make a new matrix property, either specific like `generateManifestSchema: true` or generic `doOneTimeStuff`. I'm 90% sure if you omit a matrix property it will evaluate falsey in this context.

crates/manifest/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = { workspace = true }
 [dependencies]
 anyhow = { workspace = true }
 indexmap = { version = "2", features = ["serde"] }
+schemars = { version = "0.8.21", features = ["indexmap2", "semver"] }
```
lann, line 10: What do you call a bunch of pirates plotting to overthrow their captain?

crates/manifest/src/schema/v2.rs
```diff
@@ -82,12 +87,123 @@ pub struct Trigger {
...
+fn map_of_toml_tables_schema(
+    _gen: &mut schemars::gen::SchemaGenerator,
+) -> schemars::schema::Schema {
+    schemars::schema::Schema::Object(schemars::schema::SchemaObject {
+        instance_type: Some(schemars::schema::SingleOrVec::Single(Box::new(
+            schemars::schema::InstanceType::Object,
+        ))),
+        ..Default::default()
+    })
+}
```
lann, line 131: Perhaps these should move into the `mod` below? (with their `_schema` tails removed?)

crates/manifest/src/schema/v2.rs
```diff
@@ -82,12 +87,123 @@ pub struct Trigger {
...
+
+/// TODO: there may be a way to do this without dead code using `schemars::extend` or
+/// something like that?
+#[allow(dead_code)]
+mod schema {
```
lann, line 136: Especially if the above `fn`s move in here, maybe consider breaking this out into a separate `schema/json_schema.rs` file?

crates/manifest/src/schema/v2.rs
```diff
@@ -82,12 +87,123 @@ pub struct Trigger {
...
+        #[schemars(default)]
+        http: Vec<HttpTriggerSchema>,
+        /// Redis triggers
+        #[schemars(default)]
+        redis: Vec<RedisTriggerSchema>,
+    }
+
+    #[derive(JsonSchema)]
+    #[schemars(deny_unknown_fields)]
+    struct HttpTriggerSchema {
```
lann, line 152: This isn't the first time we've had to duplicate these for crate organization reasons; I wonder if it we should throw logic to the wind and move the canonical definitions of these "special" trigger configs into this crate... :grimacing:

lann (approved):
I'd be about 30% happier if the new schema code was separated a bit more but not blocking.

crates/manifest/src/schema/v2.rs
itowlson: My puritanical soul is scandalised to hear such a thing!  But yeah the duplication is a bit regrettable.  I propose to declare it out of scope for this PR, but we should come back to it.

PR #3002, opened 2025-01-30
Include guidance in error for non-matching HTTP component export

### LINKED ISSUE #3000: Docs: could improve the error message when newer exports do not match spin supported
* Spin version (`spin --version`): spin 3.0.0 (737778e 2024-11-07)
* Installed plugin versions (`spin plugins list --installed`):
squillace@idiopath:~/work/sample-wasi-http-rust$ spin plugins list --installed
cloud 0.5.1 [installed]
js2wasm 0.6.1 [installed]
k8s 0.1.0 [installed]
kube 0.3.1 [installed]
otel 0.1.2 [installed]
py2wasm 0.3.2 [installed]

When I build and run https://github.com/yoshuawuyts/sample-wasi-http-rust it runs fine with wasmtime 29.0.1, but the command `spin up -f <path to same component/>` results in the following error:

spin up -f target/wasm32-wasip1/debug/rust_wasi_hello.wasm

- Error: Expected component to export one of `wasi:http/incoming-handler@0.2.0-rc-2023-10-18`, `wasi:http/incoming-handler@0.2.0-c-2023-11-10`, `wasi:http/incoming-handler@0.2.0`, `wasi:http/incoming-handler@0.2.1`, or `fermyon:spin/inbound-http` but it exported none of those

While this is literally true -- the newer module from cargo component build is based on exporting inbound-httpv0.2.3 which is not in the supported list in that build of Spin -- it's a tad technical. I'm thinking we might want to have an additional suggestion, something like: `This often happens when the version of exports the module offers is newer than those supported in your version of Spin. Upgrading Spin might solve the problem.`

itowlson:
Fixes #3000 

image

### DIFF
--- a/crates/http/src/trigger.rs
+++ b/crates/http/src/trigger.rs
@@ -68,7 +68,9 @@ impl HandlerType {
                 `{WASI_HTTP_EXPORT_2023_10_18}`, \
                 `{WASI_HTTP_EXPORT_2023_11_10}`, \
                 `{WASI_HTTP_EXPORT_0_2_PREFIX}.*`, \
-                 or `{SPIN_HTTP_EXPORT}` but it exported none of those"
+                 or `{SPIN_HTTP_EXPORT}` but it exported none of those. \
+                 This may mean the component handles a different trigger, or that its `wasi:http` export is newer then those supported by Spin. \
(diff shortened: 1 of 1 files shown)


squillace:
BIKESHED!!!! WHAT????

PR #3003, opened 2025-01-31
Allow `self.alt` host for self-requests

itowlson:
This is a proposal to address https://github.com/fermyon/spin-js-sdk/issues/298.

The issue there is that JS guests use the inbuilt `fetch` function to make HTTP requests. If `fetch` sees a URL without a host, it helpfully prepends the host it thinks it's running on _before_ letting Spin see it.  So for a self-request such as `fetch("/back")`, Spin sees `http://localhost:3000/back` or whatever.  Which is not on the allow list so gets the banhammer.

This PR reluctantly compromises with `fetch` by allowing self-requests via the pseudo-host `self.alt`.  So if you do `fetch("http://self.alt/back")`, Spin will treat it as a self-request: it will validate permission using the self-request permission and route it as a self request.

The developer experience is, admittedly, not very lovely.  But it hopefully gives JS folks an escape hatch.

Notes for reviewers:

* Self-request permission is still given via `allowed_outbound_hosts = ["http://self"]`, _not_ `http://self.alt`.  Should we allow `self.alt` as another way of expressing the same permission?
* I implemented this only in wasi-http because that's what JS goes through.  Should I implement it in spin-http too?  It looked faffier there because of the way the types line up, but certainly do-able.

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -152,7 +152,12 @@ async fn send_request_impl(
     let host = request.uri().host().unwrap_or_default();
     let tls_client_config = component_tls_configs.get_client_config(host).clone();
 
-    if request.uri().authority().is_some() {
+    let is_self_request = request
+        .uri()
+        .authority()
+        .is_some_and(|a| a.host() == "self.alt");
+
+    if request.uri().authority().is_some() && !is_self_request {
         // Absolute URI
         let is_allowed = outbound_allowed_hosts
             .check_url(&request.uri().to_string(), "https")
--- a/crates/factor-outbound-networking/src/config.rs
+++ b/crates/factor-outbound-networking/src/config.rs
@@ -200,7 +200,7 @@ impl HostConfig {
             return Ok(Self::Any);
         }
 
-        if host == "self" {
+        if host == "self" || host == "self.alt" {
             return Ok(Self::ToSelf);
         }
 
--- a/examples/spin-timer/Cargo.lock
+++ b/examples/spin-timer/Cargo.lock
@@ -719,6 +719,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "backon"
+version = "1.3.0"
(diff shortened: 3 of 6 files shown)


karthik2804:
@itowlson this overall looks good, an alternate suggestion that came up during the original PR adding this feature also seems to indicate that this will benefit go. 
https://github.com/fermyon/spin/pull/1710#discussion_r1310704465

karthik2804 (approved):
I am okay with adding this to just the `wasi-http`.

radu-matei (approved):
> Self-request permission is still given via allowed_outbound_hosts = ["http://self"], not http://self.alt/. Should we allow self.alt as another way of expressing the same permission?

Not a blocker; but allowing `self.alt` since it's what you use here makes sense to me.

> I implemented this only in wasi-http because that's what JS goes through. Should I implement it in spin-http too? It looked faffier there because of the way the types line up, but certainly do-able.

This only being in WASI HTTP makes perfect sense to me.

LGTM-ing since this is a pretty big blocker for the JS SDK.

Thanks!

lann:
nit: Commit message says `spin.alt` :slightly_smiling_face:

lann:
> Self-request permission is still given via allowed_outbound_hosts = ["http://self"], not http://self.alt/. Should we allow self.alt as another way of expressing the same permission?

I would say yes, to reduce confusion.

itowlson:
Added bonus host.  Note that both `self` and `self.alt` will be accepted for either relative or `self.alt` requests.

Fixed commit message - thanks for the spot @lann!

lann (approved):
If we want to get this through quickly that's fine with me, but we really should have test coverage for this.

itowlson:
> If we want to get this through quickly that's fine with me, but we really should have test coverage for this.

Yep, agree.  Note we can't run tests at the moment because the runner got lost in the repo move, but worth having for when the lights come back on...

itowlson:
Added a test which Works On My Machine even if CI is packing a sad right now.

PR #3005, opened 2025-02-04
Update Wasmtime to 29.0.1

alexcrichton:
This commit updates Wasmtime from 25.0.3 to 29.0.1, the latest release of Wasmtime. This update is accompanied with few minor changes around bindings generation such as traits no longer needing `#[async_trait]`. The `wit-parser` crate used throughout Wasmtime additionally no longer supports `float32` in WIT files and instead requires `f32` to be used (same for` float64` and `f64`)

Otherwise one of the main features is that the WITs included with Wasmtime's WASI implementation have been updated to 0.2.3 as well. This should continue to still work with guests using 0.2.2 and prior, though.


alexcrichton:
One major thing here looks like the test suite's usage of wit-bindgen at 0.16.0 will need to be updated to a more modern wit-bindgen. That's a relatively major update so it'll take me some time to work through that, and I plan on doing that as a PR separate from this.

alexcrichton:
I've rebased this on https://github.com/fermyon/spin/pull/3006 and switched to draft until that merges.

crates/factor-outbound-http/src/wasi_2023_10_18.rs
```diff
@@ -8,9 +8,7 @@ mod bindings {
 
     wasmtime::component::bindgen!({
         path: "../../wit",
-        interfaces: r#"
-            include wasi:http/proxy@0.2.0-rc-2023-10-18;
-        "#,
+        world: "wasi:http/proxy@0.2.0-rc-2023-10-18",
```
lann, line 11: Do you recall why we needed the inline interface before? (and don't now)

crates/factor-wasi/src/wasi_2023_10_18.rs
```diff
@@ -14,12 +13,7 @@ mod bindings {
 
     wasmtime::component::bindgen!({
         path: "../../wit",
-        interfaces: r#"
-            // NB: this is handling the historical behavior where Spin supported
-            // more than "just" this snaphsot of the proxy world but additionally
-            // other CLI-related interfaces.
```
lann, line 20: Is this comment no longer relevant?

lann (approved):

crates/factor-outbound-http/src/wasi_2023_10_18.rs
alexcrichton: IIRC `world` only accepted "foo" before and not "foo:bar/baz". I also think that historically there wasn't a `wasi:http/proxy` world here but it was an amalgamation of other things. EIther that or I copy/pasted from elsewhere which had these motivations and this specific place didn't. Either way though I believe the two are equivalent and `interfaces: ...` was generating an error I wasn't quite sure why but `world` seemed to fix it.

crates/factor-wasi/src/wasi_2023_10_18.rs
alexcrichton: Yeah I think that this is an outdated comment already because the comment implies there's more than one `include` or something beyond proxy but there's just the proxy world there now.

dicej (approved):

PR #3014, opened 2025-02-13
ignore gh workflow for build on .md & docs/

michelleN:


itowlson (changes requested):
I removed these intentionally, because they meant required checks didn't run, so that admin intervention was required on every docs-only PR.  See https://github.com/spinframework/spin/pull/2879.  I would really prefer not to see them back unless we have a workaround for that!

michelleN:
ah that makes sense @itowlson. Thanks will close.

PR #3016, opened 2025-02-19
Check allowed hosts before checking the `request_interceptor`

alexcrichton:
This commit updates the `request_interceptor` implementation of outbound HTTP requests to check the set of allowed hosts before calling the `request_interceptor`. This fixes behavior in Spin where service-chained requests were allowed regardless of configuration of `allowed_outbound_hosts` in `spin.toml`. A new test is added to ensure that a request to the "back" component is disallowed.

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -132,23 +132,6 @@ async fn send_request_impl(
 
     spin_telemetry::inject_trace_context(&mut request);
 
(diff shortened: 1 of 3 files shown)


lann:
Oof - I'm sure we had this implemented before. Must have been lost in the conversion to factors.

lann (approved):

itowlson (approved):
Yikes.  Thanks for adding the test!

PR #3017, opened 2025-02-21
Use connection pooling when querying CosmosDB

kate-goldenring:
I noticed that KV reads/writes had a really high latency with the CosmosDB provider. I crated some [simple test cases](https://github.com/kate-goldenring/cosmos-db-sdks-tests/tree/main) for interacting with CosmosDB and found that while the go SDK reuses connections between requests, the Rust SDK always created a new client. @lann discovered that the Azure Rust SDK disabled connection pooling because of this upstream issue: https://github.com/hyperium/hyper/issues/2312. The maintainers closed the issue deeming it to be irreproducible. I am getting latencies 3x higher without connection pooling (from 15ms reads to 60ms reads). I think it is worth enabling the pooling by creating our own reqwest client and passing it. 

I put it behind a feature flag in case we do hit the deadlock scenario that some have experienced.

~Marking this as draft for discussion~ Seems like this is a change we want

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8077,6 +8077,7 @@ dependencies = [
  "azure_data_cosmos",
  "azure_identity",
  "futures",
+ "reqwest 0.12.9",
  "serde",
  "spin-core",
  "spin-factor-key-value",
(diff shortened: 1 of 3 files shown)


rylev:
For reference [this is the PR](https://github.com/Azure/azure-sdk-for-rust/pull/1550) that disabled connection pooling in the Azure SDK.

I'd personally prefer to see this land. The hanging issue in hyper seems to be quite mysterious, and there's lots of people who _do not_ see this issue. Given that this is behind a feature flag, we can easily introduce this without changing current behavior giving folks the ability to test this out and report issues if they find them.

lann (approved):

rylev (approved):

PR #3020, opened 2025-02-21
fix: support empty files directories

kate-goldenring:
fixes https://github.com/spinframework/spin/issues/3019

Note, when directly mounting the fs with `--direct-mounts` this error is not encountered. Related https://github.com/spinframework/spin/issues/3018#issuecomment-2675096223


itowlson (approved):

itowlson:
@kate-goldenring Most of the CI failures are due to the large runners issue but I think the lint one might be real?  Otherwise happy to bypass the checks.

PR #3025, opened 2025-02-25
Update wagi.rs

guyoung:
modify parse WagiHttp request params to  command-line arguments

### DIFF
--- a/crates/trigger-http/src/wagi.rs
+++ b/crates/trigger-http/src/wagi.rs
@@ -38,13 +38,19 @@ impl HttpExecutor for WagiHttpExecutor {
         // Build the argv array by starting with the config for `argv` and substituting in
         // script name and args where appropriate.
         let script_name = uri_path.to_string();
-        let args = req.uri().query().unwrap_or_default().replace('&', " ");
-        let argv = self
-            .wagi_config
-            .argv
-            .clone()
-            .replace("${SCRIPT_NAME}", &script_name)
-            .replace("${ARGS}", &args);
+        
+        let mut args: Vec<String> = Vec::new();
+        args.push(script_name);
+
+        let parsed_url = url::Url::parse(req.uri().to_string().as_str())?;
+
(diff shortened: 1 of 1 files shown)


itowlson (changes requested):
This change seems to alter the behaviour to ignore `wagi_config`.  Can you say more about what you intend with this change?

guyoung:
In my project AIMatrices(https://github.com/guyoung/AIMatrices), a large amount of code from the Spin project is referenced. There is a requirement to run Python code using wagi, and the http requests are as follows:

- http://127.0.0.1:21500/python?-h
- http://127.0.0.1:21500/python?-c=print(1)
- http://127.0.0.1:21500/python?-c&print(1)
- http://127.0.0.1:21500/python?-c&import sys; print(sys.version)

Currently, the parameter parsing method of WagiHttpExecutor cannot meet the above http requests.

itowlson:
Thanks for the clarification.  The WAGI executor an `argv` template from configuration, and we can't just throw that away.  For example, a Spin app could set:

```
executor = { type = "wagi", argv = "my-first-value ${SCRIPT_NAME} my-middle-value ${ARGS} my-last-value"
```

and your proposed change would break that.

I'm guessing the big problem is the way that the implementation splits the post-substitution `argv` on spaces, so your query string of `http://127.0.0.1:21500/python?-c&import sys; print(sys.version)` gets turned into an `args` of `["-c", "import", "sys;", "print(sys.version)"]` or something like that?

If so, maybe the fix is to change how we use the template.  Currently we sub into the template string, then split the post-substitution `argv` on spaces to get an `args` iterator.  Maybe the better strategy is to split the template string on spaces _first,_ then splice in the array of query-string pairs at the position of the `${ARGS}` element.  That seems like it might meet your needs while retaining backward compatibility with `argv` templates.  Thoughts?

guyoung:
I agree with you

PR #3026, opened 2025-02-25
Update Cargo.toml

guyoung:
add url dependencie


itowlson (changes requested):
This crate doesn't depend on `url`.  If this is in support of your other PR, could you make it part of that PR please?

guyoung:
Pull requests #3025 dependencies url

itowlson:
Thanks for clarifying.  Could you make this part of #3025 then, please, rather than a completely separate PR?  Otherwise it will go in as an unused dependency.  Thanks!

itowlson:
Closing - if #3025 is still active then please fold into that if possible (otherwise, flag up here and we can reopen).

PR #3032, opened 2025-02-28
Fixes and enhancements to Cosmos KV host implementation

kate-goldenring:
This contains some enhancements and a couple important bug fixes

Bug fixes:
- deletes (on instances without `store_id` configured) were not working due to using the incorrect store ID
- increment was failing due to the value never being initially set. Now this sets the value if DNE -- this is the same as Spin's other KV hosts and it is impossible to set a number counter through the `set` api which requires a string value
- gets were failing on pairs that were set with increment (because the value type was a number instead of string). Deserialized into an agnostic `Keys` structure to help with this.

Enhancements:
- ~Support for writing and querying data in non-primary regions by enabling `allow_tentative_writes`. This is needed for databases configured with global replicas if not using the primary replica~ Removing this until we enable configuring region in runtime config.

### DIFF
--- a/crates/key-value-azure/Cargo.toml
+++ b/crates/key-value-azure/Cargo.toml
@@ -10,9 +10,9 @@ rust-version.workspace = true
 
 [dependencies]
 anyhow = { workspace = true }
-azure_data_cosmos = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
-azure_identity = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
-azure_core = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
+azure_data_cosmos = "0.21.0"
+azure_identity = "0.21.0"
+azure_core = "0.21.0"
 futures = { workspace = true }
 serde = { workspace = true }
 async-trait = { workspace = true }
--- a/crates/key-value-azure/src/store.rs
+++ b/crates/key-value-azure/src/store.rs
@@ -138,14 +138,18 @@ struct AzureCosmosStore {
     client: CollectionClient,
     /// An optional store id to use as a partition key for all operations.
     ///
-    /// If the store id not set, the store will use `/id` as the partition key.
+    /// If the store ID is not set, the store will use `/id` (the row key) as
+    /// the partition key. For example, if `store.set("my_key", "my_value")` is
+    /// called, the partition key will be `my_key` if the store ID is set to
+    /// `None`. If the store ID is set to `Some("myappid/default"), the
+    /// partition key will be `myappid/default`.
     store_id: Option<String>,
 }
 
 #[async_trait]
 impl Store for AzureCosmosStore {
     async fn get(&self, key: &str) -> Result<Option<Vec<u8>>, Error> {
-        let pair = self.get_pair(key).await?;
+        let pair = self.get_entity::<Pair>(key).await?;
         Ok(pair.map(|p| p.value))
     }
 
@@ -164,18 +168,20 @@ impl Store for AzureCosmosStore {
     }
 
     async fn delete(&self, key: &str) -> Result<(), Error> {
-        if self.exists(key).await? {
-            let document_client = self
-                .client
-                .document_client(key, &self.store_id)
-                .map_err(log_error)?;
-            document_client.delete_document().await.map_err(log_error)?;
+        let document_client = self
+            .client
+            .document_client(key, &self.store_id.clone().unwrap_or(key.to_string()))
+            .map_err(log_error)?;
+        if let Err(e) = document_client.delete_document().await {
+            if e.as_http_error().map(|e| e.status() != 404).unwrap_or(true) {
+                return Err(log_error(e));
+            }
         }
         Ok(())
     }
 
     async fn exists(&self, key: &str) -> Result<bool, Error> {
-        Ok(self.get_pair(key).await?.is_some())
+        Ok(self.get_entity::<Key>(key).await?.is_some())
     }
 
     async fn get_keys(&self) -> Result<Vec<String>, Error> {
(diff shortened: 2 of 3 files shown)


crates/key-value-azure/src/store.rs
```diff
@@ -353,15 +412,18 @@ impl Cas for CompareAndSwap {
 }
 
 impl AzureCosmosStore {
-    async fn get_pair(&self, key: &str) -> Result<Option<Pair>, Error> {
+    async fn get_pair<F>(&self, key: &str) -> Result<Option<F>, Error>
```
itowlson, line 415: Should we change the name of this function?  It seems like it's no longer limited to pairs.

crates/key-value-azure/src/store.rs
```diff
@@ -158,24 +159,35 @@ impl Store for AzureCosmosStore {
...
     async fn delete(&self, key: &str) -> Result<(), Error> {
-        if self.exists(key).await? {
-            let document_client = self
-                .client
-                .document_client(key, &self.store_id)
-                .map_err(log_error)?;
-            document_client.delete_document().await.map_err(log_error)?;
+        let document_client = self
+            .client
+            .document_client(key, &self.store_id.clone().unwrap_or(key.to_string()))
```
itowlson, line 171: Sanity check... `store_id` seems to be a partition key (or PK column???) but `key` seems to be a row key - is it correct to mix them?  The docs imply the default store id is `/id` rather than the current row key, but I may be misunderstanding how the store ID gets applied.

_EDIT: oh is this aligning with `Pair`'s implementation of `CosmosEntity::partition_key`?  that feels like we should make the relationship clearer then.  Maybe improve docs on `self.store_id` too?_

crates/key-value-azure/src/store.rs
```diff
@@ -216,24 +228,67 @@ impl Store for AzureCosmosStore {
...
+    // TODO: The function should parse the new value from the return response
+    // rather than sending an additional new request. However, the current SDK
+    // version does not support this.
     async fn increment(&self, key: String, delta: i64) -> Result<i64, Error> {
         let operations = vec![Operation::incr("/value", delta).map_err(log_error)?];
-        let _ = self
+        match self
             .client
-            .document_client(key.clone(), &self.store_id)
+            .document_client(&key, &self.store_id.clone().unwrap_or(key.to_string()))
```
itowlson, line 245: same question about whether `key` is the right fallback here _EDIT: and same edit!_

crates/key-value-azure/src/store.rs
```diff
@@ -216,24 +228,67 @@ impl Store for AzureCosmosStore {
...
+                        store_id: self.store_id.clone(),
+                    };
+                    if let Err(e) = self
+                        .client
+                        .create_document(counter)
+                        .is_upsert(false)
+                        .allow_tentative_writes(TentativeWritesAllowance::Allow)
+                        .await
+                    {
+                        if e.as_http_error()
```
itowlson, line 268: I found the nesting of similar-looking code hard to follow here. It would be great to pull out some chunks and/or helpers here but I appreciate it may not be easy given retries etc!

crates/key-value-azure/src/store.rs
```diff
@@ -274,8 +329,9 @@ impl CompareAndSwap {
 
 #[async_trait]
 impl Cas for CompareAndSwap {
-    /// `current` will fetch the current value for the key and store the etag for the record. The
-    /// etag will be used to perform and optimistic concurrency update using the `if-match` header.
+    /// `current` will fetch the current value for the key and store the etag
```
itowlson, line 332: Not sure why this and other doc comments needed reformatting, or is it a new lint?

itowlson (commented):
I think my comments mostly reflect my ignorance rather than actual issues.  Hopefully @devigned can provide a more informed review!

crates/key-value-azure/Cargo.toml
```diff
@@ -10,9 +10,9 @@ rust-version.workspace = true
 
 [dependencies]
 anyhow = { workspace = true }
-azure_data_cosmos = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
-azure_identity = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
-azure_core = { git = "https://github.com/azure/azure-sdk-for-rust.git", rev = "8c4caa251c3903d5eae848b41bb1d02a4d65231c" }
+azure_data_cosmos = { git = "https://github.com/azure/azure-sdk-for-rust.git", tag = "azure_data_cosmos-0.21.0"  }
```
rylev, line 13: Can we use the crates.io release instead?

crates/key-value-azure/src/store.rs
```diff
@@ -158,24 +159,35 @@ impl Store for AzureCosmosStore {
         self.client
             .create_document(pair)
             .is_upsert(true)
+            .allow_tentative_writes(TentativeWritesAllowance::Allow)
```
rylev, line 162: We should add a comment why this is needed.

crates/key-value-azure/src/store.rs
```diff
@@ -216,24 +228,67 @@ impl Store for AzureCosmosStore {
         Ok(())
     }
 
+    // WARNING: this function only works on the primary region because the
```
rylev, line 231: Nit: I think it's preferable to make this a doc comment (rather than a plain comment) so it's more easily visible outside of the source code.

crates/key-value-azure/src/store.rs
```diff
@@ -216,24 +228,67 @@ impl Store for AzureCosmosStore {
...
-                p.value.try_into().expect("incorrect length"),
-            )),
-            None => Err(Error::Other(
-                "increment returned an empty value after patching, which indicates a bug"
-                    .to_string(),
-            )),
+        {
+            Err(e) => {
+                if e.as_http_error()
+                    .map(|e| e.status())
```
rylev, line 252: It might help with @itowlson's comment directly below this to do something like:

```suggestion
            Err(e)  if e.as_http_error().map(|e| e.status() == 404).umwrap_or(false) => {
               
```

crates/key-value-azure/src/store.rs
rylev: Maybe `get_entity`?

rylev (commented):
Looking good! I think it's by and large ready to merge, but it would be nice to see the nits addressed.

crates/key-value-azure/Cargo.toml
devigned: Looks like 0.22.0 is available on crates.

crates/key-value-azure/src/store.rs
```diff
@@ -327,15 +383,18 @@ impl Cas for CompareAndSwap {
                 // attempt to replace the document if the etag matches
                 doc_client
                     .replace_document(pair)
+                    .allow_tentative_writes(TentativeWritesAllowance::Allow)
```
devigned, line 386: I'm a little concerned that by allowing tentative writes, this becomes non-atomic via last in wins default conflict resolution. It seems like this operation should be targeted at the hub rather than one of the spoke db instances. Thoughts?

devigned (commented):
I really like the improvements. I did mention concern about tentative writes in atomic operations. In the scenario where a swap occurs at roughly the same time on the same key, you could imagine 2 processes running and fetching the same etag version of the document, then the first committing the updated document to the hub while the second is tentatively committing the document to a spoke instance after the first. The second document committed seems like it would win with the [default conflict resolution protocol of last in wins](https://learn.microsoft.com/en-us/azure/cosmos-db/conflict-resolution-policies). 

I'm not sure if this is the right answer, but we may want to consider sending atomic reads and writes to the hub db instance and avoid multi-region writes.

crates/key-value-azure/Cargo.toml
kate-goldenring: I tried to update these changes to be on top of 0.22.0 when it came out a couple days ago but they removed some functionality for performing swaps using the `if-match` header. I filed an issue here: https://github.com/Azure/azure-sdk-for-rust/issues/2254

I can switch back to pointing to a commit sha if that is more robust

crates/key-value-azure/src/store.rs
kate-goldenring: This is a good point. One option would be to remove `allow_tentative_writes` from all atomic operations. Right now, through runtime config, there is no way to configure a target endpoint/region so the hub endpoint is always generated anyways. This brings up the point that we should probably hold off on `allow_tentative_writes` until we update runtime config to support configuring non-primary regions.

crates/key-value-azure/src/store.rs
kate-goldenring: I updated the `store_id` inline doc to explain that `/id` is the row key:
```
    /// An optional store id to use as a partition key for all operations.
    ///
    /// If the store id not set, the store will use `/id` (the row key) as the
    /// partition key. For example, if `store.set("my_key", "my_value")` is
    /// called, the partition key will be `my_key` if the store id is set to
    /// `None`. If the store id is set to `Some("myappid/default"), the
    /// partition key will be `myappid/default`.
    store_id: Option<String>,
}
```

crates/key-value-azure/src/store.rs
kate-goldenring: I usually lint at 80 chars so must have updated this by accident. looks like we use 100.

crates/key-value-azure/src/store.rs
kate-goldenring: updated to `get_entity`

kate-goldenring:
Updated the PR to only contain the set, delete, and get fixes. Removed non-primary region support since there is no way to configure it in runtime config

crates/key-value-azure/src/store.rs
```diff
@@ -138,14 +138,18 @@ struct AzureCosmosStore {
     client: CollectionClient,
     /// An optional store id to use as a partition key for all operations.
     ///
-    /// If the store id not set, the store will use `/id` as the partition key.
+    /// If the store ID is not set, the store will use `/id` (the row key) as
```
itowlson, line 141: Thank you Kate, this is so much clearer!

itowlson (approved):

crates/key-value-azure/Cargo.toml
lann: Is there some specific reason we need to use the repo here?

```suggestion
azure_data_cosmos = "0.21"
```

kate-goldenring: Sorry for the confusion here. I am not sure why i thought it wasn't published to crates. Updated

itowlson:
@kate-goldenring Do you have a moment to resolve the merge conflict on this?  Sorry I think there has been a bit of churn while GH actions was malingering

kate-goldenring:
Thanks for the bump @itowlson

itowlson:
Bypassing checks as only flakes and bads failing now

PR #3041, opened 2025-03-10
If a component errors after starting the response, show that error

### LINKED ISSUE #3035: HTTP streaming guest panic is not surfaced
If you have a HTTP guest which is streaming a response (using the WASI IncomingRequest/ResponseOutparam signature directly), and it panics _after_ creating and setting the OutgoingResponse but _before_ it has finished sending the response, the panic is not surfaced to the Spin console.  You need to turn on the 'warning' level of logging to see that there was a failure:

```
2025-03-04T02:32:58.051634Z  WARN spin_trigger_http::wasi: component error after response: guest invocation failed

Caused by:
    0: error while executing at wasm backtrace:
           # ... backtrace elided ...
           7: 0xacc2ca - html_rewriter_host.wasm!spin_executor::run::h0d464f805ce2c9e7
           8: 0xadd1a4 - html_rewriter_host.wasm!wasi:http/incoming-handler@0.2.0#handle
    1: wasm trap: wasm `unreachable` instruction executed
```

(By contrast, a "normal" request-response function produces a nice big `ERROR spin_trigger_http::server: Error processing request: guest invocation failed`.)

In my case something was blowing up at a point where I saw no output at all, yet Spin on its default log level didn't tell me anything was failing!

I'd suggest that we should either:

* Report a component error as an error regardless of when it falls in the response cycle; or
* Detect if the OutgoingBody has been `finish`-ed: warn if finished, error if not.

The first is obviously simpler, but I assume we had a reason for preferring warning...

itowlson:
Fixes #3035.

This makes little or no odds in non-streaming scenarios, where no guest code executes after the guest returns a response object (although SDK code may execute), and so any errors are likely to occur before the response is constructed let alone commenced.

However, in streaming scenarios, a guest may error (e.g. Rust panic) after the ResponseOutparam is set and its body taken, but before the response body is finished (or even started).  The default Spin logging settings turn this into a silent failure, which makes it hard to know where to start.  (In some cases a guest may be doing interesting things even after the response has completed, e.g. a Slack command handler doing work after the ack timeout, with intent to call back.)

This PR therefore bumps post-response guest errors to trace at Error level instead of Warning.

### DIFF
--- a/crates/trigger-http/src/wasi.rs
+++ b/crates/trigger-http/src/wasi.rs
@@ -1,3 +1,4 @@
+use std::io::IsTerminal;
 use std::net::SocketAddr;
 
 use anyhow::{anyhow, Context, Result};
@@ -144,7 +145,11 @@ impl HttpExecutor for WasiHttpExecutor {
                         Ok(())
                     }
                     .map_err(|e: anyhow::Error| {
-                        tracing::warn!("component error after response: {e:?}");
+                        if std::io::stderr().is_terminal() {
+                            tracing::error!("Component error after response started. The response may not be fully sent: {e:?}");
+                        } else {
+                            terminal::warn!("Component error after response started: {e:?}");
+                        }
                     }),
                 );
 


rylev (commented):
I'm unsure how I feel about this. Is guest code erroring an error from the perspective of the runtime? If we come from the perspective that the runtime is running untrusted code, an error in the guest is not really an error from the perspective of the runtime time.

That being said I understand the very practical need of showing errors when a guest errors. 

Warning feels semantically like the right option here. Should we instead be showing warnings by default?

lann:
I think error is appropriate - the guest code could be failing to do something semantically important like marking a message as "delivered" once the response is complete.

rylev:
I think I agree. We're always making the fundamental assumption in the `trigger-http` crate that the administrator of Spin cares about what the guest is actually doing. This isn't always the case in a spin compliant runtime, but I think it is the case here.

itowlson:
@rylev It's a good point.  The PR inadvertently takes a developer-centric point of view: I am building an app, it is not responding, but I am getting no feedback as to why.  I guess maybe a better solution is to tell me to improve the way I structure their streaming code and my logging habits.  But we report the error if a guest panics _before_ starting the response (or at least that's my impression), so it seemed odd to have a different experience for an app that goes down before the body is complete.

I did consider "downgrade to a warning once the response body is complete" but 1. I couldn't find a way to detect that at the point the warning/error is reported and 2. if the guest is still doing work after completing the body then maybe the guest cares.

But you're right about the issue of guest errors cluttering up admin logs in other hosts, and the position of this code makes it high-effort for hosts such as SpinKube to replace.  So maybe it's best to back this out, review the error-before-response behaviour, and document better guidance.

itowlson:
Maybe the behaviour should be "log it as a warning, but print an error to stderr if stderr is a terminal."

itowlson:
I've amended this to retain the current behaviour in "operator-ish" environments (so as not to spam logs with guest errors) but to report as an error when interactive (so that people like me can figure out why their app isn't responding).  Thoughts?

fibonacci1729 (approved):

PR #3043, opened 2025-03-10
Pool Postgres connections

### LINKED ISSUE #3030: Reuse connections between requests with Postgres
The Postgres host component creates a client during calls to [`open_connection`](https://github.com/spinframework/spin/blob/959075d2c592d397bc079e6defe476ee61e8a47e/crates/factor-outbound-pg/src/host.rs#L16-L17) this client can then be reused across all database interactions within that request.

Ideally, connections should be shared between requests, with clients cached.

itowlson:
Fixes #3030.

~Draft for now.  Needs more testing and some tidying but I want to see if tests pass, and to verify with factors experts if I have the pools in the right place (AppState).~

### DIFF
--- a/crates/factor-outbound-pg/src/client.rs
+++ b/crates/factor-outbound-pg/src/client.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use native_tls::TlsConnector;
 use postgres_native_tls::MakeTlsConnector;
 use spin_world::async_trait;
@@ -6,15 +6,80 @@ use spin_world::spin::postgres::postgres::{
     self as v3, Column, DbDataType, DbValue, ParameterValue, RowSet,
 };
 use tokio_postgres::types::Type;
-use tokio_postgres::{config::SslMode, types::ToSql, Row};
-use tokio_postgres::{Client as TokioClient, NoTls, Socket};
+use tokio_postgres::{config::SslMode, types::ToSql, NoTls, Row};
+
+const CONNECTION_POOL_SIZE: usize = 64;
+
+#[async_trait]
+pub trait ClientFactory: Send + Sync {
+    type Client: Client + Send + Sync + 'static;
+    fn new() -> Self;
+    async fn build_client(&mut self, address: &str) -> Result<Self::Client>;
+}
+
+pub struct PooledTokioClientFactory {
+    pools: std::collections::HashMap<String, deadpool_postgres::Pool>,
+}
+
+#[async_trait]
+impl ClientFactory for PooledTokioClientFactory {
+    type Client = deadpool_postgres::Object;
+
+    fn new() -> Self {
+        Self {
+            pools: Default::default(),
+        }
+    }
+
+    async fn build_client(&mut self, address: &str) -> Result<Self::Client> {
+        let pool_entry = self.pools.entry(address.to_owned());
+        let pool = match pool_entry {
+            std::collections::hash_map::Entry::Occupied(entry) => entry.into_mut(),
+            std::collections::hash_map::Entry::Vacant(entry) => {
+                let pool = create_connection_pool(address)
+                    .context("establishing PostgreSQL connection pool")?;
+                entry.insert(pool)
+            }
+        };
+
+        Ok(pool.get().await?)
+    }
+}
+
+fn create_connection_pool(address: &str) -> Result<deadpool_postgres::Pool> {
+    let config = address
+        .parse::<tokio_postgres::Config>()
+        .context("parsing Postgres connection string")?;
+
+    tracing::debug!("Build new connection: {}", address);
+
+    // TODO: This is slower but safer. Is it the right tradeoff?
+    // https://docs.rs/deadpool-postgres/latest/deadpool_postgres/enum.RecyclingMethod.html
+    let mgr_config = deadpool_postgres::ManagerConfig {
+        recycling_method: deadpool_postgres::RecyclingMethod::Clean,
+    };
+
+    let mgr = if config.get_ssl_mode() == SslMode::Disable {
+        deadpool_postgres::Manager::from_config(config, NoTls, mgr_config)
+    } else {
+        let builder = TlsConnector::builder();
+        let connector = MakeTlsConnector::new(builder.build()?);
+        deadpool_postgres::Manager::from_config(config, connector, mgr_config)
+    };
+
+    // TODO: what is our max size heuristic?  Should this be passed in soe that different
+    // hosts can manage it according to their needs?  Will a plain number suffice for
+    // sophisticated hosts anyway?
+    let pool = deadpool_postgres::Pool::builder(mgr)
+        .max_size(CONNECTION_POOL_SIZE)
+        .build()
+        .context("building Postgres connection pool")?;
+
+    Ok(pool)
+}
 
 #[async_trait]
 pub trait Client {
-    async fn build_client(address: &str) -> Result<Self>
-    where
-        Self: Sized;
-
     async fn execute(
         &self,
         statement: String,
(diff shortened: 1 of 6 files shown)


itowlson:
See https://github.com/spinframework/spin/issues/3030 for local performance characteristics.

itowlson:
There are now conformance tests for Postgres which will make it easier to regression-test this, but the conformance test PR is probably blocked on the data types PR, which is probably blocked on the conformance test PR.

It would be good to get some reviews on this little knot of PRs and figure out how to land them.

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,85 @@
...
+use tokio_postgres::{config::SslMode, types::ToSql, NoTls, Row};
+
+const CONNECTION_POOL_SIZE: usize = 64;
 
 #[async_trait]
-pub trait Client {
-    async fn build_client(address: &str) -> Result<Self>
-    where
-        Self: Sized;
+pub trait ClientFactory: Send + Sync {
```
lann: :factory:

crates/factor-outbound-pg/src/host.rs
```diff
@@ -71,8 +74,8 @@ fn v2_params_to_v3(
     params.into_iter().map(|p| p.try_into()).collect()
 }
 
-impl<C: Send + Sync + Client> spin_world::spin::postgres::postgres::HostConnection
-    for InstanceState<C>
+impl<CF: ClientFactory + Send + Sync> spin_world::spin::postgres::postgres::HostConnection
```
lann, line 77:
```suggestion
impl<CF: ClientFactory> spin_world::spin::postgres::postgres::HostConnection
```

crates/factor-outbound-pg/src/client.rs
lann: I think it should be safe to add `+ 'static` here, which would let us remove all the related `Send + Sync + 'static` constraints elsewhere.

crates/factor-outbound-pg/src/lib.rs
```diff
@@ -1,24 +1,26 @@
...
+pub struct OutboundPgFactor<CF = crate::client::PooledTokioClientFactory> {
+    _phantom: std::marker::PhantomData<CF>,
 }
 
-impl<C: Send + Sync + Client + 'static> Factor for OutboundPgFactor<C> {
+impl<CF: ClientFactory + Send + Sync + 'static> Factor for OutboundPgFactor<CF> {
     type RuntimeConfig = ();
-    type AppState = ();
-    type InstanceBuilder = InstanceState<C>;
+    type AppState = Arc<RwLock<CF>>;
```
lann: I'd recommend leaving interior mutability up to the trait implementation if possible; some implementations might not need to lock.

```suggestion
    type AppState = Arc<CF>;
```

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,85 @@
...
+    type Client: Client + Send + Sync + 'static;
+    fn new() -> Self;
```
lann: This would let you `#[derive(Default)]` for `PooledTokioClientFactory`

```suggestion
pub trait ClientFactory: Send + Sync + Default{
    type Client: Client + Send + Sync + 'static;
```

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,85 @@
...
+
+const CONNECTION_POOL_SIZE: usize = 64;
 
 #[async_trait]
-pub trait Client {
-    async fn build_client(address: &str) -> Result<Self>
-    where
-        Self: Sized;
+pub trait ClientFactory: Send + Sync {
+    type Client: Client + Send + Sync + 'static;
```
lann: I guess we could probably move these constraints onto `Client`, too.

crates/factor-outbound-pg/src/client.rs
rylev: I know it didn't have doc comments before, but it would be awesome if it gained some.

crates/factor-outbound-pg/src/lib.rs
rylev: This would require changing the trait to use immutable borrows which is fine, but it will complicate the implementation of the ClientFactory.

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,85 @@
...
+}
+
+fn create_connection_pool(address: &str) -> Result<deadpool_postgres::Pool> {
+    let config = address
+        .parse::<tokio_postgres::Config>()
+        .context("parsing Postgres connection string")?;
+
+    tracing::debug!("Build new connection: {}", address);
 
+    // TODO: This is slower but safer. Is it the right tradeoff?
```
lann: I think this is a good default. We can always expose other options via runtime config later if we want.

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,85 @@
...
+
+    let mgr = if config.get_ssl_mode() == SslMode::Disable {
+        deadpool_postgres::Manager::from_config(config, NoTls, mgr_config)
+    } else {
+        let builder = TlsConnector::builder();
+        let connector = MakeTlsConnector::new(builder.build()?);
+        deadpool_postgres::Manager::from_config(config, connector, mgr_config)
+    };
+
+    // TODO: what is our max size heuristic?  Should this be passed in soe that different
```
lann: nit:
```suggestion
    // TODO: what is our max size heuristic?  Should this be passed in so that different
```

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,85 @@
...
 
 #[async_trait]
-pub trait Client {
-    async fn build_client(address: &str) -> Result<Self>
-    where
-        Self: Sized;
+pub trait ClientFactory: Send + Sync {
+    type Client: Client + Send + Sync + 'static;
+    fn new() -> Self;
+    async fn build_client(&mut self, address: &str) -> Result<Self::Client>;
```
lann: nit: would this be better named `get_client`? `build` suggests to me that it always creates a new one. :shrug: 
```suggestion
    async fn get_client(&mut self, address: &str) -> Result<Self::Client>;
```

crates/factor-outbound-pg/src/lib.rs
lann: Hmm actually this might be a blocker; see below.

lann: _I have deleted several stream-of-consciousness comments about this to summarize here:_

There is a performance hazard with the current impl: this lock is held for the duration of the _async_ call to `ClientFactory::build_client`. If that async call doesn't complete immediately then calls to `open_connection` from other instances (of the same app) will be blocked.

I don't know how much this problem would apply to `PooledTokioClientFactory`; that would depend on the runtime behavior of `Pool::get`.

Pushing interior mutability down into the `ClientFactory` won't fix this on its own but it would allow a fancier impl like using a read lock to check for the entry before falling back to the write lock. Or using [moka](https://docs.rs/moka/0.12.10/moka/).

itowlson: moka worked really nicely.  Thanks for the pointer!

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,89 @@
...
+
+/// A `ClientFactory` that uses a connection pool per address.
+pub struct PooledTokioClientFactory {
+    pools: moka::sync::Cache<String, deadpool_postgres::Pool>,
+}
+
+impl Default for PooledTokioClientFactory {
+    fn default() -> Self {
+        Self {
+            pools: moka::sync::Cache::new(CONNECTION_POOL_CACHE_CAPACITY),
```
itowlson, line 34: @lann I'd value your input on whether we need to provide a way for hosts (e.g. multi-tenant environments) to configure cache (and pool) sizes and policies (e.g. idle time eviction).  But also happy to punt that to a separate PR if that's not yet obvious.

lann: I think punting on configurability is fine.

crates/factor-outbound-pg/src/host.rs
```diff
@@ -71,8 +72,8 @@ fn v2_params_to_v3(
...
```
lann: nit: Some redundant generic constraints remain in this file.

crates/factor-outbound-pg/src/client.rs
```diff
@@ -1,20 +1,89 @@
...
 };
 use tokio_postgres::types::Type;
-use tokio_postgres::{config::SslMode, types::ToSql, Row};
-use tokio_postgres::{Client as TokioClient, NoTls, Socket};
+use tokio_postgres::{config::SslMode, types::ToSql, NoTls, Row};
 
+/// Max connections in a given address' connection pool
+const CONNECTION_POOL_SIZE: usize = 64;
+/// Max addresses for which to keep pools in cache.
+const CONNECTION_POOL_CACHE_CAPACITY: u64 = 1024;
```
lann: I think we should start with a smaller number here. I think its more likely for a large number of idle connections to cause problems than for someone to actually want this many connections open for one app, and you'd almost certainly want this to be configurable at that point anyway.

lann: Picking a differently arbitrary number... 16?

lann (approved):

PR #3045, opened 2025-03-11
fix(templates): update build command for go templates

vdice:
- Updates the build command for go templates per https://github.com/spinframework/spin-go-sdk/issues/10#issuecomment-2590999195


adamreese:
That was removed because it broke `http.Transport` https://github.com/spinframework/spin/pull/2981

vdice:
@adamreese ah, thank you.  Closing!

PR #3047, opened 2025-03-12
app: Wrap inner LockedApp in Arc

lann:
This makes `App::clone` cheaper, which can be signficant for large manifests.

### DIFF
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -7,6 +7,7 @@
 #![deny(missing_docs)]
 
 use std::collections::HashSet;
+use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


rylev (approved):
This seems fine though I guess I have to ask why not make `App` no longer impl `Clone` and require users of `App` to wrap `App` in an `Arc`?

lann:
> why not make `App` no longer impl `Clone` and require users of `App` to wrap `App` in an `Arc`?

I looked at that first but it would require breaking interface changes to really take advantage of it, like changing `RuntimeFactors::configure_app` to take `Arc<App>`.

PR #3052, opened 2025-03-19
Allow "choose from list" parameters in templates

itowlson:
The UI is the usual dialoguer one:

```
$ spin new http-rust tplchoicetest
CUTLERY ME!:
  spork
> another spork
```

Activated by setting `allowed_values = [...]` on the parameter.

We don't need this for any templates in this repo, but https://github.com/spinframework/spin-js-sdk/pull/356 muses about offering a choice of router rather than separate templates.  (Although that idea has gnarlier woes than this!)  Low priority unless JS SDK decides they want to go down that route.

### DIFF
--- a/crates/templates/src/constraints.rs
+++ b/crates/templates/src/constraints.rs
@@ -3,6 +3,7 @@ use regex::Regex;
 #[derive(Clone, Debug)]
 pub(crate) struct StringConstraints {
     pub regex: Option<Regex>,
+    pub allowed_values: Option<Vec<String>>,
 }
 
 impl StringConstraints {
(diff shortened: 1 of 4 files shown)


karthik2804:
@itowlson would this make it necessary to bump the version on the manifest? What would the behavior be where the templates are newer and use the `allowed_values` field but the user has Spin `<3.2`?

itowlson:
@karthik2804 This is why we have tags (`spin/template/v...`) that direct each version of Spin to compatible templates.  We could rev the manifest version but that just shifts the problem to "my Spin 3.1 doesn't understand the new manifest version" I think?

karthik2804:
@itowlson that makes sense thanks! I forgot one would love to see this land. Various tools in the JS ecosystem already provide a setup experience like this and it would be great to have it in Spin as well!

PR #3054, opened 2025-03-21
Validate keys before persisting items in Azure CosmosDB key-value store

ThorstenHans:
This PR updates the key-value store implementation for Azure CosmosDB only. 

When calling `set`, the specified `key` is validated according to the Azure SDK documentation (https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.resource.id?view=azure-dotnet#remarks).

If the key contains an invalid char, the operation is terminated early and a corresponding error is returned.

- [x] passed `cargo build`
- [x] passed `cargo fmt`
- [x] passed `cargo clippy`
- [x] passed `make lint`
- [x] passed `make test`

### DIFF
--- a/crates/key-value-azure/src/store.rs
+++ b/crates/key-value-azure/src/store.rs
@@ -158,7 +158,7 @@ impl Store for AzureCosmosStore {
 
         if key.contains(|c| illegal_chars.contains(&c)) {
             return Err(Error::Other(format!(
-                "Key contains an illegal character. Allowed characters do not include: {}",
+                "Key contains an illegal character. Keys must not include any of: {}",
                 illegal_chars.iter().collect::<String>()
             )));
         }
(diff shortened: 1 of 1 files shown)


lann (approved):

kate-goldenring (approved):

crates/key-value-azure/src/store.rs
```diff
@@ -154,6 +154,15 @@ impl Store for AzureCosmosStore {
     }
 
     async fn set(&self, key: &str, value: &[u8]) -> Result<(), Error> {
+        let illegal_chars = ['/', '\\', '?', '#'];
+
+        if key.contains(|c| illegal_chars.contains(&c)) {
+            return Err(Error::Other(format!(
+                "Key contains an illegal character. Allowed characters do not include: {}",
```
itowlson, line 161: The wording here reads a bit awkwardly to me.  Would you be open to e.g.

```suggestion
                "Key contains an illegal character. Keys must not include any of: {}",
```

or

```
"Key contains an illegal character. The illegal characters are: {}",
```

PR #3060, opened 2025-03-25
`wasi-blobstore` draft support

### LINKED ISSUE #2877: Support WASI-blobstore
Would be nice to have spin support blob storage for unstructured objects such as images and documents.
We could use the WASI blobstore interface https://github.com/WebAssembly/wasi-blobstore

itowlson:
This is lurching towards readiness, although it still needs a lot of tidying.  E.g.

* A bunch of stuff needs renaming!  Oh goodness does it need renaming.  It's embarrassing.
* A bunch of stuff would probably benefit from splitting into modules.
* I'm not sure how to factor-test it.  I can't follow the KV tests because `InstanceBuilder::build()` doesn't actually produce a dispatcher object - instead the factor setup does some terrifying shenanigans to try to hook our streams up to Spin's WASI implementation, which means the dispatcher is produced at link time rather than build time.
* Somebody needs to look at those WASI shenanigans and tell me gently "it doesn't need to be like this Ivan."
* I'm not sure how to integration-test it.  I can integration-test the files provider no worries, but not sure about Azure and S3.
* There are several to-dos (e.g. in the file system one we need to prevent paths escaping from the "container" root path, error messages that helpfully say "oh no").

There are some underspecification issues in the current draft of wasi-blobstore: I _think_ I've made my implementations at least consistent, but I probably need to go back over and check that too.

The implementations in this PR are:

* Azure blob storage
* S3-compatible, although currently it uses the Amazon client so likely bound only to the S3 implementation
* File system, primarily for k8s persistent volumes or other attached storage

I've tested all of them but not tested them hard.

Anyway, I'm sorry this is such a beast and I'm sorry it's still messy - I wanted to do enough to have reasonable confidence in the internal provider API, and I think I'm there bar the renaming and tidying, but it means there's a lot of code, and there's a lot of scar tissue!

### DIFF
--- a/crates/factor-blobstore/src/host.rs
+++ b/crates/factor-blobstore/src/host.rs
@@ -1,0 +1,195 @@
+use anyhow::Result;
+use spin_core::wasmtime::component::ResourceTable;
+use spin_core::{async_trait, wasmtime::component::Resource};
+use spin_resource_table::Table;
+use spin_world::wasi::blobstore::{self as bs};
+use std::{collections::HashSet, sync::Arc};
+use tokio::io::{ReadHalf, SimplexStream};
+use tokio::sync::mpsc;
+use tokio::sync::RwLock;
+
+pub use bs::types::Error;
+
+mod container;
+mod incoming_value;
+mod object_names;
+mod outgoing_value;
+
+pub(crate) use outgoing_value::OutgoingValue;
+
+use crate::DelegatingContainerManager;
+
+// TODO: I feel like the notions of "container" and "container manager" are muddled.
+// This was kinda modelled on the KV StoreManager but I am not sure it has worked.
+// A "container manager" actually manages only one container, making the `get` and
+// `is_defined` functions seemingly redundant. More clarity and better definition
+// is needed here, although the existing code does work!
+//
+// Part of the trouble is, I think, that the WIT has operations for "create container"
+// etc. which implies a level above "container" but whose semantics are very poorly
+// defined (the implication in the WIT is that a `blobstore` implementation backs
+// onto exactly one provider, and if you need to deal with multiple providers then
+// you need to do some double-import trickery, which does not seem right). Clarification
+// sought via https://github.com/WebAssembly/wasi-blobstore/issues/27, so we may need
+// to do some rework once the authors define it more fully.
+
+/// Allows obtaining a container. The only interesting implementation is
+/// [DelegatingContainerManager] (which is what [BlobStoreDispatch] uses);
+/// other implementations currently manage only one container. (See comments.)
+#[async_trait]
+pub trait ContainerManager: Sync + Send {
+    async fn get(&self, name: &str) -> Result<Arc<dyn Container>, Error>;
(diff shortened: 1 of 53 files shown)


itowlson:
Okay there is some silliness going on where it has upgraded an unrelated dependency that now requires Rust 1.83, but that makes one of the examples fling a strange new lint.  I'll see if I can roll things back.  Faugh.

crates/factor-blobstore/src/lib.rs
```diff
@@ -0,0 +1,185 @@
...
+            f
+        }
+
+        let get_data_with_table = ctx.get_data_with_table_fn();
+        let closure = type_annotate(move |data| {
+            let (state, table) = get_data_with_table(data);
+            let wasi = wasmtime_wasi::WasiImpl(host::WasiImplInner {
+                ctx: &mut state.ctx,
+                table,
+            });
```
lann, line 61: I think there is something wrong here and I can't explain why things seem to be working for you.

As far as I can tell this `wasi` state is only being used to carry the `table` here; if that is correct then you could equivalently remove all of the `wasmtime_wasi` code and replace the `BlobStoreDispatch::wasi` field with just this `table` (replacing `self.wasi.table()` with `self.table`). Don't do that yet because I think this code needs to be _more_ complicated to work, not _less_. :sweat_smile: 

The table returned by `get_data_with_table` is scoped to just this factor - not shared with `WasiFactor` - which means the `input-` and `output-stream`s created by `incoming_value_consume_async` and `outgoing_value_write_body` will be in a different table space than those managed by `WasiFactor`. Since the input/output stream methods are all implemented by `WasiFactor` I don't see how any of this could possibly work. :thinking:

lann: The answer is that I forgot how the factors code works. `get_data_with_table` in fact returns a table that is shared between factors, which explains everything...

crates/factor-blobstore/src/host.rs
```diff
@@ -0,0 +1,536 @@
...
+    }
+}
+
+#[async_trait]
+pub trait Finishable: Send + Sync {
+    async fn finish(&mut self);
+}
+
+pub struct BlobStoreDispatch<'a> {
+    allowed_containers: HashSet<String>,
```
lann, line 128: I'm not _entirely_ sure that it can be but if this can be borrowed it would save a clone
```suggestion
    allowed_containers: &'a HashSet<String>,
```

lann: Actually if this one works then probably all of the `Arc`s below could just be borrows. You'd still need the locks for interior mutability of course.

itowlson: @lann I tried it and it hasn't segfaulted yet - thanks for the suggestion!

itowlson:
Closing this, not without regret: the wasi-blobstore proposal is stalled, and if we are ever to do something like this, it will look very different in the light of component model streams and futures (we could probably cannibalise some of the provider code, but little or nothing of the API).

PR #3065, opened 2025-03-29
Allow external configuration of "keep resident" options

alexcrichton:
These options control the behavior of Wasmtime's pooling allocator and can have a big effect on performance, so add environment variables for them so they can at least be tuned after-the-fact for a build of Spin.


lann (approved):

rylev:
@alexcrichton you'll need to sign your commit and this will automatically merge.

alexcrichton:
Apologies I always forget that...

PR #3073, opened 2025-04-01
Spin Improvement Proposal: Modularizing spin.toml

tschneidereit:
This PR adds a SIP describing changes that'd allow developers to split the `spin.toml` manifest up into a main application manifest and sub-manifests for individual contained components.

Signed-off-by: Till Schneidereit <till@tillschneidereit.net>


docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
+title = "SIP 022 - Modularizing spin.toml"
+template = "main"
+date = "2025-03-31T12:00:00Z"
+
+---
+
+Summary: Support moving component definit ions into separate Spin.toml files, similar to Cargo workspaces.
```
fibonacci1729, line 7:
```suggestion
Summary: Support moving component definitions into separate Spin.toml files, similar to Cargo workspaces.
```

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Sub-manifests for components
+
+In this proposal, only components (and variables only accessible locally) can be defined in sub-manifests; triggers still need to be defined in the top-level manifest.
+
+There are multiple reasons for this constraint:
+
+1. Triggers are application-level concerns; they describe the application’s structure and high-level organization. As such, keeping them “at the surface” seems right.
+2. Trigger definitions tend to be very short, so moving them into sub-manifests individually as is done for components seems very boilerplate-y.
+3. Trigger definitions don’t come with their own assets, so any folder containing a sub-manifest for them would only ever contain the manifest, and nothing else.
```
itowlson, line 182: Mostly triggers don't come with component-related config, but we do have an oddity at the moment where WAGI execution has to be specified on the _trigger_ not the _component_ even though it's the component interface that mandates it.  (I _think_ this is to allow different trigger to set different argvs on the same component. Not sure though.)

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+
+*Note that this doesn’t preclude future additions for defining nested components visible only in the scope of an individual top-level component, and destined for composition into said top-level component.*
+
+### Support for defining multiple components in the application manifest
+
+This proposal would not take away support for defining components using the existing syntax; it’d purely add support for the modular approach. The reason is that there are component definitions that are very light-weight; moving them into separate folders would add boilerplate and cognitive overhead, not reduce it.
+
+### Inheriting definitions from the application manifest
+
+Similar to [crates in a Cargo workspace](https://doc.rust-lang.org/1.85.1/cargo/reference/workspaces.html#the-package-table), component manifests can inherit certain definitions from the application manifest, such as the version, description, and authors:
```
itowlson, line 198: Perhaps say more about why you feel this would be useful?  Cargo cares because crates are units of distribution and deployment.

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+
+```toml
+...
+[variables]
+my_variable = { default = "my-value" }
+my_component_my_other_variable = { default = "some-value" }
+...
+```
+
+Any uses in template strings are adapted to use the prefixed variable names.
```
itowlson, line 256: Is this saying that Spin has to rewrite template strings to the new names as part of lockfile creation?

How does this affect providers?  Do secrets in Vault need to use the scoped name?

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+components = ["my-component"]
+```
+
+```toml
+# [project]/my-component/Spin.toml
+
+...
+# Inherit the `my_variable` variable from the application
+[variables]
+my_variable.application = true
```
itowlson, line 241: Why?  This seems like busywork.

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+
+Any uses in template strings are adapted to use the prefixed variable names.
+
+### Exposing variables to content
+
+Currently, manifests have the `[variables]` table for defining variables available to use in the manifest itself, e.g. as part of template strings. They also have `[component.name.variables]` tables for exposing variables to content.
+
+In component manifests, the role of the latter is taken by the new `[variables.exposed]` table.
+
+*Note: we should evaluate whether this table is really needed. Can we instead just expose all variables in a component manifest’s `[variables]` table to content directly?*
```
itowlson, line 264: I don't think we can.  The term "variable" has two meanings:

* A configuration knob exposed to the operator
* A value exposed to the guest

The decoupling is not just for sandboxing, but so that configuration knobs are not required to line up with what the component sees.  E.g. if I publish a component which looks at a "db-url" value, and you publish a component that looks at "db-host" and "db-port" values, the decoupling allows the app developer to define two knobs for the operator, which the application surfaces in different ways to meet different component expectations.

This may have turned out to be generality that we don't need (or won't need until a broader component ecosystem evolves and uses wasi-config), but I'd be cautious about backing it out, and doubly cautious about backing it out _for the special case of sub-manifests._  It _may_ be that sub-manifest local variables are so tightly internally coupled to their components that it's a reasonable call though - just want to be cautious in the evaluation!

itowlson (changes requested):
My sense here is of a slightly uneasy tension between "I'm just providing a handy-dandy way to split an app across files" vs "I'm providing a way to define rich components in isolation from an app."  There are a few things in the proposal that say "rich component in isolation" but there are others that say "just splitting an app."  The reason this matters is that isolated components imply different sandboxing behaviours from split apps - e.g. should a sub-manifest have access to application variables.

It might also be good to reorganise a bit so that the motivation for some of the design decisions is more visible from where the design decisions are listed.

itowlson:
I should add that I do like the implications of isolation rather than mere textual splitting.  One of the challenges I ran into with the previous attempt at this (which was basically naive inclusion) was providing good locations for errors.  Being able to parse and analyse sub-manifests separately, and combine them mostly-reliably only once parsed, should address that nicely.  I am very keen that we think about how to give good pointers in failure cases in this work!

seungjin:
Once I switched my app to a microservice-style structure, my spin.toml became bloated and complicated.
I even considered creating my own spin.toml compiler to generate a final configuration from smaller spin.toml files at build time.

docs/content/sips/022-modularizing-spin-manifests.md
kate-goldenring, line 7:
```suggestion
Summary: Support moving component definitions into separate `spin.toml` files, similar to Cargo workspaces.
```

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+
+Owner(s): [till@fermyon.com](mailto:till@fermyon.com)
+
+Created: March 31, 2025
+
+# Background
+
+As Spin applications grow in complexity, their `spin.toml` manifest files can become large and unwieldy. Large applications with many triggers and components can result in manifest files that span hundreds of lines, making them difficult to navigate, understand, and maintain. It’s very hard to keep related settings grouped, making the application architecture harder to follow.
+
+The current monolithic nature of `spin.toml` files also means that tend to be very conservative when it comes to adding additional features that’d require additions to the manifest. This proposal is directly motivated by the desire to add features that’d make `spin.toml` files in their current form substantially more unwieldy. Those will be described in separate proposals.
```
kate-goldenring, line 17:
```suggestion
The current monolithic nature of `spin.toml` files also means that they tend to be very conservative when it comes to adding additional features that’d require additions to the manifest. This proposal is directly motivated by the desire to add features that’d make `spin.toml` files in their current form substantially more unwieldy. Those will be described in separate proposals.
```

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+spin_manifest_version = ?
+
+[application]
+name = "sentiment-analysis"
+version = "0.1.0"
+authors = ["Caleb Schoepp <caleb.schoepp@fermyon.com>"]
+description = "A sentiment analysis API that demonstrates using LLM inference and
+
+# Components defined in sub-manifests must be included in this list.
+# Entries are treated as folder names, relative to the folder containing the application manifest.
```
kate-goldenring, line 106: So the directory would be laid out like follows?
```sh
├── spin.toml
├── sentiment-analysis
 │   └─── spin.toml
└── kv-explorer
     └─── spin.toml
```

Is there a way we could allow for file exact paths, too?

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+# [project]/sentiment-analysis/spin.toml
+
+# Sections and keys don't need to be scoped explicitly.
+[component]
+name = "sentiment-analysis"
+description = "Component-specific description"
+# Certain keys can be inherited from the application manifest
+authors.application = true
+version.application = true
+source = "target/spin-http-js.wasm"
```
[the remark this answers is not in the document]
kate-goldenring: It feels weird to me to have this be relative to the sub manifest. Shouldn't this already be defined in the sub manifest?

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+version = "0.1.0"
+...
+
+# [project]/my-component/Spin.toml
+...
+# Inherit the `version` field from the application
+version.application = true
+```
+
+### Handling of variables
```
kate-goldenring, line 213: I am having a hard time understanding what make variables unique over other resources, such as environment variables, files, and kv stores. Can you explain why variables are component scoped but other resources are not?

docs/content/sips/022-modularizing-spin-manifests.md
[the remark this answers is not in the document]
itowlson: @kate-goldenring Source being relative to the sub-manifest allows the sub-manifest directory to be portable and self-sufficient.  What's the alternative?   The only one I can think of is "relative to the top-level manifest" which would require the sub-manifest to make assumptions about its own path.

...or am I misunderstanding the comment?  "already be defined in the sub-manifest" makes me worry we are talking at cross purposes...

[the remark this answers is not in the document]
itowlson: I am utterly confused, sorry.  We are _not_ defining source path (for a modularised component) in the top level manifest.  We _are_ definining it in the sub-manifest (which is, presumably, in the component directory).

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+# Proposal
+
+`spin.toml` files contain multiple different kinds of items: application-level metadata, variables, trigger definitions, and component definitions. I propose to support modularizing manifests in the following way:
+
+- Manifests can only be nested one level deep: an application can have a single top-level manifest, and zero or more direct sub-manifests, which cannot have any sub-manifests themselves
+- Application-level metadata must fully be contained in the application’s top-level manifest
+- Triggers must be defined in the top-level manifest
+- Variables can be contained in all manifests
+    - A variable defined in the top-level manifest is visible in all manifests
+    - A variable defined in a sub-manifest is only visible in that manifest
```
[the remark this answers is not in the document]
fibonacci1729: What is our collective appetite 🌯 for revving the locked app format to codify these visibility rules ?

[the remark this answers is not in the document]
fibonacci1729: I _think_ my vote would be too drop these scoping rules for now and require variables be defined in the root manifest.

docs/content/sips/022-modularizing-spin-manifests.md
```diff
@@ -0,0 +1,264 @@
...
+
+Besides specific current goals, moving to a modular manifest format will bring Spin into alignment with related other formats, such as Rust’s [Cargo.toml](https://doc.rust-lang.org/cargo/reference/workspaces.html) or npm’s [package.json](https://docs.npmjs.com/cli/v11/using-npm/workspaces).
+
+Modularizing `spin.toml` files has come up previously, with an attempt at implementing them [here](https://github.com/spinframework/spin/pull/2396).
+
+# Proposal
+
+`spin.toml` files contain multiple different kinds of items: application-level metadata, variables, trigger definitions, and component definitions. I propose to support modularizing manifests in the following way:
+
+- Manifests can only be nested one level deep: an application can have a single top-level manifest, and zero or more direct sub-manifests, which cannot have any sub-manifests themselves
```
fibonacci1729, line 27: What's the motivation for restricting to 1-level nesting?

docs/content/sips/022-modularizing-spin-manifests.md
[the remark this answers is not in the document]
itowlson: @fibonacci1729 I think @tschneidereit is working on something that elaborates on his long-term "north star" thinking, which I believe is what is motivating these rules.  I feel we should hold off until we see that and can decide whether to aim for it directly (at the cost of possibly revving the lockfile) or to do something expedient to solve the immediate problem (at the expense of possible rework and possible compatibility quirks when we do move towards the aspiration).

Specifically I get hints from this document of a vision of sub-manifested components as units of packaging, so that this becomes more than spicy lexical inclusion.  And _if_ so (big if!), there is some conceptual stuff we need to understand about sandboxing and permissions and scoping (e.g. can a component sniff values from the top level without an explicit grant, how do we namespace stuff so components don't need to coordinate on naming, etc.).  I may be over-inferring though!

PR #3075, opened 2025-04-01
Spin Improvement Proposal: Supporting multiple build profiles

tschneidereit:
This PR adds a SIP describing changes that'd allow developers to define and use multiple build profiles for components in a `spin.toml` file

### DIFF
--- a/docs/content/sips/023-build-profiles.md
+++ b/docs/content/sips/023-build-profiles.md
@@ -1,0 +1,141 @@
+title = "SIP 023 - Build profiles"
+template = "main"
+date = "2025-03-31T12:00:00Z"
+
+---
+
+Summary: Support defining multiple build profiles for components, and running a Spin application in with those profiles.
+
+Owner(s): [till@fermyon.com](mailto:till@fermyon.com)
+
+Created: March 31, 2025
+
+# Background
+
+Building and running individual components or entire applications in configurations other than the default one is currently difficult to do with Spin: one has to manually edit the `[component.name.build]` tables for some or all components. This is laborious and error-prone, and can lead to accidentally committing and deploying debug/testing/profiling builds to production.
+
+# Proposal
+
+This very simple proposal contains three parts:
+
+1. Adding an optional `[component.name.profile.profile-name]` table to use for defining additional build profiles of a component
+2. Adding a `-profile [profile-name]` CLI flag and `SPIN_PROFILE` env var to `spin {build, watch, up}` to use specific build profiles of all components (where available, with fallback to the default otherwise)
+3. Adding a (repeatable) `-component-profile name=[profile-name]` CLI flag and `SPIN_COMPONENT_PROFILE="name=[profile-name],..."` env var to use specific profiles for specific components
+
+## Example
+
+The following `spin.toml` file shows how these parts are applied:
+
+```toml
+# (Source: <https://github.com/fermyon/ai-examples/blob/main/sentiment-analysis-ts/spin.toml>)
+
+spin_manifest_version = 2
+
+[application]
+name = "sentiment-analysis"
+version = "0.1.0"
+authors = ["Caleb Schoepp <caleb.schoepp@fermyon.com>"]
+description = "A sentiment analysis API that demonstrates using LLM inference and KV stores together"
+
+[[trigger.http]]
+route = "/api/..."
+component = "sentiment-analysis"
+
+[component.sentiment-analysis]
+source = "target/spin-http-js.wasm"
+allowed_outbound_hosts = []
+exclude_files = ["**/node_modules"]
+key_value_stores = ["default"]
+ai_models = ["llama2-chat"]
+
+[component.sentiment-analysis.build]
+command = "npm run build"
+watch = ["src/**/*", "package.json", "package-lock.json"]
+
+# Debug build of the `sentiment-analysis` component:
+[component.sentiment-analysis.profile.debug]
+source = "target/spin-http-js.debug.wasm"
+
+[component.sentiment-analysis.profile.debug.build]
+command = "npm run build:debug"
+watch = ["src/**/*", "package.json", "package-lock.json"]
+
+[[trigger.http]]
+route = "/..."
+component = "ui"
+
+# The `ui` component doesn't have a debug build, so the release build will always be used.
+[component.ui]
+source = { url = "<https://github.com/fermyon/spin-fileserver/releases/download/v0.0.1/spin_static_fs.wasm>", digest = "sha256:650376c33a0756b1a52cad7ca670f1126391b79050df0321407da9c741d32375" }
+allowed_outbound_hosts = []
+files = [{ source = "../sentiment-analysis-assets", destination = "/" }]
+
+[variables]
+kv_explorer_user = { required = true }
+kv_explorer_password = { required = true }
+
+[[trigger.http]]
+component = "kv-explorer"
+route = "/internal/kv-explorer/..."
+
+[component.kv-explorer]
+source = { url = "<https://github.com/fermyon/spin-kv-explorer/releases/download/v0.10.0/spin-kv-explorer.wasm>", digest = "sha256:65bc286f8315746d1beecd2430e178f539fa487ebf6520099daae09a35dbce1d" }
+allowed_outbound_hosts = ["redis://*:*", "mysql://*:*", "postgres://*:*"]
+# add or remove stores you want to explore here
+key_value_stores = ["default"]
+
+# Uses a pre-built debug version of the component.
+[component.kv-explorer.profile.debug]
+source = { url = "<https://github.com/fermyon/spin-kv-explorer/releases/download/v0.10.0/spin-kv-explorer.debug.wasm>", digest = ".." }
+
+[component.kv-explorer.variables]
+kv_credentials = "{{ kv_explorer_user }}:{{ kv_explorer_password }}"
+
+```
+
+The application defined in this manifest can be run in various configurations:
+
+- `spin up`: uses the release/default builds for everything
+- `spin up --profile debug`: uses builds of the profile named `debug` of all components that have them, default builds for the rest
+- `spin up --component-profile sentiment-analysis=debug`: uses the debug build of just the `sentiment-analysis` component, default builds for everything else
+
+## Profile-overridable fields
+
+This proposal is focused on build configurations. As such, the fields that are supported in `[component.component-name.profile.profile-name]` tables are limited to:
+
+- `source`
+- `command`
+- `watch`
+
+This set can be expanded in changes building on this initial support, as adding additional fields should always be backwards-compatible.
+
+## Profiles are all-or-nothing
+
+When running an application with a named profile applied, components that don’t define that profile fall back to the default build configuration. The same isn’t true for individual keys in a profile: if e.g. a profile contains the `source` field but no `command` field, running `spin build` will not try to run the default configuration’s build command.
+
+## Local testing only
+
+At least for now, this proposal is focused entirely on local testing: deployment through plugins is not proposed to gain support for a `--profile` flag.
+
+## Backwards-compatibility
+
+This proposal should be fully backwards-compatible.
+
+## Implementation concerns
+
+I can't see any major implementation concerns: this should hopefully be pretty straightforward.
+
+## Alternatives considered
+
+Instead of adding generalized profiles support, we could add just a hard-coded `debug` profile. This would simplify the syntax a little bit, but at the cost of flexibility.
+
+We could do so by using syntax such as
+
+```toml
+[component.sentiment-analysis.debug]
+...
+```
+
+and changing the `--profile [profile-name]` flag to instead be `--debug`, and `--component-profile sentiment-analysis=debug` be `--debug-component sentiment-analysis`.
+
+I’m not proposing this alternative because the benefits seem much smaller than the downsides.


fibonacci1729 (approved):
LGTM; proposal is quite simple and seems very straightforward to implement.

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+Instead of adding generalized profiles support, we could add just a hard-coded `debug` profile. This would simplify the syntax a little bit, but at the cost of flexibility.
+
+We could do so by using syntax such as
+
+```toml
+[component.sentiment-analysis.debug]
+...
+```
+
+and changing the `--profile [profile-name]` flag to instead be `--debug`, and `--component-profile sentiment-analysis=debug` be `--debug-component sentiment-analysis`.
```
itowlson, line 139: I'd like us to at least consider `--debug` as an alias for `--profile=debug` just for convenience's sake!  I think this is the 80% case and it would be nice to streamline it.

itowlson (changes requested):

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+
+[component.kv-explorer.variables]
+kv_credentials = "{{ kv_explorer_user }}:{{ kv_explorer_password }}"
+
+```
+
+The application defined in this manifest can be run in various configurations:
+
+- `spin up`: uses the release/default builds for everything
+- `spin up --profile debug`: uses builds of the profile named `debug` of all components that have them, default builds for the rest
```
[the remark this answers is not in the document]
tschneidereit: yeah, that's a very good point. One, and perhaps the only real one, option is to make `spin up` imply `spin build`.

This is the behavior `cargo run` has, and it makes sense to me. Cargo of course has the benefit of having much deeper understanding of what a build implies, but maybe that's not actually key here?

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+
+This proposal should be fully backwards-compatible.
+
+## Implementation concerns
+
+I can't see any major implementation concerns: this should hopefully be pretty straightforward.
+
+## Alternatives considered
+
+Instead of adding generalized profiles support, we could add just a hard-coded `debug` profile. This would simplify the syntax a little bit, but at the cost of flexibility.
```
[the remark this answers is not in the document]
tschneidereit: I'll add some language to this, yes. The key thing for now is that I know that other build tools that started with "release and debug are enough for everybody" inevitably came to the conclusion that no, they're not.

One immediate example of a useful additional profile that's still solidly in the "build tool" space is, erm, profiling: release builds with debug symbols to enable profiling tools to gather decent stacks.

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+# Background
+
+Building and running individual components or entire applications in configurations other than the default one is currently difficult to do with Spin: one has to manually edit the `[component.name.build]` tables for some or all components. This is laborious and error-prone, and can lead to accidentally committing and deploying debug/testing/profiling builds to production.
+
+# Proposal
+
+This very simple proposal contains three parts:
+
+1. Adding an optional `[component.name.profile.profile-name]` table to use for defining additional build profiles of a component
+2. Adding a `-profile [profile-name]` CLI flag and `SPIN_PROFILE` env var to `spin {build, watch, up}` to use specific build profiles of all components (where available, with fallback to the default otherwise)
```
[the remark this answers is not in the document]
tschneidereit: Same as with `spin up`, but even more strongly, I suspect that the proper answer here is "`spin [deploy by whatever name]` implies `spin build --profile=release`".

`vsce package` and `vsce publish` are behaving this way, and I think that makes all the sense in the world.

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+3. Adding a (repeatable) `-component-profile name=[profile-name]` CLI flag and `SPIN_COMPONENT_PROFILE="name=[profile-name],..."` env var to use specific profiles for specific components
+
+## Example
+
+The following `spin.toml` file shows how these parts are applied:
+
+```toml
+# (Source: <https://github.com/fermyon/ai-examples/blob/main/sentiment-analysis-ts/spin.toml>)
```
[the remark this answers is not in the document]
tschneidereit: fair enough!

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+command = "npm run build:debug"
+watch = ["src/**/*", "package.json", "package-lock.json"]
+
+[[trigger.http]]
+route = "/..."
+component = "ui"
+
+# The `ui` component doesn't have a debug build, so the release build will always be used.
+[component.ui]
+source = { url = "<https://github.com/fermyon/spin-fileserver/releases/download/v0.0.1/spin_static_fs.wasm>", digest = "sha256:650376c33a0756b1a52cad7ca670f1126391b79050df0321407da9c741d32375" }
```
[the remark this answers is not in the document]
tschneidereit: I..have no idea: this is really just copied from the source

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+
+- `source`
+- `command`
+- `watch`
+
+This set can be expanded in changes building on this initial support, as adding additional fields should always be backwards-compatible.
+
+## Profiles are all-or-nothing
+
+When running an application with a named profile applied, components that don’t define that profile fall back to the default build configuration. The same isn’t true for individual keys in a profile: if e.g. a profile contains the `source` field but no `command` field, running `spin build` will not try to run the default configuration’s build command.
```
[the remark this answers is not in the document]
tschneidereit: will expand on that, yes

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,141 @@
...
+
+## Local testing only
+
+At least for now, this proposal is focused entirely on local testing: deployment through plugins is not proposed to gain support for a `--profile` flag.
```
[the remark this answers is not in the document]
tschneidereit: Good point about env vars. I guess if we go with "we always ensure a fresh release build before publishing" we can do that by running the implied `spin build` part of the release process be forced to do a release build, env vars be damned

docs/content/sips/023-build-profiles.md
tschneidereit: I'm definitely okay with that, and will add it, yes

docs/content/sips/023-build-profiles.md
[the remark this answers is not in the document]
tschneidereit: oh wait no, this must've been some weird artefact of copying things over from notion

docs/content/sips/023-build-profiles.md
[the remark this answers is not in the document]
tschneidereit: I opted to go with the most conservative option here of rejecting the manifest if a required field is missing in any profile, but happy to discuss that further.

docs/content/sips/023-build-profiles.md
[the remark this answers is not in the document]
tschneidereit: Part of the changes I'm applying are to introduce an explicit `--release` flag, and making publishing operations run `SPIN_COMPONENT_PROFILE="" spin build --release` implicitly. I think that combination should address this concern.

tschneidereit (commented):
Thank you very much for the detailed feedback! I think I addressed all of it, but some of the ways I did probably warrant further discussion

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,146 @@
...
+
+Building and running individual components or entire applications in configurations other than the default one is currently difficult to do with Spin: one has to manually edit the `[component.name.build]` tables for some or all components. This is laborious and error-prone, and can lead to accidentally committing and deploying debug/testing/profiling builds to production.
+
+# Proposal
+
+This very simple proposal consists of four parts:
+
+1. Adding an optional `[component.name.profile.profile-name]` table to use for defining additional build profiles of a component
+2. Adding a `-profile [profile-name]` CLI flag and `SPIN_PROFILE` env var to `spin {build, watch, up}` to use specific build profiles of all components (where available, with fallback to the default otherwise)
+   1. By popular demand, the `debug` profile can be selected using the `--debug` CLI flag
```
tschneidereit, line 23: While I added the `--debug` flag for now, I think there might be a reason for not doing so after all: we might want to later add support for actual debugging to Spin, and would in that case probably regret having used this flag for profile selection.

itowlson: @tschneidereit Ah, yep. Reserving `--debug` for "I want to actually attach a debugger" is a good point.  So `spin up --build --debug` would build the debug profile and run under the debugger; but `spin up --build --profile=debug` would build the debug profile but run it without actually debugging.  The first case does feel it should (eventually) be the "common case" and get the fancy flag.

docs/content/sips/023-build-profiles.md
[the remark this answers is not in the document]
itowlson: `spin up` isn't like `cargo run` though.  It's more like `wasmtime run`.  It cares _only_ about the presence of the binary; its job is to get the binary up and running as quickly as possible.

(At least that's how it is today.  We can, of course, change that; but making it default to building would be a breaking change for certain scenarios.  Which is not a blocker, but we'd want to socialise it well in advance of the change.)

I'd be concerned about a philosophy of "build everywhere."  Okay, I built it to make sure that it typechecked and was warning free, now let me run it, huh, looks like it built again.  Push it to a registry, built again; deploy it, built _again._  For sure Rust components will fairly quickly (and _fairly_ quietly) figure out that the build is a no-op.  But rebuilding a single minimal JS component (with no changes) takes more than half a second on my computer, and produces 2/3 of a screen of npm spew   Maybe that's an extreme case but it does give me collywobbles about defaulting to rebuilding with every single operation.  I do appreciate the consistency guarantees it provides though - after all, even today we have a risk of "oops I forgot to build before pushing"!

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,146 @@
...
+   2. `release` is the default profile, but can also be explicitly named for regularity, including with the `--release` alias
```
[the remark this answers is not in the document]
kate-goldenring: I like release being the default in part because `spin build` then `spin registry push` is a promoted flow and i like orienting around defaulting to pushing release builds

radu-matei:
I believe we have merged the feature in #3246 but the SIP has not landed yet.

It also seems the implementation diverged from the SIP — would it be possible to update the SIP and merge it so we have a record of what we shipped?

Thanks!

fibonacci1729:
Cc/ @karthik2804 are you working on bringing the SIP up-to-date?

karthik2804:
@fibonacci1729 yes I am.

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+
+Created: March 31, 2025
+
+# Background
+
+Building and running individual components or entire applications in configurations other than the default one is currently difficult to do with Spin: one has to manually edit the `[component.name.build]` tables for some or all components. This is laborious and error-prone, and can lead to accidentally committing and deploying debug/testing/profiling builds to production.
+
+# Proposal
+
+This very simple proposal consists of four parts:
```
itowlson: for small values of 'four' _grin_

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
+title = "SIP 023 - Build profiles"
+template = "main"
+date = "2025-03-31T12:00:00Z"
+
+---
+
+Summary: Support defining multiple build profiles for components, and running a Spin application in with those profiles.
```
itowlson:
```suggestion
Summary: Support defining multiple build profiles for components, and running a Spin application with those profiles.
```

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Uses a pre-built debug version of the component.
+[component.kv-explorer.profile.debug]
+source = { url = ".../spin-kv-explorer.debug.wasm", digest = "..." }
+```
+
+The application defined in this manifest can be run in various configurations:
+
+- `spin up`: uses the release/default builds for everything
+- `spin up --profile debug` or `spin up --debug`: uses builds of the profile named `debug` of all components that have them, default builds for the rest
```
itowlson:
```suggestion
- `spin up --profile debug`: uses builds of the profile named `debug` of all components that have them, default builds for the rest
```

(I think we deferred the `spin up --debug` thingy to allow for more thinking)

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+- `spin up --component-profile sentiment-analysis=debug`: uses the debug build of just the `sentiment-analysis` component, default builds for everything else
```
itowlson: I didn't do this.  It can be a future enhancement if needed.

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Details of profile selection
+
+A profile can be selected `--profile=[profile-name]` CLI flag to `spin {build, up, watch, deploy}`
```
itowlson: `registry push`

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Profile-overridable fields
+
+This proposal is focused on build configurations. As such, the fields that are supported in `[component.component-name.profile.profile-name]` tables are limited to:
+
+- `source`
+- `command`
```
itowlson: `build.command`

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+- `watch`
```
itowlson: Pretty sure you can't do this

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+command = "npm run build"
+watch = ["src/**/*", "package.json", "package-lock.json"]
+
+# Debug build of the `sentiment-analysis` component:
+[component.sentiment-analysis.profile.debug]
+source = "target/spin-http-js.debug.wasm"
+
+[component.sentiment-analysis.profile.debug.build]
+command = "npm run build:debug"
+watch = ["src/**/*", "package.json", "package-lock.json"]
```
itowlson: I don't think this is allowed in a profile

docs/content/sips/023-build-profiles.md
```diff
@@ -0,0 +1,111 @@
...
+# Background
+
+Building and running individual components or entire applications in configurations other than the default one is currently difficult to do with Spin: one has to manually edit the `[component.name.build]` tables for some or all components. This is laborious and error-prone, and can lead to accidentally committing and deploying debug/testing/profiling builds to production.
+
+# Proposal
+
+This very simple proposal consists of four parts:
+
+1. Adding an optional `[component.name.profile.profile-name]` table to use for defining additional build profiles of a component
+2. Adding a `-profile [profile-name]` CLI flag to `spin {build, watch, up, deploy}` to use specific build profiles of all components (where available, with fallback to the default otherwise)
```
itowlson: SUPER MEGA NIT

```suggestion
2. Adding a `--profile [profile-name]` CLI flag to `spin {build, watch, up, deploy}` to use specific build profiles of all components (where available, with fallback to the default otherwise)
```

itowlson: also `registry push` (but maybe this intro section doesn't need to get specific about individual commands)

docs/content/sips/023-build-profiles.md
itowlson: also ULTRA NIT but I hate it when writers tell me that the thing is simple. If it's simple, I'll figure it out for myself. If it's not, you've made me feel stupid for not getting it.  I suppose it's meant to be reassuring but fie!

anyway now back to your regularly scheduled punctuation complaints

docs/content/sips/023-build-profiles.md
karthik2804: This is specifically in reference to the `watch`  right?

docs/content/sips/023-build-profiles.md
karthik2804: fixed.

docs/content/sips/023-build-profiles.md
itowlson: Correct.  Looking at the code, the only override in `build` is the `command`.  You can't override the working directory or watch list (because there seems no reason for those to vary across profiles - you're building from the same things, just in different configurations).

PR #3083, opened 2025-04-01
Generate CLI reference

itowlson:
An experiment in support of https://github.com/spinframework/spin-docs/issues/72

cc @kate-goldenring 

e.g. `spin maintenance generate-reference -o REFME.md`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,6 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 bytes = { workspace = true }
 clap = { version = "3.2.24", features = ["derive", "env"] }
-clap-markdown = { git = "https://github.com/itowlson/clap-markdown", rev = "9e484a05ac70bfb928453559fa45a49534dac121" }
 clearscreen = "3"
 comfy-table = "7"
 command-group = "2"
@@ -33,6 +32,7 @@ lazy_static = "1.5"
 levenshtein = "1"
 nix = { version = "0.29", features = ["signal"] }
 path-absolutize = "3"
+pretty_assertions = "1.3"
 regex = { workspace = true }
 reqwest = { workspace = true }
 rpassword = "7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 bytes = { workspace = true }
 clap = { version = "3.2.24", features = ["derive", "env"] }
+clap-markdown = { git = "https://github.com/itowlson/clap-markdown", rev = "9e484a05ac70bfb928453559fa45a49534dac121" }
 clearscreen = "3"
 comfy-table = "7"
 command-group = "2"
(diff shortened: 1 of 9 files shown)


src/commands/maintenance.rs
```diff
@@ -0,0 +1,32 @@
+use std::path::PathBuf;
+
+use clap::{Parser, Subcommand};
+
+/// Commands for working with OCI registries to distribute applications.
```
kate-goldenring, line 5: This looks like it needs updating

Cargo.toml
```diff
@@ -19,6 +19,7 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 bytes = { workspace = true }
 clap = { version = "3.2.24", features = ["derive", "env"] }
+clap-markdown = { git = "https://github.com/itowlson/clap-markdown", branch = "clap-v3-compat" }
```
kate-goldenring, line 22: Should we move the fork to spinframework's repo?

kate-goldenring (commented):
Thanks for looking into this @itowlson. Can you also commit the initial CLI.md?

src/commands/maintenance.rs
itowlson: Thanks!  Fixed

Cargo.toml
itowlson: I am not sure what to do with the fork.  It can't live in my personal account I agree.  But it feels awkward to have something so disconnected from Spin and so specifically about Clap in spinframework.  Maybe that's just me.  We could make it a crate in the Spin project, but then we would lose traceability to the original repo and author, which feels rude.

Anyway I'm happy to follow whatever you and other folks more steeped on open-source etiquette advise!

itowlson:
I tweaked this so it defaults to writing to stdout.  My thought is that we can have a GH workflow (or step in the release workflow) that does something like:

* cat the Bartholomew frontmatter of the CLI reference page to a file e.g. `cli-reference.md`
* `spin maintenance generate-reference >>cli-reference.md`
* PR `cli-reference.md` to the appropriate place in `spin-docs`

(We can't use `MarkdownOptions::title()` to set the front matter as it always prepends a `#` heading marker.  Although I suppose I could change that too.  Why not go wild.)

I think we've done something like the PR thing before, but I'm a bit lost on the details, sorry.

kate-goldenring:
@itowlson this all looks great. IMO, i think we can move this to ready for review and then integrate it with the docs CI as you mention

itowlson:
After discussion with @kate-goldenring I've moved the forked third-party code into the project (rather than having a separate repo unrelated to the Spin project).  I've tried to provide appropriate attribution and credit, and have reproduced the licences and copyright statements.  I've done this as a separate commit so we can easily unpick the change if I've done it wrong or I was wrong to do it!  Thoughts welcome.

fibonacci1729 (approved):

NOTICES.md
kate-goldenring, line 1: nit: i'd rename this to NOTICE

src/clap_markdown/LICENSE-MIT
kate-goldenring, line 1: when a project has multiple licenses, i believe you only need to copy over one. I think we can delete this

kate-goldenring (approved):

PR #3086, opened 2025-04-02
Generate the CLI reference as part of build

itowlson:
This is a work in progress, but I'm afraid I have to do it via PR because it's figuring out a GitHub action.  The initial cut is just to try to generate a Markdown artifact on build; once that works correctly, I'll figure out how to PR it to spin-docs as part of release (or on manual dispatch or whatever).

I apologise in advance for all the noise.


itowlson:
New plan is to tackle this from the other direction.  Closing (at least for now)

Issue #2904, opened 2024-11-04
CI: Eliminate Action warnings


vdice:
Chore issue to cleanup/eliminate warnings seen in GH workflows (for example: a recent run [of the Rust workflow on main](https://github.com/fermyon/spin/actions/runs/11673215392)).  This mostly involves bumping action versions.

PR #3089, opened 2025-04-04
Fix cache warning in GH actions

itowlson:
The last one from #2904 

I know, but my brain is not very thinkful right now


itowlson:
behold! https://github.com/spinframework/spin/actions/runs/14255593022

vdice (approved):

PR #3095, opened 2025-04-08
SQLite affected row count and last_insert_rowid

### LINKED ISSUE #3092: Can we have `affected_row_count` and `last_insert_rowid` in QueryResult?
We currently have 
```
pub struct QueryResult {
    pub columns: Vec<String>,
    pub rows: Vec<RowResult>,
}
```
https://docs.rs/spin-sdk/latest/spin_sdk/sqlite/struct.QueryResult.html
But typically database sdk gives more stuffs. (ie. https://docs.turso.tech/sdk/http/reference#response-types) 
Can we at least have `affected_row_count` and `last_insert_rowid` (more better)?

itowlson:
Fixes #3092.

**~WIP for now~**

This adds two functions to the SQLite `connection` resource:

* `changes()` - the number of rows affected by the last insert/delete/update
* `last_insert_rowid()` - the rowid of the last inserted row

The names are taken from the SQLite API, and I'm in favour of sticking with that standard rather than giving them names that might be more self-documenting (stop shuffling your feet at the back, `changes`).

WIPness: the new functions appear to work but:

* ~It needs tests~ Added to the existing SQLite test
* It needs trying with libSQL
* ~It needs a world update (for dev purposes I shoved it into the 3.0 world but that's not appropriate for shipping)~ Done - see separate comment below
* ~I want to try to remove the v2/v3 duplication~ Done (hopefully satisfactorily)

I'm putting it into PR now so that 1. the integration tests pick up any regressions and 2. if folks have other things they want to put in then we can consider rolling them into a single rev.

The new interface is backward compatible and I did consider doing one of those "since" things to make it a minor version rev, which would eliminate all the duplication, but I provisionally went for a new interface because we are trying to shift from the `fermyon:` namespace to a `spin:` namespace.  Happy to receive feedback on this!

### DIFF
--- a/wit/world.wit
+++ b/wit/world.wit
@@ -11,5 +11,6 @@ world platform {
   include fermyon:spin/platform@2.0.0;
   include wasi:keyvalue/imports@0.2.0-draft2;
   import spin:postgres/postgres@3.0.0;
+  import spin:sqlite/sqlite@3.0.0;
   import wasi:config/store@0.2.0-draft-2024-09-27;
 }
--- a/crates/factor-sqlite/src/host.rs
+++ b/crates/factor-sqlite/src/host.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 
 use spin_factors::wasmtime::component::Resource;
 use spin_factors::{anyhow, SelfInstanceBuilder};
+use spin_world::spin::sqlite::sqlite as v3;
 use spin_world::v1::sqlite as v1;
 use spin_world::v2::sqlite as v2;
 use tracing::field::Empty;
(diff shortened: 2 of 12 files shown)


itowlson:
I've tentatively created a `spin:spin@3.2.0` package to contain this.  The alternative to to retain `fermyon:spin@3`, bump it to 3.1 and use an `@since`.  This is more parsimonious and better expresses the compatibility guarantees, but doesn't contribute to moving us off the `fermyon` namespace.

I've done this as a separate commit do we can easily change or revert it independently of the functional work.  Howl at me if you have something different in mind!

wit/world.wit
```diff
@@ -1,4 +1,4 @@
-package fermyon:spin@3.0.0;
+package spin:spin@3.2.0;
```
lann, line 1:
```suggestion
package spin:up@3.2.0;
```

This may be a bit "cute" but I think it accurately reflects the use of this package, which is describing the world supported by `spin up`.

itowlson: Ooh I like `spin:up` (and agree with your objections to `spin:spin`).  Thank you!

lann (approved):

fibonacci1729 (approved):

PR #3103, opened 2025-04-10
Run cargo update on timer example

rylev:
Snuck in a free typo fix as well 😄

### DIFF
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -146,7 +146,7 @@ impl Router {
     pub fn contains_reserved_route(&self) -> bool {
         self.router
             .iter()
-            .any(|(_spec, handker)| handker.based_route.starts_with(crate::WELL_KNOWN_PREFIX))
+            .any(|(_spec, handler)| handler.based_route.starts_with(crate::WELL_KNOWN_PREFIX))
     }
 
     /// This returns the component ID that should handle the given path, or an error
--- a/crates/key-value-aws/Cargo.toml
+++ b/crates/key-value-aws/Cargo.toml
@@ -11,9 +11,11 @@ rust-version.workspace = true
 [dependencies]
 anyhow = { workspace = true }
 async-once-cell = "0.5.4"
-aws-config = "1.1.7"
+# Turn off default features to avoid pulling in "aws-smithy-runtime/default-https-client" which messes up tls provider selection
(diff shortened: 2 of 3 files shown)


rylev:
@itowlson @ogghead Unfortunately to get things to work I needed to do 2ef489c2af9801f776e94422583a7139b0ade2c7 which requires changing how the AWS SDK is compiled. We don't have a test for that functionality so I have no way to verify that the change I made doesn't break things. Do I need to create a dynamo DB account to test this or is there some easier way? If you already have an account handy, and can test these changes, that would also be super helpful!

itowlson:
@ogghead If you could test with real AWS that would be super useful.  From the comment in the code, it seems like the change affects something in TLS, and I'd like to be sure that we can still accept DynamoDB etc. TLS certs without `smithy` being in play.  We wouldn't get that validation from using a local workalike.  (It sounds like a great way to do integration testing though - thanks for highlighting it.)

@rylev Could you expand on "get things to work" please?  As far as I know this code has worked for several months - what has turned out to not work?  The commit message refers to `ring` - is this a something like we can't rev ring (or can't configure it or something) because it fights with AWS' default TLS provider?  Or is it as Lann says in #3104 that the AWS TLS crate is ropey in some environments - that does ring bells, some woe that caused us to try to avoid it in the past...?

itowlson:
Thanks for the explanation!  It all LGTM - only holding off approving until we get confirmation that it works the real AWS service.

ogghead:
Tested this against my AWS account by:
1. Checking out this branch
2. Building and installing Spin with `make build && make install`
3. Running the Spin app in https://github.com/ogghead/TestDynamo with `spin up --build --runtime-config-file rt-config.toml`

Didn't see any issues with connection to DynamoDB and data is written + read successfully. LGTM!

itowlson (approved):

itowlson:
Thanks @ogghead!

ogghead:
No problem, happy to help!

> It sounds like a great way to do integration testing though - thanks for highlighting it.

If you think it's worth adding, I'm happy to create a ticket + PR for DynamoDBLocal connectivity.

itowlson:
@ogghead You're probably the best judge of if it's worth adding. But it would be great to get this stuff under test if the local one gives reasonable confidence.  Thanks for the offer!

itowlson:
@rylev I'm going to merge this and see if it fixes our PR integration test woes.  Hope that's okay!

PR #3107, opened 2025-04-11
Less allocations in router

rylev:
This trades a bit of code complexity for less allocations in routing. Given routing can be quite performance sensitive, in my opinion this tradeoff is worth it. 

On every request we no longer:
* Clone a `RouteHandler` (3-4 `String`s)
* Clone the `trailing_wildcard` `String`

In service chaining we also no longer:
* allocate two strings in the synthetic `RouteHandler` unnecessarily. 
* unnecessarily clone the `component_id`

### DIFF
--- a/crates/http/src/routes.rs
+++ b/crates/http/src/routes.rs
@@ -165,19 +165,21 @@ impl Router {
     /// by matching wildcard patterns with the longest matching prefix.
     pub fn route<'path, 'router: 'path>(
         &'router self,
-        p: &'path str,
+        path: &'path str,
     ) -> Result<RouteMatch<'router, 'path>> {
         let best_match = self
             .router
-            .best_match(p)
-            .ok_or_else(|| anyhow!("Cannot match route for path {p}"))?;
+            .best_match(path)
+            .ok_or_else(|| anyhow!("Cannot match route for path {path}"))?;
 
         let route_handler = best_match.handler();
 
         Ok(RouteMatch {
-            route_handler: Cow::Borrowed(route_handler),
-            best_match: Some((p, best_match)),
-            trailing_wildcard: None,
+            inner: RouteMatchKind::Real {
+                route_handler,
+                best_match,
+                path,
+            },
         })
     }
 }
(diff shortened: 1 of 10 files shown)


lann (approved):

crates/http/src/routes.rs
```diff
@@ -235,25 +219,28 @@ impl fmt::Display for ParsedRoute {
...
 
 /// A routing match for a URL.
-pub struct RouteMatch {
-    route_handler: RouteHandler,
-    named_wildcards: HashMap<String, String>,
+pub struct RouteMatch<'router, 'path> {
+    /// The route handler that matched the path.
+    route_handler: Cow<'router, RouteHandler>,
+    /// An optional best match for the path.
+    best_match: Option<(&'path str, routefinder::Match<'router, 'path, RouteHandler>)>,
```
itowlson, line 226: How would you feel about replacing the tuple with a struct?  I assume this would be no more expensive, but it could make it more self-documenting.

If you're not keen, maybe tweak the doc comment to say what the tuple elements are?

crates/http/src/routes.rs
```diff
@@ -291,13 +278,45 @@ impl RouteMatch {
...
-    pub fn named_wildcards(&self) -> &HashMap<String, String> {
-        &self.named_wildcards
+    pub fn named_wildcards(&self) -> HashMap<String, String> {
+        let Some((_, best_match)) = &self.best_match else {
+            return HashMap::new();
+        };
+        best_match
+            .captures()
+            .iter()
+            .map(|(k, v)| (k.to_owned(), v.to_owned()))
```
itowlson, line 288: Is it worth exploring the possibility of optimising away these allocations?  I guess they are going to get allocated anyway when we populate the headers though...

crates/http/src/routes.rs
```diff
@@ -291,13 +278,45 @@ impl RouteMatch {
...
+        if let Some(wildcard) = &self.trailing_wildcard {
+            return wildcard.clone();
+        };
+
+        // Otherwise, we need to extract it from the best match if we have it.
+        let Some((path, best_match)) = &self.best_match else {
+            return String::new();
+        };
+        best_match
+            .captures()
```
itowlson, line 304: Is it worth further optimising this by running `Match::captures()` only once, e.g. during the construction of the `RouteMatch` object?  The `captures()` implementation is non-trivial - I'm not sure what the lifetime considerations would be but it seems like you have the `'router` and `'path` lifetimes already.

I realise this is existing code, just raising it since we are in optimisation mode!

itowlson (approved):
This all looks good - I did have some suggestions / questions, but they are non-blocking, so feel free to ignore them.

crates/http/src/routes.rs
rylev: I tweaked the code to hopefully be a bit more self documenting: https://github.com/spinframework/spin/pull/3107/commits/ee1560633bccd164646e11619158c53a51e5731b

Let me know what you think.

crates/http/src/routes.rs
rylev: Done! 🎉

itowlson (approved):
I like this - thanks for taking that second look.  The explicit distinction between the "I ran this through the router and here's what it came up with" vs "service chaining express delivery _plonk_" makes that chunk of code much safer and more self-explanatory than my Option bodge.

PR #3111, opened 2025-04-14
Support Liquid partials in templates

### LINKED ISSUE #3110: Support for using liquid partials in templates.
In the JS templates, we recently introduced an option to choose the kind of router the user would like to be bootstrapped. This currently involves writing `case` statements with blocks of code for each option ([ref](https://github.com/spinframework/spin-js-sdk/blob/main/templates/http-ts/content/src/index.ts.tmpl)). From a maintainability perspective, it would be nice if we could use partials, enabling us to have multiple files that we can include based on the user option.

itowlson:
Fixes #3110 

How it works:

* There is a new recognised directory `partials` in the template `metadata` directory
* Partials are included as `.liquid` files in the `partials` directory e.g. `fancy-router.liquid`
* Content files reference partials using the `render` directive e.g. `{% render 'fancy-router' %}`
* If a partial uses a parameter then it must be explicitly passed e.g. `{% render 'fancy-router', base-path: base-path %}`. (Yes I realise this could be annoying but it appears to be just how Liquid works. https://liquidjs.com/tags/render.html#Passing-Variables)

See `crates/templates/tests/templates/partials` for an example.

@karthik2804 @ThorstenHans If you have an opportunity to test this on the JS SDK templates then it would be super useful to have your feedback or to let me know if there are bugs which my testing didn't catch!

### DIFF
--- a/crates/templates/src/run.rs
+++ b/crates/templates/src/run.rs
@@ -19,6 +19,9 @@ use crate::{
     template::Template,
 };
 
+/// A set of partials to be included in a Liquid template.
+type PartialsBuilder = liquid::partials::EagerCompiler<liquid::partials::InMemorySource>;
+
 /// Executes a template to the point where it is ready to generate
 /// artefacts.
 pub struct Run {
@@ -104,6 +107,9 @@ impl Run {
             };
         }
 
+        let partials = self.partials()?;
+        let parser = Self::template_parser(partials)?;
+
         self.validate_provided_values()?;
 
         let files = match self.template.content_dir() {
@@ -112,7 +118,7 @@ impl Run {
                 let from = path
                     .absolutize()
                     .context("Failed to get absolute path of template directory")?;
-                self.included_files(&from, &to)?
+                self.included_files(&from, &to, &parser)?
             }
         };
 
@@ -120,7 +126,7 @@ impl Run {
             .template
             .snippets(&self.options.variant)
             .iter()
-            .map(|(id, path)| self.snippet_operation(id, path))
+            .map(|(id, path)| self.snippet_operation(id, path, &parser))
             .collect::<anyhow::Result<Vec<_>>>()?;
 
         let extras = self
@@ -151,7 +157,12 @@ impl Run {
         }
     }
 
-    fn included_files(&self, from: &Path, to: &Path) -> anyhow::Result<Vec<RenderOperation>> {
+    fn included_files(
+        &self,
+        from: &Path,
+        to: &Path,
+        parser: &liquid::Parser,
+    ) -> anyhow::Result<Vec<RenderOperation>> {
         let gitignore = ".gitignore";
         let mut all_content_files = Self::list_content_files(from)?;
         // If user asked for no_vcs
@@ -164,7 +175,7 @@ impl Run {
         let included_files =
             self.template
                 .included_files(from, all_content_files, &self.options.variant);
-        let template_contents = self.read_all(included_files)?;
+        let template_contents = self.read_all(included_files, parser)?;
         let outputs = Self::to_output_paths(from, to, template_contents);
         let file_ops = outputs
             .into_iter()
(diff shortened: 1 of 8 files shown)


crates/templates/src/run.rs
```diff
@@ -405,6 +424,29 @@ impl Run {
             .build()
             .expect("can't fail due to no partials support")
     }
+
+    fn partials(&self) -> anyhow::Result<impl liquid::partials::PartialCompiler> {
+        let mut partials = Partials::empty();
+
+        if let Some(partials_dir) = self.template.partials_dir() {
+            for partial_file in std::fs::read_dir(partials_dir)? {
+                let partial_file = partial_file?;
```
rylev, line 433: nit:

```suggestion
                let partial_file = partial_file.context("could not read partial file")?;
```

crates/templates/src/run.rs
```diff
@@ -405,6 +424,29 @@ impl Run {
             .build()
             .expect("can't fail due to no partials support")
     }
+
+    fn partials(&self) -> anyhow::Result<impl liquid::partials::PartialCompiler> {
+        let mut partials = Partials::empty();
+
+        if let Some(partials_dir) = self.template.partials_dir() {
+            for partial_file in std::fs::read_dir(partials_dir)? {
```
rylev, line 432:
```suggestion
            for partial_file in std::fs::read_dir(partials_dir).context("could not read partials directory")? {
```

crates/templates/src/run.rs
```diff
@@ -405,6 +424,29 @@ impl Run {
...
+                let partial_file = partial_file?;
+                if !partial_file.file_type().is_ok_and(|t| t.is_file()) {
+                    anyhow::bail!("Non-file in partials directory: {partial_file:?}");
```
rylev, line 435: Why bail instead of just ignore?

crates/templates/src/run.rs
```diff
@@ -405,6 +424,29 @@ impl Run {
...
+                }
+                let partial_name = partial_file
+                    .file_name()
+                    .into_string()
+                    .map_err(|f| anyhow!("Unusable partial name {f:?}"))?;
```
rylev, line 440:
```suggestion
                    .context("Unusable partial name")?;
```

crates/templates/src/run.rs
```diff
@@ -19,6 +19,8 @@ use crate::{
     template::Template,
 };
 
+type Partials = liquid::partials::EagerCompiler<liquid::partials::InMemorySource>;
```
rylev, line 22:
```suggestion
/// An in-memory eager compiler of partials
type PartialsCompiler = liquid::partials::EagerCompiler<liquid::partials::InMemorySource>;
```

crates/templates/src/run.rs
itowlson: Because if the template author put non-files (e.g. directories) in the partials directory, then presumably they had some intent for those non-files, and we can't fulfil that intent.

crates/templates/src/run.rs
itowlson: Could you elaborate on why `context` is better here?  The error case for `OsString::into_string()` is the unstringable file name, not a rich error, so I feel like `context` would lead to something like:

```
...
   1. loading template
   2: i'm mad as heck
   3: /bish/bash/bosh/zootlewurdle.liquid
   4: unusable partial name
```

which feels less obvious than "unusable partial name /bish/bash/bosh/zootlewurdle.liquid".

(Although the error case is hopefully largely moot, because WHO THE HECK SHIPS THINGS WITH UNSTRINGABLE FILE NAMES ANYWAY and NO LANN THIS IS NOT A CHALLENGE.)

karthik2804 (approved):
This works great! I tested a JS template.

Thanks!

crates/templates/src/run.rs
rylev: Ah sorry - I blindly assumed that the error was a meaningful error type and not just the filename itself. Nevermind!

rylev (approved):

fibonacci1729 (approved):

PR #3112, opened 2025-04-15
Trying to track down integration test woes

itowlson:
main: got the failure
b70e3ccf5eaa64182e06308ff761b3619c39e36c: failed
e51d608de900162c9bb08733ca16474351ae1898: failed. What the heck.  Is GH sneakily rolling forward?  Is it a cache woe?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,172 +20,172 @@ env:
   RUST_VERSION: 1.81
 
 jobs:
-  dependency-review:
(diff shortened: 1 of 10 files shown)


itowlson:
Rebased to include Ryan's subsequent updates and still no joy.  Absolutely no idea what is going on now.  The evil seems to have reached back in time and be affecting old versions.  Trying it with extremely pruned tests in the hope of things happening less slowly...

PR #3114, opened 2025-04-15
Fix the failing integration tests

itowlson:


vdice (approved):
Out of curiousity, what aspect of the failures does the wit-bindgen bump fix?  Previous version led to empty response bodies?  (I'm glancing at https://github.com/spinframework/spin/actions/runs/14460156304/job/40551157474)

itowlson:
@vdice Far more exciting than empty response headers!  If you wade through the log spew, you'll find this unlovely trace:

```
thread '<unnamed>' panicked at /home/runner/work/spin/spin/tests/test-components/helper/src/lib.rs:19:5:
misaligned pointer dereference: address must be a multiple of 0x4 but is 0xff245
```

It seems like we somehow landed on a version of wit-bindgen with a pointer bug, and that somehow that didn't manifest itself in the initial PR (or possibly even a couple after, not sure).

rylev:
🎉

PR #3122, opened 2025-04-23
Chore(governance): Propose adding js-wasi-ext

karthik2804:
This commit adds the js-wasi-ext to the spin project. This is a library compatible with WASI 0.2 that makes select Node.js APIs available in ComponentizeJS and StarlingMonkey. This library depends on the build tooling used in the Spin JS SDK. This enables having more library compatibility for npm packages with the Spin JS SDK.

TODO:
- [ ] The link in the doc is to be the URL once the repository is moved here.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -12,6 +12,7 @@ The Spin project consists of several codebases with different release cycles. Th
     - [Spin Rust SDK](https://github.com/spinframework/spin-rust-sdk)
     - [Spin .NET SDK](https://github.com/spinframework/spin-dotnet-sdk)
     - [Spin Go SDK](https://github.com/spinframework/spin-go-sdk)
+    - [JS Wasi Ext](https://github.com/spinframework/js-wasi-ext)
 - Plugins:
     - [Spin Plugins Index Repository](https://github.com/spinframework/spin-plugins)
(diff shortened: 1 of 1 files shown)


vdice (approved):

michelleN (commented):
@itowlson brought up a good question in the spin project meeting around if it makes sense to host this project to the bytecode alliance. @karthik2804 do you have any thoughts around that?

tschneidereit:
This PR is a result of conversations Karthik and I had, so I can provide some context. There are two key reasons for hosting this here instead of in the BA:
1. It relies pretty heavily on Spin's JS SDK. That could be overcome, but would be a meaningful amount of work. If that were the only issue, that'd totally be worth it. However:
2. js-wasi-ext is very much a short-term hack, not the approach we'll want to take in the long-term. We've had various conversations in the context of StarlingMonkey and ComponentizeJS about how to design a Node.js compatibility layer, and will definitely want build one, but we know that it'll look quite different from this. And in the meantime, this approach isn't easy to integrate into either project, so it'd be unclear how to even go about it.

michelleN (approved):

michelleN:
@tschneidereit that makes sense. Thanks for the context!

karthik2804:
@michelleN Can this be merged and the repo moved over? Thanks

michelleN:
@karthik2804 Transfer complete ✅

PR #3131, opened 2025-05-05
fix(templates/http-go): enable GC on builds

adamreese:
The TinyGo GC seems to be working now for wasi.


itowlson:
Does this rely on latest-and-greatest TinyGo?

adamreese:
It does. This bit us when testing streaming and it was difficult to find the issue. I figured make it default

itowlson:
@adamreese _bump_

adamreese:
Tested on previous two versions and works as expected

PR #3135, opened 2025-05-13
Add support for limiting instance memory usage in spin up

### LINKED ISSUE #3132: Add support for limiting instance memory usage in `spin up`
Some environments - such as SpinKube - may limit memory available to an instance. We should let users apply this limit in `spin up` to help find memory usage problems during development.

This could be as simple as exposing `spin_core::StoreBuilder::max_memory_size` as an environment variable or could perhaps be configurable as part of runtime config.

fibonacci1729:
Implements a CLI flag and runtime config option for configuring the maximum memory allocation limit for an instance. The limit (implemented as a new `ExecutorHooks`) gets applied to the store via implementation of `ExecutorHooks::prepare_instance`.

Closes #3132

### DIFF
--- a/crates/runtime-factors/src/lib.rs
+++ b/crates/runtime-factors/src/lib.rs
@@ -100,6 +100,10 @@ pub struct TriggerAppArgs {
     /// To run from a file, prefix the filename with @ e.g. spin up --sqlite @migration.sql
     #[clap(long = "sqlite")]
     pub sqlite_statements: Vec<String>,
+
+    /// Sets the maxmimum memory allocation limit for an instance in bytes.
+    #[clap(long, env = "SPIN_MAX_INSTANCE_MEMORY")]
+    pub max_instance_memory: Option<usize>,
 }
 
 impl From<ResolvedRuntimeConfig<TriggerFactorsRuntimeConfig>> for TriggerFactorsRuntimeConfig {
--- a/crates/runtime-config/src/lib.rs
+++ b/crates/runtime-config/src/lib.rs
@@ -45,6 +45,8 @@ pub struct ResolvedRuntimeConfig<T> {
     ///
     /// `None` is used for an "unset" log directory.
     pub log_dir: Option<PathBuf>,
+    /// The maximum memory allocation limit.
+    pub max_instance_memory: Option<usize>,
     /// The input TOML, for informational summaries.
     pub toml: toml::Table,
 }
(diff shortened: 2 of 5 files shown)


crates/runtime-factors/src/lib.rs
```diff
@@ -100,6 +100,10 @@ pub struct TriggerAppArgs {
     /// To run from a file, prefix the filename with @ e.g. spin up --sqlite @migration.sql
     #[clap(long = "sqlite")]
     pub sqlite_statements: Vec<String>,
+
+    // Sets the maxmimum memory allocation limit for an instance.
+    #[clap(long = "max-memory-size", env = "MAX_MEMORY_SIZE")]
+    pub max_memory_size: Option<usize>,
```
lann, line 106: I wounder if this should be e.g. `max-instance-memory` to make it clear that its per-instance and not the whole server?

crates/runtime-factors/src/lib.rs
```diff
@@ -100,6 +100,10 @@ pub struct TriggerAppArgs {
     /// To run from a file, prefix the filename with @ e.g. spin up --sqlite @migration.sql
     #[clap(long = "sqlite")]
     pub sqlite_statements: Vec<String>,
+
+    // Sets the maxmimum memory allocation limit for an instance.
+    #[clap(long = "max-memory-size", env = "MAX_MEMORY_SIZE")]
```
lann, line 105: Clap kebab-cases the field name by default :shrug: 
```suggestion
    #[clap(long, env = "MAX_MEMORY_SIZE")]
```

lann, line 105: I think we prefix env vars elsewhere

```suggestion
    #[clap(long = "max-memory-size", env = "SPIN_MAX_MEMORY_SIZE")]
```

lann (approved):

crates/runtime-factors/src/lib.rs
```diff
@@ -100,6 +100,10 @@ pub struct TriggerAppArgs {
     /// To run from a file, prefix the filename with @ e.g. spin up --sqlite @migration.sql
     #[clap(long = "sqlite")]
     pub sqlite_statements: Vec<String>,
+
+    // Sets the maxmimum memory allocation limit for an instance.
```
itowlson, line 104: Units?

lann: memories

fibonacci1729: In the words of my high school chemistry teacher:

"What are we talking here? Jelly beans, kumquats, frankfurters?"

lann: Given that the actual resolution here is pages it wouldn't be unreasonable for this to be KiB, though no one ever got fired for picking good old bytes...

itowlson: to be clear, I'm not asking for us to do extra work parsing kumquat/frankfurter units, bytes is fine for this level of advancedness, I just reckon we ought to say

fibonacci1729: Updated to specify `in bytes`.

Issue #3088, opened 2025-04-03
Local Service Chaining blocks spin start


seungjin:
```
> spin --version
spin 3.3.0-pre0 (66fee279 2025-04-01)
```
```
> spin plugins list --installed
check-for-update 0.1.0 [installed]
cloud 0.11.0 [installed]
platform 0.1.0 [installed]
pluginify 0.7.0 [installed]
test 0.1 [installed]
trigger-command 0.2.2 [installed]
trigger-cron 0.1 [installed]
```

Getting this error at spin startup and can't start spin up. 
```
Error: failed to parse app lock file JSON

Caused by:
    invalid version 1 != 0 at line 2 column 24
Logging component stdio to ".spin/logs/"

Serving http://127.0.0.1:3000
Available Routes:
  root: http://127.0.0.1:3000
  assets: http://127.0.0.1:3000/static (wildcard)
A trigger exited unexpectedly. Terminating.
```
only when I add this into my spin.toml
```
[component.foo]
source = "wasm/foo.wasm"
allowed_outbound_hosts = ["http://random-place.spin.internal"]
```
Other normal urls are fine, just Spin's Local Service Chaining url emits error and block spin start.

rajatjindal:
I tried this in one of my app by just modifying the spin.toml and adding `allowed_outbound_hosts = ["http://random-place.spin.internal"]`, but it starts just fine for me.

```
spin --version
spin 3.3.0-pre0 (747c9fe 2025-04-02)
```

from spin.toml
```
[component.hello]
source = "main.wasm"
allowed_outbound_hosts = ["http://random-place.spin.internal"]
```

```
spin up       
Logging component stdio to ".spin/logs/"

Serving http://127.0.0.1:3000
Available Routes:
  hello: http://127.0.0.1:3000/hello
```

rajatjindal:
do you happen to have this app published somewhere I can try from?

seungjin:
Ok. I will get a generalized example of this issue.
My app is a mix of HTTP and cron triggers in a microservice architecture.
Mixing them probably caused this issue.
I will provide a more generalized case later tonight.

seungjin:
You can check this commit/repo/branch; https://github.com/seungjin/foo.fermyon.app/commit/9559ed0324db9073c0a4f39afa45397abfd18f95

itowlson:
```
[component.cron-me]
source = "wasm/cron_me.wasm"
[...]

PR #3136, opened 2025-05-14
Fix subsets inheriting service chaining requirement from whole app

### LINKED ISSUE #2915: Feature check is not properly done at trigger level
I have a spin application with 1 HTTP triggered component and 1 MQTT component. The HTTP component uses internal service chaining but not the MQTT one. Regardless, when i go to run the app, I get the error:
```
Error: This application requires the following features that are not available in this version of the 'mqtt' trigger: local_service_chaining A trigger exited unexpectedly. Terminating.
```

itowlson:
Fixes #2915 and #3088

This explores the approach of putting host requirements on _components_ instead of on the app, meaning that app splitting and multi-trigger Just Work without any shenanigans.

The approach is much nicer than the previous attempt in #3093, but needs a bit more testing.  The error behavious for existing trigger binaries is not lovely but that may not be possible to control since those binaries are already out there (and fixes itself when the triggers are rebuilt against these versions of the Spin crates).


itowlson:
This adds a must-understand entry to the lockfile.  Thinking about what this will affect:

CNCF projects:
- SpinKube: will likely need to update to work with OCIs pushed by versions that incorporate this
- Trigger plugins: (cron, sqs, mqtt, command): should update to work with revved Spin CLI, but the situations in which un-updated triggers fail will be ones in which they fail today (albeit they may now fail with worse error messages), so I don't think this risks a huge regression - and once updated they will work in situations where un-updated ones fail

Vendor projects:
- Fermyon (Cloud, FWF): services may take this update when suitable, but client plugins must not update until the services understand the new must-understand

I don't think this needs a lockfile version bump - the existing must-understand framework handles it, although the error messaging could do with improvement.  Open to being persuaded otherwise of course though!

As well as the new integration test, I tested by rebuilding the cron trigger against the new Spin crates, and it appeared to run correctly in apps that included service-chained HTTP.  Not sure about the practicalities of adding that to the integration tests though.

lann:
> The error behavior for existing trigger binaries is not lovely

> This adds a must-understand entry to the lockfile.

I wonder if we could avoid these through the power of Making It More Complicated:

- Add per-component service chaining host requirement
  - Any requirements added here are _also_ added to host requirements (!)
- Complexify `ensure_needs_only`:
  - Check component host requirements _first_
    - Add all of these to "all component host requirements" set
  - Check app host requirements, skipping any in "all component host requirements" set 
- Don't add `must_understand` `component_host_requirements`; they should be conservatively enforced by being added to host requirements

itowlson:
@lann I don't think that works without upgrading the triggers (which are where `ensure_needs_only` lives); and if we are allowed to upgrade the triggers then the problem goes away.

(I am also not sure it works anyway.  Or I am misunderstanding it.  Consider an app with HTTP and cron components.  If a HTTP component marks itself with a SC host requirement, that bubbles up to mark the application with a SC host requirement.  Then the cron trigger looks at the cron components, does not see a SC host requirement on any of them, so does not add SC to the "all component host reqs" bucket.  Then looks at application host reqs and finds "SC required" which has not been ticked off at the component level.  So bam.  But I feel like I must be mis-reading what you're proposing.)

lann:
> I don't think that works without upgrading the triggers (which are where ensure_needs_only lives); and if we are allowed to upgrade the triggers then the problem goes away.

I'm only trying to help with the "worse error message" part here, not the "need to upgrade triggers" part; I don't think it's any worse from that perspective? :thinking: 

> I am also not sure it works anyway.

Ah, I wasn't thinking carefully enough about app splitting, but I think it can be salvaged:

- Collect component host requirements from _all_ components in the app -> "all component host requirements 
- Check app host requirements, skipping any in "all component host requirements" set
- Check component host requirements for the components you actually care about

itowlson:
Okay, I think I get you now - thanks!  Let me try to echo it back to make sure.  The proposal is that enlightened triggers figure out which application host requirements are actually component host requirements that got bubbled up, and ignore them: they then validate on any _remaining_ app host reqs, and any component host reqs from components in their subset.  Whereas unenlightened triggers just see the (bubbled) application host requirement and validate against them.

This sounds like it will work, but I'm not sure it's worth the candle what I hope will be a transitional phase.  I suppose we could make component host reqs must-understand once most our "known" triggers have completed that transition, and then get rid of the complexity.

For context, the unlovely error from un-upgraded triggers is something like "unknown untagged enum variant."  It's unlovely but I'm not sure how much effort to invest in getting rid of it.  The real fix for this is to make must-understands into strings instead of an enum - again, once that flows through to triggers this will all become a non-issue.

itowlson:
(to be clear, thank you for brainstorming ideas and I'm happy to go ahead with this if you think it's worth the extra complexity)

lann:
I don't mind not doing this; it would definitely throw another flapjack on our towering stack of complexity.

lann (approved):

PR #3137, opened 2025-05-15
ci(.github): add fossa.yml

vdice:
Adds a workflow to run FOSSA scans for this project (ref https://github.com/orgs/spinframework/projects/2/views/1?pane=issue&itemId=95403677)

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,29 @@
+name: fossa
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - v*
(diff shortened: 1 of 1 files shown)


vdice:
Are we okay with only triggering the scan on the `main` (and `v*`) branches and not on PRs from forks?  The FOSSA_API_KEY org secret isn't accessible for PRs from forks.

vdice:
Superseded by PR from origin as opposed to my fork, so we can test the workflow: https://github.com/spinframework/spin/pull/3138

PR #3143, opened 2025-05-22
Warn or error on dubious-looking file mount

itowlson:
**Motivation**

A user had a problem recently where they got a cryptic "cannot create parent directory" error while trying to `spin up` or `spin deploy`.  It turned out that they had tried to use the current directory string `"."` in a component `files` list: this caused us to generate a dubious temp directory string which the filesystem refused to create.

The user had also tried to use the `*` pattern, which successfully copied their current directory, but this was not what they really wanted or expected.

**Proposed behaviour**

This PR attempts to catch dubious-looking file mounts:

* `"."` (current directory), `".."` (parent directory), and any patterns that end in those segments are hard errors.  The error message guides users to use `{ source, destination }` (directory placement) format instead, or a `**/*` pattern.
* `"*"` (all files in current directory) and `"**/*"` (all files in or under current directory) now get a warning.  These aren't errors, we can carry out the instructions, but they are likely to be incorrect - most applications do not seek to include their own `spin.toml` or `.gitignore` files.  So these now warn (and offer equivalent, more explicit, patterns that we accept).

In both cases, the message also links to the documentation on including files with components.

**Concerns**

First off, this is a quite specific response to a rather specific issue, and I am not sure if it is an overreaction.  It may be that this is better handled by keeping the code simple and instead including the information in troubleshooting guides.

Second, this introduces hardwired output to stderr along the loader path.  We do have some precedent for this (e.g. the "Loading Wasm components is taking a few seconds" message) but I wanted to highlight it because other programs (e.g. deployment plugins) link to the loader.  An alternative approach is to make the warning cases errors on the basis that they are almost certain to be not what the user really wants and we are better off forcing them to address it than letting it ride.

All that for four lines!  Oh well, at least now y'all know.

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -485,6 +485,20 @@ impl LocalLoader {
         dest_root: &Path,
         exclude_files: &[String],
     ) -> Result<()> {
+        if glob_or_path == ".." || glob_or_path.ends_with("/..") {
+            bail!("A file pattern can't end in a parent directory path (..)\nIf you want to include a directory, use source-destination form, or a glob pattern ending in **/*.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components");
+        }
+        if glob_or_path == "." || glob_or_path.ends_with("/.") {
+            bail!("A file pattern can't end in a current directory path (.)\nIf you want to include a directory, use source-destination form, or a glob pattern ending in **/*.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components");
+        }
+
+        if glob_or_path == "*" {
+            terminal::warn!("A component is including the entire application directory as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
+        }
+        if glob_or_path == "**/*" {
+            terminal::warn!("A component is including the entire application directory tree as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./**/*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
+        }
+
         let path = self.app_root.join(glob_or_path);
         if path.exists() {
             let dest = dest_root.join(glob_or_path);
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -493,10 +493,10 @@ impl LocalLoader {
         }
 
         if glob_or_path == "*" {
-            terminal::warn!("A component is including the entire application directory as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
+            tracing::warn!(alert_in_dev = true, "A component is including the entire application directory as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
         }
         if glob_or_path == "**/*" {
-            terminal::warn!("A component is including the entire application directory tree as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./**/*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
+            tracing::warn!(alert_in_dev = true, "A component is including the entire application directory tree as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./**/*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
         }
 
         let path = self.app_root.join(glob_or_path);
--- a/crates/telemetry/src/alert_in_dev.rs
+++ b/crates/telemetry/src/alert_in_dev.rs
@@ -1,0 +1,41 @@
+//! Provides a way for warnings to be force-raised in the
+//! `spin up` environment even if RUST_LOG is not set to warn.
+//! This is useful for things that are not errors but where we
+//! want application developers to know they have a problem.
+
+use tracing::{Event, Subscriber};
+use tracing_subscriber::{filter::filter_fn, registry::LookupSpan, Layer};
+
+const ALERT_IN_DEV_TAG: &str = "alert_in_dev";
+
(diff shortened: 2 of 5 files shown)


crates/loader/src/local.rs
```diff
@@ -485,6 +485,20 @@ impl LocalLoader {
...
+        if glob_or_path == "." || glob_or_path.ends_with("/.") {
+            bail!("A file pattern can't end in a current directory path (.)\nIf you want to include a directory, use source-destination form, or a glob pattern ending in **/*.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components");
+        }
+
+        if glob_or_path == "*" {
+            terminal::warn!("A component is including the entire application directory as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
+        }
+        if glob_or_path == "**/*" {
+            terminal::warn!("A component is including the entire application directory tree as asset files. This is unlikely to be what you want.\nIf this is what you want, use the pattern \"./**/*\" to avoid this warning.\nLearn more: https://spinframework.dev/writing-apps#including-files-with-components\n");
+        }
```
lann, line 500: I wonder if these warnings should be moved to somewhere more `spin up`-specific. I believe the containerd shim shares this code.

lann: Here perhaps? https://github.com/spinframework/spin/blob/72500b233fabfaafa41e34a0a61eee6892079d86/src/commands/up.rs#L490-L497

lann: Thinking down the rabbit hole, an alternative would be to integrate `tracing` with `terminal` so that we can emit warnings anywhere we like (annotated with some appropriate field like `show_this_to_the_user_please = true` and translate those into `terminal` output. That's a lot of churn to maintain conceptual purity; what do you think?

itowlson: Should anyone doubt my zealotry for conceptual purity, I have in the past done a "state monad" in JavaScript just to have a way to pass warnings back to the correct level.  I will not be saying anything further about the 4-year gap in my resume.

I guess the tracing approach is do-able, but it's then not clear to me how a host (shim or whatever) chooses what warnings to present or not present - or is the idea that "in `spin up`, warnings are off by default, except for ones labelled as 'actually this one is not off by default'"?  This sounds like it could be more broadly useful anyway.

lann: > "in `spin up`, warnings are off by default, except for ones labelled as 'actually this one is not off by default'"?

Yes, something like that. See also https://github.com/spinframework/spin/issues/3144#issuecomment-2901213109

crates/telemetry/src/alert_in_dev.rs
```diff
@@ -0,0 +1,47 @@
...
+    event.record(&mut extractor);
+    extractor.message
+}
+
+#[derive(Default)]
+struct MessageExtractor {
+    message: Option<String>,
+}
+
+impl tracing::field::Visit for MessageExtractor {
```
itowlson, line 41: I am sure this can't be the correct way to get a value out of an event but I couldn't find a direct way.  If someone would care to educate me on the right way to approach this then I will be forever grateful _(terms and conditions apply)_

itowlson: Oh wait. Maybe the visitor is meant to terminal::warn the thing directly.  Yeah that probably makes more sense

crates/telemetry/src/alert_in_dev.rs
```diff
@@ -0,0 +1,41 @@
...
+
+fn warn(event: &Event<'_>) {
+    let mut visitor = PrintMessageAsWarning;
+    event.record(&mut visitor);
+}
+
+struct PrintMessageAsWarning;
+
+impl tracing::field::Visit for PrintMessageAsWarning {
+    fn record_debug(&mut self, field: &tracing::field::Field, value: &dyn std::fmt::Debug) {
```
lann, line 36: You may want to implement `format_str` as well (nearly identically) which would allow printing without the quotes/escaping.

itowlson: Heh, I tried that first but, weirdly, it never got hit.  I had to implement `record_debug` anyway, and `record_debug` doesn't print with quotes or escaping, so I went with it:

```
$ spin up
Warning: A component is including the entire application directory as asset files. This is unlikely to be what you want.
If this is what you want, use the pattern "./*" to avoid this warning.
Learn more: https://spinframework.dev/writing-apps#including-files-with-components
```

I can dig in more if you want though.

lann: Nope I think that's quite enough tracing for one day.

crates/telemetry/src/alert_in_dev.rs
```diff
@@ -0,0 +1,41 @@
...
+pub struct CommandLineAlertingLayer;
+
+impl<S: Subscriber> Layer<S> for CommandLineAlertingLayer {
+    fn on_event(&self, event: &Event<'_>, _ctx: tracing_subscriber::layer::Context<'_, S>) {
+        let meta = event.metadata().fields();
+        if meta.field(ALERT_IN_DEV_TAG).is_some() {
+            warn(event);
+        }
+    }
+}
```
lann, line 26: I think tracing "wants" you to do conditional logging with a [layer filter](https://docs.rs/tracing-subscriber/0.3.19/tracing_subscriber/layer/index.html#per-layer-filtering), something like:

```suggestion
    CommandLineAlertingLayer
    	.with_filter(filter_fn(|meta| {
    	    meta.field(ALERT_IN_DEV_TAG).is_some()
        }))

}

pub struct CommandLineAlertingLayer;

impl<S: Subscriber> Layer<S> for CommandLineAlertingLayer {
    fn on_event(&self, event: &Event<'_>, _ctx: tracing_subscriber::layer::Context<'_, S>) {
        warn(event);
    }
}
```

The difference is that the bare layer will - by default - register "interest" in _all_ events, preventing tracing from short-circuiting processing of events that no subscriber cares about.

lann (approved):
Well I went down yet another tracing rabbit hole. I think the point is technically correct but perhaps unlikely to make any practical difference in context so feel free to ignore it.

itowlson:
There is a quirk where if the user sets `RUST_LOG=warn` then the warning is printed twice, once using terminal and once by default subscriber.  I'm not going to tackle this because 1. it's a rare case 2. it's pretty harmless and 3. fix your `spin.toml`

PR #3146, opened 2025-05-27
outbound-networking: Add error.type to tracing::error!

lann:
This allows distinguishing between these very different kinds of errors in downstream error handling systems.

### DIFF
--- a/crates/factor-outbound-networking/src/lib.rs
+++ b/crates/factor-outbound-networking/src/lib.rs
@@ -92,8 +92,18 @@ impl Factor for OutboundNetworkingFactor {
             .expression_resolver()
             .clone();
         let allowed_hosts_future = async move {
-            let prepared = resolver.prepare().await?;
-            AllowedHostsConfig::parse(&hosts, &prepared)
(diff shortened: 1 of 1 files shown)


crates/factor-outbound-networking/src/lib.rs
```diff
@@ -92,8 +92,18 @@ impl Factor for OutboundNetworkingFactor {
             .expression_resolver()
             .clone();
         let allowed_hosts_future = async move {
-            let prepared = resolver.prepare().await?;
-            AllowedHostsConfig::parse(&hosts, &prepared)
+            let prepared = resolver.prepare().await.inspect_err(|err| {
+                tracing::error!(
```
itowlson, line 96: Does this result in a doubled error message at the console - once for the `tracing::error!` and once when the `?` bubbles back to the top and causes an error exit?

lann: It should do whatever it did before apart from the wording change on the `parse` path.

itowlson (approved):
Tried to reproduce my double-messaging worry and it didn't do it so :shipit:

PR #3150, opened 2025-06-02
Add runtime config for blocking outbound connections by CIDR

lann:
Example:

```toml
[outbound_network]
# 'private' is a keyword that blocks IPs that aren't "globally-routable"
block_networks = ["1.2.3.4/16", "private"]
```


crates/factor-outbound-networking/src/allowed_hosts.rs
```diff
@@ -209,7 +209,7 @@ impl HostConfig {
             bail!("host lists are not yet supported")
         }
 
-        if let Ok(net) = host.parse::<ipnet::IpNet>() {
+        if let Ok(net) = ip_network::IpNetwork::from_str_truncate(host) {
```
rylev, line 212: Is there any difference in behavior here? If there is, that could be a breaking change.

crates/factor-outbound-networking/src/blocked_networks.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use ip_network::IpNetwork;
+use ip_network_table::IpNetworkTable;
+
+/// A cheaply-clonable set of blocked networks
+#[derive(Clone, Default)]
+pub struct BlockedNetworks {
+    /// A set of IP networks to be blocked
+    networks: Arc<IpNetworkTable<()>>,
+    /// If true, block all non-globally-routable networks, in addition to `networks`
```
rylev, line 14: Should we link to the is_global method docs for an explanation of what "globally routable" means?

crates/factor-outbound-networking/src/runtime_config.rs
```diff
@@ -1,141 +1,37 @@
...
-}
-
-// Maps host authority -> ClientConfig
-type HostClientConfigs = Arc<HashMap<String, Arc<ClientConfig>>>;
-
-impl RuntimeConfig {
-    /// Returns runtime config with the given list of [`TlsConfig`]s. The first
-    /// [`TlsConfig`] to match an outgoing request (based on
-    /// [`TlsConfig::components`] and [`TlsConfig::hosts`]) will be used.
-    pub fn new(tls_configs: impl IntoIterator<Item = TlsConfig>) -> anyhow::Result<Self> {
```
rylev, line 26: I'm having a hard time following the changes here. We used to default to `RuntimeConfig::new([])` and now we're defaulting to the `Default` impl. Those seem pretty fundamentally different from each other. Are we sure that's correct?

crates/factor-outbound-networking/src/lib.rs
```diff
@@ -65,15 +69,19 @@ impl Factor for OutboundNetworkingFactor {
...
-        let runtime_config = match ctx.take_runtime_config() {
-            Some(cfg) => cfg,
-            // The default RuntimeConfig provides default TLS client configs
-            None => RuntimeConfig::new([])?,
-        };
+        let RuntimeConfig {
+            client_tls_configs,
+            blocked_ip_networks: block_networks,
+            block_private_networks,
+        } = ctx.take_runtime_config().unwrap_or_default();
```
rylev, line 76: As mentioned below, I'm unsure of the use of `unwrap_or_default` here is correct when previously it was defaulting to `RuntimeConfig::new([])`.

crates/factor-outbound-networking/src/runtime_config/spin.rs
```diff
@@ -41,21 +45,62 @@ impl SpinTlsRuntimeConfig {
         &self,
         table: &impl GetTomlValue,
     ) -> anyhow::Result<Option<super::RuntimeConfig>> {
-        let Some(tls_configs) = self.tls_configs_from_table(table)? else {
+        let maybe_blocked_networks = self
+            .blocked_networks_from_table(table)
+            .context("parsing blocked_networks")?;
```
rylev: Typically the error message would be something like:

```suggestion
            .context("failed to parse `blocked_networks` table")?;
```

The terse "parsing blocked_networks" will be hard to understand outside the scope of this code.

Most of the errors in this file could use a bit more context to help users understand what's happening.

rylev (approved):
The implementation mostly looks good - just one major question I left in a comment. Overall, I'm super on board with the internal changes.

I'm more unsure about the exact UX of how we're extending the runtime config yaml. I don't have specific feedback, and so I'm not going to block. My FUD mainly extends from the fact that the yaml changes are permanent and must be kept into purpiutiy. I think I'd personally have felt much more comfortable if the internal changes were landed separately from the YAML changes. I understand why they're bundled, but they don't have to be, and I think I'd like to be a bit more careful about introducing changes that directly impact users.

crates/factor-outbound-networking/src/allowed_hosts.rs
lann: Shouldn't be; `ip_network` requires "normal" cidrs by default (e.g. `1.2.0.0/16` rather than `1.2.3.4/16`) so it has this separate `from_str_truncate` to deal with the latter. In this context I think it is fine to accept either.

crates/factor-outbound-networking/src/runtime_config.rs
lann: This is some of that "not-entirely-related refactoring" I apologize for. :grimacing: 

The processing that used to happen here in `RuntimeConfig::new` moved to [`TlsClientConfigs::new`](https://github.com/spinframework/spin/pull/3150/files/94ecf44398143877b06e1f0e98b97c3baf006904#diff-42cd340729c178822197a16c7b08373bb21a04118ad475c417d1c174959bca24R17). This makes `RuntimeConfig` hold just the input config rather than also acting as the fully-processed instance state.

As a result, `RunmtimeConfig::default()` is actually sensible now.

crates/factor-outbound-networking/src/lib.rs
lann: Addressed below: https://github.com/spinframework/spin/pull/3150/files/94ecf44398143877b06e1f0e98b97c3baf006904#r2126982055

crates/factor-outbound-networking/src/runtime_config/spin.rs
lann: Ah yes I've been writing too much Go `fmt.Errorf("parsing thing: %w")` :sweat_smile:

lann: Better but still leaves something to be desired:

```
failed to parse [outbound_networking] table

Caused by:
    invalid value: string "foo", expected an IP network in CIDR notation or the keyword 'private'
    in `block_networks`
```

lann:
> I'm more unsure about the exact UX of how we're extending the runtime config yaml.

The only user-visible changes _should_ be the new `[outbound_networking]` / `blocked_networks` config. If that is what you are unsure about then we should definitely discuss before merging.

crates/factor-outbound-networking/src/blocked_networks.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+        if self.networks.longest_match(ip_addr).is_some() {
+            return true;
+        }
+        // Convert IPv4-compatible IPv6 addresses to IPv4 and check again to prevent bypass
+        if let IpAddr::V6(ipv6) = ip_addr {
+            if let Some(ipv4_compat) = ipv6.to_ipv4() {
+                return self.is_blocked(&IpAddr::V4(ipv4_compat));
+            }
+        }
```
lann, line 58: I noticed this issue when adding the link to [`ip_network::Ipv6Network::is_global`](https://docs.rs/ip_network/0.4.1/ip_network/struct.Ipv6Network.html#method.is_global) to the docs above; annoyingly, its behavior is subtly different from the unstable [`std::net::Ipv6Addr::is_global`](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#method.is_global) though that does lead to an arguably better solution here.

rylev:
> The only user-visible changes should be the new [outbound_networking] / blocked_networks config. If that is what you are unsure about then we should definitely discuss before merging.

Indeed. My concern is more of a vibe than anything (which is why I've approved this PR and don't want to block). I don't have concrete suggestions - I'm simply unsure about the API and unsure about introducing such additions to the Spin API in a PR. For example, some random questions I have:
* Should we expose this as blocking networks (as you've done here) or allowing networks (mirroring the API we expose in spin.toml)? 
  * I think I agree with your choice of blocking vs. allowing (as typically operators care more about explicitly restricting rather than explicitly allowing).
* Is the `private` keyword, the right way to block the use of all non-globally-routable networks?
  * It seems fine so 🤷 

I guess in general, I'm wondering if the user facing changes should be introduced through a SIP instead of a PR. It's a small feature so I understand why that wasn't done initially, but it feels like anything that impacts the public API of Spin should be discussed in a SIP.

lann:
> * Should we expose this as blocking networks (as you've done here) or allowing networks (mirroring the API we expose in spin.toml)?
>       
>       * I think I agree with your choice of blocking vs. allowing (as typically operators care more about explicitly restricting rather than explicitly allowing).

I thought about this a little. It seems somewhat unlikely that an operator that wanted to lock down the network _that_ much wouldn't also have complete control over the apps (and therefor `allowed_outbound_hosts`), but if there is demand for that it should be quite easy to add an `allow_hosts` field which is either mutually exclusive with `block_hosts` or does something slightly more fancy to merge the two.

lann:
>     * Is the `private` keyword, the right way to block the use of all non-globally-routable networks?
>       
>       * It seems fine so 🤷

You _could_ translate `private` into a large set of explicit CIDRs. It's really just a convenience for what (to me) seems like the most likely use-case for this feature: "don't let apps access services behind the firewall".

lann:
> I guess in general, I'm wondering if the user facing changes should be introduced through a SIP instead of a PR.

First off, this is definitely a reasonable question to ask.

> it feels like anything that impacts the public API of Spin should be discussed in a SIP.

Depending on your interpretation, "anything that impacts the public API" could apply to _a lot_ of things that we haven't covered with SIPs before, but we can certainly adjust our attitudes / policies / guidance on that if we want. I guess it didn't even occur to me to do a SIP here in part because we didn't do one for the even larger [`[[client_tls]]` feature](https://github.com/spinframework/spin/pull/2596) it lives right next to. :shrug: 

More broadly:

SIPs are [documented](https://github.com/spinframework/spin/blob/main/docs/content/sips/index.md?plain=1#L9) as being for "major work items" which...leaves a _little_ room for interpretation :sweat_smile: but I think reflects that _one_ of the goals is to avoid putting in "too much" work into an implementation before getting feedback on the design. This is important both to avoid wasting contributor's time _and_ to avoid putting too much pressure on reviewers to accept a design just because there is a lot of code already written for it. On the flip side, there is overhead in the SIP process and every extra bit of friction incrementally raises the bar for contribution - even for maintainers!

rylev:
@lann it occurs to me that perhaps one reason I'm nervous about this is that runtime-config.yaml isn't currently versioned. Perhaps if it were, I'd be less worried?

lann:
I prefer to think of it as "version 0"; you always get that one freebie before you have to start writing down explicit schema versions. :slightly_smiling_face:

PR #3152, opened 2025-06-05
Provide more guidance on invalid `allowed_outbound_hosts`

itowlson:
Fixes https://github.com/fermyon/feedback/issues/61

I added a "learn more" to the output:

```
# allowed_outbound_hosts = ["localhost"]

$ spin up

Caused by:
   2: Failed to load component `outbound-bad-wildcard`
   3: `allowed_outbound_hosts` is malformed
   4: "localhost" does not contain a scheme (e.g., 'http://' or '*://')
Learn more: https://spinframework.dev/v3/http-outbound#granting-http-permissions-to-components
```

and special-cased some things that people might guess for wildcard syntax:

```
# allowed_outbound_hosts = ["*"]

$ spin up

Caused by:
   2: Failed to load component `outbound-bad-wildcard`
   3: `allowed_outbound_hosts` is malformed
   4: "*" is not an allowed outbound host format.
Hosts must be in the form <scheme>://<host>[:<port>], with '*' wildcards allowed for each.
If you intended to allow all outbound networking, you can use '*://*:*' - this will obviate all network sandboxing.
Learn more: https://spinframework.dev/v3/http-outbound#granting-http-permissions-to-components
```


fibonacci1729 (approved):

rylev:
@itowlson drive-by thought - should we alias `*` to `*://*:*`?

lann:
> @itowlson drive-by thought - should we alias `*` to `*://*:*`?

That was suggested in [the issue](https://github.com/fermyon/feedback/issues/61#issuecomment-2945479914). [My reply](https://github.com/fermyon/feedback/issues/61#issuecomment-2946084707):

> There are two different "allow anythings": `https://*:*` (allow http to anything) and `*://*:*` allow _any_ protocol to anything (via `wasi:sockets`).
> 
> Also, we're trying to nudge people to take advantage of Spin's better-than-status-quo sandboxing; getting someone to stop and think about whether they really need `*` is part of that.
> [...]

Issue #2580, opened 2024-06-18
Add version control support to `spin new` or spin CLI 


me-diru:
I wondered if adding a flag or an interactive prompt (a few ways suggested by @calebschoepp) asking for a Git repository initialization might help people get going from the start with their Spin application and track their code from the beginning.

Maybe we can take their `github/gitlab-username` in spin.toml

calebschoepp:
I imagine it could be something like how `cargo new` works where there is a [`--vcs` flag](https://doc.rust-lang.org/cargo/commands/cargo-new.html#:~:text=the%20directory%20name.-,%2D%2Dvcs%20vcs,-Initialize%20a%20new) and a default. Likely we would start with options `none` or `git`. We can of course bikeshed the flag name and what the default is.

calebschoepp:
> Maybe we can take their github/gitlab-username in spin.toml

@me-diru Mind elaborating on what you mean by this?

itowlson:
As a heads up we already have a `--no-vcs` flag, which I think suppresses `.gitignore`.  Having a `--vcs` and `--no-vcs` which affect different things might be a bit regrettable.

me-diru:
> > Maybe we can take their github/gitlab-username in spin.toml
> 
> @me-diru Mind elaborating on what you mean by this?

I thought we could input from the users in the top-level information tin `spin.toml` o automate and set repository information in .`git` folder. However, I think `--vcs` flag that you shared covers this case as well.

itowlson:
Are you suggesting we set up git config, or set a remote?  I'd be extremely wary of that.  Running `git init` for them is about as far as I'd want to go.

calebschoepp:
+1 to be wary about setting up a remote. I think all we should do is `git init`.

me-diru:
> As a heads up we already have a `--no-vcs` flag, which I think suppresses `.gitignore`. Having a `--vcs` and `--no-vcs` which affect different things might be a bit regrettable.

thanks for pointing this out! Indeed, since we are generating a `.gitignore` with the spin new by default, maybe adding a `git init` is a good start? 

Both will be ignored with `--no-vcs` flag

me-diru:
> Are you suggesting we set up git config, or set a remote? I'd be extremely wary of that. Running `git init` for them is about as far as I'd want to go.



> +1 to be wary about setting up a remote. I think all we should do is `git init`.

Agree  :D `git init` should be enough

calebschoepp:
[...]

PR #3155, opened 2025-06-12
feat: Add automatic git repository initialization to `spin new` command

iamrajiv:
Fixes: #2580

FYI:

- Detects existing Git repositories to avoid creating nested repos
- Respects the --no-vcs flag to skip Git initialization
- Applies only to new applications, not to component additions

### DIFF
--- a/crates/templates/src/run.rs
+++ b/crates/templates/src/run.rs
@@ -474,7 +474,7 @@ impl Run {
         }
     }
 
-    // Function to check if the target directory is already in a git repositoriees
+    // Function to check if the target directory is already in a git repositories
     async fn is_already_in_git_repo(&self, dir: &std::path::Path) -> bool {
         use tokio::process::Command;
 


crates/templates/src/run.rs
```diff
@@ -429,29 +431,79 @@ impl Run {
...
-
-        if let Some(partials_dir) = self.template.partials_dir() {
-            let partials_dir = std::fs::read_dir(partials_dir)
-                .context("Error opening template partials directory")?;
-            for partial_file in partials_dir {
-                let partial_file =
-                    partial_file.context("Error scanning template partials directory")?;
-                if !partial_file.file_type().is_ok_and(|t| t.is_file()) {
-                    anyhow::bail!("Non-file in partials directory: {partial_file:?}");
+        match self.template.partials_dir() {
```
itowlson, line 434: It's not clear to me why partials support needs to change as part of this.  Should this be a separate PR or is it needed here?

crates/templates/src/run.rs
```diff
@@ -429,29 +431,79 @@ impl Run {
...
+        let mut git = Command::new("git");
+        git.arg("-C")
+            .arg(dir)
+            .arg("init");
+
+        git.output()
+            .await
+            .understand_git_result()
+            .map(|_| ())
+            .map_err(|e| anyhow::anyhow!("{}", e))
```
itowlson, line 506: This seems like it might be equivalent to `git.output().await.understand_git_result()?;` - not sure but those maps seem a bit odd.

crates/templates/src/run.rs
```diff
@@ -429,29 +431,79 @@ impl Run {
...
+        }
+
+        let target_dir = self.generation_target_dir();
+
+        if self.is_already_in_git_repo(&target_dir).await {
+            return;
+        }
+
+        if let Err(e) = self.init_git_repo(&target_dir).await {
+            eprintln!("Warning: Failed to initialize a git repository for you: {}", e);
```
itowlson, line 473: Use `terminal::warn!`

itowlson (commented):
Thanks for this Rajiv!  It basically looks good but I did have a few minor questions and nits.  Hopefully all easy to address (and they may be my lack of understanding).

crates/templates/src/run.rs
```diff
@@ -19,9 +19,6 @@ use crate::{
     template::Template,
 };
 
-/// A set of partials to be included in a Liquid template.
-type PartialsBuilder = liquid::partials::EagerCompiler<liquid::partials::InMemorySource>;
```
itowlson, line 23: Removing this is making the compile fail - I am not sure how you were even able to test it in the current state!  Whatever the intent, I suggest we back out the partials changes from this PR - I cannot imagine any reason they'd need to change for git init support.

itowlson (changes requested):

itowlson:
Incorporated into #3197

PR #3158, opened 2025-06-13
Don't mark outbound http spans as errors.

rylev:
We log errors in all the cases that are legit (non-user) related errors and so there's no need to also mark the span as errored. This prevents user related issues (e.g., timeouts) from not being bubbled up to telemetry as errors.

Additionally, we stop renaming the span after the http method which isn't very helpful.

### DIFF
--- a/crates/factor-outbound-http/src/spin.rs
+++ b/crates/factor-outbound-http/src/spin.rs
@@ -3,12 +3,12 @@ use spin_world::v1::{
     http as spin_http,
     http_types::{self, HttpError, Method, Request, Response},
 };
-use tracing::{field::Empty, instrument, Level, Span};
+use tracing::{field::Empty, instrument, Span};
 
 use crate::intercept::InterceptOutcome;
 
 impl spin_http::Host for crate::InstanceState {
-    #[instrument(name = "spin_outbound_http.send_request", skip_all, err(level = Level::INFO),
+    #[instrument(name = "spin_outbound_http.send_request", skip_all,
         fields(otel.kind = "client", url.full = Empty, http.request.method = Empty,
         http.response.status_code = Empty, otel.name = Empty, server.address = Empty, server.port = Empty))]
     async fn send_request(&mut self, req: Request) -> Result<Response, HttpError> {
@@ -116,6 +116,8 @@ fn record_request_fields(span: &Span, req: &Request) {
         Method::Head => "HEAD",
         Method::Options => "OPTIONS",
     };
+    // Set otel.name to just the method name to fit with OpenTelemetry conventions
+    // <https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name>
     span.record("otel.name", method)
         .record("http.request.method", method)
         .record("url.full", req.uri.clone());


crates/factor-outbound-http/src/spin.rs
```diff
@@ -116,8 +116,7 @@ fn record_request_fields(span: &Span, req: &Request) {
         Method::Head => "HEAD",
         Method::Options => "OPTIONS",
     };
-    span.record("otel.name", method)
```
lann, line 119: I think this is the only way to set the otel span name _dynamically_, as the tracing span name must be static: https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/index.html#special-fields

lann: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name

rylev: Ah that's what you were going for... I don't love it, but I can revert and at least leave a comment that we're trying to be spec compliant here.

rylev: Done!

lann: Its not just about the spec, it makes a difference in some o11y providers like Datadog iirc.

crates/factor-outbound-http/src/spin.rs
```diff
@@ -116,6 +116,8 @@ fn record_request_fields(span: &Span, req: &Request) {
         Method::Head => "HEAD",
         Method::Options => "OPTIONS",
     };
+    // Set otel.name to just the method name to fit with OpenTelemetry conventions
+    // <https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name>
```
lann, line 120: :+1:

crates/factor-outbound-http/src/spin.rs
```diff
@@ -3,12 +3,12 @@ use spin_world::v1::{
     http as spin_http,
     http_types::{self, HttpError, Method, Request, Response},
 };
-use tracing::{field::Empty, instrument, Level, Span};
+use tracing::{field::Empty, instrument, Span};
 
 use crate::intercept::InterceptOutcome;
 
 impl spin_http::Host for crate::InstanceState {
-    #[instrument(name = "spin_outbound_http.send_request", skip_all, err(level = Level::INFO),
```
lann, line 11: I can never _quite_ remember how all of these things translate into various outputs but I think the `name` override here helps to distinguish this from the wasi `send_request`.

...except that the wasi `send_request` has the same name as this one (oops! :sweat_smile:) https://github.com/spinframework/spin/blob/db7fd312f99f52238f41545dd582527043ae77f6/crates/factor-outbound-http/src/wasi.rs#L74

rylev: The reason I removed this is because it seems to be overwritten completely with the write to `otel.name` (at least when viewing in HoneyComb - I haven't checked in DataDog).

lann: Yes, it is likely completely replaced for any otel exporter but would still show up in the `tracing_subscriber::fmt` output(s) and be used by `RUST_LOG` filters. :shrug:

lann (approved):

PR #3163, opened 2025-06-17
fix(tests): Re-enable MQTT and Postgres tests

PhantomInTheWire:


itowlson:
We had a pass - I kicked off a re-run to see if we can get a flake

PhantomInTheWire:
@itowlson I've squashed all necessary changes. For context, all previous runs of all-integration-tests were executed on ubuntu-latest, which allowed them to run on my fork's PRs. However, I've now switched it back to the custom Spin runner.

itowlson:
Excellent, thanks!  Sorry for the slow turmarounds here, it's making me manually approve each re-run, but I will try to stay across it.

PhantomInTheWire:
Everything is in place and all the tests pass, waiting for your review @itowlson

itowlson:
Thank you @PhantomInTheWire!  I'm planning to re-run the CI a few times to see if the flakes re-appear, so please don't worry if things go back to 'running.'  I'll take a look at the changes in the meantime.

tests/runtime-tests/tests/outbound-postgres-variable-permission/spin.toml
```diff
@@ -14,5 +14,5 @@ component = "test"
 
 [component.test]
 source = "%{source=outbound-postgres}"
-allowed_outbound_hosts = ["postgres://{{ pg_host }}:%{port=5432}"]
-environment = { DB_URL = "postgres://postgres:postgres@localhost:%{port=5432}/spin_dev" }
+allowed_outbound_hosts = ["postgres://{{ pg_host }}:5432"]
```
itowlson, line 17: Are the changes away from placeholder ports needed?  I believe we did this so that test instances could be brought up on ports determined at runtime, and it would be good to keep it.  If it's not working, it would be good to understand why. (But if that's involved in the flakiness, we can probably punt on it for now.)  cc @rylev for thoughts and insight

.github/workflows/build.yml
```diff
@@ -191,6 +191,20 @@ jobs:
     # run on a larger runner for more SSD/resource access
     runs-on: ubuntu-22.04-4core-spin
     if: ${{ github.repository_owner == 'spinframework' }}
+    services:
+      postgres:
```
itowlson, line 195: Hmm, I thought we were dockerising the Postgres instance, but maybe that's been part of the problem.  Again, I'd like to get Ryan's input on making it part of the CI image instead (I'm not for or against it, I just want to hear from someone who knows this bit of the system better than I do!).

itowlson (approved):
Thanks again for looking into this!  It all looks good to me but there's a couple of things I'd like to undestand, and I really want to get feedback from @rylev as well.  But fingers crossed that you've cracked it...!

itowlson:
Bother, that's a flake in MQTT.  I saw you updated the settings for Postgres; do you think something similar needs to be done for MQTT?

.github/workflows/build.yml
PhantomInTheWire: yeah i think it was not dockerising it right in ci environments, i think it's the same problem with mqtt too, hence the hardcoded service

PhantomInTheWire:
> Bother, that's a flake in MQTT. I saw you updated the settings for Postgres; do you think something similar needs to be done for MQTT?

yeah this is my first time encountering it I think adding a mqtt service to build.yml might fix it but let me try and see if i can figure it out

PhantomInTheWire:
Update: I attempted to reproduce the flake on a fork using the ubuntu-latest runner ([link](https://github.com/PhantomInTheWire/spin/actions/runs/15758246560/attempts/7?pr=4)), running 20 jobs concurrently across multiple attempts but couldn’t trigger the issue(it did fail once on run # 2 but that was integration_tests::bad_build_test not mqtt).

Could you share how the custom Spin runner is configured? I think that could be a reason

tests/runtime-tests/tests/outbound-postgres-variable-permission/spin.toml
PhantomInTheWire: leaving the placeholder ports in place causes a runtime failure. Specifically, the test fails with:

```
postgres::Connection::open(&address) errored: 'Error::ConnectionFailed("error communicating with the server: Connection reset by peer (os error 54)")'
```

This happens sometimes unless a fixed port is provided, the dynamic port allocation logic isn’t working reliably in this test/ci context. We could dig deeper into why, but for now, hardcoding the port makes the test pass deterministically.

itowlson:
Hmm, I don't think I have access to the custom large runner configuration.  @spinframework/org-admin can one of you folks help out here?

.github/workflows/build.yml
rylev: It would be nice to be able to use docker because that's the portable option. We want to be able to run tests anywhere (as long as docker is available). That was the original intent around using docker so that we didn't have to assume anything was available on the host machine (besides docker and python).

If we move CI to use host based services, we lose the ability to verify that things are working correctly.

tests/runtime-tests/tests/outbound-postgres-variable-permission/spin.toml
rylev: It makes the test pass in CI, but it will break tests running locally :-)

We should figure out why things are not working in CI.

vdice:
@PhantomInTheWire @itowlson sorry for the delay.  Here's the configuration for the `ubuntu-22.04-4core-spin` runner:

```
Platform: Linux x64
Size: 4-cores · 16 GB RAM · 150 GB SSD
Public IP: Disabled
Network Configuration: Disabled
Image source: GitHub-owned
Image name: Ubuntu 22.04
```

As seen, it is a GitHub-provided runner, so that high-level config is all we've really specified.

PR #3166, opened 2025-06-18
Remove wasm-pkg-common as dependency

### LINKED ISSUE #3165: Remove the `wasm_pkg_common` dependency from `spin-manifest` (or make it optional)
The `spin-manifest` crate takes on `wasm_pkg_common` as a dependency for two simple definitions:
* `PackageRef`
* `Registry`

Unfortunately `wasm_pkg_common` is a very heavy weight dependency bringing in for example `tokio`. This is an issue for `spin-test` which needs to compile `spin-manifest` to wasm and cannot, for example, bringing `tokio` in.

rylev:
Fixes #3165

I'm not sure if this is the way we'd like to address this issue. It does involve copying a few definitions from `wasm-pkg-common`, but perhaps that's easiest?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5162,6 +5162,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "mio"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "78bed444cc8a2160f01cbcf811ef18cac863ad68ae8ca62092e8db51d51c761c"
+dependencies = [
+ "libc",
+ "wasi 0.11.0+wasi-snapshot-preview1",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
 name = "monostate"
 version = "0.1.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


lann:
`wasm-pkg-common` has a `tokio` feature that could maybe be turned off?

lann:
It looks like it hasn't been tested and is probably broken but that can probably be resolved relatively easily.

itowlson (approved):

fibonacci1729 (approved):

rylev:
For `spin-test` the real need is that `spin-manifest` can be compiled to the `wasm32-wasip1` target. Unfortunately, `wasm-pkg-common` without tokio still cannot be compiled to wasm because of a dependency on the `home` crate which does not support wasi. I've opened an issue on the `wasm-pkg-common` repo: https://github.com/bytecodealliance/wasm-pkg-tools/issues/162

lann:
[`wasm-pkg-common` 0.11.0 is released](https://crates.io/crates/wasm-pkg-common/0.11.0) and [should build on wasm32-wasip1](https://github.com/bytecodealliance/wasm-pkg-tools/blob/ac85d149d7247c563f143d63309dceceb38930cd/.github/workflows/ci.yml#L43).

rylev:
Thanks @lann!! Closing in favor of https://github.com/spinframework/spin/pull/3171

PR #3167, opened 2025-06-18
Move outbound networking config to new crate

rylev:
Configuration for outbound networking is used in a lot of places outside of the outbound networking factor (e.g., in `spin-test). Instead of requiring pulling in the entire factor (with its heavy dependencies), we allow functionality related to the config to be pulled in separately.


lann (approved):
Assuming no sneaky refactoring ...

itowlson:
YESSSSSSSSSSSSSSSSS

PR #3169, opened 2025-06-19
Move router to its own separate crate

rylev:
Same idea as #3167 - this allows the router to be used in wasm (e.g., in `spin-test`).


crates/routes/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "spin-routes"
```
lann: 🤷
```suggestion
name = "spin-http-routes"
```

lann (approved):

PR #3173, opened 2025-06-23
Warn if `spin deploy` rather than `spin cloud deploy`

itowlson:
Spin provides `deploy` and `login` commands.  These are an inheritance from the original development at Fermyon, and act as shortcuts to the equivalent commands in Fermyon's `cloud` plugin.

As part of vendor neutrality, we want to shift away from the assumption of a default vendor plugin, but we need to respect backward compatibility at minor releases, and signpost breaking changes well before major releases.

This PR proposes a first step on that transition.  With this PR, the `spin deploy` and `spin login` commands:

* Continue to call their equivalents in the Fermyon `cloud` plugin.  (Backward compatibility.)
* Allow the defaulting to `cloud` to be overridden via the new `SPIN_DEPLOY_PLUGIN` environment variable.  (Provide a pathway for decoupling.)
* Warn if the `SPIN_DEPLOY_PLUGIN` environment variable is not set.  (Signpost future breaking change.)

A future major release will be able to remove the backcompat default (first bullet) and to hard error if the EV is not set (third bullet).

**Alternatives**

We could remove the proposed environment variable, and tell people that in future they will have to run plugin commands explicitly (`spin whatever deploy`).  This has the advantages that: 1. less code; 2. commands are unambiguous and do not depend on the environment.  I quite like the `spin deploy` shortcut which is why I've proposed the EV as a way to keep using it, but if other folks prefer to just get rid of the shortcuts, I'm totally okay with that.

**Behaviour**

With no environment variable:

```
$ spin deploy
Warning: `spin deploy` will soon need to told which deployment plugin to use.
Run a plugin command (e.g. `spin cloud deploy`), or set the `SPIN_DEPLOY_PLUGIN` environment variable, instead.
Defaulting to `cloud` plugin.

# ... normal `spin cloud deploy` output ...
```

Still with no environment variable:

```
$ spin cloud deploy

# ... normal `spin cloud deploy` output ...
```

With environment variable (normally if someone wanted this they'd set it in `win.ini` or whatever the kids use nowadays but you get what I mean):

```
$ SPIN_DEPLOY_PLUGIN=cloud spin deploy

# ... normal `spin cloud deploy` output ...
```

### DIFF
--- a/src/commands/cloud.rs
+++ b/src/commands/cloud.rs
@@ -4,7 +4,7 @@ use clap::Args;
 
 #[derive(Debug, Args, PartialEq)]
 #[clap(
-    about = "Package and upload an application to the Fermyon Cloud.",
+    about = "Package and upload an application to a deployment environment.",
     allow_hyphen_values = true,
     disable_help_flag = true
 )]
@@ -16,7 +16,7 @@ pub struct DeployCommand {
 
 #[derive(Debug, Args, PartialEq)]
 #[clap(
-    about = "Log into the Fermyon Cloud.",
+    about = "Log into a deployment environment.",
     allow_hyphen_values = true,
     disable_help_flag = true
 )]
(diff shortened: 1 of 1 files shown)


src/commands/cloud.rs
```diff
@@ -26,18 +26,43 @@ pub struct LoginCommand {
...
         cmd.append(&mut self.args.clone());
         execute_external_subcommand(cmd, app).await
     }
 }
+
+fn deployment_plugin(cmd: &str) -> anyhow::Result<String> {
+    match std::env::var(DEPLOY_PLUGIN_ENV) {
+        Ok(v) => Ok(v),
+        Err(std::env::VarError::NotPresent) => {
+            terminal::warn!("`spin {cmd}` will soon need to told which deployment plugin to use.\nRun a plugin command (e.g. `spin {DEFAULT_DEPLOY_PLUGIN} {cmd}`), or set the `{DEPLOY_PLUGIN_ENV}` environment variable, instead.\nDefaulting to `{DEFAULT_DEPLOY_PLUGIN}` plugin.");
```
kate-goldenring:
```suggestion
            terminal::warn!("`spin {cmd}` will soon need to be told which deployment plugin to use.\nRun a plugin command (e.g. `spin {DEFAULT_DEPLOY_PLUGIN} {cmd}`), or set the `{DEPLOY_PLUGIN_ENV}` environment variable, instead.\nDefaulting to `{DEFAULT_DEPLOY_PLUGIN}` plugin.");
```

kate-goldenring (approved):
@itowlson I really like this approach of using an env var for now and agree with the phase out approach. Thank you for laying it out and the motivation behind it.

PR #3174, opened 2025-06-23
Enable HTTP2 for outgoing requests

fibonacci1729:
~This is draft for now (waiting for the release of `Wasmtime v35.0.0`) but nevertheless putting this up to get 👀 and early feedback.~

This PR enables outgoing http2 connections in the outbound HTTP factor. ~The upgrade to HTTP2 happens if the `TE` header is present and equal to `trailers` in the outgoing request headers.~ This makes it possible to support gRPC clients in Spin components.

UPDATE: This now uses ALPN negotiation to decide if HTTP2 is used. Additionally, I added an environment variable `SPIN_OUTBOUND_H2C_PRIOR_KNOWLEDGE` for convenience when testing against non-tls http2 servers. 

Now that we're not relying on the presence of `te` to signal upgrade to HTTP2 this PR is no longer dependent upon wasmtime v35.


crates/factor-outbound-http/src/wasi.rs
```diff
@@ -242,6 +243,11 @@ async fn send_request_handler(
         return Err(ErrorCode::DestinationIpProhibited);
     }
 
+    let is_http2 = request
+        .headers()
+        .get(http::header::TE)
+        .is_some_and(|te| te == "trailers");
```
itowlson: I believe the `TE` header might contain multiple values, e.g. the example `TE: trailers, deflate;q=0.5` from the MDN page.  Is that allowed?  If so, will it be picked up?

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -297,14 +319,25 @@ async fn send_request_handler(
...
         let tcp_stream = TokioIo::new(tcp_stream);
-        let (sender, conn) = timeout(
-            connect_timeout,
-            // TODO: we should plumb the builder through the http context, and use it here
-            hyper::client::conn::http1::handshake(tcp_stream),
-        )
-        .await
-        .map_err(|_| ErrorCode::ConnectionTimeout)?
-        .map_err(hyper_request_error)?;
+        let (sender, conn) = if is_http2 {
```
itowlson, line 338: Is it feasible to deduplicate this now rather large and complicated block of code with the one around line 289-307?

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -317,19 +350,21 @@ async fn send_request_handler(
...
-                .path_and_query()
-                .map(|p| p.as_str())
-                .unwrap_or("/"),
-        )
-        .build()
-        .expect("comes from valid request");
+    if !is_http2 {
+        // at this point, the request contains the scheme and the authority, but
+        // the http packet should only include those if addressing a proxy, so
+        // remove them here, since SendRequest::send_request does not do it for us
```
itowlson, line 372: This is my own ignorance but if we change the URI to remove the authority then how does it know where to send the request?

itowlson (commented):
I'm not qualified to review all the HTTP shenanigans but ~I didn't let that stop me~ I did have a couple of comments/questions.

crates/factor-outbound-http/src/wasi.rs
lann: You're going to love this!

https://httpwg.org/specs/rfc9113.html#ConnectionSpecific

> The only exception to this is the TE header field, which MAY be present in an HTTP/2 request; when it is, it MUST NOT contain any value other than "trailers".

In practice, HTTP/1.1 clients "almost never" set `TE: trailers` and HTTP/2 gRPC clients _must_ set `TE: trailers`. If someone wants to send an HTTP/1.1 request with `trailers` they have a workaround: `TE: chunked, trailers`.

crates/factor-outbound-http/src/wasi.rs
lann: This is annoying coupled to wasmtime-wasi-http which has earlier copied the authority into the `host` header: https://github.com/bytecodealliance/wasmtime/blob/3632683b38011cd12839c57cbbc4fa03e834b111/crates/wasi-http/src/http_impl.rs#L71

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -366,3 +438,16 @@ fn dns_error(rcode: String, info_code: u16) -> ErrorCode {
         info_code: Some(info_code),
     })
 }
+
+#[derive(Clone, Copy, Debug)]
+struct TokioExec;
```
lann: It looks like this crate already transitively depends on `hyper-util`; could probably use https://docs.rs/hyper-util/0.1.14/hyper_util/rt/tokio/struct.TokioExecutor.html?

fibonacci1729: ah yes! I couldn't find where that was; i'll update.

crates/factor-outbound-http/src/wasi.rs
fibonacci1729: Going to pop this in a comment on the check so future passerbys don't need to re-research this.

tschneidereit:
The code looks good to me, though for an actual review I'd defer to Lann and others.

A more general question: would it be viable to default to http2 instead of only opting in for `TE: trailers`? Or are there substantial downsides to doing so?

dicej:
> A more general question: would it be viable to default to http2 instead of only opting in for `TE: trailers`? Or are there substantial downsides to doing so?

From https://w3techs.com/technologies/details/ce-http2:

> HTTP/2 is used by 33.3% of all the websites.

Another option would be to start with HTTP/1 and request an upgrade to HTTP/2; not sure if that would require an extra round-trip, though.  Could also send both an HTTP/1 request and a HTTP/2 request concurrently and let them race, but that means double the network traffic.  Not sure what the state of the art is for this kind of thing.

lann:
> A more general question: would it be viable to default to http2 instead of only opting in for `TE: trailers`? Or are there substantial downsides to doing so?

You know I was about to complain that hyper deprecated its support for http1+http2 clients but then I actually stopped and thought about what that entails and realized I was overthinking things:

- Set `["h2", "http/1.1"]` for ALPN unconditionally
- Immediately after [the TLS connect step](https://github.com/spinframework/spin/blob/bf339fcf4754964f675ff28dd74123facb295db8/crates/factor-outbound-http/src/wasi.rs#L282-L285) you should have access to a `&ClientConnection` via [`TlsStream::get_ref`](https://docs.rs/tokio-rustls/0.26.2/tokio_rustls/client/struct.TlsStream.html#method.get_ref)
- Check the [`ClientConnection::alpn_protocol`](https://docs.rs/rustls/0.23.22/rustls/client/struct.ClientConnection.html#method.alpn_protocol); if it is `h2` use http2, otherwise use http1

tschneidereit:
@lann that was my expectation: that we can open up protocol negotiation and use h2 where possible, instead of defaulting to only trying to negotiate h1.1. If that works, then I think it'd be good for performance, too

lann:
What should happen if `TE: trailers` is set and http1 is negotiated? I guess that could be an error but maybe a warning would be just as useful?

tschneidereit:
I think a warning is fine for this: ISTM that if you're setting that header, you're doing so because you know you're talking to something that supports GRPC

dicej:
@lann Thanks for explaining how HTTP/1+HTTP/2 connections are established in the real world.  I forgot it happened at the TLS layer.  I guess that means we'd stick with plain HTTP/1 for `http://` URLs.

lann:
For debugging/test purposes it might be nice to have an env var like `SPIN_OUTBOUND_H2C_PRIOR_KNOWLEDGE="<host>:<port>"` where if it matches the authority exactly you get plaintext http/2.

Cargo.toml
```diff
@@ -187,3 +187,9 @@ blocks_in_conditions = "allow"
 [[bin]]
 name = "spin"
 path = "src/bin/spin.rs"
+
+# Remove this once Wasmtime 35.0.0 is released
```
dicej: Given that this is only needed for `te` header support, could we separate this part out into a different PR?

Cargo.toml
```diff
@@ -187,3 +187,9 @@ blocks_in_conditions = "allow"
...
+[patch.crates-io]
+wasmtime = { git = "https://github.com/fibonacci1729/wasmtime", branch = "allow-te" }
+wasmtime-wasi = { git = "https://github.com/fibonacci1729/wasmtime", branch = "allow-te" }
+wasmtime-wasi-http = { git = "https://github.com/fibonacci1729/wasmtime", branch = "allow-te" }
```
lann: This needs to be reverted.

dicej (commented):
LGTM overall; thanks for doing this!

I agree with @itowlson that the duplication in `send_request_handler` should be addressed to ease the maintenance burden, if practical.

PR #3175, opened 2025-06-24
PostgreSQL UUID and JSONB support

### LINKED ISSUE #1551: uuid data type on spin_sdk::pg threw an exception
Problems: create any columns in Postgres with `uuid` data type:

On code:

```rs
let order_id = uuid::Uuid::new_v4().clone();
let order_id_param = ParameterValue::Binary(order_id.as_ref()); 
```

if not using Binary then I don't know how can we assign `uuid` to what kind of data type which spin supported at https://github.com/fermyon/spin/blob/main/crates/outbound-pg/src/lib.rs#L145

Then save it using `spin_sdk::pg`

```rs
let order_nrow_executed = pg::execute(&address, insert_order_sql, &insert_order_params)?;
```

It compiled successfully, but when runs

```
Handler returned an error: Postgres error: Error { kind: ToSql(0), cause: Some(WrongType { postgres: Uuid, rust: "alloc::vec::Vec<u8>" }) }
```

My question is there is any way to work around or handle this case?

### LINKED ISSUE #2936: Feature request: Support for JSONB in postgresql
Hi there,

can you please give me an advice how to store JSONB (binary) values into postgresql?

## Table
```sql
create table tb_test
(
  "value" jsonb,
);
```

## What I tried 
```rust
let params = vec![
  ParameterValue::Binary(json_string.into_bytes()),
];
connection.execute("INSERT INTO tb_test VALUES ($1)", &params);
```

error output:
```sh
Error::QueryFailed("Error { kind: ToSql(1), cause: Some(WrongType { postgres: Jsonb, rust: \"alloc::vec::Vec<u8>\" }) }")
```

also as string does not work
```rust
let params = vec![
  ParameterValue::Str(json_string.clone()),
];
connection.execute("INSERT INTO tb_test VALUES ($1)", &params);
```

error output:
```sh
fail: Error::QueryFailed("Error { kind: ToSql(1), cause: Some(WrongType { postgres: Jsonb, rust: \"alloc::string::String\" }) }")
```

## environment
- spin: 3.0.0
- postgresql: 17
- rust

### LINKED ISSUE #2938: Example: Provide a postgresql example with error handling
Hi,

I got stuck to figure out how to get detailed postgresql error message.

function to store data into postgres:
```rust
fn store_new_thing(thing: String) -> Result<()> {
    let db_connection = variables::get("db_connection")?;
    let connection = Connection::open(&db_connection)?;

    let params = vec![ParameterValue::Str(thing.clone())];
    connection.execute(SQL_CREATE_NEW_THING, &params)?;

    Ok(())
}
````

usage 
```rust
match store_new_thing(serialized) {
    Ok(_) => println!("store new thing succeed"),
    Err(error) => panic!("Failed to save new thing: {:?}", error),
}
```

I got this output but I want to check the reason. So how can I access for instance the `severity` or `message` or `routine`...
`println!("{}", error.causes())` returns None and the function is deprecated.

```
{
  kind: Db,
  cause: Some(DbError {
    severity: \"ERROR\",
    parsed_severity: Some(Error),
    code: SqlState(E23505),
    message: \"duplicate key value violates unique constraint \\\"new_things_pkey\\\"\",
    detail: Some(\"Key (id)=(...) already exists.\"), 
    hint: None,
    position: None,
    where_: Some(\"SQL function \\\"fn_insert_new\\\" statement 1\"),
    schema: Some(\"public\"),
    table: Some(\"new_things\"),
    column: None, datatype: None, constraint: Some(\"tb_new_things_pkey\"),
    file: Some(\"nbtinsert.c\"),
    line: Some(673),
    routine: Some(\"_bt_check_unique\")
  })
}
```

* Spin version 3.0.0

itowlson:
Fixes #1551.  Fixes #2936.  Fixes #2938.


wit/deps/spin-postgres@4.0.0/postgres.wit
```diff
@@ -0,0 +1,130 @@
...
+      boolean(bool),
+      int8(s8),
+      int16(s16),
+      int32(s32),
+      int64(s64),
+      floating32(f32),
+      floating64(f64),
+      str(string),
+      binary(list<u8>),
+      date(tuple<s32, u8, u8>), // (year, month, day)
```
lann, line 71: A tuple that needs a comment to explain its elements sounds like it wants to be a record . 🙂

itowlson: Yes: unfortunately, this was a mistake I made in v3 and I don't see value in changing it for v4 (yes, I know I am allowed to).

itowlson:
This PR now also adds the following types:

* NUMERIC.  Currently I have this represented at the WIT level as a `string`, which I realise will be controversial.  The reason for suggesting this was that the decimal representation is highly portable and unambiguous.  After looking at how Go and Rust decimal libraries construct their types, I was concerned that a more sensible mantissa-scale approach would likely be error prone.  But I'm very open to better ideas (heck, I'm open to worse ones if you can think of any).
* INTERVAL
* Arrays: INT4[], INT8[], NUMERIC[], TEXT[].  Only single level supported, and these are done with `option` lists to allow for NULLs.  (It is technically inconsistent to use option-none rather than the DbNull value, but WIT doesn't allow for recursive variants, so here we are.  Again, open to better ideas.)
* Ranges: INT4RANGE, INT8RANGE, NUMRANGE
* UUID and JSONB as mentioned in the title

cc @tharpooljha

There is still significant tidying to do: the new types make the Postgres code more complicated and easier to miss cases, and I had to add some helper types DIDN'T I THE ORPHAN RULE so it's pretty messy right now.  I'll work on cleaning it up but anyway here it is so far.

wit/deps/spin-postgres@4.0.0/postgres.wit
```diff
@@ -0,0 +1,145 @@
...
+      range-int32(tuple<option<tuple<s32, range-bound-kind>>, option<tuple<s32, range-bound-kind>>>),
+      range-int64(tuple<option<tuple<s64, range-bound-kind>>, option<tuple<s64, range-bound-kind>>>),
+      range-decimal(tuple<option<tuple<string, range-bound-kind>>, option<tuple<string, range-bound-kind>>>),
+      array-int32(list<option<s32>>),
+      array-int64(list<option<s64>>),
+      array-decimal(list<option<string>>),
+      array-str(list<option<string>>),
+      interval(interval),
+      db-null,
+      unsupported,
```
lann: Would it make sense to have a e.g. `raw(list<u8>)`?

itowlson: Could you say more about what you're envisaging?

lann: Many postgres drivers provide some mechanism to extend their builtin types by converting to/from raw postgres values: 
https://docs.rs/tokio-postgres/latest/tokio_postgres/types/trait.FromSql.html
https://pkg.go.dev/github.com/jackc/pgx/v4#hdr-Raw_Bytes_Mapping
https://www.psycopg.org/docs/advanced.html#type-casting-from-sql-to-python

lann: I guess `raw` would effectively be "other".

itowlson: Gotcha.  Intercepting that may require us a bit more FromSql magic, I am not sure - we access values via `get`, which expects to return into a FromSql implementation that matches the column type, but maybe it will let us return anything into a `Vec<u8>`, or otherwise I can create a fake FromSql that accepts all types and just captures the raw bytes.  I will take a look!

lann: Another thought: maybe this "other" variant could be a _resource_ that additionally exposes type information about the value, like https://docs.rs/tokio-postgres/latest/tokio_postgres/types/struct.Type.html#method.oid

itowlson: Your lightest whim is my command (subject to applicable law and terms of service).  I provisionally put the type name rather than the OID but if you have opinions then bring 'em on, I am utterly naive about this stuff.

lann: I don't really know how these are used in practice either, I'm just trying to save you from spending the rest of your days adding support for an neverending list of increasingly obscure data types. 😅

itowlson: Indeed - this has me thinking "what if I rewrote the whole interface to just tunnel EVERYTHING as blobs and let the guest language use whatever decoder is idiomatic" so your plan to free me from this rabbit hole may have backfired...

(Fear not.  I do not intend to do this as part of this PR.  Nor without a _lot_ more research into the factoring of guest language DB libraries.)

itowlson:
@rylev How would you advise committing tests for this?  From conversation it sounds like they should go in `conformance-tests`; can I do that before landing this in Spin?

rylev:
@itowlson indeed - adding this to the conformance tests seems like the right way to go. Consumers of conformance tests are in charge of digesting the version of the tests that they want so it should be fine to add new tests whenever.

itowlson:
I've taken this opportunity to change the query error type to be structured when structured information is available (#2938).  I've kept the old textual error available via the `as-text` field in order to facilitate porting existing `pg3` code; also note that not all provider errors are "DB errors" so we retain an "I can only give you the error text" variant as well.  Sorry if this is too fiddly.

Cargo.toml
```diff
@@ -124,7 +124,7 @@ base64 = "0.22"
 bytes = "1"
 chrono = "0.4"
 clap = "3.2"
-conformance-tests = { git = "https://github.com/fermyon/conformance-tests", rev = "ecd22a45bcc5c775a56c67689a89aa4005866ac0" }
+conformance-tests = { git = "https://github.com/itowlson/conformance-tests", rev = "fa529a698616116beb024881cfd6fdeda7f39415" }
```
itowlson, line 127: Once conformance test changes are merged, and a new `tests.tar.gz` generated, we can reset to the main fork; at that point, Postgres tests should run.  (They can't run without a merge because they use test definitions from the  `tests.tar.gz` of the conformance-tests repo canary release.)

fibonacci1729 (approved):

PR #3176, opened 2025-06-26
Selective error spans in variables and key-value

rylev:
We currently set spans as errored in variables and key-value on any error. This includes errors that are purely caused by bad app input - for example key-value's `Error::NoSuchStore`. 

## Open Question

The current behavior is arguably correct in the Spin CLI local developer setting where the operator of Spin and the author of the app that it's running are probably the same. However, in other setups like SpinKube, this assumption breaks down - the author of the app running and the operator of Spin are probably not the same.

We could consider making this configurable where Spin CLI will continue to mark _all_ errors as errored spans, but other uses of these factors could change to only mark spans as errored when its clearer that it was not caused by app behavior.

### DIFF
--- a/crates/factor-key-value/src/host.rs
+++ b/crates/factor-key-value/src/host.rs
@@ -2,7 +2,7 @@ use super::{Cas, SwapError};
 use anyhow::{Context, Result};
 use spin_core::{async_trait, wasmtime::component::Resource};
 use spin_resource_table::Table;
-use spin_telemetry::OpenTelemetrySpanExt as _;
+use spin_telemetry::traces::{self, Fault};
 use spin_world::v2::key_value;
 use spin_world::wasi::keyvalue as wasi_keyvalue;
 use std::{collections::HashSet, sync::Arc};
@@ -72,10 +72,7 @@ impl KeyValueDispatch {
     pub fn get_store<T: 'static>(&self, store: Resource<T>) -> anyhow::Result<&Arc<dyn Store>> {
         let res = self.stores.get(store.rep()).context("invalid store");
         if let Err(err) = &res {
-            let current_span = tracing::Span::current();
-            current_span.set_status(spin_telemetry::opentelemetry::trace::Status::error(
-                err.to_string(),
-            ));
+            traces::mark_as_error(err, Some(Fault::Host));
         }
         res
     }
@@ -190,12 +187,11 @@ impl key_value::HostStore for KeyValueDispatch {
 
 /// Make sure that infrastructure related errors are tracked in the current span.
 fn track_error_on_span(err: Error) -> Error {
-    if let Error::Other(_) | Error::StoreTableFull = &err {
-        let current_span = tracing::Span::current();
-        current_span.set_status(spin_telemetry::opentelemetry::trace::Status::error(
-            err.to_string(),
-        ));
-    }
+    let fault = match err {
+        Error::NoSuchStore | Error::AccessDenied => Fault::Guest,
+        Error::StoreTableFull | Error::Other(_) => Fault::Host,
+    };
+    traces::mark_as_error(&err, Some(fault));
     err
 }
 
--- a/crates/factor-key-value/src/host.rs
+++ b/crates/factor-key-value/src/host.rs
@@ -2,10 +2,11 @@ use super::{Cas, SwapError};
 use anyhow::{Context, Result};
 use spin_core::{async_trait, wasmtime::component::Resource};
 use spin_resource_table::Table;
+use spin_telemetry::OpenTelemetrySpanExt as _;
 use spin_world::v2::key_value;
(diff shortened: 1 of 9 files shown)


crates/telemetry/src/traces.rs
```diff
@@ -61,3 +62,32 @@ pub(crate) fn otel_tracing_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
...
+        }
+    }
+}
+
+/// Marks the current span as an error with the given error message and optional fault type.
+pub fn mark_as_error<E: std::fmt::Display>(err: &E, fault: Option<Fault>) {
+    let current_span = tracing::Span::current();
+    current_span.set_status(opentelemetry::trace::Status::error(err.to_string()));
+    if let Some(fault) = fault {
+        current_span.set_attribute("error.fault", fault.as_str());
```
lann: Is there a difference between "`error.fault` not set" and "`error.fault = 'host'`"?

crates/factor-key-value/src/host.rs
```diff
@@ -190,6 +205,7 @@ fn to_wasi_err(e: Error) -> wasi_keyvalue::store::Error {
 }
 
 impl wasi_keyvalue::store::Host for KeyValueDispatch {
+    #[instrument(name = "wasi_key_value.open", skip(self, identifier), fields(otel.kind = "client"))]
```
lann:
```suggestion
    #[instrument(name = "wasi_key_value.open", skip_all, fields(otel.kind = "client"))]
```

hack/o11y-stack/docker-compose.yaml
```diff
@@ -1,4 +1,3 @@
-version: "3"
```
lann, line 1: Intentional?

crates/factor-key-value/src/host.rs
```diff
@@ -69,7 +70,11 @@ impl KeyValueDispatch {
     }
 
     pub fn get_store<T: 'static>(&self, store: Resource<T>) -> anyhow::Result<&Arc<dyn Store>> {
-        self.stores.get(store.rep()).context("invalid store")
+        let res = self.stores.get(store.rep()).context("invalid store");
+        if let Err(err) = &res {
+            traces::mark_as_error(err, Some(Fault::Host));
+        }
+        res
```
lann: Setting `Fault::Host` unconditionally seems like it shouldn't be necessary (see other comment about host vs empty).

crates/telemetry/src/traces.rs
rylev: I'm not sure - for the use case I have in mind, not really, but I generally prefer being explicit when possible.

hack/o11y-stack/docker-compose.yaml
rylev: Yea, there was a warning from docker compose that this field is deprecated and should be removed.

crates/factor-key-value/src/host.rs
lann: Separately, if we're going to be doing this repeatedly it might be nice to have a `ResultExt` with some helper methods like `.mark_as_host_error()` (`.blame_host()`?) :shrug:

crates/telemetry/src/traces.rs
lann: I guess it means "we thought about it and we blame the host" vs "we haven't thought about it".

rylev: Talked a bit about this offline with Lann. I'd like to treat not being set as "the question wasn't even asked". If we determine that this always gets interpreted the same as "host", we could potentially remove the host fault.

crates/factor-key-value/src/host.rs
```diff
@@ -69,7 +70,11 @@ impl KeyValueDispatch {
     }
 
     pub fn get_store<T: 'static>(&self, store: Resource<T>) -> anyhow::Result<&Arc<dyn Store>> {
-        self.stores.get(store.rep()).context("invalid store")
+        let res = self.stores.get(store.rep()).context("invalid store");
+        if let Err(err) = &res {
+            traces::mark_as_error(err, Some(Blame::Host));
+        }
+        res
```
lann, line 77: Just FYI: I recently realized that [`Result::inspect_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err) has been stable for "long enough" now :slightly_smiling_face:

crates/factor-variables/src/host.rs
```diff
@@ -20,6 +21,7 @@ impl variables::Host for InstanceState {
 }
 
 impl v1::config::Host for InstanceState {
+    #[instrument(name = "spin_config.get", skip(self), fields(otel.kind = "client"))]
```
lann, line 24: Out of scope: we probably shouldn't be tracing keys...

lann (approved):

PR #3179, opened 2025-06-30
Fuller and more explicit docs in manifest schema

itowlson:
Examples:

image

image

Unfortunately, Taplo has a bug (https://github.com/tamasfe/taplo/issues/788) which means that for array fields like these, the hover only works if the array is empty.  Which rather compromises the usefulness!  Should I push ahead adding docs like these to more fields all the same?

### DIFF
--- a/crates/manifest/src/schema/common.rs
+++ b/crates/manifest/src/schema/common.rs
@@ -5,43 +5,96 @@ use serde::{Deserialize, Serialize};
 
 use wasm_pkg_common::{package::PackageRef, registry::Registry};
 
-/// Variable definition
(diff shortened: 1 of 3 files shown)


rylev:
This seems useful even if the IDE help doesn't display properly.

itowlson:
I HAVE DEFEATED TAPLO, although I expect to see a bit of "sadly shaking head while murmuring 'but at what cost'" in the reviews.

image

lann (approved):
> "Your scientists were so preoccupied with whether they could, they didn’t stop to think if they should."
\- Albert Einstein, probably

PR #3180, opened 2025-06-30
Add schema directive to template manifests

itowlson:
As of Spin 3.3 we include a `spin.toml` JSON schema as part of each release.  This PR proposes linking that schema from template manifests: this provides (a measure of) validation and code completion out of the box (well, if the user has Even Better TOML or its equivalent installed).

image

In this PR, I link to the manifest via a versioned release URL.  We get this for free, but it does mean we need to update the templates whenever we update the schema doc (or on every release, if we want to avoid folks going "I'm on Spin 3.5, why did you give me this OLD schema"), and that users need to rev the link to get schema updates.  If we wanted to invest a little more in the plumbing, we could upload schemas to a version-neutral location so users could automatically inherit improvements (or could still lock to a version via a GH URL).  Thoughts?

### DIFF
--- a/templates/http-c/content/spin.toml
+++ b/templates/http-c/content/spin.toml
@@ -1,3 +1,5 @@
+#:schema https://schemas.spinframework.dev/spin/manifest-v2/latest.json
+
 spin_manifest_version = 2
 
 [application]
--- a/templates/http-go/content/spin.toml
+++ b/templates/http-go/content/spin.toml
@@ -1,3 +1,5 @@
+#:schema https://schemas.spinframework.dev/spin/manifest-v2/latest.json
+
 spin_manifest_version = 2
 
 [application]
--- a/templates/http-grain/content/spin.toml
+++ b/templates/http-grain/content/spin.toml
@@ -1,3 +1,5 @@
+#:schema https://schemas.spinframework.dev/spin/manifest-v2/latest.json
+
 spin_manifest_version = 2
 
 [application]
--- a/templates/http-php/content/spin.toml
+++ b/templates/http-php/content/spin.toml
@@ -1,3 +1,5 @@
+#:schema https://schemas.spinframework.dev/spin/manifest-v2/latest.json
+
 spin_manifest_version = 2
 
 [application]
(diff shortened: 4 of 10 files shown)


lann:
> upload schemas to a version-neutral location so users could automatically inherit improvements (or could still lock to a version via a GH URL). Thoughts?

I'm in favor of this. Three options come to mind:

- Github Page from this repo, e.g. `spinframework.github.io/spinframework/schemas/...`; probably the easiest to get started with since it can be built in this repo
- New subdomain e.g. `schemas.spinframework.dev/...` which could be hosted by Github Pages pretty easily; would probably make the most sense as a new e.g. `spin-schemas` repo; technically could be in this repo but that would be a little weird since you can only host one site per repo and schemas would "use up" that slot (though this could be migrated later)
- Existing docs site, e.g. `spinframework.dev/schemas/...`; would require some integration with https://github.com/spinframework/spin-docs

lann:
After editing the comment above several times I have developed An Opinion:

Let's create a subdomain like `schemas.spinframework.dev` pointing at Github Pages built from a branch in this repo. If we ever _really_ want to use this repo's Page for something else we can migrate the schemas somewhere else such as Github Pages in another repo by redirecting the subdomain.

lann:
^ I'm going to get started on this just because there is some lead time on the DNS update; we can always bail out if we don't like this approach.

lann:
This is now live at https://schemas.spinframework.dev/spin/manifest-v2/latest.json

We should update the release workflow to push to the `gh-pages-schemas` branch (note: this is an "orphan" branch so use a separate checkout for it).

itowlson:
@lann Thanks for setting that up.  I've updated the PR to reflect the official location.

Would you be open to uploading the canary schema in place of the 3.3.1 schema?  I believe it's the same schema but:

* Field descriptions explain what the field is, cover value formats and options, and link to the docs
* It (somewhat) works around the Taplo bug of not showing schema tooltips on non-empty arrays

Or is this something I can do myself by checking out the GH pages branch and copying the new schema into that?

(The PR is good to go regardless.)

lann:
> Or is this something I can do myself by checking out the GH pages branch and copying the new schema into that?

Yep, that.

Issue #2470, opened 2024-04-25
Hardcoded default port numbers prevent concurrent execution - consider testing if port is in use


garikAsplund:
I had a couple projects open at the same time and noticed that the default ports are hardcoded and thus collide.

<img width="403" alt="Screenshot 2024-04-24 at 9 18 38 PM" src="https://github.com/fermyon/spin/assets/111464359/dd470064-4bdb-4e93-950d-04d367cd7d91">
<img width="732" alt="Screenshot 2024-04-24 at 9 18 28 PM" src="https://github.com/fermyon/spin/assets/111464359/8b670ce3-0caa-44b7-a452-04b2314ce7bb">

I'm using spin 2.4.2.

I'd be willing to give this a shot tomorrow. The issue is in [crates/trigger-http/src/lib.rs](https://github.com/fermyon/spin/blob/main/crates/trigger-http/src/lib.rs). 

I propose amending the serve and serve_tls functions to go through ~10 ports if the one listening is in use. This would not just work for default_value but also if the "--listen ADDRESS" is also in use.

lann:
I don't think we should do this by default; if an app starts successfully but the port is hosting a different app than expected you could end up with frustration at best and a security issue at worst.

From personal experience I also just think this would be very surprising behavior; do you have any examples of other tools that work this way?

garikAsplund:
No worries. I'm used to webdev and [Vite has this feature](https://github.com/vitejs/vite/blob/f71ba5b94a6e862460a96c7bf5e16d8ae66f9fe7/packages/vite/src/node/http.ts#L141-L175). 

<img width="396" alt="Screenshot 2024-04-25 at 8 19 43 AM" src="https://github.com/fermyon/spin/assets/111464359/1d3ca385-f0a8-458e-8c4c-e2c516ab33d0">


It might not be that common of an issue and it's an easy enough quickfix by adding the --listen flag.

lann:
> Vite has this feature.

I've used Vite and never noticed this! :sweat_smile:

As pointed out out-of-band, another option that sort-of works today is `spin up --listen localhost:0`, which picks a random unused port. I say "sort-of" because it doesn't currently log the actual selected port, it just says `http://127.0.0.1:0` which isn't very helpful. _That_ is definitely something we should try to fix.

lann:
https://github.com/fermyon/spin/issues/2471

garikAsplund:
Thanks! I'll take a look.

itowlson:
[...]

PR #3182, opened 2025-07-03
adding custom port in cli opt-in

seun-ja:
Aims to resolve  issue [#2470](https://github.com/spinframework/spin/issues/2470)

Custom port can be provided in CLI, optional though

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -50,6 +50,9 @@ pub struct CliArgs {
     /// The path to the certificate key to use for https, if this is not set, normal http will be used. The key should be in PKCS#8 format
     #[clap(long, env = "SPIN_TLS_KEY", requires = "tls-cert")]
     pub tls_key: Option<PathBuf>,
+
+    #[clap(long = "find-free-port")]
+    pub find_free_port: bool,
 }
 
 impl CliArgs {
@@ -73,6 +76,7 @@ pub struct HttpTrigger {
     /// If the port is set to 0, the actual address will be determined by the OS.
     listen_addr: SocketAddr,
     tls_config: Option<TlsConfig>,
+    find_free_port: bool,
 }
 
 impl<F: RuntimeFactors> Trigger<F> for HttpTrigger {
@@ -82,7 +86,14 @@ impl<F: RuntimeFactors> Trigger<F> for HttpTrigger {
     type InstanceState = ();
 
     fn new(cli_args: Self::CliArgs, app: &spin_app::App) -> anyhow::Result<Self> {
-        Self::new(app, cli_args.address, cli_args.into_tls_config())
+        let find_free_port = cli_args.find_free_port;
+
+        Self::new(
+            app,
+            cli_args.address,
+            cli_args.into_tls_config(),
+            find_free_port,
+        )
     }
 
     async fn run(self, trigger_app: TriggerApp<F>) -> anyhow::Result<()> {
@@ -104,12 +115,14 @@ impl HttpTrigger {
         app: &spin_app::App,
         listen_addr: SocketAddr,
         tls_config: Option<TlsConfig>,
+        find_free_port: bool,
     ) -> anyhow::Result<Self> {
         Self::validate_app(app)?;
 
         Ok(Self {
             listen_addr,
             tls_config,
+            find_free_port,
         })
     }
 
@@ -121,8 +134,14 @@ impl HttpTrigger {
         let Self {
             listen_addr,
             tls_config,
+            find_free_port,
         } = self;
-        let server = Arc::new(HttpServer::new(listen_addr, tls_config, trigger_app)?);
+        let server = Arc::new(HttpServer::new(
+            listen_addr,
+            tls_config,
+            find_free_port,
+            trigger_app,
+        )?);
         Ok(server)
     }
 
--- a/crates/trigger-http/src/server.rs
+++ b/crates/trigger-http/src/server.rs
@@ -1,4 +1,10 @@
-use std::{collections::HashMap, future::Future, io::IsTerminal, net::SocketAddr, sync::Arc};
+use std::{
+    collections::HashMap,
+    future::Future,
+    io::{ErrorKind, IsTerminal},
+    net::SocketAddr,
+    sync::Arc,
+};
 
 use anyhow::{bail, Context};
 use http::{
@@ -41,12 +47,16 @@ use crate::{
     Body, NotFoundRouteKind, TlsConfig, TriggerApp, TriggerInstanceBuilder,
 };
 
+pub const MAX_RETRIES: u16 = 10;
+
 /// An HTTP server which runs Spin apps.
 pub struct HttpServer<F: RuntimeFactors> {
     /// The address the server is listening on.
     listen_addr: SocketAddr,
     /// The TLS configuration for the server.
     tls_config: Option<TlsConfig>,
+    /// Whether to find a free port if the specified port is already in use.
+    find_free_port: bool,
     /// Request router.
     router: Router,
     /// The app being triggered.
@@ -62,6 +72,7 @@ impl<F: RuntimeFactors> HttpServer<F> {
     pub fn new(
         listen_addr: SocketAddr,
         tls_config: Option<TlsConfig>,
+        find_free_port: bool,
         trigger_app: TriggerApp<F>,
     ) -> anyhow::Result<Self> {
         // This needs to be a vec before building the router to handle duplicate routes
@@ -129,6 +140,7 @@ impl<F: RuntimeFactors> HttpServer<F> {
         Ok(Self {
             listen_addr,
             tls_config,
+            find_free_port,
             router,
             trigger_app,
             component_trigger_configs,
@@ -138,18 +150,55 @@ impl<F: RuntimeFactors> HttpServer<F> {
 
     /// Serve incoming requests over the provided [`TcpListener`].
     pub async fn serve(self: Arc<Self>) -> anyhow::Result<()> {
-        let listener = TcpListener::bind(self.listen_addr).await.with_context(|| {
-            format!(
-                "Unable to listen on {listen_addr}",
-                listen_addr = self.listen_addr
-            )
-        })?;
+        let listener: TcpListener = if self.find_free_port {
+            self.search_for_free_port().await?
+        } else {
+            TcpListener::bind(self.listen_addr).await.map_err(|err| {
+                if err.kind() == ErrorKind::AddrInUse {
+                    anyhow::anyhow!("{} is already in use. To have Spin search for a free port, use the --find-free-port option.", self.listen_addr)
+                } else {
+                    anyhow::anyhow!("Unable to listen on {}: {err:?}", self.listen_addr)
+                }
+            })?
+        };
+
         if let Some(tls_config) = self.tls_config.clone() {
             self.serve_https(listener, tls_config).await?;
         } else {
             self.serve_http(listener).await?;
         }
         Ok(())
+    }
+
+    async fn search_for_free_port(&self) -> anyhow::Result<TcpListener> {
+        let mut found_listener = None;
+        let mut addr = self.listen_addr;
+
+        for _ in 1..=MAX_RETRIES {
+            if addr.port() == u16::MAX {
+                anyhow::bail!(
+                    "Couldn't find a free port as we've reached the maximum port number. Consider retrying with a lower base port."
+                );
+            }
+
+            match TcpListener::bind(addr).await {
+                Ok(listener) => {
+                    found_listener = Some(listener);
+                    break;
+                }
+                Err(err) if err.kind() == ErrorKind::AddrInUse => {
+                    addr.set_port(addr.port() + 1);
+                    continue;
+                }
+                Err(err) => anyhow::bail!("Unable to listen on {addr}: {err:?}",),
+            }
+        }
+
+        found_listener.ok_or_else(|| anyhow::anyhow!(
+            "Couldn't find a free port in the range {}-{}. Consider retrying with a different base port.",
+            self.listen_addr.port(),
+            self.listen_addr.port() + MAX_RETRIES
+        ))
     }
 
     async fn serve_http(self: Arc<Self>, listener: TcpListener) -> anyhow::Result<()> {
--- a/tests/testing-framework/src/runtimes/in_process_spin.rs
+++ b/tests/testing-framework/src/runtimes/in_process_spin.rs
@@ -104,7 +104,7 @@ async fn initialize_trigger(
     .await?;
 
     let app = spin_app::App::new("my-app", locked_app);
-    let trigger = HttpTrigger::new(&app, "127.0.0.1:80".parse().unwrap(), None)?;
+    let trigger = HttpTrigger::new(&app, "127.0.0.1:80".parse().unwrap(), None, false)?;
     let mut builder = TriggerAppBuilder::<_, FactorsBuilder>::new(trigger);
     let trigger_app = builder
         .build(


seun-ja:
I came across this error after running `make test`. 

```bash
error: 1 target failed:
    `-p spin-factor-outbound-http --test factor_test`
make: *** [test-unit] Error 101
```

 All tests passed. So I don't know it this is something that can be ignored

lann:
@seun-ja Could you please take a look at https://spinframework.dev/v3/contributing-spin#committing-and-pushing-your-changes? This project requires commit signing _and_ sign-off (two entirely separate things).

crates/trigger-http/src/lib.rs
```diff
@@ -50,6 +50,10 @@ pub struct CliArgs {
     /// The path to the certificate key to use for https, if this is not set, normal http will be used. The key should be in PKCS#8 format
     #[clap(long, env = "SPIN_TLS_KEY", requires = "tls-cert")]
     pub tls_key: Option<PathBuf>,
+
+    /// The port to listen on
+    #[clap(long = "port", env = "SPIN_HTTP_LISTEN_PORT")]
+    pub port: Option<u16>,
```
itowlson: Could you expand on how you see this interacting with `--listen`?  The way I usually override the port is e.g. `spin up --listen 127.0.0.1:4000` - is this envisaged as a shortcut to that?  Or a way to set the address and port separately?

crates/trigger-http/src/server.rs
```diff
@@ -138,16 +148,31 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                    Ok(listener)
+                } else {
+                    Err(anyhow::anyhow!("Unable to listen on {}", self.listen_addr))
+                }
+            }
+        };
+
         if let Some(tls_config) = self.tls_config.clone() {
-            self.serve_https(listener, tls_config).await?;
+            self.serve_https(listener?, tls_config).await?;
```
itowlson: I'm not sure I love having to `?` the listener everywhere it's used.  Why not stick with the previous pattern of failing errors _before_ assigning the `listener` variable?

itowlson (commented):
I'm not sure this addresses #2470 after all - that issue asks for Spin to automatically find a free port number (which we are dubious about wanting as default), whereas this PR provides another way to select a fixed port number.

Perhaps it would be useful to say more about the motivation for this change?

seun-ja:
In hindsight, this isn't an ideal approach.

I didn't take into consideration that `--listen` is already custom. All I need to do here is just check for the error kind that I already used in the last commit, check if it is `AddrInUse` and use a wildcard 0 so the OS can just assign one.

itowlson:
@seun-ja Makes sense!  It might be worth reading through the comment thread on that issue though, if you haven't already, as there were some doubts/discussions - e.g. I know @lann was reluctant to have this behaviour by default.  Thanks for looking into it!

lann:
IMO if you'd be happy with a random port then you should just listen on `:0` to begin with, but then again I probably won't be using this feature anyway. :slightly_smiling_face: 

Based on the discussion in the issue I'd suggest something like this:

- Add a new `--find-free-port` flag
- If `bind` fails with `AddrInUse` and the flag was set, add 1 to the port and try again
- Fail after 10 tries

karthik2804:
Would be a good idea to prompt the user if we are on an interactive CLI where we can have a confirm prompt with the user if they want to use a different port?

itowlson:
@karthik2804 Because multi-trigger-type applications exist, I'd be inclined to adopt a philosophy that triggers should not prompt for input.  Otherwise we risk having to manage several child processes all prompting at once, which is likely to be a bit chaotic.  Or partial starts where one trigger is processing events but another is stuck at a prompt.  (Sure, we could do a little dance to inform a trigger that it's the only one so it's allowed to prompt, but this makes both `up` and the trigger more complicated, and means the trigger behaves differently depending on single vs multiple.)  Open to thoughts of course but for now I'd suggest we stick to a helpful failure message.

crates/trigger-http/src/lib.rs
```diff
@@ -82,7 +86,14 @@ impl<F: RuntimeFactors> Trigger<F> for HttpTrigger {
...
 
     fn new(cli_args: Self::CliArgs, app: &spin_app::App) -> anyhow::Result<Self> {
-        Self::new(app, cli_args.address, cli_args.into_tls_config())
+        let find_free_port = cli_args.find_free_port;
+
+        Self::new(
+            app,
+            cli_args.address,
+            cli_args.into_tls_config(),
+            find_free_port,
```
itowlson, line 95: You can probably write `cli_args.find_free_port` - the local variable doesn't really add any explanatory power.

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,43 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                        addr.set_port(addr.port() + 1);
+
+                        match TcpListener::bind(addr).await {
+                            Ok(listener) => {
+                                found_listener = Some(listener);
+                                break;
+                            }
+                            Err(err) => {
+                                if err.kind() == ErrorKind::AddrInUse {
+                                    continue;
```
itowlson: This routine gets into some quite deep nesting (a match inside a loop inside an if inside a match) and for me it was hard to follow.  Is it feasible to break out a helper function?

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,43 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                                }
+                                return Err(anyhow::anyhow!("Unable to listen on {}", addr));
```
itowlson: If there's a non-AddrInUse error on one of the scanned ports, this will produce a strange error message e.g. the user said to listen on port 4000 and gets an error about port 4007.

itowlson: Is it correct that we bail here?  Why would an error on port 3002 stop us from trying port 3003?  (Is it, perhaps, that errors other than AddrInUse imply we've gone to a bad place?)

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,43 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
-            )
-        })?;
+        let listener: TcpListener = match TcpListener::bind(self.listen_addr).await {
+            Ok(listener) => listener,
+            Err(err) => {
+                if self.find_free_port && err.kind() == ErrorKind::AddrInUse {
+                    let mut found_listener = None;
+                    for _ in 1..=9 {
+                        let mut addr = self.listen_addr;
+                        addr.set_port(addr.port() + 1);
```
itowlson: What happens if `addr.port()` is close to `U16_MAX`?  Will this overflow round to 0?  Is that expected behaviour?

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,43 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
-                listen_addr = self.listen_addr
-            )
-        })?;
+        let listener: TcpListener = match TcpListener::bind(self.listen_addr).await {
+            Ok(listener) => listener,
+            Err(err) => {
+                if self.find_free_port && err.kind() == ErrorKind::AddrInUse {
+                    let mut found_listener = None;
+                    for _ in 1..=9 {
+                        let mut addr = self.listen_addr;
```
itowlson: This reads as if you are resetting `addr` each time through the loop, which means that each iteration is trying `self.listen_addr` with +1 to its port.  Have you confirmed that if e.g. 3000-3002 are all in use then this does find 3003?  I may be misreading for sure

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,43 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                    match found_listener {
+                        Some(listener) => listener,
+                        None => {
+                            return Err(anyhow::anyhow!(
+                                "All retries failed. Unable to bind to a free port"
+                            ));
+                        }
+                    }
+                } else {
+                    return Err(anyhow::anyhow!("Unable to listen on {}", self.listen_addr));
```
itowlson: Consider mentioning the `--find-free-port` option e.g.
```suggestion
                    return Err(anyhow::anyhow!("Unable to listen on {}. To have Spin search for a free port, use the --find-free-port option.", self.listen_addr));
```

tests/testing-framework/src/runtimes/in_process_spin.rs
```diff
@@ -104,7 +104,7 @@ async fn initialize_trigger(
     .await?;
 
     let app = spin_app::App::new("my-app", locked_app);
-    let trigger = HttpTrigger::new(&app, "127.0.0.1:80".parse().unwrap(), None)?;
+    let trigger = HttpTrigger::new(&app, "127.0.0.1:80".parse().unwrap(), None, true)?;
```
itowlson: Do we want to end up on a random port here?  I am not sure if the testing infra will be happy with that.  Although I guess it should never happen.  But it might be better for it to fail here if it did.

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,43 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                                return Err(anyhow::anyhow!("Unable to listen on {}", addr));
+                            }
+                        }
+                    }
+
+                    match found_listener {
+                        Some(listener) => listener,
+                        None => {
+                            return Err(anyhow::anyhow!(
+                                "All retries failed. Unable to bind to a free port"
```
itowlson: From the user point of view, they told us to find a free port, not to "retry" anything, so I'd suggest we re-word this error message in those terms e.g.
```suggestion
                                "Couldn't find a free port in the range {min}-{max}. Consider retrying with a different base port."
```

or something like that?

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,43 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                            Err(err) => {
+                                if err.kind() == ErrorKind::AddrInUse {
+                                    continue;
+                                }
+                                return Err(anyhow::anyhow!("Unable to listen on {}", addr));
+                            }
+                        }
+                    }
+
+                    match found_listener {
```
itowlson: You can achieve the same effect with `found_listener.ok_or_else(...)?` - your call which you think is more readable!

itowlson (commented):
Some style and wording nits, but my main concern is if the loop is fully doing what you intend.  That may be a code reading fail on my part though!  Great to see this progressing!

crates/trigger-http/src/lib.rs
seun-ja: that was my initial thoughts as well, but the issue is `cli_args.into_tls_config()` consumes `self`. this approach is a way to move `find_free_port` value into a new variable. 

The readability is agreeable not the best especially in reference to the pattern already in the code but that seems like the way out here unless I would have to do some adjustment to  `into_tls_config()`

crates/trigger-http/src/server.rs
seun-ja: There should be.. would look into it

crates/trigger-http/src/server.rs
seun-ja: It only bails if the error kind isn't `AddrInUse`, so it doesn't stop us from trying if 3002 isn't available.

crates/trigger-http/src/server.rs
seun-ja: I should take that into consideration. Thanks for pointing this out

crates/trigger-http/src/lib.rs
itowlson: That makes sense - no need to change it then - thanks for explaining!

crates/trigger-http/src/server.rs
itowlson: I understand that it continues on `AddrInUse`.  I am asking why it _bails_ on other errors.  Is the assumption that other errors are unrecoverable and there is no point searching further?

seun-ja: There are probably other recoverable errors. I only focused on `AddrInUse` for the scope of this issue. 

Should I add the ones that could be recoverable here, or just create a new issue for that?

itowlson: If it's going to turn into a faff then don't worry about it - stick with what you have and we can always re-evaluate if something comes up (i.e. no need to raise an issue either).

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,46 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+        let listener: TcpListener = match TcpListener::bind(self.listen_addr).await {
+            Ok(listener) => listener,
+            Err(err) => {
+                if self.find_free_port && err.kind() == ErrorKind::AddrInUse {
+                    let mut found_listener = None;
+                    let mut addr = self.listen_addr;
+                    for _ in 1..=9 {
+                        addr.set_port(addr.port() + 1);
+
+                        if !addr.port() == u16::MAX {
```
itowlson: The `!` operator here is bitwise complement.  Is that what you intended?

seun-ja: I missed that, was initially trying a `while` loop. Would fix and push a fix now

crates/trigger-http/src/server.rs
itowlson: I don't think this check addresses the concern.  Look at line 158, which happens _before_ this check.  What happens if `addr.port()` returns `U16::MAX`?  What value gets passed to `set_port` in that case?

Also, `(u16::MAX + 1) as u32` is 0 in release builds and panic in debug builds, neither of which feels like what you intend...?

itowlson: I'd suggest instead checking if `addr.port() == u16::MAX` _before_ incrementing (at the very top of the loop), and bailing if so.

seun-ja: Yeah, that makes sense.

crates/trigger-http/src/server.rs
itowlson: I think there's still value in breaking out a helper.  The long and complicated if and loop make it very hard (at least for me) to see what is basically simple "if it succeeds, use it; if it doesn't, search or bail."  Consider e.g.

```
let listener = match TcpListener::bind(self.listen_addr).await {
    Ok(listener) => listener,
    Err(err) if self.find_free_port && err.kind() == ErrorKind::AddrInUse => self.search_for_free_port().await?,
    Err(err) => anyhow::bail!(...)
}

fn search_for_free_port(&self) -> anyhow::Result<TcpListener> {
    // declare mutables
    // loop
    // ok or error
}
```

There's probably more breaking out you could do but to me this greatly clarifies the top-level logic, and gives a meaningful name to that long chunk for which I currently have to infer the intention from the code.

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,46 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                                }
+                                anyhow::bail!("Unable to listen on {}", self.listen_addr);
+                            }
+                        }
+                    }
+
+                    found_listener.ok_or_else(|| anyhow::anyhow!(
+                        "Couldn't find a free port in the range {}-{}. Consider retrying with a different base port.",
+                        self.listen_addr.port(),
+                        self.listen_addr.port() + 9
```
itowlson: Duplicate magic number (with line 157).  Make this a const, or if you adopt the helper function, it could even be a parameter passed to the helper.

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,46 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                    ))?
+                } else {
+                    anyhow::bail!("Unable to listen on {}. To have Spin search for a free port, use the --find-free-port option.", self.listen_addr)
```
itowlson: It occurs to me this can happen even if they passed `--find-free-port` (if the error was something other than AddrInUse) (at least I think it can, I'm finding it tricky to visually match which if or case we're in here).  So we probably need to distinguish those cases I'm afraid - sorry for the bad steer.

Also, we should surface what the actual error is, because there are few things more maddening than being told 
"computer says no" and not being told _why._

If you decide to go with my suggestion about breaking out the loop logic, this might be fairly easy e.g.

```
match TcpListener::bind(self.listen_addr).await {
    Ok(listener) => listener,
    Err(err) if err.kind() == ErrorKind::AddrInUse && self.find_free_port => self.search_for_free_port().await?,
    Err(err) if err.kind() == ErrorKind::AddrInUse => anyhow::bail!(/* use --find-free-port */),
    Err(err) => anyhow::bail!(/* --find-free-port wouldn't have helped, just report the error */)
}
```

but not sure, would need some testing and another pair of eyes for sure!

itowlson (changes requested):
Thanks for all the updates!  I think the logic is sound (except for one nitty detail of an error message which is not a blocker), but I remain really troubled by the long nameless block containing the search logic.  It is long enough and nested enough that I find it really hard to be confident about following the logic, and in particular to visually match open and close braces to know what scope any given line is in, especially towards the end of the block.  I've suggested a way that seems a lot clearer to me (and that, I think, helps with the nitty error thing), but I'm not wedded to it, and it doesn't fully solve the problem: if you have better ideas then go for it.  But I do think it would be desirable to do _something_ unless it turns out impractical.  Which it sometimes does, so please push back if it's going to be a big pain!

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,46 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+        let listener: TcpListener = match TcpListener::bind(self.listen_addr).await {
+            Ok(listener) => listener,
+            Err(err) => {
+                if self.find_free_port && err.kind() == ErrorKind::AddrInUse {
+                    let mut found_listener = None;
+                    let mut addr = self.listen_addr;
+                    for _ in 1..=9 {
+                        if addr.port() == u16::MAX {
+                            anyhow::bail!(
+                                "Couldn't find a free port as we've reached the maximum port number. Consider retrying with a lower base port."
```
lann, line 160: Why recommend trying with a _lower_ port?

crates/trigger-http/src/server.rs
lann: Yeah I've never found a nice way to do retry loops without a separate function so you can return from the middle of the loop.

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +148,46 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+                            Ok(listener) => {
+                                found_listener = Some(listener);
+                                break;
+                            }
+                            Err(err) => {
+                                if err.kind() == ErrorKind::AddrInUse {
+                                    continue;
+                                }
+                                anyhow::bail!("Unable to listen on {}", self.listen_addr);
+                            }
```
lann: It's a little confusing to use the _original_ port in this error message and we should include the source error.

Separately/optionally, you could move the `if/continue` out to a match guard clause:

```suggestion
                            Err(err) if err.kind() == ErrorKind::AddrInUse => {
                            	// Try the next port
                            }
                            Err(err) => {
                                anyhow::bail!("Unable to listen on {addr}: {err:?}");
                            }
```

lann (changes requested):

crates/trigger-http/src/server.rs
lann: I think it might be easier to follow if the `find_free_port` check moved to the outside, e.g.

```rust
let listener = if self.find_free_port {
    // try 10 times here
} else {
    // only try once here
    TcpListener::bind(self.listen_addr)...
};
```

That would make it easier to have the error messages make sense in the two different cases.

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +150,18 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
-            )
-        })?;
+        let listener: TcpListener = if self.find_free_port {
+            self.search_for_free_port().await?
+        } else {
+            TcpListener::bind(self.listen_addr).await.map_err(|err| {
+                if err.kind() == ErrorKind::AddrInUse {
+                    anyhow::anyhow!("Unable to listen on {}. To have Spin search for a free port, use the --find-free-port option.", self.listen_addr)
+                } else {
+                    anyhow::anyhow!("Unable to listen on {}", self.listen_addr)
```
itowlson: I still feel like we are dropping the actual error here - am I missing something?

crates/trigger-http/src/server.rs
```diff
@@ -152,6 +170,36 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
+        let mut addr = self.listen_addr;
+
+        for _ in 1..=MAX_RETRIES {
+            if addr.port() == u16::MAX {
+                anyhow::bail!(
+                    "Couldn't find a free port as we've reached the maximum port number. Consider retrying with a lower base port."
+                );
+            }
+
+            addr.set_port(addr.port() + 1);
```
itowlson: I think the ordering here will cause us to start searching at requested port plus 1, rather than at the requested port, even if the requested port is free.  Am I misreading?

crates/trigger-http/src/server.rs
```diff
@@ -138,12 +150,18 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
-                "Unable to listen on {listen_addr}",
-                listen_addr = self.listen_addr
-            )
-        })?;
+        let listener: TcpListener = if self.find_free_port {
+            self.search_for_free_port().await?
+        } else {
+            TcpListener::bind(self.listen_addr).await.map_err(|err| {
+                if err.kind() == ErrorKind::AddrInUse {
+                    anyhow::anyhow!("Unable to listen on {}. To have Spin search for a free port, use the --find-free-port option.", self.listen_addr)
```
itowlson: It occurs to me that since we now know the kind of error we can reword this to e.g.

```suggestion
                    anyhow::anyhow!("{} is already in use. To have Spin search for a free port, use the --find-free-port option.", self.listen_addr)
```

lann (approved):

PR #3186, opened 2025-07-08
Upgrade wasmtime to v35.0.0

fibonacci1729:
~Upgrade to the `wasmtime v35.0.0` release branch; FWIW i'm fine leaving this sit while waiting for the crates.io release in approximately 2 weeks and switching then.~

Upgrade to the `v35.0.0` crates.io release of `wasmtime`

Notably this brings in a change to wasmtime to unforbid the `te` header which allows for gRPC clients to work in Spin components.


fibonacci1729:
Waiting to rebase onto https://github.com/spinframework/spin/pull/3187

Cargo.toml
```diff
@@ -172,9 +172,9 @@ wasm-metadata = "0.235"
 wasm-pkg-client = "0.11"
 wasm-pkg-common = "0.11"
 wasmparser = "0.235"
-wasmtime = "34.0.0"
-wasmtime-wasi = "34.0.0"
-wasmtime-wasi-http = "34.0.0"
+wasmtime = { git = "https://github.com/bytecodealliance/wasmtime", package = "wasmtime", branch = "release-35.0.0" }
```
itowlson: Is it correct to link to a `branch` (which is mutable) rather than a commit SHA?  This could result in different builds happening against different upstreams couldn't it?

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -49,7 +49,10 @@ where
     wasmtime_wasi_http::bindings::http::outgoing_handler::add_to_linker::<_, HasHttp>(
         linker, get_http,
     )?;
-    wasmtime_wasi_http::bindings::http::types::add_to_linker::<_, HasHttp>(linker, get_http)?;
+    let link_opts = wasmtime_wasi_http::bindings::http::types::LinkOptions::default();
+    wasmtime_wasi_http::bindings::http::types::add_to_linker::<_, HasHttp>(
+        linker, &link_opts, get_http,
```
itowlson: Does it work to say `&Default::default()` for the second arg, and remove the explicit `link_opts` object?  I'm not familiar enough to be sure, but if it does it might make intent clearer?

itowlson (commented):
If Alex has approved it then I reckon I can't add much but I've added it anyway!

lann:
> FWIW i'm fine leaving this sit while waiting for the crates.io release in approximately 2 weeks and switching then.

Unless there is a compelling need for this before then let's do that. Should we move this PR back to draft in the meantime?

Cargo.toml
fibonacci1729: Good catch; resolving this as we'll update to use the crates.io release once it's out.

alexcrichton: While this isn't a precise lock it's similar to a version requirement such as `wasmtime = "34.0.0"` in that the version requirement listed is actually `wasmtime = ">= 34.0.0, < 35.0.0"` which is any possible range of versions. For Wasmtime's case in particular while it's possible we do API-breaking things on a release branch it's also pretty unlikely, so this is more-or-less the same as a crates.io dep in that regard. That being said if there's no urgency I think it's also reasonable to wait for the publish too.

fibonacci1729:
Now that `wasmtime v35.0.0` is released to crates.io i am going to mark this ready for review.

itowlson (approved):
:shipit: :shipit: :shipit:

PR #3192, opened 2025-07-11
isolate and handle unix error

### LINKED ISSUE #3161: Improve error messages when trigger subprocess killed by signal
We've had a couple of similar cases where MacOS was killing the spin trigger subprocess due to a misconfiguration: https://github.com/spinframework/spin/issues/2553 https://github.com/spinframework/spin/issues/3154

The error message from Spin is very unhelpful: `subprocess exited with status: unknown`

It looks like we can probably do better based on the docs for [`ExitStatus.code`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code):

> On Unix, this will return None if the process was terminated by a signal. [ExitStatusExt](https://doc.rust-lang.org/std/os/unix/process/trait.ExitStatusExt.html) is an extension trait for extracting any such signal, and other details, from the ExitStatus.

`ExitStatusExt` is unix-only so we'd have to `#[cfg(unix)]` a bit of logic but it should let us provide a more helpful error here.

(there is also a windows-only `ExitStatusExt` but it doesn't appear to expose anything useful here)

seun-ja:
Aims to Fix #3161

### DIFF
--- a/src/subprocess.rs
+++ b/src/subprocess.rs
@@ -4,19 +4,45 @@
 /// When this error is encountered the cli will exit with the status code
 /// instead of printing an error,
 #[derive(Debug)]
-pub struct ExitStatusError {
-    status: Option<i32>,
+pub enum ExitStatusError {
+    /// The subprocess exited with a specific exit code.
+    ExitCode(i32),
+    /// The subprocess was exited by a signal. Only available for `Unix` based systems.
+    Signal(i32),
+    /// Catchall for general errors. Error code `1`
+    Unknown,
 }
 
 impl ExitStatusError {
+    #[cfg(unix)]
     pub(crate) fn new(status: std::process::ExitStatus) -> Self {
-        Self {
-            status: status.code(),
+        use std::os::unix::process::ExitStatusExt as _;
+
+        if let Some(code) = status.code() {
+            Self::ExitCode(code)
+        } else if let Some(signal) = status.signal() {
+            Self::Signal(signal)
+        } else {
+            Self::Unknown
+        }
+    }
+
+    #[cfg(not(unix))]
+    pub(crate) fn new(status: std::process::ExitStatus) -> Self {
+        if let Some(code) = status.code() {
+            Self::ExitCode(code)
+        } else {
+            Self::Unknown
         }
     }
 
     pub fn code(&self) -> i32 {
-        self.status.unwrap_or(1)
+        match self {
+            Self::ExitCode(code) => *code,
+            Self::Signal(signal) => 128 + *signal,
+            Self::Unknown => 1,
+        }
+        .min(255)
     }
 }
 
@@ -24,11 +50,12 @@ impl std::error::Error for ExitStatusError {}
 
 impl std::fmt::Display for ExitStatusError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let _ = write!(f, "subprocess exited with status: ");
-        if let Some(status) = self.status {
-            writeln!(f, "{status}")
-        } else {
-            writeln!(f, "unknown")
+        let _ = write!(f, "subprocess exited with ");
+
+        match self {
+            Self::ExitCode(code) => write!(f, "exit code: {code}"),
+            Self::Signal(signal) => write!(f, "signal: {signal}"),
+            Self::Unknown => write!(f, "unknown status"),
         }
     }
 }


src/subprocess.rs
```diff
@@ -9,12 +11,20 @@ pub struct ExitStatusError {
...
     pub(crate) fn new(status: std::process::ExitStatus) -> Self {
         Self {
             status: status.code(),
         }
     }
 
+    #[cfg(unix)]
+    pub(crate) fn new(status: std::process::ExitStatus) -> Self {
+        Self {
+            status: status.code().or_else(|| status.signal()),
```
itowlson: This conflates a process exit code with an operating system signal number, and won't help with the original problem - it will just make it appear as if the subprocess has exited with a weird exit code.

A possible way forward is to make `ExitStatusError` an _enum_ e.g.

```
pub enum ExitStatusError {
    ExitCode(i32),
    Signal(i32),
    Unknown,
}
```

Then the message can report which case was in play, distinguishing the signal case from the process exit case.

~(I am not sure if we can translate the `i32` signal number into a meaningful-ish string - that would be great if we could.)~ _(ETA from light googling this looks like it would be a pain but maybe someone knows better)_

src/subprocess.rs
```diff
@@ -1,32 +1,44 @@
...
+            Self::ExitCode(code)
+        } else if let Some(signal) = status.signal() {
+            Self::Signal(signal)
+        } else {
+            Self::Unknown
         }
     }
 
-    #[cfg(unix)]
+    #[cfg(windows)]
```
lann: nit: Just in case someone wants to try compiling spin to wasm some day we could make things slightly easier for them. :slightly_smiling_face: 

```suggestion
    #[cfg(not(unix))]
```

src/subprocess.rs
```diff
@@ -1,32 +1,44 @@
...
         }
     }
 
     pub fn code(&self) -> i32 {
-        self.status.unwrap_or(1)
+        match self {
+            Self::ExitCode(code) => *code,
+            Self::Signal(signal) => *signal,
+            Self::Unknown => 1,
+        }
```
lann: It would be good to at least try to avoid confusion between exit codes and signals, even though its only slightly within our control here.

https://tldp.org/LDP/abs/html/exitcodes.html
> 128+n Fatal error signal "n"
> 255 Exit status out of range
```suggestion
        // Exit code conventions: https://tldp.org/LDP/abs/html/exitcodes.html
        match self {
            Self::ExitCode(code) => *code,
            Self::Signal(signal) => 128 + *signal,
            Self::Unknown => 1,
        }.min(255)
```

src/subprocess.rs
```diff
@@ -35,10 +47,11 @@ impl std::error::Error for ExitStatusError {}
...
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let _ = write!(f, "subprocess exited with status: ");
-        if let Some(status) = self.status {
-            writeln!(f, "{status}")
-        } else {
-            writeln!(f, "unknown")
+
+        match self {
+            Self::ExitCode(code) => writeln!(f, "{code}"),
+            Self::Signal(signal) => writeln!(f, "{signal}"),
```
lann: We should expose the new detail we have:

```suggestion
            Self::Signal(signal) => writeln!(f, "signal {signal}"),
```

src/subprocess.rs
seun-ja: Thanks for sharing the docs, I've been curious to know why the unknown was initially set at 1. Makes more sense now!

src/subprocess.rs
```diff
@@ -25,10 +51,11 @@ impl std::error::Error for ExitStatusError {}
 impl std::fmt::Display for ExitStatusError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let _ = write!(f, "subprocess exited with status: ");
```
itowlson: This is extremely nitty but `status: signal 7` reads a wee bit oddly to me.  How would you feel about e.g. "subprocess exited with exit code 5", "subprocess exited with signal 7", "subprocess exited with unknown status"?  Something like

```
let _ = write!(f, "subprocess exited with ");
match self {
    Self::ExitCode(code) => write!(f, "exit code {code}"),
    Self::Signal(signal) => write!(f, "signal {signal}"),
    Self::ExitCode(code) => write!(f, "unknown status"),
}
```

Definitely not a blocker though and there may be others who prefer to leave as is!

itowlson (commented):
I separated out the Zig CI stuff into a single-purpose PR, so please revert all GH workflow changes out of this one - thanks!

But the signals stuff looks great - thank you for improving this!  Happy to approve once the CI stuff rolls back.

src/subprocess.rs
seun-ja: Yes, this is better. Would go for this, thanks!

.github/workflows/build.yml
itowlson, line 1: There shouldn't be any changes needed in the build.yml I think?

seun-ja: It's only formatting. My IDE auto-formatted the file when I worked on it. The CI still works.

Should I revert the changes?

itowlson: Yes please.  We mostly try to avoid PRs containing unrelated changes to different files, even formatting, because it makes the file history harder to follow.

PR #3197, opened 2025-07-14
Initialise git repo in new app

### LINKED ISSUE #2580: Add version control support to `spin new` or spin CLI

itowlson:
Fixes #2580 

This picks up on work done in #3155 by @iamrajiv.  The core of that PR was fine, but unfortunately there were some unrelated blocking problems, and the PR has since gone inactive.

### DIFF
--- a/crates/templates/src/git.rs
+++ b/crates/templates/src/git.rs
@@ -1,4 +1,6 @@
-use std::io::ErrorKind;
+use std::{io::ErrorKind, path::Path, process::Stdio};
+
+use anyhow::Context;
 
 // TODO: the following and the second half of plugins/git.rs are duplicates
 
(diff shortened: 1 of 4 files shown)


crates/templates/src/git.rs
```diff
@@ -46,3 +46,24 @@ impl UnderstandGitResult for Result<std::process::Output, std::io::Error> {
...
+pub(crate) async fn is_in_git_repo(dir: &Path) -> anyhow::Result<bool> {
+    let mut cmd = tokio::process::Command::new("git");
+    cmd.arg("-C")
+        .arg(dir)
+        .arg("rev-parse")
+        .arg("--git-dir")
+        .stdout(Stdio::null())
+        .stderr(Stdio::null());
+
+    let status = cmd.status().await?;
```
rylev: nit: use of `context` would likely help here as the error message from the status call is likely to not be very helpful out of context.

crates/templates/src/run.rs
```diff
@@ -352,6 +355,32 @@ impl Run {
         }
     }
 
+    async fn initialise_git(&self) -> anyhow::Result<()> {
```
rylev: Nit:

```suggestion
    async fn maybe_initialise_git(&self) -> anyhow::Result<()> {
```

crates/templates/src/run.rs
```diff
@@ -352,6 +355,32 @@ impl Run {
...
+            return Ok(());
+        }
+
+        if self.options.no_vcs {
+            return Ok(());
+        }
+
+        let target_dir = self.generation_target_dir();
+
+        let skip_initing_repo = git::is_in_git_repo(&target_dir).await.unwrap_or(true);
```
rylev, line 369: Should we warn if `is_in_git_repo` errors?

rylev (approved):

iamrajiv (approved):
lgtm sorry i missed the notice for the review comment @itowlson thanks for the fix

adamreese:
@itowlson Rebasing on main will fix the CI errors now

crates/templates/src/run.rs
itowlson: I would say not: we don't want to error on "git not installed," and other cases seem unlikely.  We could trace or something if you feel strongly?

PR #3199, opened 2025-07-14
Smoke test local templates as well as remote ones

itowlson:
The template smoke tests currently validate that changes to Spin don't break the templates on `main`.  However, if a PR changes the templates, they _don't_ check that those changes still work.  This has been a source of occasional pain, because when broken templates get into main (because of not being tested), that causes all _subsequent_ PRs to fail (because they run tests using the now-broken templates in main).

This PR tests both local and main templates, to catch both possible sources of sadness.

~WIP to see if the technique works: if it does, it will need some clearer distinction in the failure messages, and then rolling out to other templates.  (There will also be a certain amount of fault injection along the way.)~

ETA: To be clear, this does not remedy the problem of "broken templates in main turn all PRs red, even the ones that fix the broken templates," but it hopefully reduces the risk of broken templates getting into main.

Also, the current tests mean that if JS or Python break _their_ templates, then Spin PRs turn red, and this does not address that.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -459,12 +459,22 @@ mod integration_tests {
         http_smoke_test_template(
             "http-rust",
             None,
-            None,
             &[],
             |_| Ok(()),
             HashMap::default(),
             "Hello World!",
         )
+        .context("template on main")?;
+        http_smoke_test_template(
+            "http-rust",
+            Some(crate::testcases::TemplateSource::Local),
+            &[],
+            |_| Ok(()),
+            HashMap::default(),
+            "Hello World!",
+        )
+        .context("local template")?;
+        Ok(())
     }
 
     #[test]
@@ -603,8 +613,10 @@ mod integration_tests {
         ]);
         http_smoke_test_template(
             "http-py",
-            Some("https://github.com/spinframework/spin-python-sdk"),
-            Some("v2.0"),
+            Some(crate::testcases::TemplateSource::Url {
+                repo: "https://github.com/spinframework/spin-python-sdk",
+                branch: Some("v2.0"),
+            }),
             &[],
             prebuild,
             env_vars,
@@ -617,7 +629,6 @@ mod integration_tests {
     fn http_c_template_smoke_test() -> anyhow::Result<()> {
         http_smoke_test_template(
             "http-c",
-            None,
             None,
             &[],
             |_| Ok(()),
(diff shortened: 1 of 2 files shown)


itowlson:
Okay, the errors are not very lovely but they do happen and it does tell you whether your PR made local or main emit smoke.  This is from a previous run of this PR (https://github.com/spinframework/spin/actions/runs/16280659281/job/45969436053) where I deliberately broke the Go template (also tried with Rust):

```
test integration_tests::http_go_template_smoke_test ... FAILED
Error: local template

Caused by:
    'cd "/tmp/t77ee-b" && CARGO_TARGET_DIR="/tmp/t77ee-b/target" PATH="target/debug:target/release:/home/runner/.cargo/bin:/home/runner/work/spin/spin:/opt/hostedtoolcache/node/22.17.0/x64/bin:/opt/hostedtoolcache/binaryen/116/x86_64/binaryen-version_116/bin:/opt/hostedtoolcache/tinygo/0.37.0/amd64/tinygo/bin:/home/runner/go/bin:/opt/hostedtoolcache/go/1.23.10/x64/bin:/home/runner/work/_temp/0c3644da-bccf-4298-b5fd-cafd82c90f37/zig-x86_64-linux-0.14.1:/snap/bin:/home/runner/.local/bin:/opt/pipx_bin:/home/runner/.cargo/bin:/home/runner/.config/composer/vendor/bin:/usr/local/.ghcup/bin:/home/runner/.dotnet/tools:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/home/runner/work/spin/spin/target/release" "/home/runner/work/spin/spin/target/release/spin" "build"' failed with status code Some(1)
    stdout:
    Building component test with `tinygo build -target=wasip1 -buildmode=c-shared -no-debug -o main.wasm .`
    
    stderr:
    # github.com/test
Error:     main.go:13:5: w.Headdddddddddddddder undefined (type "net/http".ResponseWriter has no field or method Headdddddddddddddder)
    Error: Build command for component test failed with status Exited(1)
```

I've only done Go and Rust because the JS and Python ones aren't in this repo, and the other languages are relatively small and not seeing a lot of updates at the moment.  If we want to do more templates then I can do so.

vdice (approved):

PR #3200, opened 2025-07-15
Remove WASI P1 compatibility check tooling

itowlson:
This removes the modsurfer package which could be used to check that WASI Preview 1 modules had the correct ABI.  The Spin world has moved on a lot and is now defined in terms of the Component Module and WASI Preview 2.


adamreese:
The only other mention I found was https://www.fermyon.com/blog/introducing-spin-v11#other-improvements

fibonacci1729 (approved):

PR #3201, opened 2025-07-17
added support for otlp headers

### LINKED ISSUE #2811: Support for OTEL_EXPORTER_OTLP_HEADERS/OTEL_EXPORTER_OTLP_{TRACE,LOG,METRICS}_HEADERS
Some managed providers require specifying custom headers for managing authentication when shipping telemetry data without a local relaying agent (e.g. Honeycomb) - we should also support that with Spin. (afaict we don't currently plumb those up anywhere when looking in https://github.com/fermyon/spin/blob/ba32722671752321ab7a34b52216559aacacfa51/crates/telemetry/src/env.rs)

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#specifying-headers-via-environment-variables

seun-ja:
Fixes #2811

### DIFF
--- a/crates/telemetry/src/propagation.rs
+++ b/crates/telemetry/src/propagation.rs
@@ -4,9 +4,15 @@ use opentelemetry::{
 };
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
+use crate::env::OtelExporterOtlpHeaders;
+
 /// Injects the current W3C TraceContext into the provided request.
 pub fn inject_trace_context<'a>(req: impl Into<HeaderInjector<'a>>) {
     let mut injector = req.into();
+    for (key, value) in OtelExporterOtlpHeaders::headers() {
+        injector.set(&key, value);
+    }
+
     global::get_text_map_propagator(|propagator| {
         let context = tracing::Span::current().context();
         propagator.inject_context(&context, &mut injector);
--- a/crates/telemetry/src/env.rs
+++ b/crates/telemetry/src/env.rs
@@ -1,9 +1,10 @@
-use std::env::VarError;
+use std::{collections::HashMap, env::VarError};
 
 use opentelemetry_otlp::{
-    OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_EXPORTER_OTLP_LOGS_ENDPOINT,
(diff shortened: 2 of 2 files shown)


lann:
Did anyone try this _without_ any code changes? I think the opentelemetry crate already checks _some_ of the standard env vars...

lann:
e.g. https://github.com/open-telemetry/opentelemetry-rust/blob/189078d7a92e534d67137ffaed8e35456eaa589b/opentelemetry-otlp/src/exporter/tonic/mod.rs#L343

seun-ja:
Yeah, looks like they all are checked. Judging by the code base.

seun-ja:
@lann would you suggest #2811 be closed?

crates/telemetry/src/propagation.rs
```diff
@@ -4,9 +4,15 @@ use opentelemetry::{
...
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
+use crate::env::OtelExporterOtlpHeaders;
+
 /// Injects the current W3C TraceContext into the provided request.
 pub fn inject_trace_context<'a>(req: impl Into<HeaderInjector<'a>>) {
     let mut injector = req.into();
+    for (key, value) in OtelExporterOtlpHeaders::headers() {
+        injector.set(&key, value);
+    }
```
calebschoepp, line 14: We wouldn't want to inject the headers here. This is for propagating context to outgoing requests. `OTEL_EXPORTER_OTLP_TRACES_HEADERS` is for setting headers on requests made to the OTel collector to export traces.

calebschoepp:
@seun-ja the next step here would be to run Spin off main and see if it respects these env vars already. You could point Spin at a collector that is actually just a [proxy](https://mitmproxy.org/) to see if the headers are being set or not.

seun-ja:
It respects it, I ran it on main and was able to see the key-pair data parsed into the header.

sample env data provided:
```env
export OTEL_EXPORTER_OTLP_TRACES_HEADERS="authorization=Bearer trace-token,x-trace-id=abc123"
```

result:
Screenshot 2025-08-12 at 3 45 38 PM

Apologies for the late response

cc: @calebschoepp

PR #3203, opened 2025-07-18
added flag for clearing log files at restart

### LINKED ISSUE #3130: Clear log files when re-building locally
If I am rebuilding my application via `spin build --up` to rebuild locally, the logs don't get cleared when the server restarts, which is annoying. By default they should be cleared, because I want to simulate how the server reacts on a new deployment. 

Maybe there can be a config flag for users who really want this behavior, but by default I'd expect the logs to clear when rebuilding and re-starting the sever.

seun-ja:
Aims to fix #3130

### DIFF
--- a/crates/runtime-config/src/lib.rs
+++ b/crates/runtime-config/src/lib.rs
@@ -99,6 +99,7 @@ where
         local_app_dir: Option<PathBuf>,
         provided_state_dir: UserProvidedPath,
         provided_log_dir: UserProvidedPath,
+        refresh_logs: bool,
     ) -> anyhow::Result<Self> {
         let toml = match runtime_config_path {
             Some(runtime_config_path) => {
@@ -117,8 +118,13 @@ where
             }
             None => Default::default(),
         };
-        let toml_resolver =
-            TomlResolver::new(&toml, local_app_dir, provided_state_dir, provided_log_dir);
+        let toml_resolver = TomlResolver::new(
+            &toml,
+            local_app_dir,
+            provided_state_dir,
+            provided_log_dir,
+            refresh_logs,
+        );
 
         Self::new(toml_resolver, runtime_config_path)
     }
@@ -192,6 +198,8 @@ pub struct TomlResolver<'a> {
     state_dir: UserProvidedPath,
     /// Explicitly provided log directory.
     log_dir: UserProvidedPath,
+    /// Whether to refresh logs when restarted.
+    refresh_logs: bool,
 }
 
 impl<'a> TomlResolver<'a> {
@@ -201,12 +209,14 @@ impl<'a> TomlResolver<'a> {
         local_app_dir: Option<PathBuf>,
         state_dir: UserProvidedPath,
         log_dir: UserProvidedPath,
+        refresh_logs: bool,
     ) -> Self {
         Self {
             table: TomlKeyTracker::new(table),
             local_app_dir,
             state_dir,
             log_dir,
+            refresh_logs,
         }
     }
 
@@ -268,7 +278,17 @@ impl<'a> TomlResolver<'a> {
         }
 
         match log_dir {
+            UserProvidedPath::Provided(p) if self.refresh_logs => Ok(Some({
+                let _ = std::fs::remove_dir_all(&p);
+
+                std::path::absolute(p)?
+            })),
             UserProvidedPath::Provided(p) => Ok(Some(std::path::absolute(p)?)),
+            UserProvidedPath::Default if self.refresh_logs => Ok(self.state_dir()?.map(|p| {
+                let _ = std::fs::remove_dir_all(&p);
+
+                p.join("logs")
+            })),
             UserProvidedPath::Default => Ok(self.state_dir()?.map(|p| p.join("logs"))),
             UserProvidedPath::Unset => Ok(None),
         }
@@ -619,6 +639,7 @@ mod tests {
             None,
             UserProvidedPath::Default,
             UserProvidedPath::Default,
+            false,
         )
     }
 
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -27,6 +27,7 @@ pub use stdio::StdioLoggingExecutorHooks;
 pub use summary::{KeyValueDefaultStoreSummaryHook, SqliteDefaultStoreSummaryHook};
 
 pub const APP_LOG_DIR: &str = "APP_LOG_DIR";
+pub const REFRESH_LOGS: &str = "REFRESH_LOGS";
 pub const DISABLE_WASMTIME_CACHE: &str = "DISABLE_WASMTIME_CACHE";
 pub const FOLLOW_LOG_OPT: &str = "FOLLOW_ID";
 pub const WASMTIME_CACHE_FILE: &str = "WASMTIME_CACHE_FILE";
@@ -53,6 +54,14 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
         env = "SPIN_LOG_DIR",
     )]
     pub log: Option<PathBuf>,
+
+    /// Whether to refresh logs when restarted.
+    #[clap(
+        name = REFRESH_LOGS,
+        short = 'r',
+        long = "refresh-logs",
+    )]
+    pub refresh_logs: bool,
 
     /// Disable Wasmtime cache.
     #[clap(
@@ -139,6 +148,8 @@ pub struct FactorsConfig {
     pub follow_components: FollowComponents,
     /// Log directory for component stdout/stderr.
     pub log_dir: UserProvidedPath,
+    /// Whether to refresh logs when restarted.
+    pub refresh_logs: bool,
 }
 
 /// An empty implementation of clap::Args to be used as TriggerExecutor::RunConfig
@@ -220,6 +231,7 @@ impl<T: Trigger<B::Factors>, B: RuntimeFactorsBuilder> FactorsTriggerCommand<T, 
             local_app_dir: local_app_dir.clone(),
             follow_components,
             log_dir,
+            refresh_logs: self.refresh_logs,
         };
 
         let run_fut = builder
--- a/crates/runtime-factors/src/build.rs
+++ b/crates/runtime-factors/src/build.rs
@@ -28,6 +28,7 @@ impl RuntimeFactorsBuilder for FactorsBuilder {
             config.local_app_dir.clone().map(PathBuf::from),
             config.state_dir.clone(),
             config.log_dir.clone(),
+            config.refresh_logs,
         )?;
 
         runtime_config.summarize(config.runtime_config_file.as_deref());


crates/trigger/src/cli.rs
```diff
@@ -54,6 +55,14 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
     )]
     pub log: Option<PathBuf>,
 
+    /// Whether to refresh logs when restarted.
```
itowlson: I found the word "refresh" a bit odd here, although I can see the connection when I squint.  At least in help text, I think we should be more explicit about what the flag does, e.g.

```suggestion
    /// If set, Spin deletes the log files before starting the application.
```

(or 'clears the log files'?  Not sure which is better)

crates/trigger/src/cli.rs
```diff
@@ -54,6 +55,14 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
...
+    #[clap(
+        name = REFRESH_LOGS,
+        short = 'r',
```
itowlson: I wouldn't spend a valuable shortcode on this.

crates/runtime-config/src/lib.rs
```diff
@@ -192,6 +198,8 @@ pub struct TomlResolver<'a> {
     state_dir: UserProvidedPath,
     /// Explicitly provided log directory.
     log_dir: UserProvidedPath,
+    /// Whether to refresh logs when restarted.
+    refresh_logs: bool,
```
itowlson: A bit strange to see this in a struct called TomlResolver - it's not like it has anything to do with resolving TOML.  But I guess maybe the log directory is already non-TOML-related and the struct has just grown beyond its original name?  Not sure.

crates/runtime-config/src/lib.rs
```diff
@@ -268,7 +278,17 @@ impl<'a> TomlResolver<'a> {
         }
 
         match log_dir {
+            UserProvidedPath::Provided(p) if self.refresh_logs => Ok(Some({
+                let _ = std::fs::remove_dir_all(&p);
```
itowlson: This seems to end up duplicating the check and the deletion code.  It seems like something you could do outside the `match`, if the `match` successfully resolves a path.

But I see this is in a function whose job is to "get the configured log directory."  Getting path to a directory should _not_ have a potential side effect of deleting that directory - that's super surprising behaviour.

(Similarly, notice that this function does not _create_ the directory.  It just gets the path.)

So I think this code needs to move somewhere else, at which point hopefully we will know where the log directory is (if we have one) and can clear it without resorting to repeated guards or code.

crates/trigger/src/cli.rs
```diff
@@ -54,6 +55,14 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
     )]
     pub log: Option<PathBuf>,
 
+    /// Whether to refresh logs when restarted.
+    #[clap(
+        name = REFRESH_LOGS,
```
itowlson: Maybe consider a `SPIN_` prefix for this?  Going by the original issue, some folks will be treating this as a user preference and setting it in `.profile` rather than on the `spin up` command line, so it would be good to have it make sense out of the Spin context perhaps?

(sorry forgot to think of this while reviewing)

crates/runtime-config/src/lib.rs
seun-ja: Somewhat, yes. 

All but `refresh_logs` still source the default value from `table: TomlKeyTracker<'a>`, so the initial name makes sense. However, with other elements added, a change of name would be better, something like `RuntimeConfigurationResolver`. Would that be fine, or what would you suggest?

itowlson: I should have come back to this comment after writing the one about the location of the delete code - sorry.

We should first figure out a good place for the delete code to live, whether in this struct impl or elsewhere.  Then that will guide us as to whether this flag needs to live here or not.

crates/runtime-config/src/lib.rs
```diff
@@ -147,6 +143,11 @@ where
 
         let toml = toml_resolver.toml();
         let log_dir = toml_resolver.log_dir()?;
+
+        if refresh_logs {
+            let _ = Self::resolve_log_dir(&log_dir);
```
itowlson: I'm sorry to be a nag but once again this is kind of jamming the delete code in somewhere that yes, it's convenient right now because this code knows what the log directory is, but will be surprising to anyone debugging in a few months.  This is the _constructor_ for the runtime configuration.  It has nothing to do with creating or deleting directories.

What I really think we should do before diving into writing code is some design:

* What is the desired behaviour of the feature?
* What considerations do we have to think about when implementing the feature?
* What code object(s) are responsible for the behaviour?

Let me give some examples.

* The current behaviour is to delete the logs directory.  Suppose, for some reason known only to myself, I set the logs directory to be my application directory (`spin up --log-dir . --refresh-logs`).  Oops, Spin just deleted my application directory.  Should the behaviour be to delete individual log files rather than the whole directory?
* The current implementation deletes the directory as part of trigger startup.  That means that if I have a multi-trigger application, each trigger will perform a directory delete as it starts.  Suppose one trigger fails early, and another trigger starts slowly.  The second trigger blew away my logs for the failure.
* The existing log file creation happens in `ComponentStdioWriter`, in the `spin_trigger::cli::stdio` module.  Does deleting existing logs fit nicely with the existing responsibilities of that type or module?  I'm not saying it does, I'm just giving it as an example of a place I'd look at before the runtime-config subsystem.

I don't know if there are other factors to consider because I've not done any design on this, I just feel like this might run smoother if you were gathering feedback on a holistic proposal rather than trying to make local tweaks in response to a stream of comments.  That said, this is your project and if you feel more comfortable working at the code level then no worries!

seun-ja: Taking a further look at this issue in the broader context (apologies for not doing this earlier), I am struggling to find an intersection between `up`'s subsequent calls/methods and `trigger`'s subcommand where `refresh-log` and `log-dir` can be accessed.

Here's an approach I'm considering
- Ideally, the refresh should be done before the triggers are started when the up command is called
https://github.com/seun-ja/spin/blob/9d09742a860ffea180d3eadd547424c41b577312/src/commands/up.rs#L239
- handling it in ComponentStdioWriter could have been idle, as it has a connection to trigger commands, it wouldn't work because it is called on every request, hence if placed there, it deletes and recreates the file, not ideal

Would you be open to a pair?

itowlson: Good catch on `ComponentStdioWriter` and sorry for the bad steer.  I do think that module as a whole will repay attention.  `StdioLoggingExecutorHooks` is the other half of the equation, and that does get to participate in trigger startup.  Looking at other hooks (e.g. `InitialKvSetter`), `configure_app` is where they do their "do once at startup" work - we'd have to figure out the list of log files relevant to the trigger, and I'm not sure if the `ConfiguredApp` is pared back to just the trigger components at this point.  But testing will clarify that (or more code reading!).

crates/trigger/src/cli/stdio.rs
```diff
@@ -328,3 +369,11 @@ fn bullet_list<S: std::fmt::Display>(items: impl IntoIterator<Item = S>) -> Stri
         .collect::<Vec<_>>()
         .join("\n")
 }
+
+fn santized_log_path(
```
itowlson:
```suggestion
fn sanitized_log_path(
```

crates/trigger/src/cli/stdio.rs
```diff
@@ -95,6 +103,39 @@ impl<F: RuntimeFactors, U> ExecutorHooks<F, U> for StdioLoggingExecutorHooks {
...
+                .collect();
+
+            for sanitized_component_id in sanitized_component_ids {
+                if let Some(stdout_log_path) = santized_log_path(
+                    self.log_dir.as_deref(),
+                    &sanitized_component_id,
+                    STDOUT_LOG_FILE_SUFFIX,
+                ) {
+                    if let Err(err) = std::fs::File::create(stdout_log_path) {
+                        //TODO: figure out if we want to return error here
```
itowlson: As you have it now, `create` _will_ fail if the logs dir doesn't already exist, and that shouldn't be considered an error or warning.  But you might want to move it to after the "create the log directory" bit so that it doesn't fail...

crates/trigger/src/cli/stdio.rs
```diff
@@ -95,6 +103,39 @@ impl<F: RuntimeFactors, U> ExecutorHooks<F, U> for StdioLoggingExecutorHooks {
         configured_app: &spin_factors::ConfiguredApp<F>,
     ) -> anyhow::Result<()> {
         self.validate_follows(configured_app.app())?;
+
+        if self.refresh_log {
```
itowlson: This is a very long block which now dominates the `configure_app` function.  Consider breaking out e.g. `if self.refresh_log { self.truncate_log_files(); }`

crates/trigger/src/cli/stdio.rs
```diff
@@ -95,6 +103,39 @@ impl<F: RuntimeFactors, U> ExecutorHooks<F, U> for StdioLoggingExecutorHooks {
...
+            let sanitized_component_ids: Vec<String> = configured_app
+                .app()
+                .components()
```
itowlson: Did we confirm that `.components()` returns _only_ the components being run by _this_ trigger?

crates/trigger/src/cli.rs
```diff
@@ -54,6 +55,13 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
     )]
     pub log: Option<PathBuf>,
 
+    /// If set, Spin deletes the log files before starting the application.
+    #[clap(
+        name = SPIN_REFRESH_LOGS,
+        long = "refresh-logs",
```
itowlson: I'd really like to align on the language here.  Is the consensus for "refresh logs"?  It feels ambiguous to me and I'd prefer "clear logs" or "truncate logs" or something.  (But if we do make a change then the language needs to line up everywhere, not just on this line - field names, EVs, docs, etc.)

crates/trigger/src/cli.rs
```diff
@@ -54,6 +55,13 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
     )]
     pub log: Option<PathBuf>,
 
+    /// If set, Spin deletes the log files before starting the application.
```
itowlson: Given the proposed implementation, consider e.g.

```suggestion
    /// If set, Spin clears the log files before starting the application.
```

(since we no longer delete them)

crates/trigger/src/cli.rs
```diff
@@ -139,6 +147,8 @@ pub struct FactorsConfig {
     pub follow_components: FollowComponents,
     /// Log directory for component stdout/stderr.
     pub log_dir: UserProvidedPath,
+    /// If set, Spin deletes the log files before starting the application.
```
itowlson:
```suggestion
    /// If set, Spin clears the log files before starting the application.
```

itowlson (commented):
Thanks for this - it looks good!  I had a couple of style and naming nits, which are non-blockers but I'd be keen to see in; but my only remaining functional concern is the multi-trigger-type scenario.  If you could validate the each trigger only truncates _its own_ log files then that would be very reassuring!

Indeed, generally, since this is hard to write unit tests for, it would also be very useful to have some notes in the PR about what you tested and how.

Thanks again for persevering with this!

crates/trigger/src/cli/stdio.rs
seun-ja: Yes, returns only this trigger

itowlson: Excellent!

crates/trigger/src/cli/stdio.rs
```diff
@@ -86,6 +94,35 @@ impl StdioLoggingExecutorHooks {
...
+            .map(|c| sanitize_filename::sanitize(c.locked.id.clone()))
+            .collect();
+
+        for sanitized_component_id in sanitized_component_ids {
+            if let Some(stdout_log_path) = sanitized_log_path(
+                self.log_dir.as_deref(),
+                &sanitized_component_id,
+                STDOUT_LOG_FILE_SUFFIX,
+            ) {
+                let _ = std::fs::File::create(stdout_log_path);
```
itowlson: I think you don't need the `let` here - the `_` suffices as a discard indicator.
```suggestion
                _ = std::fs::File::create(stdout_log_path);
```

(I may be mis-remembering though)

crates/trigger/src/cli/stdio.rs
```diff
@@ -86,6 +94,35 @@ impl StdioLoggingExecutorHooks {
...
+            }
+
+            if let Some(stderr_log_path) = sanitized_log_path(
+                self.log_dir.as_deref(),
+                &sanitized_component_id,
+                STDERR_LOG_FILE_SUFFIX,
+            ) {
+                let _ = std::fs::File::create(stderr_log_path);
```
itowlson: same comment (and same disclaimer!)

itowlson (approved):
Looks good - thanks!  It would be great if you could write something about testing but the code looks fine!

crates/trigger/src/cli/stdio.rs
```diff
@@ -86,6 +95,24 @@ impl StdioLoggingExecutorHooks {
             _ => Ok(()),
         }
     }
+
+    fn truncate_log_files(&self, log_dir: &Option<PathBuf>) {
```
itowlson: This truncates all log-like files in the directory - as discussed before, this potentially runs into a problem with multi-trigger if one trigger process emits a log while the other is still starting.

Maybe that's not a consideration, we probably have race conditions in multi-trigger already (like KV init), but I'd definitely seek a second opinion...

crates/trigger/src/cli/stdio.rs
```diff
@@ -95,11 +122,16 @@ impl<F: RuntimeFactors, U> ExecutorHooks<F, U> for StdioLoggingExecutorHooks {
...
     ) -> anyhow::Result<()> {
         self.validate_follows(configured_app.app())?;
+
         if let Some(dir) = &self.log_dir {
             // Ensure log dir exists if set
             std::fs::create_dir_all(dir)
                 .with_context(|| format!("Failed to create log dir {}", quoted_path(dir)))?;
 
+            if self.truncate_log {
+                self.truncate_log_files(&self.log_dir);
```
itowlson: You can pass `dir` here and save the redundant pattern match in `truncate_log_files`.

crates/trigger/src/cli/stdio.rs
lann: We have a race condition in a _single_ trigger; each instance opens a separate handle to the log file and writes to it without any locking. In the context of `spin up` this behavior seems fine to me.

crates/trigger/src/cli/stdio.rs
```diff
@@ -86,6 +95,23 @@ impl StdioLoggingExecutorHooks {
...
+            for entry in entries.flatten() {
+                let path = entry.path();
+                let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
+                    continue;
+                };
+
+                if name.ends_with(&format!("{STDOUT_LOG_FILE_SUFFIX}.txt"))
+                    || name.ends_with(&format!("{STDERR_LOG_FILE_SUFFIX}.txt"))
+                {
+                    _ = std::fs::File::create(log_dir.join(name))
```
itowlson: Could this use the `path` variable (set on line 102) instead of regenerating the path from a `join`?

crates/trigger/src/cli/stdio.rs
```diff
@@ -86,6 +95,23 @@ impl StdioLoggingExecutorHooks {
             _ => Ok(()),
         }
     }
+
+    fn truncate_log_files(&self, log_dir: &Path) {
```
itowlson: Does this use the `&self` argument?

crates/trigger/src/cli/stdio.rs
```diff
@@ -86,6 +95,23 @@ impl StdioLoggingExecutorHooks {
...
+        if let Ok(entries) = log_dir.read_dir() {
+            for entry in entries.flatten() {
+                let path = entry.path();
+                let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
```
itowlson, line 103: Does this need the `file_name()`?  It seems like the `ends_with` tests would work just as well on the full path.

itowlson (commented):
Looks good and thanks Lann for the second opinion on the potential race.  Happy to go with this approach but I'm afraid I have yet a few more comments on the implementation details...

crates/trigger/src/cli/stdio.rs
seun-ja: Yes, it does. Apparently, `ends_with` under `path()` only checks path segment equality as compared to `file_path()` which checks str match

itowlson (approved):

PR #3206, opened 2025-07-22
Ensure valid wasms when publishing app

fibonacci1729:
This PR ensures that each `source` field and depenency source specified in a Spin component is a valid Wasm binary.

### DIFF
--- a/crates/oci/src/validate.rs
+++ b/crates/oci/src/validate.rs
@@ -1,0 +1,166 @@
+use anyhow::{bail, Context, Result};
+use spin_common::{ui::quoted_path, url::parse_file_url};
+use spin_locked_app::locked::{LockedComponent, LockedComponentSource};
+
+/// Validate that all Spin components specify valid wasm binaries in both the `source`
+/// field and for each dependency.
+pub async fn ensure_wasms(component: &LockedComponent) -> Result<()> {
+    // Ensure that the component source is a valid wasm binary.
+    let bytes = read_component_source(&component.source).await?;
+    if !is_wasm_binary(&bytes) {
+        bail!(
+            "Component {} source is not a valid .wasm file",
+            component.id,
+        );
+    }
+
+    // Ensure that each dependency is a valid wasm binary.
+    for (dep_name, dep) in &component.dependencies {
+        let bytes = read_component_source(&dep.source).await?;
+        if !is_wasm_binary(&bytes) {
+            bail!(
+                "dependency {} for component {} is not a valid .wasm file",
+                dep_name,
+                component.id,
+            );
+        }
+    }
+    Ok(())
+}
+
+fn is_wasm_binary(bytes: &[u8]) -> bool {
+    wasmparser::Parser::is_component(bytes)
+        || wasmparser::Parser::is_core_wasm(bytes)
+        || wat::parse_bytes(bytes).is_ok()
+}
+
+async fn read_component_source(source: &LockedComponentSource) -> Result<Vec<u8>> {
+    let source = source
+        .content
+        .source
+        .as_ref()
+        .context("LockedComponentSource missing source field")?;
+
+    let path = parse_file_url(source)?;
+
+    let bytes: Vec<u8> = tokio::fs::read(&path).await.with_context(|| {
+        format!(
+            "failed to read component source from disk at path {}",
+            quoted_path(&path)
+        )
+    })?;
+    Ok(bytes)
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::from_json;
+    use spin_locked_app::locked::LockedComponent;
+    use tokio::io::AsyncWriteExt;
+
+    #[tokio::test]
+    async fn ensures_valid_wasm_binaries() {
+        let working_dir = tempfile::tempdir().unwrap();
+
+        macro_rules! make_locked {
+            ($source:literal, $($dep_name:literal=$dep_path:literal),*) => {
+                from_json!({
+                    "id": "jiggs",
+                    "source": {
+                        "content_type": "application/wasm",
+                        "source": format!("file://{}", working_dir.path().join($source).to_str().unwrap()),
+                        "digest": "digest",
+                    },
+                    "dependencies": {
+                        $(
+                            $dep_name: {
+                                "source": {
+                                    "content_type": "application/wasm",
+                                    "source": format!("file://{}", working_dir.path().join($dep_path).to_str().unwrap()),
+                                    "digest": "digest",
+                                },
+                            }
+                        ),*
+                    }
+                })
+            };
+        }
+
+        let make_file = async |name, content| {
+            let path = working_dir.path().join(name);
+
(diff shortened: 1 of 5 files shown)


crates/oci/src/validate.rs
```diff
@@ -0,0 +1,193 @@
...
+mod test {
+    use super::*;
+    use crate::from_json;
+    use spin_locked_app::locked::LockedComponent;
+    use tokio::io::AsyncWriteExt;
+
+    #[tokio::test]
+    async fn ensures_valid_wasm_binaries() {
+        let working_dir = tempfile::tempdir().unwrap();
+        let _ = tokio::fs::create_dir(working_dir.path()).await;
```
itowlson: Is this needed?  I thought `tempdir()` created the temporary directory.

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        // valid component source
```
itowlson: The preamble for creating test files consumes a lot of space which, for me, somewhat obscures the actual test part of the test.  I wonder if we could break out a helper so that this could be condensed down to e.g.

```
let make_file = |name, content| { ... }  // closure would let us close over working_dir. or wrapper struct + method

let valid_wasm_component_file = make_file("component.wasm", b"...").await;
// etc.
```

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,193 @@
...
+            },
+            TestCase {
+                name: "Valid Spin component source field with valid wasm dependency",
+                locked_component: from_json!({
+                    "id": "component",
+                    "source": {
+                        "content_type": "application/wasm",
+                        "source": format!("file://{}", working_dir.path().join("component.wasm").to_str().unwrap()),
+                        "digest": "digest",
+                    },
```
itowlson: Again, there's a lot of repetition in the test case construction which obscures what is actually being tested - I'm afraid my eyes slid off this rather!  What if e.g.

```
// let make_locked = |wasm_file, deps_wasm_files| { ... }

TestCase {
  name: "biscuits",
  locked_component: make_locked("component.wasm", &["component.wasm"],
  valid: true,
}
```

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,193 @@
...
+                "dependency {} for component {} is not a valid .wasm file",
+                dep_name,
+                component.id,
+            );
+        }
+    }
+    Ok(())
+}
+
+fn is_wasm_binary(bytes: &[u8]) -> bool {
```
itowlson, line 31: Just to confirm given the function name - I _think_ we may actually allow WAT text as well as the binary encoding - `wasmparser::Parser` will handle that correctly anyway right?

(That said, I don't think we document allowing WAT, and I don't think I've ever tried it, so it's not a big deal either way...)

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,193 @@
+use anyhow::{bail, Context, Result};
+use spin_common::{ui::quoted_path, url::parse_file_url};
+use spin_locked_app::locked::{LockedComponent, LockedComponentSource};
+
+// Validate that all Spin components specify valid wasm binaries in both the `source`
```
itowlson: channelling my inner Ryan to propose that these might as well be doc comments

itowlson (approved):
The actual code that does the actual thing looks fine, but I did have a good old moan about the tests I'm afraid.  Not a blocker but would be nice to tighten them up if you have the bandwidth.

crates/oci/src/validate.rs
fibonacci1729: Good catch.

fibonacci1729: I'm not sure if `*.wat` is supported currently but i'll include the test anyway.

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,174 @@
...
+        // valid module source using WAT
+        make_file("module.wat", b"(module)").await;
+        // valid component source
+        make_file("component.wasm", b"\x00\x61\x73\x6D\x0D\x00\x01\x00").await;
+        // valid core module source
+        make_file("module.wasm", b"\x00\x61\x73\x6D\x01\x00\x00\x00").await;
+        // invalid wasm binary
+        make_file("invalid.wasm", b"not a wasm file").await;
+
+        macro_rules! make_locked {
```
itowlson, line 66: nit: I wonder if it would be good to move this (and `TestCase`?) somewhere else in the function so they don't break the top-to-bottom flow so much.  Definitely not a biggie though

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,174 @@
...
+        #[derive(Clone)]
+        struct TestCase {
+            name: &'static str,
+            locked_component: LockedComponent,
+            valid: bool,
+        }
+
+        let tests: Vec<TestCase> = vec![
+            TestCase {
+                name: "Valid Spin component source field with component WAT",
```
itowlson: nit: consider removing `field` from these names (and `source`??) - to me it doesn't feel like it adds anything

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,174 @@
...
+                name: "Valid Spin component source field with core module",
+                locked_component: make_locked!("module", "module.wasm",),
+                valid: true,
+            },
+            TestCase {
+                name: "Valid Spin component source field with valid wasm dependency",
+                locked_component: make_locked!(
+                    "component",
+                    "component.wasm",
+                    "test:comp2" = "component.wasm"
```
itowlson: OH NICE.  Much better than mine!

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,174 @@
...
+        make_file("component.wasm", b"\x00\x61\x73\x6D\x0D\x00\x01\x00").await;
+        // valid core module source
+        make_file("module.wasm", b"\x00\x61\x73\x6D\x01\x00\x00\x00").await;
+        // invalid wasm binary
+        make_file("invalid.wasm", b"not a wasm file").await;
+
+        macro_rules! make_locked {
+            ($id:literal, $source:literal, $($dep_name:literal=$dep_path:literal),*) => {
+                from_json!({
+                    "id": $id,
```
itowlson: It doesn't look like we display the ID directly, nor the error which contains the ID - how about simplifying the call site by hardwiring the ID to "component" or "alleged-component" or something?

crates/oci/src/validate.rs
```diff
@@ -0,0 +1,166 @@
...
+    use tokio::io::AsyncWriteExt;
+
+    #[tokio::test]
+    async fn ensures_valid_wasm_binaries() {
+        let working_dir = tempfile::tempdir().unwrap();
+
+        macro_rules! make_locked {
+            ($source:literal, $($dep_name:literal=$dep_path:literal),*) => {
+                from_json!({
+                    "id": "jiggs",
```
itowlson, line 69: YES

PR #3213, opened 2025-08-05
Spin up should check required variables

### LINKED ISSUE #3115: `spin up` should check required variables
Currently, the `spin up` command does not check if **required** variables are provided when launching the application.

Consider a Spin application with the following manifest:

```toml
spin_manifest_version = 2

[application]
name = "test"
version = "0.1.0"
description = ""

[variables]
foo = { required = true }

[[trigger.http]]
route = "/..."
component = "test"

[component.test]
source = "target/wasm32-wasip1/release/test.wasm"
allowed_outbound_hosts = []

[component.test.variables]
foo = "{{foo}}"

[component.test.build]
```

```rust
use spin_sdk::http::{IntoResponse, Request, Response};
use spin_sdk::{http_component, variables};
#[http_component]
fn handle_test_var(_: Request) -> anyhow::Result<impl IntoResponse> {
    match variables::get("foo") {
        Ok(value) => println!("{}", value),
        Err(r) => println!("{}", r),
    };
    Ok(Response::new(200, ()))
}
```

## Expected behavior

I expect `spin up` to bail out and prevent from launching the app - if the required variable `foo` is not specified (e.g., by setting the `SPIN_VARIABLE_FOO` environment variable)

## Actual behavior

I can start the app locally using `spin up`. Sending a request to the app (which tries to load the `foo` variable will result in a runtime error:

```
Error::Provider("no provider resolved required variable \"foo\"")
```

## Spin Version

I see this behavior with current spin version (tested `3.0.0`, `3.1.2` and `3.2.0`)

seun-ja:
Closes #3115 

The PR checks if the variable default isn't provided when required. The check is handled by a new hook: `VariableSorterExecutorHooks`.

Inspired by @itowlson's [comment](https://github.com/spinframework/spin/issues/3115#issuecomment-3272520127), new metadata has been added to `Provider`s.


crates/loader/src/local.rs
```diff
@@ -135,6 +136,21 @@ impl LocalLoader {
         })
     }
 
+    fn env_checker((key, val): (String, Variable)) -> anyhow::Result<(String, Variable)> {
```
itowlson: This function name needs to be more descriptive.  What is it checking?

itowlson: Not clear why this takes a tuple rather than two parameters?

crates/loader/src/local.rs
```diff
@@ -88,7 +89,7 @@ impl LocalLoader {
 
         let variables = variables
             .into_iter()
-            .map(|(name, v)| Ok((name.to_string(), locked_variable(v)?)))
+            .map(|(name, v)| Self::env_checker((name.to_string(), locked_variable(v)?)))
```
itowlson: Doing this here means appears to mean we can't even build the LockedApp without having the variables available.  How does this worth with e.g. `spin registry push` or `spin cloud deploy`?

crates/loader/src/local.rs
```diff
@@ -135,6 +136,21 @@ impl LocalLoader {
         })
     }
 
+    fn env_checker((key, val): (String, Variable)) -> anyhow::Result<(String, Variable)> {
+        if val.default.is_none() {
+            if std::env::var(env_key(None, key.as_ref())).is_err() {
```
itowlson: You can't assume that variables will come from the host environment.  They could come from Vault or Azure KeyVault - for example passwords or tokens would typically be stashed in vaults and it would be rude to demand users create dummy EVs for them.  Variable sources are defined in the runtime config file - you can't validate them until you have the runtime config providers lined up.

itowlson: How do you know that the user is using the default prefix?

itowlson: The environment provider appears to support a `dotenv_path` setting to merge values from a file - not clear that this is respected here?

crates/loader/src/local.rs
itowlson: Maybe a nit, but doing this in the loader produces an error message that implies we can't understand the manifest (usually meaning invalid TOML or incorrect schema): that's potentially misleading.

itowlson (changes requested):

karthik2804:
I think adding a new hook before execution of the CLI for the variables factor in [crates/trigger/src/cl](https://github.com/spinframework/spin/tree/main/crates/trigger/src/cli) as discussed offline earlier might be a better place to run the check.

seun-ja:
Interestingly, the test failing in CI doesn't fail locally. Investigating it at the moment.

Cargo.toml
```diff
@@ -64,12 +64,13 @@ spin-oci = { path = "crates/oci" }
 spin-plugins = { path = "crates/plugins" }
 spin-runtime-factors = { path = "crates/runtime-factors" }
 spin-telemetry = { path = "crates/telemetry", features = [
-  "tracing-log-compat",
+    "tracing-log-compat",
```
itowlson: There's a few of these unrelated layout changes - maybe turn off auto formatting for this file?

crates/factors-test/src/lib.rs
```diff
@@ -100,6 +101,8 @@ impl<T: RuntimeFactors> TestEnvironment<T> {
 pub async fn build_locked_app(manifest: &toml::Table) -> anyhow::Result<LockedApp> {
     let toml_str = toml::to_string(manifest).context("failed serializing manifest")?;
     let dir = tempfile::tempdir().context("failed creating tempdir")?;
+    // `foo` variable is set to require. As we're not providing a default value, env is checked.
+    std::env::set_var(env_key(None, "foo"), "baz");
```
itowlson: I still say we should be able to transform a TOML manifest into a LockedApp without doing validation on variables.

crates/trigger/src/cli/variables.rs
```diff
@@ -0,0 +1,44 @@
+use spin_core::async_trait;
+use spin_factors::RuntimeFactors;
+use spin_factors_executor::ExecutorHooks;
+use spin_variables::{VariableProviderConfiguration, VariableSourcer};
+
+/// Implements TriggerHooks, sorting required variables
```
itowlson: Sorting?  Maybe comment on why they need sorting?

crates/trigger/src/cli/variables.rs
```diff
@@ -0,0 +1,44 @@
...
+pub struct VariableSorterExecutorHooks {
```
itowlson: Not sure about the naming here.  Why is it an "Executor" hook?

crates/trigger/Cargo.toml
```diff
@@ -32,7 +32,9 @@ spin-factor-wasi = { path = "../factor-wasi" }
 spin-factors = { path = "../factors" }
 spin-factors-executor = { path = "../factors-executor" }
 spin-telemetry = { path = "../telemetry" }
+spin-variables = { path = "../variables" }
 tokio = { workspace = true, features = ["fs", "rt"] }
+toml = { workspace = true }
```
itowlson: I think Ryan had tried to push TOML dependencies into the runtime config crate, so that the trigger crate was abstracted from the physical representation of the runtime config?  (For e.g. SpinKube scenarios.)  I am not sure though.

crates/trigger/src/cli/variables.rs
```diff
@@ -0,0 +1,44 @@
...
+    ) -> anyhow::Result<()> {
+        for (key, variable) in configured_app.app().variables() {
+            self.variable_env_checker(key.clone(), variable.clone())?;
+        }
+        Ok(())
+    }
+}
+
+impl VariableSourcer for VariableSorterExecutorHooks {
+    fn variable_env_checker(&self, key: String, val: spin_app::Variable) -> anyhow::Result<()> {
```
itowlson: `variable_env_checker` is a noun.  I think the idea here is something like `check_variable_has_value` or `check_variable_resolves`.

crates/trigger/src/cli/variables.rs
```diff
@@ -0,0 +1,44 @@
...
+
+        if let Some(config) = configs.into_iter().next() {
+            let (dotenv_path, prefix) = match config {
+                VariableProviderConfiguration::Env(env) => (env.dotenv_path, env.prefix),
+                _ => (None, None),
+            };
+            return self.check(key, val, dotenv_path, prefix);
+        }
+
+        Err(anyhow::anyhow!("No environment variable provider found"))
```
itowlson: Not sure what "provider" refers to in this context.  Have I misunderstood the purpose of this function and it's actually trying to check if the env provider is present?

crates/variables/src/lib.rs
```diff
@@ -70,3 +92,34 @@ impl VariableProviderConfiguration {
...
+    ) -> anyhow::Result<()> {
+        if val.default.is_some() {
+            return Ok(());
+        }
+
+        if let Some(path) = dotenv_path {
+            _ = std::env::set_current_dir(path);
+        }
+
+        match std::env::var(env_key(prefix, &key)) {
```
itowlson: You can't do this.  **There are providers other than the environment one.**  It is legal for a variable to come from Vault or Azure KeyVault or the static provider rather than the environment one.  I tried to stress this in my previous review, sorry - this is fundamental.

itowlson (changes requested):
I feel like I stuffed up communication in my previous review, sorry.  The Spin variables system allows the user to configure variables to be read from multiple different sources.  **You cannot rely on the environment being the only source.**  You need to rethink this PR to completely forget about the environment.  It needs to go through a configured `ProviderResolver` or whatever it's called at the moment, the thing that gets set up in the `VariablesFactor` and checks all configured sources for values.  _That_ will do the environment lookup, and any other lookups - the validation code should _not_ be attempting a partial emulation of its behaviour, you should just use what exists.

crates/trigger/src/cli/variables.rs
seun-ja: Yeah, sorting wasn't the right choice of words. It doesn't sort, but basically checks for variables based on the configuration available for each variable provider. Guess the right word would be "checker"

crates/trigger/src/cli/variables.rs
seun-ja: The naming choice was to reflect its implementation of the `ExecutorHooks` trait

crates/trigger/src/cli/variables.rs
seun-ja: I'd change that. It's meant to be "no default variable provided"

crates/trigger/src/cli/variables.rs
itowlson: Oh, it's called ExecutorHooks now... fair enough then but probably good to change the comment! _grin_

seun-ja:
Closing this PR in favour of #3265. The commit history is messy and almost impossible to resolve

PR #3215, opened 2025-08-06
Factor out variable azure/vault providers into separate crates

fibonacci1729:
This PR factors out the azure and vault providers into separate crates.


crates/runtime-config/src/lib.rs
```diff
@@ -463,6 +469,63 @@ fn sqlite_config_resolver(
     ))
 }
 
+/// Resolves a runtime configuration for the variables factor from a TOML table.
```
fibonacci1729: This code moving here was a necessary evil to decouple `spin-variables` from `spin-variables-vault` and `spin-variables-azure`

lann: Maybe move this all into `variables.rs`?

lann (approved):
This deviates from how we've done runtime config for kv and sqlite but I think its actually closer to the direction I want to go anyway: https://github.com/spinframework/spin/issues/3151

fibonacci1729:
> Not a blocker, but was there a particular reason not to break out the env and static providers at the same time?

No particular reason. I'll do that now.

PR #3216, opened 2025-08-06
Allow providing static variables via the cli

fibonacci1729:


crates/runtime-factors/src/build.rs
```diff
@@ -23,12 +24,26 @@ impl RuntimeFactorsBuilder for FactorsBuilder {
         config: &FactorsConfig,
         args: &Self::CliArgs,
     ) -> anyhow::Result<(Self::Factors, Self::RuntimeConfig)> {
-        let runtime_config = ResolvedRuntimeConfig::<TriggerFactorsRuntimeConfig>::from_file(
-            config.runtime_config_file.clone().as_deref(),
-            config.local_app_dir.clone().map(PathBuf::from),
-            config.state_dir.clone(),
-            config.log_dir.clone(),
-        )?;
+        let mut runtime_config: ResolvedRuntimeConfig<TriggerFactorsRuntimeConfig> =
```
itowlson: Is the new type annotation needed?

crates/runtime-factors/src/build.rs
```diff
@@ -23,12 +24,26 @@ impl RuntimeFactorsBuilder for FactorsBuilder {
...
+        let cli_static_variables = args.get_variables()?.clone();
+        let cli_static_variables_provider = StaticVariablesProvider::new(cli_static_variables);
+
+        // Insert the parsed static variables provided via cli arguments
+        // into the set of variable providers with highest precedence.
+        runtime_config
+            .runtime_config
+            .variables
+            .as_mut()
+            .unwrap()
```
itowlson, line 43: unwrap spidey sense tingling.  How do we know this is safe?

crates/runtime-factors/src/lib.rs
```diff
@@ -119,3 +138,17 @@ impl TryFrom<TomlRuntimeConfigSource<'_, '_>> for TriggerFactorsRuntimeConfig {
...
+
+impl TriggerAppArgs {
+    /// Parse all variable sources into a single merged map.
+    pub fn get_variables(&self) -> anyhow::Result<&HashMap<String, String>> {
+        if self.variables_cache.get().is_none() {
+            let mut variables = HashMap::new();
+            for source in &self.variable {
+                variables.extend(source.get_variables()?);
+            }
+            self.variables_cache.set(variables).unwrap();
```
itowlson, line 150: If somehow the cache was already set this would panic.  Is that the desired behaviour?  I know it should be impossible so I'm not too fussed, just checking.

crates/runtime-factors/src/lib.rs
```diff
@@ -104,6 +107,22 @@ pub struct TriggerAppArgs {
...
+    /// path to a JSON or TOML file may be given as `@file.json` or
+    /// `@file.toml`.
+    ///
+    /// This option may be repeated. If the same key is specified multiple times
+    /// the last value will be used.
+    #[clap(long, multiple = true, value_parser = clap::value_parser!(VariableSource),
+        value_name = "KEY=VALUE | @FILE.json | @FILE.toml")]
+    pub variable: Vec<VariableSource>,
+
+    /// Cache variables to avoid reading files twice
```
itowlson, line 123: How often do we read the files?  Looking at the code it seems like we read them once at trigger startup and shovel them into the static provider.  Is this saying that we reinitialise the runtime config on every request?

crates/variables-static/src/source.rs
```diff
@@ -0,0 +1,116 @@
...
+    TomlFile(PathBuf),
+}
+
+impl VariableSource {
+    pub fn get_variables(&self) -> anyhow::Result<HashMap<String, String>> {
+        match self {
+            VariableSource::Literal(key, val) => Ok([(key.to_string(), val.to_string())].into()),
+            VariableSource::JsonFile(path) => {
+                let json_bytes = std::fs::read(path)
+                    .with_context(|| format!("Failed to read '{}'.", path.display()))?;
```
itowlson: We usually use `quoted_path` (from `spin_common`) for displaying paths - it is longer but it ensures consistency of style.

crates/variables-static/src/source.rs
```diff
@@ -0,0 +1,116 @@
...
+impl FromStr for VariableSource {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Some(path) = s.strip_prefix('@') {
+            let path = PathBuf::from(path);
+            match path.extension().and_then(|s| s.to_str()) {
+                Some("json") => Ok(VariableSource::JsonFile(path)),
+                Some("toml") => Ok(VariableSource::TomlFile(path)),
+                _ => bail!("variable files must end in .json or .toml"),
```
itowlson, line 43: Should this say which file was being naughty?  Not sure if clap already does this for us

crates/variables-static/src/source.rs
```diff
@@ -0,0 +1,116 @@
...
+    JsonFile(PathBuf),
+    TomlFile(PathBuf),
+}
+
+impl VariableSource {
+    pub fn get_variables(&self) -> anyhow::Result<HashMap<String, String>> {
+        match self {
+            VariableSource::Literal(key, val) => Ok([(key.to_string(), val.to_string())].into()),
+            VariableSource::JsonFile(path) => {
+                let json_bytes = std::fs::read(path)
```
itowlson, line 17: ~Is this an absolute path by this point?  If not, am I right that we are interpreting it relative to current working directory (rather than manifest dir) and is that okay?~

~(It probably is.  Paths _in the manifest_ have to be treated as relative to the manifest dir, but this is a path on the command line.)~

Never mind.  I've convinced myself that relative to CWD is correct.  Sorry for the noise.

crates/variables-static/src/source.rs
```diff
@@ -0,0 +1,116 @@
...
+                Ok(toml_vars)
+            }
+        }
+    }
+}
+
+impl FromStr for VariableSource {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
itowlson, line 37: I am not sure what the best solution for this is but I am wary of the static variables provider defining the command line syntax for variable sources - that feels more like a UI concern.

itowlson (commented):
I might need reassuring about a couple of wobbles but generally this looks great

crates/runtime-factors/src/build.rs
fibonacci1729: Nope; must have clicked wrong when jumping to the reference.

crates/runtime-factors/src/lib.rs
fibonacci1729: I'm not sure I understand; the cache is only is set/initialized when `is_none()`.

crates/runtime-factors/src/lib.rs
fibonacci1729: Once on startup but i don't think it's safe to assume that `get_variables` will only ever be called once throughout the code base.

crates/runtime-factors/src/lib.rs
itowlson: I guess the use of the Cell is misleading me.  I was assuming that was used because there was a possibility of race conditions, but maybe it's just to allow interior mutability.  But I'm happy if this isn't a concern!

crates/variables-static/src/source.rs
fibonacci1729: Verified that It does.

```
$ spin up  --variable @noext -f ...
error: Invalid value "@noext" for '--variable <KEY=VALUE | @FILE.json | @FILE.toml>...': variable files must end in .json or .toml

```

crates/variables-static/src/source.rs
fibonacci1729: yeah +1 I agree with your parenthetical but good consideration.

crates/variables-static/src/source.rs
fibonacci1729: It probably is but given that we don't a `cli-common` crate at the moment I think this is OK for now. This seems like the sanest place to share this code across projects.

crates/runtime-factors/src/lib.rs
fibonacci1729: It's using a [OnceCell](https://doc.rust-lang.org/beta/std/cell/struct.OnceCell.html) _which can nominally be written to only once_.

crates/variables-static/src/source.rs
```diff
@@ -0,0 +1,116 @@
+use spin_factors::anyhow::{self, bail, Context as _};
+use std::{collections::HashMap, path::PathBuf, str::FromStr};
+
+#[derive(Clone, Debug)]
+pub enum VariableSource {
+    Literal(String, String),
+    JsonFile(PathBuf),
+    TomlFile(PathBuf),
```
radu-matei, line 9: ref https://github.com/spinframework/spin/issues/3214#issuecomment-3162383120

crates/runtime-factors/src/build.rs
lann: IT IS KNOWN. This is a wart of the derived factors runtime config struct; the fields are filled in incrementally when it is being constructed (somewhere deep in `ResolvedRuntimeConfig::from_file` above) but out here they are always `Some`.

crates/runtime-factors/src/lib.rs
lann: It would be slightly better to do 
```suggestion
            self.variables_cache.set(variables).get_or_init(|| variables);
```
which would be thread safe.

crates/runtime-factors/src/lib.rs
lann: Given that it isn't needed in this PR I'd probably drop the cache and let any future consumer of `get_variables` deal with caching if it needs it.

fibonacci1729: It's not needed directly but there is work being done to validate that variables are present ahead-of-time. I think leaving the caching makes sense.

lann: Oh wait for some reason I thought this was in `spin-variables-static`; IGNORE.

crates/variables-static/src/source.rs
fibonacci1729: Given the comment there i'm going to resolve this for now.

crates/runtime-factors/src/lib.rs
fibonacci1729: I'm not sure how that works; if this needs to be thread-safe i can change it to a [OnceLock](https://doc.rust-lang.org/beta/std/sync/struct.OnceLock.html)

RE (not sure how it works): `.set(variables)` returns a `Result<(), T>` or did you mean `self.variables_cache.get_or_init(|| variables)`?

fibonacci1729: Woulda been nice to use the nightly `get_or_try_init`:

```
      self.variables_cache.get_or_try_init(|| {
            let mut variables = HashMap::new();
            for source in &self.variable {
                variables.extend(source.get_variables()?);
            }
            Ok(variables)
        })
```

lann: ohh right its a `OnceCell` so `!Sync` anyway; yeah `set` should be fine.

PR #3217, opened 2025-08-07
Add SIP for target world validation

tschneidereit:
This SIP describes and motivates the functionality introduced in #2806.

[Rendered](https://github.com/tschneidereit/spin/blob/be389fbc94d5a7734879394c4302e4b66db9a51c/docs/content/sips/022-validate-target-environment.md)

### DIFF
--- a/docs/content/sips/022-validate-target-environment.md
+++ b/docs/content/sips/022-validate-target-environment.md
@@ -1,0 +1,74 @@
+title = "SIP 022 - Target Environment Validation"
+template = "main"
+date = "2025-08-07T00:00:00Z"
+---
+
+Summary: Provide a mechanism to validate that a Spin application is compatible with its intended deployment environment(s).
+
+Owner(s): till@tillschneidereit.net
+
+Created: Aug 7, 2025
+
+## Background
+
+When developing a Spin application, developers might use features that are only available in a specific Spin environment. For example, an application might use a custom trigger or a host capability that is only available in a specific deployment environment, or in a self-hosted Spin instance with specific plugins.
+
+Currently, there is no way to declare the intended deployment environment for an application and have Spin validate that the application is compatible with it. This can lead to situations where an application builds successfully but fails at runtime in the target environment due to missing dependencies, unsupported features, or other incompatibilities.
+
+This SIP proposes a new mechanism to address this problem by allowing developers to specify the target environments for their applications and have Spin validate the compatibility of the application with those environments at build time.
+
+## Proposal
+
+The proposal is to introduce a new `targets` field in the `[application]` section of the `spin.toml` manifest file. This field will contain a list of identifiers for the target environments. Spin will use these identifiers to fetch environment definition files and validate the application against them.
+
+### The `targets` field in `spin.toml`
+
+The `targets` field is a list of strings or tables that identify the target environments. It can be specified in the `[application]` section of `spin.toml` like this:
+
+```toml
+[application]
+name = "my-app"
+version = "1.0.0"
+targets = ["spin-up:3.3", "spinkube:0.4", { registry = "ghcr.io/my-org/environments", id = "my-env:1.0" }, { path = "local-env.toml" }]
+```
+
+Each entry in the list can be:
+- A string that identifies an environment in the default registry (e.g., `"spin-up:3.3"`).
+- A table with `registry` and `id` keys to identify an environment in a custom OCI registry.
+- A table with a `path` key to specify a local environment definition file.
+
+### Environment Definition Files
+
+An environment definition is a TOML file that describes the capabilities of a Spin environment. It specifies the triggers that are available in the environment, and for each trigger, the WIT worlds that are compatible with it and the host capabilities that it supports.
+
+Here is an example of an environment definition file:
+
+```toml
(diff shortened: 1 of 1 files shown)


itowlson (approved):

docs/content/sips/022-validate-target-environment.md
```diff
@@ -0,0 +1,75 @@
...
+```toml
+[application]
+name = "my-app"
+version = "1.0.0"
+targets = ["spin-up:3.3", "spinkube:0.4", { registry = "ghcr.io/my-org/environments", id = "my-env:1.0" }, { path = "local-env.toml" }]
+```
+
+Each entry in the list can be:
+- A string that identifies an environment in the default registry (e.g., `"spin-up:3.3"`).
+- A table with `registry` and `id` keys to identify an environment in a custom OCI registry.
```
lann, line 37: We should permit `wkg` registry names (bare domain) here with the `/.well-known/wasm-pkg/registry.json` indirection.

docs/content/sips/022-validate-target-environment.md
```diff
@@ -0,0 +1,75 @@
...
+Here is an example of an environment definition file:
+
+```toml
+# spin-up.3.2.toml
+[triggers.http]
+worlds = ["spin:up/http-trigger@3.2.0", "spin:up/http-trigger-rc20231018@3.2.0"]
+capabilities = ["local_service_chaining"]
+
+[triggers.redis]
+worlds = ["spin:up/redis-trigger@3.2.0"]
```
lann, line 53: This makes me wonder if we should accept e.g. `{ trigger_type = "redis", world = "spin:up/redis-trigger@3.2.0" }` as a valid target.

lann: I guess a slightly more verbose but flexible way to get the same effect would be allowing inline environment definitions. :shrug:

lann: Or maybe the target could be _just_ the world reference? While that wouldn't have explicit trigger types those can be sort of inferred from the manifest itself. :thinking:

docs/content/sips/022-validate-target-environment.md
lann: Alternatively we could specify a slightly different syntax that implies this lookup, like

```toml
targets = ["my.org/my-env@1.0"]
```

This would look up the oci prefix at `https://my.org/.well-known/wasm-pkg/registry.json` and use it to construct an oci reference like `ghcr.io/my-org/my-env:1.0`. If we wanted to mechanically avoid conflicts between environments and regular wasm packages we could stick another substring in there e.g. `ghcr.io/my-org/my-env:spin-env-1.0` or `ghcr.io/my-org/spin-env/my-env:1.0` or something.

itowlson: @lann These entries point to TOML documents not WIT documents.  (The TOML documents then point onward to WIT documents, which _are_ typically `wkg` package references.)

docs/content/sips/022-validate-target-environment.md
itowlson: @lann I am wary of that for two reasons:

1. It breaks down as soon as multi-trigger-type enters play - you find yourself needing to infer which world corresponds to which trigger type.  We tried that and it was terrible.
2. It breaks the "I want to make sure I can deploy to..." abstraction.  The goal here is for the developer to say "I want this to run on Wasmtime 35, Spin 3.2, and SpinKube 3.1."  If I want to target `spin:up/http-trigger@3.2.0`, and figure out for myself whether that's compatible with SpinKube 3.1, then I have the tools to do that.  The idea is to take care of "knowing which worlds a host supports" for you.

I'm not opposed to inline environments as an escape hatch, but I'd like them to be defined in the same way as other environments, and at that point I don't see a vast amount of value compared to putting them in a `my-environment.toml` file.

docs/content/sips/022-validate-target-environment.md
lann: Sure; I'm suggesting that we could reuse the domain -> OCI mapping that wkg already implements to allow nicer specifications of targets from registries.

itowlson: Ah sorry thanks for the clarification

fibonacci1729 (approved):

docs/content/sips/022-validate-target-environment.md
itowlson: This feels like a convenience feature we could add without breaking back compat if the feature achieves traction?  Or are we sufficiently confident that other folks will be publishing environment TOMLs that we need to offer this in the first drop?

itowlson:
@lann There's a couple of conversation chains here that end with me saying "could we not, or at least not now."  Do those remain concerns for you, or is it okay to merge as is, and iterate on improvements as the spirit moves us?

itowlson:
Seems like this possibly got lost while Fermyon folks were distracted by, y'know, the things, and has been sitting here for 9 months (oops).  @lann the outstanding question is with you: are your comments blockers, or can we merge and iterate later?

lann:
That's @lann-9-months ago. I don't know that guy.

No, no blockers.

PR #3219, opened 2025-08-07
Allow optional description for manifest variables

### LINKED ISSUE #3218: [feature request]: optional description for application variables
The variable schema in `spin.toml` should have an optional `description` field. This field would be used by tools that render the application manifest and help operators and developers understand the application post-deployment.

https://github.com/spinframework/spin/blob/0a4c05ac67f8cfb26f714cf7cbff754dc34f1922/crates/manifest/src/schema/common.rs#L10-L37

fibonacci1729:
closes #3218

### DIFF
--- a/crates/manifest/src/schema/common.rs
+++ b/crates/manifest/src/schema/common.rs
@@ -11,6 +11,13 @@ use super::json_schema;
 #[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
 #[serde(deny_unknown_fields)]
 pub struct Variable {
+    /// A brief description of the variable.
+    ///
+    /// Example: `description = "This is a variable"`
+    ///
+    /// Learn more: https://spinframework.dev/variables#adding-variables-to-your-applications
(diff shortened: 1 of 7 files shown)


crates/manifest/src/schema/common.rs
```diff
@@ -11,6 +11,13 @@ use super::json_schema;
 #[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
 #[serde(deny_unknown_fields)]
```
lann, line 12: @radu-matei Note this means you won't be able to use this field with old versions of Spin.

radu-matei: Yeah, saw that. It makes sense.

crates/manifest/tests/ui/maximal.json
```diff
@@ -21,6 +21,7 @@
   },
   "variables": {
     "var_one": {
+      "description": "Test me like one of your French strings!",
```
adamreese, line 24: Of course, Rose! 🌹

PR #3220, opened 2025-08-07
Shell completions

itowlson:
I should know by now that these PRs only end in tears.

This uses a technique explored in https://github.com/karthik2804/spin-autocomplete.  The idea is, instead of generating completions directly or going through the flawed clap-complete, we generate a YAML file (it's JSON, but JSON is YAML so I don't have to deal with _actual_ YAML), then we can run [completely](https://github.com/bashly-framework/completely) over it, thus:

```bash
spin maintenance generate-shell-completions -o completely.yaml  # the name matters
docker run --rm --user $(id -u):$(id -g) --volume "$PWD:/app" -e COMPLETELY_OUTPUT_PATH=spin_comps.sh dannyben/completely generate
source ./spin_comps.sh
```

So we can get a high level of control over the completions, but without a separate binary or handcrafting bash.

Note as per Karthik's repo this relies on `bashcompinit` to work on zsh - sorry.

Draft because file completions are not cat-like and I am recruiting fellow believers to fix my ignorance.

### DIFF
--- a/src/commands/build.rs
+++ b/src/commands/build.rs
@@ -22,6 +22,7 @@ pub struct BuildCommand {
         short = 'f',
         long = "from",
         alias = "file",
+        value_hint = clap::ValueHint::AnyPath,
     )]
     pub app_source: Option<PathBuf>,
 
--- a/src/commands/doctor.rs
+++ b/src/commands/doctor.rs
@@ -17,7 +17,8 @@ pub struct DoctorCommand {
         name = APP_MANIFEST_FILE_OPT,
         short = 'f',
         long = "from",
-        alias = "file"
+        alias = "file",
+        value_hint = clap::ValueHint::AnyPath,
     )]
     pub app_source: Option<PathBuf>,
 }
--- a/src/commands/maintenance.rs
+++ b/src/commands/maintenance.rs
@@ -9,6 +9,8 @@ pub enum MaintenanceCommands {
     GenerateReference(GenerateReference),
     /// Generate JSON schema for application manifest.
     GenerateManifestSchema(GenerateSchema),
+    /// Generate a `completely` file which can then be processed into shell completions.
+    GenerateShellCompletions(GenerateCompletions),
 }
 
 impl MaintenanceCommands {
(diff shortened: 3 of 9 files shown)


lann:
Feel free to reject this suggestion as naive and/or too terrible to contemplate:

Would it be plausible to use the clap _builder_ API to construct a parallel `Command` that reuses those bits of derived clap types that aren't broken and manually reconstruct (or omit) the bits that are broken? (and then use e.g. clap_complete)

lann:
...though it occurs to me that you may be complaining about clap_complete itself and not our weird subcommand shenanigans.

itowlson:
I would dearly love to use `clap_complete`.  Unfortunately:

1. `clap_complete` doesn't have the vocabulary to express things like "list of templates" or "list of plugins."  It only allows the `ValueHint` enum.
2. The version of `clap_complete` that we are, for now, stuck with ignores `ValueHint` and completes EVERYTHING with paths.  (This is fixed in v4.x, but that would require "up"-grading to clap 4.)

So I don't see a path to success with it for now, even with constructing a parallel Command tree, sorry.

itowlson:
In shock news, `completely` has also broken my heart.

The trouble is that for parameter completion it relies on a first-last pattern, e.g. `spin up*--from`, which works great for the first `TAB` after the expression, but apparently not so great for subsequent tabs, because the last thing on the line is no longer `--from` but the initial completion.  The impact is that if you have a file nested in a directory, completion stops at the directory e.g. if you are trying to get to `foo/spin.toml` you go `spin up --from fo<TAB>`, and it completes `foo` then _leaves a space_ instead of including a `/` and allowing you to continue typing `s<TAB>` - you have to backspace over the space, then type `/s<TAB>` which is quite the speed bump.

In theory `-o bashdefault -o default` would fix the "not continuing" thing but in practice I think the first-last thing is messing it up.

By contrast, `clap_complete` uses a cleverer technique for figuring out which flag governs the current completion, then goes "compgen -f -o bashdefault -o default YOLO" so that _everything_ gets completed as files.

Well, time to put this away for another year I guess, until I feel ready to fall in love with another completions technology that will also disappoint me.

itowlson:
Superseded by #3481.

PR #3223, opened 2025-08-11
expressions: Fix Display for Template escaping

lann:
`// Rust format strings escape "{" with "{{"", so "{{" becomes "{{{{"`


lann:
Split out of #3168; merging on the basis of that approval

PR #3224, opened 2025-08-11
Validate allowed_outbound_hosts template with defaults

lann:
If all variables in a template have default values then validate with those defaults.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8584,6 +8584,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "spin-common",
+ "spin-expressions",
  "spin-locked-app",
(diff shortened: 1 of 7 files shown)


itowlson:
Could you add more about the intent here?  My memory is that we resolve the AOHs during trigger startup using all configured providers - it woud be useful to capture the reason for validating against the defaults rather than against the actual values.

lann:
This is primarily about catching problems when pushing to OCI, e.g. `spin registry push` for SpinKube. As a side benefit it can also catch bogus default variable values if you happen to be testing locally with e.g. env provider variables.

fibonacci1729 (approved):

PR #3228, opened 2025-08-13
cli: Resolve clap/deprecated warnings

lann:


src/clap_markdown/mod.rs
```diff
@@ -517,19 +503,17 @@ fn write_arg_markdown(buffer: &mut String, arg: &clap::Arg) -> fmt::Result {
     //--------------------
 
     let possible_values = arg
-        .get_possible_values()
-        .unwrap_or_default()
+        .get_value_parser()
+        .possible_values()
         .into_iter()
+        .flatten()
```
lann, line 509: This was the only substantive change in this file. The rest was rustfmt.

itowlson:
Could you say more about the motivation here?  Deprecation doesn't seem like a substantial concern when we're intentionally staying on an old package version, so I'm not clear on the value of turning on the deprecation warnings.

lann:
I've done this twice now on clap 4 upgrade attempts. :shrug:

Issue #2444, opened 2024-04-09
HTTPS Connection reuse from Spin worker


dlozina:
Hi all,
I am looking into possibly having two outbound HTTPS calls from my Spin worker or Webasembly module in general to the same server and reusing the connection. 
It would be beneficial so we don't have to do a handshake to the server again, and requests over the network would be faster for that use-case. 
If I understood correctly, there is no technical reason why we can't support connection reuse with `wasi-http`, I have not found a good code example of the implementation for that use case.
Is this something that is planned to be implemented in Spin? Could you please point me to some examples that cover this use case?
Thank you!

itowlson:
@dicej Is this something @dlozina could access by wit-bindgen-ing the wasi-http outgoing-handler interface directly into their project, or is it something the Spin host would need to (invisibly) do?

dicej:
This requires a Spin change.  Currently, we use the default outgoing request handler provided by `wasmtime-wasi-http` (https://github.com/fermyon/spin/blob/b86af6c7188e84f5824f7eef757c65ff2ad265c4/crates/core/src/lib.rs#L236), which creates a new TCP connection and sends the request using `hyper`: https://github.com/bytecodealliance/wasmtime/blob/de0a9d7ce959d83a336d4f13487cf00e69e9efb2/crates/wasi-http/src/types.rs#L114.  In order to reuse requests, we'd need to provide our own handler instead.

One appealing option would be to use `reqwest`, which has built-in support for connection pooling.  However, as of this writing it does not yet support trailers, which means it can't be used to fully implement `wasi:http/outgoing-handler`.  Using trailers is pretty rare in the real world, though, so perhaps that's an acceptable tradeoff.

dlozina:
Hi @itowlson and @dicej , first of all thank you for the extremely fast response! 🙏🏻 
I agree with @dicej , trailer support is not a dealbreaker ATM.
@dicej does this boils down changing hyper with reqwest?

dicej:
@dlozina We'll need to create an outgoing http implementation from scratch based on `reqwest` instead of using the one provided by `wasmtime-wasi-http`, which is based on `hyper`.  It shouldn't be terribly difficult.  FWIW, I created one for `wasi:http@0.3.0-draft` which we could use as a starting point: https://github.com/dicej/isyswasfa/blob/df9a153fafbbdf532d811bdc53e61ff7368e4898/cli/src/main.rs#L126-L244

[...]

PR #3229, opened 2025-08-13
enable connection pooling and reuse for outbound wasi-http

dicej:
This updates `factor-outbound-http/wasi.rs` to use `hyper_util::client::legacy::Client`, which does connection pooling by default. `Client` requires a `tower_service::Service` implementation for creating new connections, which in turn required moving code around to accomodate a different flow of control.  It also forces us to work at a slightly higher level of abstraction, so we don't have as much fine-grained control as before, notably:

- We can't make HTTP/2-specific request tweaks like we did before (e.g. removing or adding `host` headers, or tweaking the path-with-query).  I'm mainly leaving it up to `hyper-util` to do the right thing here :crossed_fingers:.

- We can't provide an `AbortOnDropJoinHandle` when constructing an `IncomingResponse` object, which may mean the connection continues to exist after the response is dropped, but that's kind of the point of connection pooling anyway.

Open questions: do we want to make this configurable, and if so, should it be enabled or disabled by default?

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -135,15 +135,22 @@ impl WasiHttpView for WasiHttpImplInner<'_> {
         request: Request<wasmtime_wasi_http::body::HyperOutgoingBody>,
         config: wasmtime_wasi_http::types::OutgoingRequestConfig,
     ) -> wasmtime_wasi_http::HttpResult<wasmtime_wasi_http::types::HostFutureIncomingResponse> {
+        let connection_pooling = self.state.connection_pooling;
+        let builder = move || {
+            let mut builder = Client::builder(TokioExecutor::new());
+            if !connection_pooling {
+                builder.pool_max_idle_per_host(0);
+            }
+            builder
+        };
+
         let http_clients = self
             .state
             .wasi_http_clients
             .get_or_insert_with(|| HttpClients {
-                http1: Client::builder(TokioExecutor::new()).build(HttpConnector),
-                http2: Client::builder(TokioExecutor::new())
-                    .http2_only(true)
-                    .build(HttpConnector),
-                https: Client::builder(TokioExecutor::new()).build(HttpsConnector),
+                http1: builder().build(HttpConnector),
+                http2: builder().http2_only(true).build(HttpConnector),
+                https: builder().build(HttpsConnector),
             })
             .clone();
 
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -1,23 +1,41 @@
-use std::{error::Error, future::Future, pin::Pin, sync::Arc, time::Duration};
+use std::{
+    error::Error,
+    future::Future,
+    io::IoSlice,
+    pin::Pin,
+    sync::Arc,
+    task::{Context, Poll},
+    time::Duration,
+};
 
-use anyhow::Context;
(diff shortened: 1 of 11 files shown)


dicej:
Fixes #2444

fibonacci1729 (approved):

fibonacci1729:
Verified no regressions w.r.t outbound-http2/gRPC

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -104,6 +135,25 @@ impl WasiHttpView for WasiHttpImplInner<'_> {
...
+            builder
+        };
+
+        let http_clients = self
+            .state
+            .wasi_http_clients
+            .get_or_insert_with(|| HttpClients {
+                http1: builder().build(HttpConnector),
+                http2: builder().http2_only(true).build(HttpConnector),
+                https: builder().build(HttpsConnector),
```
lann, line 153: `Builder::build` takes `&self` (cloning internally) so you could reuse one `builder` here if you wanted. 🤷

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -114,13 +164,198 @@ impl WasiHttpView for WasiHttpImplInner<'_> {
...
+#[derive(Clone)]
+struct ConnectOptions {
+    blocked_networks: BlockedNetworks,
+    connect_timeout: Duration,
+}
+
+// We must use task-local variables for these config options when using
+// `hyper_util::client::legacy::Client::request` because there's no way to plumb
+// them through as parameters.  Moreover, if there's already a pooled connection
+// ready, we'll reuse that and ignore these options anyway.
```
lann, line 184: I'd probably take the tiny theoretical performance hit and use a `OnceLock` for building the clients to avoid this.

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -114,13 +164,198 @@ impl WasiHttpView for WasiHttpImplInner<'_> {
                     self.state.request_interceptor.clone(),
                     self.state.self_request_origin.clone(),
                     self.state.blocked_networks.clone(),
+                    http_clients,
```
lann, line 167: This is mostly out of scope so feel free to ignore but it seems clear that these `state` fields should clearly be their own struct with `send_request` as a method. I have no idea why I didn't do that last time I was in here...

crates/factor-outbound-http/src/wasi.rs
dicej: I'm not seeing how a `OnceLock` would avoid the need for task-local variables.  We only build the clients once per instance, but each outbound request could have its own `connect_timeout`, for example.  I.e. we can't just build a client with a fixed `connect_timeout` and use it for all requests.  Or am I misunderstanding what you're proposing?

crates/factor-outbound-http/src/wasi.rs
dicej: Yeah, but then the code would be order-dependent -- I'd have to be sure to build the `http2` one last so that the `http2_only` setting doesn't affect the other ones.  I'd prefer to contain the side effect by using separate builders unless there's a good reason not to.

crates/factor-outbound-http/src/wasi.rs
lann: Ahhh OK yes that makes sense. I think I'm seeing some of the reasons why this is "`legacy`" 😬

PR #3232, opened 2025-08-15
Appease clippy

lann:


src/commands/new.rs
```diff
@@ -140,15 +140,13 @@ impl TemplateNewCommandCore {
...
 
-        let template = match &template_id {
+        let template = match template_id {
             Some(template_id) => match template_manager
-                .get(template_id)
+                .get(&template_id)
                 .with_context(|| format!("Error retrieving template {template_id}"))?
             {
                 Some(template) => template,
-                None => match prompt_template(&template_manager, &variant, &[template_id.clone()])
```
lann, line 149: Clippy wanted to use some fancy `std::slice::from_ref` malarkey to avoid this clone but I showed it what for.

fibonacci1729 (approved):

PR #3233, opened 2025-08-15
selects only id when checking existence of data

### LINKED ISSUE #2921: Cosmos key-value store `exists` downloads data
The Cosmos key-value provider's `exists` method calls `get_pair`, which always fully materialises the value blob on the client.  This is potentially slow for large values (and could incur egress costs, although we'd expect this provider to be used only within Azure I imagine).  It might be worth implementing an additional method that selects only the `id`, and reimplementing `exists` in terms of that.

seun-ja:
Fixes #2921

PR sieves out redundant data, only checks for `id`.  Does the same for `get_keys_query`, only querying for `c.id` and `c.store_id` (needed for deserialising to `Key` struct), thereby excluding possible heavy data from `value`

### DIFF
--- a/crates/key-value-azure/src/store.rs
+++ b/crates/key-value-azure/src/store.rs
@@ -458,13 +458,19 @@ impl AzureCosmosStore {
     }
 
     fn get_id_query(&self, key: &str) -> String {
-        let mut query = format!("SELECT c.id, c.store_id FROM c WHERE c.id='{key}'");
+        let mut query = match self.store_id {
+            Some(_) => format!("SELECT c.id, c.store_id FROM c WHERE c.id='{key}'"),
+            None => format!("SELECT c.id FROM c WHERE c.id='{key}'"),
+        };
         self.append_store_id(&mut query, true);
         query
     }
 
     fn get_keys_query(&self) -> String {
-        let mut query = "SELECT c.id, c.store_id FROM c".to_owned();
+        let mut query = match self.store_id {
+            Some(_) => "SELECT c.id, c.store_id FROM c".to_owned(),
+            None => "SELECT c.id FROM c".to_owned(),
+        };
         self.append_store_id(&mut query, false);
         query
     }
--- a/crates/key-value-azure/src/store.rs
+++ b/crates/key-value-azure/src/store.rs
@@ -190,7 +190,18 @@ impl Store for AzureCosmosStore {
     }
 
     async fn exists(&self, key: &str) -> Result<bool, Error> {
-        Ok(self.get_entity_by_id::<Key>(key).await?.is_some())
+        let mut stream = self
+            .client
+            .query_documents(Query::new(self.get_id_query(key)))
+            .query_cross_partition(true)
+            .max_item_count(1)
+            .into_stream::<Key>();
+
+        match stream.next().await {
+            Some(Ok(res)) => Ok(!res.results.is_empty()),
+            Some(Err(e)) => Err(log_error(e)),
+            None => Ok(false),
+        }
     }
 
     async fn get_keys(&self) -> Result<Vec<String>, Error> {
@@ -409,28 +420,6 @@ impl AzureCosmosStore {
         let query = self
             .client
             .query_documents(Query::new(self.get_query(key)))
-            .query_cross_partition(true)
-            .max_item_count(1);
-
-        // There can be no duplicated keys, so we create the stream and only take the first result.
-        let mut stream = query.into_stream::<F>();
-        let Some(res) = stream.next().await else {
-            return Ok(None);
-        };
-        Ok(res
-            .map_err(log_error)?
-            .results
-            .first()
-            .map(|(p, _)| p.clone()))
-    }
-
-    async fn get_entity_by_id<F>(&self, key: &str) -> Result<Option<F>, Error>
-    where
-        F: CosmosEntity + Send + Sync + serde::de::DeserializeOwned + Clone,
-    {
-        let query = self
-            .client
-            .query_documents(Query::new(self.get_id_query(key)))
             .query_cross_partition(true)
             .max_item_count(1);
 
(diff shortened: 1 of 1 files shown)


crates/key-value-azure/src/store.rs
```diff
@@ -424,6 +424,28 @@ impl AzureCosmosStore {
             .map(|(p, _)| p.clone()))
     }
 
+    async fn get_entity_by_id<F>(&self, key: &str) -> Result<Option<F>, Error>
```
lann: nit: This is a little bit of a confusing method on its own; why would I want to get the ID from the ID? Maybe just 
```suggestion
    async fn exists(&self, key: &str) -> Result<bool, Error>
```

lann: Actually you could probably just inline this method into `Store::exists` above. It looks like that was already done for `get_many`.

crates/key-value-azure/src/store.rs
```diff
@@ -446,8 +468,14 @@ impl AzureCosmosStore {
...
 
+    fn get_id_query(&self, key: &str) -> String {
+        let mut query = format!("SELECT c.id FROM c WHERE c.id='{key}'");
+        self.append_store_id(&mut query, true);
+        query
+    }
+
     fn get_keys_query(&self) -> String {
-        let mut query = "SELECT * FROM c".to_owned();
+        let mut query = "SELECT c.id, c.store_id FROM c".to_owned();
```
lann, line 467: Is there an issue here with certain mandatory metadata being omitted from results?

seun-ja: Once your query includes `store_id`, those mandatory metadata won't be required. was tested last week

lann: > Once your query includes store_id, those mandatory metadata won't be required.

What about the `SELECT c.id` in `get_id_query` above?

crates/key-value-azure/src/store.rs
kate-goldenring: +1

crates/key-value-azure/src/store.rs
```diff
@@ -424,6 +424,28 @@ impl AzureCosmosStore {
...
+        F: CosmosEntity + Send + Sync + serde::de::DeserializeOwned + Clone,
+    {
+        let query = self
+            .client
+            .query_documents(Query::new(self.get_id_query(key)))
+            .query_cross_partition(true)
+            .max_item_count(1);
+
+        // There can be no duplicated keys, so we create the stream and only take the first result.
+        let mut stream = query.into_stream::<F>();
```
kate-goldenring: I would expect this to fail to serialize to a `Key` since the `get_id_query` does not select `store_id` which is a required field of `Key`. This should probably not be a generic function since the query is not generic (by no longer selecting all).

crates/key-value-azure/src/store.rs
```diff
@@ -446,8 +468,14 @@ impl AzureCosmosStore {
         query
     }
 
+    fn get_id_query(&self, key: &str) -> String {
+        let mut query = format!("SELECT c.id FROM c WHERE c.id='{key}'");
```
kate-goldenring:
```suggestion
        let mut query = format!("SELECT c.id, c.store_id FROM c WHERE c.id='{key}'");
```

crates/key-value-azure/src/store.rs
seun-ja: Yes, that should include `store_id` as well. Added it

kate-goldenring: After taking another look at the `Key` structure, i noticed that `store_id` is optional. Since `store_id` is optional, if the entry does not have a `store_id` will this query fail? This query may need to be dependent on that `Option`:
```suggestion
        let mut query = match self.store_id {
            Some(_) => "SELECT c.id, c.store_id FROM c".to_owned(),
            None => "SELECT c.id FROM c".to_owned(),
        };
```

Same goes for the `get_id_query`.

seun-ja: `c.store_id` seems to be required in queries; if not, it would require including default metadata in the query.

In case where `store_id` doesn't have an entry, it should map to `None`.

Also, if we don't include `store_id` in the query, I think we might have a deserialization error when mapping to `Key`

kate-goldenring: not all Keys have a `store_id`. If `app_id` is not set on start up, then id is used as partition key:
```
impl CosmosEntity for Key {
    type Entity = String;

    fn partition_key(&self) -> Self::Entity {
        self.store_id.clone().unwrap_or_else(|| self.id.clone())
    }
}
```

crates/key-value-azure/src/store.rs
```diff
@@ -446,8 +457,20 @@ impl AzureCosmosStore {
...
+        self.append_store_id(&mut query, true);
+        query
+    }
+
     fn get_keys_query(&self) -> String {
-        let mut query = "SELECT * FROM c".to_owned();
+        let mut query = match self.store_id {
+            Some(_) => "SELECT c.id, c.store_id FROM c".to_owned(),
+            None => "SELECT c.id FROM c".to_owned(),
+        };
```
kate-goldenring: @seun-ja i tested this on a cosmos container that uses `id` as partition key (and no store_id) and your simplified query works. Sorry to shift you to more complexity. I think we can revert to your original approach of: 
```suggestion
        let mut query = "SELECT c.id, c.store_id FROM c".to_owned();
```

crates/key-value-azure/src/store.rs
```diff
@@ -446,8 +457,20 @@ impl AzureCosmosStore {
         query
     }
 
+    fn get_id_query(&self, key: &str) -> String {
+        let mut query = match self.store_id {
+            Some(_) => format!("SELECT c.id, c.store_id FROM c WHERE c.id='{key}'"),
+            None => format!("SELECT c.id FROM c WHERE c.id='{key}'"),
+        };
```
kate-goldenring: Same as below. Can revert this to your original approach
```suggestion
        let mut query = format!("SELECT c.id, c.store_id FROM c WHERE c.id='{key}'");
```

kate-goldenring (approved):

PR #3238, opened 2025-08-20
Implement open ai api llm backend

### LINKED ISSUE #3211: Implement OpenAI API LLM backend
From what I can tell it is more-or-less the defacto standard "do LLM remotely" API today: https://github.com/openai/openai-openapi

seun-ja:
This PR aims to resolve #3211.  Remote LLM implementation now provides options: the default option and OpenAI.

The default represents the current option, and as the name states, it's the default in case none is provided.

It's expandable while preserving the original API and not breaking it.

Concerns arise regarding `wasi::llm` parameters that are not compatible with OpenAI's API documentation, which may necessitate adjustments to the current definitions available in wit.. Documented [here](https://github.com/spinframework/spin/issues/3211#issuecomment-3206284505)

[Update]

In reference to @rylev's [comment](https://github.com/spinframework/spin/issues/3211#issuecomment-3219947039) there would be no need for a change in the wit file, just going to use what's available.


itowlson:
It would be useful to see an example of how a user selects this back-end in their runtime config.  It might also be useful to add some discussion to the PR of the tradeoffs you considered when bundling this in with the existing `cloud-gpu` remote back-end vs. having it as a separate (typed) back end.

seun-ja:
The PR is going with bundling the open-ai api backend into the existing `llm-remote-http` crate because:

- They both share lots of common resource, therefore, avoid unneccessary cyclic import issue
- Interegrating it with existing won't break the current set-up
- Keeps the current streamlined options, `Spin` (local) and `RemoteHttp`, while still making the later customizable and expandable depending on the optional field added to the runtime-config

-----

@itowlson, I just realised in my PR main comment I said "breaking it", apologies, I meant "not breaking it" 😄

crates/factor-llm/src/spin.rs
```diff
@@ -132,6 +133,7 @@ impl LlmCompute {
 pub struct RemoteHttpCompute {
     url: Url,
     auth_token: String,
+    custom_llm: Option<String>,
```
karthik2804: I think we want this to be an enum which would allow us to constrain it to only the options implemented.

crates/llm-remote-http/src/open_ai/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+        );
+        spin_telemetry::inject_trace_context(&mut headers);
+
+        let chat_url = url
+            .join("/v1/chat/completions")
+            .map_err(|_| wasi_llm::Error::RuntimeError("Failed to create URL".to_string()))?;
+
+        tracing::info!("Sending remote inference request to {chat_url}");
+
+        let body = CreateChatCompletionRequest {
```
karthik2804, line 64: I think we should use the generate API instead of Chat completions here because our interface does not really lend itself to having multiple roles. The `wit` interface aligns more closely with the generate API instead of chat completions.

seun-ja: I have looked into this option; the endpoint is `/v1/responses`. While it is more compatible with the current `wit` interface, I noticed an unusual behaviour with this endpoint while testing it. 

For example, it didn't work with `gpt-4`. I got this error
```bash
"error": {
    "message": "The model `gpt-4` does not exist or you do not have access to it.",
    "type": "invalid_request_error",
    "param": null,
    "code": "model_not_found"
 }
```
But when I tried using it with `gpt-4o`, it worked. Got a different error message and rightly so 🫣
```bash
"error": {
    "message": "You exceeded your current quota, please check your plan and billing details. For more information on this error, read the docs: https://platform.openai.com/docs/guides/error-codes/api-errors.",
    "type": "insufficient_quota",
    "param": null,
    "code": "insufficient_quota"
  }
```
I tried with Ollama using the `gpt-oss:20b` model,  got this error:
``` bash
404 page not found
```
Dived deep and realised Ollama doesn't have that endpoint. However, Ollama has the `/generate` endpoint, but not compatible with what we need here. Returns this:
```bash
{
    "model": "gpt-oss:20b",
    "created_at": "2025-08-28T21:17:25.432385Z",
    "response": "",
    "done": true,
    "done_reason": "load"
}
```

In conclusion, I think the current setup is fine, the `User` role matches the expected behaviour, similar to what the default remote LLM does.

crates/factor-llm/src/spin.rs
```diff
@@ -132,6 +133,8 @@ impl LlmCompute {
 pub struct RemoteHttpCompute {
     url: Url,
     auth_token: String,
+    #[serde(default)]
+    custom_llm: CustomLlm,
```
karthik2804: Probably calling it backend might make more sense than custom_llm as this refers to the backing API.

crates/llm-remote-http/src/lib.rs
```diff
@@ -50,130 +117,106 @@ struct EmbeddingResponseBody {
...
+            usage: wasi_llm::EmbeddingsUsage {
+                prompt_token_count: value.usage.prompt_tokens,
+            },
+        }
+    }
+}
+
+#[derive(Debug, Default, serde::Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum CustomLlm {
```
karthik2804: This probably needs a better name. Something similar to the backend or something along those lines.

dicej (approved):
LGTM.  I don't have much LLM API experience, so I can't comment on the details of how this uses the OpenAI APIs, but the structure of the code seems reasonable to me.

Is there any way we can automate testing this without necessarily spending money on an OpenAI account for CI, and without writing a mock API server from scratch (which might not be very useful anyway)?

seun-ja:
Thanks for the review @dicej.

There could be a way. Would look into that.

On the other hand, because the API is compatible with other similar LLM APIs, such as Ollama, and it's free.

kmk142789 (approved):

crates/llm-remote-http/src/lib.rs
```diff
@@ -38,6 +68,43 @@ struct InferResponseBody {
     usage: InferUsage,
 }
 
+#[derive(Deserialize)]
+struct CreateChatCompletionResponse {
+    /// A unique identifier for the chat completion.
+    #[serde(rename = "id")]
+    _id: String,
```
itowlson: Not sure why we're deserialising these presumably unused fields?  I see you'd need them with `deny_unknown_fields` but that doesn't seem to be turned on...?

(This seems to be a general pattern throughout.  If we need it then no worries, just looks odd to me.)

crates/llm-remote-http/src/lib.rs
```diff
@@ -50,130 +117,106 @@ struct EmbeddingResponseBody {
...
+    fn embeddings(&self) -> Vec<Vec<f32>> {
+        self.data
+            .iter()
+            .map(|embedding| embedding.embedding.clone())
+            .collect()
+    }
+}
+
+#[derive(Deserialize)]
+struct OpenAIEmbeddingUsage {
```
itowlson: By the name, this seems specific to the OpenAI backend, which makes me wonder if `CreateEmbeddingResponse` is also OpenAI-specific.  I'd expect OpenAI stuff to be in `open_ai.rs` but I'm not sure I'm understanding this right.

crates/llm-remote-http/src/open_ai.rs
```diff
@@ -0,0 +1,182 @@
...
+
+        let body = CreateEmbeddingRequest {
+            input: data,
+            model,
+            encoding_format: None,
+            dimensions: None,
+            user: None,
+        };
+
+        let chat_url = self
```
itowlson: `chat_url` seems a misleading name for this.

crates/llm-remote-http/src/open_ai.rs
```diff
@@ -0,0 +1,182 @@
...
+            "authorization",
+            HeaderValue::from_str(&format!("bearer {}", self.auth_token)).map_err(|_| {
+                wasi_llm::Error::RuntimeError("Failed to create authorization header".to_string())
+            })?,
+        );
+        spin_telemetry::inject_trace_context(&mut headers);
+
+        let chat_url = self
+            .url
+            .join("/v1/chat/completions")
```
itowlson: This string appears in multiple places: should we pull it out to a `const`?

crates/llm-remote-http/src/open_ai.rs
```diff
@@ -0,0 +1,182 @@
...
+            input: data,
+            model,
+            encoding_format: None,
+            dimensions: None,
+            user: None,
+        };
+
+        let chat_url = self
+            .url
+            .join("/v1/embeddings")
```
itowlson: Again, consider creating a `const`

crates/llm-remote-http/src/open_ai/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    dimensions: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    user: Option<String>,
+}
+
+#[derive(serde::Deserialize)]
+#[serde(untagged)]
+enum CreateChatCompletionResponses {
+    Success(CreateChatCompletionResponse),
```
itowlson: This reinforces my concern about the location of `CreateChatCompletionResponse`

crates/llm-remote-http/src/open_ai.rs
```diff
@@ -0,0 +1,182 @@
...
+    encoding_format: Option<EncodingFormat>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    dimensions: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    user: Option<String>,
+}
+
+#[derive(serde::Deserialize)]
+#[serde(untagged)]
+enum CreateChatCompletionResponses {
```
itowlson: Confused by this name.  It appears to contain only one response?

crates/llm-remote-http/src/open_ai.rs
```diff
@@ -0,0 +1,182 @@
...
+    Success(CreateChatCompletionResponse),
+    Error { error: ResponseError },
+}
+
+#[derive(serde::Deserialize)]
+#[serde(untagged)]
+enum CreateEmbeddingResponses {
```
itowlson: Again puzzled by the plural

crates/llm-remote-http/src/open_ai/schemas.rs
itowlson, line 1: What is it the schema for?  Is it OpenAI-specific?  All these message types seem a bit randomly scattered across the files: could we clarify the rhyme and reason, e.g. so maintainers know where to put new types?

seun-ja: Yes, it is OpenAI-specific schemas (I'm probably using the word incorrectly). 

I thought having it inside of `open_ai.rs` might just make the file a bit crowded.

itowlson: What I'd suggest:

* Move the file into a new `open_ai` folder
* Add `mod schemas;` to the top of `open_ai.rs`
* Change references in `open_ai.rs` from `crate::schemas` to just `schemas`

This will make the schemas module a sub-module of the `open_ai` module which is a nice pattern for private stuff which would be too clutterful in the main file of the module.

crates/llm-remote-http/src/open_ai/mod.rs
```diff
@@ -0,0 +1,266 @@
+#![allow(dead_code)]
```
itowlson: Where is the dead code?  Why is it needed?  If it is neede3d, could we limit this declaration to specific parts of the module rather than the whole thing (so that we don't lose checking on the whole module)?

crates/llm-remote-http/src/open_ai/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    fn url(&self) -> Url {
+        self.url.clone()
+    }
+}
+
+#[derive(Serialize, Debug)]
+struct CreateChatCompletionRequest {
```
itowlson: I am still unclear on when I should put message types in this file vs when I should put them in `schemas.rs`.

crates/llm-remote-http/src/open_ai/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+#[serde(untagged)]
+enum CreateEmbeddingResponseKind {
+    Success(CreateEmbeddingResponse),
+    Error { error: ResponseError },
+}
+
+#[derive(Deserialize)]
+struct CreateChatCompletionResponse {
+    /// A unique identifier for the chat completion.
+    id: String,
```
itowlson: I am confused.  Is this why you allowed dead code?  If we can't just omit the unused fields, then your previous strategy was right; but I would have thought we could omit them e.g.

```
struct CCCR {
  choices: ...,
  usage: ...,
}
```

But if that doesn't work then the `_id` prefix kludge is better than the dead code kludge.

seun-ja: Deserialization would fail if these fields are not provided, which is why I can't ignore them.

I'd revert to the underscore prefix

itowlson: I don't understand _why_ deserialisation would fail.  By default, if there's no Rust field that matches a JSON field, the JSON field is ignored.  You need to explicitly opt into the "fail on unknown JSON" behaviour, because you never know when the server will add a new field; and as far as I can tell you haven't done that.  (You had me doubting myself but I tested it: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=14e024bd7b123e568dd368bc31dcbe95)

I feel like I'm missing something here - could you clarify the nature of the failure?  Thanks!

seun-ja: You're absolutely right, apologies. 

I was confusing it with some issues I encountered earlier in this PR.

crates/llm-remote-http/src/open_ai/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+            input: data,
+            model,
+            encoding_format: None,
+            dimensions: None,
+            user: None,
+        };
+
+        let url = self
+            .url
+            .join("EMBEDDINGS_ENDPOINT")
```
itowlson:
```suggestion
            .join(EMBEDDINGS_ENDPOINT)
```

This raises some concerns about testing for me: I can't see how this could have worked.

I appreciate automated testing is hard for a service like this but can you confirm that you've tested the current code manually (both OpenAI and the changes to classic) and maybe capture what tests you performed?

seun-ja: Thanks for catching that.

For testing, I manually test using the example provided in this PR [./examples/open-ai-rust]. It only tests for inference, which I still used to test before the last commit. 

For embedded, I also used it but modified it slightly to ensure it worked. Used Ollama locally (their API is modelled closely to OpenAI's).

crates/llm-remote-http/src/open_ai/schemas.rs
```diff
@@ -0,0 +1,243 @@
...
+
+#[derive(Deserialize)]
+struct EmbeddingUsage {
+    prompt_tokens: u32,
+}
+
+impl From<CreateChatCompletionResponse> for wasi_llm::InferencingResult {
+    fn from(value: CreateChatCompletionResponse) -> Self {
+        Self {
+            text: value.choices[0].message.content.clone(),
```
itowlson: Do we need to defend against `choices` being empty?  Or is it guaranteed to have an element?

itowlson (approved):

crates/llm-remote-http/src/open_ai/schemas.rs
seun-ja: While it isn't explicitly stated in the [documents](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices) that there will be at least one, it is expected to always have at least one if it's successful.

I suppose it costs nothing to add a safeguard against such an event.

itowlson (approved):

karthik2804:
I am late here, but to answer
>Is there any way we can automate testing this without necessarily spending money on an OpenAI account for CI, and without writing a mock API server from scratch (which might not be very useful anyway)?

We could technically run ollama with a small model to verify in CI integration tests but whether it is worth it, I am not sure.

Issue #3153, opened 2025-06-11
Debug server port forwarding


lann:
Summarizing [@itowlson's description](https://cloud-native.slack.com/archives/C089NJ9G1V0/p1748916014786839) of the problem:

https://github.com/bytecodealliance/StarlingMonkey/pull/218 started adding support for running a JS debugger in the guest, which requires that the debugger be able to make an outbound socket connection to a debug server (typically on `localhost`). Spin's `allow_outbound_hosts` security mechanism makes this somewhat awkward in practice, because a developer would need to either _always_ allow connections to some local port (whether actively debugging or not) or would have to e.g. edit the `spin.toml` while debugging, which is annoying and error prone.

My proposal (originally [here](https://cloud-native.slack.com/archives/C089NJ9G1V0/p1749151348045449?thread_ts=1749149997.064809&cid=C089NJ9G1V0)):

- Pick a domain namespace like `*.spin-debug.alt` (.alt is a [TLD reserved for this sort of thing](https://www.rfc-editor.org/rfc/rfc9476.html))
- Configure debugger(s) to point at fixed address(es), e.g. `js.spin-debug.alt:9229`
- Teach `spin up` a new flag like `--debug-server js=127.0.0.1:9229`
- Spin resolves `js.spin-debug.alt` to some appropriate [reserved IP](https://en.wikipedia.org/wiki/Reserved_IP_addresses) internally and forwards guest connections to that IP to the configured matching `--debug-server` (note `js` from the `--debug-server` value above)

lann:
I think my only hesitation with this one right now is that I have a vague feeling it could be turned into a more generally useful feature if it wasn't specialized to "debug servers". 🤷

itowlson:
Could you say more about:

> Configure debugger(s) to point at fixed address(es), e.g. `js.spin-debug.alt:9229`

Can we rely on debuggers to be configurable in this way, and if so is there a standard way to configure them?  My mental model of this is a bit foggy, sorry.

PR #3246, opened 2025-08-27
Build profiles

itowlson:
Incomplete draft for discussion.  This diverges from [Till's draft SIP](https://github.com/spinframework/spin/pull/3075) in some ways, mostly cosmetic and open to change; but also somewhat in preferring to constrain variation between profiles.

Example:

```toml
[component.build-profile-test]
source = "target/wasm32-wasip1/release/build_profile_test.wasm"
profile.debug.source = "target/wasm32-wasip1/debug/build_profile_test.wasm"
allowed_outbound_hosts = []
[component.build-profile-test.build]
profile.debug.command = "cargo build --target wasm32-wasip1"
```

This proposal uses a partial approach to the possible "am I running the same profile I built" problem.  When you do a build, it records the built profile in a "last profile built" file.  When you do a run, it checks if the profile being run matches the last built one, and prints a warning if not.  This is certainly far from reliable: if you build both debug and release profiles, and then run both of them, you'll get a spurious warning.

If we're okay with this as a basis for further development, we need to look at what other fields should be allowed to vary by profile, e.g.:

* `allowed_outbound_hosts` - you should be able to specify additional hosts.  I'd be inclined not to say you can specify an entirely separate collection, because the 90% case is going to "I need to do sockets to a debugger."  (Things like "I want to talk to the dev database not the prod database" should be handled by variables.)
* `dependencies` - should be able to replace a dependency, e.g. to use its debug build or a stub/mock.  If I squint I can imagine use cases for extra dependencies, in case a component has debug-only imports, not sure
* `environment` - should be able to replace an EV or set additional EVs
* KV/SQLite/LLM - my inclination that varying these by profile is probably an anti-pattern but I am open to being talked round if someone has a concrete use case - otherwise propose deferring to later
* `variables` - not clear why you would, I guess there could be a knob to control profile-specific behaviour? Propose defer
* `files` - not clear why you would, but I guess I could imagine a special build needing some additional asset?  Again my inclination would be to defer until someone needs it

Possible further features:

* whole components being conditional on the profile - e.g. include KV explorer only in certain build configs
* this implies conditional triggers and I am not sure how much I love where this is going

Anyway here it is for discussion, please be gentle with me

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -16,14 +16,18 @@ use subprocess::{Exec, Redirection};
 
 use crate::manifest::component_build_configs;
 
+const LAST_BUILD_PROFILE_FILE: &str = "last-build.txt";
+const LAST_BUILD_ANON_VALUE: &str = "<anonymous>";
+
 /// If present, run the build command of each component.
 pub async fn build(
     manifest_file: &Path,
+    profile: Option<&str>,
     component_ids: &[String],
     target_checks: TargetChecking,
     cache_root: Option<PathBuf>,
 ) -> Result<()> {
-    let build_info = component_build_configs(manifest_file)
+    let build_info = component_build_configs(manifest_file, profile)
         .await
         .with_context(|| {
             format!(
@@ -54,6 +58,10 @@ pub async fn build(
     // If the build failed, exit with an error at this point.
     build_result?;
 
+    if let Err(e) = save_last_build_profile(&app_dir, profile) {
+        tracing::warn!("Failed to save build profile: {e:?}");
+    }
+
     let Some(manifest) = build_info.manifest() else {
         // We can't proceed to checking (because that needs a full healthy manifest), and we've
         // already emitted any necessary warning, so quit.
@@ -63,6 +71,7 @@ pub async fn build(
     if target_checks.check() {
         let application = spin_environments::ApplicationToValidate::new(
             manifest.clone(),
+            profile,
             manifest_file.parent().unwrap(),
         )
         .await
@@ -90,8 +99,19 @@ pub async fn build(
 /// Run all component build commands, using the default options (build all
 /// components, perform target checking). We run a "default build" in several
 /// places and this centralises the logic of what such a "default build" means.
-pub async fn build_default(manifest_file: &Path, cache_root: Option<PathBuf>) -> Result<()> {
-    build(manifest_file, &[], TargetChecking::Check, cache_root).await
+pub async fn build_default(
+    manifest_file: &Path,
+    profile: Option<&str>,
+    cache_root: Option<PathBuf>,
+) -> Result<()> {
+    build(
+        manifest_file,
+        profile,
+        &[],
+        TargetChecking::Check,
+        cache_root,
+    )
+    .await
 }
 
 fn build_components(
@@ -316,6 +336,69 @@ fn sort(components: Vec<ComponentBuildInfo>) -> (Vec<ComponentBuildInfo>, bool) 
     }
 }
 
+/// Saves the build profile to the "last build profile" file.
+pub fn save_last_build_profile(app_dir: &Path, profile: Option<&str>) -> anyhow::Result<()> {
+    let app_stash_dir = app_dir.join(".spin");
+    let last_build_profile_file = app_stash_dir.join(LAST_BUILD_PROFILE_FILE);
+
+    // This way, if the user never uses build profiles, they won't see a
+    // weird savefile that they have no idea what it is.
+    if profile.is_none() && !last_build_profile_file.exists() {
+        return Ok(());
+    }
+
+    std::fs::create_dir_all(&app_stash_dir)?;
+    std::fs::write(
+        &last_build_profile_file,
+        profile.unwrap_or(LAST_BUILD_ANON_VALUE),
+    )?;
+
+    Ok(())
+}
+
+/// Reads the last build profile from the "last build profile" file.
+pub fn read_last_build_profile(app_dir: &Path) -> anyhow::Result<Option<String>> {
+    let app_stash_dir = app_dir.join(".spin");
+    let last_build_profile_file = app_stash_dir.join(LAST_BUILD_PROFILE_FILE);
+    if !last_build_profile_file.exists() {
+        return Ok(None);
+    }
+
+    let last_build_str = std::fs::read_to_string(&last_build_profile_file)?;
+
+    if last_build_str == LAST_BUILD_ANON_VALUE {
+        Ok(None)
+    } else {
+        Ok(Some(last_build_str))
+    }
+}
+
+/// Prints a warning to stderr if the given profile is not the same
+/// as the most recent build in the given application directory.
+pub fn warn_if_not_latest_build(manifest_path: &Path, profile: Option<&str>) {
+    let Some(app_dir) = manifest_path.parent() else {
+        return;
+    };
+
+    let latest_build = match read_last_build_profile(app_dir) {
+        Ok(profile) => profile,
+        Err(e) => {
+            tracing::warn!(
+                "Failed to read last build profile: using anonymous profile. Error was {e:?}"
+            );
+            None
+        }
+    };
+
+    if profile != latest_build.as_deref() {
+        let profile_opt = match profile {
+            Some(p) => format!(" --profile {p}"),
+            None => "".to_string(),
+        };
+        terminal::warn!("You built a different profile more recently than the one you are running. If the app appears to be behaving like an older version then run `spin up --build{profile_opt}`.");
+    }
+}
+
 /// Specifies target environment checking behaviour
 pub enum TargetChecking {
     /// The build should check that all components are compatible with all target environments.
@@ -343,7 +426,7 @@ mod tests {
     #[tokio::test]
     async fn can_load_even_if_trigger_invalid() {
         let bad_trigger_file = test_data_root().join("bad_trigger.toml");
-        build(&bad_trigger_file, &[], TargetChecking::Skip, None)
+        build(&bad_trigger_file, None, &[], TargetChecking::Skip, None)
             .await
             .unwrap();
     }
@@ -351,7 +434,7 @@ mod tests {
     #[tokio::test]
     async fn succeeds_if_target_env_matches() {
         let manifest_path = test_data_root().join("good_target_env.toml");
-        build(&manifest_path, &[], TargetChecking::Check, None)
+        build(&manifest_path, None, &[], TargetChecking::Check, None)
             .await
             .unwrap();
     }
@@ -359,7 +442,7 @@ mod tests {
     #[tokio::test]
     async fn fails_if_target_env_does_not_match() {
         let manifest_path = test_data_root().join("bad_target_env.toml");
-        let err = build(&manifest_path, &[], TargetChecking::Check, None)
+        let err = build(&manifest_path, None, &[], TargetChecking::Check, None)
             .await
             .expect_err("should have failed")
             .to_string();
@@ -374,9 +457,11 @@ mod tests {
     #[tokio::test]
     async fn has_meaningful_error_if_target_env_does_not_match() {
         let manifest_file = test_data_root().join("bad_target_env.toml");
-        let manifest = spin_manifest::manifest_from_file(&manifest_file).unwrap();
+        let mut manifest = spin_manifest::manifest_from_file(&manifest_file).unwrap();
+        spin_manifest::normalize::normalize_manifest(&mut manifest, None).unwrap();
         let application = spin_environments::ApplicationToValidate::new(
             manifest.clone(),
+            None,
             manifest_file.parent().unwrap(),
         )
         .await
--- a/crates/manifest/src/normalize.rs
+++ b/crates/manifest/src/normalize.rs
@@ -9,9 +9,10 @@ use anyhow::Context;
 /// - Inline components in trigger configs are moved into top-level
 ///   components and replaced with a reference.
 /// - Any triggers without an ID are assigned a generated ID.
-pub fn normalize_manifest(manifest: &mut AppManifest) -> anyhow::Result<()> {
+pub fn normalize_manifest(manifest: &mut AppManifest, profile: Option<&str>) -> anyhow::Result<()> {
     normalize_trigger_ids(manifest);
     normalize_inline_components(manifest);
+    apply_profile_overrides(manifest, profile);
     normalize_dependency_component_refs(manifest)?;
     Ok(())
 }
@@ -107,6 +108,44 @@ fn normalize_trigger_ids(manifest: &mut AppManifest) {
     }
 }
 
+fn apply_profile_overrides(manifest: &mut AppManifest, profile: Option<&str>) {
+    let Some(profile) = profile else {
+        return;
+    };
+
+    for (_, component) in &mut manifest.components {
+        let Some(overrides) = component.profile.get(profile) else {
+            continue;
+        };
+
+        if let Some(profile_build) = overrides.build.as_ref() {
+            match component.build.as_mut() {
+                None => {
+                    component.build = Some(crate::schema::v2::ComponentBuildConfig {
+                        command: profile_build.command.clone(),
+                        workdir: None,
+                        watch: vec![],
+                    })
+                }
+                Some(build) => {
+                    build.command = profile_build.command.clone();
+                }
+            }
+        }
+
+        if let Some(source) = overrides.source.as_ref() {
+            component.source = source.clone();
+        }
+
+        component.environment.extend(overrides.environment.clone());
+
+        component
+            .dependencies
+            .inner
+            .extend(overrides.dependencies.inner.clone());
+    }
+}
+
 use crate::schema::v2::{Component, ComponentDependency, ComponentSource};
 
 fn normalize_dependency_component_refs(manifest: &mut AppManifest) -> anyhow::Result<()> {
@@ -192,6 +231,7 @@ fn ensure_is_acceptable_dependency(
         tool: _,
         dependencies_inherit_configuration: _,
         dependencies,
+        profile: _,
     } = component;
 
     if !ai_models.is_empty() {
@@ -262,7 +302,7 @@ mod test {
         })
         .unwrap();
 
-        normalize_manifest(&mut manifest).unwrap();
+        normalize_manifest(&mut manifest, None).unwrap();
 
         let dep = manifest
             .components
@@ -302,7 +342,7 @@ mod test {
         })
         .unwrap();
 
-        normalize_manifest(&mut manifest).unwrap();
+        normalize_manifest(&mut manifest, None).unwrap();
 
         let dep = manifest
             .components
@@ -348,7 +388,7 @@ mod test {
         })
         .unwrap();
 
-        normalize_manifest(&mut manifest).unwrap();
+        normalize_manifest(&mut manifest, None).unwrap();
 
         let dep = manifest
             .components
--- a/crates/manifest/src/schema/v2.rs
+++ b/crates/manifest/src/schema/v2.rs
@@ -52,6 +52,25 @@ impl AppManifest {
         }
         Ok(())
     }
+
+    /// Whether any component in the application defines the given profile.
+    /// Not every component defines every profile, and components intentionally
+    /// fall back to the anonymouse profile if they are asked for a profile
+    /// they don't define. So this can be used to detect that a user might have
+    /// mistyped a profile (e.g. `spin up --profile deugb`).
+    pub fn ensure_profile(&self, profile: Option<&str>) -> anyhow::Result<()> {
+        let Some(p) = profile else {
+            return Ok(());
+        };
+
+        let is_defined = self.components.values().any(|c| c.profile.contains_key(p));
+
+        if is_defined {
+            Ok(())
+        } else {
+            Err(anyhow!("Profile {p} is not defined in this application"))
+        }
+    }
 }
 
 /// App details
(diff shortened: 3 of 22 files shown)


lann:
> This diverges from Till's draft SIP in some ways, mostly cosmetic and open to change; but also somewhat in preferring to constrain variation between profiles.

I think I prefer the SIP's TOML structure with everything for component `X` / profile `Y` going under `component.X.profile.Y`. I agree that we should constrain variation between profiles regardless of the structure.

lann:
> I need to do sockets to a debugger.

I think we should do this as a separate feature: https://github.com/spinframework/spin/issues/3153

itowlson:
Updated with reworked manifest:

```toml
[component.build-profile-test]
source = "target/wasm32-wasip1/release/build_profile_test.wasm"
allowed_outbound_hosts = []
[component.build-profile-test.build]
[component.build-profile-test.profile.debug]
source = "target/wasm32-wasip1/debug/build_profile_test.wasm"
build.command = "cargo build --target wasm32-wasip1"
```

I dislike the verbosity and repetition of the way TOML does this.  It takes a lot to make me yearn for JSON but man this comes close.  But the clarify benefit of the grouping overrides the verbosity issue.

(again, will do squashes once we have agreement on the approach)

itowlson:
Done env vars and deps, holding off on allowed hosts: if we can do the special-case thing for debug connections, then I reckon we can defer allowed hosts in general until someone comes up with a use case.

crates/manifest/src/schema/v2.rs
```diff
@@ -394,7 +413,141 @@ pub struct Component {
...
+    /// The command or commands to build the component in a named profile. If multiple commands
+    /// are specified, they are run sequentially from left to right.
+    ///
+    /// Example: `build.command = "cargo build"`
+    ///
+    /// Learn more: https://spinframework.dev/build#setting-up-for-spin-build
+    command: super::common::Commands,
+}
+
+impl Component {
```
lann: This brings to mind a famous quote, by Abraham Lincoln I believe...

`#deep-inside-joke`

crates/manifest/src/schema/v2.rs
lann, line 1: Another "I wonder if it would be better" here:

Would it be reasonable to handle profiles in a new [`normalize_manifest` pass](https://github.com/spinframework/spin/blob/0a7665e480277af59d395399cb580059c5eb5aaa/crates/manifest/src/normalize.rs#L11) (i.e. `normalize_manifest(manifest, profile)`)? It would have the benefit and/or fatal flaw of hiding most of the profile logic from the rest of the downstream code.

crates/manifest/src/schema/v2.rs
itowlson: Yes: the trouble is that we've already rather violated President Lincoln's decree.  We use serialisation types throughout Spin's application logic, and now the choice is major rework of multiple subsystems for no functional benefit, or papering over where we've coupled to a serialisation format that no longer gives us the correctness guarantees we seek.  Making dangerous fields as private as possible and providing safer accessors seems like the lowest risk...

I did explore keeping `Component` as POD and doing something at the application level so that the `components` collection returned pre-profiled components, which would avoid a lot of this nonsense in favour of a single piece of once-and-for-all nonsense.  It didn't really seem to work out, but I can revisit it if you feel that's a better strategy.

There, that will teach you to make light inside jokes.

lann: > It didn't really seem to work out, but I can revisit it if you feel that's a better strategy.

Speak of the devil: https://github.com/spinframework/spin/pull/3246/files/ea595fcfd5df993eb55155409fb56eaf37f4d715#r2320229292

crates/manifest/src/schema/v2.rs
itowlson: I explored this but had trouble getting it to work safely.  This may point to a deeper problem of normalisation not being sufficiently embedded in the loader logic, but again this is a risk of world+dog being able to go "eh I'll just deserialise from TOML."  The loader pipeline goes from TOML to LockedApp but there are things in the dev tooling that care about AppManifests and they tend to bypass the correct load sequence because it's not readily accessible.  But like I say this may be an opportunity to improve the load sequence and make better guarantees, it will just likely be more work.

lann: Could you say more about the troubles here? I can see a few potential problems along those lines but it isn't clear to me where things would be _worse_ by normalizing in `spin-manifest` rather than `spin-loader`.

itowlson: If I remember rightly, the trouble was that some places were deserialising manifests rather than asking a loader to do it for them (because the loader produced a LockedApp and they wanted a Manifest).  And this is hard to stop because deserialisability is a public feature.  You need to make the DTO private and force consumers to go through a loader which can then enforce normalisation.  (But some things _do_ have a legit interest in the raw manifest.  So, choices, choices.)

Let me look again.  I didn't spend much time on it and my memory of what I was trying is not great.

crates/build/src/lib.rs
```diff
@@ -215,6 +232,56 @@ fn construct_workdir(app_dir: &Path, workdir: Option<impl AsRef<Path>>) -> Resul
...
+
+    let latest_build = read_last_build_profile(app_dir);
+
+    let is_match = match (profile, latest_build) {
+        (None, None) => true,
+        (Some(_), None) | (None, Some(_)) => false,
+        (Some(p), Some(latest)) => p == latest,
+    };
+
+    if !is_match {
```
lann: I'm _reasonably_ confident this is equivalent (and less-not-inarguably better):

```suggestion
    if profile != latest_build {
```

crates/build/src/lib.rs
```diff
@@ -215,6 +232,56 @@ fn construct_workdir(app_dir: &Path, workdir: Option<impl AsRef<Path>>) -> Resul
...
+pub fn read_last_build_profile(app_dir: &Path) -> Option<String> {
+    let app_stash_dir = app_dir.join(".spin");
+    let last_build_profile_file = app_stash_dir.join(LAST_BUILD_PROFILE_FILE);
+    let last_build_str = std::fs::read_to_string(&last_build_profile_file).ok()?;
+
+    if last_build_str == LAST_BUILD_ANON_VALUE {
+        None
+    } else {
+        Some(last_build_str)
+    }
```
lann, line 373: Not _inarguably_ better, but another option:


```suggestion
    (last_build_str != LAST_BUILD_ANON_VALUE).then_some(last_build_str)
```

crates/build/src/lib.rs
itowlson: FACEPALM

Thank you!

crates/build/src/lib.rs
itowlson: Maybe it's too much Go programming but I find this a bit obscure and would prefer to keep spelling it out.  Appreciate the education though - just give me a couple of years to internalise it!

crates/build/src/lib.rs
```diff
@@ -215,6 +232,50 @@ fn construct_workdir(app_dir: &Path, workdir: Option<impl AsRef<Path>>) -> Resul
     Ok(cwd)
 }
 
+/// Saves the build profile to the "last build profile" file.
+/// Errors are ignored as they should not block building.
```
lann: Perhaps propagate errors here and log at the call site?

crates/build/src/lib.rs
```diff
@@ -215,6 +232,50 @@ fn construct_workdir(app_dir: &Path, workdir: Option<impl AsRef<Path>>) -> Resul
...
+    _ = std::fs::create_dir_all(&app_stash_dir);
+    let last_build_profile_file = app_stash_dir.join(LAST_BUILD_PROFILE_FILE);
+    _ = std::fs::write(
+        &last_build_profile_file,
+        profile.unwrap_or(LAST_BUILD_ANON_VALUE),
+    );
+}
+
+/// Reads the last build profile from the "last build profile" file.
+/// Errors are ignored.
```
lann: Ditto propagate/log.

crates/build/src/lib.rs
```diff
@@ -15,14 +15,18 @@ use subprocess::{Exec, Redirection};
 
 use crate::manifest::component_build_configs;
 
+const LAST_BUILD_PROFILE_FILE: &str = "last-build.txt";
+const LAST_BUILD_ANON_VALUE: &str = "<anonymous>";
```
lann, line 20: :bike::house: 

```suggestion
const LAST_BUILD_NONE_VALUE: &str = "<none>";
```

crates/build/src/lib.rs
```diff
@@ -215,6 +232,50 @@ fn construct_workdir(app_dir: &Path, workdir: Option<impl AsRef<Path>>) -> Resul
     Ok(cwd)
 }
 
+/// Saves the build profile to the "last build profile" file.
+/// Errors are ignored as they should not block building.
+pub fn save_last_build_profile(app_dir: &Path, profile: Option<&str>) {
+    let app_stash_dir = app_dir.join(".spin");
+    _ = std::fs::create_dir_all(&app_stash_dir);
+    let last_build_profile_file = app_stash_dir.join(LAST_BUILD_PROFILE_FILE);
```
lann: Small nicety: do nothing in the common case:

```suggestion
pub fn save_last_build_profile(app_dir: &Path, profile: Option<&str>) {
    let app_stash_dir = app_dir.join(".spin");
    let last_build_profile_file = app_stash_dir.join(LAST_BUILD_PROFILE_FILE);
    if profile.is_none() && !last_build_profile_file.exists() {
        return;
    }
    _ = std::fs::create_dir_all(&app_stash_dir);
```

src/commands/registry.rs
```diff
@@ -84,9 +89,11 @@ impl Push {
         notify_if_nondefault_rel(&app_file, distance);
 
         if self.build {
-            spin_build::build_default(&app_file, self.cache_dir.clone()).await?;
+            spin_build::build_default(&app_file, self.profile(), self.cache_dir.clone()).await?;
         }
 
+        spin_build::warn_if_not_latest_build(&app_file, self.profile());
```
lann: This seems slightly easier to reason about than relying on the implicit write and read. :shrug: 

```suggestion
        } else {
            spin_build::warn_if_not_latest_build(&app_file, self.profile());
        }
```

src/commands/up.rs
```diff
@@ -191,8 +198,11 @@ impl UpCommand {
         }
 
         if self.build {
-            app_source.build(&self.cache_dir).await?;
+            app_source.build(self.profile(), &self.cache_dir).await?;
         }
+
+        app_source.warn_if_not_latest_build(self.profile());
```
lann: Ditto

```suggestion
        } else {
            app_source.warn_if_not_latest_build(self.profile());
        }
```

crates/manifest/src/normalize.rs
```diff
@@ -103,3 +104,45 @@ fn normalize_trigger_ids(manifest: &mut AppManifest) {
...
+
+fn apply_profile_overrides(manifest: &mut AppManifest, profile: Option<&str>) {
+    let Some(profile) = profile else {
+        return;
+    };
+
+    for (_, component) in &mut manifest.components {
+        let Some(overrides) = component.profile.get(profile) else {
+            continue;
+        };
```
lann, line 119: Should it be an error (or warning) to use a profile that doesn't appear in the manifest?

lann: Oh look there it is just below...

crates/manifest/src/normalize.rs
```diff
@@ -103,3 +104,45 @@ fn normalize_trigger_ids(manifest: &mut AppManifest) {
...
+        for (name, value) in &overrides.environment {
+            component.environment.insert(name.clone(), value.clone());
+        }
+
+        for (reference, value) in &overrides.dependencies.inner {
+            component
+                .dependencies
+                .inner
+                .insert(reference.clone(), value.clone());
+        }
```
lann: I think these ought to work :shrug: 

```suggestion
        component.environment
            .extend(overrides.environment.iter().cloned());

        component.dependencies.inner
            .extend(overrides.dependencies.inner.iter().cloned());
```

(and really in this context it would be completely fine to just `.clone()` the entire `Map`s instead of `.iter().cloned()`)

crates/build/src/lib.rs
itowlson: It's not necessarily "none" (which to me implies no last build).  It's usually "the anonymous profile."

lann: Ah ok that makes sense.

lann:
[This comment](https://github.com/spinframework/spin/pull/3246#discussion_r2404662387) helped clarify the design a bit and brings up a brace of questions:

- Should last profile be tracked per-component? `spin build --component-id <component-id>` could cause this to be different for different components.
- Should build failures be recorded as a separate "last build" value? In general if a build fails you can't know if the build state represents the old profile, the new profile, or something else entirely.

If we don't want to address these problems directly we could maybe mitigate them in common cases by 1) removing `last-build.txt` immediately before builds and 2) not writing `last-build.txt` unless all components were built.

tschneidereit:
I haven't done a real code review, and certainly shouldn't be the one to sign off on all implementation aspects. With that said: this looks great, and I'm very excited we'll soon have profiles support! 🥳

Could you perhaps go through the SIP and comment on any differences between what's proposed there and implemented here? I think that'd make it clearer where we are, and move design discussions, if any, to the right venue.

itowlson:
@lann The "last build" is a heuristic effort to mitigate the "I built Profile A then ran Profile B and none of my changes are showing up" problem discussed on the SIP (without either 1. adopting the "always build" nuclear option or 2. building a whole series of file scans and timestamp checks).  It's certainly not perfect (e.g. I built Profile A, then built Profile B, ran Profile A - I get the warning even though Profile A is still up to date) but the intention is just to give folks a heads up in the commonest mistake case.

My inclination is that if folks are doing selective builds then they're likely thinking about their build-run cycle enough that they'll be conscious of "oh these things might be out of sync," so I'm not greatly inclined to make things more complicated in order to support that.  I'd propose that if users find it to be a problem then we can address it in a follow-up.

The point about failed builds is valid - we could have built 9/10 components in Profile B, then failed on the last 1/10, leaving the file saying Profile A.  Again, though, my inclination is that in a failed build situation you have a mix of what is current anyway, so a gentle warning is likely to be the least of your surprises.  But maybe I'm taking an overly simplistic view here.

itowlson:
I just re-read my previous message and noticed "The point about failed builds is valid" which I realised could come across as "not like your other point," which was not my intention.  Both of your points were good.  Sorry about that.

itowlson:
@tschneidereit Comparison to the draft SIP (https://github.com/spinframework/spin/pull/3075):

* No support for `SPIN_PROFILE` environment variable
* No support for selecting the `debug` profile with `--debug`
* No aliasing of `release` to the anonymous profile
* No support for "profile per component" (`--component-profile`)
* No automatic release build before registry push
* Allows pushing non-default profile
* Allows override of `environment` and `dependencies`
* Profiles are not 'atomic' - all overrides are optional
* Heuristic check for "is this profile's build current"

tschneidereit:
> * No support for `SPIN_PROFILE` environment variable
> * No support for selecting the `debug` profile with `--debug`
> * No aliasing of `release` to the anonymous profile
> * No support for "profile per component" (`--component-profile`)

These all seem okay to omit, yeah.

> * No automatic release build before registry push

This I feel quite strongly is important to do: I'm almost entirely certain that I myself would get this wrong and push debug builds at some point in the future.
> 
> * Allows pushing non-default profile

By "allows" do you mean "will push whatever was built last", or "adds a --profile option to the CLI subcommand"? If the latter, I think that's actively good, in particular for workflows involving test deployments. If the former, see my comment above.

> * Allows override of `environment` and `dependencies`

Excellent!

> * Profiles are not 'atomic' - all overrides are optional

I keep going back and forth on this one, which probably means that either version is okay. Non-atomic profiles are more ergonomic for sure, whereas atomic ones feel a bit less foot-gunny to me. But I can't really articulate all that well why, so let's ignore the FUD.

> * Heuristic check for "is this profile's build current"

We should make sure that that heuristic is well documented somewhere, lest our future selves have no idea what is going on and make incompatible changes.


All in all, I think only the build-before-push thing needs (more than documentation) changes, or more discussion. I can update the SIP for the other things.

itowlson:
> > * Allows pushing non-default profile
> 
> By "allows" do you mean "will push whatever was built last", or "adds a --profile option to the CLI subcommand"? If the latter, I think that's actively good, in particular for workflows involving test deployments. If the former, see my comment above.

It adds a `--profile` option to `spin registry push`.  The default remains to push the anonymous (presumably release) profile.

> > * Profiles are not 'atomic' - all overrides are optional
> 
> I keep going back and forth on this one, which probably means that either version is okay. Non-atomic profiles are more ergonomic for sure, whereas atomic ones feel a bit less foot-gunny to me. But I can't really articulate all that well why, so let's ignore the FUD.

Adding environment and dependency overrides forced non-atomicity.  And then it sorta kinda made sense to me that a profile might involve only overriding a dependency or EV, and not change the binary at all.  I agree it's a matter of taste - if we see it causing problems we can be more forceful about it in the next major version.

tschneidereit:
> And then it sorta kinda made sense to me that a profile might involve only overriding a dependency or EV, and not change the binary at all.

Yeah, I agree that that makes sense. One thing we could consider is to introduce an explicit `inherits` property like cargo has, so one is forced to choose which profile to inherit stuff from.

That does seem like a nice thing to have, honestly. The major issue I'd see is that it only really fully works if it's a required field, as in cargo. Which raises the question of how to refer to the anonymous profile.

And that, in turn, brings me to another point: I think instead of `anonymous` calling that profile `default` might be better? That is what it effectively is, since it's used if not profile is mentioned. And `inherits = "default"` sure looks better to me than `inherits = "anonymous"`

karthik2804:
I am in agreement with @tschneidereit's idea of having an `inherits` field making things clearer and easier to follow. It also makes sense to me to call it "default" instead of "anonymous" for the unnamed profile. 

> >  No automatic release build before registry push. 
> 
> This I feel quite strongly is important to do: I'm almost entirely certain that I myself would get this wrong and push debug builds at some point in the future.

I do not feel as strongly here with deploy/push taking into consideration the `profile` arg and also profile names being arbitrary strings means we would have to special case things.

> Allows override of environment and dependencies

I feel like allowing outbound hosts would be useful thinking about it from the perspective of the JS debugger. I see the suggested approach is https://github.com/spinframework/spin/issues/3153 but to me that feels like more things for the user to setup every time when they need to debug a component. They would need to setup the debugger port as well as also the profile to choose the binary built with debugging enabled. 

I see an argument as to why we may want to enable overriding variables in profiles in case where we want different defaults. This does not seem super important because the variables can be overriden with environment variables anyway but I think it is a nice to have.

tschneidereit:
> profile names being arbitrary strings means we would have to special case things.

Can you say more about what we'd need to special-case? I'm imagining something like `spin deploy --profile=gofastest`, where the `default` profile is used if `--profile` isn't specified. This would then internally run `spin build --profile=gofastest` before actually deploying.

If we end up doing this, we should definitely make it possible to skip, with something like `--skip-build`.

karthik2804:
I may have misunderstood the intent here. I interpreted the "automatic release build" to imply that there would be a specific release profile (which is what would required special-casing). I do not have a strong opinion on if we should require build before deploy but that seems like a problem that can be punted as it does not change existing behavior where deploy does not build.

tschneidereit:
> that seems like a problem that can be punted as it does not change existing behavior where deploy does not build.

I don't think it can be, because the introduction of profiles changes the picture. Yes, it's currently possible to accidentally deploy a stale build. I think that's not ideal and we probably should already be doing an automatic build. But with profiles the failure mode is very different: the developer might be ensuring very carefully that they do a build, but accidentally use the wrong profile to do so.

One thing I'm just realizing though is that profiles (can, at least) specify different locations for their outputs. It seems likely that they would in practice, too. That means the main risk is that the likelihood of deploying stale builds increases, not that a debug build might accidentally be deployed—which would be much worse.

I still think an automatic build would be very highly desirable, but this realization at least allays my biggest fears.

karthik2804:
>One thing I'm just realizing though is that profiles (can, at least) specify different locations for their outputs. It seems likely that they would in practice, too. That means the main risk is that the likelihood of deploying stale builds increases, not that a debug build might accidentally be deployed—which would be much worse.

This is what I assumed and hence the lack of strong opinion on my end. I think automatic builds would be good though yes but personally don't feel strong enough to block on it.

itowlson:
`inherits` feels like a _significant_ increase in complexity (topologically sorting profiles, cycle detection, chaining overrides).  Could we articulate what functional benefit justifies that complexity?  Could we consider deferring it until someone needs it?

karthik2804:
To me, it feels like a clean way to describe if a profile is fully isolated or if it is just overriding the default profile.

I do agree on the point of the increase in complexity. Could a middle ground be that `inherit` supports only boolean values and only allows inheriting from the "default" profile.

As I write this now, I realize that this is describing new behavior of profiles being completely isolated and not just overriding the default 😅.

tschneidereit:
I think deferring `inherits` is fine, as long as without it each profile is a blank slate, fwiw. Implicitly inheriting anything at all from _somewhere_ would mean a later introduction of `inherits` is a breaking change.

> Could a middle ground be that `inherit` supports only boolean values and only allows inheriting from the "default" profile.

If we were to special-case the `default` profile as a source of inheritance, I think we shouldn't do it with a boolean, but with the string `"default"` being the only permitted value to begin with.

karthik2804:
I finally played around with this once again and now I have changed my mind about the `inherits`. The current state of things seem ideal. I like the idea of every profile option being an optional override. 

The only feedback I have here is that I would really prefer to be able to override the `allowed_outbound_hosts` specifically coming from the JS debugger usecase. I see that we want to address it [seperately](https://github.com/spinframework/spin/issues/3153) but I am not too convinced about the value from that. To me personally it seems cleaner for it to be enabled by profiles. @lann Can you expand on why we do not want to do it here?

This looks great and I would love for it to land!

lann:
Debugging just seems orthogonal to app configuration to me. I've run debuggers in production a fair number of times to track down particularly wiley bugs.

karthik2804:
I could potentially dream up a usecase where we have build profiles that enable building with different features where said one of the optional features is exporting telemetry data. I want to only allow list the host in the case where the app is built with the `telemetry` feature.  And enabling debugging just happens to be a side benefit. Or is there a strong reason for us to not do this?

lann:
> exporting telemetry data

You mean like wasi-otel, which is configured outside of the manifest? 🙂

lann:
To be clear I'm not opposed to using profiles to enable debugging; I just think #3153 would still be useful.

One incremental change we could make here would be to add support for an `allowed_outbound_hosts` "no-op" value, e.g. `allowed_outbound_hosts = [""]`, which you could then use as your default `debug_host` variable. Technically you could get this effect today by using an unroutable IPv6 "black hole" address like `100::1`, though that might be more confusing.

itowlson:
The reason for not doing AOH is that we primarily want this variation to be driven by variables, so that the component talks to the host given in the variable instead of needing to have separate logic.

But debugging is a great example of when that doesn't work - it happens below the component level.

I'd be wary of making AOH overridable for this, because if means that either:

1. there's a lot of repetition if there are multiple hosts; or
2. the profile value needs to merge instead of replace, which is different behaviour from other fields.

I somewhat like Lann's suggestion of variable values which resolve to 'ignore this', so you could have a single list of AOH in the base component, with `{{ debug_host }}` resolving to the "ignore" string or localhost per profile.  I'm not sure empty string works for this, though, because I'm not sure if we can reliably tell this apart from "variable is not set."

I suspect there is also some fiddliness around application variable default values not currently being profileable, at least I don't think they are, which is what you'd want to make this transparent.  (I may be confusing myself though.)

karthik2804:
Lann's suggestion is interesting. Even without modifying what is acceptable in thefield, I think we can make it work by using if clauses using the liquid templating engine we already have. I'll try verify that and if it works, I'm happy to punt on allowed outbound hosts.

itowlson:
I don't think AOH uses Liquid templating.

itowlson:
(force push was just a rebase, and profilifying target env signatures which didn't exist when I started this PR)

karthik2804:
> I don't think AOH uses Liquid templating.
You are right I was misremembering. 

I like the approach @lann suggested as well.

itowlson:
> I like the approach @lann suggested as well.

There are a couple of Lann suggestions floating around - could you clarify which one you are proposing?

karthik2804:
@itowlson I like the one where we allow `allowed_outbound_host` to have a nullable value which would mean we can punt on making that field profileable.

itowlson:
> @itowlson I like the one where we allow `allowed_outbound_host` to have a nullable value which would mean we can punt on making that field profileable.

AOH uses variables from the application, not the component, and application-level stuff is not currently profileable.  So we would need to either:

1. Have users pass in the debugger host value via `-v` (e.g. `spin up --profile debug -v debug_host="tcp://127.0.0.1:*"`); or
2. Support profiling application variables (so that the value can be defaulted differently).

The first one is easy and mostly-safe from a design point of view (the only work needed is to make AOH ignore empty strings, which I am suspicious of but could go along with).  But it's a bit of a faff for users!

The second is more convenient for users, but requires another area where we have to consult the profiles system.  I am wary of doing this for the specific purpose of JS debugging, especially when it only provides us with a rather kludgy solution (and one where it becomes tied to the profile which, as Lann notes, is not necessarily desirable).

My current preference is 3. let's not hold the PR up on JS debugging.  We can add AOH empty string forgiveness and/or app variable profiling as follow-up work if we want them: they would be non-breaking.  But if we feel this is something we need to solve before we can land the PR, then I'm happy to entertain options!

karthik2804:
@itowlson I am in agreement, let us on punt on the  design for enabling debugger to a future PR.

itowlson:
What is required to land this?

PR #3247, opened 2025-08-27
outbound-http: Improve test assertion logging

lann:
I give up; here's some better logging instead.

Sneakily introduce assert_matches! and assert_not_matches! macros.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8517,6 +8517,7 @@ dependencies = [
  "reqwest 0.12.9",
  "rustls 0.23.18",
  "serde",
+ "spin-common",
  "spin-factor-outbound-networking",
  "spin-factor-variables",
  "spin-factors",
--- a/crates/common/src/assert.rs
+++ b/crates/common/src/assert.rs
@@ -1,0 +1,33 @@
+//! Assertion macros.
+
+/// Asserts that the expression matches the pattern.
+///
+/// This is equivalent to `assert!(matches!(...))` except that it produces nicer
+/// errors.
+#[macro_export]
+macro_rules! assert_matches {
+    ($expr:expr, $pat:pat $(,)?) => {{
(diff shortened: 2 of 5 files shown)


itowlson (approved):
"sneakily"

lann:
> "sneakily"

Did you notice?!

itowlson:
Notice what?

lann:
I have a hunch.

Some systems timeout instead of immediately failing requests to "discard prefix" IPv6 addrs (`100::1`). "I know!", I thought to myself, very cleverly, "I'll set a very low timeout and ignore timeouts for those tests!"

https://github.com/spinframework/spin/blob/d1eb9acd42addcb4782afd0a1f1fbe0654063cb4/crates/factor-outbound-http/tests/factor_test.rs#L141

"None of the tests actually touch the network, so short timeouts can't possibly cause problem!"

OR CAN THEY? (dunn dunn duuuuuunn)

https://github.com/spinframework/spin/blob/d1eb9acd42addcb4782afd0a1f1fbe0654063cb4/crates/factor-outbound-http/tests/factor_test.rs#L83

"So what? That request gets rejected by the private IPs check before it gets sent."

BUT NOT BEFORE THE SYSTEM DNS RESOLVER GETS A WHACK AT IT! (_gasp!_)

https://github.com/spinframework/spin/blob/d1eb9acd42addcb4782afd0a1f1fbe0654063cb4/crates/factor-outbound-http/src/wasi.rs#L201

"But...but...it's just `localhost`! It still won't touch the network!" (_desperation setting in_)

THIS IS A HEAVILY LOADED GITHUB ACTIONS RUNNER! ONCE YOU'VE `await`ED ANYTHING ALL YOUR PLANS ARE FOR NAUGHT!

Anyway its just a theory.

PR #3248, opened 2025-08-27
trigger-http: Fix otel context propagation

lann:
The debug_span! was disabled by default tracing config, preventing otel context from being propagated through it.

### DIFF
--- a/crates/trigger-http/src/wasi.rs
+++ b/crates/trigger-http/src/wasi.rs
@@ -93,7 +93,7 @@ impl HttpExecutor for WasiHttpExecutor<'_> {
             HandlerType::Wagi(_) => unreachable!("should have used WagiExecutor instead"),
         };
 
-        let span = tracing::debug_span!("execute_wasi");
+        let span = tracing::info_span!("execute_wasi");
         let handle = task::spawn(
             async move {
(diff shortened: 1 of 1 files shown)


lann:
I made a hot mess of two parallel branches...

calebschoepp (commented):
image

This span was debug intentionally so that it wouldn't pollute the traces developers see. 	`execute_wasi` is redundant with `execute_wasm_component foo`.

Is there another way we could solve this problem?

lann:
We could try removing it, just do in_current_span 🤷

calebschoepp:
@lann were you going to try that or were you wanting me to try that?

lann:
Either is fine with me I just might not get to it immediately...

calebschoepp:
Tested `in_current_span()`. We correctly get the only two spans we want. I still need to test if this fixes the issue Andrew was experiencing.

image

calebschoepp:
~Did some testing and it seems like neither the info span or `in_current_span` fix actually do anything to help with the issue Andrew was observing of the otel context not being propagated.~

Actually I was just testing it wrong. Both ways work so I propose we go forward `in_current_span`.

lann:
#3260

PR #3250, opened 2025-08-28
Static HTTP responses

### LINKED ISSUE #3242: proposal: Static responses for http triggers
A few use cases:

- `/robots.txt` and similar small "well known" URLs
- Temporarily blocking a specific page for maintenance, etc.
- Simple redirects (a subset of what https://github.com/spinframework/spin-redirect can do)

Examples:

```toml
[[trigger.http]]
route = "/robots.txt"
static_response = {
    # Could be optional, defaulting to 200
    status_code = 200
    body = "User-agent: *\nDisallow: "
}

# Oops I had a typo in my newsletter
[[trigger.http]]
route = "/my_blorg_post"
static_response = {
    status_code = 301
    location = "/my_blog_post"
}
```

itowlson:
Fixes #3242.  Example `spin.toml` fragment:

```toml
[[trigger.http]]
route = "/test"
static_response = { body = "hello!\n\nand more hello!\n" }

[[trigger.http]]
route = "/fermyon-me-up"
static_response = { status_code = 302, headers = { location = "https://www.fermyon.com" } }
```

This is, in some ways, a bit unwieldy, because the assumption that a trigger maps to a component is fairly deeply embedded.  We seem to need a new handler type that can never be executed, and there is some faff around assigning a fake component ID, because everything currently depends on component IDs for lookup.  If folks feel we should attempt a larger refactor then I can give it a go before fleeing into the hills.

An implication of this is that applications with static responses can't reliably interact with selective deployment (`spin up -c`).  I believe that `spin up -c` retains triggers associated with the selected components, and static response triggers are associated with no component.  An alternative design that would avoid this would be to have static responses as a pseudo-component type rather than a trigger behaviour (that is, the trigger still points to a named component, but the component is... not a component! ha ha!) - however, this feels extremely can-of-worms-tastic, as the component schema is embedded fairly deeply through Spin.  Again, I can investigate it if that's something we want to explore.

I did consider synthesising an in-memory (or temporary filesystem) component that would respond with the static response, but that seemed like it just created large amounts of hairiness elsewhere.  However, it would allow applications that used static responses to be deployed to existing hosts.

As it is, I think we need a `host_requirements` entry in the lockfile to ensure that hosts understand the new field.  (And this field is app level because we don't have trigger-level `must_understand`, so it won't be bypassed by selective deployment, even though selective deployment exterminates static response triggers.)  Although maybe I am wrong because this is a schema change so hosts will automatically fail by syntax rather than needing to be warned about semantics?  I've put it in a separate commit in case we can back it out.

Draft because I need to write tests because I always forget to do those first, but ready for initial feedback on the approach and on the host-requirements stuff.

### DIFF
--- a/crates/http/src/config.rs
+++ b/crates/http/src/config.rs
@@ -2,16 +2,29 @@ use serde::{Deserialize, Serialize};
 use spin_http_routes::HttpTriggerRouteConfig;
 
 /// Configuration for the HTTP trigger
-#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct HttpTriggerConfig {
     /// Component ID to invoke
-    pub component: String,
+    pub component: Option<String>,
+    /// Static response to send
+    pub static_response: Option<StaticResponse>,
     /// HTTP route the component will be invoked for
     pub route: HttpTriggerRouteConfig,
     /// The HTTP executor the component requires
     #[serde(default)]
     pub executor: Option<HttpExecutorType>,
+}
+
+impl HttpTriggerConfig {
+    pub fn lookup_key(&self, trigger_id: &str) -> anyhow::Result<crate::routes::TriggerLookupKey> {
+        match (&self.component, &self.static_response) {
+            (None, None) => Err(anyhow::anyhow!("Triggers must specify either component or static_response - {trigger_id} has neither")),
+            (Some(_), Some(_)) => Err(anyhow::anyhow!("Triggers must specify either component or static_response - {trigger_id} has both")),
+            (Some(c), None) => Ok(crate::routes::TriggerLookupKey::Component(c.to_string())),
+            (None, Some(_)) => Ok(crate::routes::TriggerLookupKey::Trigger(trigger_id.to_string())),
+        }
+    }
 }
 
 /// The executor for the HTTP component.
@@ -62,6 +75,32 @@ impl Default for WagiTriggerConfig {
             entrypoint: WAGI_DEFAULT_ENTRYPOINT.to_owned(),
             argv: WAGI_DEFAULT_ARGV.to_owned(),
         }
+    }
+}
+
+/// A static response to be served directly by the host
+/// without instantiating a component.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct StaticResponse {
+    #[serde(default)]
+    status_code: Option<u16>,
+    #[serde(default)]
+    headers: indexmap::IndexMap<String, String>,
+    #[serde(default)]
+    body: Option<String>,
+}
+
+impl StaticResponse {
+    pub fn status(&self) -> u16 {
+        self.status_code.unwrap_or(200)
+    }
+
+    pub fn headers(&self) -> impl Iterator<Item = (&String, &String)> {
+        self.headers.iter()
+    }
+
+    pub fn body(&self) -> Option<&String> {
+        self.body.as_ref()
     }
 }
 
--- a/crates/routes/src/lib.rs
+++ b/crates/routes/src/lib.rs
@@ -21,8 +21,8 @@ pub struct Router {
 /// What a route maps to
 #[derive(Clone, Debug)]
 struct RouteHandler {
-    /// The component ID that the route maps to.
-    component_id: String,
+    /// The handler identifier (typically component ID) that the route maps to.
+    lookup_key: TriggerLookupKey,
     /// The route, including any application base.
     based_route: Cow<'static, str>,
     /// The route, not including any application base.
@@ -30,6 +30,25 @@ struct RouteHandler {
     /// The route, including any application base and capturing information about whether it has a trailing wildcard.
     /// (This avoids re-parsing the route string.)
     parsed_based_route: ParsedRoute,
+}
+
+/// An identifier that can be returned from a RouteMatch and used to look up the trigger
+/// that handles the route.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub enum TriggerLookupKey {
+    /// The route is handled by the specified Wasm component.
+    Component(String),
+    /// The route is handled directly within the specified trigger.
+    Trigger(String),
+}
+
+impl std::fmt::Display for TriggerLookupKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Component(id) => f.write_str(id),
+            Self::Trigger(id) => f.write_str(id),
+        }
+    }
 }
 
 /// A detected duplicate route.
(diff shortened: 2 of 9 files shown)


lann:
> As it is, I think we need a host_requirements entry in the lockfile to ensure that hosts understand the new field.

I think we're OK here; I'm reasonably confident that `HttpTriggerConfig` has had `deny_unknown_fields` at least since manifest v2, and even if it didn't it would be invalid without a `component`.

crates/http/src/config.rs
```diff
@@ -2,16 +2,39 @@ use serde::{Deserialize, Serialize};
...
-    #[serde(default)]
-    pub executor: Option<HttpExecutorType>,
+    /// How to handle requests to this route
+    #[serde(flatten)]
+    pub handler: HttpTriggerHandler,
+}
+
+/// Configuration for the HTTP trigger
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case", untagged, deny_unknown_fields)]
```
lann: Serde claims that any combination of `deny_unknown_fields` and `flatten` is "not supported" (read: "buggy"). Some alternatives:

- Duplicate `route` in the two variants
- BYO `Deserialize` for `HttpTriggerRouteConfig`
- Go back to a single struct with `component: Option<String>` and `static_response: Option<...>` and validate that exactly one is set later in the trigger

lann:
Based on gazing at the code for a bit my hunch is that having no component associated with the trigger will ultimately be best.

- The way we map components to triggers is already ambiguous today; e.g. you _should_ be able to have two wagi triggers with different `argv`s pointing at the same component but currently that wouldn't work. The `Router` probably ought to map routes to the trigger config rather than the component directly; this refactor doesn't look _too_ bad but I have only gazed...

- Given the intended uses of selective deployments I think it would be fine to retain static response triggers unconditionally. If something _really_ needs selective static responses I don't think it would be _too_ hard to add support for selective deployment of triggers, i.e. `spin up --trigger-id`.

- I think I may have previously suggested adding a new `HttpExecutorType` for static responses but I would now like to recant. Instead I would suggest handling these somewhere in the [vicinity of `handle_trigger_route`](https://github.com/spinframework/spin/blob/996d50ac6a9289cc1819a47bcd88b0da33c568c0/crates/trigger-http/src/server.rs#L263-L285).

itowlson:
> Instead I would suggest handling these somewhere in the [vicinity of handle_trigger_route](https://github.com/spinframework/spin/blob/996d50ac6a9289cc1819a47bcd88b0da33c568c0/crates/trigger-http/src/server.rs#L263-L285).

I think that's what I do but I think I needed a placeholder executor type because that stuff gets set up before I can arrive there... maybe I'm wrong, I did go through some iterations and might have ended up with dead code

lann:
> I think that's what I do but I think I needed a placeholder executor type because that stuff gets set up before I can arrive there

If you have a dummy component ID then yeah you would need a matching executor; that whole comment apples only to an alternate-universe no-component-id PR.

itowlson:
I have backed out the host requirement, reverted to using Options for the trigger config, and used a type for the trigger lookup key rather than synthesising a fake component ID.  (The RouteMatch can't return the TriggerConfig directly because circularity.  We could have it return the trigger ID rather than the component ID, but at the moment we use the route match key for printing component names against routes, so some refactoring would be needed there.)

Kept it as two commits for now, will squash if this looks less unacceptable.

crates/http/src/config.rs
```diff
@@ -2,18 +2,31 @@ use serde::{Deserialize, Serialize};
...
+    pub static_response: Option<StaticResponse>,
     /// HTTP route the component will be invoked for
     pub route: HttpTriggerRouteConfig,
     /// The HTTP executor the component requires
     #[serde(default)]
     pub executor: Option<HttpExecutorType>,
 }
 
+impl HttpTriggerConfig {
+    pub fn lookup_key(&self, trigger_id: &str) -> anyhow::Result<crate::routes::TriggerLookupKey> {
```
lann, line 20: > SERIALISATION TYPES SHOULD BE FOR 1. SERIALISING 2. DESERIALISING 3. NOTHING ELSE AND THE MOST IMPORTANT OF THESE IS NOTHING ELSE

– Abraham Lincoln (I think)

crates/trigger-http/src/server.rs
```diff
@@ -284,15 +299,48 @@ impl<F: RuntimeFactors> HttpServer<F> {
...
-            component_id = component_id
+            component_id = lookup_key.to_string()
         );
 
+        let trigger_config = self.component_trigger_configs.get(lookup_key).unwrap();
+
+        match (&trigger_config.component, &trigger_config.static_response) {
+            (Some(component), None) => {
+                Self::respond_wasm_component(
+                    self,
```
lann: This _should_ work?
```suggestion
                self.respond_wasm_component(
```

crates/trigger-http/src/server.rs
```diff
@@ -121,22 +120,16 @@ impl<F: RuntimeFactors> HttpServer<F> {
```
lann, line 120: I wonder if it would be easier to partition into `component_trigger_configs` and `static_trigger_configs` here?

crates/trigger-http/src/server.rs
itowlson: I swear this did not work when I first tried it.  Obviously it has been fixed in some sort of stealth update since two days ago.  Thanks for the prod!

crates/trigger-http/src/server.rs
itowlson: @lann I tried this in the "Plan C" commit.  I'm not sure it's easier.  It does front load some of the fiddliness, but I can't imagine it's a big perf benefit, and in terms of code clarity I think it just moves complexity around.  But I've put it up so you can take a look and let me know if this is what you were envisaging and if so how you feel about it.

(I realise "Plan C" still needs a bit of polishing if we decide to go this route.)

lann: This suggestion was just a hunch. If it isn't clearly better then pick your favorite.

crates/http/src/config.rs
```diff
@@ -65,6 +78,32 @@ impl Default for WagiTriggerConfig {
     }
 }
 
+/// A static response to be served directly by the host
+/// without instantiating a component.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct StaticResponse {
```
lann, line 84: Should we require at least one of these to be set? It's certainly _logically valid_ to use all defaults but seems unlikely to be intentional.

crates/routes/src/lib.rs
```diff
@@ -419,11 +440,46 @@ impl<T: Into<String>> From<T> for HttpTriggerRouteConfig {
 mod route_tests {
     use super::*;
 
+    // This is a bit lazy but it saves a lot of faff in the conventional case
+    impl From<&str> for TriggerLookupKey {
+        fn from(value: &str) -> Self {
+            component_key(value)
+        }
+    }
```
lann: This is _pretty_ surprising when reading the tests. Maybe you should be even lazier?

```suggestion
    fn test_router(component_routes: impl IntoIterator<Item = (&str, &str)>) -> Router {
        ...
    }
```

crates/http/src/config.rs
itowlson: My inclination is to punt this: there are all sorts of things that are unlikely to be intentional, and this seems like a low-likelihood omission which will be spotted as soon as you use the route.

(I admit bias because I started looking at it and it turned into more of a faff than I expected, although some of that may have been overthinking.)

PR #3255, opened 2025-08-28
Workflow to bump rust templates sdk dependency

fibonacci1729:
The other half of https://github.com/spinframework/spin-rust-sdk/pull/84 so we don't forget to bump the rust templates when an sdk is released. A good follow up would be to extend this to support other languages (alternatively we could have one of these for the finite set of sdk languages we support).

### DIFF
--- a/.github/workflows/rust-sdk-release.yml
+++ b/.github/workflows/rust-sdk-release.yml
@@ -1,0 +1,39 @@
+name: Bump Spin Rust Templates SDK Dependency
+
+on:
+  repository_dispatch:
+    types:
+      - rust-sdk-release
+
+jobs:
+  create-pr:
+    name: Create PR with Spin Rust Templates SDK Dependency Bump
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Bump Rust Templates SDK Dependency
+        shell: bash
+        run: ./scripts/bump-rust-template-sdk.sh ${{ github.event.client_payload.version }}
+
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.PASSPHRASE }}
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          commit-message: "chore(rust-templates): bump Spin Rust SDK to ${{ github.event.client_payload.version }}"
+          title: "chore(rust-templates): bump Spin Rust SDK to ${{ github.event.client_payload.version }}"
+          body: Update the Spin Rust Templates SDK dependency to ${{ github.event.client_payload.version }}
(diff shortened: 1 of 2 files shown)


scripts/bump-rust-tpl-sdk-dep.sh
itowlson, line 1: I would be in favour of spelling this out e.g. `bump-rust-template-sdk`.  Nobody is typing it: there is no value in saving characters.

Signed, someone who grew up with 8.3 file names

scripts/bump-rust-tpl-sdk-dep.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
itowlson: Brief comment at the top saying what it does?

scripts/bump-rust-tpl-sdk-dep.sh
```diff
@@ -0,0 +1,34 @@
...
+# Ensure version is an 'official' release
+if [[ ! "${VERSION}" =~ ^[0-9]+.[0-9]+.[0-9]+$ ]]
+then
+  echo "VERSION doesn't match [0-9]+.[0-9]+.[0-9]+ and may be a prerelease; skipping."
+  exit 1
+fi
+
+
+# Update the version in the Cargo.toml.tmpl files for each Rust template
+# find templates -type f -path "templates/*-rust/content/Cargo.toml.tmpl" -exec $SED_INPLACE "s/spin-sdk = \".*\"/spin-sdk = \"$VERSION\"/" {} +
```
itowlson: still needed? or is this WIP

itowlson (approved):

.github/workflows/rust-sdk-release.yml
```diff
@@ -0,0 +1,39 @@
...
+      - uses: actions/checkout@v4
+
+      - name: Bump Rust Templates SDK Dependency
+        shell: bash
+        run: ./scripts/bump-rust-template-sdk.sh ${{ github.event.client_payload.version }}
+
+      - name: Import GPG key
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
```
vdice, line 22: Smol nit on the secret keys in this workflow.  I know we've used these same names before ([ha!](https://github.com/spinframework/homebrew-tap/blob/main/.github/workflows/bump-spin.yml)), but wondering if we should take the chance to rename to be more helpful in determining their use at a glance?  What do you think about:

`SPINFRAMEWORKBOT_GPG_PRIVATE_KEY`
`SPINFRAMEWORKBOT_GPG_PRIVATE_KEY_PASSPHRASE`
`SPINFRAMEWORKBOT_PAT` <-- not sure if we should make this even more specific for this workflow if we might need different PATs w/ different perms in the future, though this is probably over-thinking; we can always revisit as needed.

vdice: Or `PULL_REQUESTER`/`PR_AUTHOR` (or similar) if `SPINFRAMEWORKBOT` is too specific (eg maybe for testing on forks 😉)?

fibonacci1729: I agree but would vote for doing this as a follow up; would you mind filing an issue?

vdice (approved):

PR #3257, opened 2025-08-29
Add spin-trigger-cron to release notes for tracking

fibonacci1729:


fibonacci1729:
I'll follow up with releasing the cron trigger tomorrow.

PR #3259, opened 2025-08-29
Adds support for special case : with test

### LINKED ISSUE #3256: manifest: ai_model parsing too restrictive
Validation of llm model entries supplied in the `ai_models` configuration of a spin manifest is too restrictive, preventing use of legitimate models.

Some examples:

```
$ spin build
Building component open-ai-rust with `cargo build --target wasm32-wasip1 --release`
    Finished `release` profile [optimized] target(s) in 0.17s
TOML parse error at line 18, column 13
   |
18 | ai_models = ["gpt-oss:20b"]
   |             ^^^^^^^^^^^^^^^
'-'-separated words may only contain alphanumeric ASCII; got ':'
```
```
$ spin build
Building component open-ai-rust with `cargo build --target wasm32-wasip1 --release`
    Finished `release` profile [optimized] target(s) in 0.17s
Warning: The manifest has errors not related to the Wasm component build. Error details:
TOML parse error at line 18, column 13
   |
18 | ai_models = ["gpt-4"]
   |             ^^^^^^^^^
'-'-separated words must start with an ASCII letter; got '4'
```

Granted, Spin must first support a model before it can actually be used, but this issue blocks applications from using them when/if support is added (eg see @seun-ja's work in https://github.com/spinframework/spin/pull/3238 re: these particular `gpt*` models)

seun-ja:
Aims to resolve #3256

Relaxes rules to accommodate LLM models. Also includes tests and updates current ones

### DIFF
--- a/crates/manifest/src/schema/v2.rs
+++ b/crates/manifest/src/schema/v2.rs
@@ -797,6 +797,21 @@ mod tests {
     }
 
     #[test]
+    fn test_valid_kebab_ids() {
+        for valid in [
+            "default",
+            "mixed-case-words",
+            "letters1-then2-numbers345",
+            "gpt-oss:20b",
+            "gpt-4",
+        ] {
+            if let Err(err) = KebabId::try_from(valid.to_string()) {
+                panic!("{valid:?} should be value: {err:?}");
+            }
+        }
+    }
+
+    #[test]
     fn test_invalid_snake_ids() {
         for invalid in [
             "",
--- a/crates/manifest/tests/ui/invalid/bad_kebab_component_ref.err
+++ b/crates/manifest/tests/ui/invalid/bad_kebab_component_ref.err
@@ -2,5 +2,5 @@ TOML parse error at line 7, column 13
   |
 7 | component = "1-2-3"
   |             ^^^^^^^
-'-'-separated words must start with an ASCII letter; got '1'
+'-'-separated words must start with an ASCII letter or digit; got '1'
 
--- a/crates/manifest/tests/ui/invalid/bad_variable_name.err
+++ b/crates/manifest/tests/ui/invalid/bad_variable_name.err
@@ -2,4 +2,4 @@ TOML parse error at line 7, column 1
   |
 7 | "bad+name" = { required = true }
   | ^^^^^^^^^^
-'_'-separated words may only contain alphanumeric ASCII; got '+'
+'_'-separated words may only contain alphanumeric ASCII (plus ':' inside words); got '+'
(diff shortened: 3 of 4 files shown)


vdice (commented):
I'm not sure if this is the ideal way to go about the fix.  Here's my initial thought:

With the current version, all keys/values in the manifest are now exempt from the exceptions added here.  This is problematic specifically for Spin application names (and maybe for other keys; I haven't yet looked in depth).  We have logic elsewhere (eg `spin new` and `spin deploy`) with more restrictive naming rules, which I think we want to continue to uphold.  For instance, admitting the `:` character in an app name would cause its OCI artifact reference to no longer be valid, and `spin deploy` would fail.

I'm not quite sure about the genesis of forbidding a segment from starting with a number, though @lann may recall.  It's possible that is a restriction that could be lifted...

Hopefully @lann (or @itowlson?) can advise here as I don't have a whole lot of experience in this part of the code.  My naive thought was we'd update logic (either in the serde crate here or maybe the toml parser?) to specifically ignore `ai_models` string values.  Or, if ignoring is too lax, basically just apply the exceptions you've added here to these string values, but continuing to preserve the default restrictions for everything else in the manifest.

fibonacci1729:
At the moment, I don't see much value in restricting the IDs of `ai_models` to be kebab-cased (iirc this was put into place in anticipation of value imports in the component model which won't land for a longish time). I think we can just use `String` in lieu of `KebabId` [here](https://github.com/spinframework/spin/blob/main/crates/manifest/src/schema/v2.rs#L376) and remove the compat enforcement [here](https://github.com/spinframework/spin/blob/main/crates/manifest/src/compat.rs#L45).

itowlson:
@seun-ja Unless there's value in having the history (or it's hard to do) I'd suggest squashing the commits here.

fibonacci1729 (approved):
Looks good to me! Maybe just update the title to reflect the new changes.

itowlson:
@seun-ja Could you GPG sign please?  Thanks!

seun-ja:
@itowlson GPG signed.

Out of curiosity, could it be because I squashed my commit that could have made it seem like I didn't GPG sign? The commit even passed the DCO check

itowlson:
@seun-ja Oh, that's possible.  Squashing would likely have retained the signed-off-by text which is the DCO, but if you're manually GPG-signing then it's a new commit and so presumably _waves hands_ a new cryptographic thingummy needs to be whatsited.

If you want, you can turn on automatic GPG signing by setting `commit.gpgsign=true` in your git config - I have this and so I never have to think about whether I need to re-sign after a merge/rebase.

Anyway thanks for doing it, I see the Verified flag and I've queued an auto merge.

PR #3263, opened 2025-09-05
trigger-http: Don't panic on unknown service chaining component ID

lann:
With `allowed_outbound_hosts = ["*.spin.internal"]` we don't have a component ID to validate until the actual service chaining request is attempted. In the non-service-chaining path the request ID must exist because it was routed to, so we `unwrap`ped instead of erroring.


lann:
I want to mitigate this panic now but a more robust approach might be to catch it in the `allowed_outbound_hosts` check: https://github.com/spinframework/spin/issues/3264

itowlson (approved):

PR #3265, opened 2025-09-10
Adds new dynamism metadata for Providers

### LINKED ISSUE #3115: `spin up` should check required variables

seun-ja:
New logic only allows no default variable before runtime if the Provider is of a Dynamic type

Closes #3115

The PR checks if the variable default isn't provided when required. The check is handled by a new hook: VariableSorterExecutorHooks.

Inspired by @itowlson's https://github.com/spinframework/spin/issues/3115#issuecomment-3272520127, new metadata has been added to Providers.


seun-ja:
I noticed something unusual about the `test_vault_config_provider` test

It is meant to target `VaultVariablesProvider`; however, it ends up targeting `StaticVariablesProvider`.

With the new setup, `VaultVariablesProvider` is treated as a `Dynamic` variable provider, while `StaticVariablesProvider` is regarded as a `Static` variable provider. Therefore, even without providing the data in `spin.toml`, the test should still pass because the check should be bypassed, as it is assumed to be available at runtime. When debugging the reason for the initial test failure, I realised the test was targeting `StaticVariablesProvider`.

Is it by design that it was treated as `StaticVariablesProvider`?

cc: @rylev

crates/expressions/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = { workspace = true }
 anyhow = { workspace = true }
 async-trait = { workspace = true }
 futures = { workspace = true }
+serde = { workspace = true }
```
itowlson: I don't think anything we're doing here should need to introduce a `serde` dependency into `expressions`?

crates/expressions/src/provider.rs
```diff
@@ -9,4 +10,17 @@ use crate::Key;
 pub trait Provider: Debug + Send + Sync {
     /// Returns the value at the given config path, if it exists.
     async fn get(&self, key: &Key) -> anyhow::Result<Option<String>>;
+    fn kind(&self) -> &ProviderVariableKind;
```
itowlson: It's probably more conventional to return a value here rather than a reference.

crates/expressions/src/provider.rs
```diff
@@ -9,4 +10,17 @@ use crate::Key;
...
+}
+
+/// The dynamism of a Provider.
+#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
```
itowlson: This shouldn't need to be Deserialize: it's inherent to the provider type, not something we ever read out of TOML or JSON.

crates/expressions/src/provider.rs
itowlson: Technically this doesn't need `&self` for functionality, but I guess maybe it needs it so it can be called polymorphically via a boxed reference?

crates/trigger/src/cli/variable.rs
```diff
@@ -0,0 +1,21 @@
...
+#[spin_core::async_trait]
+impl<F: RuntimeFactors, U> ExecutorHooks<F, U> for VariablesPreparationExecutorHook {
+    async fn configure_app(
+        &self,
+        configured_app: &spin_factors::ConfiguredApp<F>,
+    ) -> anyhow::Result<()> {
+        let variables_factor = configured_app.app_state::<VariablesFactor>()?;
+
+        let expression_resolver = variables_factor.expression_resolver();
+        expression_resolver.pre_runtime_prepare().await?;
```
itowlson: Not sure what `pre_runtime_prepare` means in this context.  My understanding is that you intend to check if variables references are doomed to fail: can we find a name that expresses that?

crates/trigger/src/cli/variable.rs
```diff
@@ -0,0 +1,21 @@
+use spin_factor_variables::VariablesFactor;
+use spin_factors::RuntimeFactors;
+use spin_factors_executor::ExecutorHooks;
+
+/// An executor hook that prepares the variables factor before runtime execution.
+pub struct VariablesPreparationExecutorHook;
```
itowlson: It's not clear to me that there is any "preparation" going on here.  What do you mean by "prepare/preparation" in this context?  Does the word accurately represent that?

crates/variables-azure/src/lib.rs
```diff
@@ -151,6 +153,10 @@ impl Provider for AzureKeyVaultProvider {
             .context("Failed to read variable from Azure Key Vault")?;
         Ok(Some(secret.value))
     }
+
+    fn kind(&self) -> &ProviderVariableKind {
+        &self.kind
```
itowlson: Surely Azure Key Vault is always dynamic?

crates/variables-env/src/lib.rs
```diff
@@ -132,6 +135,10 @@ impl Provider for EnvVariablesProvider {
     async fn get(&self, key: &Key) -> anyhow::Result<Option<String>> {
         tokio::task::block_in_place(|| self.get_sync(key))
     }
+
+    fn kind(&self) -> &ProviderVariableKind {
+        &self.kind
```
itowlson: The environment provider is always static: there's no need for a member variable to track that.

crates/variables-vault/src/lib.rs
```diff
@@ -52,4 +53,8 @@ impl Provider for VaultVariablesProvider {
             Err(e) => Err(e).context("Failed to check Vault for config"),
         }
     }
+
+    fn kind(&self) -> &ProviderVariableKind {
+        &self.kind
```
itowlson: Again, Hashi Vault is always dynamic.  This is not something users can override by saying otherwise: it is in the nature of the provider.

examples/vault-variable-test/runtime_config.toml
```diff
@@ -3,3 +3,4 @@ type = "vault"
 url = "http://127.0.0.1:8200"
 token = "root"
 mount = "secret"
+kind = "dynamic"
```
itowlson: remove

tests/testcases/vault-variables-test/runtime_config.toml
```diff
@@ -3,3 +3,4 @@ type = "vault"
 url = "http://127.0.0.1:%{port=8200}"
 token = "root"
 mount = "secret"
+kind = "dynamic"
```
itowlson: remove

tests/testcases/vault-variables-test/spin.toml
```diff
@@ -6,7 +6,7 @@ trigger = { type = "http" }
 version = "0.1.0"
 
 [variables]
-password = { required = true }
+password = { default = "test_password" }
```
itowlson: This feels like it obviates the point of the test

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,30 @@ impl ProviderResolver {
...
+    /// Prepares the resolver by attempting to resolve all variables, printing warnings for any
+    /// that cannot be resolved.
+    pub async fn pre_runtime_prepare(&self) -> Result<()> {
+        for name in self.internal.variables.keys() {
+            self.check_variable_existence(name).await?;
+        }
+        Ok(())
+    }
+
+    async fn check_variable_existence(&self, key: &str) -> Result<()> {
```
itowlson: `check_variable_exists` reads slightly simpler to me, but I think the name is misleading, because the intent here is not to check if the variable exists, it is to test if the variable _cannot ever_ exist.

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,30 @@ impl ProviderResolver {
...
+        for provider in &self.providers {
+            if provider.kind() == &ProviderVariableKind::Dynamic {
+                continue;
+            }
+
+            match provider.get(&Key(key)).await {
+                Ok(Some(_)) => return Ok(()),
+                Err(_) | Ok(None) => return self.internal.resolve_variable(key).map(|_| ()),
```
itowlson: I am pretty sure the logic here is wrong.  Suppose I have a variable `myvar` and there are two static providers in play.  `myvar` exists in Provider B but not in Provider A.  If Provider A is hit first, it will say `myvar` doesn't exist, even though Provider B would have actually satisfied it.

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,30 @@ impl ProviderResolver {
...
+        for name in self.internal.variables.keys() {
+            self.check_variable_existence(name).await?;
+        }
+        Ok(())
+    }
+
+    async fn check_variable_existence(&self, key: &str) -> Result<()> {
+        for provider in &self.providers {
+            if provider.kind() == &ProviderVariableKind::Dynamic {
+                continue;
```
itowlson: I am pretty sure the logic here is wrong.  If there is _any_ dynamic provider in play, then we _cannot_ assert that the variable is doomed.  We do _not_ go on and try to find a static provider (and we definitely do _not_ fail if no static provider has the variable).

crates/expressions/src/lib.rs
itowlson: It would be good for the PR description to show the output on failure - I am not sure what message this would produce and whether it would be suitable for startup time.

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,30 @@ impl ProviderResolver {
         Ok(PreparedResolver { variables })
     }
 
+    /// Prepares the resolver by attempting to resolve all variables, printing warnings for any
```
itowlson: This documentation appears to be incorrect?  I don't think anything is "prepared" and I don't see where anything is "printed"?

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,30 @@ impl ProviderResolver {
...
+    /// that cannot be resolved.
+    pub async fn pre_runtime_prepare(&self) -> Result<()> {
```
itowlson: Naming

itowlson:
I'm looking at `test_vault_config_provider` and it appears to bring in the Vault provider via runtime config (https://github.com/spinframework/spin/blob/d1eb9acd42addcb4782afd0a1f1fbe0654063cb4/tests/integration.rs#L552, https://github.com/spinframework/spin/blob/main/tests/testcases/vault-variables-test/runtime_config.toml).  Can you say more about what you're seeing that makes you think it's bypassing Vault?

If it's only this:

> With the new setup, VaultVariablesProvider is treated as a Dynamic variable provider, while StaticVariablesProvider is regarded as a Static variable provider. Therefore, even without providing the data in spin.toml, the test should still pass because the check should be bypassed, as it is assumed to be available at runtime.

then that I think could be an error in your validation code (where you check the static providers even if dynamic providers are in play).

You did also mention that in debugging you saw it going to the environment provider - that's not unexpected in itself because the providers are checked in priority order - you should be concerned only if you _don't_ see it going to the Vault provider.

itowlson:
I should maybe expand that penultimate comment.  When a developer makes a change, and an existing test fails (especially one in the area they're changing), they should treat that as "I broke existing behaviour" until proven otherwise.  They should not assume that their changes are correct and have shown up a flaw in the test.

That's not to say that existing tests never have flaws.  They do!  They have loads of flaws!  And sometimes it turns out that the code change _is_ correct and the test had a hidden flaw.  But it's wise to start from a position of "the problem is likely in the new, untested code" before resorting to "fixing" the test.

crates/expressions/src/lib.rs
seun-ja: Yeah, no printing anymore. I initially did a printing, would change that. Thanks

seun-ja:
> Can you say more about what you're seeing that makes you think it's bypassing Vault?

As you rightly pointed out, the validation logic was flawed. So the reason for seeing a Static was an expected occurrence, as it is part of the available Providers. The logic sees Static first and just returns an error, as it couldn't find a variable.

The new logic seems to have resolved that issue and should do what is expected. However, there needs to be some updates to the logic, as some `integration_tests` tests (different from the previous ones). Will continue on that tomorrow.

itowlson:
Before we go further could you write some unit tests please?  For example, I'm guessing the change to explicitly check for default values was motivated by discovering a case where a variable should have been allowed but wasn't.  Which is a good catch... but... there's no test to identify the case and prevent a regression.  Test cases allow the reviewer to look at the covered cases and go either "okay I can see this all works and am reviewing only for maintainability" or "huh but what happens in this weird case, I need to review for correctness."  Thanks.

seun-ja:
@itowlson - I just realised there's no way you could have known I included the unit tests you suggested I add, I'm sorry. Here's informing you.

Thanks for the review!

itowlson:
Thanks for that - it definitely moves things in the right direction.  Unfortunately, the current tests are hard to read and understand because:

* The names are vague.  `validate_variable_existence_successful` - okay I assume this is testing that a "valid" result appears in some circumstance, but what is that circumstance?  Some tests tell me part of the context; others don't.
  * A starting point here is to write the name in if-then format and not worry about the length.  `if_there_is_a_single_static_provider_and_it_does_not_contain_a_required_variable_then_validation_fails`.  `if_there_is_a_dynamic_provider_then_validation_succeeds_even_if_a_static_provider_without_the_variable_is_in_play`.  Then prune those down to focus on the differences.  Or turn them into comments.  Or assertion messages.
* The test code is dominated by largely, but not entirely, boilerplate setup.  This makes it hard to see what is different (and what differences are important) between the tests.  Okay, this test asserts success and this test asserts failure, but what the scenario that each one is covering?
  * Can we pull out common boilerplate into a fixture helper, so that the test code focuses _only_ on the scenario?
  * Can we unit-test in the `expressions` crate rather than needing the whole application setup ceremony involved in using a `VariablesFactor`?
 
In addition, from what I can tell, the tests don't cover many scenarios.  I appreciate that the current format is quite labour-intensive and so I understand why you focused on a few cases.  But for example there seems to be nothing covering "there's more than one provider in play" which is core and which was one of the bugs I think we had in a previous iteration.

My suggestion for moving this forward would be:

* Figure out a way to write the test cases really really simply.
* Then go through the list in https://github.com/spinframework/spin/pull/3265#pullrequestreview-3207620874 and write those cases.

What I'd like the tests to look like is something like:

```
#[tokio::test]
async fn only_static_provider_without_variable_is_invalid() {
  let resolver = make_resolver(
    &[StaticProvider::with_variable("a", "123")],
  );

  assert!(!resolver.may_resolve("b").await);
}

#[tokio::test]
async fn dynamic_provider_and_static_provider_without_variable_is_valid() {
  let resolver = make_resolver(
    &[DynamicProvider, StaticProvider::with_variable("a", "123")],
  );

  assert!(resolver.may_resolve("b").await);
}
```

(you'll probably need an extra knob to set up which vars have defaults, although that could be hardwired?  something to consider in the design)

I'd find that way easier to understand, and so way easier to check if 1. the test is doing what it says it's doing and 2. whether the test suite covers the cases I'm concerned about.

Hope this helps, happy to chat and iterate further if you want to bounce ideas before diving in!

itowlson:
Adding to the above: it may well be worth keeping a factor test or hook test or whatever to verify that the factor or hook or whatever _calls_ the validator.  And that test _will_ likely be quite ceremonious, and that's okay.  But that test doesn't need to cover a variety of cases: it's testing only that the call _happens,_ and can trust the `expressions` unit tests to test that the call _does what we want._  Hope that makes sense...

seun-ja:
Thanks for the feedback, @itowlson. 

Made a new commit that updates the test to
- stay in the `expression` crate
- separated boilerplate, making the test simpler.
- It also covers more possible situations. Following your suggestion [here](https://github.com/spinframework/spin/pull/3265#pullrequestreview-3207620874)

I'd love to receive more feedback on ways to improve this PR. Thank you!

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,260 @@
...
+
+struct StaticProvider {
+    resolver: ProviderResolver,
+}
+
+impl StaticProvider {
+    fn with_variables(
+        variable_key: Option<String>,
+        default: Option<String>,
+        other_providers: Option<Box<dyn Provider>>,
```
itowlson: Why does a provider have other providers?  The ProviderResolver can have multiple providers: there's no need for a Provider to have its own sub-providers.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,260 @@
...
+
+        Self { resolver }
+    }
+}
+
+struct DynamicProvider {
+    resolver: ProviderResolver,
+}
+
+impl DynamicProvider {
```
itowlson: I am confused.  Neither of these providers appears to implement Provider?

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn single_static_provider_with_no_variable_provided_is_valid() -> anyhow::Result<()> {
+    let static_provider = StaticProvider::with_variables(None, None, None);
+
+    static_provider
+        .resolver
+        .validate_variable_existence()
```
itowlson: Oh this validates _all_ variables known to the resolver!  The name implies _one_ variable.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn if_there_is_a_dynamic_provider_and_a_static_provider_then_validation_succeeds_even_if_a_static_provider_with_the_variable_is_in_play(
+) -> anyhow::Result<()> {
+    let dynamic_provider = DynamicProvider::with_variables(
+        Some("baz".to_string()),
+        Some("foo".to_string()),
+        Some(Box::new(StaticMockProvider)),
+    );
```
itowlson: This feels like you are putting a lot of ceremony in the test (which is where you want to avoid ceremony and focus on the scenario and outcome).  You control `::with_variables`: make it take `&str` instead of `Option<String>`.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,260 @@
...
+        .is_err());
+
+    Ok(())
+}
+
+#[derive(Debug)]
+struct StaticMockProvider;
+
+#[spin_world::async_trait]
+impl Provider for StaticMockProvider {
```
itowlson: Oh this is the provider?  I am not sure why we have StaticProvider and StaticMockProvider

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,260 @@
...
+        }
+    }
+
+    fn kind(&self) -> ProviderVariableKind {
+        ProviderVariableKind::Static
+    }
+}
+
+#[derive(Debug)]
+struct ExtraStaticMockProvider;
```
itowlson: ...and ExtraStaticMockProvider!

I'd ask you to think about someone (possibly yourself) trying to maintain this in a few months.  They're not in the moment like you are - they're going to grapple with "okay what is the difference between these three "static provider" things and how do they relate and when do I use one rather than the other."  Naming is a huge part of guiding the reader towards understanding.

(And none of us are perfect at it, because it's hard!)

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,260 @@
...
+    static_provider
+        .resolver
+        .validate_variable_existence()
+        .await?;
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn static_provider_with_two_static_providers_neither_has_data() -> anyhow::Result<()> {
```
itowlson: I'm appreciating the test names - these are much more helpful.

But "static provider with two static providers" mystifies me.  I think what this is talking about is the case where there are two static providers?  So just "two static providers" would do, you don't need another provider to "with" them.

crates/expressions/tests/validation_test.rs
itowlson, line 1: There's already a `tests` module in `expressions/lib.rs` - maybe use that instead of creating a new file?

itowlson: Actually now I look at this I can't figure out how it's getting incorporated.  Or where the conditional compilation directive is.  Is this some Cargo convention I've not learned about?

itowlson (changes requested):
Thanks for this update.  I'm definitely appreciating the new unit tests - lots more coverage, more clearly named, easier to see what's being tested.

But there's still a lot of confusion about other naming, and the tests are still a bit cluttery to read.  Why are there two ways of constructing a resolver?  Why are they called Provider?  Why don't they return the resolver directly?  Why do they take values that are inconvenient to construct?

What I'd suggest you aim for is something like:

```
let resolver = make_resolver(...);
resolver.validate().await
```

and look at how you can keep the `make_resolver` stuff as simple as possible _at the call site_.  E.g.

```
let resolver = make_resolver(
  &[("a", Some("hello")), ("b", None)],  // the variables with their defaults
  &[&StaticProvider::with("b", "something"), &DynamicProvider]  // the providers in play
);

// or

let resolver = make_resolver(&[("a", Some("hello")])
  .with_provider(StaticProvider::...)
  .with_provider(DynamicProvider);
```

And the providers should be actual providers:

```
struct StaticProvider {
  known: HashMap<Key, String>
}

impl Provider for StaticProvider {
  fn get(&self, key) {
    self.known.get(key)  // yep you'll need to massage this
  }
}

impl Provider for DynamicProvider {
  fn get(&self, _) {
    panic!("this should never be called during validation");
  }
}
```

Hope this helps - happy to chat if you want more detail or to explore possible approaches.

crates/expressions/tests/validation_test.rs
seun-ja: I also just realised it's possible. There are a couple of tests in Spin that are similar. Apparently, if the folder is named "tests", Cargo automatically incorporates it.

The reason I created a new file for it was that I thought separating the tests would be better, as the tests in lib, while similar, test for different purposes. They also have a different boilerplate.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    fn with_provider(mut self, provider: Box<dyn Provider>) -> Self {
+        self.providers.push(provider);
+        self
+    }
+
+    fn make_resolver(
+        self,
+        key: Option<&str>,
```
itowlson: This shouldn't need to be an Option.  How about if you did `.with_variable(name, default)` (in the same way as `with_provider`), then `make_resolver()` with no args: that would allow you to specify zero, one or many variables.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,193 @@
...
+struct ResolverTester {
+    providers: Vec<Box<dyn Provider>>,
+}
+
+impl ResolverTester {
+    fn new() -> Self {
+        Self::default()
+    }
+
+    fn with_provider(mut self, provider: Box<dyn Provider>) -> Self {
```
itowlson: I think you can have this take an `impl Provider` instead of a `Box<dyn>` - that would save the caller having to box it at the call site (you would box it inside `with_provider`).

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        Ok(provider_resolver)
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn single_static_provider_with_no_variable_provided_is_valid() -> anyhow::Result<()> {
+    let resolver = ResolverTester::new()
+        .with_provider(Box::new(StaticMockProvider::with_variables("foo", "bar")))
```
itowlson: Just call it `StaticProvider`.  The `Mock` isn't really adding anything in the test context (and "mock" is a testing term of art with a specific meaning that doesn't apply here):

```suggestion
        .with_provider(StaticProvider::with_variable("foo", "bar")))
```

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+}
+
+#[derive(Debug)]
+struct DynamicMockProvider;
+
+#[spin_world::async_trait]
+impl Provider for DynamicMockProvider {
+    async fn get(&self, _key: &Key) -> anyhow::Result<Option<String>> {
+        Ok(None)
```
itowlson: I'd suggest `panic!("validation should never call `get` for a dynamic provider")` so if a later maintainer incorrectly changes the validation routine they get a nice test failure to warn them off!

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,193 @@
...
+    Ok(())
+}
+
+#[derive(Debug)]
+struct StaticMockProvider {
+    variables: HashMap<String, Option<String>>,
+}
+
+impl StaticMockProvider {
+    fn with_variables(key: &str, value: &str) -> Self {
```
itowlson:
```suggestion
    fn with_variable(key: &str, value: &str) -> Self {
```

(only one variable)

Or have it take a slice of tuples but that's going to be a bit noisy!

crates/expressions/Cargo.toml
```diff
@@ -12,5 +12,6 @@ spin-locked-app = { path = "../locked-app" }
 thiserror = { workspace = true }
 
 [dev-dependencies]
+spin-world = { path = "../world" }
```
itowlson: Is this dependency still needed?

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+    providers: Vec<Box<dyn Provider>>,
+    variables: HashMap<String, Variable>,
+}
+
+impl ResolverTester {
+    fn new() -> Self {
+        Self::default()
+    }
+
+    fn with_dynamic_provider(mut self) -> Self {
```
itowlson: The change to bake provider kinds into the resolver seems unlovely - I appreciate the goal of making the tests easier to read but for me this begins to move into then territory of "hiding what we are actually testing."  (It also seems to rule out a category of tests along the lines of "a single provider has one variable but not another".)

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+            },
+        );
+        self
+    }
+
+    fn make_resolver(self) -> anyhow::Result<ProviderResolver> {
+        let mut provider_resolver = ProviderResolver::new(self.variables)?;
+
+        for provider in self.providers {
+            provider_resolver.add_provider(provider as _);
```
itowlson: Is `as _` still needed?

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        Ok(provider_resolver)
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn single_static_provider_with_no_variable_provided_is_valid() -> anyhow::Result<()> {
```
itowlson: Name here is a bit confusing.  The provider provides a variable.  It's more that the _resolver_ has no variables.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        self
+    }
+
+    fn with_static_provider(mut self, key: &str, value: Option<&str>) -> Self {
+        self.providers
+            .push(Box::new(StaticProvider::with_variable(key, value)));
+        self
+    }
+
+    fn with_variable(mut self, (key, default): (&str, Option<&str>)) -> Self {
```
itowlson: No point tupling these when they are method args - this was needed if we were going to put them in a slice (to group key-default pairs) but now it's implicit by them being args to the same method call.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        .with_variable(("bar", None))
+        .make_resolver()?;
+
+    assert!(resolver.validate_variables().await.is_err());
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn if_there_is_a_dynamic_provider_then_validation_succeeds_even_if_a_static_provider_without_the_variable_is_in_play(
```
itowlson: The test name is misleading - no static provider is in play in this code.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        .with_variable(("bar", None))
+        .make_resolver()?;
+
+    resolver.validate_variables().await?;
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn if_there_is_a_dynamic_provider_and_a_static_provider_then_validation_succeeds_even_if_a_static_provider_without_the_variable_is_in_play(
```
itowlson: I feel "even if a static provider is in play" covers the "and a static provider" part of the name.

```suggestion
async fn if_there_is_a_dynamic_provider_then_validation_succeeds_even_if_a_static_provider_without_the_variable_is_in_play(
```

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        .with_variable(("baz", None))
+        .make_resolver()?;
+
+    resolver.validate_variables().await?;
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn if_there_is_a_dynamic_provider_and_a_static_provider_then_validation_succeeds_even_if_a_static_provider_with_the_variable_is_in_play(
```
itowlson: But in this case the static provider does _not_ have the variable?  The resolver needs `baz` and the provider has `foo`.

(There's a case that names that guide the writer/reader might help avoid errors like this - it's hard to have intuition about things called `foo` and `baz`.)

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+        .with_variable(("foo", None))
+        .make_resolver()?;
+
+    resolver.validate_variables().await?;
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn if_there_is_two_static_providers_where_first_provider_does_not_have_data_while_second_provider_does(
```
itowlson: Perhaps a comment to distinguish this from the previous test.  (I know the case here is "we don't barf just because the first one did, we check all providers" and agree it's worth testing - but it may be less obvious to a future reader.)

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,204 @@
...
+    let resolver = ResolverTester::new()
+        .with_static_provider("foo", Some("bar"))
+        .with_static_provider("baz", Some("hay"))
+        .with_variable(("hello", None))
+        .make_resolver()?;
+
+    assert!(resolver.validate_variables().await.is_err());
+
+    Ok(())
+}
```
itowlson, line 229: I think you previously had a test for "providers don't have the data but there is a default value"?  That test is needed.

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,54 @@ impl ProviderResolver {
         Ok(PreparedResolver { variables })
     }
 
+    /// Validates `Provider`(s) that are `ProviderVariableKind::Static` provide their variables at startup.
+    pub async fn validate_variables(&self) -> Result<()> {
```
itowlson: I'd like to change the name away from `validate` because that word is strongly overloaded.  What are we checking here?  That no variable is doomed?

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,54 @@ impl ProviderResolver {
         Ok(PreparedResolver { variables })
     }
 
+    /// Validates `Provider`(s) that are `ProviderVariableKind::Static` provide their variables at startup.
```
itowlson: This comment is misleading.

itowlson (commented):
This is looking good - the checking logic is clear and the tests are readable and give good coverage.  Just a few naming, style and coverage nits.  Thanks!

crates/expressions/Cargo.toml
seun-ja: Yes, here:

https://github.com/seun-ja/spin/blob/9c7c74285555f6c23834ad3e21564a9633764c7a/crates/expressions/tests/validation_test.rs#L195

itowlson: But the `Provider` trait that you're implementing uses `async_trait` directly rather than the `spin-world` re-export: https://github.com/spinframework/spin/blob/main/crates/expressions/src/provider.rs.  Are you _sure_ you need the `spin-world` re-export?

(There is also a re-export from `spin-locked-app` which is already a dependency.)

crates/expressions/src/lib.rs
```diff
@@ -90,6 +92,54 @@ impl ProviderResolver {
         Ok(PreparedResolver { variables })
     }
 
+    /// Ensures that all required variables are resolvable at startup
+    pub async fn ensure_required_variables_resolved(&self) -> Result<()> {
```
itowlson: I'd go with the form in the comment:

```suggestion
    pub async fn ensure_required_variables_resolvable(&self) -> Result<()> {
```

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,259 @@
...
+            Some("super-secret-key"),
+        ))
+        .with_variable("database_host", None)
+        .make_resolver()?;
+
+    resolver.ensure_required_variables_resolved().await?;
+
+    Ok(())
+}
+// This is a bit of an edge case, but we want to ensure that if there are two or more static providers
```
itowlson: It's not really an "edge case": there's no reason to expect that the first provider in line is the one with the answer!  I'd just drop this phrasing:

```suggestion
// Ensure that if there are two or more static providers
```

itowlson:
Sorry to be fussy but the commit message doesn't really reflect the purpose or impact of the commit, only a minor part of the implementation.  Could you change it to e.g. "Check required variables are potentially resolvable" or something similar please?  Thanks.

crates/expressions/tests/validation_test.rs
```diff
@@ -0,0 +1,267 @@
...
+        .with_variable("api_key", None)
+        .make_resolver()?;
+
+    resolver.ensure_required_variables_resolvable().await?;
+
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn if_there_is_a_dynamic_provider_and_a_static_provider_then_validation_succeeds_even_if_a_static_provider_with_the_variable_is_in_play(
```
itowlson:
```suggestion
async fn if_there_is_a_dynamic_provider_and_a_static_provider_then_validation_succeeds_even_if_a_static_provider_without_the_variable_is_in_play(
```

or am I misreading?

itowlson (approved):
One test name looks iffy.  It's not a blocker but if you have a moment to check then that would be awesome.  Thanks!

crates/expressions/tests/validation_test.rs
seun-ja: The wording is confusing, apologies. Although it has a variable, the naming doesn't need to say "`static_provider_with_the_variable`" 

I think `if_there_is_a_dynamic_provider_and_a_static_provider_then_validation_succeeds_even_if_there_is_a_variable_in_play` might be a better name

PR #3266, opened 2025-09-11
Expand wildcard service chaining during `spin up`

### LINKED ISSUE #3264: Consider expanding `*.spin.internal` in `allowed_outbound_hosts`
https://github.com/spinframework/spin/pull/3263 mitigates a panic when a service chaining request is made to invalid host with `allowed_outbound_hosts = ["*.spin.internal"]`. I think it would be more robust to handle this by expanding `["*.spin.internal"]` into `["component-1.spin.internal", "component-2.spin.internal", ...]` at load time.

itowlson:
Fixes #3264.

There are several ways to approach this.  I don't much love this but adopted it in order to preserve the error in app splitting scenarios.

An alternative would be to do it as part of TOML manifest normalisation before we even get to the locking stage: this would result in OCI artifacts having the wildcard pre-expanded which I dunno if it's a good thing or a bad thing or just a thing.  However it would result in app splitting giving an error like "component alice service chains to component bob which doesn't exist," which seems less clear and helpful than "you can't use a wildcard if you're gonna app split."

(Note that even if we did pre-expand OCI artifacts we'd still need to either do this as well, or deal with the possibility of wildcards at run time trigger, because existing OCI artifacts could contain star chains.)

Anyway open to feedback and discussion as ever

### DIFF
--- a/crates/outbound-networking-config/src/allowed_hosts.rs
+++ b/crates/outbound-networking-config/src/allowed_hosts.rs
@@ -465,12 +465,14 @@ impl AllowedHostsConfig {
     pub fn parse<S: AsRef<str>>(
         hosts: &[S],
         resolver: &spin_expressions::PreparedResolver,
+        component_ids: &[String],
     ) -> anyhow::Result<AllowedHostsConfig> {
         let partial = Self::parse_partial(hosts)?;
         let allowed = partial
             .into_iter()
             .map(|p| p.resolve(resolver))
             .collect::<anyhow::Result<Vec<_>>>()?;
+        let allowed = Self::expand_wildcard_service_chaining(allowed, component_ids);
         Ok(Self::SpecificHosts(allowed))
     }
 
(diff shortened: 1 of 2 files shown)


lann:
Another alternative to consider might be to normalize this in `AllowedHostsConfig::parse`, which would require passing in an extra `component_ids` param, but that seems feasible looking at its one ("real") call site.

crates/outbound-networking-config/src/allowed_hosts.rs
```diff
@@ -465,12 +465,14 @@ impl AllowedHostsConfig {
     pub fn parse<S: AsRef<str>>(
         hosts: &[S],
         resolver: &spin_expressions::PreparedResolver,
+        component_ids: &[String],
```
lann, line 468: nit: Can the copy be avoided?
```suggestion
        component_ids: &[&str],
```

itowlson: I haven't yet found a way to do so.  This was what I originally tried.  Unfortunately the compiler doesn't like it because it reckons the `PrepareContext` that the IDs are borrowed from doesn't live long enough (although this may be a spurious report from its _other_ error which is that it results in the PrepareContext having mutable and immutable borrows at the same time).  I tried it again and still no joy.

I don't trust my expertise enough to say the copy _can't_ be avoided.  It sure seems like it _should_ be possible.  But I haven't managed to find a way.  I'll merge as is, but if you do spot a way, I'm ready to learn!

PR #3268, opened 2025-09-11
factor-outbound-http: Refactor wasi impls

lann:
Putting my commits [where my fingers type](https://github.com/spinframework/spin/pull/3229#discussion_r2277782673).

This is split into ~3~ 5 commits to ease review:

- A small test change trying to squash a persistent flake
- Reorganizing `send_request_impl`, inlining `send_request_handler` and making it a method on a new struct that holds most of its inputs. This doesn't do anything functionally but I think it helps distinguish the data that was copied from the instance state and may aid future refactoring.
- Reordering some top level items in the file to better fit (what I think of as) a common convention, roughly "things higher in the file call into things lower in the file".
- Merging the separate `TLS_CLIENT_CONFIG` task-local into `CONNECT_OPTIONS`, makes `connect_tcp` and `connect_tls` methods on `ConnectOptions`, and makes a bunch of miscellaneous simplifications (which I now realize could/should have done in a _fourth_ commit; sorry!).
- Another refactor splitting the `send` method into a few parts to help avoid confusion around error handling.

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -378,8 +378,12 @@ impl ConnectOptions {
 
         let mut socket_addrs = tokio::net::lookup_host(host_and_port)
             .await
-            .map_err(|_| dns_error("address not available".into(), 0))?
+            .map_err(|err| {
+                tracing::debug!(?host_and_port, ?err, "Error resolving host");
+                dns_error("address not available".into(), 0)
+            })?
             .collect::<Vec<_>>();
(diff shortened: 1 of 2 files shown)


lann:
Bisecting, the test failure surprisingly happens in the _first_ commit. :thinking:

lann:
OK I'm pretty sure this is just a subtle change to the way errors got propagated with the nested `Result`s.

lann:
Squashing that bug inspired/compelled me to do another round of refactoring to get rid of the nested `Result`s.

lann:
The last CI run succeeded; I just needed to sign my commits.

PR #3269, opened 2025-09-11
factor-outbound-http: Add InterceptRequest::override_connect_host

lann:
Draft pending merge of https://github.com/spinframework/spin/pull/3268 which it is based on. See the [last commit](https://github.com/spinframework/spin/pull/3269/commits/640087fa48aa8653f32171e2b710234c9f3adecf).

This allows an `OutboundHttpInterceptor` to override the host address used for the outbound socket connection only; it does not otherwise affect e.g. TLS server name checking or the HTTP host header.

### DIFF
--- a/crates/factor-outbound-http/tests/factor_test.rs
+++ b/crates/factor-outbound-http/tests/factor_test.rs
@@ -4,12 +4,14 @@ use anyhow::bail;
 use http::{Request, Uri};
 use spin_common::{assert_matches, assert_not_matches};
 use spin_factor_outbound_http::{
+    intercept::{InterceptOutcome, InterceptRequest, OutboundHttpInterceptor},
     ErrorCode, HostFutureIncomingResponse, OutboundHttpFactor, SelfRequestOrigin,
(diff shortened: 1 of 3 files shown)


crates/factor-outbound-http/tests/factor_test.rs
```diff
@@ -128,9 +158,9 @@ async fn test_instance_state(
 fn test_request_config() -> OutgoingRequestConfig {
     OutgoingRequestConfig {
         use_tls: false,
-        connect_timeout: Duration::from_millis(1),
-        first_byte_timeout: Duration::from_millis(1),
-        between_bytes_timeout: Duration::from_millis(1),
+        connect_timeout: Duration::from_millis(10),
+        first_byte_timeout: Duration::from_millis(10),
+        between_bytes_timeout: Duration::from_millis(10),
```
lann, line 163: I think this has been a source of flake in the past where the system dns resolver occasionally takes >1ms to resolve `localhost`.

PR #3272, opened 2025-09-15
Add WASIp3 support

dicej:
This adds wasi:{cli,clocks,filesystem,http,random,sockets}@0.3.0-rc-2025-08-15 support, including inbound and outbound HTTP requests.

I've smoke tested this manually with a few WASIp3 components and plan to port over some or all of the existing WASIp2 tests.

Note that this builds on https://github.com/spinframework/spin/pull/3271

TODO:
  - Port wasi:http@0.2.x tests
  - Add `Instrument::in_current_span` calls where necessary (e.g. in wasip3.rs)
  - Update to next release candidate once the `release-37.0.0` branch supports it
  - Switch to the final 37.0.0 release once it's available


crates/core/tests/integration_test.rs
```diff
@@ -169,44 +169,3 @@ async fn run_test(
         .0
         .map_err(|()| anyhow::anyhow!("command failed"))
 }
-
-// Write with `print!`, required for test output capture
-struct TestWriter(tokio::io::Stdout);
```
dicej, line 175: Clippy pointed out this was unused, so I removed it.

crates/trigger-http/src/lib.rs
```diff
@@ -153,9 +154,13 @@ impl HttpTrigger {
         }
         if let Some(TriggerMetadata { base: Some(base) }) = app.get_trigger_metadata("http")? {
             if base == "/" {
-                tracing::warn!("This application has the deprecated trigger 'base' set to the default value '/'. This may be an error in the future!");
+                tracing::warn!(
```
dicej, line 157: Sorry for the formatting churn here.  This might be because my editor runs `rustfmt` automatically using the latest edition, then I go back and run `cargo fmt` before committing, which formats using the project's edition, but apparently the round trip has side effects.

crates/trigger-http/src/wasip3.rs
```diff
@@ -0,0 +1,109 @@
...
+        let body = body.map_err(spin_factor_outbound_http::p2_to_p3_error_code);
+        let request = http::Request::from_parts(request, body);
+        let (request, request_io_result) = types::Request::from_http(request);
+        let request = wasi_http::<F>(store.data_mut())?.table.push(request)?;
+
+        let guest = self.0.load(&mut store, &instance)?;
+
+        let (tx, rx) = oneshot::channel();
+        task::spawn(
+            async move {
```
dicej, line 52: I'm guessing I need to sprinkle some `.in_current_span()` calls here, but not sure where.  Do we need them on every `async` block and every call to an async function?

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -580,3 +717,203 @@ fn dns_error(rcode: String, info_code: u16) -> ErrorCode {
         info_code: Some(info_code),
     })
 }
+
+pub fn p2_to_p3_error_code(code: p2_types::ErrorCode) -> p3_types::ErrorCode {
```
alexcrichton, line 750: This conversion back and forth is a bit unfortunate, so in the future one possibility might be to make the source of truth whatever hyper's native error is and then there's just a shim converting that to either p2 or p3, but no need for p2/p3 to meet each other here

crates/trigger-http/src/wasip3.rs
```diff
@@ -0,0 +1,109 @@
...
+    async fn execute<F: RuntimeFactors>(
+        &self,
+        instance_builder: TriggerInstanceBuilder<'_, F>,
+        route_match: &RouteMatch<'_, '_>,
+        mut req: http::Request<Body>,
+        client_addr: SocketAddr,
+    ) -> Result<http::Response<Body>> {
+        let spin_http::routes::TriggerLookupKey::Component(component_id) = route_match.lookup_key()
+        else {
+            anyhow::bail!("INCONCEIVABLE");
```
alexcrichton: debugging error message?

alexcrichton (approved):
Looks good to me on all the Wasmtime bits, thanks! I'll defer to others for the Spin-related bits though.

crates/trigger-http/src/wasip3.rs
dicej: I copied-and-pasted that from https://github.com/spinframework/spin/blob/5430ac1ca4e94caaf58b2c39f7573a9e3d7b197c/crates/trigger-http/src/wasi.rs#L34-L54, meaning to factor it out to be reused, but then forgot.  I'll do that.

Regarding the `bail` itself: yeah, it seemed a little weird to me, like it was meant to be an `unreachable!()` instead.  @itowlson this appears to have come from https://github.com/spinframework/spin/commit/058b1751ceba004d5addd45030d964e7a063a313; would it make sense to change it to `unreachable!()`, i.e. to indicate that if that code _is_ reached, it's due to a bug in Spin rather than anything the user did wrong?

dicej: I just pushed a commit to deduplicate the code and change the `bail` to an `unreachable`.  @itowlson let me know if you have concerns.

crates/factor-outbound-http/src/wasi.rs
dicej: Yeah, I even considered updating `wasmtime-wasi-http` to reuse the p2 `ErrorCode` (and its payloads) for the p3 bindings, since they're currently identical, but I know that Dan has proposed making some changes for p3, so they might not be identical for long.

dicej: I'll see how much refactoring would be needed to avoid the p2<->p3 conversion, and whether we can do it without changing any `wasmtime-wasi-http` APIs.

dicej: OK, I tried, but it got ugly, so I'm going to stick with the current approach for this PR.  I've opened https://github.com/spinframework/spin/issues/3274 to track further work and eliminate the need for the conversion functions.

crates/trigger-http/src/wasip3.rs
alexcrichton: oh oops sorry I tried to search for this already in the repo but I probably looked at an outdated branch or something

itowlson: @dicej oops, yes.  `unreachable!` would be correct; the unfortunate (but hopefully accurate) error message slipped through review.  Thanks for the fix.

crates/http/src/trigger.rs
```diff
@@ -63,6 +70,7 @@ impl HandlerType {
                     `{WASI_HTTP_EXPORT_2023_10_18}`, \
                     `{WASI_HTTP_EXPORT_2023_11_10}`, \
                     `{WASI_HTTP_EXPORT_0_2_PREFIX}.*`, \
+                    `{WASI_HTTP_EXPORT_0_3_PREFIX}.*`, \
```
tschneidereit: Should this also use the snapshot approach we used for the p2 snapshots? I don't think we want to commit to making the snapshots backwards compatible among each other, after all.

dicej: Yeah, good point.  At the time of writing, there was only one p3 snapshot in existence, but that's not true anymore as of yesterday.  I'll change this.

crates/trigger-http/src/wasip3.rs
lann: You typically only need it for the outermost `async` block within a `spawn`. If you peel back several layers of indirection it is essentially pulling the "current span" (handle) from a thread-local and then wrapping this `async`'s Future in an [`Instrumented`](https://docs.rs/tracing/0.1.41/tracing/instrument/struct.Instrumented.html) that sets that thread-local every time it is polled.

dicej:
~I'm not able to repro the CI error on Mac or Linux.  I tried rerunning the job, but it failed the second time.  Not sure what's up.~ UPDATE: fixed

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -44,16 +48,140 @@ use crate::{
...
+                Output = Result<
+                    (
+                        http::Response<BoxBody<Bytes, p3_types::ErrorCode>>,
+                        Box<dyn Future<Output = Result<(), p3_types::ErrorCode>> + Send>,
+                    ),
+                    TrappableError<p3_types::ErrorCode>,
+                >,
+            > + Send,
+    > {
+        // TODO: do we neeed to do anything with `fut`?
```
itowlson: Do we know the status of this TODO?

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -72,6 +199,17 @@ where
...
     )?;
 
+    fn get_http_p3<C>(store: &mut C::StoreData) -> p3::WasiHttpCtxView<'_>
+    where
+        C: spin_factors::InitContext<OutboundHttpFactor>,
+    {
+        let (state, table) = C::get_data_with_table(store);
+        p3::WasiHttpCtxView { ctx: state, table }
+    }
+    let get_http_p3 = get_http_p3::<C> as fn(&mut C::StoreData) -> p3::WasiHttpCtxView<'_>;
```
itowlson, line 218: super nit: Maybe a blank line between the inline function and the next chunk?

itowlson (approved):
Not sure my approval is worth much but here it is anyway!  Great to see this, excited to give it a try!

crates/factor-outbound-http/src/wasi.rs
dicej: Good question.  It's [also ignored in `wasmtime-wasi`](https://github.com/bytecodealliance/wasmtime/blob/7cebfa206fe4a40ab54e9862f30b05c5fefb9043/crates/wasi-http/src/p3/mod.rs#L109), ~and it seems redundant to me.  I assume it represents an error reading the request body, but that should be reported via the `BoxBody<Bytes, ErrorCode>` in the request; no need for a separate parameter.  I'll ask Roman (who wrote that code) about it.~ Nevermind; see my comment below.

dicej: Oops, I should have read the docs more carefully.  That future may be used to learn about errors the guest might encounter while doing whatever it decides to do with the response we return here.  Arguably we don't care, though -- we've done our job producing the response, and we can't do anything about it here if the guest has trouble e.g. forwarding it elsewhere.  I'll update the comment to say so.

itowlson: I don't quite understand, but probably don't need to, except to ask: would that include guest errors produced after the response has been finished?  (The reason I ask is e.g. the Slack bot scenario, where the bulk of the work may happen after the ack response has been returned: current Spin can be a bit weird about errors in that phase if I remember rightly.)

(I realise "a bit weird about" is, uh, a bit vague, but I have forgotten the details.)

itowlson: Never mind, I'm an idiot: this is outbound HTTP.  Sorry for the noise.

dicej: > would that include guest errors produced after the response has been finished?

The guest can send an error to `fut` at any time, i.e. before, during, or after the response has arrived in full and been delivered to the guest; there's nothing in the API that constraints when or why the guest would do that.  In any case, there's not much we can do about it here 🤷 .

crates/trigger-http/src/wasi.rs
```diff
@@ -17,6 +17,34 @@ use wasmtime_wasi_http::{bindings::Proxy, body::HyperIncomingBody as Body, WasiH
...
+    let headers = prepare_request_headers(req, route_match, client_addr)?;
+    req.headers_mut().clear();
+    req.headers_mut()
+        .extend(headers.into_iter().filter_map(|(n, v)| {
+            let Ok(name) = n.parse::<HeaderName>() else {
+                return None;
+            };
+            let Ok(value) = HeaderValue::from_bytes(v.as_bytes()) else {
+                return None;
+            };
```
lann, line 41: Just FYI

```suggestion
            let name = n.parse::<HeaderName>().ok()?;
            let value = HeaderValue::from_bytes(v.as_bytes()).ok()?;
```

PR #3273, opened 2025-09-16
Adds allow_unset field for variables

### LINKED ISSUE #3053: allow variables to be optional without a default value
currently variable needs to be either provided a default value or made `required` in `spin.toml`. I would like to keep my default value in code near to where it is actually used. one of the reason I want to store my default values in code is because my project can run as a independent service or as a spin app and I don't want to duplicate the default values in two places.

I wanted to hear if that is something we may consider to relax or are there other reasons that we recommend configuring the default value in `spin.toml`?

seun-ja:
Aims to resolve #3053

Offers an option for variables unset

### DIFF
--- a/crates/loader/src/local.rs
+++ b/crates/loader/src/local.rs
@@ -607,8 +607,8 @@ fn locked_metadata(
 
 fn locked_variable(variable: v2::Variable) -> Result<locked::Variable> {
     ensure!(
-        variable.required ^ variable.default.is_some(),
-        "must be `required` OR have a `default`"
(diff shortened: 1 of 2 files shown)


lann (changes requested):
We don't have a way to represent "default unset" in spin:variables/get. I think currently it would just return `error.undefined` which can't be distinguished from mistyping the name.

As-is I think this change isn't worth it when the alternative of `default = ""` will probably handle the vast majority of cases where this would be useful. If we really want to do this I think `spin:variables` should be revised to explicitly represent this case through e.g. a new error variant.

PR #3258, opened 2025-08-29
Use wasip2 target in templates and examples

fibonacci1729:

PR #3276, opened 2025-09-17
chore(examples): Update Spin Rust SDK and other dependencies

### LINKED ISSUE #3275: Understanding the differences between Spin and Wasmtime in wasip2 imports resolution
Hello everyone,

I am relatively new to Spin, so I am trying to understand how to use it. This issue is mainly a request for a technical explanation more than a problem with Spin.

I was studying the http-rust example (https://github.com/spinframework/spin/tree/main/examples/http-rust).
I noticed that while Spin can run the compiled module, Wasmtime cannot. When running the module with Wasmtime, I get:

```
Error: failed to run main module 'http_rust.wasm'

Caused by:
    0: failed to instantiate "http_rust.wasm"
    1: unknown import: 'wasi:io/streams@0.2.0-rc-2023-10-18::[resource-drop]output-stream' has not been defined
```

Since Spin is built upon Wasmtime, why is Spin able to provide this import while Wasmtime isn't? Wasmtime includes a wasi-io crate in its codebase (https://github.com/bytecodealliance/wasmtime/tree/3cc0023cb60295ee15931d40b266d723f4598eba/crates/wasi-io), so I would expect it to be able to run the example module as well.

Does anyone know why there is this difference in behavior between Spin and Wasmtime when it comes to resolving wasip2 imports?

Thanks in advance for your help :)

tschneidereit:
The Rust examples were all using Spin SDK 2.2.0, which targets a release candidate of WASIp2, not the final version. That means they only ran in Spin, not other WASI runtimes such as Wasmtime.

This commit updates all applicable examples to the latest Spin SDK and changes the build target to `wasm32-wasip2` to directly build components.

Fixes #3275


fibonacci1729 (approved):

fibonacci1729:
Closing #3258 in favor of this one. 

Also, would you mind updating gh workflows (`build.yml` and `release.yml`) to use the `wasm32-wasip2` target?

examples/spin-timer/app-example/src/lib.rs
```diff
@@ -1,9 +1,7 @@
 wit_bindgen::generate!({
     world: "spin-timer",
     path: "..",
-    exports: {
-        world: MySpinTimer
-    }
+    generate_all,
 });
```
fibonacci1729, line 6: This file is missing `export!(MySpinTimer)`

fibonacci1729 (changes requested):
Once the gh workflows are updated and the comment about missing `export!` is addressed the tests should pass and this looks good to me!

examples/spin-timer/app-example/src/lib.rs
tschneidereit: So it does!

thorstenhans:
This is great! 

With the samples compiling to `wasm32-wasip2`, I guess we would also want to update all the templates and `spin doctor`?

If so, I'm happy to help - especially with the templates. For `spin doctor` I see two ways of addressing this:

- we could flag manifests that have `wasm32-wasip1` as target for Rust components and check if the `wasm32-wasip2` target is installed
- we could report a suggestion and ask users to manually update the target in their build instructions and keep on checking for both Rust targets installed (maybe we're hitting cases where auto-upgrading from `p1` to `p2` could result in errors)

tschneidereit:
> I guess we would also want to update all the templates and `spin doctor`?

I think updating the templates makes sense, yes. Brian's #3258 (which I had missed—apologies!) did that, though I think there were a few more parts of the templates that need updating.

I'm less sure about `spin doctor`: there's not necessarily anything really wrong with targeting wasip1, and for now that's the only way to use Wizer. (That latter part will soon change, though.)

radu-matei (approved):

PR #3279, opened 2025-09-18
factor-outbound-http: Update override_connect_host

lann:
- Rename to override_connect_addr
- Change from String to SocketAddr
- Fix bug in host resolution for ipv6 addresses
- Use multi-threaded executor in tests in a desperate attempt to finally fix flaky tests

### DIFF
--- a/crates/factor-outbound-http/src/intercept.rs
+++ b/crates/factor-outbound-http/src/intercept.rs
@@ -1,3 +1,5 @@
+use std::net::SocketAddr;
+
 use http::{Request, Response};
 use http_body_util::{BodyExt, Full};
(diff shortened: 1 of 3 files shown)


itowlson (approved):

lann:
On further reflection I think the override should take a `SocketAddr` instead...

lann:
Sorry for the churn; it occurred to be that there was some ambiguity in what it means to override _with_ an `Authority` so I punted all that ambiguity to the poor interceptor.

itowlson (approved):

radu-matei (approved):

lann:
Last update just adds commit signature

PR #3280, opened 2025-09-30
Skip `load_instance_pre` for components not referenced by the trigger

itowlson:
This arose from prototyping WASI P3 middleware.  It was convenient to represent my middleware components as untriggered Spin components so they could participate in `spin build`, but then the app wouldn't load because the middleware components imported `next` (aka `origin`), which Spin couldn't satisfy.

I do worry that this solution might have unintended consequences, since it puts our instance map out of sync with our component list.  If we're concerned, I can close this and build my middleware using additional `build.command`s and avoid the Spin components.  But I think Till wants to be able to build library components within Spin anyway?  So we may need to figure it out at some point anyway.

Anyway very open to thoughts...


lann (approved):

tschneidereit:
> I think Till wants to be able to build library components within Spin anyway?

And indeed, I ran into the same issue in that pursuit just now. Thank you for fixing this!

PR #3282, opened 2025-10-01
variables: Refactor Provider::kind into ::may_resolve

lann:
I'd like to avoid the implicit calls to `get` required by the current implementation. I'm not attached to specifics here otherwise.

### DIFF
--- a/crates/expressions/src/lib.rs
+++ b/crates/expressions/src/lib.rs
@@ -11,8 +11,6 @@ pub use provider::Provider;
 use template::Part;
 pub use template::Template;
 
-use crate::provider::ProviderVariableKind;
-
 /// A [`ProviderResolver`] that can be shared.
 pub type SharedPreparedResolver =
     std::sync::Arc<std::sync::OnceLock<std::sync::Arc<PreparedResolver>>>;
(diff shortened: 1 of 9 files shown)


crates/expressions/src/provider.rs
```diff
@@ -9,15 +9,13 @@ use crate::Key;
...
-    /// Variable must be declared on start
-    Static,
-    /// Variable can be made available at runtime
-    #[default]
-    Dynamic,
+    /// Returns true if the given key _might_ be resolvable by this Provider.
+    ///
+    /// Dynamic resolvers will typically return true unconditionally, which is
+    /// the default implementation.
+    fn may_resolve(&self, key: &Key) -> bool {
```
lann, line 17: I'm trying to capture some subtlety in the name here: a Provider should only return `false` if it "knows" resolution would later fail; it is always safe to return `true`.

crates/expressions/src/lib.rs
```diff
@@ -92,50 +90,25 @@ impl ProviderResolver {
...
-                {
-                    resolved = true;
-                    break;
-                }
-            }
-
-            if !resolved {
-                unvalidated_keys.push(key);
+        let mut unresolvable_keys = vec![];
+        for key in self.internal.required_variables() {
```
itowlson, line 96: One thing I liked about the previous approach was that it allowed for a quick upfront check: do we have any dynamic providers in the mix, if so don't bother with checking individual variables.  But in practice I guess the performance overhead of always checking every variable is insignificant.

crates/expressions/src/lib.rs
```diff
@@ -92,50 +90,25 @@ impl ProviderResolver {
         Ok(PreparedResolver { variables })
     }
 
-    /// Ensures that all required variables are resolvable at startup
+    /// Ensures that all required variables are not unresolvable
     pub async fn ensure_required_variables_resolvable(&self) -> Result<()> {
```
itowlson: This might no longer need to be async?

lann: Indeed!

crates/expressions/src/lib.rs
lann: I wouldn't be surprised if we saved more cycles in a single CI run by these tests no longer needing to set up a tokio runtime than we ever lose to this deoptimization. :slightly_smiling_face:

itowlson: _chuckle_

PR #3284, opened 2025-10-02
Sort builds if dependency relation between manifest components

### LINKED ISSUE #3283: Build dependencies before triggered components
It's useful to be able to have dependencies built as part of `spin build`.  However, for type information to flow correctly (e.g. in the presence of a Rust build script or a JS `witDependencies`-like section), it's desirable to build those dependencies before we build the primary components.

(To be clear, this is infrastructural work to support other intended tooling: it won't necessarily make a difference today, but it goes some way to enabling stuff that will make a difference tomorrow.)

cc @tschneidereit

itowlson:
Fixes #3283.

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -7,6 +7,7 @@ mod manifest;
 use anyhow::{anyhow, bail, Context, Result};
 use manifest::ComponentBuildInfo;
 use spin_common::{paths::parent_dir, ui::quoted_path};
+use spin_manifest::schema::v2;
 use std::{
     collections::HashSet,
     path::{Path, PathBuf},
@@ -124,6 +125,15 @@ fn build_components(
         return Ok(());
     }
 
+    // If dependencies are being built as part of `spin build`, we would like
+    // them to be rebuilt earlier (e.g. so that consumers using the binary as a source
+    // of type information see the latest interface).
+    let (components_to_build, has_cycle) = sort(components_to_build);
+
+    if has_cycle {
+        tracing::debug!("There is a dependency cycle among components. Spin cannot guarantee to build dependencies before consumers.");
+    }
+
     components_to_build
         .into_iter()
         .map(|c| build_component(c, app_dir))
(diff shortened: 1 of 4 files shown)


crates/build/src/lib.rs
```diff
@@ -124,6 +125,12 @@ fn build_components(
         return Ok(());
     }
 
+    let (components_to_build, has_cycle) = sort(components_to_build);
```
rylev, line 131: We should comment why we're sorting. This code is pretty opaque when you don't have the context of #3283.

crates/build/src/lib.rs
```diff
@@ -124,6 +125,12 @@ fn build_components(
...
+
+    if has_cycle {
+        terminal::warn!("Cyclic dependency detected");
```
rylev: While this just be more confusing than actionable? What do we expect users to do with this information?

itowlson: @rylev Good question.  It's less "actionable" than it is "if you see errors where a consumer is not 'seeing' changes to the dependency interface then this could be why."  But maybe it's so pathological a case that it's not worth flagging...

rylev: Seems like something only someone with deep knowledge would be able to take action on which would indicate `debug` being a more appropriate logging level.

tschneidereit: I think this will become a thing we'll have to show more prominently before too long. Or rather, I think it should :)

Specifically, if and when we add features that require build outputs of dependencies as inputs to the dependee's build process.

It might make sense to look at the error message cargo reports for cyclic dependencies between crates as inspiration.

itowlson: @tschneidereit In the interests of getting this PR over the line, I propose we cross that bridge when we come to it...

tschneidereit: Oh yes, my apologies: I hadn't meant to say that this needs to happen now or otherwise nothing can possibly land.

rylev (approved):

PR #3285, opened 2025-10-02
Set a limit on max number of concurrent outbound http requests

### LINKED ISSUE #3281: Limit the number of concurrent outbound HTTP requests
I'm trying to limit the number of outbound requests guests can make to make sure no single guest can exhaust resources on the host system. Ideally, I'd be able to limit the number of client ports used by a guest application because I believe this (besides open file descriptors) is the resource I'm really worried about exhausting. However, I believe that this number is buried deep in the internals of hyper, and we don't really have access to it from inside Spin. 

So the next best idea is to just limit the number of outbound requests which should be a fairly close approximation of what we want. I have a [branch](https://github.com/spinframework/spin/pull/new/max-number-requests) with a POC implementation that just adds an `AtomicUsize` to the instance state and increment/decrements this number before and after each request. Additionally it checks this number against some configured max and rejects the request if we're currently doing too many. 

Does this sound like a good approach? If so, I will clean up the branch and open a PR. Otherwise, I'd love to brainstorm a better approach.

cc @dicej who recently added connection pooling and @lann who I know will be interested.

rylev:
Fixes #3281 

Wraps outbound requests in a semaphore with configurable permit size. This allows runtime config to decide how many outbound requests may be made at a give time.

### DIFF
--- a/crates/factor-outbound-http/src/spin.rs
+++ b/crates/factor-outbound-http/src/spin.rs
@@ -98,6 +98,8 @@ impl spin_http::Host for crate::InstanceState {
         });
 
         // If we're limiting concurrent outbound requests, acquire a permit
+        // Note: since we don't have access to the underlying connection, we can only
+        // limit the number of concurrent requests, not connections.
         let permit = match &self.concurrent_outbound_connections_semaphore {
             Some(s) => s.acquire().await.ok(),
             None => None,
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -457,18 +457,14 @@ impl RequestSender {
             None
         };
 
-        // If we're limiting concurrent outbound requests, acquire a permit
-        let permit = match self.concurrent_outbound_connections_semaphore {
-            Some(s) => s.acquire_owned().await.ok().map(Arc::new),
-            None => None,
-        };
         let resp = CONNECT_OPTIONS.scope(
             ConnectOptions {
                 blocked_networks: self.blocked_networks,
                 connect_timeout,
                 tls_client_config,
                 override_connect_addr,
-                permit,
+                concurrent_outbound_connections_semaphore: self
+                    .concurrent_outbound_connections_semaphore,
             },
             async move {
                 if use_tls {
@@ -556,10 +552,8 @@ struct ConnectOptions {
     tls_client_config: Option<TlsClientConfig>,
     /// If set, override the address to connect to instead of using the given `uri`'s authority.
     override_connect_addr: Option<SocketAddr>,
-    /// A permit for this connection
-    ///
-    /// If there is a permit, it should be dropped when the connection is closed.
-    permit: Option<Arc<OwnedSemaphorePermit>>,
+    /// A semaphore to limit the number of concurrent outbound connections.
+    concurrent_outbound_connections_semaphore: Option<Arc<Semaphore>>,
 }
 
 impl ConnectOptions {
@@ -603,6 +597,12 @@ impl ConnectOptions {
             return Err(ErrorCode::DestinationIpProhibited);
         }
 
+        // If we're limiting concurrent outbound requests, acquire a permit
+        let permit = match &self.concurrent_outbound_connections_semaphore {
+            Some(s) => s.clone().acquire_owned().await.ok(),
+            None => None,
+        };
+
         let stream = timeout(self.connect_timeout, TcpStream::connect(&*socket_addrs))
             .await
             .map_err(|_| ErrorCode::ConnectionTimeout)?
@@ -614,7 +614,7 @@ impl ConnectOptions {
             })?;
         Ok(PermittedTcpStream {
             inner: stream,
-            _permit: self.permit.clone(),
+            _permit: permit,
         })
     }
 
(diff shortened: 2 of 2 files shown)


lann:
It just occured to me how we could limit actual connections instead of requests:

```rust
struct PermittedTcpStream {
  inner: TcpStream,
  permit: OwnedSemaphorePermit,
}

// simple wrappers just like our existing `RustlsStream`
impl AsyncRead for PermittedTcpStream { ... }
impl AsyncWrite for PermittedTcpStream { ... }
```

The permit would be acquired out in `RequestSender::send_request` and passed in to `ConnectOptions` and used to construct the `PermittedTcpStream` in `HttpConnector`. The permit would automatically be dropped when the stream is dropped.

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -320,11 +326,17 @@ impl RequestSender {
...
+        let permit = match &self.concurrent_outbound_requests_semaphore {
+            Some(s) => s.acquire().await.ok(),
+            None => None,
+        };
         let resp = timeout(first_byte_timeout, resp)
             .await
             .map_err(|_| ErrorCode::ConnectionReadTimeout)?
             .map_err(hyper_legacy_request_error)?
             .map(|body| body.map_err(hyper_request_error).boxed());
+        drop(permit);
```
dicej: We should either do as Lann suggested and limit by connection rather than request, or else attach `permit` to the response body so that it is only dropped when the body is dropped (which could be long after we've returned from this function).

rylev: I've landed on Lann's suggestion as I think that aligns more closely with what we're trying to control here (overused TCP connections).

crates/factor-outbound-http/src/lib.rs
```diff
@@ -53,13 +54,15 @@ impl Factor for OutboundHttpFactor {
...
         Ok(AppState {
-            wasi_http_clients: wasi::HttpClients::new(connection_pooling),
-            connection_pooling,
+            wasi_http_clients: wasi::HttpClients::new(config.connection_pooling_enabled),
+            connection_pooling_enabled: config.connection_pooling_enabled,
+            concurrent_outbound_connections_semaphore: config
+                .max_concurrent_connections
+                // Permit count is the max concurrent connections + 1.
+                // i.e., 0 concurrent connections means 1 total connection.
+                .map(|n| Arc::new(Semaphore::new(n + 1))),
```
lann, line 65: Not a huge deal since we're talking about rather large limits but I don't understand this interpretation. I would expect "0 concurrent connections" to mean "you can't ever connect" (which should perhaps be invalid).

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -448,12 +457,18 @@ impl RequestSender {
             None
         };
 
+        // If we're limiting concurrent outbound requests, acquire a permit
+        let permit = match self.concurrent_outbound_connections_semaphore {
+            Some(s) => s.acquire_owned().await.ok().map(Arc::new),
+            None => None,
+        };
```
lann: I think this needs to be moved down into `ConnectOptions::connect_tcp` or it will still be limiting requests instead of connections.

crates/factor-outbound-http/src/lib.rs
rylev: I guess I have a different interpretation. A concurrent request is by definition one that is happening during another request. 0 concurrent requests would therefore mean "no requests happening at the same time" not "no requests at all". I think we should just pick which ever interpretation is most convenient and use that.

crates/factor-outbound-http/src/wasi.rs
rylev: Ah yea of course!

rylev: Should be good now: https://github.com/spinframework/spin/pull/3285/commits/a472018a12f49006befc833608acb8736fd19439

crates/factor-outbound-http/src/lib.rs
lann: How many requests are involved in "2 concurrent requests"?

calebschoepp: > A concurrent request is by definition one that is happening during another request

That is strictly true, but I think that doing the +1 makes everything else more confusing. The first request becomes concurrent once there is two total requests.

I'd vote for removing the `+ 1`

rylev:
@lann ping on a review.

crates/factor-outbound-http/src/spin.rs
```diff
@@ -92,13 +91,21 @@ impl spin_http::Host for crate::InstanceState {
...
-            if !self.connection_pooling {
+            if !self.connection_pooling_enabled {
                 builder = builder.pool_max_idle_per_host(0);
             }
             builder.build().unwrap()
         });
 
+        // If we're limiting concurrent outbound requests, acquire a permit
+        // Note: since we don't have access to the underlying connection, we can only
+        // limit the number of concurrent requests, not connections.
```
lann, line 102: Teeechnically we probably could via https://docs.rs/reqwest/0.12.24/reqwest/struct.ClientBuilder.html#method.connector_layer but given that this interface is effectively deprecated this seems fine.

lann:
> metrics

Tracking: https://github.com/spinframework/spin/issues/3307

PR #3287, opened 2025-10-03
MacOS 13 retirement

itowlson:
GitHub is retiring its MacOS 13 runner.  The process will start in about a month: at that point, GitHub will start doing "brownouts" during which they will randomly fail jobs that use it.  Planned brownout periods are:

* November 4, 14:00 UTC - November 5, 00:00 UTC
* November 11, 14:00 UTC - November 12, 00:00 UTC 
* November 18, 14:00 UTC - November 19, 00:00 UTC   
* November 25, 14:00 UTC - November 26, 00:00 UTC

The image will be fully retired "by December 4th."

The first two periods are in the run-up or around Kubecon, which is a likely target time for releases.  We have had previous experience of trying to push out a release while a brownout OS was in play and well let's not do that again eh.

**The impact of this change as currently implemented will be that we do not ship MacOS x64 builds going forward.**  If we want to keep doing Mac x64 builds, GitHub says we can use the `macos-15-intel` runner, which will see us out another couple of years.  My inclination is to use this opportunity to drop Mac x64 now, so that if anyone is relying on it we find out about _before_ the MacOS 15 escape hatch closes for good.  But other folks may find that too aggressive.  If we would rather change this PR to move to the MacOS 15 runner then let me know and I can do that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -115,7 +115,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [windows-latest, macos-13, macos-14]
+        os: [windows-latest, macos-14]
     steps:
       - uses: actions/checkout@v3
 
@@ -124,7 +124,7 @@ jobs:
         with:
           rust: true
           rust-wasm: true
-          rust-cache: "${{ matrix.os == 'windows-latest' }} || ${{ matrix.os == 'macos-13' }}"
+          rust-cache: "${{ matrix.os == 'windows-latest' }}"
           openssl-windows: "${{ matrix.os == 'windows-latest' }}"
 
       - name: Cargo Build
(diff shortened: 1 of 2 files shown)


fibonacci1729 (approved):
+1 to your strategy

itowlson:
@rajatjindal gave feedback that he still needed MacOS x64 builds, so I have changed this to instead bump to the final such image.  GitHub have indicated that this image will be removed by Q4 2027 so we should still be planning on deprecating this build (and so users should be planning for hardware migration or self-build before it becomes a crisis).

itowlson:
Ugh there's a mandatory/branch protection check for macos-13.

itowlson:
...and exterminated.  I'll re-establish a macos-15-intel rule once some of the active PRs have completed (to minmise busywork rebases)

PR #3295, opened 2025-10-08
Update WIT world to WASI P3 RC

itowlson:
This is my best guess at what our world looks like with the Wasmtime 37 update.  It's been a long time since we updated the core `platform` WASI reference so I'm not super confident but I _think_ `wasi:cli/imports` and `wasi:http/handler` are all we need.  I'm not quite sure if we need a 3.5 world that exports HTTP P2: apps targeting `spin:up@3.4` will be fine but I'm not sure if an app using the P2 handler signature could target `spin:up@3.5` (and work by inclusion of the 0.2.0 deps), or whether it would need to (since presumably such an app would be using P2 imports as well???).  So @fibonacci1729 @dicej if you could take a critical look that would be super helpful... thanks!

### DIFF
--- a/crates/world/src/lib.rs
+++ b/crates/world/src/lib.rs
@@ -12,6 +12,7 @@ wasmtime::component::bindgen!({
         include fermyon:spin/platform@3.0.0;
         include spin:up/platform@3.2.0;
         include spin:up/platform@3.4.0;
+        include spin:up/platform@3.5.0;
         include wasi:keyvalue/imports@0.2.0-draft2;
     }
     "#,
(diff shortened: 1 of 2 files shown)


fibonacci1729 (approved):

dicej:
@itowlson regarding your questions and concerns about which interfaces to include in which worlds, etc.: I think what you've done here is sufficient; if we find we need more worlds and/or a different selection of interfaces in a given world, we can add those.  Meanwhile, an SDK can always define their own world(s) corresponding to a unique subset of the interfaces Spin 3.5.x supports, according to what that SDK provides bindings for; i.e. they need not be restricted to one of the worlds defined here.

PR #3300, opened 2025-10-13
Workflow for publishing spin:up WIT package

fibonacci1729:
Simple manual dispatch workflow for publishing the `spin:up` WIT package. 

~Draft for now until I amend the release notes.~

Will follow up with separate PR amending the release notes once the first release is done.

### DIFF
--- a/scripts/build-and-publish-wit.sh
+++ b/scripts/build-and-publish-wit.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+# This script builds and publishes the spin:up WIT package
+# NOTE: The package name and version are inferred from the
+# encoded wasm binary.
+set -euo pipefail 
+
+# Build the package
+wasm-tools component wit wit/ -w -o spin_up_wit.wasm
+
+# Publish to registry
+wkg publish --registry spinframework.dev spin_up_wit.wasm
--- a/.github/workflows/publish-wit.yml
+++ b/.github/workflows/publish-wit.yml
@@ -1,0 +1,39 @@
+name: Publish WIT to registry
+
+on:
+  workflow_dispatch:
+
+jobs:
+  publish-wit:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      packages: write
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install cargo-binstall
+        uses: cargo-bins/cargo-binstall
+
+      - name: Install wasm-tools
+        shell: bash
+        run: cargo binstall wasm-tools
+
+      - name: Install wkg
+        shell: bash
+        run: cargo binstall wkg
+
+      - name: Login to the GitHub registry
+        uses: docker/login-action
+        with:
+          registry: ghcr.io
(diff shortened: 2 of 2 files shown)


scripts/build-and-publish-wit.sh
```diff
@@ -0,0 +1,15 @@
...
+set -euo pipefail 
+
+# Read only the first line (the package declaration)
+line=$(head -n 1 "wit/world.wit")
+
+# Parse the version from the WIT file
+version=$(echo "$line" | sed -E 's/.*@([0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?(\+[0-9A-Za-z.-]+)?).*/\1/')
+
+# Build the package
+wasm-tools component wit wit -w -o spin:up@$version.wasm
```
itowlson: Nit: Is the distinctive name needed?  `wkg publish` says"the package name and version will be inferred from the Wasm file" and I assumed that was the file content rather than the name.  But since I can't get the damn thing to work locally I can't confirm.

scripts/build-and-publish-wit.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Publish to GHCR
+wkg publish --registry ghcr.io spin:up@$version.wasm
```
itowlson: I think this might need to be `--registry spinframework.dev`?  I thought wkg ran against wasm-pkg registry indirections rather than against the OCI layer directly.  I might be wrong though.

itowlson:
Oh the other thing I couldn't remember from my previous testing (and now can't verify because nothing ever works) was if wkg packages arrived public or private.  GHCR tends to default to private and I can't remember if wkg fixes that.

scripts/build-and-publish-wit.sh
fibonacci1729: Good catch, looking at the wkg code it appears to extrapolate the package spec from the wasm so I can remove this nonsense! Ill verify tomorrow when at computer and not big phone.

fibonacci1729: Updated

fibonacci1729:
@itowlson good question; i'm not sure. I think it's something we can investigate once this lands and we try to publish the first package. cc/ @lann if you have thoughts on this.

lann:
`wkg` just speaks generic OCI; it doesn't know anything about GitHub. If ghcr.io packages default to private then that's what will happen.

_Edit: but I believe you only have to change the visibility for a package once._

scripts/build-and-publish-wit.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# This script builds and publishes the spin:up WIT package
+# NOTE: The package name and version are inferred from the
```
itowlson, line 3: Not any more _grin_

itowlson (approved):

scripts/build-and-publish-wit.sh
fibonacci1729: AHHHHHHH

fibonacci1729: Wait, no. This is still accurate.

Issue #3293, opened 2025-10-08
Write and publish environment TOMLs for `spin up`


itowlson:
In order to start making use of the target environments feature, we need environment TOMLs (that is, trigger type-to-compatible worlds mappings) to be in an OCI registry.  The default location is `ghcr.io/spinframework/environments`.

We should decide/define/create:

* Where to maintain the TOMLs that are part of CNCF definitions (bearing in mind there will also be SpinKube: e.g. `spin up` in the Spin/SpinKube in the SK repo, or both in a separate repo, or...)
* How to handle third-party TOMLs (PRed into the Spin org and pushed to the default registry, or separate and pushed to owner's registry, or some intermediate which I dunno what it would look like)
* Processes for updating the TOMLs with new Spin releases (particularly around timing once we start offering environments in templates)
* Automation around pushing new TOMLs to OCI

PR #3301, opened 2025-10-14
Environment definitions starter pack

itowlson:
This provides TOML environment definitions for Spin 3.2 and 3.4.  (It will also need a definition for 3.5, once #3295 is agreed.)

Proposed plan of attack: 

* Once #3300 merges, we can run that against the 3.4 and 3.2 branches to get those WITs into the default `spinframework.dev` registry.
* Then we can test these TOMLs using local files against all of our templates and any known good applications. This is important because during dev I ran into problems with apps that targeted older worlds and we want to make sure that these (within reason) do not give spurious errors.  My dev test suggested that we might need to retrofit some compat stuff to the top level worlds which I really hope we don't have to do but well this is where we find out I guess.
* Then we can merge them.  _mumble mumble automation mumble CI mumble muttered cursing_
* Then we can push them to the default OCI registry.
* Then we can test them again using just env names instead of file references.  Which will work first time and will produce no mysterious inconsistencies, none do you hear me, but you know, good form.
* And then we can tell people about them!

### DIFF
--- a/.github/workflows/publish-env-def.yml
+++ b/.github/workflows/publish-env-def.yml
@@ -1,0 +1,37 @@
+name: Publish environment definition to registry
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "spin-up.(ver).toml version, e.g. 3.4, 3.5"
+        required: true
+        type: string
+
+jobs:
+  publish-env-def:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      packages: write
+      contents: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install oras
+        uses: oras-project/setup-oras@v1
+
+      - name: Login to the GitHub registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: spinframeworkbot
+          password: ${{ secrets.SPIN_WIT_PAT }}
+
+      - name: Publish environment TOML
+        shell: bash
+        env:
+          VERSION: ${{ inputs.version }}
+        run: ./scripts/push-env-def.sh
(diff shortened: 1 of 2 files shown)


itowlson:
#3293 had an open question about where we want to keep these but it's not had much traction so I'm parking them here for now but if people have other preferences then whatever

itowlson:
@fibonacci1729 with the 3.2-2.5 packages uploaded I think this is ready to go but I am not sure if the script/workflow is right or how to test it.  The script works (mutatis mutandis) locally though...

.github/workflows/publish-env-def.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          registry: ghcr.io
+          username: spinframeworkbot
+          password: ${{ secrets.SPIN_WIT_PAT }}
+
+      - name: Publish environment TOML
+        shell: bash
+        env:
+          VERSION: ${{ inputs.version }}
+        run: ./scripts/build-and-publish-wit.sh
```
fibonacci1729: Should this be `./script/push-env-def.sh`?

.github/workflows/publish-env-def.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "spin-up.(ver).toml version, e.g. 3.4, 3.5"
+        required: true
+        type: string
+
+jobs:
+  publish-wit:
```
fibonacci1729:
```suggestion
  publish-env-def:
```

.github/workflows/publish-env-def.yml
itowlson: Bloody hell.  Come on, Ivan's brain.  Thank you

fibonacci1729 (approved):
LGTM!

PR #3305, opened 2025-10-15
Fix the publish-wit ci workflow

fibonacci1729:
I think this ought to do the trick.


itowlson (approved):
> I think this ought to do the trick.

Yes, it looks like it _ought_ to, but on current trends it will make YAML interpret it as choreography notation for an interpretive dance. :shipit:

PR #3313, opened 2025-10-20
Outbound http permits metrics

rylev:
Partially fixes https://github.com/spinframework/spin/issues/3307

I didn't implement a gauge of `available_permits` for two reasons:
* We don't yet support gauges in `spin_telemetry`
* The semaphore has no API for getting the number of total permits which means we have to thread that book keeping through which I didn't feel like doing.

Curious what folks think about this.

### DIFF
--- a/crates/telemetry/src/metrics.rs
+++ b/crates/telemetry/src/metrics.rs
@@ -61,6 +61,8 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
 ///
 /// The increment may only be an i64 or f64. You must not mix types for the same metric.
 ///
+/// Takes advantage of counter support in [tracing-opentelemetry](https://docs.rs/tracing-opentelemetry/0.32.0/tracing_opentelemetry/struct.MetricsLayer.html).
+///
 /// ```no_run
 /// # use spin_telemetry::metrics::counter;
 /// counter!(spin.metric_name = 1, metric_attribute = "value");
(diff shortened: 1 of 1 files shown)


crates/factor-outbound-http/src/lib.rs
```diff
@@ -142,6 +142,66 @@ impl InstanceState {
...
+            // No available permits right now; wait for one
+            Err(tokio::sync::TryAcquireError::NoPermits) => {
+                waited = true;
+                acquire().await.map_err(|_| ())
+            }
+            Err(_) => Err(()),
+        };
+        if permit.is_ok() {
+            spin_telemetry::monotonic_counter!(
+                outbound_http.acquired_permits = 1,
```
lann: 🚲🏚️ a little more verbosity might help if you don't already know what you're looking at; `concurrent_connection_permits_acquired`?

lann:
I'll take it over nothing! It would be good to have `available_permits` as a _leading_ indicator of (potential) problems but the `waited` count should be useful on its own as a _trailing_ indicator.

rylev:
@lann @calebschoepp the only thing I'm unsure about is whether `monotonic_counter!` is the right way to emit metrics like these. It seems like the implementation is simply using `tracing::trace!` which I don't think would do what we want.

calebschoepp:
> @lann @calebschoepp the only thing I'm unsure about is whether `monotonic_counter!` is the right way to emit metrics like these. It seems like the implementation is simply using `tracing::trace!` which I don't think would do what we want.

It should work. It is using [`tracing_opentelemetry::MetricsLayer`](https://docs.rs/tracing-opentelemetry/0.32.0/tracing_opentelemetry/struct.MetricsLayer.html). Maybe a comment could be thrown in `spin_telemetry` to make that more obvious? Also looking at the docs I linked it seems to support gauges so you could unblock yourself on that front

rylev:
@calebschoepp can I get another review? I didn't end up using the gauge macro in this PR because I don't have time to thread that through, but I think we don't need to delay merging this any more.

calebschoepp (approved):

rylev:
@lann ping - looks like I need you to give this a 👍

fibonacci1729:
@rylev feel free to merge !

PR #3320, opened 2025-10-23
bump deps with `cargo audit fix`

dicej:
This addresses all the vulnerabilities flagged by `cargo audit` which are fixable without API-breaking updates.  We'll need to upgrade to newer, API-breaking versions to address the remaining vulnerabilities.


radu-matei:
@dicej -- could you please create an issue so we track the ones not addressed here?
Thanks!

dicej:
> @dicej -- could you please create an issue so we track the ones not addressed here? Thanks!

https://github.com/spinframework/spin/issues/3321

PR #3326, opened 2025-10-26
Expose per-component CPU and memory usage

### LINKED ISSUE #2933: Expose CPU and memory usage on per-component execution spans
It would be nice if we could see the amount of memory and CPU that was used when a component is executed.

I believe exposing CPU time requires a CallHook which means we'll probably want to put it behind a feature flag or benchmark it to make sure it doesn't regress performance.

asteurer:
This is not currently working. Just starting the conversation. Closes #2933

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8283,6 +8283,7 @@ dependencies = [
  "spin-doctor",
  "spin-environments",
  "spin-factor-outbound-networking",
+ "spin-factors-executor",
  "spin-http",
  "spin-loader",
  "spin-locked-app",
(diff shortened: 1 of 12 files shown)


asteurer:
@lann When you get a chance, I could use some help figuring out what is preventing CPU and Memory usage from being recorded and printed. Also, I would love your thoughts on this initial approach.

asteurer:
Also, I relied heavily on code that Till wrote: https://github.com/spinframework/spin/compare/main...tschneidereit:spin:resource-usage

asteurer:
I'm going to spend some more time iterating on this privately

PR #3331, opened 2025-10-28
Fix the template bump script to accept leading 'v'

fibonacci1729:
The rust SDK dispatches an event with version payload that includes a leading `v`.

### DIFF
--- a/scripts/bump-rust-template-sdk.sh
+++ b/scripts/bump-rust-template-sdk.sh
@@ -14,7 +14,7 @@ trap 'find templates -name "*.bak" -delete' EXIT
 usage() {
   echo "Usage: $0 <VERSION>"
   echo "Updates the Rust templates SDK dependency to the specified version"
-  echo "Example: $0 6.0.0"
+  echo "Example: $0 v6.0.0"
 }
 
 if [[ $# -ne 1 ]]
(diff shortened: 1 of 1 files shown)


scripts/bump-rust-template-sdk.sh
```diff
@@ -24,12 +24,14 @@ then
...
-if [[ ! "${VERSION}" =~ ^[0-9]+.[0-9]+.[0-9]+$ ]]
+if [[ ! "${VERSION}" =~ ^v[0-9]+.[0-9]+.[0-9]+$ ]]
 then
-  echo "VERSION doesn't match [0-9]+.[0-9]+.[0-9]+ and may be a prerelease; skipping."
+  echo "VERSION doesn't match v[0-9]+.[0-9]+.[0-9]+ and may be a prerelease; skipping."
   exit 1
 fi
 
+# Strip the leading v for Cargo.toml
+CARGO_VERSION="${VERSION#v}"
```
itowlson: nit: this name read to me like "the version of Cargo" rather than "the version that works in `Cargo.toml`"

fibonacci1729: Changed to `STRIPPED_VERSION`.

PR #3340, opened 2025-10-28
Upgrade watchexec to 8.x and other audit adventures

itowlson:
In the hope that `cargo audit` will be less mad at us about it

### DIFF
--- a/crates/oci/src/utils.rs
+++ b/crates/oci/src/utils.rs
@@ -1,61 +1,59 @@
 //! Utilities related to distributing Spin apps via OCI registries
 
 use anyhow::{Context, Result};
-use async_compression::tokio::bufread::GzipDecoder;
-use async_compression::tokio::write::GzipEncoder;
-use async_tar::Archive;
+use flate2::read::GzDecoder;
+use flate2::write::GzEncoder;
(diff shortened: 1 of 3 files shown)


itowlson:
Added fixes for `url` and `async-tar`

crates/oci/src/utils.rs
```diff
@@ -1,61 +1,61 @@
 //! Utilities related to distributing Spin apps via OCI registries
 
 use anyhow::{Context, Result};
-use async_compression::tokio::bufread::GzipDecoder;
-use async_compression::tokio::write::GzipEncoder;
-use async_tar::Archive;
+// use async_compression::tokio::bufread::GzipDecoder;
+// use async_compression::tokio::write::GzipEncoder;
```
calebschoepp: Is keeping this around intentional?

calebschoepp (approved):

crates/oci/src/utils.rs
itowlson: Oops - thanks!

calebschoepp:
@dicej my ✅ isn't worth much round these parts. LGTM. Mind giving Ivan an approval that will let him merge.

PR #3346, opened 2025-11-04
WASI OTel

asteurer:
# Overview
This is an implementation of [WASI OTel](https://github.com/calebschoepp/wasi-otel) which gives guest applications the ability to export traces, metrics and logs with [OpenTelemetry SDKs](https://github.com/calebschoepp/opentelemetry-wasi).

We see two options for adding this to Spin:
- Require the user to compile Spin with configuration flags that will enable this feature.
- Have this feature available out-of-the-box and add an experimental-features flag that the user will pass to spin up.

To help ease adoption, we would prefer to pass a flag to spin up, but we know there was some previous work around this with wasip3 so we're happy to defer to that process.

Some additional notes:
- In the root `Cargo.toml`, you will notice that certain experimental features are being enabled for the `opentelemetry-sdk` dependency, and that we are using `reqwest-client` instead of `reqwest-blocking-client` in the `opentelemetry` dependency. These are all to prevent runtime errors that come from the default OpenTelemetry dependency configurations conflicting with Spin's async runtime.
- `wasi-otel` has been moved to Phase 1.

# Usage
To try the features added in this PR, do the following:

1. Build Spin from [this branch](https://github.com/asteurer/spin/tree/wasi-otel):
```bash
git clone --branch wasi-otel --depth 1 https://github.com/asteurer/spin
cd spin
cargo install --path .
spin plugin update
spin plugin install otel
```

2. Clone [OpenTelemetry WASI](https://github.com/calebschoepp/opentelemetry-wasi) and try the Rust example applications:
```bash
git clone https://github.com/calebschoepp/opentelemetry-wasi
spin otel setup

cd opentelemetry-wasi/rust/examples/spin-basic
spin otel up -- --build --experimental-wasi-otel

# In a different terminal window
spin otel open jaeger
spin otel open grafana
curl localhost:3000
```

### DIFF
--- a/crates/factor-key-value/src/host.rs
+++ b/crates/factor-key-value/src/host.rs
@@ -1,6 +1,7 @@
 use super::{Cas, SwapError};
 use anyhow::{Context, Result};
 use spin_core::{async_trait, wasmtime::component::Resource};
+use spin_factor_otel::OtelFactorState;
 use spin_resource_table::Table;
 use spin_telemetry::traces::{self, Blame};
 use spin_world::v2::key_value;
@@ -49,23 +50,31 @@ pub struct KeyValueDispatch {
     manager: Arc<dyn StoreManager>,
     stores: Table<Arc<dyn Store>>,
     compare_and_swaps: Table<Arc<dyn Cas>>,
+    otel: OtelFactorState,
 }
 
 impl KeyValueDispatch {
     pub fn new(allowed_stores: HashSet<String>, manager: Arc<dyn StoreManager>) -> Self {
-        Self::new_with_capacity(allowed_stores, manager, DEFAULT_STORE_TABLE_CAPACITY)
+        Self::new_with_capacity(
+            allowed_stores,
+            manager,
+            DEFAULT_STORE_TABLE_CAPACITY,
+            Default::default(),
+        )
     }
 
     pub fn new_with_capacity(
         allowed_stores: HashSet<String>,
         manager: Arc<dyn StoreManager>,
         capacity: u32,
+        otel: OtelFactorState,
     ) -> Self {
         Self {
             allowed_stores,
             manager,
             stores: Table::new(capacity),
             compare_and_swaps: Table::new(capacity),
+            otel,
         }
     }
 
@@ -113,6 +122,7 @@ impl key_value::Host for KeyValueDispatch {}
 impl key_value::HostStore for KeyValueDispatch {
     #[instrument(name = "spin_key_value.open", skip(self), err, fields(otel.kind = "client", kv.backend=self.manager.summary(&name).unwrap_or("unknown".to_string())))]
     async fn open(&mut self, name: String) -> Result<Result<Resource<key_value::Store>, Error>> {
+        self.otel.reparent_tracing_span();
         Ok(async {
             if self.allowed_stores.contains(&name) {
                 let store = self.manager.get(&name).await?;
@@ -135,6 +145,7 @@ impl key_value::HostStore for KeyValueDispatch {
         store: Resource<key_value::Store>,
         key: String,
     ) -> Result<Result<Option<Vec<u8>>, Error>> {
+        self.otel.reparent_tracing_span();
         let store = self.get_store(store)?;
         Ok(store.get(&key).await.map_err(track_error_on_span))
     }
@@ -146,6 +157,7 @@ impl key_value::HostStore for KeyValueDispatch {
         key: String,
         value: Vec<u8>,
     ) -> Result<Result<(), Error>> {
+        self.otel.reparent_tracing_span();
         let store = self.get_store(store)?;
         Ok(store.set(&key, &value).await.map_err(track_error_on_span))
     }
@@ -156,6 +168,7 @@ impl key_value::HostStore for KeyValueDispatch {
         store: Resource<key_value::Store>,
         key: String,
     ) -> Result<Result<(), Error>> {
+        self.otel.reparent_tracing_span();
         let store = self.get_store(store)?;
         Ok(store.delete(&key).await.map_err(track_error_on_span))
     }
@@ -166,6 +179,7 @@ impl key_value::HostStore for KeyValueDispatch {
         store: Resource<key_value::Store>,
         key: String,
     ) -> Result<Result<bool, Error>> {
+        self.otel.reparent_tracing_span();
         let store = self.get_store(store)?;
         Ok(store.exists(&key).await.map_err(track_error_on_span))
     }
@@ -175,6 +189,7 @@ impl key_value::HostStore for KeyValueDispatch {
         &mut self,
         store: Resource<key_value::Store>,
     ) -> Result<Result<Vec<String>, Error>> {
+        self.otel.reparent_tracing_span();
         let store = self.get_store(store)?;
         Ok(store.get_keys().await.map_err(track_error_on_span))
     }
--- a/crates/factor-otel/src/host.rs
+++ b/crates/factor-otel/src/host.rs
@@ -1,0 +1,120 @@
+use crate::InstanceState;
+use anyhow::anyhow;
+use anyhow::Result;
+use opentelemetry::trace::TraceContextExt;
+use opentelemetry_sdk::error::OTelSdkError;
+use opentelemetry_sdk::logs::LogProcessor;
+use opentelemetry_sdk::metrics::exporter::PushMetricExporter;
+use opentelemetry_sdk::trace::SpanProcessor;
+use spin_world::wasi;
+use tracing_opentelemetry::OpenTelemetrySpanExt;
+
+impl wasi::otel::tracing::Host for InstanceState {
+    async fn on_start(&mut self, context: wasi::otel::tracing::SpanContext) -> Result<()> {
+        // If the host does not have tracing enabled we just no-op
+        let Some(tracing_state) = self.tracing_state.as_ref() else {
+            return Ok(());
+        };
+        let mut tracing_state = tracing_state.write().unwrap();
+
+        // Before we do anything make sure we track the original host span ID for reparenting
+        if tracing_state.original_host_span_id.is_none() {
+            tracing_state.original_host_span_id = Some(
+                tracing::Span::current()
+                    .context()
+                    .span()
+                    .span_context()
+                    .span_id(),
+            );
+        }
+
+        // Track the guest spans context in our ordered map
+        let span_context: opentelemetry::trace::SpanContext = context.into();
+        tracing_state
+            .guest_span_contexts
+            .insert(span_context.span_id(), span_context);
+
+        Ok(())
+    }
+
+    async fn on_end(&mut self, span_data: wasi::otel::tracing::SpanData) -> Result<()> {
+        // If the host does not have tracing enabled we just no-op
+        let Some(tracing_state) = self.tracing_state.as_ref() else {
+            return Ok(());
+        };
+        let mut tracing_state = tracing_state.write().unwrap();
+
+        let span_context: opentelemetry::trace::SpanContext = span_data.span_context.clone().into();
+        let span_id: opentelemetry::trace::SpanId = span_context.span_id();
+
+        if tracing_state
+            .guest_span_contexts
+            .shift_remove(&span_id)
+            .is_none()
+        {
+            Err(anyhow!("Trying to end a span that was not started"))?;
+        }
+
+        tracing_state.span_processor.on_end(span_data.into());
+
+        Ok(())
+    }
+
+    async fn outer_span_context(&mut self) -> Result<wasi::otel::tracing::SpanContext> {
+        Ok(tracing::Span::current()
+            .context()
+            .span()
+            .span_context()
+            .clone()
+            .into())
+    }
+}
+
+impl wasi::otel::metrics::Host for InstanceState {
+    async fn export(
+        &mut self,
+        metrics: wasi::otel::metrics::ResourceMetrics,
+    ) -> spin_core::wasmtime::Result<std::result::Result<(), wasi::otel::metrics::Error>> {
+        // If the host does not have metrics enabled we just no-op
+        let Some(metric_exporter) = self.metric_exporter.as_ref() else {
+            return Ok(Ok(()));
+        };
+
+        match metric_exporter.export(&mut metrics.into()).await {
+            Ok(_) => Ok(Ok(())),
+            Err(e) => match e {
+                OTelSdkError::AlreadyShutdown => {
+                    let msg = "Shutdown has already been invoked";
+                    tracing::error!(msg);
+                    Ok(Err(msg.to_string()))
+                }
+                OTelSdkError::InternalFailure(e) => {
+                    let detailed_msg = format!("Internal failure: {}", e);
+                    tracing::error!(detailed_msg);
+                    Ok(Err("Internal failure.".to_string()))
+                }
+                OTelSdkError::Timeout(d) => {
+                    let detailed_msg = format!("Operation timed out after {} seconds", d.as_secs());
+                    tracing::error!(detailed_msg);
+                    Ok(Err("Operation timed out.".to_string()))
+                }
+            },
+        }
+    }
+}
+
+impl wasi::otel::logs::Host for InstanceState {
+    async fn on_emit(
+        &mut self,
+        data: wasi::otel::logs::LogRecord,
+    ) -> spin_core::wasmtime::Result<()> {
+        // If the host does not have logs enabled we just no-op
+        let Some(log_processor) = self.log_processor.as_ref() else {
+            return Ok(());
+        };
+
+        let (mut record, scope) = spin_world::wasi_otel::parse_wasi_log_record(data);
+        log_processor.emit(&mut record, &scope);
+        Ok(())
+    }
+}
--- a/crates/factor-otel/src/lib.rs
+++ b/crates/factor-otel/src/lib.rs
@@ -1,0 +1,286 @@
+mod host;
+
+use anyhow::bail;
+use indexmap::IndexMap;
+use opentelemetry::{
+    trace::{SpanContext, SpanId, TraceContextExt},
+    Context,
+};
+use opentelemetry_otlp::MetricExporter;
+use opentelemetry_sdk::{
+    logs::{log_processor_with_async_runtime::BatchLogProcessor, LogProcessor},
+    resource::{EnvResourceDetector, ResourceDetector, TelemetryResourceDetector},
+    runtime::Tokio,
+    trace::{span_processor_with_async_runtime::BatchSpanProcessor, SpanProcessor},
+    Resource,
+};
+use spin_factors::{Factor, FactorData, PrepareContext, RuntimeFactors, SelfInstanceBuilder};
+use spin_telemetry::{
+    detector::SpinResourceDetector,
+    env::{otel_logs_enabled, otel_metrics_enabled, otel_tracing_enabled, OtlpProtocol},
+};
+use std::sync::{Arc, RwLock};
+use tracing_opentelemetry::OpenTelemetrySpanExt;
+
+pub struct OtelFactor {
+    span_processor: Option<Arc<BatchSpanProcessor<Tokio>>>,
+    metric_exporter: Option<Arc<MetricExporter>>,
+    log_processor: Option<Arc<BatchLogProcessor<Tokio>>>,
+    enable_interface: bool,
+}
+
+impl Factor for OtelFactor {
+    type RuntimeConfig = ();
+    type AppState = ();
+    type InstanceBuilder = InstanceState;
+
+    fn init(&mut self, ctx: &mut impl spin_factors::InitContext<Self>) -> anyhow::Result<()> {
+        // Only link the WASI OTel bindings if experimental support is enabled. This means that if
+        // the user tries to run a guest component that consumes the WASI OTel WIT without enabling
+        // experimental support they'll see an error like "component imports instance
+        // `wasi:otel/tracing@0.2.0-draft`"
+        if self.enable_interface {
+            ctx.link_bindings(
+                spin_world::wasi::otel::tracing::add_to_linker::<_, FactorData<Self>>,
+            )?;
+            ctx.link_bindings(
+                spin_world::wasi::otel::metrics::add_to_linker::<_, FactorData<Self>>,
+            )?;
+            ctx.link_bindings(spin_world::wasi::otel::logs::add_to_linker::<_, FactorData<Self>>)?;
+        }
+        Ok(())
+    }
+
+    fn configure_app<T: spin_factors::RuntimeFactors>(
+        &self,
+        _ctx: spin_factors::ConfigureAppContext<T, Self>,
+    ) -> anyhow::Result<Self::AppState> {
+        Ok(())
+    }
+
+    fn prepare<T: spin_factors::RuntimeFactors>(
+        &self,
+        _: spin_factors::PrepareContext<T, Self>,
+    ) -> anyhow::Result<Self::InstanceBuilder> {
+        if !self.enable_interface {
+            return Ok(InstanceState::default());
+        }
+
+        // Warn the user if they enabled experimental support but didn't supply any environment variables
+        if self.span_processor.is_none()
+            && self.metric_exporter.is_none()
+            && self.log_processor.is_none()
+        {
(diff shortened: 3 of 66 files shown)


calebschoepp:
@lann I think you would likely be the most relevant reviewer for this

crates/factor-otel/src/lib.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub(crate) original_host_span_id: Option<SpanId>,
+}
+
+/// Manages access to the OtelFactor state for the purpose of maintaining proper span
+/// parent/child relationships when WASI Otel spans are being created.
+pub struct OtelContext {
+    pub(crate) state: Option<Arc<RwLock<State>>>,
+}
+
+impl OtelContext {
```
lann: There is already https://docs.rs/opentelemetry/latest/opentelemetry/context/struct.Context.html; maybe `OtelState` or `OtelFactorState` or `SpinOtel` or something?

crates/telemetry/src/env.rs
```diff
@@ -19,7 +19,7 @@ const SPIN_DISABLE_LOG_TO_TRACING: &str = "SPIN_DISABLE_LOG_TO_TRACING";
 /// - `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
 ///
 /// Note that this is overridden if OTEL_SDK_DISABLED is set and not empty.
-pub(crate) fn otel_tracing_enabled() -> bool {
+pub fn otel_tracing_enabled() -> bool {
```
lann, line 22: It isn't immediately obvious why these visibility changes are necessary?

crates/world/src/conversions.rs
```diff
@@ -553,3 +553,495 @@ mod llm {
...
+            $exemplar_type:ty
+        ) => {
+            $wasi_exemplar_list
+                .iter()
+                .map(|e| {
+                    let span_id: [u8; 8] = e
+                        .span_id
+                        .as_bytes()
+                        .try_into()
+                        .expect("Span ID is longer than 8 bytes");
```
lann: Is this `expect` (and the one below) actually impossible?

crates/trigger-http/src/wasi.rs
```diff
@@ -107,6 +107,7 @@ impl HttpExecutor for WasiHttpExecutor<'_> {
             HandlerType::Wagi(_) => unreachable!("should have used WagiExecutor instead"),
         };
 
+        let span = tracing::debug_span!("execute_wasi");
```
lann: Is this intentional here?

crates/world/src/conversions.rs
```diff
@@ -553,3 +553,495 @@ mod llm {
         }
     }
 }
+
+mod otel {
```
lann: My immediate hot take is that we should break this out into a separate wasi-otel crate. We've been discussing removing this `spin-world` crate entirely and I think these conversion impls are a good example of why we'd rather split up these import bindings.

lann (commented):
This was mostly just a shallow pass here with a few surface-level comments.

Looking at how this code is working, I think there may be a viable approach that wouldn't require changes to the outbound factors. I only have a minute right now but briefly:

- Use a [task-local](https://docs.rs/tokio/latest/tokio/task/struct.LocalKey.html) to manage state through the host-guest-host burger
- Use an [otel span processor](https://docs.rs/opentelemetry_sdk/0.31.0/opentelemetry_sdk/trace/trait.SpanProcessor.html) that uses that task-local to conditionally reparent spans

crates/trigger-http/src/wasi.rs
asteurer: Doesn't seem like it's necessary. Removing it unless @calebschoepp would like it included.

crates/world/src/conversions.rs
asteurer: I created a separate crate for `wasi-otel`; however, the `src/lib.rs` file is an exact copy of the `world/src/lib.rs` file. It works; however, I could use some assistance tailoring the `bindgen!` macro to `wasi-otel`.

crates/telemetry/src/env.rs
asteurer: @calebschoepp When you have a sec, would you mind addressing this?

crates/factor-otel/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+            span_processor: self.span_processor.clone(),
+            metric_exporter: self.metric_exporter.clone(),
+        })
+    }
+}
+
+impl OtelFactor {
+    pub fn new() -> anyhow::Result<Self> {
+        // This is a hack b/c we know the version of this crate will be the same as the version of Spin
+        let spin_version = env!("CARGO_PKG_VERSION").to_string();
```
rylev: We can probably just take the version as an argument to the `new` constructor and do this 'hack' from inside the actual Spin crate.

crates/factor-otel/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+    ///
+    /// This MUST be called at the very start of the function before any awaits.
+    pub fn reparent_tracing_span(&self) {
+        // If state is None then we want to return early b/c the factor doesn't depend on the
+        // Otel factor and therefore there is nothing to do
+        let state = if let Some(state) = self.state.as_ref() {
+            state.read().unwrap()
+        } else {
+            return;
+        };
```
rylev: Nit:
```suggestion
        let Some(state) = self.state.as_ref() { state.read().unwrap() } else { return };
```

crates/factor-otel/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// guest span is created and then the host is re-entered to fetch a key value.
+    ///
+    /// ```text
+    /// | GET /... _________________________________|
+    ///    | execute_wasm_component foo ___________|
+    ///       | my_guest_span ___________________|
+    ///          | spin_key_value.get |
+    /// ```
+    ///
+    ///  Setting the guest spans parent as the host is enabled through current_span_context.
```
rylev:
```suggestion
    /// Setting the guest spans parent as the host is enabled through current_span_context.
```

crates/factor-otel/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        // Ensure that we are not reparenting the original host span
+        if let Some(original_host_span_id) = state.original_host_span_id {
+            if tracing::Span::current()
+                .context()
+                .span()
+                .span_context()
+                .span_id()
+                .eq(&original_host_span_id)
+            {
+                panic!("Incorrectly attempting to reparent the original host span. Likely `reparent_tracing_span` was called in an incorrect location.")
```
rylev: Can you talk a bit about the choice to panic here rather than return an error?

crates/factor-otel/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// span. [`OtelFactorState::reparent_tracing_span`] handles this by reparenting the current span to
+    /// be a child of the last active guest span (which is tracked internally in the otel factor).
+    ///
+    /// Note that if the otel factor is not in your [`RuntimeFactors`] than this is effectively a
+    /// no-op.
+    ///
+    /// This MUST only be called from a factor host implementation function that is instrumented.
+    ///
+    /// This MUST be called at the very start of the function before any awaits.
+    pub fn reparent_tracing_span(&self) {
```
rylev, line 256: Does this mean the host loses these spans for its own tracing? Ideally we would emit these spans both for host _and_ guest tracing.

crates/factor-otel/src/host.rs
```diff
@@ -0,0 +1,88 @@
+use crate::InstanceState;
+use anyhow::anyhow;
+use anyhow::Result;
+use opentelemetry::trace::TraceContextExt;
+use opentelemetry_sdk::error::OTelSdkError;
+use opentelemetry_sdk::metrics::exporter::PushMetricExporter;
+use opentelemetry_sdk::trace::SpanProcessor;
+use tracing_opentelemetry::OpenTelemetrySpanExt;
+// TODO: This feels weird. I'm wondering if it can be fixed by wrangling the bindgen macro in `crates/wasi_otel/lib.rs`.
```
rylev: Nit: comments like this saying things like "this feels weird" is rarely helpful. It's usually not clear to readers (likely even the author themselves in the future) what feels weird. We should either address this issue now, write a more descriptive comment, or remove the comment entirely.

tests/test-components/components/otel-smoke-test/src/lib.rs
rylev, line 1: This app doesn't seem to do anything specific with wasi. Why do we need this instead of using any of the existing components?

tests/test-components/components/wasi-otel-tracing/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+anyhow = "1"
+http = "0.2"
+opentelemetry = "0.30.0"
+opentelemetry_sdk = "0.30.0"
+opentelemetry-wasi = { git = "https://github.com/calebschoepp/opentelemetry-wasi", rev = "60df119fbc03ed83d0e2cc1300f6a8eb055d23f7" }
```
rylev: I assume the plan is to upstream this back somewhere? Do we know where?

tests/test-components/components/wasi-otel-tracing/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    let req = Request::builder()
+        .method(Method::Get)
+        .uri("https://asdf.com")
+        .build();
+    let _res: Response = spin_sdk::http::send(req).await.unwrap();
+}
+
+// TODO: Test what happens if start is called but not end
+// TODO: Test what happens if end is called but not start
+// TODO: What happens if child span outlives parent
```
rylev: Let's answer all these questions

crates/factor-otel/src/host.rs
asteurer: Yeah, this was intended as something to solicit help from reviewers, not to be merged. If it's not something we want to worry about, happy to remove it.

crates/factor-otel/src/lib.rs
asteurer: I **_think_** I did this correctly. Will you double check and let me know?

crates/factor-otel/src/lib.rs
asteurer: @calebschoepp I changed this to an error. Let me know if you would prefer to have it panic.

crates/factor-otel/src/lib.rs
asteurer: @calebschoepp Will you address this when you have a sec?

tests/test-components/components/otel-smoke-test/src/lib.rs
asteurer: @calebschoepp Will you address this when you have a sec?

tests/test-components/components/wasi-otel-tracing/Cargo.toml
asteurer: @calebschoepp Will you address this when you have a sec?

crates/telemetry/src/env.rs
asteurer: Actually, I'm just going to revert this. @calebschoepp let me know if you had something else in mind.

crates/factor-otel/src/lib.rs
calebschoepp: I'm wondering if it should still panic. If we're ever reparenting the original host span something has gone wrong and there is a bug in our code (like it says probably in where we called `reparent_tracing_span`). My logic is lets fail fast and catch those bugs.

Alternatively we could just log an error and bail out of the function and do nothing in this case.

Returning an error doesn't make sense b/c the callee of `reparent_tracing_span` can't meaningfully do anything about it.

I would vote for the panic.

crates/factor-otel/src/lib.rs
calebschoepp: @rylev your question implies that we emit host and guest spans to different places. That is not the case in the current design.

If we wanted to change things such that host spans could be sent somewhere different than guest spans, then yes we would want to make a copy of the host spans to parent the guest spans and only reparent those ones.

tests/test-components/components/wasi-otel-tracing/Cargo.toml
calebschoepp: No clear plans yet. Into all the language specific OTel repos is probably the long term plan but I expect that to take a long time. A more medium term plan would be `calebschoepp/opentelemetry-wasi` goes to to live in the OTel github organization.

tests/test-components/components/otel-smoke-test/src/lib.rs
calebschoepp: I'm not sure why this is being modified actually. I thought this smoke test was already checked in to main.

crates/trigger-http/src/wasi.rs
calebschoepp: So long as we still get the traces we expect I'd say chuck it.

tests/test-components/components/otel-smoke-test/src/lib.rs
asteurer: @calebschoepp I'm going to delete this for now. Let me know if you think there's a need to build additional smoke tests.

calebschoepp:
@lann do you mind taking another pass at this?

crates/factor-key-value/src/host.rs
```diff
@@ -49,23 +50,26 @@ pub struct KeyValueDispatch {
     manager: Arc<dyn StoreManager>,
     stores: Table<Arc<dyn Store>>,
     compare_and_swaps: Table<Arc<dyn Cas>>,
+    otel_context: Option<OtelFactorState>,
```
lann: I think we renamed something at some point and now these field names are a little confusing; can we rename them e.g.

```suggestion
    otel_state: Option<OtelFactorState>,
```

crates/factor-key-value/src/host.rs
```diff
@@ -49,23 +50,26 @@ pub struct KeyValueDispatch {
...
 }
 
 impl KeyValueDispatch {
     pub fn new(allowed_stores: HashSet<String>, manager: Arc<dyn StoreManager>) -> Self {
-        Self::new_with_capacity(allowed_stores, manager, DEFAULT_STORE_TABLE_CAPACITY)
+        Self::new_with_capacity(allowed_stores, manager, DEFAULT_STORE_TABLE_CAPACITY, None)
```
lann: Since `OtelFactorState` already manages an `Option` internally I think we could avoid some noise here by adding a `#[derive(Default)]` to `OtelFactorState` which would let us remove the `Option` from `otel_context` here and then this becomes:

```suggestion
        Self::new_with_capacity(allowed_stores, manager, DEFAULT_STORE_TABLE_CAPACITY, Default::default())
```

crates/factor-key-value/src/host.rs
```diff
@@ -113,6 +117,9 @@ impl key_value::Host for KeyValueDispatch {}
 impl key_value::HostStore for KeyValueDispatch {
     #[instrument(name = "spin_key_value.open", skip(self), err, fields(otel.kind = "client", kv.backend=self.manager.summary(&name).unwrap_or("unknown".to_string())))]
     async fn open(&mut self, name: String) -> Result<Result<Resource<key_value::Store>, Error>> {
+        if let Some(otel_context) = self.otel_context.as_ref() {
+            otel_context.reparent_tracing_span();
+        }
```
lann: With the above change these conditionals could be removed:


```suggestion
        otel_context.reparent_tracing_span();
```

crates/telemetry/src/lib.rs
```diff
@@ -53,6 +53,15 @@ pub use propagation::inject_trace_context;
 /// spin_telemetry::metrics::monotonic_counter!(spin.metric_name = 1, metric_attribute = "value");
 /// ```
 pub fn init(spin_version: String) -> anyhow::Result<()> {
+    // This filter globally filters out spans produced by wasi_http so that they don't conflict with
+    // the behaviour of the wasi-otel factor.
```
lann: Unless I'm missing something this comment isn't really accurate - this filter would remove _all_ "wit-bindgen export" spans. Could you say more about why this is necessary? Hopefully we can find a more targetted way to mitigate this.

lann (commented):
This review might take a while - leaving an initial batch of comments now but I'll have to come back to it.

crates/factor-otel/src/lib.rs
calebschoepp: Closing this as it is not relevant to the existing design

crates/factor-key-value/src/host.rs
asteurer: The `StoreManager` type was giving me issues with the `#[derive(Default)]` macro, so I created a default workaround that includes a new type: 

```rust
/// This is enables the KeyValueDispatch type to implement a default method.
/// It is not intended to be useable.
pub struct NoOpStoreManager;
```

Let me know if there are concerns with this path.

asteurer: Ope, I misread the suggestion. Disregard the above.

crates/telemetry/src/lib.rs
calebschoepp: Dug more into this. Discovered that in fact this was not necessary anymore so I removed it. Along the way I found that we weren't properly reparenting some of the outbound wasi http spans so I fixed that up too.

crates/factor-otel/src/lib.rs
```diff
@@ -31,9 +35,19 @@ impl Factor for OtelFactor {
...
 
     fn init(&mut self, ctx: &mut impl spin_factors::InitContext<Self>) -> anyhow::Result<()> {
-        ctx.link_bindings(wasi_otel::wasi::otel::tracing::add_to_linker::<_, FactorData<Self>>)?;
-        ctx.link_bindings(wasi_otel::wasi::otel::metrics::add_to_linker::<_, FactorData<Self>>)?;
-        ctx.link_bindings(wasi_otel::wasi::otel::logs::add_to_linker::<_, FactorData<Self>>)?;
+        // Only link the WASI OTel bindings if experimental support is enabled. This means that if
+        // the user tries to run a guest component that consumes the WASI OTel WIT without enabling
+        // experimental support they'll see an error like "component imports instance
+        // `wasi:otel/tracing@0.2.0-draft`"
+        if self.experimentally_enabled {
```
lann: Micro-nit: no need to change it now but within the implementation I'd name this something closer to what its actually doing, like `enable_interface` or something. The "experiemental" part only really makes sense from the outside. 🤷

crates/factor-key-value/src/host.rs
```diff
@@ -113,6 +122,7 @@ impl key_value::Host for KeyValueDispatch {}
 impl key_value::HostStore for KeyValueDispatch {
     #[instrument(name = "spin_key_value.open", skip(self), err, fields(otel.kind = "client", kv.backend=self.manager.summary(&name).unwrap_or("unknown".to_string())))]
     async fn open(&mut self, name: String) -> Result<Result<Resource<key_value::Store>, Error>> {
+        self.otel_state.reparent_tracing_span();
```
itowlson: nitty suggestion: from a reader's point of view, `_state` seems like clutter here.  I'd suggest `self.otel.reparent...()`.  If someone cares that it's the 'state' of anything, the type will tell them.

crates/factor-otel/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+        // Ensure that we are not reparenting the original host span
+        if let Some(original_host_span_id) = tracing_state.original_host_span_id {
+            if tracing::Span::current()
+                .context()
+                .span()
+                .span_context()
+                .span_id()
+                .eq(&original_host_span_id)
+            {
+                panic!("Incorrectly attempting to reparent the original host span. Likely `reparent_tracing_span` was called in an incorrect location.");
```
itowlson: We're happy that panicking is the right thing to do?  Rather than a debug-only assert or something?

crates/runtime-factors/src/lib.rs
```diff
@@ -93,6 +98,10 @@ pub struct TriggerAppArgs {
     #[clap(long = "allow-transient-write")]
     pub allow_transient_write: bool,
 
+    /// [Experimental] Enable experimental WASI OTel support. Backwards compatibility of the WIT is not guaranteed.
+    #[clap(long = "experimental-wasi-otel")]
+    pub experimental_wasi_otel: bool,
```
itowlson, line 103: We've had some discussion around how to manage experimental feature flags, although I'm not sure we've reached any conclusion.  See https://github.com/spinframework/spin/issues/3318.  @fibonacci1729 used an feature-specific flag for https://github.com/spinframework/spin/pull/3377 so... I guess this is where we ended up.  And this lines up with that.  So I guess we've reachd a de facto conclusion. :shipit: :shipit: :shipit:

crates/runtime-factors/src/lib.rs
```diff
@@ -93,6 +98,10 @@ pub struct TriggerAppArgs {
     #[clap(long = "allow-transient-write")]
     pub allow_transient_write: bool,
 
+    /// [Experimental] Enable experimental WASI OTel support. Backwards compatibility of the WIT is not guaranteed.
```
itowlson, line 101: This could maybe do with some help as to what environment variables need to be on/off for it to be effective?  Or would OTel fans just know?

There seems to be some stuff about separately enabling metrics and logs via environment variables... I assume that is normal and that these are standard OTel env vars?  Because otherwise surfacing them as flag options might be nice.

And I am wondering if someone sets the CLI flag but OTel is ineffectual because of env vars - should we print a startup warning or is this expected and benign?

crates/factor-key-value/src/host.rs
calebschoepp: Good call. Change made.

crates/factor-otel/src/lib.rs
calebschoepp: Good call. Debug assert seems more appropriate here

crates/runtime-factors/src/lib.rs
calebschoepp: Yep everything is standard OTel environment variables. I think between the warning I've added below and documentation for tracing it should hopefully be clear enough to users.

> And I am wondering if someone sets the CLI flag but OTel is ineffectual because of env vars - should we print a startup warning or is this expected and benign?

Good call, I've added a warning in that scenario.

wit/deps/otel/world.wit
```diff
@@ -0,0 +1,8 @@
+package wasi:otel@0.2.0-draft;
```
itowlson: `0.2.0-draft` may be a moving target.  Is there a way we can identify a snapshot, e.g. date, or would that break things?

wit/world.wit
```diff
@@ -12,6 +12,7 @@ world platform {
   import wasi:http/outgoing-handler@0.2.6;
   include wasi:cli/imports@0.3.0-rc-2026-01-06;
   import wasi:http/handler@0.3.0-rc-2026-01-06;
+  include wasi:otel/imports@0.2.0-draft;
```
itowlson: We might want to mark this `@unstable` though I dunno how meaningful that is in this context.  Or whether it can be applied to an `include`.  But maybe something to consider.

wit/deps/otel/world.wit
calebschoepp: Yes. We've started cutting [tags](https://github.com/WebAssembly/wasi-otel/releases/tag/v0.2.0-rc.2) for WASI OTel so I'll refer to that.

wit/world.wit
alexcrichton: Are the WITs here shipped to other locations, such as the sdks? If it's just used here in this repo then `@unstable` doesn't really buy much since it's probably going to be specified in the list of features in the `bindgen!` call in this repo. 

If the WITs are shipped to other locations, though, then `@unstable` is good to have here since it'll ensure that other locations don't have this by default (without opting-in). It might be worth double-checking though that `@unstable` works on `include` though since I know that had some weird funny business in the past but I think we've fixed a number of those things over time

calebschoepp: I don't believe the WITs are being shipped anywhere, but I'm not super knowledgeable about this. @itowlson do you have any idea?

FWIW I see an unstable flag on another include in the world so it probably works 🤷. What should the feature for the unstable be if we do it?

calebschoepp: I've added the `@unstable`.

PR #3349, opened 2025-11-05
Allow configurable max buffer size for HTTP connections

fibonacci1729:
This introduces `SPIN_HTTP1_MAX_BUF_SIZE` to allow configuring the underlying maximum read buffer size for HTTP1 connections.


itowlson (approved):
One day (O HAPPY DAY) we should come up with some principles about when to use trigger args vs trigger config vs runtime config vs pure environment variables.

Today, however, need not be that day.

(This thought is prompted by the assumption that the user in the linked issue is going to want their application to _always_ run with this buffer size.  But it seems a bit low level for a trigger config setting, so 🤷.)

PR #3351, opened 2025-11-10
core: Add test_memory_consumed

lann:
This test verifies that `memory_consumed` accounts for initial memory reservations from instantiation.


lann:
`memory_consumed` is a method on `StoreLimitsAsync` which is what is implemented in this file. Basically I wrote this test to confirm my understanding and then realized I might as well commit it. :shrug:

PR #3356, opened 2025-11-18
Include WASI 0.2.6 WITs

itowlson:
This aligns the Spin 3.5 WITs with what we actually support.  This is to address an issue where compilers (e.g. Rust with the `wasm32-wasip2` target) may emit up-level imports that run fine on Spin but fail target world validation.  The risk of regressions seems low because 1. backward compatible and 2. Spin never actually uses these WITs, it just provides whatever the `wasi` crate gives it.  This is more about a correct reference that we can publish for validation and/or use for SDKs etc.

We should figure out whether to retcon the Spin 3.4 WITs too.

Draft for feedback and discussion (and integration testing!).


wit/world.wit
```diff
@@ -8,6 +8,8 @@ world http-trigger {
 
 /// The imports needed for a guest to run on a Spin host
 world platform {
+  include wasi:cli/imports@0.2.6;
+  import wasi:http/outgoing-handler@0.2.6;
```
fibonacci1729, line 12: For posterity, why did we decide on `0.2.6`?

fibonacci1729: Ok I see, `wasmtime v37` uses `0.2.6` across the board it seems so 👍

fibonacci1729:
I think it would make sense to do this for whatever target envs we have published.

fibonacci1729:
For current published envs we have remaining:

```
v3.2.0 -> wasi@0.2.3 (wasmtime 29.0.1)
v3.4.0 -> wasi@0.2.6 (wasmtime 36.0.1)
```

Good news is that updating `v3.4.0` should be a step easier because we already include `wasi@0.2.6` in this PR. I'll take a tab at PRing for `v3.2.0`

wit/deps/clocks/world.wit
```diff
@@ -1,6 +1,11 @@
-package wasi:clocks@0.2.0;
```
fibonacci1729, line 1: For consistency, can we name this folder `wit/deps/clocks@0.2.6`

itowlson: Oops, thanks for catching that Brian!

PR #3359, opened 2025-11-20
Update to Wasmtime 38.0.4 and wasm-tools 240

alexcrichton:
No major changes here, just a routine update. This notably does not update to today's release of Wasmtime 39.0.0 since I some changes in Wasmtime are necessary, but that'll get fixed by 40.0.0


alexcrichton:
I think CI should be good to go now, had to hold back wasm-tools by one version. CI will need a restart though to retry the spurious failure

alexcrichton:
Oh oops, I ned to sign anyway!

PR #3361, opened 2025-11-24
Enable running `wkg oci` packages as bare Wasm

### LINKED ISSUE #3360: Support running components that export `wasi:http/incoming-handler` from registri
`spin up` expects OCI references to resolve to artifacts that contain a `spin.toml` file.  It chokes if you pass an OCI reference that resolves to a simple `wasi:http/incoming-handler` exporter. Instead you have to `wkg fetch` and then run from the local artifact.

```console
$ spin up --from ghcr.io/bowlofarugula/fluid:0.0.1
Error: failed to decode locked app from "/Users/ian/Library/Caches/spin/registry/manifests/ghcr.io/bowlofarugula/fluid/0.0.1/config.json"

Caused by:
      missing field `spin_lock_version` at line 1 column 239
$ wkg oci pull ghcr.io/bowlofarugula/fluid:0.0.1
Successfully wrote ghcr.io/bowlofarugula/fluid:0.0.1 to bowlofarugula_fluid.wasm
$ spin up --from bowlofarugula_fluid.wasm

Serving http://127.0.0.1:3000
Available Routes:
  http-trigger1-component: http://127.0.0.1:3000 (wildcard)
I'll update the doc to show the flow Spin expects with spin registry.
```

itowlson:
Fixes #3360.

I tested this against `wkg oci push`:

```
$ wkg oci push ghcr.io/itowlson/v3vanilla40:1 ./target/wasm32-wasip2/release/v3vanilla40.wasm
$ spin up -f ghcr.io/itowlson/v3vanilla40:1
```

Note that I've only tested to the "hello world" level: if we believe we need to try this on a wider range of binaries then let's chat about desirable test coverage.  I'm also not sure if there's a good way to automate testing - open to advice!

It would be nice to also do it for proper packages (`wkg publish`) but that's probably going to need a bit more investigation and thinky-designy-sort-of-work.

### DIFF
--- a/crates/oci/src/loader.rs
+++ b/crates/oci/src/loader.rs
@@ -14,6 +14,14 @@ pub struct OciLoader {
     working_dir: PathBuf,
 }
 
+/// The artifact loaded by the `OciLoader`
+pub enum ExecutableArtifact {
+    /// The OCI reference contained a Spin application
+    Application(LockedApp),
+    /// The OCI reference contained a Wasm package
+    Package(PathBuf),
+}
+
 impl OciLoader {
     /// Creates a new OciLoader which builds temporary mount directory(s) in
     /// the given working_dir.
@@ -23,9 +31,13 @@ impl OciLoader {
     }
 
     /// Pulls and loads an OCI Artifact and returns a LockedApp with the given OCI client and reference
-    pub async fn load_app(&self, client: &mut Client, reference: &str) -> Result<LockedApp> {
+    pub async fn load_app(
+        &self,
+        client: &mut Client,
+        reference: &str,
+    ) -> Result<ExecutableArtifact> {
         // Fetch app
-        client.pull(reference).await.with_context(|| {
+        let manifest = client.pull(reference).await.with_context(|| {
             format!("cannot pull Spin application from registry reference {reference:?}")
         })?;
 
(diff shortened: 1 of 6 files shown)


crates/oci/src/loader.rs
```diff
@@ -34,42 +46,58 @@ impl OciLoader {
...
-            self.resolve_component_content_refs(component, cache)
-                .await
-                .with_context(|| {
-                    format!("failed to resolve content for component {:?}", component.id)
-                })?;
+        let locked_json: serde_json::Value = serde_json::from_slice(&locked_content)
+            .with_context(|| format!("OCI config {} is not JSON", quoted_path(&lockfile_path)))?;
+
+        if locked_json.get("spin_lock_version").is_some() {
+            let mut locked_app = LockedApp::from_json(&locked_content).with_context(|| {
```
itowlson, line 68: This chunk (to line 89) has not changed from line 51ff in the existing code: it has just moved down and been indented because of going inside the "is it a LockedApp" check.  If I've made this function too long and/or intricate (as I strongly suspect I have), yell at me and I'll split the arms of the `if` into separate helpers.

crates/oci/src/loader.rs
```diff
@@ -14,6 +14,14 @@ pub struct OciLoader {
     working_dir: PathBuf,
 }
 
+/// The artefact loaded by the `OciLoader`
+pub enum ExecutableArtefact {
```
lann: femto-nit: I'm broadly in favor of spelling things as many ways as possible but `artifact` shows up in the OCI spec (as `artifactType`). 🤷

lann (approved):

fibonacci1729 (approved):

crates/oci/src/loader.rs
```diff
@@ -34,42 +46,58 @@ impl OciLoader {
...
+            Ok(ExecutableArtefact::Application(locked_app))
+        } else {
+            if manifest.layers.len() != 1 {
+                anyhow::bail!(
+                    "expected single layer in OCI package, found {} layers",
+                    manifest.layers.len()
+                );
+            }
+            let layer = &manifest.layers[0]; // guaranteed safe by previous check
+            let wasm_path = cache.wasm_path(&layer.digest);
```
kate-goldenring, line 98: Is it worth checking that this component is in fact executable? Aka checking whether the appropriate trigger can run it? I guess that comes down the line and we'd want to do that for spin components too.

itowlson: @kate-goldenring Yes, the trigger takes care of that (and is the only thing that has the knowledge to do it I think).

crates/oci/src/loader.rs
itowlson: Excellent nitting - fixed and thanks for catching it.

src/commands/up.rs
```diff
@@ -467,10 +467,18 @@ impl UpCommand {
                     .await
                     .context("cannot create registry client")?;
 
-                let locked_app = OciLoader::new(working_dir)
+                let artefact = OciLoader::new(working_dir)
                     .load_app(&mut client, reference)
                     .await?;
-                ResolvedAppSource::OciRegistry { locked_app }
+
+                match artefact {
```
jsturtevant: just for consistency
```suggestion
                match artifact {
```

itowlson: That's what I get for using these fancy modern refactoring tools instead of tried and trusted search and replace.  Thanks for the catch @jsturtevant, fixed (and a couple of doc strings too).

itowlson:
@kate-goldenring integration test added (although right now it is going about as well as you'd expect, fie)

PR #3365, opened 2025-12-12
Expose CPU and memory usage on per-component execution spans

### LINKED ISSUE #2933: Expose CPU and memory usage on per-component execution spans

asteurer:
# Overview
This is a feature that enables the collection and emission of CPU and memory usage for each component.
Closes #2933

## Notes about the collected metrics
This measures 3 things:
- Memory usage from component instantiation
- Memory usage from component execution
- The total CPU time elapsed actively running guest code in this instance.

Although [it was proposed](https://github.com/spinframework/spin/issues/2933#issuecomment-2498733789) to measure CPU cycles, I think that that measuring execution time is sufficient for now.

### Prerequisites
- [Spin OTel plugin](https://github.com/fermyon/otel-plugin)
- A Spin application that generates CPU and memory usage - Feel free to use [this one](https://github.com/asteurer/spin-component-metrics-example)

### Generate usage metrics
```sh
cd spin-component-metrics-example
spin otel setup
spin otel open grafana
spin build
spin otel up

# In a different terminal
curl localhost:3000
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ watchexec-filterer-globset = "8.0"
 spin-app = { path = "crates/app" }
 spin-build = { path = "crates/build" }
 spin-common = { path = "crates/common" }
+spin-factors-executor = { path = "crates/factors-executor" }
 spin-doctor = { path = "crates/doctor" }
 spin-environments = { path = "crates/environments" }
 spin-factor-outbound-networking = { path = "crates/factor-outbound-networking" }
@@ -106,12 +107,14 @@ vergen = { version = "^8.2.1", default-features = false, features = [
 ] }
 
 [features]
-default = ["llm"]
+default = ["llm", "cpu-time-metrics"]
 all-tests = ["extern-dependencies-tests"]
 extern-dependencies-tests = []
 llm = ["spin-runtime-factors/llm"]
 llm-metal = ["llm", "spin-runtime-factors/llm-metal"]
 llm-cublas = ["llm", "spin-runtime-factors/llm-cublas"]
+# This enables the collection and emission CPU time elapsed per component execution.
+cpu-time-metrics = ["spin-factors-executor/cpu-time-metrics"]
 
 [workspace]
 members = [
--- a/crates/factors-executor/Cargo.toml
+++ b/crates/factors-executor/Cargo.toml
@@ -13,6 +13,8 @@ anyhow = { workspace = true }
 spin-app = { path = "../app" }
 spin-core = { path = "../core" }
 spin-factors = { path = "../factors" }
+spin-telemetry = { path = "../telemetry" }
+tracing = { workspace = true }
 
 [dev-dependencies]
 spin-factor-wasi = { path = "../factor-wasi" }
@@ -21,3 +23,6 @@ tokio = { workspace = true, features = ["macros", "rt"] }
 
 [lints]
 workspace = true
+
+[features]
+cpu-time-metrics = ["spin-core/call-hook"]
--- a/crates/factors-executor/src/lib.rs
+++ b/crates/factors-executor/src/lib.rs
@@ -1,8 +1,9 @@
+use std::time::{Duration, Instant};
 use std::{collections::HashMap, sync::Arc};
 
 use anyhow::Context;
 use spin_app::{App, AppComponent};
-use spin_core::{async_trait, Component};
+use spin_core::{async_trait, wasmtime::CallHook, Component};
 use spin_factors::{
     AsInstanceState, ConfiguredApp, Factor, HasInstanceBuilder, RuntimeFactors,
     RuntimeFactorsInstanceState,
@@ -255,9 +256,24 @@ impl<T: RuntimeFactors, U: Send> FactorsInstanceBuilder<'_, T, U> {
             core: Default::default(),
             factors: self.factors.build_instance_state(self.factor_builders)?,
             executor: executor_instance_state,
+            cpu_time_elapsed: Duration::from_millis(0),
+            cpu_time_last_entry: None,
+            memory_used_on_init: 0,
+            component_id: self.app_component.id().into(),
         };
         let mut store = self.store_builder.build(instance_state)?;
+
+        #[cfg(feature = "cpu-time-metrics")]
+        store.as_mut().call_hook(|mut store, hook| {
+            CpuTimeCallHook.handle_call_event::<T, U>(store.data_mut(), hook)
+        });
+
         let instance = self.instance_pre.instantiate_async(&mut store).await?;
+
+        // Track memory usage after instantiation in the instance state.
+        // Note: This only applies if the component has initial memory reservations.
+        store.data_mut().memory_used_on_init = store.data().core_state().memory_consumed();
+
         Ok((instance, store))
     }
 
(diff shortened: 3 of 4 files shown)


crates/factors-executor/Cargo.toml
```diff
@@ -21,3 +23,6 @@ tokio = { workspace = true, features = ["macros", "rt"] }
 
 [lints]
 workspace = true
+
+[features]
+call-hook = ["spin-core/call-hook"]
```
lann: I think this could use a more descriptive name, e.g.:


```suggestion

[features]
cpu-time-metrics = ["spin-core/call-hook"]
```

Cargo.toml
```diff
@@ -112,6 +113,8 @@ extern-dependencies-tests = []
 llm = ["spin-runtime-factors/llm"]
 llm-metal = ["llm", "spin-runtime-factors/llm-metal"]
 llm-cublas = ["llm", "spin-runtime-factors/llm-cublas"]
+# This enables the collection and emission CPU time elapsed per component execution.
+call-hook = ["spin-factors-executor/call-hook"]
```
lann: Matching the name change above:


```suggestion
# This enables the collection and emission CPU time elapsed per component execution.
cpu-time-metrics = ["spin-factors-executor/cpu-time-metrics"]
```

crates/factors-executor/src/lib.rs
```diff
@@ -282,6 +328,42 @@ pub struct InstanceState<T, U> {
...
+    /// The memory (in bytes) consumed on initialization.
+    pub mem_usg_init: u64,
+}
+
+impl<T, U> Drop for InstanceState<T, U> {
+    fn drop(&mut self) {
+        // Record the component execution time.
+        #[cfg(feature = "call-hook")]
+        spin_telemetry::metrics::histogram!(
+            spin.component_cpu_time = self.cpu_time_elapsed.as_micros() as f64,
```
lann:
```suggestion
            spin.component_cpu_time = self.cpu_time_elapsed.as_secs_f32(),
```

or `_f64` if the metrics system requires that, but f32 is plenty for this metric.

crates/factors-executor/src/lib.rs
```diff
@@ -282,6 +328,42 @@ pub struct InstanceState<T, U> {
...
+            // According to the OpenTelemetry spec, instruments measuring durations should use "s" as the unit.
+            // See https://opentelemetry.io/docs/specs/semconv/general/metrics/#units
+            unit = "s"
+        );
+
+        // Record the component memory consumed on initialization.
+        spin_telemetry::metrics::histogram!(
+            spin.component_init_memory_used = self.mem_usg_init,
+            component_id = self.component_id,
+            unit = "bytes"
```
lann:
```suggestion
            unit = "By"
```

...apparently.

OpenTelemetry recommends using [UCUM units](https://ucum.org/ucum) (the "c/s" column). This is a nicer reference and shows one place where it actually matters: https://github.com/open-telemetry/opentelemetry-rust/blob/07bb613d0601589e03cb220f139e967eb317923d/opentelemetry-prometheus/src/utils.rs#L55

crates/factors-executor/src/lib.rs
```diff
@@ -282,6 +328,42 @@ pub struct InstanceState<T, U> {
...
+        );
+
+        // Record the component memory consumed during execution.
+        spin_telemetry::metrics::histogram!(
+            spin.component_exec_memory_used = self.core.memory_consumed(),
+            component_id = self.component_id,
+            unit = "bytes"
```
lann:
```suggestion
            unit = "By"
```

crates/factors-executor/src/lib.rs
```diff
@@ -282,6 +328,42 @@ pub struct InstanceState<T, U> {
...
     factors: T,
     executor: U,
+    /// The last time guest code started running in this instance.
+    pub cpu_time_last_entry: Option<Instant>,
+    /// The total CPU time elapsed actively running guest code in this instance.
+    pub cpu_time_elapsed: Duration,
+    /// The component ID.
+    pub component_id: String,
+    /// The memory (in bytes) consumed on initialization.
+    pub mem_usg_init: u64,
```
lann: A couple of suggestions:

- Prefer slightly more verbose names over uncommon abbreviations `usg` -> `usage`
- Fields should default to private unless they need to be `pub`; every field you expose makes it (potentially) a little harder to refactor later. You could argue that `cpu_time_elapsed` might make sense as `pub` but I'd let it wait until something actually needs it.
- I think its nice to group read-only and mutated fields separately; I vaguely think this is somewhat conventional in Rust but it also might just be personal preference

```suggestion
    /// The component ID.
    component_id: String,

    /// The last time guest code started running in this instance.
    cpu_time_last_entry: Option<Instant>,
    /// The total CPU time elapsed actively running guest code in this instance.
    cpu_time_elapsed: Duration,
    /// The memory (in bytes) consumed on initialization.
    mem_usage_init: u64,
```

crates/factors-executor/src/lib.rs
```diff
@@ -282,6 +328,42 @@ pub struct InstanceState<T, U> {
...
+        // Record the component memory consumed on initialization.
+        spin_telemetry::metrics::histogram!(
+            spin.component_init_memory_used = self.mem_usg_init,
+            component_id = self.component_id,
+            unit = "bytes"
+        );
+
+        // Record the component memory consumed during execution.
+        spin_telemetry::metrics::histogram!(
+            spin.component_exec_memory_used = self.core.memory_consumed(),
```
lann: Somewhat bike-shedy here but I'm not sure `exec` makes things more clear here, and could be a little confusing if interpreted as being _separate_ from `init` when its actually both.

```suggestion
            spin.component_memory_used = self.core.memory_consumed(),
```

lann (commented):
Looking good overall; some mostly-superficial suggestions.

I wonder if we should enable this by default for `spin up`; there _is_ a small performance penalty but I'm not sure anyone will actually use this otherwise... 🤔

asteurer:
# Benchmarking
I used the [oha](https://github.com/hatoo/oha) tool to benchmark. The performance differences seem to be small enough to justify including the `cpu-time-metrics` flag by default.

## Spin v3.5.1
- Requests: 10,000
- Connections: 1
### Summary
  Success rate: 100.00%
  Total:        3567.8610 ms
  Slowest:      2.9617 ms
  Fastest:      0.3008 ms
  Average:      0.3454 ms
  Requests/sec: 2802.7998

## Spin v3.6.0-pre0 with `cpu-time-metrics` flag
- Requests: 10,000
- Connections: 1
### Summary
  Success rate: 100.00%
  Total:        3699.8777 ms
  Slowest:      3.2799 ms
  Fastest:      0.3153 ms
  Average:      0.3586 ms
  Requests/sec: 2702.7

lann:
Would you mind doing one more comparison just between the branch as-is and with [the lines that set up the call hook](https://github.com/spinframework/spin/pull/3365/changes#diff-0a8021a17474d3461c85930aab439bcc8ba5d0a9dbe378c52c9f2975245ec073R267-R269) commented out? If most of the overhead is in actually _running_ the call hook then we could maybe have the feature enabled by default but still opt-in at runtime.

asteurer:
# Benchmarking Part II
## With [the lines that set up the call hook](https://github.com/spinframework/spin/pull/3365/changes#diff-0a8021a17474d3461c85930aab439bcc8ba5d0a9dbe378c52c9f2975245ec073R267-R269)
- Requests: 10,000
- Connections: 1
### Summary
  Success rate: 100.00%
  Total:        3744.0815 ms
  Slowest:      2.9720 ms
  Fastest:      0.3169 ms
  Average:      0.3622 ms
  Requests/sec: 2670.8820

## Without [the lines that set up the call hook](https://github.com/spinframework/spin/pull/3365/changes#diff-0a8021a17474d3461c85930aab439bcc8ba5d0a9dbe378c52c9f2975245ec073R267-R269)
- Requests: 10,000
- Connections: 1
### Summary
  Success rate: 100.00%
  Total:        3616.5206 ms
  Slowest:      2.9669 ms
  Fastest:      0.3033 ms
  Average:      0.3499 ms
  Requests/sec: 2765.0886

lann:
OK not exactly scientific but it seems like the bulk of the latency comes from actually running the call hook. I think it probably makes sense to enable the cargo feature by default for `spin` but require opt-in to actually enable the call hook. I'd be fine with doing that in this PR or a followup; up to you.

asteurer:
Yeah, I would prefer to tackle the runtime flag in another PR

asteurer:
@lann If this looks good, would you mind approving so I can merge?

asteurer:
@lann I don't have merge permissions. Do you mind?

PR #3366, opened 2025-12-15
Postgres custom root certificates

itowlson:
Partially fixes https://github.com/fermyon/feedback/issues/64 - this handles the custom root certs side, but not the `sslmode=verify-full` part.

Credit and thanks to @kate-goldenring for figuring out a test case.

## How to specify custom roots

In this PR, trusted roots certificates are referenced in the runtime config file:

```
[postgres]
root_certificates = ["postgres-ssl/ca.crt"]
```

Things to consider:

* This leaves no trace in the manifest that an application might require custom certificates: deployment tools therefore have no way of knowing that they need to also prepare custom certs.  (Although this is probably reasonable.  The deployment environment will likely connect to a different Postgres host, which may or may not require a custom CA when accessed from the deployment environment.)

* This makes it clear that the custom certificates will only be when making Postgres connections.  But this seems potentially weird: if I trust these CAs, why would I not trust them for all my network traffic?  But I am not sure how to set this up.  Sockets have some existing runtime-config network config stuff but it's not clear to me whether that applies to Postgres (e.g. I don't think the Postgres library allows for client certificates) and if so how to re-use it.  But very open to feedback on this: it feels like trust is trust and would be good not to have that partitioned by API.

## Testing

I don't currently have a good way to test this.  Kate wrote an excellent and convenient manual test case, which I've committed as part of this draft, but really we'd want a Docker-based test to run as part of integration.  The existing PG tests are done via conformance-test, but that doesn't seem like a natural home for this (not to mention I'd prefer not to have to deal with the circularity problem again), and yet I'm hazy on how to write PG tests _outside of_ conformance-test.  So... open to advice/suggestions on this.


kate-goldenring:
Thank you for putting this together @itowlson! Is there a reason why we cannot use the existing `client_tls` runtime config? Is it too weird validating the associated host in this scenario?

lann:
It should be possible to use the existing client TLS runtime config but it would require either a) adding support for `native-tls` there or b) switching from `postgres-native-tls` to `postgres-rustls`.

For b) I think the rough approach would be something like:

1. In `OutboundPgFactor::prepare` get a handle to the `OutboundNetworkingFactor`'s `ComponentTlsClientConfigs`:
   https://github.com/spinframework/spin/blob/0ec6bc47426d2f80a376c07a22d08af73082bff2/crates/factor-outbound-http/src/lib.rs#L73
   https://github.com/spinframework/spin/blob/0ec6bc47426d2f80a376c07a22d08af73082bff2/crates/factor-outbound-http/src/lib.rs#L76

1. Pass that through in `InstanceState` (as already done with `OutboundAllowedHosts`) and then at the point where you need TLS config call `get_client_config` to get `TlsClientConfig`: https://github.com/spinframework/spin/blob/0ec6bc47426d2f80a376c07a22d08af73082bff2/crates/factor-outbound-http/src/wasi.rs#L455

1. Trade your `TlsClientConfig` for a `TlsConnector`, something like: 
   ```rust
	let tls_connector = tokio_rustls::TlsConnector::from(tls_client_config.inner().clone())
   ```

1. Then it should be similar to this PR, except with `postgres-rustls` instead: https://docs.rs/postgres_rustls/latest/postgres_rustls/struct.TlsConnector.html

I think a) would be similar but the runtime config parsing would need to be reworked a bit to allow conversion to either `rustls` or `native-tls` config.

itowlson:
The trouble with using the TLS configs from the outbound networking factor is that they are component specific and available at instance state level, but Postgres uses an app-level connection pool.  I guess we could make it pool per component - technically it feels a bit naughty for one component's InstanceState to bleed through to another's but in this case it would be benign.  I am not sure how much faff would be involved, but probably not an excessive amount.  (We could even make it pool per instance but that obviates a lot of the value of pooling.)

kate-goldenring:
> I guess we could make it pool per component - technically it feels a bit naughty for one component's InstanceState to bleed through to another's but in this case it would be benign. I am not sure how much faff would be involved, but probably not an excessive amount. (We could even make it pool per instance but that obviates a lot of the value of pooling.)

I think component level connection pool seems right to me, especially since we allow outbound hosts at the component level. I do not think it should be per instance -- I agree that we want to reuse connections at that level if possible

itowlson:
Yeah I started looking at pool-per-component and that seemed okay but then I ran into another problem: the Postgres connection string can contain multiple hosts, but the outbound networking stuff requires us to pick a host before it will give us a TLS configuration.  Currently investigating to see if this is bridgeable.

kate-goldenring:
> the Postgres connection string can contain multiple hosts

This is strange -- does it create a connection pool amongst those hosts? Is this specific to the pg lib we use or the pg standard? If the former, maybe we only allow info for one host?

itowlson:
It's the tokio-postgres config: https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html#method.get_hosts

itowlson:
The examples (https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html#examples) show multi-host things.  I think it's a preference/fallback kind of thing - I can no longer find the docs but I think I saw something saying it would try the hosts in order.

_ETA:_ yikes you can even do it in the URL style (`postgresql://user@host1:1234,host2,host3:5678`).  I bet we don't handle that gracefully!

kate-goldenring:
> ETA: yikes you can even do it in the URL style (postgresql://user@host1:1234,host2,host3:5678). I bet we don't handle that gracefully!

Definitely yikes! Could we start by limiting this and then maybe be more permissive in the future if there is a feature request. I know this means we need a custom parser for url style.

itowlson:
Okay I think we do handle URL style correctly, because we use the `Config` parser (which handles both connection string and URL style) for permission checking - we never try to parse the URL ourselves - and tokio-postgres takes care of the rest.

We could bypass the bug by checking if the `.hosts()` value has length 1, and barfing if not.  If I'm right about multiple hosts currently working, that would be a breaking change.  And it feels arse about face to restrict multi-host just because our internal TLS API needs to know the host up front - maybe we need to dig into the outbound-networking TLS API a bit deeper...?

lann:
Based on docs (and not writing a lick of code) it seems like it ought to be possible to write an implementation of [`tokio_postgres::tls::MakeTlsConnect`](https://docs.rs/tokio-postgres/latest/tokio_postgres/tls/trait.MakeTlsConnect.html) that wraps [`postgres_rustls::MakeTlsConnector`](https://docs.rs/postgres_rustls/latest/postgres_rustls/struct.MakeTlsConnector.html) which should give us the single hostname we need.

PR #3367, opened 2025-12-18
Disable tls 1.2

loganaden:
TLS 1.2 is not getting updates. It might be time to consider using only tls 1.3 as default.


lann:
TLS 1.2 is still widely deployed and cannot be disabled in a general-purpose tool.

PR #3377, opened 2026-01-21
Allow enabling experimental wasm features

fibonacci1729:
Example: 
``` 
spin up \
   --experimental-wasm-feature gc \
   --experimental-wasm-feature exceptions \
   --experimental-wasm-feature function-references \
   --experimental-wasm-feature reference-types \
...
```

* Feature gated via `experimental-wasm-features`.
* Enabled for canary builds

### DIFF
--- a/crates/trigger/Cargo.toml
+++ b/crates/trigger/Cargo.toml
@@ -13,6 +13,8 @@ rust-version.workspace = true
 # `ComponentLoader::enable_loading_aot_compiled_components`
 # documentation for more information about the safety risks.
 unsafe-aot-compilation = []
+# Enables configuring unstable Wasm features.
+experimental-wasm-features = []
 
 [dependencies]
 anyhow = { workspace = true }
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -10,6 +10,8 @@ use std::path::PathBuf;
 use std::{future::Future, sync::Arc};
 
 use anyhow::{Context, Result};
+#[cfg(feature = "experimental-wasm-features")]
+use clap::ValueEnum;
 use clap::{Args, IntoApp, Parser};
 use spin_app::App;
 use spin_common::sloth;
@@ -111,6 +113,10 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
     )]
     pub runtime_config_file: Option<PathBuf>,
 
+    #[cfg(feature = "experimental-wasm-features")]
+    #[clap(long, value_enum)]
+    pub experimental_wasm_feature: Vec<ExperimentalWasmFeature>,
+
     /// Set the application state directory path. This is used in the default
     /// locations for logs, key value stores, etc.
     ///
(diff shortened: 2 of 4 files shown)


crates/trigger/Cargo.toml
```diff
@@ -13,6 +13,7 @@ rust-version.workspace = true
 # `ComponentLoader::enable_loading_aot_compiled_components`
 # documentation for more information about the safety risks.
 unsafe-aot-compilation = []
+experimental-wasm-features = []
```
itowlson, line 17: Worth a docs comment or not?

crates/trigger/src/cli.rs
```diff
@@ -133,6 +139,15 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
     pub launch_metadata_only: bool,
 }
 
+#[cfg(feature = "experimental-wasm-features")]
+#[derive(Clone, Debug, ValueEnum)]
+pub enum ExperimentalWasmFeature {
+    Gc,
+    ReferenceTypes,
```
itowlson, line 146: Does clap know to punctuate multi-word enum values?

crates/trigger/src/cli.rs
```diff
@@ -111,6 +113,10 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
     )]
     pub runtime_config_file: Option<PathBuf>,
 
+    #[cfg(feature = "experimental-wasm-features")]
+    #[clap(long, value_enum)]
+    pub experimental_wasm_features: Vec<ExperimentalWasmFeature>,
```
itowlson: This is multi-use, right?  If so, it should probably be surfaced as a singular e.g. `--experimental-wasm-feature gc --experimental-wasm-feature reference-types` (because each occurrence is one feature)

crates/trigger/src/cli.rs
```diff
@@ -212,6 +227,26 @@ impl<T: Trigger<B::Factors>, B: RuntimeFactorsBuilder> FactorsTriggerCommand<T,
             config.disable_pooling();
         }
 
+        #[cfg(feature = "experimental-wasm-features")]
+        {
+            let wasmtime_config = config.wasmtime_config();
+            for (i, wasm_feature) in self.experimental_wasm_features.iter().enumerate() {
+                if i == 0 {
```
itowlson: This feels rather more complicated and less self-explanatory than e.g. `if !self.experimental_wasm_features.is_empty() { warn(...) }` outside the loop

crates/trigger/src/cli.rs
```diff
@@ -212,6 +227,26 @@ impl<T: Trigger<B::Factors>, B: RuntimeFactorsBuilder> FactorsTriggerCommand<T,
...
+                    tracing::warn!("You are using experimental Wasm features. These are unstable and may go away in a future release");
```
itowlson: Do you want `tracing::warn` or `terminal::warn` here?  I believe `tracing::warn` is not shown by default and I'm not sure if that was your intention.  (But if it's not important enough to be shown by default then I'd just drop the warning.  The user already knows they typed this long-ass flag, they're not going to turn on logging to help them realise it.)

crates/trigger/src/cli.rs
fibonacci1729: The intention was to enable `--experimental-wasm-features gc,reference-types,...`. I forgot the `value_delimiter` incantation.

itowlson: We normally use (or at least allow) multi-thingy whatsits instead of honkin' CSV strings.  Is there a reason for diverging from that here?

fibonacci1729: Thats fair if there is a precedent. I was just going off https://github.com/spinframework/spin/issues/3364#issuecomment-3617385233 and https://github.com/spinframework/spin/issues/3364#issuecomment-3623069188

fibonacci1729: Updated to use `--experimental-wasm-feature`

PR #3378, opened 2026-01-21
Extract dependency WITs during build

### LINKED ISSUE #3376: Emit WITs during build
We want to make it more seamless for application components to consume dependencies: that is, once you add the dependencies to your `spin.toml`, the imported interfaces become readily available for you to program and build against in your main component.  We've explored various ways of doing this, but the current favourite is to have `spin build` extract the dependency WITs and emit them to a well-known location.

We have prior art here from the `spin deps` plugin, but we now want that functionality in the Spin CLI and integrated into `spin build`.  A consequence of this is language neutrality: the `spin deps` plugin had language commands that did some nice work on the code side, but Spin generally tries not to know about specific languages.

We need to design it so it can be called from outside `spin build`.  We've always had a philosophy that you can build Spin apps using your toolchain of choice, whether that's Spin, make, or whatever.  So we should expose the "create or update WITs" functionality in such a way that other toolchains can use it (although `spin build` will do it automatically).  Ideally it should be possible to run continuously or via LSP so that WITs are updated as dependencies are added and removed (and so compiler errors and code completion update), but this may be impractical: in any event, it's not required on day one.

Language templates or SDKs will want to update to integrate seamlessly with the generated WITs but that's out of scope for this issue.

We should agree where to store extracted WITs.  `spin deps` put them under `.spin`, which is natural - it is a bit of mix of various kinds of data but we can create a subdirectory to isolate the extracted WITs.

itowlson:
Fixes #3376.

WIP. Current status: 1. ~contains gross, gross hacks~ replaced with BEAUTIFUL AND ELEGANT hacks, and 2. incomplete.

Work needed:

* ~Handle `export`~
* ~Test that `include` is handled correctly~ I think this works but a second tester would be very welcome
* Unit tests (and e2e tests and examples)
* ~Provide a way to invoke it outside of build~ (can be deferred if naming is hard, which it always is)

### DIFF
--- a/crates/build/src/lib.rs
+++ b/crates/build/src/lib.rs
@@ -25,6 +25,7 @@ pub async fn build(
     profile: Option<&str>,
     component_ids: &[String],
     target_checks: TargetChecking,
+    wit_generation: GenerateDependencyWits,
     cache_root: Option<PathBuf>,
 ) -> Result<()> {
     let build_info = component_build_configs(manifest_file, profile)
@@ -37,7 +38,26 @@ pub async fn build(
         })?;
     let app_dir = parent_dir(manifest_file)?;
 
-    let build_result = build_components(component_ids, build_info.components(), &app_dir);
+    let components_to_build = components_to_build(component_ids, build_info.components())?;
+
+    if wit_generation.generate() {
+        let wit_gen_errs = regenerate_wits(&components_to_build, &app_dir).await;
+        if !wit_gen_errs.is_empty() {
+            terminal::warn!("One or more components specified dependencies for which Spin couldn't generate import interfaces.");
+            eprintln!(
+                "If these components rely on Spin-generated interfaces they may fail to build."
+            );
+            eprintln!(
+                "Otherwise, to skip interface generation, use the --skip-generate-wits flag."
+            );
+            eprintln!("Error details:");
+            for (component, err) in wit_gen_errs {
+                terminal::einfo!("{component}:", "{err:#}");
+            }
+        }
+    }
+
+    let build_result = build_components(components_to_build, &app_dir);
 
     // Emit any required warnings now, so that they don't bury any errors.
     if let Some(e) = build_info.load_error() {
@@ -109,16 +129,16 @@ pub async fn build_default(
         profile,
         &[],
         TargetChecking::Check,
+        GenerateDependencyWits::Generate,
         cache_root,
     )
     .await
 }
 
-fn build_components(
+fn components_to_build(
     component_ids: &[String],
     components: Vec<ComponentBuildInfo>,
-    app_dir: &Path,
-) -> Result<(), anyhow::Error> {
+) -> anyhow::Result<Vec<ComponentBuildInfo>> {
     let components_to_build = if component_ids.is_empty() {
         components
     } else {
@@ -139,6 +159,40 @@ fn build_components(
             .collect()
     };
 
+    Ok(components_to_build)
+}
+
+#[must_use]
+async fn regenerate_wits(
+    components_to_build: &[ComponentBuildInfo],
+    app_root: &Path,
+) -> Vec<(String, anyhow::Error)> {
+    let mut errors = vec![];
+
+    for component in components_to_build {
+        let component_dir = match component.build.as_ref().and_then(|b| b.workdir.as_ref()) {
+            None => app_root.to_owned(),
+            Some(d) => app_root.join(d),
+        };
+        let dest_file = component_dir.join("spin-dependencies.wit");
+        let extract_result = spin_dependency_wit::extract_wits_into(
+            component.dependencies.inner.iter(),
+            app_root,
+            dest_file,
+        )
+        .await;
+        if let Err(e) = extract_result {
+            errors.push((component.id.clone(), e));
+        }
+    }
+
+    errors
+}
+
+fn build_components(
+    components_to_build: Vec<ComponentBuildInfo>,
+    app_dir: &Path,
+) -> anyhow::Result<()> {
     if components_to_build.iter().all(|c| c.build.is_none()) {
         println!("None of the components have a build command.");
         println!("For information on specifying a build command, see https://spinframework.dev/build#setting-up-for-spin-build.");
@@ -414,6 +468,21 @@ impl TargetChecking {
     }
 }
 
+/// Specifies dependency WIT generation behaviour
+pub enum GenerateDependencyWits {
+    /// The build should generate WITs for component dependencies.
+    Generate,
+    /// The build should not generate WITs.
+    Skip,
+}
+
+impl GenerateDependencyWits {
+    /// Should the build generate dependency WITs?
+    fn generate(&self) -> bool {
+        matches!(self, Self::Generate)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -426,26 +495,47 @@ mod tests {
     #[tokio::test]
     async fn can_load_even_if_trigger_invalid() {
         let bad_trigger_file = test_data_root().join("bad_trigger.toml");
-        build(&bad_trigger_file, None, &[], TargetChecking::Skip, None)
-            .await
-            .unwrap();
+        build(
+            &bad_trigger_file,
+            None,
+            &[],
+            TargetChecking::Skip,
+            GenerateDependencyWits::Skip,
+            None,
+        )
+        .await
+        .unwrap();
     }
 
     #[tokio::test]
     async fn succeeds_if_target_env_matches() {
         let manifest_path = test_data_root().join("good_target_env.toml");
-        build(&manifest_path, None, &[], TargetChecking::Check, None)
-            .await
-            .unwrap();
+        build(
+            &manifest_path,
+            None,
+            &[],
+            TargetChecking::Check,
+            GenerateDependencyWits::Skip,
+            None,
+        )
+        .await
+        .unwrap();
     }
 
     #[tokio::test]
     async fn fails_if_target_env_does_not_match() {
         let manifest_path = test_data_root().join("bad_target_env.toml");
-        let err = build(&manifest_path, None, &[], TargetChecking::Check, None)
-            .await
-            .expect_err("should have failed")
-            .to_string();
+        let err = build(
+            &manifest_path,
+            None,
+            &[],
+            TargetChecking::Check,
+            GenerateDependencyWits::Skip,
+            None,
+        )
+        .await
+        .expect_err("should have failed")
+        .to_string();
 
         // build prints validation errors rather than returning them to top level
         // (because there could be multiple errors) - see has_meaningful_error_if_target_env_does_not_match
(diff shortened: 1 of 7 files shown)


itowlson:
I think `include` might be a non-issue because Spin dependencies are on interfaces or packages, not on worlds, and interfaces don't `include` as far as I can tell (only worlds do).

I do determine the exported interfaces by looking at worlds, but I believe it's safe to ignore `include` directives for this, because an interface can be exported through an `include` if there's a world that exports it, and because we iterate the worlds we will pick it up from there.

I am not sure if there is some edge case when a component exports interfaces from other packages but that seems like a potential issue regardless of `include`.  Bother.

crates/build/src/lib.rs
```diff
@@ -33,7 +34,24 @@ pub async fn build(
...
+    let components_to_build = components_to_build(component_ids, build_info.components())?;
+
+    if wit_generation.generate() {
+        let wit_gen_errs = regenerate_wits(&components_to_build, &app_dir).await;
+        if !wit_gen_errs.is_empty() {
+            terminal::warn!("One or more components specified dependencies for which Spin couldn't generate import bindings.");
+            eprintln!(
+                "If these components rely on Spin-generated bindings they may fail to build."
+            );
+            eprintln!("Otherwise, to skip binding generation, use the --skip-generate-wits flag.");
```
karthik2804: is it accurate to call it bindngs? I am not really sure how to articulate it but to me it feels like bindings are what the language toolchain will generate based on the importized wit. This is a very loosely held opinion, so feel free to ignore.

itowlson: Yeah I was trying to find a wording that didn't presume intimate knowledge of Component Model internals.  "Bindings" was the least worst one I could think of (because the _effect in your program_ is that yay you have bindings or boo you don't) but I agree it's technically incorrect (the worst kind of incorrect).  I'm very open to suggestions for this!

fibonacci1729: If I squint bindings feels ok to me. I might just replace "bindings" with "interfaces". I'm sympathetic to not wanting to presuppose too much CM knowledge but interfaces in this context, at least to me, transcends the WIT specifics.

crates/dependency-wit/src/lib.rs
```diff
@@ -0,0 +1,481 @@
...
+fn importize(
+    decoded: DecodedWasm,
+    world: Option<&str>,
+    out_world_name: Option<&String>,
+) -> anyhow::Result<DecodedWasm> {
+    let (mut resolve, world_id) = match (decoded, world) {
+        (DecodedWasm::Component(resolve, world), None) => (resolve, world),
+        (DecodedWasm::Component(..), Some(_)) => {
+            anyhow::bail!(
+                "the `--importize-world` flag is not compatible with a \
```
fibonacci1729: We probably don't want to surface this error back to the user.

I _think_ what we can do here IIUC is what is done in the following case-block:

```
(DecodedWasm::Component(resolve, id), world) => {
  let world = resolve.select_world(&[id], world)?;
  (resolve, world)
}
```

itowlson: Good catch - I had to copy a bunch of stuff from wasm-tools because it was in the CLI but not the library, and was just too darn diligent about preserving it.  Well, that's my story and I'm sticking to it.

itowlson: @fibonacci1729 The proposed fix doesn't work because `id` is a WorldId and `select_world` requires a PackageId, and I'm not sure how to address that.  Can you expand on your intention here?

fibonacci1729: Yeah definitely confusing.

I might rewrite `importize` here to be as minimal for what we actually need at this time. AFAICT `importize` is only ever called with `world=None`.

```
fn importize(
    decoded: DecodedWasm,
    out_world_name: Option<&String>,
) -> anyhow::Result<DecodedWasm> {
    let (mut resolve, world_id) = match decoded {
        DecodedWasm::Component(resolve, world) => (resolve, world),
        DecodedWasm::WitPackage(resolve, id) => {
            let world = resolve.select_world(&[id], None)?;
            (resolve, world)
        }
    };
    resolve
        .importize(world_id, out_world_name.cloned())
        .context("failed to move world exports to imports")?;

    Ok(DecodedWasm::Component(resolve, world_id))
}
```

`wit-bindgen` has a few knobs i don't fully understand.

crates/dependency-wit/src/lib.rs
```diff
@@ -0,0 +1,484 @@
...
+
+fn read_wasm(wasm_bytes: &[u8]) -> anyhow::Result<DecodedWasm> {
+    if wasmparser::Parser::is_component(wasm_bytes) {
+        wit_component::decode(wasm_bytes)
+    } else {
+        let (wasm, bindgen) = wit_component::metadata::decode(wasm_bytes)?;
+        if wasm.is_none() {
+            anyhow::bail!(
+                "input is a core wasm module with no `component-type*` \
+                    custom sections meaning that there is not WIT information; \
```
fibonacci1729: nit: 
```suggestion
                    custom sections meaning that there is no WIT information; \
```

crates/dependency-wit/src/lib.rs
```diff
@@ -0,0 +1,484 @@
...
+        let importised = importize(decoded, None, Some(&impo_world))
+            .with_context(|| format!("failed to map importize dependency {dependency_name}"))?;
+
+        let imports = match import_name {
+            None => all_imports(&importised),
+            Some(itf) => one_import(&importised, itf.as_ref()),
+        };
+
+        // Capture WITs for all packages used in the importised thing.
+        // Things like WASI packages may be depended on my multiple packages
```
fibonacci1729: nit: 

```suggestion
        // Things like WASI packages may be depended on by multiple packages
```

crates/dependency-wit/src/lib.rs
```diff
@@ -0,0 +1,484 @@
...
+        let remap = aggregating_resolve.merge(importised.resolve().clone())?;
+        for iid in imports {
+            let mapped_iid = remap.map_interface(iid, Span::default())?;
+            let wk = wit_parser::WorldKey::Interface(mapped_iid);
+            let world_item = wit_parser::WorldItem::Interface {
+                id: mapped_iid,
+                stability: wit_parser::Stability::Unknown,
+                span: Span::default(),
+            };
+            aggregating_resolve
```
fibonacci1729: If two different dependencies import the same interface (e.g. both depend on `wasi:http/types`), this code inserts into `[aggregating_resolve.worlds[...].imports` with the same `WorldKey`. If the map already contains that key, the second insert silently overwrites it. This works by coincidence (both values should be equivalent), but i think it would be wise to debug-assert that the existing value matches the new one which would guard against subtle breakage.

crates/dependency-wit/src/lib.rs
```diff
@@ -0,0 +1,484 @@
...
+}
+
+fn as_interface(wi: &wit_parser::WorldItem) -> Option<wit_parser::InterfaceId> {
+    match wi {
+        wit_parser::WorldItem::Interface { id, .. } => Some(*id),
+        _ => None,
+    }
+}
+
+fn one_import(wasm: &DecodedWasm, name: &str) -> Vec<wit_parser::InterfaceId> {
```
fibonacci1729: I can't quite convince myself this _can't_ happen but if the interface name here doesn't match anything, the dependency is silently skipped. I think we should consider returning an error or logging a warning here that no import was added since this seems to indicate a misconfiguration at the manifest level.

crates/dependency-wit/src/lib.rs
```diff
@@ -0,0 +1,484 @@
...
+
+    Ok(buf)
+}
+
+fn munge_aliased_export(
+    decoded: DecodedWasm,
+    export: &str,
+    new_name: &DependencyName,
+) -> anyhow::Result<DecodedWasm> {
+    // TODO: I am not sure how `export` is meant to work if you are
```
fibonacci1729: This can't actually happen given the validation here: https://github.com/spinframework/spin/blob/13a2a3580f3ac134925757718d10ed36cc4972e6/crates/manifest/src/schema/v2.rs#L471

crates/dependency-wit/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+spin-manifest = { path = "../manifest" }
+spin-serde = { path = "../serde" }
+tokio = { workspace = true, features = ["rt", "time"] }
+wasmparser = { workspace = true }
+wit-component = { workspace = true }
+wit-parser = { workspace = true }
+
+[dev-dependencies]
+tempfile = { workspace = true }
+wasm-pkg-common = { workspace = true }
```
fibonacci1729: I think this is a dead dependency

crates/dependency-wit/src/lib.rs
itowlson: Confirmed that it can happen (e.g. `"foo:bar/interface-that-doesnt-exist@1.0.0" = { ... }`) and turned it into an error.

Crossing my fingers for no false positives...

crates/dependency-wit/src/lib.rs
itowlson: VINDICATI-- uh I mean thanks for the pointer (and for having already dealt to it)

itowlson:
Moving to draft until #3421 is resolved.  I believe the code is in a reasonable state, but if we are reworking dependencies then some of it becomes moot or will require rework.

crates/dependency-wit/src/lib.rs
```diff
@@ -0,0 +1,489 @@
...
+        let decoded = read_wasm(&wasm_bytes)?;
+        let decoded = match export {
+            None => decoded,
+            Some(export) => {
+                munge_aliased_export(decoded, &export, dependency_name).with_context(|| {
+                    format!("failed to map named export {export} to dependency {dependency_name}")
+                })?
+            }
+        };
+        let impo_world = format!("impo-world{index}");
```
fibonacci1729, line 67: Can you elaborate on why this extra world is necessary? I'm trying to wrap my head around the shape of the output.

itowlson: I believe I needed to provide a world to the `importize` function.  But maybe not?

(The shape of the output is https://github.com/itowlson/spin-composy-build-script-madness/blob/main/comp-consumer-build-rs-test/spin-dependencies.wit.)

fibonacci1729: Ah i see, likely to avoid collisions with multiple importized worlds. If `None` is specified you end up with `world root-importized { ... }`.

crates/dependency-wit/src/lib.rs
fibonacci1729: Thanks for addressing this!

crates/dependency-wit/src/lib.rs
itowlson: ...or multiple `root-importized`!  That was why, I remember now!

Issue #3380, opened 2026-01-26
outbound-pg: Avoid emitting errors with unredacted connection string


lann:
The connection string will often include a password which should be redacted.

I don't see any really straightforward way to do this today; `tokio_postgres`'s [`Config`](https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html) type properly redacts the password but it only has a very verbose `Debug` impl, not `Display`.

PR #3382, opened 2026-01-27
Fix misleading error on bad PostgreSQL connection string

### LINKED ISSUE #3379: outbound-pg: Wrong error on malformed connection string
The first place that `connection.open` can fail is when checking the host against `allowed_outbound_hosts`: https://github.com/spinframework/spin/blob/a12c2bf7fc85e538254e04f5a36ac01975061ed2/crates/factor-outbound-pg/src/host.rs#L85-L93

This `is_allowed_address` function attempts to parse the connection string and on parse failure returns `Ok(false)`: https://github.com/spinframework/spin/blob/a12c2bf7fc85e538254e04f5a36ac01975061ed2/crates/factor-outbound-pg/src/host.rs#L42-L44

...which then gets translated into a `ConnectionFailed("address {address} is not permitted")` error, which is wrong.

I'm not sure why `is_allowed_address` returns `Ok(false)` on parse error; it should probably either propagate the parse error or - if there is reason not to leak those details - turn it into a fixed e.g. `"invalid connection string"`.

itowlson:
Fixes #3379.

Bad connection string: `Error::ConnectionFailed("invalid connection string: unknown option 'honk'")`

Denied host: `Error::ConnectionFailed("address pantoufles.com is not permitted")`

### DIFF
--- a/crates/factor-outbound-pg/src/host.rs
+++ b/crates/factor-outbound-pg/src/host.rs
@@ -38,10 +38,19 @@ impl<CF: ClientFactory> InstanceState<CF> {
             .ok_or_else(|| v4::Error::ConnectionFailed("no connection found".into()))
     }
 
-    async fn is_address_allowed(&self, address: &str) -> Result<bool> {
-        let Ok(config) = address.parse::<tokio_postgres::Config>() else {
-            return Ok(false);
-        };
+    #[allow(clippy::result_large_err)]
+    async fn ensure_address_allowed(&self, address: &str) -> Result<(), v4::Error> {
+        fn conn_failed(message: impl Into<String>) -> v4::Error {
+            v4::Error::ConnectionFailed(message.into())
+        }
+        fn err_other(err: anyhow::Error) -> v4::Error {
+            v4::Error::Other(err.to_string())
+        }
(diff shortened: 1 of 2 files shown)


crates/factor-outbound-pg/src/host.rs
```diff
@@ -55,15 +64,22 @@ impl<CF: ClientFactory> InstanceState<CF> {
                             .or_else(|| if ports.len() == 1 { ports.get(1) } else { None });
```
lann, line 64: Completely unrelated to this pr but since I'm here and on my phone: surely this is a bug?

itowlson:
This might have fixed at least one occurrence of #3380 as well.

crates/factor-outbound-pg/src/host.rs
itowlson: Could you elaborate on which bit is wrong?

lann: `if ports.len() == 1 { ports.get(1) }`

Unless I'm badly confused this would always return `None`. Ought to be `.get(0)` or better `.first()`.

PR #3383, opened 2026-01-30
Spin 3.6 world for new WASI snapshot

itowlson:
We recently merged Wasmtime 41 and the new WASIp3 snapshot.  This changes the Spin world, so we prefer to create a new package rather than amend the existing one.

In this case it's a bit awkward because going forward we do not support the Spin 3.5-specific worlds that contained the 2025 snapshot.  I have, for now, kept the Spin 3.5 world around with deprecations, which means I needed to bring back the 2025 WITs.  Longer term we will want to remove the Spin 3.5 world so that we can remove the 2025 WITs (and, presumably, something similar with 3.6 and the Jan 2026 WITs).  Perhaps we should purge the Spin 3.5 world right now, but my gentle soul revolts at simply _disappearing_ a version in a minor release - but then again I had to remove it from the world.wit and `world` bindgen so perhaps it is more confusing that the WITs stay around.

Anyway at least now we have something that we can argue about so hurrah for that and an eclair for both of us.

### DIFF
--- a/crates/http/src/trigger.rs
+++ b/crates/http/src/trigger.rs
@@ -38,7 +38,7 @@ const WASI_HTTP_EXPORT_2023_11_10: &str = "wasi:http/incoming-handler@0.2.0-rc-2
 /// The `incoming-handler` export prefix for all `wasi:http` 0.2 versions
 const WASI_HTTP_EXPORT_0_2_PREFIX: &str = "wasi:http/incoming-handler@0.2";
 /// The `handler` export `wasi:http` version 0.3.0-rc-2025-08-15
-const WASI_HTTP_EXPORT_0_3_0_RC_2025_09_16: &str = "wasi:http/handler@0.3.0-rc-2025-09-16";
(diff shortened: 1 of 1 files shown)


itowlson:
The testing for this was in https://github.com/spinframework/spin-rust-sdk/pull/110 and https://github.com/itowlson/spin-middleware-terrifying-nonsense.  So I think it works.  But Spin itself doesn't particularly use the world, so...

itowlson:
@alexcrichton @fibonacci1729 @dicej also suggests dropping the 3.5 package: https://github.com/spinframework/spin-rust-sdk/pull/110#issuecomment-3844136429

If folks can reassure me that there are no compatibility implications for people using SDKs/writing apps that target the `spin:up@3.5.0` package _but do not use the experimental features_ then I'm good to go with this.  I _think_ it will work because the binaries end up targeting interfaces, none of which are in `spin:up@3.5.0`, rather than worlds (which are).  Right?

itowlson:
okay I tested with 3.5 fully purged and it was dandy. I am reassured.

itowlson:
Asking for re-reviews because purge - thanks and sorry for the noise!

itowlson:
sneaked in a 25ectomy in the trigger

PR #3384, opened 2026-02-03
HTTP middleware using WASIp3

### LINKED ISSUE #3291: WASI P3 HTTP middleware
As part of WASI P3 work, we want to explore implementing HTTP middleware using component composition.  The purpose is to allow application developers to use off-the-shelf or custom components in a pipeline to add or customise behaviour without needing to bake it into the core application component.

For example, the following site requires a CORS check and authentication, and enriches the request with geolocation information.  None of these are core application concerns, and they could ideally be pulled off the shelf rather than custom coded.

Image

## Manifest

Spin cannot naturally do this in the current dependency model, because dependencies model only imports of the application component.  (It can do one level of middleware unnaturally, by making the main component the middleware and having it depend on the application, but 1. that currently allows only one level and 2. that's unnatural, ew.)

So to support middleware we have to decide:

1. Do we try to wedge it into the current dependencies section?
2. If not, where do we want to put it?

The trouble with using the dependencies section is that it's a map, whereas the middleware pipeline is a sequence.  All middleware components use the same interface, and the order of plugging them together is significant.  I'm sure some creative soul has an idea for how to overcome this mismatch, but it defeated me.  Ideas welcome!

If we can't use the dependencies section, where do we put it?  The two options appear to be the trigger or the component:

```toml
# Trigger?
[[trigger.http]]
route = "/..."
component = "cat-videos"
pipeline = ["cors", "auth", "geo"]

# Or alternative trigger?
[[trigger.http]]
route = "/..."
component = ["cors", "auth", "geo", "cat-videos"]

# Or component?
[component.cat-videos]
source = "catvids.wasm"
pipeline = ["cors", "auth", "geo"]
```

I'm attracted to putting it on the trigger, because:

1. It's HTTP specific - it can be part of the HTTP trigger schema rather than the common component schema
2. It frames middleware as part of the serving pipeline rather than as inherent to the component

(okay mostly 1 I admit)

However, all our composition and Wasm loading infrastructure is component centric.  During prototyping (https://github.com/itowlson/spin/tree/http-middleware) I found it really hard to see how the relevant subsystems could reach into trigger-specific config this way - for example how would `spin registry push` know that this field on the HTTP trigger required binaries to be resolved and packaged?

So for prototyping I've been putting it on the component.  But I'd value discussion and feedback on this.

## Interface

Middleware needs to be able to pass a request on to the next entry in the chain.  In principle this could be done by composing each component's `wasi:http/handler` import onto the next one's export, then the middleware code calling `handle` on the incoming request to pass it on.  But many middleware items will want to do their own HTTP stuff, e.g. authenticating using OIDC or whatever.

https://github.com/WebAssembly/wasi-http/issues/167 proposes adding an `origin` interface to the WASI-HTTP proxy world.  This would be an additional interface identical to `handler`, but with the semantics of "pass it on along the chain" rather than "fling it across the network."  My prototype called this `next` and had it in a Spin-specific package, but the concept is the same.  If `origin` lands soon enough then we can use that, otherwise we use our own and adopt the spec one when it arrives.

Unlike dependencies, Spin (or the HTTP trigger, or whatever) needs special knowledge of the middleware origin/next interface, because it's not amenable to plug composition (and we don't want devs to have to specify the mapping on every pipeline entry!).

## Permissions

Currently, dependencies get the choice of "main component permissions or nothing," governed by the `dependencies_inherit_configuration` flag.  We should not expect fine-grained permissions in the middleware timeframe, so middleware will likely inherit the same permissions behaviour, governed by the same flag.  This fits nicely with "middleware is an attribute of the component" but not so nicely with it being on the trigger (because if the component needs to allow-list the OIDC provider then it is not middleware-agnostic after all).

This is a bit of a brain dump and I am sure other brains have thought about this a lot more than mine so please weigh in.

itowlson:
Fixes #3291.

~This currently includes #3383; once that lands I will rebase.  For now just ignore the `wit` directory.~ Done!

Example usage: https://github.com/itowlson/spin-middleware-terrifying-nonsense

Setting it up in spin.toml:

```
[[trigger.http]]
route = "/..."
component = "middleware-terrifying-nonsense"
components.middleware = ["middlybiddly", "middlybiddly2"]
executor = { type = "wasip3-unstable" }
```

Implementing a middleware component in Rust (today: we'd presumably provide a wrapper for the onbound `handle` import):

```
#[spin_sdk::http_wasip3::http_service]
async fn handle(mut request: Request) -> impl IntoResponse {
    munge(&mut request);
    wasi::http0_3_0_rc_2026_01_06::handler::handle(request.into_request().unwrap()).await
    // could process the response here if needed
}
```

Needs lots of renaming, tests, etc. but anyway getting it into the system.  (and yes, yes, I'll squash the commits)


itowlson:
Updated to rework OCI precomposition.  The way I did it before resulted in two code paths which were essentially the same but couldn't be exactly the same.  Now there is only one code path and it is _exactly the same as itself._

I am getting increasingly fearful that I am getting lost in a labyrinth of my own changes and although it all seems to work _for me_ I would really value somebody else looking at it and kicking the tyres because what if I am not testing it right.  I was sure that this last rework had broken _something,_ and yet seemingly not, which makes me worry whether I am testing what I think I'm testing.

itowlson:
One thing that is troubling me with all this is isolation behaviour.  Like if you want a middleware to call an auth provider, then it needs network permissions.  So:

1. the _main_ component has to declare network permissions
2. the middleware then inherits _all_ the component's network permissions

We can probably work around 1 in the increasingly terrifying normalisation step but 2 is a gazillion times more complicated.  The only way around that today is to use a service chaining approach where it passes through the host rather than being directly composed.  (Which is do-able, but a non-trivial rewrite, and we'd need to check that this would work with downstream hosts who don't support service chaining.)

crates/trigger/src/cli.rs
```diff
@@ -214,12 +219,38 @@ impl<T: Trigger<B::Factors>, B: RuntimeFactorsBuilder> FactorsTriggerCommand<T,
             App::new(locked_url, locked)
         };
 
+        // Handle --precompose-only
+        if self.precompose_only {
+            let Some(precompose_component_id) = self.precompose_component_id.as_ref() else {
+                anyhow::bail!("got --precompose-only but no --precompose-component-id");
+            };
```
fibonacci1729, line 226: Was this added to aid in the research? I'm not sure what's happening here.

crates/trigger/src/lib.rs
```diff
@@ -50,6 +50,12 @@ pub trait Trigger<F: RuntimeFactors>: Sized + Send {
         Ok(())
     }
 
+    /// An object which composes extras onto the primary component.
+    /// TODO: the combination of functions and objects and traits is a bit funny and we may/should be able to streamline it.
+    fn complicator() -> impl spin_factors_executor::Complicator {
```
fibonacci1729: I know `complicator` is a placeholder so I'll suggest we use something generic here like `fn compose_extras() -> impl spin_factors_executor::ComposeExtras`.

crates/trigger-http/src/middleware.rs
```diff
@@ -0,0 +1,176 @@
...
+        compose_middlewares(component, middleware_blobs).await
+    }
+}
+
+async fn compose_middlewares<'a>(
+    primary: Vec<u8>,
+    middleware_blobs: impl Iterator<Item = &'a ComplicationData>,
+) -> anyhow::Result<Vec<u8>> {
+    const MW_NEXT_INBOUND: &str = "wasi:http/handler@0.3.0-rc-2026-01-06";
+    const MW_NEXT_OUTBOUND: &str = "wasi:http/handler@0.3.0-rc-2026-01-06";
```
fibonacci1729: Can we unify these into a single `const MW_HANDLER_INTERFACE: &str = "wasi:http/handler@0.3.0-rc-2026-01-06";`?

fibonacci1729 (commented):
Early comments.

crates/loader/src/local.rs
```diff
@@ -540,6 +544,133 @@ impl LocalLoader {
...
+/// We want all component/composition graph information to be in the component,
+/// because the component ID is how Spin looks this stuff up. So if a trigger
+/// contains a `components` table, e.g. specifying middleware, we want to move
+/// that to the component.
+///
+/// But it's possible to have two triggers pointing to the same primary component,
+/// but with different middleware. In this case, we will synthesise a component
+/// for each such trigger, with the same main configuration but with its own
+/// "extra" components.
+fn reassign_extra_components_from_triggers(mut locked: LockedApp) -> LockedApp {
```
fibonacci1729: I think we can avoid cloning the locked app if we break this up into 3 parts:

1. Collect a lightweight `TriggerInfo` struct containing the `trigger_id`, `component_id` and whether "extras" are present.
2. Use those to determine splitting, then clone only the individual components that need it.
3. Move extras from triggers to components.

Something like (untested):
```rust
/// We want all component/composition graph information to be in the component,
/// because the component ID is how Spin looks this stuff up. So if a trigger
/// contains a `components` table, e.g. specifying middleware, we want to move
/// that to the component.
///
/// But it's possible to have two triggers pointing to the same primary component,
/// but with different middleware. In this case, we will synthesise a component
/// for each such trigger, with the same main configuration but with its own
/// "extra" components.
fn reassign_extra_components_from_triggers(mut locked: LockedApp) -> LockedApp {
    use std::collections::{HashMap, HashSet};

    fn trigger_component_id(trigger: &LockedTrigger) -> Option<String> {
        trigger
            .trigger_config
            .get("component")
            .and_then(|v| v.as_str())
            .map(|s| s.to_string())
    }

    fn extra_components(trigger: &LockedTrigger) -> Option<&ValuesMap> {
        trigger
            .trigger_config
            .get("components")
            .and_then(|v| v.as_object())
    }

    fn has_extra_components(trigger: &LockedTrigger) -> bool {
        extra_components(trigger).is_some_and(|xcs| !xcs.is_empty())
    }

    // Collect lightweight trigger metadata to determine which components
    // need splitting. This avoids cloning the entire LockedApp.
    struct TriggerInfo {
        trigger_id: String,
        component_id: String,
        has_extras: bool,
    }

    let trigger_infos: Vec<TriggerInfo> = locked
        .triggers
        .iter()
        .filter_map(|t| {
            trigger_component_id(t).map(|cid| TriggerInfo {
                trigger_id: t.id.clone(),
                component_id: cid,
                has_extras: has_extra_components(t),
            })
        })
        .collect();

    // Group triggers by component ID, find components that need splitting
    // (multiple triggers reference the same component AND at least one has extras).
    let mut cid_triggers: HashMap<&str, Vec<&TriggerInfo>> = HashMap::new();
    for info in &trigger_infos {
        cid_triggers
            .entry(&info.component_id)
            .or_default()
            .push(info);
    }

    let mut seen = HashSet::new();
    let mut disambiguator = 0;

    // For components referenced by multiple triggers where at least
    // one has extras, create synthetic component clones so each trigger gets
    // its own composition graph.
    let needs_splitting: Vec<_> = cid_triggers
        .into_iter()
        .filter(|(_, triggers)| {
            triggers.len() > 1 && triggers.iter().any(|t| t.has_extras)
        })
        .collect();

    for (cid, triggers) in &needs_splitting {
        for info in triggers {
            if !info.has_extras {
                // Unenriched triggers can continue pointing to the original component.
                continue;
            }
            let mut synthetic_id = format!("{cid}-for-{}", info.trigger_id);
            if seen.contains(&synthetic_id) {
                disambiguator += 1;
                synthetic_id = format!("{synthetic_id}-d{disambiguator}");
            }
            seen.insert(synthetic_id.clone());

            let component = locked
                .components
                .iter()
                .find(|c| c.id == **cid)
                .expect("trigger references non-existent component")
                .clone();

            let mut synthetic = component;
            synthetic.id = synthetic_id.clone();
            locked.components.push(synthetic);

            // Update the trigger to point to the new synthetic component
            let trigger = locked
                .triggers
                .iter_mut()
                .find(|t| t.id == info.trigger_id)
                .expect("trigger disappeared during splitting");
            trigger
                .trigger_config
                .as_object_mut()
                .expect("trigger config should be an object")
                .insert("component".into(), synthetic_id.into());
        }
    }

    // Move extras from triggers onto their respective components.
    // Now each enriched trigger has its own component, so composition graphs
    // are uniquely identified by component ID.
    for trigger in &mut locked.triggers {
        if let Some(extras) = extra_components(trigger) {
            if let Some(component_id) = trigger_component_id(trigger) {
                if let Some(component) = locked.components.iter_mut().find(|c| c.id == component_id)
                {
                    component
                        .metadata
                        .insert("trigger-extras".into(), extras.clone().into());
                    component.metadata.insert(
                        "resolve-extras-using".into(),
                        trigger.trigger_type.clone().into(),
                    );
                    trigger
                        .trigger_config
                        .as_object_mut()
                        .expect("trigger config should be an object")
                        .remove("components");
                }
            }
        }
    }

    locked
}

```

crates/loader/src/local.rs
```diff
@@ -941,4 +1079,181 @@ mod test {
...
+        assert_eq!(5, locked_app.components.len());
+    }
+
+    #[tokio::test]
+    async fn enriched_lockfile_multiple_enriched_triggers_per_component_get_split() {
+        let (_wd, locked_app) = load_test_case("extra-components", "three-to-one.toml")
+            .await
+            .unwrap();
+        assert_eq!(4, locked_app.triggers.len());
+        // Splitting should resultm in triggers pointing to different IDs, but the same primary source
```
fibonacci1729:
```suggestion
        // Splitting should result in triggers pointing to different IDs, but the same primary source
```

crates/trigger/src/loader.rs
```diff
@@ -90,17 +128,69 @@ impl<T: RuntimeFactors, U> spin_factors_executor::ComponentLoader<T, U> for Comp
...
+    for (role, role_components) in extras {
+        let components = role_components
+            .as_array()
+            .context("extra components should have been an array")?;
+        let mut complications_for_role = Vec::with_capacity(components.len());
+
+        for component_ref in components {
+            let component_ref = component_ref
+                .as_str()
+                .context("middleware should be strings curently")?;
```
fibonacci1729:
```suggestion
                .context("middleware should be strings currently")?;
```

crates/environments/src/loader.rs
```diff
@@ -209,6 +210,19 @@ impl<'a> spin_compose::ComponentSourceLoader for ComponentSourceLoader<'a> {
             .with_context(|| format!("componentizing {}", quoted_path(&path)))?;
         Ok(component.into())
     }
+
+    async fn load_source(&self, source: &Self::Source) -> anyhow::Result<Vec<u8>> {
+        let path = self
+            .wasm_loader
+            .load_component_source("bippety-boppety", source)
```
fibonacci1729:
```suggestion
            .load_component_source("<extra-component>", source)
```

crates/trigger/src/loader.rs
fibonacci1729: Also i might say: `"middleware references should be strings currently"`

crates/loader/src/local.rs
```diff
@@ -941,4 +1079,181 @@ mod test {
...
+        let (_wd, locked_app) = load_test_case("extra-components", "three-to-one.toml")
+            .await
+            .unwrap();
+        assert_eq!(4, locked_app.triggers.len());
+
+        for t in &locked_app.triggers {
+            assert!(t.trigger_config.get("components").is_none());
+        }
+
+        println!("{}", serde_json::to_string_pretty(&locked_app).unwrap());
```
fibonacci1729: TODO: delete this `println!`

crates/manifest/src/schema/v2.rs
```diff
@@ -664,11 +665,26 @@ mod one_or_many {
         D: Deserializer<'de>,
     {
         let value = toml::Value::deserialize(deserializer)?;
-        if let Ok(val) = T::deserialize(value.clone()) {
-            Ok(vec![val])
+        if let Some(arr) = value.as_array() {
```
fibonacci1729, line 991: Maybe delete the old path commented out below and add a note here:

```
// NOTE: We explicitly check for array first rather than trying T::deserialize
// first, because toml's serde impl will treat an array as a sequence of fields
// to be assigned to struct members (e.g. Component), producing nonsensical results.
```

crates/oci/src/client.rs
```diff
@@ -365,15 +365,25 @@ impl Client {
         let mut components = Vec::new();
         let mut layers = Vec::new();
 
+        let temp_dir = tempfile::tempdir().unwrap();
+        let working_dir = temp_dir.path();
+        let locked_url = write_locked_app(&locked, working_dir).await.unwrap();
```
fibonacci1729: Are these `.unwrap()`s safe? Or should we `.context(...)?` here?

fibonacci1729:
> One thing that is troubling me with all this is isolation behaviour. Like if you want a middleware to call an auth provider, then it needs network permissions. So:
> 
>     1. the _main_ component has to declare network permissions
> 
>     2. the middleware then inherits _all_ the component's network permissions
> 
> 
> We can probably work around 1 in the increasingly terrifying normalisation step but 2 is a gazillion times more complicated. The only way around that today is to use a service chaining approach where it passes through the host rather than being directly composed. (Which is do-able, but a non-trivial rewrite, and we'd need to check that this would work with downstream hosts who don't support service chaining.)

Capability isolation remains a thorn for composition. We either need to accept that middlewares inherit capability from the primary component (or alternatively deny-all by default via some adapter like we do for dependencies with an opt-in) or what you suggest with service-chaining (which I prefer less for the reason you postulate).

crates/trigger-http/src/middleware.rs
```diff
@@ -0,0 +1,176 @@
...
+        if middlewares.is_empty() {
+            return Ok(component);
+        }
+
+        let middleware_blobs = middlewares.iter().map(|cm| &cm.data);
+        compose_middlewares(component, middleware_blobs).await
+    }
+}
+
+async fn compose_middlewares<'a>(
```
fibonacci1729: I didn't want to just push a commit but i would simplify this to use wac-graph (addressing the TODO). You can delete the `wasm-compose`, `wasmparser` and `tempfile` dependencies using this.

Note: I added `chain` as a general purpose thing thinking it would be useful to upstream but for now it can live here.

```rust
/// Chain a list of component packages into a middleware pipeline.
///
/// `packages` is ordered from **outermost** (first to receive a request) to
/// **innermost** (the final handler).  Every component except the last must
/// import a name equal to `import_name` and every component must export a name
/// equal to `export_name`.  In the common middleware pattern these are the same
/// (e.g. both `"handle"`), but they can differ if the WIT uses separate names.
///
/// Returns the [`NodeId`] of the alias for the outermost component's export,
/// ready to be passed to [`CompositionGraph::export`].
///
/// # Errors
///
/// Returns an error if fewer than two packages are provided, or if any
/// alias / argument wiring step fails.
fn chain(
    graph: &mut CompositionGraph,
    packages: &[PackageId],
    import_name: &str,
    export_name: &str,
) -> anyhow::Result<NodeId> {
    if packages.len() < 2 {
        bail!("chain requires at least 2 packages, got {}", packages.len());
    }

    // Start from the innermost component (last in the list) and work outward.
    // The innermost component is instantiated first with no wiring — its
    // unsatisfied imports (if any) will become implicit imports of the
    // composed component.
    let mut iter = packages.iter().rev();
    let innermost = *iter.next().unwrap();
    let mut instance = graph.instantiate(innermost);
    let mut upstream_handle = graph.alias_instance_export(instance, export_name)?;

    // For each remaining component (moving outward), instantiate it and
    // wire the previous component's export into its import.
    for &pkg in iter {
        instance = graph.instantiate(pkg);
        graph.set_instantiation_argument(instance, import_name, upstream_handle)?;
        upstream_handle = graph.alias_instance_export(instance, export_name)?;
    }

    Ok(upstream_handle)
}

async fn compose_middlewares<'a>(
    primary: Vec<u8>,
    middleware_blobs: impl Iterator<Item = &'a ComplicationData>,
) -> anyhow::Result<Vec<u8>> {
    const MW_HANDLER_INTERFACE: &str = "wasi:http/handler@0.3.0-rc-2026-01-06";

    let mut graph = CompositionGraph::new();
    let mut package_ids: Vec<PackageId> = Vec::new();

    // Register middleware packages (outermost → innermost order).
    for (index, blob) in middleware_blobs.enumerate() {
        let bytes: Vec<u8> = match blob {
            ComplicationData::InMemory(data) => data.clone(),
            ComplicationData::OnDisk(path) => tokio::fs::read(path)
                .await
                .with_context(|| format!("reading middleware from {}", path.display()))?,
        };
        let name = format!("middleware{index}");
        let package = Package::from_bytes(&name, None, bytes, graph.types_mut())
            .context("parsing middleware component")?;
        package_ids.push(graph.register_package(package)?);
    }

    // Register the primary component (innermost in the chain).
    let package = Package::from_bytes("primary", None, primary, graph.types_mut())
        .context("parsing primary component")?;
    package_ids.push(graph.register_package(package)?);

    // Wire the pipeline: outermost middleware → … → primary.
    let outermost_export = chain(
        &mut graph,
        &package_ids,
        MW_HANDLER_INTERFACE,
        MW_HANDLER_INTERFACE,
    )?;

    // Export the outermost handler as the composed component's export.
    graph.export(outermost_export, MW_HANDLER_INTERFACE)?;

    Ok(graph.encode(Default::default())?)
}
```

crates/trigger/src/cli.rs
itowlson: This is used by the OCI subsystem during precomposition.  `spin registry push` needs to precompose Wasm binaries for cases like SpinKube where composition at load time isn't available.  But the OCI subsystem doesn't know how to deal with HTTP middleware because that's an HTTP concern, not an OCI concern.  So when the OCI subsystem sees complications in a trigger, such as HTTP middleware, it issues issues a command to the relevant trigger saying "hey, you're so smart, you precompose this for me."

crates/trigger/src/lib.rs
itowlson: Yeah I think I originally had "extras" but if anything I hated that even more than "complication."  I am not sure what a good name would be!

crates/trigger-http/src/middleware.rs
itowlson: We probably can.  Originally they were different, and I guess I wanted to retain clarity about which context we were in at any given point in case we ever needed to prise them apart again.  But yeah that does seem unlikely.

crates/manifest/src/schema/v2.rs
itowlson: > nonsensical

would you compromise on "the valuable guano of our much-admired chiropterans"

crates/trigger-http/src/middleware.rs
itowlson: Thank you @fibonacci1729 - this is so much clearer and well-structured than what I had - I really appreciate you taking the time to figure it out and share it!

crates/manifest/src/schema/v2.rs
fibonacci1729: Prefer it, actually.😂

crates/loader/src/local.rs
itowlson: I had another look at this, and ended up tackling it a different way - I've moved this huge pile of code to its own module, broke out the helper functions, defaffed the synthetic ID stuff, and broke out `needs_splitting` (among other bits) as an additional helper.  This limits the clone antics to a few IDs and things in the `needs_splitting` function.  (It does a bit clone more stuff than your version but honestly with all the nonsense we get up to during load I am not going to shed any tears over this.)

I reckon this makes the main flow way clearer, although if the borrow checker were not intent on THWARTING ME AT EVERY TURN I could have made it even less ghastly...

itowlson:
Bumped to the Spin 4.0 WASI version.  Also bumped the sample to the SDK `factor` branch for now.  It appears to work!  (with the usual caveats)

itowlson:
Rebased.  Not sure what is happening with that CI failure but it seems unrelated.

@fibonacci1729 Can we figure out remaining work required on this?  I know there is some stuff around permissions but not sure if there's anything else.

itowlson:
A maybe-silly-maybe-serious challenge with middleware permissions is:

* We are implementing middleware declarations via the `trigger.components` collection.
* `trigger.components` declares that a value in the map is a `OneOrManyComponentSpec` (i.e. `ComponentSpec` or `Vec<ComponentSpec>`)
* `ComponentSpec` doesn't provide a way to declare permissions

If we modelled middleware as a the same way as dependencies then this wouldn't be an issue, but `trigger.components` has pre-existing schema (even though that syntax was never used).  So we have choices (yay):

* retroactively change the schema of `trigger.components` to be dependency-like - technically a breaking change but the field is undocumented, no trigger used it, and it didn't work correctly so eh
  * although some dependency options don't make sense e.g. version
  * we don't know future uses for the `components` field so we don't want to force it to be dep-like just because that works for this one use case
* retroactively declare the schema of `trigger.components` to be 'components-plus' i.e. `ComponentSpec` augmented with dep-like permissions fields
  * some component fields make even less sense e.g. KV stores
* retroactively declare the schema of `trigger.components` to be "any combination of the above" (i.e. dep schema, component schema, or an array containing a mix of both)
  * since a dep can point to a path or component I'm not sure this gains us anything
* declare a different well-known map under `trigger`, which does _not_ have existing component schema and we can make it what we will
  * it needs to be well-known so that normalisation knows how to process it and OCI knows to compose it
  * it's still not clear what "what we will" would look like given we don't know other uses for the thing
  * I'm having the devil's own time coming up with a good name (as fans of the code in this PR will know)

My instinct based on this use case is "change `trigger.components` to be some variant of the deps schema" but I'm conscious this is a dramatic change of plan from what the shadowy architects of that field originally envisaged, and may therefore not reflect future needs.

itowlson:
Oh well that's weird.  That `bad_build_test` failure is repeatable.  But it occurs only in MacOS CI.  And I've no real idea how this PR would cause it.

fibonacci1729:
@itowlson My instincts align with yours. A strawman proposal:

## Example

```toml
[[trigger.http]]
route = "/..."
component = "main-handler"
components.middleware = [
  # 1. bare string: reference + deny-all (unchanged shorthand)
  "logging-mw",
  # 2. reference + inheritance policy (NEW table form)
  { component = "auth-mw", inherit_configuration = ["allowed_outbound_hosts"] },
  { component = "cache-mw", inherit_configuration = true },
]
```

## Schema

```rust
/// Component reference, reference-with-policy, or inline definition.
#[derive(Clone, Debug, Serialize, Deserialize, JsonSchema)]
#[serde(deny_unknown_fields, untagged, try_from = "toml::Value")]
#[schemars(schema_with = "json_schema::id_or_component")]
pub enum ComponentSpec {
    /// `"component-id"` — reference with deny-all (unchanged).
    Reference(KebabId),
    /// `{ component = "component-id", inherit_configuration = ... }` (NEW).
    ReferenceWithPolicy {
        /// The ID of the referenced component.
        component: KebabId,
        /// Which of the parent component's configurations this reference may
        /// inherit. Omitted/`false` = deny-all; `true` = inherit all;
        /// `["key", ...]` = inherit the named subset.
        inherit_configuration: Option<InheritConfiguration>,
    },
    /// `{ source = ..., ... }` — inline definition (unchanged).
    Inline(Box<Component>),
}
```

## Deserializion

```rust
impl TryFrom<toml::Value> for ComponentSpec {
    type Error = toml::de::Error;

    fn try_from(value: toml::Value) -> Result<Self, Self::Error> {
        use serde::de::Error as _;

        if value.is_str() {
            return Ok(ComponentSpec::Reference(KebabId::deserialize(value)?));
        }

        let Some(table) = value.as_table() else {
            // Not a string and not a table: let `Component` produce the error.
            return Ok(ComponentSpec::Inline(Box::new(Component::deserialize(value)?)));
        };

        let has_component = table.contains_key("component");
        let has_source = table.contains_key("source");

        match (has_component, has_source) {
            // Ambiguous: surface a precise error rather than a confusing
            // "unknown field `component`" from the inline `Component` path.
            (true, true) => Err(toml::de::Error::custom(
                "component spec cannot set both `component` (reference) and `source` (inline)",
            )),
            // Reference-with-policy form.
            (true, false) => {
                ReferenceWithPolicy::deserialize(value).map(|ReferenceWithPolicy { component, inherit_configuration }| {
                    ComponentSpec::ReferenceWithPolicy { component, inherit_configuration }
                })
            }
            // Inline component form (has `source`, or empty/other table that
            // `Component` will validate and reject as appropriate).
            (false, _) => Ok(ComponentSpec::Inline(Box::new(Component::deserialize(value)?))),
        }
    }
}
```

Disambiguation rule: a table with a `component` key (and no `source`) is a `ReferenceWithPolicy`; a table with a `source` key is an `Inline` component; a table with **both** is rejected with an explicit error. 

This mirrors how `ComponentDependency::AppComponent` is distinguished (`component = "..."`) from path/url/version dependency forms.

itowlson:
@lann I believe you were the shadowy architect of the `trigger.components` field. Does Brian's proposal align with your plans for the field?

lann:
I don't have any concerns from a compat perspective - as you say this was never used. It does seem a little confusing for the schema of `components.*` to be different than `component` - would this be too weird?

```toml
[[trigger.http]]
...
dependencies.middleware = [
  { component = "auth-mw", inherit_configuration = ["allowed_outbound_hosts"] },
]
```

itowlson:
@fibonacci1729 Is it intentional in your proposal that an inline component can't have permissions?  That would rule out e.g. `{ source = "third-party.wasm", inherit_configuration = ["allowed_outbound_hosts"] }`.

Really if we are reworking the field then the inline should not be allowed its own permission fields like a full component - it should inherit only.

itowlson:
> It does seem a little confusing for the schema of `components.*` to be different than `component` - would this be too weird?
> 
> ```toml
> [[trigger.http]]
> dependencies.middleware = [ ... ]
> ```

Yeah that is basically my option 4 (although I am queasy about repurposing the name "dependencies", but that is why we have bikesheds).

_(ETA: although for sure the schema is close to dependencies)_

fibonacci1729:
No that was not intentional and i think effectively rules out any approaches that don't introduce a new schema. Since to solve this, we would have to add `inherit_configuration` to `Component` which smells like schema pollution.

fibonacci1729:
+1 to `dependencies.middleware` I think that makes the most sense here.

Edit: Or introduce a `MiddlewareSpec`

lann:
I wouldn't be too upset with just calling the field `middleware`. Lots of people will associate it with just HTTP but there isn't really anything conceptually wrong with other trigger types having middleware. 🤷

itowlson:
Okay I have switched over to `dependencies.middleware` and implemented inherited permissions.  Local dev seems to work on my machine (I still need to test OCI), although there is some tidying to do, and the crate dependency graph for the HTTP trigger feels like it has gotten rather, er, compromised.  And of course we need about a gazillion unit and integration tests which I am enthusiastically looking forward to writing.

In the meantime, if anyone wants to kick tyres, https://github.com/itowlson/spin-middleware-terrifying-nonsense continues to show the way.

_(ETA: tested OCI and WOMM)_

itowlson:
Added some integration tests (at great personal peril, I might modestly note).  These are rather closely based on my existing test case, though, so I'd value guidance on whether they're enough or what other cases I should be covering.

crates/loader/tests/extra-components/inoffensive.toml
```diff
@@ -0,0 +1,29 @@
+spin_manifest_version = 2
+
+[application]
+name = "file-errors"
+
+[[trigger.http]]
+route = "/a"
+component = "a"
+components.middleware = ["m1", "m2"]
```
fibonacci1729: Lots of these test files in this directory still use `components.middleware`

itowlson: Ugh, well that tells me my tests aren't doing their job.  Thanks for the spot!

_(edit: although this is a 'nothing weird happens' test so not sure I could have made it fail)_

crates/loader/src/local/trigger_components.rs
```diff
@@ -0,0 +1,152 @@
...
+        .collect();
+    cid_to_triggers
+        .into_iter()
+        .filter(|(_, triggers)| triggers.len() > 1 && triggers.iter().any(has_trigger_deps))
+        .collect::<HashMap<_, _>>()
+}
+
+fn move_deps_from_triggers_to_components(locked: &mut LockedApp) {
+    for trigger in &mut locked.triggers {
+        if !trigger.trigger_dependencies.is_empty()
```
fibonacci1729: Should we use `has_trigger_deps` here? I think we will fall through here when the map has one empty-vec entry.

crates/loader/tests/extra-components/inoffensive.toml
itowlson: Brian, can you elaborate on "lots"?  I couldn't find any others and am concerned I am missing something - sorry.

crates/manifest/src/schema/v2.rs
```diff
@@ -844,10 +985,18 @@ mod one_or_many {
...
+            arr.iter()
+                .map(|v| T::deserialize(v.clone()))
+                .collect::<Result<Vec<_>, _>>()
+                .map_err(serde::de::Error::custom)
         } else {
-            Vec::deserialize(value).map_err(serde::de::Error::custom)
+            T::deserialize(value)
+                .map(|v| vec![v])
+                .map_err(serde::de::Error::custom)
```
fibonacci1729, line 999: Would you mind adding a small unit test for the one-or-may change?

crates/loader/tests/extra-components/inoffensive.toml
fibonacci1729: Oh no you're right -- it's just this one. My gh view was wonky and looked like multiple.

crates/trigger/src/lib.rs
```diff
@@ -51,6 +51,12 @@ pub trait Trigger<F: RuntimeFactors>: Sized + Send {
         Ok(())
     }
 
+    /// An object which composes extras onto the primary component.
+    /// TODO: the combination of functions and objects and traits is a bit funny and we may/should be able to streamline it.
+    fn trigger_dependencies_composer() -> impl spin_factors_executor::TriggerDependenciesComposer {
+        // the do-nothing unit composer
+    }
```
fibonacci1729, line 58: 🤔 does this compile?

crates/trigger/src/loader.rs
```diff
@@ -90,17 +122,60 @@ impl<T: RuntimeFactors, U> spin_factors_executor::ComponentLoader<T, U> for Comp
...
-            })?;
+        let composed = self.load_composed(component, complicator).await?;
 
         let component = spin_core::Component::new(engine, composed)
             .with_context(|| format!("failed to compile component from {}", quoted_path(&path)))?;
         Ok(component)
     }
 }
+
+pub(crate) async fn load_complications(
```
fibonacci1729: Would still prefer `load_compose_extras` over `complications` as fun as that is.

fibonacci1729:
This is looking great! I'm gonna kick the tires and dig in a little deeper to understand everything but i left some obvious first comments.

crates/trigger/src/lib.rs
itowlson: Yes (well, it does both locally and in CI). TriggerDependenciesComposer is implemented for the unit type.

fibonacci1729: Oh TIL I guess? I would have thought you needed to explicitly return the unit type here.

crates/trigger/src/loader.rs
itowlson: I've been using `trigger_dependencies` rather than `extras`, now we have a name for it.  Thanks for the catch.

crates/trigger/src/lib.rs
itowlson: If memory serves, clippy gets mad if you have the temerity to try.

itowlson:
Latest force-push is just squashing commits.

fibonacci1729 (approved):

PR #3385, opened 2026-02-09
Re-export the correct export in a dependency

fibonacci1729:
This fixes an issue where, when using component dependencies, the name of the item being exported from a dependency was incorrectly assumed to always be the name of the import.


fibonacci1729:
I definitely agree. At the very least I filed https://github.com/spinframework/spin/issues/3386 to improve the logging here in the interim and https://github.com/spinframework/spin/issues/3387

PR #3400, opened 2026-02-16
Add SIP for environment variable isolation

tschneidereit:
This SIP proposes a new component isolation mechanism for environment variables, allowing components to have their own isolated view of the environment. This is achieved by synthesizing an isolator component applying filtering to expose only environment variables with a specified prefix to each component in a composition.

[Rendered](https://github.com/tschneidereit/spin/blob/4a9842bb2baab24af481d7d028b9b79fef1f127b/docs/content/sips/022-environment-variable-isolation.md)

### DIFF
--- a/docs/content/sips/022-environment-variable-isolation.md
+++ b/docs/content/sips/022-environment-variable-isolation.md
@@ -1,0 +1,233 @@
+title = "SIP 022 - Environment Variable Isolation for Composed Components"
+template = "main"
+date = "2026-02-15T00:00:00Z"
+---
+
+Summary: Automatically isolate WASI environment variables within a component composition so that each component in a dependency graph sees only its own declared variables, despite sharing a single host process.
+
+Owner(s): till@tillschneidereit.net
+
+Created: Feb 15, 2026
+
+## Background
+
+Spin applications can compose multiple WebAssembly components together using component dependencies (see [SIP 020](../sips/020-component-dependencies.md)). When a parent component declares a dependency, Spin statically composes the two components into a single Wasm component at build time or at load time.
+
+A consequence of this composition is that all components in the resulting graph share the same host `wasi:cli/environment` import. Any environment variables set on the parent component are visible to every dependency, and dependencies cannot have their own distinct environment configurations. This is a problem for several reasons:
+
+1. **Leaking configuration.** A parent component's secrets or internal configuration (e.g. API keys set via environment variables) are visible to dependencies that should not see them.
+
+2. **No per-dependency configuration.** Reusable library components that are parameterised by environment variables cannot be given different values when the same library is used by different parent components, or when a dependency needs different settings than its parent.
+
+3. **Name collisions.** Two components that both read `DATABASE_URL` will see the same value, with no way to differentiate.
+
+## Proposal
+
+To address these issues, this SIP proposes to automatically isolate the environment variables of a component and all its dependencies from each other.
+
+At a high level, the isolation works by adding prefixes based on the components' name / dependency IDs when loading an application manifest, and then filtering by those prefixes in a synthesized component that interposes imports of `wasi:cli/environment`, removing the prefix of allowed variables before forwarding them.
+
+This process introduces a single very small (about 1.5KB to 2KB) core WebAssembly module to the composed graph, containing a 1-page (64KB) linear memory. Actual resource allocation should normally be a small fraction of those 64KB, unless the component graph has large amounts of data passed in as env vars.
+
+### Manifest changes
+
+A new `environment` field is added to each dependency specification within the `[component.<id>.dependencies]` section. This field is an inline table of string key-value pairs, identical in syntax to the existing component-level `environment` field:
+
+```toml
+spin_manifest_version = 2
+
+[application]
+name = "my-app"
+version = "1.0.0"
+
+[[trigger.http]]
+route = "/..."
+component = "main"
+
+[component.main]
+source = "main.wasm"
+environment = { GREETING = "hello from main", MAIN_ONLY = "secret" }
+
+[component.main.dependencies."acme:lib/helper"]
+component = "helper"
+environment = { GREETING = "hello from helper", HELPER_SETTING = "42" }
+
+[component.helper]
+source = "helper.wasm"
+```
+
+Additionally, components are now allowed to specify their own `environment` table and still be used as dependencies, instead of being rejected. When a dependency references another component in the same application (the `component = "..."` form), the referenced component may itself declare an `environment`. In that case, the two sets of variables are merged: the referenced component's `environment` provides the base values, and the dependency specification's `environment` provides overrides and additions. This merge happens during manifest normalisation, before any composition takes place.
+
+For example, given:
+
+```toml
+[component.helper]
+source = "helper.wasm"
+environment = { GREETING = "default greeting", LOG_LEVEL = "info" }
+
+[component.main.dependencies."acme:lib/helper"]
+component = "helper"
+environment = { GREETING = "overridden greeting" }
+```
+
+The effective environment for the `helper` dependency is `{ GREETING = "overridden greeting", LOG_LEVEL = "info" }`.
+
+The `environment` field is available on all dependency forms (`component`, `path`, `url`, and registry version specs), consistent with the existing `export` field.
+
+### How isolation works
+
+Spin automatically activates environment variable isolation for compositions with at least one dependency. The mechanism works in two coordinated stages:
+
+#### Stage 1: Prefixed environment variables (manifest load time)
+
+At manifest load time, Spin's WASI factor prepends a deterministic prefix to every environment variable key before placing it in the WASI context. The prefix is derived from the component's ID (for the main component) or the dependency name (for dependencies):
+
+- Convert the last segment of the name (after the final `/` if present) to uppercase.
+- Replace non-alphanumeric characters with underscores.
+- Append a trailing underscore.
+
+Examples:
+| Component / Dependency Name | Prefix |
+|---|---|
+| `main` | `MAIN_` |
+| `my-service` | `MY_SERVICE_` |
+| `acme:lib/helper` | `HELPER_` |
+| `hello:components/dependable` | `DEPENDABLE_` |
+
+The resulting flat list of environment variables is used in the locked application, so except for stage 2 below, all further parts of the processing pipeline aren't changed in any way, which results in the WASI environment being populated with the prefixed variables.
+
+Given the manifest example above, the actual WASI environment would contain:
+
+```
+MAIN_GREETING=hello from main
+MAIN_MAIN_ONLY=secret
+HELPER_GREETING=hello from helper
+HELPER_HELPER_SETTING=42
+```
+
+**NOTE**: Dynamically provided environment variables, e.g. those added using `spin up -e SOME=var` aren't automatically prefixed, and as a result not visible to any component, unless they already have the right prefix. E.g. to expose `SOME` to the `main` component, it has to be changed to `MAIN_SOME`.
+
+#### Stage 2: Isolator component (composition time)
+
+At composition time, Spin generates and wires in a synthetic WebAssembly component — the **env-isolator** — that interposes on each component's `wasi:cli/environment` import. For each component in the composition, the isolator:
+
+1. Intercepts calls to `get-environment`.
+2. Filters the flat environment to only those keys matching the component's prefix.
+3. Strips the prefix before returning the key-value pairs.
+
+The result is that each component sees only its own environment variables, with unprefixed keys, as if those were the only variables that existed.
+
+The isolator also passes through `get-arguments` and `initial-cwd` without modification, since these are process-level values that do not benefit from per-component filtering.
+
+Components that do not import `wasi:cli/environment` are skipped; no isolator is wired up for them.
+
+### Activation rule
+
+Environment variable isolation is activated **automatically** for components with at least one dependency. When active, **all** components in the composition are isolated — including the main component. A component with no declared `environment` will see an empty environment when isolation is active.
+
+## Implementation
+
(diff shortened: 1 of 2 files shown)


tschneidereit:
Oh, one very important thing I forgot to mention: I'm not entirely sure myself whether this should be accepted as-is. Perhaps the better thing to do would be to address component isolation more comprehensively: instead of doing env vars now, and something else later, we'd make one big change applying isolation to everything where it makes sense. That'd include things like Spin's own `variables` interface, outbound allow lists for sockets & http, kv store buckets, db interfaces, etc.

The downside is that that is quite substantially more work, and requires a lot more design decisions than are needed just for env vars. I think a lot of the isolation approach here would still apply, and many considerations would stay the same. But certainly not all would, and we'd all but certainly need to make pretty intrusive changes to interfaces implemented in factors, as opposed to just having to change manifest loading and component composition.

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+- Convert the last segment of the name (after the final `/` if present) to uppercase.
+- Replace non-alphanumeric characters with underscores.
+- Append a trailing underscore.
+
+Examples:
+| Component / Dependency Name | Prefix |
+|---|---|
+| `main` | `MAIN_` |
+| `my-service` | `MY_SERVICE_` |
+| `acme:lib/helper` | `HELPER_` |
```
itowlson, line 94: `HELPER_` or `ACME_LIB_HELPER_`?

Also I believe we allow "import all" deps so the dependency could be `acme:lib` (check with Brian though).

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+Given the manifest example above, the actual WASI environment would contain:
+
+```
+MAIN_GREETING=hello from main
+MAIN_MAIN_ONLY=secret
+HELPER_GREETING=hello from helper
+HELPER_HELPER_SETTING=42
+```
+
+**NOTE**: Dynamically provided environment variables, e.g. those added using `spin up -e SOME=var` aren't automatically prefixed, and as a result not visible to any component, unless they already have the right prefix. E.g. to expose `SOME` to the `main` component, it has to be changed to `MAIN_SOME`.
```
itowlson, line 108: This seems like a breaking change.

The current behaviour is that `spin up -e SOME=var` makes `SOME` appear in _all_ components of the application.  The proposed new behaviour would require separate `THING_SOME` for _each_ component.

We should consider instead having Spin auto-munge `-e` variables so that they are applied to all components.

(If we _do_ intentionally want to make `-e` selective - which should probably be a separate discussion - I'd look to decouple the external representation from the internal one.  At the moment, we can change the internal representation as we see fit: I'd be very cautious about making it contractual.)

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+
+At composition time, Spin generates and wires in a synthetic WebAssembly component — the **env-isolator** — that interposes on each component's `wasi:cli/environment` import. For each component in the composition, the isolator:
+
+1. Intercepts calls to `get-environment`.
+2. Filters the flat environment to only those keys matching the component's prefix.
+3. Strips the prefix before returning the key-value pairs.
+
+The result is that each component sees only its own environment variables, with unprefixed keys, as if those were the only variables that existed.
+
+The isolator also passes through `get-arguments` and `initial-cwd` without modification, since these are process-level values that do not benefit from per-component filtering.
```
itowlson, line 120: My first reaction is "why am I letting Semi-Trusted Steve scope out **MY** argument list" but I may be misunderstanding - is the idea that I can still _block_ access via the normal "no inherit config" mechanism, but if I _grant_ access then there is no point customising?  If so then no worries.

docs/content/sips/022-environment-variable-isolation.md
itowlson, line 108: As the isolator is only in play for components with at least one dependency, this suggests that `SOME=var` will work for components _without_ dependencies?  I'm not enthusiastic about having to guess which components have dependencies and customise syntax for them alone.

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+
+Components that do not import `wasi:cli/environment` are skipped; no isolator is wired up for them.
+
+### Activation rule
+
+Environment variable isolation is activated **automatically** for components with at least one dependency. When active, **all** components in the composition are isolated — including the main component. A component with no declared `environment` will see an empty environment when isolation is active.
```
itowlson, line 126: > A component with no declared `environment` will see an empty environment when isolation is active.

This is the same as the current behaviour right?  I am not quite sure what you are seeking to highlight here.

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+
+- **`isolator::generate_isolator(targets, wasi_env_version) -> Vec<u8>`** — Generates the isolator component. Each `IsolationTarget` specifies a name and prefix. The generated component imports `wasi:cli/environment@{version}` and exports flat functions (`environment-{name}-get-environment`, `environment-{name}-get-arguments`, `environment-{name}-get-cwd`) for each target.
+
+- **`wrapper::build_env_wrapper_component(name, wasi_env_version) -> Vec<u8>`** — Generates a small adapter component per target that imports the flat functions from the isolator and re-exports them bundled as a `wasi:cli/environment` instance, which can be wired into the target component's import. Note that this wrapper component only contains component model instructions, no core WebAssembly bytecode, and is fully erased at link time without leaving overhead.
+
+- **`core_module`** — Generates the core Wasm module that performs the actual byte-level prefix matching and stripping in linear memory, operating on the canonical ABI representation of `list<tuple<string, string>>`.
```
itowlson, line 142: I gazed at this for a while trying to get my head around the relationship between `isolator` and `wrapper` and `core_module` and I think I understand why we need `wrapper` but I'm not getting the distinction between `core_module` and `isolator`.  Perhaps expand on this or draw a diagram, or possibly drop this section as more detail than we need at this level?

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+
+The implementation spans five crates:
+
+### `spin-env-isolator` (new crate)
+
+A new `crates/env-isolator/` crate containing the core Wasm generation logic. It has no dependency on Spin's application model and can be tested independently. It provides:
+
+- **`compute_prefix(id: &str) -> String`** — The shared prefix derivation function used by both the composition pipeline and the runtime factor, guaranteeing agreement.
+
+- **`isolator::generate_isolator(targets, wasi_env_version) -> Vec<u8>`** — Generates the isolator component. Each `IsolationTarget` specifies a name and prefix. The generated component imports `wasi:cli/environment@{version}` and exports flat functions (`environment-{name}-get-environment`, `environment-{name}-get-arguments`, `environment-{name}-get-cwd`) for each target.
```
itowlson, line 138: How do we manage evolving this over time?  That is, if a new WASI preview comes out, how are we reminded to update this?

How does it interact with version compatibility stuff?  That is, do we need to recognise minor versions or only major versions? (or whatever WIT calls them, you know, the things where you can use 4.0.0 and 4.1.0 interchangeably apart from the `since` stuff)

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Design decisions
+
+### Automatic activation
+
+Environment isolation is activated implicitly when any component in a composition imports `wasi:cli/environment`, without an opt-out. This choice follows from the goal to make safe behavior the default or even enforced, but could be weakened by providing an opt-out.
+
+### Prefix is not user-configurable
+
+The prefix is derived deterministically from the component or dependency name. Exposing the prefix to users would add configuration surface for an implementation detail that has no user-visible effect (components never see the prefix). Keeping it internal simplifies the UX and prevents misconfiguration.
```
itowlson, line 215: I agree it should be user-configurable but the stuff about `spin up -e` gives me the impression that it _does_ have a user-visible effect.  (But as mentioned above I think we _should_ avoid it having a user-visible effect.)

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+dependable's env vars: GREETING='hello from dependable', DEPENDABLE_ONLY='only visible to dependable'
+```
+
+Each component sees only its own variables. The shared key `GREETING` has a different value for each, and `MAIN_ONLY` / `DEPENDABLE_ONLY` are invisible to the other component.
+
+## Design decisions
+
+### Automatic activation
+
+Environment isolation is activated implicitly when any component in a composition imports `wasi:cli/environment`, without an opt-out. This choice follows from the goal to make safe behavior the default or even enforced, but could be weakened by providing an opt-out.
```
itowlson, line 211: Yeah I think care is needed here.  Currently `dependencies_inherit_configuration = true` gives a dep access to _all_ the composition's EVs.  It's not clear to me what the semantics are after this change.

The combinations (and my guess at behaviour) are:

1.  "inherit = false" and no `env` section(s) - no EVs for dep (isolator in play but blocking everything)
2. "inherit = false" and yes `env` section(s) - only specified EVs, isolator in play
3. "inherit = true" and no `env` secition - ??? all inherited EVs as current ???
4. "inherit = true" and yes `env` secition - ??? all inherited EVs _and_ all dep EVs??? ???

But I am not quite sure what you envision for scenarios 3 and 4.  If we change scenario 3 then it's breaking (which I'm not opposed to, we just need to plan and communicate).

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+
+The prefix is derived deterministically from the component or dependency name. Exposing the prefix to users would add configuration surface for an implementation detail that has no user-visible effect (components never see the prefix). Keeping it internal simplifies the UX and prevents misconfiguration.
+
+### Only environment variables are isolated
+
+This proposal isolates only `wasi:cli/environment`. Other WASI capabilities (filesystem, networking, etc.) continue to be managed by the existing `dependencies_inherit_configuration` / deny-all adapter mechanism from SIP 020. Isolating those would require different interposition strategies and is out of scope.
+
+### Backward compatibility
+
+There are two compatibility-breaking aspects to this change:
```
itowlson, line 223: Oops, should have read ahead before leaving those comments!

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+1. The main component's `environment` isn't visible to dependencies anymore, so existing applications that assume it is won't work as expected.
+2. Environment variables provided dynamically, e.g. using the `spin up -e` need to be prefixed to be exposed to specific components.
```
itowlson, line 225: As noted above I do not love this one especially if the way we do it couples the internal representation to the UI.

docs/content/sips/022-environment-variable-isolation.md
```diff
@@ -0,0 +1,233 @@
...
+
+At the purely syntactical manifest level the change is backwards-compatible: the `environment` table in a dependency spec is optional, as is the `env` table in the lock file.
+
+## Future possibilities
+
+- **Template expressions in environment.** The `environment` field currently accepts only static strings. A future enhancement could allow Spin template expressions (e.g. `{{ my_variable }}`), leveraging the existing variables/expressions system for dynamic configuration.
```
itowlson, line 231: This feature is completely independent of isolation though isn't it?

itowlson (commented):
Looks good - thanks for noodling through all the details on this!  I had a few thoughts, mostly about UI, compatibility (which turned out to be covered later!), and a couple of things that I was unclear on.  But generally yep this looks splendid!

docs/content/sips/022-environment-variable-isolation.md
itowlson: Another potential woe:

Imagine _two_ components, `alice` and `bob`.  Both import `carol`.

With the current scheme, the syntax for setting environment variables on `carol` would be `spin up -e CAROL_WHATEVER=thingy`.

But there are two `carol`s in play.  We don't want to couple them together.  So we need further disambiguation at the CLI. (Of course this is not an issue for the internal representation, as it's scoped to a particular composition).

docs/content/sips/022-environment-variable-isolation.md
fibonacci1729: ^ is correct: `acme:lib` is an acceptable form

kate-goldenring:
> Oh, one very important thing I forgot to mention: I'm not entirely sure myself whether this should be accepted as-is. Perhaps the better thing to do would be to address component isolation more comprehensively: instead of doing env vars now, and something else later, we'd make one big change applying isolation to everything where it makes sense. That'd include things like Spin's own `variables` interface, outbound allow lists for sockets & http, kv store buckets, db interfaces, etc.
> 
> The downside is that that is quite substantially more work, and requires a lot more design decisions than are needed just for env vars. I think a lot of the isolation approach here would still apply, and many considerations would stay the same. But certainly not all would, and we'd all but certainly need to make pretty intrusive changes to interfaces implemented in factors, as opposed to just having to change manifest loading and component composition.

@tschneidereit , to your point here https://github.com/spinframework/spin/pull/3400#issuecomment-3909077630, while supporting all component resources on a first implementation may be hard, could we aim to support both environment variable and spin variables isolation in the first implementation? That will catch pains early and help with terminology around the released feature (explaining the diff between envs and vars from the start)

itowlson:
@tschneidereit Should we close this and start a new one for the multiple imports approach, or is this one still relevant?

PR #3401, opened 2026-02-16
feat: Implement environment variable isolation for components

tschneidereit:
This implements the SIP introduced in #3400.

Note: I made extensive use of GitHub Copilot in implementing this, but made sure that I fully understand the entire implementation. Additionally, I ensured that everything is thoroughly documented and tested, since some of the implementation is very subtle and low-level.

In particular, the isolation itself is implemented by synthesizing a new component containing two, also synthesized, core modules, plus a wrapper component for each target component. This code makes substantial use of the `wasm-encoder` crate to generate the necessary WebAssembly components and modules, and the `wasmparser` crate to validate the generated components.

By comparison, the integration into Spin is delightfully simple: manifest loading and composition are the only two pieces that needed updating, with everything else not needing to have any awareness of the isolation at all.


crates/compose/src/lib.rs
```diff
@@ -213,6 +213,12 @@ impl<'a, L: ComponentSourceLoader> Composer<'a, L> {
 
         let prepared = self.prepare_dependencies(world_id, component).await?;
 
+        // Apply env isolation when there is at least one dependency.
+        if self.should_apply_env_isolation(&prepared) {
+            self.apply_env_isolation(component.id(), instantiation_id, world_id, &prepared)
```
itowlson, line 218: I am not sure if I am misunderstanding this or if I have missed the relevant change elsewhere, but this seems to happen _after_ the deny-all adapter has been applied (in `prepare_dependencies`).  So my understanding would be that when the dep called `get-environment`, that would be denied by the adapter and never reach the isolator.  I know you've tested this and it works, so this is probably more for my education, but what am I missing?  (Sorry if this is a stupid question.)

crates/compose/src/lib.rs
```diff
@@ -466,6 +472,115 @@ impl<'a, L: ComponentSourceLoader> Composer<'a, L> {
...
+
+    /// Returns true when environment isolation should be applied.
+    ///
+    /// Environment isolation is applied if there is at least one dependency.
+    // Note: In principle this is overly conservative since we could check
+    // whether the main component or any dependency imports `wasi:cli/environment`,
+    // but in practice almost all components will import `wasi:cli/environment`,
+    // so we go with this simpler heuristic for now.
+    fn should_apply_env_isolation(&self, prepared: &IndexMap<String, DependencyInfo>) -> bool {
+        prepared.values().len() > 0
```
itowlson, line 492: Any particular reason for preferring this over `!prepared.is_empty()`?  Is the call to `values()` significant?  If so maybe comment on why?

crates/compose/src/lib.rs
```diff
@@ -466,6 +472,115 @@ impl<'a, L: ComponentSourceLoader> Composer<'a, L> {
...
+    }
+
+    /// Apply environment variable isolation.
+    ///
+    /// Generates an isolator component shared across all targets, plus a small
```
itowlson, line 497: Does "target" mean "dependency" here?

(Edit: I think is means "component" in the sense of any component, main or dependency.)

crates/compose/src/lib.rs
```diff
@@ -466,6 +472,115 @@ impl<'a, L: ComponentSourceLoader> Composer<'a, L> {
...
+        prepared: &IndexMap<String, DependencyInfo>,
+    ) -> anyhow::Result<()> {
+        use spin_env_isolator::isolator::{generate_isolator, IsolationTarget};
+        use spin_env_isolator::wrapper::build_env_wrapper_component;
+
+        // Collect isolation targets (components that import wasi:cli/environment).
+        // Each target tracks the specific wasi:cli/environment version it imports,
+        // since different components may import different versions.
+        let mut targets = Vec::new();
+        let mut target_instances: Vec<(String, NodeId, String)> = Vec::new();
```
itowlson, line 514: This might merit being a struct with field names, in particular because I for one will mix up the order of the two strings!

crates/compose/src/lib.rs
```diff
@@ -466,6 +472,115 @@ impl<'a, L: ComponentSourceLoader> Composer<'a, L> {
...
+        // For each target, create a wrapper and wire isolator → wrapper → target.
+        // Each wrapper uses the target's own wasi:cli/environment version.
+        for (target_name, target_instance, target_wasi_env_version) in &target_instances {
+            let wrapper_bytes = build_env_wrapper_component(target_name, target_wasi_env_version)?;
+            let wrapper_pkg_name = format!("env-wrapper-{target_name}");
+            let (_, wrapper_instance) =
+                self.register_package(&wrapper_pkg_name, None, wrapper_bytes)?;
+
+            // Wire isolator → wrapper (filtered get-environment)
+            let export_name = format!("environment-{target_name}-get-environment");
```
itowlson, line 566: Two sources of truth for this format string?  (cf isolator line 180)

crates/compose/src/lib.rs
```diff
@@ -482,6 +597,18 @@ struct DependencyInfo {
...
 
+/// Returns true if `candidate` is a higher semver version than `current`
+/// (or if `current` is `None`).
+fn is_higher_version(candidate: &str, current: Option<&str>) -> bool {
+    let Some(current) = current else {
+        return true;
+    };
+    match (Version::parse(candidate), Version::parse(current)) {
+        (Ok(c), Ok(cur)) => c > cur,
+        _ => candidate > current,
```
itowlson, line 608: This is a lexical comparison?  I'm a bit doubtful about that.  I'd feel safer with something like "bad candidate => false, bad current => true" maybe?

crates/env-isolator/src/isolator.rs
```diff
@@ -0,0 +1,220 @@
...
+use wasm_encoder::{
+    Alias, CanonicalOption, ComponentBuilder, ComponentExportKind, ComponentOuterAliasKind,
+    ComponentTypeRef, ComponentValType, ExportKind, InstanceType, ModuleArg, PrimitiveValType,
+};
+
+use crate::core_module::{build_memory_module, generate_env_filter_module};
+
+/// Information about a component that needs environment isolation.
+pub struct IsolationTarget {
+    /// Logical name for this component (e.g., "main", "dep").
```
itowlson, line 18: What is "logical" in this context?  This isn't the actual name?

crates/env-isolator/src/isolator.rs
```diff
@@ -0,0 +1,220 @@
...
+        kind: ComponentOuterAliasKind::Type,
+        count: 1,
+        index: get_args_func_type,
+    });
+    env_instance_type.alias(Alias::Outer {
+        kind: ComponentOuterAliasKind::Type,
+        count: 1,
+        index: get_cwd_func_type,
+    });
+    env_instance_type.export("get-environment", ComponentTypeRef::Func(0));
```
itowlson, line 98: This is stretching my ability to follow so sorry if this is already happening but: I would feel happier if there were a check that the chosen version of `wasi:cli/environment` is compatible with these functions and an _explicit, specific error_ if not.  I realise it is vanishingly unlikely for `wasi:cli/environment` to change but if it ever does then I would like not to have to diagnose a weird composition error about a thing that I never nominated for composition - I would like to know "the isolator doesn't support your version of wasi:cli/environment, call Brian and tell him to get with the programme."

Bonus points for a test that runs this exercise against all the versions of `wasi:cli` in the Spin world.

crates/env-isolator/src/isolator.rs
```diff
@@ -0,0 +1,220 @@
...
+            &format!("environment-{}-get-environment", target.name),
+            ComponentExportKind::Func,
+            lifted_get_env,
+            None,
+        );
+    }
+
+    let bytes = builder.finish();
+
+    #[cfg(debug_assertions)]
```
itowlson, line 189: Am I right in thinking that in release builds an invalid isolator will be returned and allowed to participate in composition?  Yes, I know it can't happen, but again, if it _did_, would the error be clear and crisp?

crates/env-isolator/src/core_module.rs
```diff
@@ -0,0 +1,1091 @@
...
+        let get_env_0 = instance
+            .get_typed_func::<(), i32>(&mut store, "get-environment-0")
+            .unwrap();
+        let result_ptr = get_env_0.call(&mut store, ()).unwrap();
+        let vars = read_env_result(&memory, &store, result_ptr);
+
+        assert_eq!(vars.len(), 1);
+        assert_eq!(vars[0], ("X".to_string(), "match".to_string()));
+    }
+}
```
itowlson, line 1091: I have no idea how to review this code, so I'll leave that to others, but my confidence in its as a black box would be increased by fuzzing or property-based testing or something.

I worry that we are going to struggle to maintain this if it ever goes wrong or needs changes.  I guess the project has at least 3 collaborators at the moment who would be able to do it, so the bus factor is not an immediate concern?

crates/env-isolator/src/isolator.rs
```diff
@@ -0,0 +1,220 @@
...
+    // --- Alias get-environment from imported instance ---
+    let host_get_env = builder.alias_export(
+        host_env_instance,
+        "get-environment",
+        ComponentExportKind::Func,
+    );
+
+    // --- Embed core modules ---
+    let total_prefix_bytes: usize = targets.iter().map(|t| t.prefix.len()).sum();
+    let aux_module = build_memory_module(total_prefix_bytes as u32);
```
itowlson, line 119: Is it possible to clarify the role of the "aux" module?

crates/env-isolator/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_compute_prefix() {
+        assert_eq!(compute_prefix("my-app"), "MY_APP_");
+        assert_eq!(compute_prefix("worker"), "WORKER_");
+        assert_eq!(compute_prefix("main"), "MAIN_");
+        assert_eq!(compute_prefix("foo:bar/baz"), "BAZ_");
```
itowlson, line 50: As noted on the SIP, I am not sure we can rely on interface names alone.  If I depend on `sandro:botticelli/draw` and `john:wayne/draw` I don't necessarily want them to share EVs.

crates/env-isolator/src/wrapper.rs
```diff
@@ -0,0 +1,154 @@
...
+//! away by the component linker.
+
+use anyhow::Result;
+use wasm_encoder::{
+    Alias, Component, ComponentAliasSection, ComponentExportKind, ComponentExportSection,
+    ComponentImportSection, ComponentInstanceSection, ComponentOuterAliasKind, ComponentTypeRef,
+    ComponentTypeSection, ComponentValType, InstanceType, PrimitiveValType,
+};
+
+/// Build a wrapper component that imports a filtered `get-environment` from
```
itowlson, line 22: I am confused at this point.  I thought the wrapper was importing the `environment-whatever-get-environment` thingy and mapping that to its exported `get-environment`.  Are we saying that the isolator exports a `get-environment` which is already filtered?

crates/loader/src/local.rs
```diff
@@ -239,6 +240,45 @@ impl LocalLoader {
...
+    ) -> BTreeMap<String, String> {
+        if manifest_deps.inner.is_empty() {
+            // Don't prefix env vars if there are no dependencies.
+            return component_environment.into_iter().collect();
+        }
+
+        let mut env = BTreeMap::new();
+
+        // Prefix the component's own env vars.
+        let main_prefix = spin_env_isolator::compute_prefix(component_id);
```
itowlson, line 263: Oof, having the loader depend on the isolator feels kind of icky.  Could this be pushed down to something that thinks about isolation and composition?  I am not sure.  But we have kind tried to keep the loader's job to "prepare the lockfile and the assets and stuff."

For example, if we do it here, then the post-munge environment will (in a non-precompose situation) become part of the OCI manifest.  That makes it a contract with future versions of Spin.  Could we make it so that the (uncomposed) OCI manifest preserved the information with maximal fidelity, and flattened it to the munged representation only at the time of use?

(I admit this is mostly an emotional reaction to depending on something as terrifying as the isolator crate, but the OCI thing is a real concern.)

src/commands/up.rs
```diff
@@ -520,7 +520,10 @@ impl UpCommand {
     }
 
     fn update_locked_app(&self, locked_app: &mut LockedApp) {
-        // Apply --env to component environments
+        // Apply --env to component environments.
+        // Dynamic env vars are added as-is: if env isolation is active,
```
itowlson, line 524: I don't think  we've previously used the term "dynamic" for env vars.  Maybe "EVs passed on the command line"?

src/commands/up.rs
```diff
@@ -520,7 +520,10 @@ impl UpCommand {
...
+        // the env map is already prefixed from lock time, and the isolator
```
itowlson, line 525:
```suggestion
        // environment variable names are already prefixed in the lockfile, and the isolator
```

(I had to read the original a few times, sorry.)

src/commands/up.rs
```diff
@@ -520,7 +520,10 @@ impl UpCommand {
...
+        // will route vars by prefix. The user provides the prefix they want.
```
itowlson, line 526: > The user provides the prefix they want.

I am not sure what this is getting at.  The user has no control over the prefix.  They have to use the prescribed prefix for the component or dep that will be consuming the EV.

src/commands/up.rs
```diff
@@ -924,4 +927,78 @@ mod test {
         assert_eq!("-L", groups[2][0]);
         assert_eq!("/fie", groups[2][1]);
     }
+
+    #[test]
+    fn dynamic_env_vars_passed_through_unchanged() {
```
itowlson, line 932:
```suggestion
    fn command_line_env_vars_passed_through_unchanged() {
```

itowlson, line 932: I don't get the significance of "passed through unchanged."  Why would they be anything other than "unchanged"?  What are you testing here?

~Is this a loader test, perhaps, that the bit where it munges env names from the manifest skips names supplied on the CLI?~  (No, because the loader never sees the CLI vars.  Okay I think I see what this is testing now but I'm not sure I can think of a good name for it.)

src/commands/up.rs
```diff
@@ -924,4 +927,78 @@ mod test {
...
+                )]
+                .into_iter()
+                .collect(),
+                host_requirements: Default::default(),
+            }],
+        };
+
+        let cmd = UpCommand {
+            env: vec![
+                ("MAIN_FOO".into(), "bar".into()),
```
itowlson, line 979: As per comments on SIP, I find this very awkward.  I am concerned about the coupling of internal and external naming.  I am bothered that the operator needs to know the internal composition structure.  And I dislike that if you have a dep-less component that uses EVs, adding a dep to it is a breaking change to the application (because the existing EV names stop working).  I would really like us to explore alternatives here, please.

itowlson:
@tschneidereit Are you planning to pick this up as is, or should we close it in favour of the planned multiple-imports approach?

tschneidereit:
I'm planning to pick this up again, but indeed with an approach based on named imports. About 50% to 75% of the changes here should stay relevant, but I'll close this particular PR for now.

PR #3453, opened 2026-04-09
SIP-024: spin deps cli dx

fibonacci1729:
This SIP proposes a Dx for guiding a user into adding a dependency to a component in a Spin.toml. It builds on https://github.com/spinframework/spin/pull/3445 and the implementation https://github.com/spinframework/spin/pull/3450.

### DIFF
--- a/docs/content/sips/024-spin-deps-cli-dx.md
+++ b/docs/content/sips/024-spin-deps-cli-dx.md
@@ -1,0 +1,248 @@
+title = "SIP 024 - spin deps cli dx"
+template = "main"
+date = "2026-04-09T00:00:00Z"
+
+---
+
+Summary: A CLI command (`spin deps add`) for adding component dependencies to a Spin application, with interactive prompts for selecting components, exports, and capability inheritance.
+
+Owner(s): [brian.hardock@fermyon.com](mailto:brian.hardock@fermyon.com)
+
+Created: April 9, 2026
+
+# Background
+
+[SIP 020](docs/content/sips/020-component-dependencies.md) introduced the concept of component dependencies in Spin, allowing developers to compose components together by declaring dependencies in `spin.toml`. [SIP 023](docs/content/sips/023-granular-capability-inheritance.md) extended this with per-dependency, granular capability inheritance — replacing the all-or-nothing `dependencies_inherit_configuration` boolean with a flexible `inherit_configuration` field that accepts `true`, `false`, or a list of specific capability keys.
+
+However, authoring the dependency entries by hand requires understanding the TOML schema, knowing which exports a component offers, and correctly configuring capability inheritance — all of which are error-prone.
+
+`spin deps add` provides a guided CLI experience for adding a component dependency. It resolves the source, inspects the Wasm component's exports, and writes the correct entry into `spin.toml`, along with regenerating the `spin-dependencies.wit` file.
+
+# Proposal
+∏
+## Command Syntax
+
+```
+spin deps add <source> [options]
+```
+
+### Source Formats
+
+The `<source>` positional argument accepts three forms:
+
+| Form | Example | Description |
+|------|---------|-------------|
+| Local path | `./my-component.wasm` | A path to a Wasm component on disk |
+| HTTP URL | `https://example.com/component.wasm` | A remote Wasm component (requires `--digest`) |
+| Registry reference | `aws:client@1.0.0` | A package from a component registry |
+
+### Options
+
+| Flag | Description |
+|------|-------------|
+| `--to <component-id>` | Target component to add the dependency to. Prompted if omitted and the app has multiple components. |
+| `-f, --from <path>` | Path to the `spin.toml` manifest. Defaults to the current directory. |
+| `--export <name>` | Export to use from the dependency. Prompted if omitted and the component has multiple exports. |
+| `-d, --digest <sha256>` | SHA-256 digest for verifying HTTP downloads. Required for HTTP sources. |
+| `-r, --registry <url>` | Override the default registry. Only applies to registry sources. |
+| `--inherit <value>` | Capability inheritance: `true`/`all`, `false`/`none`, or comma-separated capabilities. Prompted if omitted and the dependency requires capabilities. |
+
+## Interactive Prompts
+
+When optional flags are omitted, `spin deps add` presents interactive prompts to guide the developer through each decision. The following sections illustrate the prompt flow.
+
+### Step 1: Select the target component
+
+If `--to` is omitted and the application has more than one component, the user is prompted:
+
+```
+$ spin deps add aws:client@1.0.0
+
+? Which component should the dependency be added to?
+> api-server
+  worker
+  dashboard
+```
+
+If the application has exactly one component, it is selected automatically.
+
+### Step 2: Select the export
+
+The command inspects the resolved Wasm component to enumerate its exports. If `--export` is omitted, the prompt flow depends on the number of packages and interfaces.
+
+#### Single export — auto-selected
+
+If the component exports only one interface, it is selected automatically with no prompt.
+
+#### Multiple packages — select a package first
+
+If the component exports interfaces from multiple packages, the user first selects a package:
+
+```
+? Which package should be used?
+> aws:client@1.0.0
+  aws:util@1.0.0
+```
+
+#### Within-package selection — all or a specific interface
+
+After a package is selected (or if there is only one), the user chooses between all exports from that package or a single specific interface:
+
+```
+? Which export should be used?
+> All from aws:client@1.0.0
+  aws:client/s3@1.0.0
+  aws:client/dynamodb@1.0.0
+  aws:client/sqs@1.0.0
+```
+
+Selecting **"All from aws:client@1.0.0"** records `aws:client@1.0.0` as the dependency name (a package-level selector). Selecting a specific interface records that interface (e.g. `aws:client/s3@1.0.0`).
+
+#### Explicit `--export` flag
+
+The `--export` flag accepts the same forms:
+
+- **Specific interface:** `--export aws:client/s3@1.0.0`
+- **Package selector:** `--export aws:client@1.0.0` (selects all matching exports)
+- **Plain name:** `--export my-export`
+
+### Step 3: Select capability inheritance
+
+The command inspects the dependency's imports and matches them against known capability sets (e.g. `allowed_outbound_hosts`, `ai_models`, `key_value_stores`) using semver-compatible matching. If the dependency requires any capabilities and `--inherit` is omitted, the user is prompted:
+
+```
+This dependency requires the following capabilities: allowed_outbound_hosts, ai_models
+
+? Select capabilities to inherit from the parent component
+> All
+  allowed_outbound_hosts
+  ai_models
+```
+
+Selecting **"All"** sets `inherit_configuration = true` in the manifest. Selecting individual capabilities records them as a list (e.g. `inherit_configuration = ["allowed_outbound_hosts"]`). Selecting nothing results in no inheritance.
+
+#### Explicit `--inherit` flag
+
+- `--inherit true` or `--inherit all` → inherits all capabilities
+- `--inherit false` or `--inherit none` → inherits nothing
+- `--inherit allowed_outbound_hosts,ai_models` → inherits only those capabilities
+
+### Step 4: Write to manifest and regenerate WIT
+
+After all selections are made, the command:
+
+1. Serializes the dependency into the `[component.<id>.dependencies]` table in `spin.toml`
+2. Regenerates `spin-dependencies.wit` in the component's build directory
+3. Prints a confirmation message
+
+```
+Added aws:client@1.0.0 to component 'api-server'
+
+NOTE: This dependency requires the following capabilities: allowed_outbound_hosts, ai_models
+You may need to add configuration for these capabilities to your component.
+```
+
+## End-to-End Examples
+
+### Fully interactive
+
+```
+$ spin deps add aws:client@1.0.0
+
+? Which component should the dependency be added to?
+> api-server
+
+? Which package should be used?
+> aws:client@1.0.0
+
+? Which export should be used?
+> aws:client/s3@1.0.0
+
+This dependency requires the following capabilities: allowed_outbound_hosts
+
+? Select capabilities to inherit from the parent component
+> allowed_outbound_hosts
+
+Added aws:client/s3@1.0.0 to component 'api-server'
+
+NOTE: This dependency requires the following capabilities: allowed_outbound_hosts
+You may need to add configuration for these capabilities to your component.
+```
+
+### Fully non-interactive
+
+```
+$ spin deps add aws:client@1.0.0 \
+    --to api-server \
+    --export aws:client/s3@1.0.0 \
+    --inherit allowed_outbound_hosts
+
+Added aws:client/s3@1.0.0 to component 'api-server'
+
+NOTE: This dependency requires the following capabilities: allowed_outbound_hosts
+You may need to add configuration for these capabilities to your component.
+```
+
+### Local component with all capabilities
+
+```
+$ spin deps add ./my-component.wasm --to worker --export my-export --inherit all
+
+Added my-export to component 'worker'
+```
+
+### HTTP source
+
+```
+$ spin deps add https://example.com/component.wasm \
+    --digest abc123... \
+    --to dashboard \
+    --export foo:bar/baz@0.1.0 \
+    --inherit false
+
+Added foo:bar/baz@0.1.0 to component 'dashboard'
+```
+
+## Resulting Manifest Entries
+
+The command produces entries in `spin.toml` matching the schema defined in [SIP 020](docs/content/sips/020-component-dependencies.md) and the per-dependency `inherit_configuration` field introduced in [SIP 023](docs/content/sips/023-granular-capability-inheritance.md):
+
+```toml
+# Package-level selector with full inheritance
+[component.api-server.dependencies]
+"aws:client@1.0.0" = { version = "=1.0.0", package = "aws:client", inherit_configuration = true }
+
+# Specific interface with selective inheritance
+[component.api-server.dependencies]
+"aws:client/s3@1.0.0" = { version = "=1.0.0", package = "aws:client", inherit_configuration = ["allowed_outbound_hosts"] }
+
+# Local dependency with no inheritance
+[component.worker.dependencies]
+"my-export" = { path = "my-component.wasm" }
+
+# HTTP dependency
+[component.dashboard.dependencies]
+"foo:bar/baz@0.1.0" = { url = "https://example.com/component.wasm", digest = "sha256:abc123..." }
+```
+
+## Capability Detection
+
+The command detects required capabilities by inspecting the dependency's component-level imports and matching them against the capability sets defined in [SIP 023](docs/content/sips/023-granular-capability-inheritance.md) using **semver-compatible** matching (via `wac_graph::types::are_semver_compatible`). This means a dependency importing `wasi:http/outgoing-handler@0.2.7` correctly matches the `allowed_outbound_hosts` capability set even though the set is defined with `@0.2.6`.
+
+The recognized capability sets are:
+
+| Capability | Example interfaces |
+|---|---|
+| `ai_models` | `fermyon:spin/llm` |
+| `allowed_outbound_hosts` | `wasi:http/outgoing-handler`, `wasi:sockets/tcp`, `fermyon:spin/mqtt` |
+| `environment` | `wasi:cli/environment` |
+| `files` | `wasi:filesystem/preopens` |
+| `key_value_stores` | `fermyon:spin/key-value` |
+| `sqlite_databases` | `fermyon:spin/sqlite` |
+| `variables` | `fermyon:spin/variables` |
+
+## Potential Future Work
+
+### Multiple selections within a single package
+
+The current design allows selecting either **all** exports from a package or a **single** specific interface. A natural extension would be to support selecting **multiple** (but not all) interfaces from the same package in a single invocation. For example, a multi-select prompt could allow the user to pick both `aws:client/s3@1.0.0` and `aws:client/dynamodb@1.0.0` without selecting the entire `aws:client@1.0.0` package. This would generate one dependency entry per selected interface and avoid requiring the user to run `spin deps add` multiple times for the same package.
--- a/docs/content/sips/024-spin-deps-cli-dx.md
+++ b/docs/content/sips/024-spin-deps-cli-dx.md
@@ -19,7 +19,7 @@ However, authoring the dependency entries by hand requires understanding the TOM
 `spin deps add` provides a guided CLI experience for adding a component dependency. It resolves the source, inspects the Wasm component's exports, and writes the correct entry into `spin.toml`, along with regenerating the `spin-dependencies.wit` file.
 
 # Proposal
-∏
+
 ## Command Syntax
 
 ```
@@ -40,9 +40,9 @@ The `<source>` positional argument accepts three forms:
 
 | Flag | Description |
 |------|-------------|
-| `--to <component-id>` | Target component to add the dependency to. Prompted if omitted and the app has multiple components. |
-| `-f, --from <path>` | Path to the `spin.toml` manifest. Defaults to the current directory. |
-| `--export <name>` | Export to use from the dependency. Prompted if omitted and the component has multiple exports. |
+| `--to <component-id>` | Parent component to add the dependency to. Prompted if omitted and the app has multiple components. |
+| `-f, --file <path>` | Path to the `spin.toml` manifest. Defaults to the current directory. |
+| `--export <name>` | Export to use from the dependency. Prompted if omitted and the dependency has multiple exports. |
 | `-d, --digest <sha256>` | SHA-256 digest for verifying HTTP downloads. Required for HTTP sources. |
 | `-r, --registry <url>` | Override the default registry. Only applies to registry sources. |
 | `--inherit <value>` | Capability inheritance: `true`/`all`, `false`/`none`, or comma-separated capabilities. Prompted if omitted and the dependency requires capabilities. |
@@ -218,7 +218,7 @@ The command produces entries in `spin.toml` matching the schema defined in [SIP 
 
 # Local dependency with no inheritance
 [component.worker.dependencies]
-"my-export" = { path = "my-component.wasm" }
+"my-export" = { path = "./my-component.wasm" }
 
 # HTTP dependency
 [component.dashboard.dependencies]
@@ -227,19 +227,7 @@ The command produces entries in `spin.toml` matching the schema defined in [SIP 
 
 ## Capability Detection
 
-The command detects required capabilities by inspecting the dependency's component-level imports and matching them against the capability sets defined in [SIP 023](docs/content/sips/023-granular-capability-inheritance.md) using **semver-compatible** matching (via `wac_graph::types::are_semver_compatible`). This means a dependency importing `wasi:http/outgoing-handler@0.2.7` correctly matches the `allowed_outbound_hosts` capability set even though the set is defined with `@0.2.6`.
-
-The recognized capability sets are:
-
-| Capability | Example interfaces |
-|---|---|
-| `ai_models` | `fermyon:spin/llm` |
-| `allowed_outbound_hosts` | `wasi:http/outgoing-handler`, `wasi:sockets/tcp`, `fermyon:spin/mqtt` |
-| `environment` | `wasi:cli/environment` |
-| `files` | `wasi:filesystem/preopens` |
-| `key_value_stores` | `fermyon:spin/key-value` |
-| `sqlite_databases` | `fermyon:spin/sqlite` |
-| `variables` | `fermyon:spin/variables` |
+The command detects required capabilities by inspecting the dependency's component-level imports and matching them against the capability sets defined in [SIP 023](docs/content/sips/023-granular-capability-inheritance.md) using **semver-compatible** matching. This means a dependency importing `wasi:http/outgoing-handler@0.2.7` correctly matches the `allowed_outbound_hosts` capability set even though the set is defined with `@0.2.6`.
 
 ## Potential Future Work
 
--- a/docs/content/sips/024-spin-deps-cli-dx.md
+++ b/docs/content/sips/024-spin-deps-cli-dx.md
@@ -112,7 +112,9 @@ This dependency requires the following capabilities: allowed_outbound_hosts, ai_
   [ ] ai_models
 ```
 
-The prompt is a multi-select (checkboxes). Selecting **all** listed capabilities sets `inherit_configuration = true` in the manifest. Selecting a subset records them as a list (e.g. `inherit_configuration = ["allowed_outbound_hosts"]`). Selecting nothing results in no inheritance.
+The prompt is a multi-select (checkboxes). The selected capabilities are always recorded as an explicit list — for example, selecting both options writes `inherit_configuration = ["allowed_outbound_hosts", "ai_models"]`. Selecting nothing omits the field, resulting in no inheritance.
+
+The command never writes `inherit_configuration = true`, even when every currently-listed capability is selected. `true` is a wildcard that would also grant whatever capabilities a *future* version of the dependency imports. Recording the explicit list instead ensures that choosing `allowed_outbound_hosts` and `ai_models` today does not silently opt the dependency into, say, `key_value_stores` if a later release starts importing it. (The manifest schema still accepts `true` for users who deliberately want that behavior; `spin deps add` simply does not generate it.)
 
 ### Step 4: Write to manifest and regenerate WIT
 
@@ -188,10 +190,10 @@ If the selected trigger already has middleware, the command displays the current
   existing-auth
 ▶ [ensure-admin]  ← new
   existing-logger
-  ─── component ───
+  ─── application component ───
 ```
 
-By default — and whenever the trigger has no existing middleware — the entry is appended to the end of the pipeline (closest to the component).
+By default — and whenever the trigger has no existing middleware — the entry is appended to the end of the pipeline (closest to the application component).
 
 ### Step M3: Select capability inheritance
 
@@ -270,6 +272,8 @@ Run `spin build` to generate language bindings for the new dependency.
 
 ### Middleware (interactive)
 
+Here the `/admin/...` route already has an `authn` middleware, so the position prompt appears; the new authorization middleware is placed after it:
+
 ```
 $ spin deps add ./ensure-admin.wasm
 
@@ -279,8 +283,9 @@ Detected HTTP middleware (imports and exports wasi:http/handler).
 > /admin/...
 
 ? Position the middleware in the pipeline (use ↑↓ to move):
+  authn
 ▶ [ensure-admin]  ← new
-  ─── component ───
+  ─── application component ───
 
 This middleware requires the following capabilities: allowed_outbound_hosts
 
@@ -308,9 +313,9 @@ Added middleware 'ensure-admin' to the trigger for route '/admin/...'
 The command produces entries in `spin.toml` matching the schema defined in [SIP 020](docs/content/sips/020-component-dependencies.md) and the per-dependency `inherit_configuration` field introduced in [SIP 023](docs/content/sips/023-granular-capability-inheritance.md):
 
 ```toml
-# Package-level selector with full inheritance
+# Package-level selector inheriting explicitly-chosen capabilities
 [component.api-server.dependencies]
-"aws:client@1.0.0" = { version = "=1.0.0", package = "aws:client", inherit_configuration = true }
+"aws:client@1.0.0" = { version = "=1.0.0", package = "aws:client", inherit_configuration = ["allowed_outbound_hosts", "ai_models"] }
 
 # Specific interface with selective inheritance
 [component.api-server.dependencies]
--- a/docs/content/sips/024-spin-deps-cli-dx.md
+++ b/docs/content/sips/024-spin-deps-cli-dx.md
@@ -4,9 +4,9 @@ date = "2026-04-09T00:00:00Z"
 
 ---
 
-Summary: A CLI command (`spin deps add`) for adding component dependencies to a Spin application, with interactive prompts for selecting components, exports, and capability inheritance. The command also detects when the resolved component is HTTP middleware and guides attaching it to a trigger selected by route.
+Summary: A CLI command (`spin deps add`) for adding component dependencies to a Spin application, with interactive prompts for selecting components, interfaces, and capability inheritance. The command also detects when the resolved component is HTTP middleware and guides attaching it to a trigger selected by route.
 
-Owner(s): [bhardock@akamai.com](mailto:bhardock@akamai.com)
+Owner(s): Brian Hardock
 
 Created: April 9, 2026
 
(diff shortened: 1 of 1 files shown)


docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+? Which export should be used?
+> All from aws:client@1.0.0
+  aws:client/s3@1.0.0
+  aws:client/dynamodb@1.0.0
+  aws:client/sqs@1.0.0
+```
+
+Selecting **"All from aws:client@1.0.0"** records `aws:client@1.0.0` as the dependency name (a package-level selector). Selecting a specific interface records that interface (e.g. `aws:client/s3@1.0.0`).
+
+#### Explicit `--export` flag
```
itowlson, line 101: ~This is _not_ surfaced interactively, right?  (I would _like_ it not to be - I just want to check we are on the same page.)~

oh wait, this is the interface on the _left_ hand side, not the `export =` option?  That coincidence of terminology is going to be a trap

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+### Step 3: Select capability inheritance
+
+The command inspects the dependency's imports and matches them against known capability sets (e.g. `allowed_outbound_hosts`, `ai_models`, `key_value_stores`) using semver-compatible matching. If the dependency requires any capabilities and `--inherit` is omitted, the user is prompted:
+
+```
+This dependency requires the following capabilities: allowed_outbound_hosts, ai_models
+
+? Select capabilities to inherit from the parent component
+> All
+  allowed_outbound_hosts
```
itowlson, line 118: Presumably these will be check boxes rather than a single-select?

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+  ai_models
+```
+
+Selecting **"All"** sets `inherit_configuration = true` in the manifest. Selecting individual capabilities records them as a list (e.g. `inherit_configuration = ["allowed_outbound_hosts"]`). Selecting nothing results in no inheritance.
+
+#### Explicit `--inherit` flag
+
+- `--inherit true` or `--inherit all` → inherits all capabilities
+- `--inherit false` or `--inherit none` → inherits nothing
+- `--inherit allowed_outbound_hosts,ai_models` → inherits only those capabilities
```
itowlson, line 128: Or `--inherit allowed_outbound_hosts --inherit ai_models`, right?  (that is, as well as not instead of)

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+
+The command inspects the resolved Wasm component to enumerate its exports. If `--export` is omitted, the prompt flow depends on the number of packages and interfaces.
+
+#### Single export — auto-selected
+
+If the component exports only one interface, it is selected automatically with no prompt.
+
+#### Multiple packages — select a package first
+
+If the component exports interfaces from multiple packages, the user first selects a package:
```
itowlson, line 79: I worry with this that we may be asking a question the user doesn't know how to answer.  I want the authentication interface, is that in `client` or `util`?  If I guess wrong, is there a way back?

Possible alternative approaches:

1. Show them a list of _everything_ they could import - all the interfaces plus an "everything" for each package.
2. When they come to interface selection, offer a "The interface I'm looking for isn't here. Take me back to package selection" option

Or maybe this isn't a concern and we have high confidence that users will know what they are looking for and just want to save on ye olde typing, I am not sure.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+> aws:client@1.0.0
+  aws:util@1.0.0
+```
+
+#### Within-package selection — all or a specific interface
+
+After a package is selected (or if there is only one), the user chooses between all exports from that package or a single specific interface:
+
+```
+? Which export should be used?
```
itowlson, line 92: Nit: I would avoid saying 'export' here (because I want to _import_ something, where is the choice for that).  Consider e.g. "Which interface do you want to import?" or some better wording that avoids the whole import/export debacle altogether (a la "which interface do you want to use" although that's hardly going to set the poetry world alight either)

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+## Capability Detection
+
+The command detects required capabilities by inspecting the dependency's component-level imports and matching them against the capability sets defined in [SIP 023](docs/content/sips/023-granular-capability-inheritance.md) using **semver-compatible** matching (via `wac_graph::types::are_semver_compatible`). This means a dependency importing `wasi:http/outgoing-handler@0.2.7` correctly matches the `allowed_outbound_hosts` capability set even though the set is defined with `@0.2.6`.
+
+The recognized capability sets are:
+
+| Capability | Example interfaces |
+|---|---|
+| `ai_models` | `fermyon:spin/llm` |
+| `allowed_outbound_hosts` | `wasi:http/outgoing-handler`, `wasi:sockets/tcp`, `fermyon:spin/mqtt` |
```
itowlson, line 237: `wasi:http/client`, `spin:postgres`, `fermyon:spin/mysql`, `spin:mqtt`, Redis itfs

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+| `environment` | `wasi:cli/environment` |
+| `files` | `wasi:filesystem/preopens` |
+| `key_value_stores` | `fermyon:spin/key-value` |
+| `sqlite_databases` | `fermyon:spin/sqlite` |
+| `variables` | `fermyon:spin/variables` |
```
itowlson, line 242: All of these have `spin:*` forms as of Spin 4 (actually SQLite as of Spin 3.2 I think)

itowlson (commented):
LGTM.  I like the capabilities stuff; in fact I almost wonder if I should be able to run the capabilities stuff outside of doing an add.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: They're equivalent when selecting interfaces. I couldnt think of a better name here that encapsulated packages and interfaces. How do you feel about `--use`?

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: Indeed

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: Wait once again it seems I am utterly confused about a feature we have been shipping for a gazillion years - I thought `export` was about remapping interfaces... _sigh_

I was initially going to suggest `--interface`, but I now realise that if you want all a package's APIs, you have to tell it the package name if you don't want it to interact.  `--use` seems a bit open-ended but could be an option.  We could try `--import` perhaps?  It's not great though.  Well maybe it's okay, I originally thought "ugh, importing exports" but well, huh, that is what importers do, so maybe it makes sense?

Or send up the BIKESHED BAT-SIGNAL and let Lann come up with something.

fibonacci1729: I think import works. We could also keep export for the remapping scenario you mention which is the point of it. All I meant was that if the LHS is an interface, it implies that an export with that name exists with that name so its technically equivalent.

fibonacci1729: I.e. `"a:b/c" = { export = none, ... }` is equivalent to  `"a:b/c" = { export = "a:b/c", ... }`

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: Good call out. Ill have a tinker and see if i could smooth this out. I think listing everything with all options for each package is the way to go.

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: Aha, I think I see what you mean now.  Thanks!

fibonacci1729: We could allow add a mutually exclusive --package/-p, --interface/-i.

fibonacci1729: Actually never mind, I forgot about plain names.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Source Formats
+
+The `<source>` positional argument accepts three forms:
+
+| Form | Example | Description |
+|------|---------|-------------|
+| Local path | `./my-component.wasm` | A path to a Wasm component on disk |
+| HTTP URL | `https://example.com/component.wasm` | A remote Wasm component (requires `--digest`) |
+| Registry reference | `aws:client@1.0.0` | A package from a component registry |
```
karthik2804, line 37: Should we also support adding by component ID? I am imagining a case where we have a component that we already build as part of the manifest that we would like to add as dependency to another.

rmarx: Strong agree on this one. I don't want to have to know/copy-paste the build output path of my Rust component to add it as a dependency of my TS component in the same spin project. I want that to be auto-figured out from the component ID

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Options
+
+| Flag | Description |
+|------|-------------|
+| `--to <component-id>` | Target component to add the dependency to. Prompted if omitted and the app has multiple components. |
+| `-f, --from <path>` | Path to the `spin.toml` manifest. Defaults to the current directory. |
```
rmarx: Nit: This feels a bit weird... I get that it's consistent with e.g., `spin build`, but there it makes more semantic sense (you build "from" a .toml). 

In this setup, it's a bit weird given that you're adding dependencies TO something.

In `spin add` the `-f` is `--file` not `--from`, which would make more sense here as well imo.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+| `--export <name>` | Export to use from the dependency. Prompted if omitted and the component has multiple exports. |
```
rmarx:
```suggestion
| `--export <name>` | Export to use from the dependency. Prompted if omitted and the dependency has multiple exports. |
```

docs/content/sips/024-spin-deps-cli-dx.md
rmarx: I was thinking something similar, but you might run into an issue where there's just a TON of exports in a given dependency, and navigating them manually would get annoying.

Counterpoint is of course that that won't happen super-often and you only need to do this once/infrequently, so I'd also be in favor of having just 1 flat list of everything with the "All from DEP" in the list as just another option.

docs/content/sips/024-spin-deps-cli-dx.md
rmarx: I had a similar feeling when seeing the CLI argument is called `--export` instead of `--import`... it depends on how you want to look at it of course.

However, since I'm `adding` a dependency `to` a specific component, I probably want that component to `import`/use stuff from the dependency (I don't necessarily want to think about the stuff the dependency `export`s).

Kind of like a github PR should really be a Merge Request instead? :D (imo)

Either way, if we keep `--export` (instead of `--import` or `--interfaces`) I agree it would be more sensible to have some other language here like suggested by @itowlson 

EDIT: just now reading the whole discussion on this below :)

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+> api-server
+  worker
+  dashboard
+```
+
+If the application has exactly one component, it is selected automatically.
+
+### Step 2: Select the export
+
+The command inspects the resolved Wasm component to enumerate its exports. If `--export` is omitted, the prompt flow depends on the number of packages and interfaces.
```
rmarx: I'm wondering if this document should also detail the failure path? Or if there's somewhere else that describes exactly how you "resolve" a Wasm component from an HTTP URL/registry and what happens if it can't be found / if it doesn't actually export anything that can be used?

What would be the error message shown if the dependency can't be found? If it's somehow incompatible (e.g., wasm-bindgen .wasm from HTTP URL)?

When does spin try the resolution? Before or after or parallel with showing the interactive prompt for `--to`? Is there a progress-bar or something for large/slow downloads?

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+  aws:client/sqs@1.0.0
+```
+
+Selecting **"All from aws:client@1.0.0"** records `aws:client@1.0.0` as the dependency name (a package-level selector). Selecting a specific interface records that interface (e.g. `aws:client/s3@1.0.0`).
+
+#### Explicit `--export` flag
+
+The `--export` flag accepts the same forms:
+
+- **Specific interface:** `--export aws:client/s3@1.0.0`
```
rmarx: EDIT: just read the bottom of the doc saying that multiple exports are not yet in-scope. Decided to leave the comment though, as I do need we can/should make it a bit clearer in this doc/eventual documentation that it's intended you can only add 1 interface at a time. 

Just want to confirm if/how multiple exports work. I assume that 

`--export aws:client/s3@1.0.0,aws:client/sqs@1.0.0`
and
`--export aws:client/s3@1.0.0 --export aws:client/sqs@1.0.0`

would both work? Or is it always just 1 export at a time?

If there are multiple possible at once, then some text above needs to be adjusted (e.g., "Which exportS should be used").

If however you can only do 1 interface at a time, that can probably also be made a bit clearer in the text/description that people would run multiple `spin deps add` commands for multiple exports, even if from the same package.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+|------|---------|-------------|
+| Local path | `./my-component.wasm` | A path to a Wasm component on disk |
+| HTTP URL | `https://example.com/component.wasm` | A remote Wasm component (requires `--digest`) |
+| Registry reference | `aws:client@1.0.0` | A package from a component registry |
+
+### Options
+
+| Flag | Description |
+|------|-------------|
+| `--to <component-id>` | Target component to add the dependency to. Prompted if omitted and the app has multiple components. |
```
rmarx:
```suggestion
| `--to <component-id>` | Parent component to add the dependency to. Prompted if omitted and the app has multiple components. |
```

To be more consistent with later prose (i.e., inheriting "parent" component capabilities)

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+```
+This dependency requires the following capabilities: allowed_outbound_hosts, ai_models
+
+? Select capabilities to inherit from the parent component
+> All
+  allowed_outbound_hosts
+  ai_models
+```
+
+Selecting **"All"** sets `inherit_configuration = true` in the manifest. Selecting individual capabilities records them as a list (e.g. `inherit_configuration = ["allowed_outbound_hosts"]`). Selecting nothing results in no inheritance.
```
rmarx: So is that `inherit_configuration = []` or simply no `inherit_configuration`?

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+After all selections are made, the command:
+
+1. Serializes the dependency into the `[component.<id>.dependencies]` table in `spin.toml`
+2. Regenerates `spin-dependencies.wit` in the component's build directory
+3. Prints a confirmation message
+
+```
+Added aws:client@1.0.0 to component 'api-server'
+
+NOTE: This dependency requires the following capabilities: allowed_outbound_hosts, ai_models
```
rmarx: Is this printed only for capabilities that were not inherited? 

Would be confusing if you explicitly chose `--inherit all` or `--inherit ai_models` and then still saw this kind of message at the end?

The example below does indicate this warning would be present if inheriting, so I think that's not ideal.

rmarx: Reading it a bit more, it seems it's saying you might need to actually fill in `ai_models` on the PARENT component, even if you added it as --inherit here.

Conceptually that makes sense, but it's confusing imo. 

Is there a way to check if the parent component has the capabilities already filled out and only print this type of warning if it doesn't?

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+[component.api-server.dependencies]
+"aws:client@1.0.0" = { version = "=1.0.0", package = "aws:client", inherit_configuration = true }
+
+# Specific interface with selective inheritance
+[component.api-server.dependencies]
+"aws:client/s3@1.0.0" = { version = "=1.0.0", package = "aws:client", inherit_configuration = ["allowed_outbound_hosts"] }
+
+# Local dependency with no inheritance
+[component.worker.dependencies]
+"my-export" = { path = "my-component.wasm" }
```
rmarx:
```suggestion
"my-export" = { path = "./my-component.wasm" }
```

Nit for consistency

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,248 @@
...
+- `--inherit true` or `--inherit all` → inherits all capabilities
+- `--inherit false` or `--inherit none` → inherits nothing
+- `--inherit allowed_outbound_hosts,ai_models` → inherits only those capabilities
+
+### Step 4: Write to manifest and regenerate WIT
+
+After all selections are made, the command:
+
+1. Serializes the dependency into the `[component.<id>.dependencies]` table in `spin.toml`
+2. Regenerates `spin-dependencies.wit` in the component's build directory
```
rmarx, line 124: Reading this made me wonder what happens if a user wants to REMOVE a dependency instead of adding it (obviously not covered by `spin deps ADD` but still).

Is it just removing the line in the .toml and `spin build` picks that up and re-generates the `spin-dependencies.wit` to remove the dependency? Or do you need a separate command to regen `spin-dependencies.wit`?

In the first case: why would regeneration be part of `spin deps add` instead of just `spin build`?
In the second case: do we have a separate `spin deps remove` or something to help re-trigger?

rmarx (commented):
Hey all,

Thanks for working on this important part of the spin ecosystem.

I added a few comments, but most of them are nits and only few have to do with the underlying proposal/design, which in general I think is solid. 

I'm quite new to spin (Akamai employee working with the Fermyon team) so I might have made some wrong assumptions/basic errors (e.g., not knowing how external dependency resolution error handling is done, not knowing if there's a way to remove added dependencies, ...). Feel free to correct me where wrong :) 

One of the bigger missing pieces in this proposal as a higher-level user is the tie-in with WIT and auto-generation of programming language bindings when a dependency is added. I reckon that's documented elsewhere / seen as a separate project (which is fine), but I might have expected just a bit more tie-in to how that's supposed to interact here. Currently the text just talks about re-generating the `spin-dependencies.wit` file, but it's not entirely clear what that does exactly/how that flows into getting the dependency actually ready for use from within other components. If another command is needed (say... `spin deps bind` who knows) it would be good if that next step would e.g., be mentioned in the output text after a `spin deps add` potentially).

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: @rmarx We don't usually go down to the level of "error message text" in SIPs.  We generally approach these as high-level design discussions rather than detailed functional specifications.  If we cover things like error messages it tends to be as "and the error message should include such-and-such piece of diagnostic information."

HTTP and registry resolution are established code paths in Spin (and specifically in the dependency resolution subsystem) - I imagine we'll follow those.

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: @rmarx They are equivalent.  (Well, there is an edge case in which they are not equivalent.  But we should catch that first and not ask the question if we are in that edge case!  @fibonacci1729 I'm thinking of where the component already has a `dependencies_inherit_configuration` field.)

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: @rmarx Hmm, good question.  If the dep used the `llm` API, you would need the parent component to have `ai_models` so that the dep had something to inherit, _and_ you would need the dep to inherit that setting.

We need to print a warning if the dep uses `llm` and parent has `ai_models` but the user _declined_ to inherit it.  We might ideally print a slightly different warning if the dep uses `llm` and inherits `ai_models` but the parent doesn't set `ai_models`.  And maybe another one if the dep doesn't inherit _and_ the parent doesn't set it...!

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: @rmarx If memory serves, one of the motivations for `add` regenerating `spin-dependencies.wit` was so that you got code completion in your editor without having to perform an artificial build cycle first.  Another reason for having an `add` command rather than typing into TOML was to analyse the incoming dep and help you get it correctly configured.

These considerations largely go away for the remove case.  You can just delete the line in the TOML (for me at least, probably easier than typing the CLI incantation).  Okay, code completion might briefly offer you APIs you no longer want to use (until the first build).  But configuration concerns go away, and your editor will not sulk about your code.

So it's not clear that `remove` adds value.  For myself, I wouldn't invest effort until and unless users asked for it.  I do realise it feels a bit asymmetric though...

itowlson:
@rmarx Thanks for asking about how interfaces flow from `spin-dependencies.wit` into application code.  It's a good question and not well documented.  The answer is that Spin punts this to languages.

For example, the Rust SDK has a macro that generates bindings from `spin-dependencies.wit` if it exists.  I think Karthik developed something that sits in the JavaScript toolchain although I am not sure of details.  So the intent is that it should magically work if you use the templates.  We're not there yet - I don't think all templates are enlightened (possibly none are yet, I think JS stuff is a bit in limbo over WASI P3 stuff).  But the basic principle is that Spin makes the dependencies WIT available to the language tools, and the language tools can pull that in in whatever way makes sense for that language.  That avoids Spin itself having to know how to generate bindings for however many languages support the Component Model!

So all of that is out of scope for this SIP, but it's fair to say the reason it's out of scope is folk knowledge rather than explicitly documented.

karthik2804:
Here is test case as sample for how it will work in JS - https://github.com/spinframework/spin-js-sdk/tree/main/test/deps-test

https://github.com/spinframework/spin-js-sdk/blob/643f08206354bc43e24ac52d60adc3b9822c559d/test/deps-test/package.json#L8-L10 is the important bit where we have a NPM script for generating the bindings that gets called with the build step first before building the component.

docs/content/sips/024-spin-deps-cli-dx.md
rmarx: Yeah, I suspected something like this, so chuck it up to me being unfamiliar with the wider project/process :)

docs/content/sips/024-spin-deps-cli-dx.md
rmarx: Yeah, strongly in favor of more fine-grained warnings here :)

Another thing I was thinking about just now re-reading this: what happens if the user removes something related to this from the .toml down the line (e.g., removes `ai_models` from the parent while keeping it as a requirement/inherits on the dependency). Will `spin build` do a similar check as `spin deps add` to detect that and issue a warning (as it probably should)?

docs/content/sips/024-spin-deps-cli-dx.md
rmarx: Ok, thanks for that, again a case of me missing some context on the overall project then.

To be clear: I wasn't advocating for `remove`. I was just wondering what the removal process looks like. If it's indeed as simple as changing the .toml and `spin build` will take take of the rest automagically (minus the autocomplete), that's fine with me.

rmarx:
Hey @itowlson and @karthik2804, thanks for the added context here :) 

While I of course still agree this is/should be a separate thing, I'm not _entirely_ agreed with "that is out of scope for this SIP"... (though all I'm really asking for is a "Don't forget to run `spin build` to generate language bindings" part of the output :))


In [one of the comments above](https://github.com/spinframework/spin/pull/3453#discussion_r3215494220) where I asked about maybe needing a `spin deps remove`, I got the impression that updating `spin-dependencies.wit` during `spin deps add` was **mostly** to provide IDE support for the new dependencies before a new `spin build` was executed. 

With Karthik's example above though (and hopefully correctly interpreting @itowlson), it's clear that the generation of the language bindings only happens during the build process. Put differently IIUC: after `spin deps add` the programmer would conceptually get partial IDE support via `spin-dependencies.wit` but would need `spin build` for "full" IDE support in the programming language.

If this is correctly interpreted on my end, then I **think we can make this clearer to the programmer by explicitly mentioning the recommendation to run `spin build` after `spin deps add`.** 


To re-iterate, I think it's fine to do the binding generation etc. at build time. I'm just saying that it might be confusing to newer users that are unfamiliar with the ecosystem (and that are maybe just copy-pasting a `spin deps` command from a README), that the dependency isn't immediately "fully" usable (e.g., from a TypeScript component because the bindings haven't been generated yet). I think I can say that fairly confidently since I remember how confused I was with all of this a year ago (and still a bit tbh :D). 

TL;DR: if `spin deps add` doesn't do the complete logic that is needed to make the dependency directly callable from the parent component in userland code, imo that should be clear(er) to the end user.

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: That's a good question.  Naively, the behaviour would be that you inherit the parent's `ai_models` permissions, which are "nothing."  So if the dep really did use the `llm` interface it would get an "access denied" error for all models at runtime, which... is correct, but the reason might be non-obvious.  So yeah probably helpful to emit the same warning at build time as we would have done at add time.  @fibonacci1729 what do you reckon?  Effort?

itowlson:
> it's clear that the generation of the language bindings only happens during the build process

Well - and I apologise for the pedantry - it depends.  But yes in the general case the user _might_ need to build or refresh their IDE or whatever to get the bindings to regenerate.  (The Rust language server is maddening in that in theory it can pick up changes without a build... but in practice it often doesn't pick up changes _even after_ a build, because the build and language server are completely independent.  Whereas in JS the IDE will _never_ pick up without a build - barring watch shenanigans - but a build will reliably make them available in the IDE.)  So we probably want a dreadfully vague and generic message along the lines of "you may need to build or reload the project to see your bindings."

I hope that makes sense - I do sympathise with the confusion!

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: I think the best we could do at build time is check if a particular capability was inherited and issue a warning iff the configuration is the empty set.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: That's a good point @rmarx. I don't anticipate dependencies that will have an overwhelmingly large # of exports so I agree with your counterpoint.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: Good catch.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: Could you say more about the edge case? I'm not sure i'm following.

itowlson: The edge case is where `dependencies_inherit_configuration` is set.  In that case, no `inherit_configuration` is acceptable, and `inherit_configuration = []` is an error (because it implies "inherit nothing" which contradicts `dependencies_inherit_configuration`).  So if DIC is already set on the component, `spin dep add` shouldn't ask about capabilities.

fibonacci1729:
@itowlson in my last commit i added a sketch of middleware flow. If you could take a look, thanks!

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,386 @@
+title = "SIP 024 - spin deps cli dx"
+template = "main"
+date = "2026-04-09T00:00:00Z"
+
+---
+
+Summary: A CLI command (`spin deps add`) for adding component dependencies to a Spin application, with interactive prompts for selecting components, exports, and capability inheritance. The command also detects when the resolved component is HTTP middleware and guides attaching it to a trigger selected by route.
+
+Owner(s): [brian.hardock@fermyon.com](mailto:brian.hardock@fermyon.com)
```
itowlson: Might want to remove the email address (or if you prefer to have it in then update it)

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,386 @@
...
+| `--export <name>` | Export to use from the dependency. Component dependencies only. Prompted if omitted and the dependency has multiple exports. |
+| `-d, --digest <sha256>` | SHA-256 digest for verifying HTTP downloads. Required for HTTP sources. |
+| `-r, --registry <url>` | Override the default registry. Only applies to registry sources. |
+| `--inherit <value>` | Capability inheritance: `true`/`all`, `false`/`none`, or comma-separated capabilities. Prompted if omitted and the dependency requires capabilities. |
+| `--middleware` | Force treating the dependency as HTTP middleware. By default this is auto-detected from the component's imports and exports. |
+| `--route <route>` | HTTP route of the trigger to attach middleware to. Middleware only. Prompted if omitted and the app has multiple HTTP triggers. |
+| `--trigger <id>` | Target a trigger by its `id` instead of its route. Middleware only; useful for private endpoints, which have no route. |
+| `--before <ref>`, `--after <ref>` | Position the middleware before/after an existing entry in the trigger's pipeline. Middleware only. Defaults to appending to the end of the pipeline. |
```
itowlson: Your concerns about the panoply of flags are well founded.  "Oh you pass `--to` unless it's middleware in which case you pass `--route` unless it's a private endpoint in which case you pass `--trigger`."  It's too much, and the typing is onerous (chances of my correctly typing out a route first time: zero).

And then we get into which flags can be used when, like middleware can't have `--export`.

And _then_ (sorry) what do before and after look like?  I already have an authorisation middleware from `{ registry = "spinframework.example.com", package = "auth:auth", version = "1.0.0" }`, what to I write to insert my bot detection middleware before that?  I'm struggling to think of anything that would be practical, but perhaps I'm missing what you're thinking?

I feel like this is reaching the point of "just make it interactive only, and if a scripting/CI scenario comes up then consider adding flags for that then."  Thoughts?

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: This remains a concern for me.

docs/content/sips/024-spin-deps-cli-dx.md
itowlson: This remains a source of potential confusion for me.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,386 @@
...
+
+Middleware is attached to a trigger, not a component. If `--route` is omitted and the application has more than one HTTP trigger, the user selects the trigger by its route:
+
+```
+$ spin deps add ./ensure-admin.wasm
+
+Detected HTTP middleware (imports and exports wasi:http/handler).
+
+? Which HTTP route should the middleware be added to?
+> /admin/...
```
itowlson, line 174: This could be multi-select, although I guess that would mean we cannot have the nice ordering and inherits stuff.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,386 @@
...
+  /api/...
+  (private) [component: health-check]
+```
+
+If the application has exactly one HTTP trigger, it is selected automatically. Private endpoints (which have no route) are listed by their component id and can be targeted non-interactively with `--trigger <id>`.
```
itowlson: I am not sure we should be encouraging middleware on private endpoints (in fact I am wondering if I should ban it).  Internally the presence of middleware can result in transformations which can affect the component names used for service chaining.

(In fact wait don't service chaining requests go straight to the component?  They don't pass through any trigger, and middleware is attached to triggers...)

But also logically, in a service chaining scenario, you're already inside the walled garden - SC is more like composition than like a routed HTTP request - you should have run your middleware on the way into the service, not between two components within the service.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,386 @@
...
+
+### Step M2: Choose the pipeline position
+
+The `middleware` field is an ordered pipeline. A request flows through the middlewares from front to back before reaching the application component, and the response flows from back to front. For example, authentication middleware should sit ahead of authorisation middleware. If the selected trigger already has middleware, the user chooses where the new entry goes:
+
+```
+? Where should this middleware run in the pipeline?
+> Last (closest to the component)
```
itowlson: This is going to get quite confusing with longer pipelines (will each entry have before and after list items? won't some of those be equivalent?).  I wonder if a widget would work that shows the current pipeline with a placeholder for the new thing and the user uses arrow keys to move it up and down a la

```
  existing thing 1
  existing thing 2
[ the new thing      ]

# user hits arrow up

  existing thing 1
[ the new thing      ]
  existing thing 2
```

kind of thing maybe?  (of course would need some chrome to show the direction of the pipeline)

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,386 @@
...
+
+### Step M3: Select capability inheritance
+
+Capability inheritance works exactly as for component dependencies (see [Step 3](#step-3-select-capability-inheritance)), producing an `inherit_configuration` value on the middleware entry. Because middleware is attached to a trigger rather than a component, the capabilities it inherits come from whichever component the trigger runs, so the command emphasises that **every** component served by the trigger must grant those capabilities:
+
+```
+This middleware requires the following capabilities: allowed_outbound_hosts
+
+? Select capabilities to inherit from the parent component
+> allowed_outbound_hosts
```
itowlson: This definitely needs to be multi-select.

It may be worth thinking ahead to how this will look in a hypothetical future where middlewares can have their own independent capabilities, rather than inheriting.  I guess maybe just an option for "I'm gonna enter custom permissions" and then it tells you to bust TOML?

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,386 @@
...
+
+### Step M4: Write to manifest
+
+The command appends the middleware entry to the selected trigger's `dependencies.middleware` array and prints a confirmation. Unlike the component-dependency flow, no `spin-dependencies.wit` is generated — middleware is composed onto the component by the HTTP trigger at load time and does not satisfy a component import.
+
+```
+Added middleware './ensure-admin.wasm' to the trigger for route '/admin/...'
+
+NOTE: This middleware requires the following capabilities: allowed_outbound_hosts
+Ensure every component served by this route grants these capabilities.
```
itowlson: Every component served by this route - I thought a route mapped to a single component?

itowlson: Why are we telling the user to ensure it?  Can't we check?  I had assumed the list of "things you can inherit" was populated from the component at the destination of the route being gemiddlewared.

itowlson (commented):
I left a comment on the CLI experience and the middleware stuff, and started looking at the other things but I don't think they've changed?  At least I ran into a couple of things I was worried about last review and they still seem possible concerns.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: Made the experience interactive only and left flags for future work.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: I considered that too but yeah it gets pretty gnarly.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: Disallowing adding middlewares to private endpoints in the flow.

fibonacci1729:
Ok @itowlson in my last commit the experience is now entirely interactive with all the bells and whistles suggested around multi selection etc. PTAL.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,373 @@
...
+
+```
+This dependency requires the following capabilities: allowed_outbound_hosts, ai_models
+
+? Select capabilities to inherit from the parent component
+  [ ] allowed_outbound_hosts
+  [ ] ai_models
+```
+
+The prompt is a multi-select (checkboxes). Selecting **all** listed capabilities sets `inherit_configuration = true` in the manifest. Selecting a subset records them as a list (e.g. `inherit_configuration = ["allowed_outbound_hosts"]`). Selecting nothing results in no inheritance.
```
itowlson: I don't think full inheritance is safe here.  If I selected to inherit AOH and AI, then inherit me AOH and AI.  (The concern is around future versions: if v1.1 of the dep imports the KV interface, my selection of AOH and AI shouldn't opt it into KV as well.)

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,373 @@
...
+The `middleware` field is an ordered pipeline. A request flows through the middlewares from front to back before reaching the application component, and the response flows from back to front. For example, authentication middleware should sit ahead of authorization middleware.
+
+If the selected trigger already has middleware, the command displays the current pipeline and lets the user position the new entry using arrow keys:
+
+```
+? Position the middleware in the pipeline (use ↑↓ to move):
+  existing-auth
+▶ [ensure-admin]  ← new
+  existing-logger
+  ─── component ───
```
itowlson: "component" is overloaded ("wait, I thought the thing I was adding was a component") - maybe e.g. "application component," "request handler," "HTTP service," something else?

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,373 @@
...
+```
+$ spin deps add ./ensure-admin.wasm
+
+Detected HTTP middleware (imports and exports wasi:http/handler).
+
+? Which HTTP route should the middleware be added to?
+> /admin/...
+
+? Position the middleware in the pipeline (use ↑↓ to move):
+▶ [ensure-admin]  ← new
```
itowlson, line 287: I think you said this would be omitted if the pipeline was currently empty.

docs/content/sips/024-spin-deps-cli-dx.md
```diff
@@ -0,0 +1,373 @@
...
+
+### Step 3: Select capability inheritance
+
+The command inspects the dependency's imports and matches them against known capability sets (e.g. `allowed_outbound_hosts`, `ai_models`, `key_value_stores`) using semver-compatible matching. If the dependency requires any capabilities, the user is prompted:
+
+```
+This dependency requires the following capabilities: allowed_outbound_hosts, ai_models
+
+? Select capabilities to inherit from the parent component
+  [ ] allowed_outbound_hosts
```
itowlson, line 111: Open question: should we default these boxes to ticked?  We know that the dep is likely to need them, and so if the user leaves them unticked there's a good chance it will fail in weird ways.

I am not sure whether this goes against our normal posture of "opt in to access" though.  On the one hand, yes it does.  On the other hand, it's still asking the human to grant informed permission, it's just giving them a helping hand.

I guess the big risk is that people won't read the list of permissions they're granting.  We'll train them to click through and then "hey I thought I was depending on a random number generator and now it's exfiltrated my database."

itowlson (commented):
This looks really good.  Losing the scriptability makes the command a lot simpler and more approachable.  I saw export mapping had dropped out too, which I think is the right decision for now.  I did leave a couple of questions and nitpicks but the core flows feel nicely focused and unambiguous - thanks for nutting this out and for the thoughtful iteration.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: Yeah I had considered defaulting them to "selected" but the risk you mention is my main concern.

docs/content/sips/024-spin-deps-cli-dx.md
fibonacci1729: "application component" makes sense to me.

fibonacci1729:
@itowlson I think i addressed everything!

itowlson (approved):
LGTM - thanks!

PR #3459, opened 2026-04-14
Add content-length header attributes to outbound HTTP spans

ChihweiLHBird:
## Description

Records `http.request.header.content-length` and `http.response.header.content-length` on outbound HTTP spans when the header is present.

Uses `set_attribute()` from `tracing_opentelemetry::OpenTelemetrySpanExt` because the OTel attribute names contain hyphens, which `tracing`'s `#[instrument]` field system cannot represent.

I didn't implement a test because I thought it's not worth to add another expensive integration test case just for a new field sent to OTel collector.

I used a small HTTP component and [otel-desktop-viewer](https://github.com/CtrlSpice/otel-desktop-viewer) to verify this works as expected:

```rust
#[http_component]
async fn handle_test_spin_rust(req: Request) -> anyhow::Result<impl IntoResponse> {
    println!("Handling request to {:?}", req.header("spin-full-url"));

    let outbound_request = Request::get("https://httpbin.org/get")
        .header("accept", "application/json")
        .build();
    let outbound_response: Response = spin_sdk::http::send(outbound_request)
        .await
        .context("failed to call https://httpbin.org/get")?;
    let outbound_status = outbound_response.status();
    let outbound_body_len = outbound_response.body().len();
    
    Ok(Response::builder()
        .status(200)
        .body(format!(
            "Hello World!\nhttpbin.org/get status: {outbound_status}\nhttpbin.org/get body bytes: {outbound_body_len}"
        ))
        .build())
}
```

Addresses part of https://github.com/spinframework/spin/issues/3188 (Option 1).

### Changes

- `crates/factor-outbound-http/Cargo.toml` — added `tracing-opentelemetry` dependency
- `crates/factor-outbound-http/src/wasi.rs` — added `record_content_length_header` helper and call sites for request (after interceptor) and response

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -12,7 +12,11 @@ use std::{
 
 use bytes::Bytes;
 use futures::channel::oneshot;
-use http::{header::HOST, uri::Scheme, Uri};
+use http::{
+    header::{CONTENT_LENGTH, HOST},
+    uri::Scheme,
+    HeaderMap, Uri,
+};
 use http_body::{Body, Frame, SizeHint};
 use http_body_util::{combinators::UnsyncBoxBody, BodyExt};
 use hyper_util::{
@@ -35,7 +39,7 @@ use tokio::{
 };
 use tokio_rustls::client::TlsStream;
 use tower_service::Service;
-use tracing::{field::Empty, instrument, Instrument};
+use tracing::{field::Empty, instrument, Instrument, Span};
 use wasmtime::component::HasData;
 use wasmtime_wasi::TrappableError;
 use wasmtime_wasi_http::{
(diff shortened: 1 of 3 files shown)


crates/factor-outbound-http/src/wasi.rs
```diff
@@ -453,6 +459,12 @@ impl RequestSender {
             }
         }
 
+        record_content_length_header(
+            &span,
+            request.headers(),
+            "http.request.header.content-length",
+        );
+
```
ChihweiLHBird, line 467: I intentionally put this after interceptor, but I wasn't confident about this decision. I would like to hear more opinion on this.

calebschoepp: Assuming you've manually tested and values look correct this seems fine to me. Before after the interceptor shouldn't matter too much from my understanding.

calebschoepp (approved):
This change seems reasonable to me. LGTM

crates/factor-outbound-http/src/wasi.rs
ChihweiLHBird: Thank you!

PR #3462, opened 2026-04-16
spin-telemetry: use rustls for OTLP HTTP exporter

TheRayquaza:
Hey spin folks 👋
I recently run into the following error when trying to target a https OTEL endpoint:

```
2026-04-16T16:28:45.980443Z ERROR opentelemetry_sdk:  name="BatchSpanProcessor.Flush.ExportError" 
reason="InternalFailure(\"reqwest::Error { kind: Request, url: "https://....", source: 
hyper_util::client::legacy::Error(Connect, Ssl(Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 
167772294, library: "SSL routines", function: "tls_post_process_server_certificate", reason: "certificate verify 
failed", file: "ssl/statem/statem_clnt.c", line: 2124 }]))) }, X509VerifyResult { code: 19, error: "self-signed certificate 
in certificate chain" })) }\")" Failed during the export process
```
 
Tell me If I am wrong:

In minimal OCI containers (distroless, Alpine without ca-certificates), the Spin host process has no system CA bundle, which breaks TLS  verification for any component relying on the OS cert store.

Worth noting: `rustls-platform-verifier`, the current default for outbound HTTP, also reads the OS store so it probably has the same issue in distroless environments (though I haven't tested it 😄).                          

Would it make sense to make `webpki-roots` the default or at least an easy fallback ?


crates/telemetry/src/lib.rs
```diff
@@ -108,3 +108,12 @@ pub fn init(spin_version: String) -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Build a reqwest::Client that explicitly uses rustls as the TLS backend.
+///
```
calebschoepp, line 113:
```suggestion
```

calebschoepp (commented):
From an OTel perspective the changes look fine to me. 

Regarding rustls and cert stuff I'm out of my depth and would like some feedback from other maintainers with more knowledge in the area. My gut says that at the very least when there are OS certs we can rely on we should use them over rustls.

itowlson:
+ @lann as the TLS opinion-haver

itowlson:
@TheRayquaza Could you run `make lint` over this please?  The build is failing on some (very) minor Clippy warnings.  Thanks!

itowlson:
I'm afraid CI is still failing on the DCO check.  Looks like the penultimate commit might have come via the GitHub suggestion UI?  I am not if that applies signoff.  Sorry about this...

lann:
FWIW `rustls-platform-verifier` is meant to fall back to [`webpki-roots`](https://docs.rs/webpki-roots/latest/webpki_roots/) if the underlying OS has none configured.

lann:
I'd like to retain support for native roots. Luckily I think this is actually pretty easy, by enabling the `rustls-tls-native-roots` feature of `reqwest` ([here](https://github.com/spinframework/spin/blob/main/Cargo.toml#L170)); this should enable both native and webpki roots by default.

therayquaza:
Hey @lann thanks for reviewing this. Just added the `rustls-tls-native-roots`
@itowlson Also squashed and signed the commit

itowlson:
Thanks @TheRayquaza 

@lann if you're happy with this then could you give it a stamp of approval please? thanks

lann (approved):
Thanks!

PR #3463, opened 2026-04-16
feat(trigger-http): add `SPIN_HTTP_IDLE_INSTANCE_TIMEOUT_SECS` env var

TheRayquaza:
cc https://github.com/spinframework/containerd-shim-spin/pull/426 for the intended usage

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -134,7 +134,7 @@ pub struct CliArgs {
     /// This may be specified either as a single time value or as a range,
     /// e.g. 1..8s.  If it's a range, a value will be selected from that range
     /// at random for each new instance.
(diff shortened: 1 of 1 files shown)


itowlson:
@TheRayquaza @kate-goldenring Looking at the discussion on https://github.com/spinframework/containerd-shim-spin/pull/426, I'm not clear if this is still needed, or if the shim will handle it by setting the `CliArgs` struct fields from code.  (It feels a bit weird for the CLI to have to environment-variable-ify anything that SpinKube wants to set, but if that's where we are then so be it.)

If it _is_ still needed, do we think there should be corresponding environment variables for other instance reuse parameters?

TheRayquaza:
in fact not needed anymore 👌 closing

PR #3464, opened 2026-04-16
When WAGI converts string to HeaderValue, error rather than unwrap

itowlson:
Minimal impact since the request handling task is already closing in on completion at this point (and the task level panics are translated into join errors), but just a bit tidier.

### DIFF
--- a/crates/http/src/wagi/mod.rs
+++ b/crates/http/src/wagi/mod.rs
@@ -243,57 +243,54 @@ pub fn compose_response(stdout: &[u8]) -> Result<Response<Body>, Error> {
     let mut res = Response::new(body::full(buffer.into()));
     let mut sufficient_response = false;
     let mut explicit_status_code = false;
-    parse_cgi_headers(String::from_utf8(out_headers)?)
-        .iter()
-        .for_each(|h| {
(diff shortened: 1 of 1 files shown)


crates/http/src/wagi/mod.rs
```diff
@@ -243,57 +243,54 @@ pub fn compose_response(stdout: &[u8]) -> Result<Response<Body>, Error> {
...
-                    tracing::debug!(status_code, "Raw status code");
-                    match status_code.parse::<StatusCode>() {
-                        Ok(code) => *res.status_mut() = code,
-                        Err(e) => {
-                            tracing::warn!("Failed to parse code: {}", e);
-                            *res.status_mut() = StatusCode::BAD_GATEWAY;
-                        }
+    for h in parse_cgi_headers(String::from_utf8(out_headers)?).iter() {
+        use hyper::header::{CONTENT_TYPE, LOCATION};
+        match h.0.to_lowercase().as_str() {
```
itowlson, line 248: A lot of this diff is indent changes due to using a `for` loop instead of `for_each` combinator - not sure if there's a way to make it highlight the _actual_ changes...

fibonacci1729 (approved):

PR #3465, opened 2026-04-16
Use macro instead of checking bindings in

fibonacci1729:
Meant to do this before merging. Use the macro instead of pre-generating the bindings.


itowlson (approved):
Hah, and I meant to ask about this before merging.  What a team

Issue #3456, opened 2026-04-09
Switch to Rust's 2024 edition


dicej:
Currently Spin uses the 2021 edition, but the 2024 edition has been out for a while.

I'll admit this is, in part, selfishly motivated because my editor is set up to run `rustfmt` using the 2024 style by default, and it's annoying to have to reformat back to the 2021 style prior to committing.  Also, prominent dependencies like Wasmtime have been on 2024 for a while.

Actually doing this will be a large-ish diff that touches most of the `.rs` files due to the aforementioned formatting style changes, which is annoying, but we're going to want to do it eventually.

PR #3477, opened 2026-04-21
Update `schemars`

itowlson:
This emerged from looking into #3456.  Rust 2024 reserves the `gen` keyword, which was a module in `schemars` 0.8.  So to avoid reserved word woe in 2024, we need to bump schemars.  Is it just a few renames?  Reader, it is not.

I have diffed the JSON Schema documents produced by Spin main and Spin with this change, and there are some structural differences that make it hard to verify that the behaviour is the same.  But they look plausible.

To be clear, there is _no_ motivation to do this unless we are planning a move to Rust 2024.  This is just clearing some ground to make that less painful.

### DIFF
--- a/crates/serde/src/version.rs
+++ b/crates/serde/src/version.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 /// FixedVersion represents a version integer field with a const value.
 #[derive(Clone, Debug, Default, Serialize, Deserialize, JsonSchema)]
 #[serde(into = "usize", try_from = "usize")]
-#[schemars(with = "usize", range = (min = V, max = V))]
+#[schemars(with = "usize")]
 pub struct FixedVersion<const V: usize>;
 
 impl<const V: usize> From<FixedVersion<V>> for usize {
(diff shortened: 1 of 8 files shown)


itowlson:
Good news, the new schema crashes Even Better TOML.  Well back to the detail reading I guess.

_ETA:_ but, maddeningly, the Newtonsoft validator is fine with it

itowlson:
You will be astonished to learn that PEBKAC.

This seems to be working in Even Better TOML now.

crates/serde/src/version.rs
```diff
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 /// FixedVersion represents a version integer field with a const value.
 #[derive(Clone, Debug, Default, Serialize, Deserialize, JsonSchema)]
 #[serde(into = "usize", try_from = "usize")]
-#[schemars(with = "usize", range = (min = V, max = V))]
+// #[schemars(with = "usize", range(min = V, max = V))]
```
dicej: Why is this attribute (and the one on `FixedVersionBackwardCompatible` below) commented out?

itowlson: Because they don't compile and I forgot to delete them.  🤦  Thanks for catching that!

dicej (approved):

fibonacci1729 (approved):

PR #3478, opened 2026-04-22
Update to Rust 2024

### LINKED ISSUE #3456: Switch to Rust's 2024 edition

itowlson:
Futurama welcome to the world of tomorrow

This incorporates a functional update from #3477.  We can merge that as part of this PR if folks prefer, but I'd value a review on #3477 because it makes non-mechanical changes (whereas the changes here are essentially mechanical).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1505,12 +1505,12 @@ dependencies = [
 
 [[package]]
 name = "cargo_toml"
-version = "0.17.2"
+version = "0.22.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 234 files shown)


fibonacci1729 (approved):

dicej (approved):
🎉

PR #3480, opened 2026-04-23
Fix test for `spin up` JSON output

### LINKED ISSUE #3437: Flake in `integration_tests::http_smoke_test_json_format`
```
Error: Test failed to run: Actual output:

  Caused by: stdout was not valid JSON
  Caused by: EOF while parsing a value at line 1 column 0
```

Not sure what's going on here.  I've not seen it fail locally (although I don't run the integration tests very often locally).  In CI, it mostly passes, but I've seen it fail this way a couple of times now.  It doesn't seem to be a blocker for now, but capturing it to investigate.

itowlson:
Fixes #3437... I hope.

This fix is still a bit heuristic.  The test tries to read stdout once the Spin listener port opens.  This, I think, causes a race, because the output is also printed to stdout once the Spin listener port opens.  When the read wins, stdout is empty.  My fix is to wait until stdout is non-empty (by polling it until it is non-empty).  This potentially still has a race if output is partially written when a poll occurs, but the window is hopefully a bit smaller.

I was never able to reproduce the issue locally, so there's no proof this will make things better - maybe I've misunderstood the nature of the flake, or the "partially written" race is just as cruel as the "empty" race.  But this seemed like a minimal-investment first try!

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -108,7 +108,7 @@ mod integration_tests {
             },
             ServicesConfig::none(),
             move |env| {
-                let stdout = env.runtime_mut().stdout().to_owned();
+                let stdout = env.runtime_mut().wait_for_non_empty_stdout()?.to_owned();
 
                 let check_json = || -> anyhow::Result<()> {
                     let parsed: serde_json::Value =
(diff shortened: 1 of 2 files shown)


fibonacci1729 (approved):
Lets give it a go!

PR #3482, opened 2026-04-24
Set blame on span for postgres and mysql

rylev:
This sets the `blame` otel field to allow consumers of the otel data to differentiate between errors blamed on the host vs those blamed on the guest. 

My motivation here is for allowing my monitoring set up to alert when I see a spike in host errors but ignore any spike in guest errors.


lann (approved):
LGTM but I'm a little surprised this actually works - I'd expect the error to bubble up and end up _also_ being emitted on outer spans but without the blame annotation. 🤷

rylev:
@lann It works because `mark_as_error` sets the status on `tracing::Span::current()` at the exact point the error is constructed while still executing inside the instrumented span (e.g. the execute or query span). By the time the error propagates to an outer span, we've already annotated the right span.

PR #3486, opened 2026-04-27
Add http-rust-p2 compatibility template

fibonacci1729:


templates/http-rust-p2/metadata/spin-template.toml
```diff
@@ -0,0 +1,13 @@
+manifest_version = "1"
+id = "http-rust"
+description = "WASI Preview 2 HTTP request handler using Rust"
```
itowlson:
```suggestion
description = "HTTP request handler compatible with Spin 3.x hosts"
```

or "maximising platform compatibility" something - but to help answer "when/why would I use this instead of the normal one"

itowlson (approved):
Thanks!  The content LGTM but I had a non-blocking suggestion on the description.

itowlson:
Oh you will need to update `TPLS_IN_THIS` in the templates crate unit tests (it asserts that the right number of things have been installed and so numbers need to line up)

templates/http-rust-p2/metadata/spin-template.toml
```diff
@@ -0,0 +1,13 @@
+manifest_version = "1"
+id = "http-rust"
```
itowlson: Oh this needs to be `http-rust-p2`.  Sorry, I always forget the ID isn't automatically the same as the directory!

itowlson (approved):

PR #3488, opened 2026-04-29
reorg http-rust p2 & p3 templates

fibonacci1729:
Renames `http-rust-p2` -> `http-rust`
Adds `http-rust-p3`


itowlson:
This makes me sad.  If we have to make P3 non-default then should we have `http-rust-p2` and `http-rust-p3` i.e. no default rather than blessing P2 as the default?  But what a miserable experience for users having to know about this stuff (and gosh, what a shambles it's going to make of the documentation).

PR #3490, opened 2026-04-30
Upgrading templates should remove ones that no longer exist

itowlson:
~We haven't run into this situation yet as far as I know.  But...~ Uh, Ivan. We totally run into it with the old `http-rust-wasip3-unstable` template!

~(WIP.)~

I tested this by installing the 3.6 templates, then installing the main templates, and verifying that `http-rust-wasip3-unstable` was reported as removed (and no longer appeared in `spin templates list`). Then went back to 3.6 and `http-rust-p2` was correctly removed (and p3-unstable correctly reinstated).  Then did `spin templates upgrade` and verified again.

However, this is all "works on my machine."  Now, the feature applies only to remote (git) template stores (we don't really have origin tracking / upgrading for local directories), And we don't want to have the templates unit tests reaching out to GitHub.  So unit testing is likely ruled out.  I am not sure whether it is worth writing it as an integration test?

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -68,6 +68,8 @@ pub enum SkippedReason {
     AlreadyExists,
     /// The template was skipped because its manifest was missing or invalid.
     InvalidManifest(String),
+    /// The template was removed from the source but could not be removed locally.
+    CouldNotRemove,
 }
 
 /// The results of installing a set of templates.
@@ -76,6 +78,8 @@ pub struct InstallationResults {
     pub installed: Vec<Template>,
     /// The templates that were skipped during the install operation.
     pub skipped: Vec<(String, SkippedReason)>,
+    /// The templates that were removed during the install operation.
+    pub removed: Vec<String>,
 }
 
 /// The result of listing templates.
@@ -139,6 +143,21 @@ impl TemplateManager {
         let mut installed = vec![];
         let mut skipped = vec![];
 
+        let mut local_but_not_source = if let Some(upgrading_repo) = source.as_git_url() {
+            let local = self
+                .list()
+                .await
+                .map(|list| list.templates)
+                .unwrap_or_default();
+            local
+                .into_iter()
+                .filter(|t| t.is_from_source_repo(upgrading_repo))
+                .map(|t| t.id().to_string())
+                .collect()
+        } else {
+            std::collections::HashSet::new()
+        };
+
         for template_dir in template_dirs {
             let install_result = self
                 .install_one(&template_dir, options, source, reporter)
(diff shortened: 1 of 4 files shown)


itowlson:
@fibonacci1729 If this lands before the Spin 4 launch, it may be worth backporting, so that users upgrading from Spin 3.5/3.6 no longer see the 'unstable' template.  I appreciate this is additional release overhead though.

crates/templates/src/manager.rs
```diff
@@ -147,15 +166,37 @@ impl TemplateManager {
...
+            }
+        }
+
+        let mut removed = match &options.exists_behaviour {
+            ExistsBehaviour::Skip => vec![],
+            ExistsBehaviour::Update => local_but_not_source.into_iter().collect::<Vec<_>>(),
+        };
+
+        for id in &removed {
+            if self.uninstall(id).await.is_err() {
```
fibonacci1729: It's a little weird that if uninstall fails here the id is kept in both skipped and removed.

itowlson: Bah, that's me being an idiot.  Will fix - thanks!

itowlson: Okay, this is hard to test for because it's difficult to force an uninstall failure, but I've updated it and think it's right now.  And it made the naming better in the process so that's a bonus win.

src/commands/templates.rs
```diff
@@ -163,6 +163,7 @@ impl Install {
     fn print_installed_templates(&self, installation_results: &InstallationResults) {
         let templates = &installation_results.installed;
         let skipped = &installation_results.skipped;
+        let removed = &installation_results.removed;
 
         if templates.is_empty() && skipped.is_empty() {
```
fibonacci1729, line 168: Adding `&& removed.is_empty()` would make this a little more robust.

src/commands/templates.rs
```diff
@@ -377,6 +397,25 @@ impl Upgrade {
 
         println!();
 
+        if !removed.is_empty() {
+            println!(
+                "Removed {} templates no longer in repository/ies",
```
fibonacci1729:
```suggestion
                "Removed {} template(s) no longer in repository/ies",
```

Consistency with other messages elsewhere here.

src/commands/templates.rs
```diff
@@ -611,6 +655,7 @@ fn skipped_reason_text(reason: &SkippedReason) -> String {
     match reason {
         SkippedReason::AlreadyExists => "Already exists".to_owned(),
         SkippedReason::InvalidManifest(msg) => format!("Template load error: {msg}"),
+        SkippedReason::CouldNotRemove => "Deleted but could not remove local copy".to_owned(),
```
fibonacci1729: nit: I think `Removed from source but could not delete locally` is a little more clear.

src/commands/templates.rs
itowlson: I am not sure that is correct.  If for some reason a repo used to have templates in it and now doesn't, and I upgrade there, I should probably say "hey that repo didn't have any templates in it" as well as "I removed foo and bar."  But I'm open to being persuaded!

fibonacci1729: If a re-install of an already-fully-up-to-date repo somehow yields only removals (edge case, but possible if every still-existing template was filtered/skipped silently; currently they end up in skipped, so probably safe today), the removals would be hidden and the user would see "contained no templates". It's cheap to make robust. But probably a nit.

src/commands/templates.rs
itowlson: I changed it to `No longer exists in source, but could not uninstall`

crates/templates/src/template.rs
```diff
@@ -257,6 +257,11 @@ impl Template {
         }
     }
 
+    pub(crate) fn is_from_source_repo(&self, source_repo: &url::Url) -> bool {
+        self.source_repo()
+            .is_some_and(|r| r == source_repo.as_str())
```
fibonacci1729, line 262: This feels a little brittle. What happens if i installed with `https://github.com/foo/bar` and then re-run with `https://github.com/foo/bar.git`?

itowlson: Then you will be in no worse a position than you were before this PR. _grin_  I am not sure how much value there is in trying to get across all the ways one can refer to the same Git repo (like... there's `ssh` and `git@` stuff and all sorts isn't there?) - for something non-critical like this I we rather just do the 90% case, and consider patching it up if folks run into the cases.

fibonacci1729 (approved):

PR #3491, opened 2026-04-30
outbound-networking: Turn `use_webpki_roots` back on by default

lann:
https://github.com/spinframework/spin/pull/3426 added support for `rustls-platform-verifier` as `use_platform_roots` and enabled it by default, but it also disabled `use_webpki_roots` by default causing a regression in `containerd-shim-spin` (and potentially other consumers).

- Changes the default to enable both `use_platform_roots` _and_ `use_webpki_roots`.
- Switch from `webpki-roots` to `webpki-root-certs`, necessary to make enabling both options work correctly.
- Simplify some logic - enabled by the crate change above.
- Improve `ClientTlsRuntimeConfig` doc comments.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8867,7 +8867,7 @@ dependencies = [
  "tracing",
  "url",
  "wasmtime-wasi",
- "webpki-roots",
+ "webpki-root-certs",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


vdice (commented):
This makes sense to me, but I know @kate-goldenring has been in this area in more depth than I have.

Just a quick note to mention that it looks like `tls::tests::test_minimal_config` needs to be updated/adjusted to pass.

kate-goldenring (approved):

PR #3493, opened 2026-05-06
Update go.sum and Redis build command

itowlson:
@fibonacci1729 @adamreese these changes made the Go templates build and run locally for me (although I don't think CI will run for them because I'm PRing to your branch)


fibonacci1729 (approved):

adamreese:
Good catch on the go.sum. That wasn't a very helpful error

PR #3495, opened 2026-05-06
Testing Go P3 CI failure

itowlson:
Trying this to main rather than to Brian's PR so that I can get the integration test result


fibonacci1729 (approved):
If this works, feel free to supersede my PR!

itowlson:
Okay it looks like this works, BUT the changes to Go version mean the template on main (TinyGo) no longer works.  We consider this a test failure for Spin runtime changes, but in this case it's ignorable because the template we broke is exactly the one we're replacing.  My evidence for this is that if I comment out the "test template on main" cases, it passes.  So I am uncommenting them, which will cause CI to fail, and then I'll force merge because I believe the commented-out case is evidence that we will be in a good place after.  Fingers crossed.

@fibonacci1729 @adamreese I'm not planning to follow this deep into the evening, so if you come in tomorrow morning and main is broken then please accept my apologies!

PR #3496, opened 2026-05-06
fix(outbound-http): ensure tracing span records land on send_request span for P3

ChihweiLHBird:
## Summary

Fix a tracing span propagation bug in the WASI Preview 3 outbound HTTP path.

`p3::WasiHttpHooks::send_request` returns a boxed future. The `#[instrument(name = "spin_outbound_http.send_request")]` span exits immediately, so later `Span::record(...)` calls (e.g. `url.full`, `server.address`) inside the async work were lost.

Wrap the async block with `.in_current_span()` so the instrumented span remains current when the future is polled. This matches the P2 implementation.

```bash
cargo test -p spin-factor-outbound-http p3_send_request_propagates_span_to_async_work
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8821,6 +8821,7 @@ dependencies = [
  "tower-service",
  "tracing",
  "tracing-opentelemetry",
+ "tracing-subscriber",
  "wasmtime",
  "wasmtime-wasi",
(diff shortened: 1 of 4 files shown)


calebschoepp (approved):
LGTM

PR #3497, opened 2026-05-06
fix(templates): Update build command for Go component

adamreese:
@fibonacci1729 @itowlson Looks like we missed this one.


fibonacci1729 (approved):

itowlson:
Bother.  Sorry about that @adamreese - thanks for fixing it.

PR #3499, opened 2026-05-06
Fix incorrect URL being shown to user during just-in-time plugin install

### LINKED ISSUE #3498: plugins: offer install URL may be incorrect
Unsure if this is a spin or spin-plugins issue, but the URL printed during the installation offer for a missing plugin doesn't always match what is fetched.  For example, seen here using spin version `spin 4.1.0-pre0 (ec18b602 2026-05-05)` and fetching the `aka` and `cloud` plugins:

- `aka-0.4.0` is mentioned, but latest installed.  Latest is good - would expect the `aka-0.7.1` URL to be printed
```
$ spin aka -V
Warning: `aka` is not a known Spin command, but there is a plugin with that name.
The plugin has the Apache-2.0 license and would download from https://s3.amazonaws.com/fermyon-neutrino-artifacts-public/plugin/aka-0.4.0-macos-aarch64.tar.gz
Would you like to install this plugin and run it now? yes
Plugin 'aka' was installed successfully!

Description:
        Akamai Functions plug-in.

Homepage:
        https://developer.fermyon.com/wasm-functions/index

spin-aka 0.7.1 (cec416c 2026-05-06)
```

- `cloud-v0.2.0` is mentioned, but latest installed.  Latest is good - would expect the `cloud-v0.11.0` URL to be printed
```
$ spin cloud -V
Warning: `cloud` is not a known Spin command, but there is a plugin with that name.
The plugin has the Apache-2.0 license and would download from https://github.com/fermyon/cloud-plugin/releases/download/v0.2.0/cloud-v0.2.0-macos-aarch64.tar.gz
Would you like to install this plugin and run it now? yes
Plugin 'cloud' was installed successfully!

Description:
        Commands for publishing applications to the Fermyon Cloud.

Homepage:
        https://github.com/fermyon/cloud-plugin

spin-cloud 0.11.0 (c86d7c8 2025-03-26)
```

itowlson:
Fixes #3498 

I wasn't able to repro the original bug report, but I was able to simulate it by reversing the order of sorting, so I'm optimistic this is correct.  @vdice If you still have your problematic environment available then it would be awesome if you could test this there - thanks!

### DIFF
--- a/src/commands/external.rs
+++ b/src/commands/external.rs
@@ -214,12 +214,20 @@ fn match_catalogue_plugin(
     plugin_store: &PluginStore,
     plugin_name: &str,
 ) -> Option<spin_plugins::manifest::PluginManifest> {
+    use itertools::Itertools;
+
     let Ok(known) = plugin_store.catalogue_manifests() else {
         return None;
     };
-    known
+    let candidates = known
         .into_iter()
-        .find(|m| m.name() == plugin_name && m.has_compatible_package())
+        .filter(|m| m.name() == plugin_name && m.has_compatible_package());
+
+    // We care about getting the latest because it's the return of this that is used
+    // to show the plugin installation offer to the user (e.g. URL).
+    let mut candidates_latest_first =
+        candidates.sorted_by(|a, b| b.compare_versions(a).unwrap_or(std::cmp::Ordering::Equal));
+    candidates_latest_first.next()
 }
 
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

vdice:
Working great on a fresh Ubuntu VM as well as my Mac!  Thanks @itowlson.

PR #3500, opened 2026-05-11
open issues in js and python sdk repos to tag templates on new releases

karthik2804:
This PR adds a job to release CI that opens issues in the Python and JS SDK repos to remind us to tag templates whenever there is a release of Spin. We have run into some issues with the 2 repos where the templates have not been tagged for the latest release and which leads to the templating subsystem pulling from `main` which may be in a non working state.

This is an attempt to at least connect the release flow to the SDK repos but I am definitely open to better alternative ideas.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -386,6 +386,51 @@ jobs:
           event-type: spin-release
           client-payload: '{"version": "${{ github.ref_name }}"}'
 
+  notify-template-repos:
+    name: Open issues in template repos for new release
+    needs: create-gh-release
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'spinframework' && startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - name: Check if this is at least a minor release
+        id: check-minor
+        shell: bash
+        run: |
+          TAG="${{ github.ref_name }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -386,6 +386,51 @@ jobs:
...
+            fi
+          else
+            echo "is_minor=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Open issues in template repos
+        if: steps.check-minor.outputs.is_minor == 'true'
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.DEST_REPO_ACCESS_TOKEN }}
```
karthik2804, line 417: If we do decide to go with this PR, we would either need to create a new token with fine grained permissions for the 2 repos or expand the access of this token.

vdice: We should be fine to update the access for this token to include both sdk repos.

vdice (approved):
Nice, this will be helpful.

karthik2804:
I do not have merge rights in this repo. so if all looks good and the token permissions are updated, it would be great to get this merged.

vdice:
@karthik2804 ok, token/perms should now be updated.  Merge when ready -- though we wouldn't see this in action until the next minor Spin release, eh?

Issue #3494, opened 2026-05-06
Target environments: suggest templates and plugins


itowlson:
This follows on from conversations with @fibonacci1729 and @tschneidereit (who came up with more of these ideas than I did!).

Right now the target environments system is the proverbial ambulance at the bottom of the cliff - it can tell you that your code won't run where you want to run it, but that's all.

But if we centred target environments in the application creation workflow, they could act as a hub for the stuff you need to successfully build for that environment.  The specific ideas we've discussed are:

* **A target environment could point to a set of templates** that are known to work in that environment.  For example, suppose that Bob's Discount Cloud supports only WAGI.  Then running `spin new -E bdc@latest` would offer you a selection of sparkling WAGI templates with none of your new-fangled component model nonsense.

* **A target environment could contain a list of plugins** that are required to work with the environment.  For example, suppose you need the `bobcloud` plugin to deploy to or manage applications on Bob's Discount Cloud.  Then `spin new -E bdc@latest` could check if the `bobcloud` plugin is installed, and if not install it from the plugins registry.

(Note that the latter is a significant and possibly surprising new behaviour of the `spin new` command.  `spin new` has previously confined its antics to the application directory: now it would be installing software on your machine.  We would want to ensure that the user was clearly informed and felt that they had consented.)

This would set the user up so that 1. the application was guaranteed to work in the environment and 2. they had the tools necessary to get it up and running there.

I did a POC if the 'target environment templates' idea and it appeared to work, but it was only a POC - there are several questions left to explore.  The plugins idea should be fairly simple provided we have a good solution to the informed consent aspect.

fibonacci1729:
Thanks for capturing all of this @itowlson!

Just to capture this, from our offline sync earlier, that a design consideration we would need to address is how to handle templates when multiple target environments are used.

tschneidereit:
> * **A target environment could contain a list of plugins** that are required to work with the environment.  For example, suppose you need the `bobcloud` plugin to deploy to or manage applications on Bob's Discount Cloud.  Then `spin new -E bdc@latest` could check if the `bobcloud` plugin is installed, and if not install it from the plugins registry.

One way to address the concerns around surprising behavior is to not auto-install plugins. Instead, `spin new -E [env]` could print something like
> Use `spin plugin list - E [env]` to see plugins for the `[env]` environment

or alternatively, assuming that most environment definitions would only list a single plugin, if that single plugin isn't already installed, it could print something like
[...]

PR #3505, opened 2026-05-12
Plugin manager refactoring

itowlson:
This emerged from prototyping on #3494.  The original plugin code was fairly simple and well structured, but over time it has accreted a bunch of stuff that bypasses abstractions and entangles concerns, to the point where I was struggling to figure out how and where to add some stuff I needed.

This PR aims to restore a division of responsibilities between:

* The Manager - the thing that applications interact with to list, install, remove, and execute plugins without needing to know internal implementation details
* The Store - the thing that knows how to lay files out on disk
* The Catalogue (new) - the snapshot of the remote `spin-plugins` repo - knows about available plugins and updating (although updating remains a slightly fuzzy responsibility, but that's not a huge concern for me)

I have made the `store` module private, so that application code is not tempted to perform logic directly against the disk layout: the manager should surface all that at a layer of abstraction that insulates the app from knowledge of JSON and directories and related nibbly bits.

I also did some minor renaming, and added some doc comments where I was struggling with names.

I don't imagine I've eliminated all the tangly bits, in fact I know some bits that still feel a bit of a lash-up (I think largely because of the badger), but for me this surfaces a clearer API to consumers of the crate.  Feedback is, of course, welcome!

### DIFF
--- a/crates/plugins/src/catalogue.rs
+++ b/crates/plugins/src/catalogue.rs
@@ -1,0 +1,108 @@
+use crate::{git::GitSource, manifest::PluginManifest};
+use anyhow::Context;
+use semver::Version;
+use std::path::{Path, PathBuf};
+use url::Url;
+
+const SPIN_PLUGINS_REPO: &str = "https://github.com/spinframework/spin-plugins/";
+
+pub(crate) fn plugins_repo_url() -> Result<Url, url::ParseError> {
+    Url::parse(SPIN_PLUGINS_REPO)
+}
+
+/// The local clone of the spin-plugins repo.
+pub struct Catalogue {
+    git_root: PathBuf,
+    manifests_root: PathBuf,
+}
+
+// Name of directory containing the installed manifests
+const LOCAL_CATALOGUE_MANIFESTS_DIRECTORY: &str = "manifests";
+
+impl Catalogue {
+    pub fn new(git_root: PathBuf) -> Self {
+        let manifests_root = git_root.join(LOCAL_CATALOGUE_MANIFESTS_DIRECTORY);
+        Self {
+            git_root,
+            manifests_root,
+        }
+    }
+
+    pub fn manifests(&self) -> anyhow::Result<Vec<PluginManifest>> {
+        // Structure:
+        // CATALOGUE_DIR (spin/plugins/.spin-plugins/manifests)
+        // |- foo
+        // |  |- foo@0.1.2.json
+        // |  |- foo@1.2.3.json
+        // |  |- foo.json
+        // |- bar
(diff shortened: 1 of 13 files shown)


crates/plugins/src/catalogue.rs
```diff
@@ -0,0 +1,107 @@
...
+        let catalogue_manifests_dir = &self.manifests_root;
+
+        // Catalogue directory doesn't exist so likely nothing has been installed.
+        if !catalogue_manifests_dir.exists() {
+            return Ok(Vec::new());
+        }
+
+        let plugin_dirs = catalogue_manifests_dir
+            .read_dir()
+            .context("reading manifest catalogue at {catalogue_dir:?}")?
```
fibonacci1729:
```suggestion
            .context(format!("reading manifest catalogue at {catalogue_dir:?}"))?
```

crates/plugins/src/catalogue.rs
```diff
@@ -0,0 +1,107 @@
...
+    git_root: PathBuf,
+    manifests_root: PathBuf,
+}
+
+// Name of directory containing the installed manifests
+const LOCAL_CATALOGUE_MANIFESTS_DIRECTORY: &str = "manifests";
+
+impl Catalogue {
+    pub fn new(git_root: PathBuf) -> Self {
+        Self {
```
fibonacci1729, line 25: nit: unnecessary clone

```
let manifests_root = git_root.join(LOCAL_CATALOGUE_MANIFESTS_DIRECTORY);
Self { git_root, manifests_root }
```

crates/plugins/src/manifest.rs
```diff
@@ -57,11 +55,14 @@ impl PluginManifest {
...
         self.packages.iter().any(|p| p.matches_current_os_arch())
     }
+
     pub fn is_compatible_spin_version(&self, spin_version: &str) -> bool {
         is_version_compatible_enough(&self.spin_compatibility, spin_version).unwrap_or(false)
     }
-    pub fn is_installed_in(&self, store: &PluginStore) -> bool {
-        match store.read_plugin_manifest(&self.name) {
+
+    // TODO: this should be a PluginManager function I'm pretty sure
```
fibonacci1729: Agreed

crates/plugins/src/catalogue.rs
itowlson: _facepalm_

fibonacci1729 (approved):

itowlson:
CI failure is due to an unrelated change in the JS SDK (https://github.com/spinframework/spin-js-sdk/issues/506) so I will bypass.

PR #3506, opened 2026-05-12
feat(factor-outbound-http): report `http.response.body.size` for WASI p3 responses

### LINKED ISSUE #3188: Include body sizes in outbound HTTP telemetry
This would be useful for egress bandwidth accounting.

The easy version of this would be to just report body sizes (e.g. https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/#http-request-body-size) based off of `content-length` headers, which would work for "most" requests. The harder but more complete version would be to use custom `Body` wrapper(s) to record the actual bytes written, which would additionally work for streaming/chunked-encoding bodies that have no content-length.

ChihweiLHBird:
## Description
Adds the stable OpenTelemetry semantic convention `http.response.body.size` to the `spin_outbound_http.send_request` span for outbound HTTP calls made through the WASI p3 interface.
### Key points
- The attribute records the **actual** number of response body bytes delivered to the guest (counted frame-by-frame during streaming), not a `Content-Length` guess.
- Clean end-of-stream (`Ready(None)` or `is_end_stream()`) triggers the recording exactly once via a deferred `Span::record` on the instrumented span.
- On body read errors the size field is intentionally left unset (to keep the metric untainted) and a warning is emitted with `tracing::warn!()`.

Partially resolve https://github.com/spinframework/spin/issues/3188 (Option 2).

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -10,7 +10,7 @@ use std::{
     time::Duration,
 };
 
-use bytes::Bytes;
+use bytes::{Buf, Bytes};
 use futures::channel::oneshot;
 use http::{
     HeaderMap, Uri,
@@ -138,6 +138,7 @@ impl p3::WasiHttpHooks for InstanceHttpHooks {
             url.full = Empty,
             http.request.method = %request.method(),
             otel.name = %request.method(),
+            http.response.body.size = Empty,
             http.response.status_code = Empty,
             server.address = Empty,
             server.port = Empty,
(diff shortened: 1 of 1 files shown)


calebschoepp (commented):
Seems fine from an OTel perspective. Maybe @dicej can comment from a P3/HTTP perspective.

dicej (approved):
LGTM, thanks.

This is a good reminder that we need more test coverage for this kind of thing.  I see there is a basic `otel_smoke_test` in `tests/integration.rs`, but it doesn't currently do much.  Expanding that test to check for e.g. `http.response.body.size` and make it run both p2 and p3 components seems appropriate.

We might also be able to unit test this kind of thing using e.g. https://crates.io/crates/tracing-fluent-assertions, although I don't personally have experience with that.

Issue #3504, opened 2026-05-11
Migrate target environment TOMLs to a git repo


itowlson:
Spin currently by default looks for target environment definition TOMLs in a registry (`ghcr.io/spinframework/envs` or something like that).  This means downstream hosts with their own environment definitions need to either sweet-talk the project into hosting their TOMLs, or contend with long registry references pointing to their own registry.

Sweet talk is hard to automate, and with the proposal for `spin new -E` (#3494), long registry references become unattractive.

We should therefore consider migrating target environments to a Git repo.  Downstreams could then PR into that repo.  It still involves a step on the Spin project side to approve and merge PRs, but it's a process which is much simpler, smoother, and more familiar than getting stuff into the project's GHCR.  The model here is plugins, where the project hosts a Git repo of JSON manifests which vendors and ICs can PR into.

itowlson:
I created https://github.com/spinframework/spin-environments as a first cut of this.

PR #3508, opened 2026-05-13
Host target environment TOMLs in git instead of registry

itowlson:
This is the underlying code for #3504.

The corresponding Git repo is https://github.com/spinframework/spin-environments.

bikeshed go brrr


itowlson:
I know this needs work on error messages and stuff but I want to agree the approach and iron out any other concerns before I invest too much time in those!

crates/manifest/src/schema/v2.rs
```diff
@@ -739,15 +739,12 @@ impl ComponentDependencies {
...
-        registry: String,
-        /// Environment definition document name e.g. `my-spin-env:1.2`. For hosted environments
-        /// where you always want `latest`, omit the version tag e.g. `my-host`.
-        id: String,
+    /// in the default environment catalogue (the `spin-environments` repo, `env` directory).
+    Catalogue(String),
+    /// An environment definition doc HTTP URL
+    Http {
+        /// The environment document URL e.g. `https://github.com/me/environments/blob/main/target-envs/spin-up.3.6.toml`.
+        url: String,
```
karthik2804, line 747: This is a breaking change right?

itowlson: Yes, explicit registry refs will break.  I can either put them back or we can roll this into a 4.0.1 and pretend it's part of the 4.0 family of breaking changes.

Default registry refs will NOT break because all existing default registry envs will be in the repo, and I translate syntax for compatibility.

fibonacci1729 (approved):
Looks good!

PR #3509, opened 2026-05-14
`spin new -E` and `spin plugins -E`

### LINKED ISSUE #3494: Target environments: suggest templates and plugins

itowlson:
Fixes https://github.com/spinframework/spin/issues/3494

This adds:

* `spin new -E` - looks for the template in the repo pointed to by the given environment
* `spin plugins install -E` - install the plugins suggested in the given environment, or the environment for the current app if `-E` with no env name
* `spin plugins list -E` - list the plugins suggested in the given environment (again default to current app)

Testing things I have to remove:

* ~The fake local env file~ Done
* ~Interpreting the `-E` as a file ref instead of a catalogue ref~ Now defaults to catalogue ref. Use `https:` prefix to interpret as URL, `file:` prefix for file path

Need:

* ~A way for env authors to test against a local file or a GH fork~ `-E file:foo.toml` / `-E https://example.com/foo.toml`
* ~A way to view and install the env plugins if you don't want to do so at `spin new` time~ Done, at least I think it's done, my brain is fried
* ~Define `--accept-defaults` behaviour when plugins are in play~ `--accept-defaults` will NOT suppress the plugin install prompt
* @kate-goldenring has reported seeing it offering to install already-installed plugins but I cannot reproduce this (and cannot see a code path that could open it up)


itowlson:
Sorry I thought I had clippied this.  I have stuff going on which mean I have uncommitted stuff in my working copy, so I will fix the lints in the morning along with those.

itowlson:
For my testing I have used https://raw.githubusercontent.com/itowlson/spin/4f12a8b60dd5d66ed79df89e50fddf00f18519c3/target-envs/env-mc-envface.toml (note this is removed from the PR so you will need to address it at that specific commit).

src/commands/plugins.rs
```diff
@@ -88,9 +93,18 @@ pub struct Install {
         long = "url",
         conflicts_with = PLUGIN_LOCAL_PLUGIN_MANIFEST_OPT,
         conflicts_with = PLUGIN_NAME_OPT,
+        conflicts_with = PLUGIN_TARGET_ENV_OPT,
     )]
     pub remote_manifest_src: Option<Url>,
 
+    /// The Spin platform or runtime for which you want to list plugins.
```
fibonacci1729:
```suggestion
    /// The Spin platform or runtime for which you want to install plugins.
```

src/commands/plugins.rs
```diff
@@ -157,6 +178,133 @@ impl Install {
...
+                .is_ok();
+            if !did_install && show_manual_help {
+                eprintln!(
+                    "Plugin `{plugin}` installation failed. You can try manually using `spin plugins install -E`."
+                );
+            }
+        }
+        if chosen.len() < plugins.len() && show_manual_help {
+            eprintln!(
+                "You can review and install unchosen recommanded plugins using `spin plugins` with the `-E` option"
```
fibonacci1729:
```suggestion
                "You can review and install unchosen recommended plugins using `spin plugins` with the `-E` option"
```

src/commands/plugins.rs
```diff
@@ -607,6 +769,10 @@ pub struct List {
     /// The format in which to list the plugins.
     #[clap(value_enum, long, default_value_t = ListFormat::default())]
     pub format: ListFormat,
+
+    /// The Spin platform or runtime for which you want to list plugins.
+    #[clap(long, short = 'E', num_args = 0..=1, default_value(FLAG_NOT_PRESENT), default_missing_value(FLAG_PRESENT_BUT_NO_VALUE), group = "which")]
+    target_environment: String,
```
fibonacci1729, line 775: What happens if I `spin plugins list --installed -E`? It looks like `-E` is ignored. We should either take the intersection or toss a `conflicts_with` in here.

crates/environments/src/environment/env_loader.rs
```diff
@@ -72,31 +73,34 @@ async fn load_environment<'a>(
...
+    let (name, env_def, relative_to_dir) = load_environment_def(env_id, app_dir).await?;
+    let env =
+        load_environment_from_env_def(&name, env_def, &relative_to_dir, cache, lockfile).await?;
+    Ok((env_id, env))
+}
+
+pub async fn load_environment_def(
+    env_id: &TargetEnvironmentRef,
+    app_dir: &Path,
+) -> Result<(String, EnvironmentDefinition, Option<PathBuf>), anyhow::Error> {
```
fibonacci1729: Could we return a named struct for this tuple?

crates/plugins/src/manager.rs
```diff
@@ -130,6 +130,24 @@ impl PluginManager {
...
+    /// This is roughly equivalent to `spin plugins install <name>` with no options.
+    pub async fn install_latest(&self, name: &str, spin_version: &str) -> anyhow::Result<String> {
+        let manifest_location = ManifestLocation::PluginsRepository(PluginRef {
+            name: name.to_string(),
+            version: None,
+        });
+        let plugin_manifest = self
+            .get_manifest(&manifest_location, false, spin_version, &None)
+            .await?;
+        let plugin_package = plugin_manifest.get_package().unwrap();
```
fibonacci1729: Is this `unwrap()` safe? If so, mind leaving a comment.

src/commands/plugins.rs
itowlson: `group = "which"` handles that - same as `--installed --all`

crates/environments/src/environment/env_loader.rs
itowlson: Curse you for telling me to do the thing I knew I ought to do but was too lazy to both-- I mean why yes, an excellent idea.

crates/plugins/src/manager.rs
itowlson: I don't think it is!  Fixed, thanks!

src/commands/plugins.rs
fibonacci1729: ~It seems like down below if `-E` and `--installed` are both set, we only show the `--installed` which is a bit confusing.~

Nevermind! My goof.

fibonacci1729 (approved):
Thanks for this! Works great for me.

PR #3510, opened 2026-05-14
Adopt `opentelemetry-semantic-conventions` for OTel attribute names

ChihweiLHBird:
## Description

Adopt `opentelemetry-semantic-conventions` 0.28 for stable runtime span attribute names on outbound HTTP (`factor-outbound-http`) and the HTTP trigger (`trigger-http`), replacing string-literal field keys with crate constants where available.

Bump workspace `tracing` from 0.1.41 to 0.1.44 so `#[instrument(fields(...))]` accepts braced semconv constants (older `tracing-attributes` only parsed dotted field names there).

`otel.kind`, `otel.name`, and incubating `http.response.body.size` stay literals (with a short comment on body size). Spin-specific attributes (e.g. `error.blame`) are unchanged.

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -26,9 +26,7 @@ use hyper_util::{
     },
     rt::{TokioExecutor, TokioIo},
 };
-use opentelemetry_semantic_conventions::attribute::{
-    HTTP_REQUEST_METHOD, HTTP_RESPONSE_STATUS_CODE, SERVER_ADDRESS, SERVER_PORT, URL_FULL,
-};
+use opentelemetry_semantic_conventions::attribute as otel_attribute;
 use spin_factor_outbound_networking::{
     ComponentTlsClientConfigs, TlsClientConfig,
     config::{allowed_hosts::OutboundAllowedHosts, blocked_networks::BlockedNetworks},
@@ -138,14 +136,14 @@ impl p3::WasiHttpHooks for InstanceHttpHooks {
         skip_all,
         fields(
             otel.kind = "client",
-            {URL_FULL} = Empty,
-            {HTTP_REQUEST_METHOD} = %request.method(),
+            {otel_attribute::URL_FULL} = Empty,
+            {otel_attribute::HTTP_REQUEST_METHOD} = %request.method(),
             otel.name = %request.method(),
             // Incubating convention; not yet a stable `opentelemetry_semantic_conventions` constant.
             http.response.body.size = Empty,
-            {HTTP_RESPONSE_STATUS_CODE} = Empty,
-            {SERVER_ADDRESS} = Empty,
-            {SERVER_PORT} = Empty,
+            {otel_attribute::HTTP_RESPONSE_STATUS_CODE} = Empty,
+            {otel_attribute::SERVER_ADDRESS} = Empty,
+            {otel_attribute::SERVER_PORT} = Empty,
         )
     )]
     #[allow(clippy::type_complexity)]
@@ -404,12 +402,12 @@ impl p2::WasiHttpHooks for InstanceHttpHooks {
         skip_all,
         fields(
             otel.kind = "client",
-            {URL_FULL} = Empty,
-            {HTTP_REQUEST_METHOD} = %request.method(),
+            {otel_attribute::URL_FULL} = Empty,
+            {otel_attribute::HTTP_REQUEST_METHOD} = %request.method(),
             otel.name = %request.method(),
-            {HTTP_RESPONSE_STATUS_CODE} = Empty,
-            {SERVER_ADDRESS} = Empty,
-            {SERVER_PORT} = Empty,
+            {otel_attribute::HTTP_RESPONSE_STATUS_CODE} = Empty,
+            {otel_attribute::SERVER_ADDRESS} = Empty,
+            {otel_attribute::SERVER_PORT} = Empty,
         )
     )]
     fn send_request(
@@ -491,12 +489,12 @@ impl RequestSender {
         // Backfill span fields after potentially updating the URL in the interceptor
         let span = tracing::Span::current();
         if let Some(addr) = override_connect_addr {
-            span.record(SERVER_ADDRESS, addr.ip().to_string());
-            span.record(SERVER_PORT, addr.port());
+            span.record(otel_attribute::SERVER_ADDRESS, addr.ip().to_string());
+            span.record(otel_attribute::SERVER_PORT, addr.port());
         } else if let Some(authority) = request.uri().authority() {
-            span.record(SERVER_ADDRESS, authority.host());
+            span.record(otel_attribute::SERVER_ADDRESS, authority.host());
             if let Some(port) = authority.port_u16() {
-                span.record(SERVER_PORT, port);
+                span.record(otel_attribute::SERVER_PORT, port);
             }
         }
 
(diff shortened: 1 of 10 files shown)


lann:
FYI: https://docs.rs/tracing/latest/tracing/index.html#recording-fields:~:text=Constant%20expressions%20can,the%20constant%E2%80%99s%20name.

> Constant expressions can also be used as field names. Constants must be enclosed in curly braces ({}) to indicate that the value of the constant is to be used as the field name, rather than the constant’s name.

ChihweiLHBird:
> FYI: https://docs.rs/tracing/latest/tracing/index.html#recording-fields:~:text=Constant%20expressions%20can,the%20constant%E2%80%99s%20name.
> 
> > Constant expressions can also be used as field names. Constants must be enclosed in curly braces ({}) to indicate that the value of the constant is to be used as the field name, rather than the constant’s name.

Thanks! Let me try that. Do we want to include migrations for them in this PR as well?

lann:
> Do we want to include migrations for them in this PR as well?

Up to you.

ChihweiLHBird:
Note that there are other experimental or deprecated semantic conventions that I left as string literals and didn't migrate to the constants in this PR. And it should be ready for review now. Let me know if anything else is needed.

lann:
I probably should have started with this: looking at the diff now I'm not sure this is a net improvement. Avoiding typos is good, but there is also a downside here in that the constants make it harder to see the actual field names. That wouldn't be so bad if they were simply UPPERCASED versions of the field names but they also flatten periods into underscores... 🤔

ChihweiLHBird:
@lann thanks for the thought! That's totally fair and I actually had a similar concern about it because this introduces another dependency.

The upside is the omitted key is still the dotted conventions and benefits of safer compile-time typo checks.
I personally feel it's easy to view the value of the constants too.

Another benefit is that the `opentelemetry-semantic-conventions` crate also makes us aware of the status of each semantic contention, like whether it's deprecated or experimental; so, we can avoid adding a deprecated key when building some new stuff.

I am okay if the team decide to not accept this, and feel free to close the PR in that case.

ChihweiLHBird:
> Another benefit is that the `opentelemetry-semantic-conventions` crate also makes us aware of the status of each semantic contention, like whether it's deprecated or experimental; so, we can avoid adding a deprecated key when building some new stuff.

An example of this is [`db.system`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-system), which is still widely used in `spin` but is actually deprecated. And [it's marked as deprecated in the crate](https://docs.rs/opentelemetry-semantic-conventions/latest/opentelemetry_semantic_conventions/attribute/constant.DB_SYSTEM.html) so triggers warnings in compile time, is used without `#![allow(deprecated)]`.

crates/trigger-http/src/instrument.rs
```diff
@@ -1,5 +1,8 @@
 use anyhow::Result;
 use http::Response;
+use opentelemetry_semantic_conventions::attribute::{
```
calebschoepp: Maybe we could just import all of `opentelemetry_semantic_conventions::attribute` with an alias of `otel_attribute` or something like that.

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -397,12 +404,12 @@ impl p2::WasiHttpHooks for InstanceHttpHooks {
         skip_all,
         fields(
             otel.kind = "client",
```
calebschoepp, line 404: Is this one not a convention?

calebschoepp (commented):
Did you make sure you found all possible attributes to replace?

Overall the advantages of getting warnings when things are deprecated seems worth it.

crates/factor-outbound-http/src/wasi.rs
ChihweiLHBird: Unfortunately, `otel.kind` is not an official convention and ~~is a `spin` specific thing...~~
You can see all `otel.*` keys in the conventions on [this page](https://opentelemetry.io/docs/specs/semconv/registry/attributes/otel/) and it doesn't contain `otel.kind`.

ChihweiLHBird:
@calebschoepp thanks for the review!

> Did you make sure you found all possible attributes to replace?

There are some experimental keys and some deprecated key I didn't replace with constants. The experimental keys require an additional feature of the crate, and the deprecate keys requir `#![allow(deprecated)]` in `lib.rs`. I can add them if you think that's fine. Personally I lean to leave it as is OR only add experimental feature. We can migrate the deprecated keys in the future then replace them with constants; so, we won't have to declare allowing any deprecated item in `lib.rs`

crates/trigger-http/src/instrument.rs
ChihweiLHBird: Good idea! Let me try that

crates/factor-outbound-http/src/wasi.rs
ChihweiLHBird: ~~I would say another benefit of using the official convention crate is we can realize things like this earlier~~

lann: `otel.*` is from `tracing-opentelemetry`: https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/#special-fields

ChihweiLHBird: @lann Thanks! Nice to know it

lann: If we're switching to constants for standard keys it would probably be wise for us to create constants for these too, e.g. `OTEL_KIND`, `OTEL_NAME`. (`tracing-opentelemetry` doesn't seem to define them)

Scope creepily: I wouldn't be sad to have macros for these. We use `otel.name` and `otel.kind` quite a bit so e.g.:

```rust
#[instrument(
  otel_kind!("client"),
  // Lots of this pattern in crates/factor-outbound-redis/src/host.rs
  otel_name!("GET {key}"),
)]
```

expanding to

```rust
#[instrument(
  {$crate::something_goes_here::OTEL_KIND} = "client",
  // Lots of this pattern in crates/factor-outbound-redis/src/host.rs
  {$crate::something_goes_here::OTEL_NAME} = format!("GET {path}"),
)]
```

Note: if you haven't seen it, `$crate` is a magic macro variable: https://doc.rust-lang.org/reference/macros-by-example.html#r-macro.decl.meta.dollar-crate

lann:
I was on the fence so @calebschoepp can be the tie-breaker; I delegate my review to him. 🙂

crates/factor-outbound-http/src/wasi.rs
ChihweiLHBird: > Scope creepily: I wouldn't be sad to have macros for the otel.* fields. We use otel.name and otel.kind quite a bit

This looks like something for `tracing-opentelemetry` to add.

crates/trigger-http/src/instrument.rs
ChihweiLHBird: Getting an alias to work in a macro is more complex than I thought. It would require `use ::opentelemetry_semantic_conventions::attribute as otel_attribute;` in the macro block and also in the modules where the macro is used. Can I leave this as is?

ChihweiLHBird: Using this pattern for all others seem to be fine and I pushed it here: https://github.com/spinframework/spin/pull/3510/commits/0b5917b8ac80be056f72d2b388bf8a2b45686b7e

calebschoepp (approved):
This seems net better to me

PR #3512, opened 2026-05-18
`spin add` should check existing `targets` for templates

itowlson:
If an existing application declares exactly one target environment, and that target environment declares templates, `spin add` should use those templates.

This turned out to not play very nicely with the existing flow.  I've gone with it for now, but this may be pointing us to want to rework some of the `new/add` stuff more substantially - some of the functions are getting a bit long and having to check a lot of cases.  But I don't think we need to tackle this now.

### DIFF
--- a/src/commands/new.rs
+++ b/src/commands/new.rs
@@ -150,7 +150,7 @@ impl TemplateNewCommandCore {
         variant: TemplateVariantInfo,
     ) -> Result<()> {
         let (template_manager, suggested_plugins) =
-            env_templates_and_plugins(target_environment).await?;
+            env_templates_and_plugins(target_environment, &variant).await?;
 
         let (name, template_id) = self.resolve_name_template_syntax(&template_manager, &variant)?;
 
@@ -284,11 +284,25 @@ impl TemplateNewCommandCore {
 
 async fn env_templates_and_plugins(
     target_environment: Option<&String>,
+    variant: &TemplateVariantInfo,
 ) -> anyhow::Result<(TemplateManager, Vec<String>)> {
-    Ok(match target_environment {
-        Some(env) => {
-            //   - resolve the TE
-            let env_ref = parse_env(env);
+    let target_env_ref = match variant {
+        TemplateVariantInfo::NewApplication => target_environment.map(|te| parse_env(te)),
+        TemplateVariantInfo::AddComponent { manifest_path } => {
+            match infer_target_env_from_file(manifest_path) {
+                Ok(t) => t,
+                Err(_) => {
+                    terminal::warn!(
+                        "Couldn't determine target environment; using your installed templates.\n"
+                    );
+                    None
+                }
+            }
+        }
+    };
+
+    Ok(match target_env_ref {
+        Some(env_ref) => {
             let loaded_env =
                 spin_environments::load_environment_def(&env_ref, &std::env::current_dir()?)
                     .await?;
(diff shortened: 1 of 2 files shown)


src/commands/new.rs
```diff
@@ -327,6 +350,25 @@ async fn env_templates_and_plugins(
...
+    let manifest = spin_manifest::manifest_from_file(manifest_path).ok()?;
+    match manifest.application.targets.len() {
+        0 => None,
+        1 => Some(manifest.application.targets[0].clone()),
+        _ => {
+            terminal::warn!("This application targets multiple environments.");
+            eprintln!(
+                "Spin doesn't currently support environment-specific templates for multiple environments."
+            );
+            eprintln!("Templates will instead be taken from the installed template set.");
```
kate-goldenring, line 373: What is the reason that we cannot support `spin add` for spin apps with multiple targets. Is the idea that it is impossible to take the intersection of two templates directories? This almost suggests that each template should have a universal id. I can see many users adding Spin CLI and SpinKube envs by default to most apps.

src/commands/new.rs
```diff
@@ -313,11 +319,28 @@ async fn env_templates_and_plugins(
...
-                    "The {env} environment doesn't list any associated templates. Run `spin new` without `-E` to use generic templates."
-                );
+                match variant {
+                    TemplateVariantInfo::NewApplication => {
+                        anyhow::bail!(
+                            "The {env_ref} environment doesn't list any associated templates. Run `spin new` without `-E` to use generic templates."
+                        );
+                    }
+                    TemplateVariantInfo::AddComponent { .. } => {
+                        // `spin add` doesn't have an option for skipping a templateless env (the equivalent of `spin new` without `-E`).
```
kate-goldenring, line 337: This comment seems to suggest there is a todo or something to be done here. But it seems expected that `spin add` infers `E` from the `spin.toml`. My only concern here is that the `eprintln` seems to indicate it is errorsome for a target env to not provide templates. is that the case?

kate-goldenring (commented):
Just a couple clarifying questions. Should any tests be updated/added?

src/commands/new.rs
```diff
@@ -327,6 +350,25 @@ async fn env_templates_and_plugins(
     })
 }
 
+fn infer_target_env_from_file(
+    manifest_path: impl AsRef<Path>,
+) -> Option<spin_manifest::schema::v2::TargetEnvironmentRef> {
+    let manifest = spin_manifest::manifest_from_file(manifest_path).ok()?;
```
fibonacci1729: Is `ok()?` correct? A malformed `spin.toml` becomes indistinguishable from "no targets" and the user silently gets the generic template set in this case.

fibonacci1729 (approved):

src/commands/new.rs
itowlson: See the four-ish messages beginning at https://github.com/spinframework/spin/issues/3494#issuecomment-4401954591.  It's not a great situation I admit.

src/commands/new.rs
itowlson: It's not errorsome, but it does mean we can no longer guarantee in advance that the templates are vetted to be suitable for the given TE.  Like if you targeted Spin 3.6, and that didn't declare any templates, you'd get the Spin 4 templates, which might land you with WASI P3.  Hence the warning.  I'm happy to reword it.

(That said, my feeling is leaning to "a TE probably _should_ include templates because otherwise it's saying that some random version of the Spin templates will work with it.")

src/commands/new.rs
itowlson: My intent was to not block the user if their template wasn't perfect, but yeah, maybe I should simply dig TOML and not schema-validate.

(And if it's invalid TOML then we probably want to fail early because I think the manifest edit stage later in the add will fail in that case.)

itowlson: Ugh new plan I am not going to dig TOML because of the various enum cases.  I will add a warning.

PR #3523, opened 2026-05-21
`spin new -E`: update unversioned environments if not recent

itowlson:
The `spin new -E` feature assumed that environments were immutable - that a new environment could get a new version.  This is true for self-host platforms such as the CLI and SpinKube.  However, for hosted environments it leads to a weird UI.  Let us return to Bob's Discount Cloud, the premier bargain-basement Spin host.  At any give time, BDC supports a specific environment.  But the current thing requires users to pick a version. At first Bob tells them to write `spin new -E bdc@1`, which is a bit weird but uh okay; but then later Bob must raze all those docs from the Internet and get them to write `spin new -E bdc@2`, instead.  Really Bob wants users to be able to write `spin up -E bdc` and get whatever is current in his Discount Cloud.

We could tackle this by forcing an update of the environments repo on every `get`.  However, this introduces a delay.  If you're on airplane wifi, that could be quite a loooooooooong delay.

This PR proposes to track the recency of the catalogue snapshot, and to update it only if 1. the env being referenced is unversioned; _and_ 2. the snapshot is older than 1 hour.  In addition, such updates will be timed out after 2 seconds to avoid slow networks making them painful.

This hopefully minimises user annoyance; and hopefully Bob's updates will not usually be time-sensitive to the hour (if they are, Bob will have to ask users to purge their envs cache, or to `spin new -E some-fake-env` to force a fetch: we can look at adding commands for this if need be).

(And... I realised while working on this that `spin new` currently always does a git fetch anyway, to force-install templates from the referenced repo.  So maybe this was a pointless optimisation.  But I'd like to fix that soon, so I'm not going to bother backing it out...)

Tested by pointing a build at a fork of the environments repo, dialling the recency timeout down to a few minutes, and logging when pulls did and didn't happen (as well as verifying that changes were reflected in the `spin new` UI).

### DIFF
--- a/crates/environments/src/environment/catalogue.rs
+++ b/crates/environments/src/environment/catalogue.rs
@@ -1,3 +1,5 @@
+use std::time::{Duration, SystemTime};
+
 const SPIN_ENV_REPO: &str = "https://github.com/spinframework/spin-environments";
 const ENVS_DIR_IN_REPO: &str = "envs";
 
@@ -7,6 +9,8 @@ pub struct Catalogue {
 }
 
 static CATALOGUE_UPDATE_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+const JUST_IN_TIME_UPDATE_TIMEOUT: Duration = Duration::from_secs(2);
+const RECENCY_WINDOW: Duration = Duration::from_hours(1);
 
 impl Catalogue {
     pub fn try_default() -> anyhow::Result<Self> {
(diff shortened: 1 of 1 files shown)


crates/environments/src/environment/catalogue.rs
```diff
@@ -100,6 +171,10 @@ fn sans_version(id: &str) -> &str {
     }
 }
 
+fn is_unversioned(id: &str) -> bool {
```
kate-goldenring, line 172: We should update the readme of `spin-environments` to specify that versioning is not needed.

crates/environments/src/environment/catalogue.rs
```diff
@@ -17,29 +21,96 @@ impl Catalogue {
...
+
     pub async fn update(&self) -> anyhow::Result<()> {
         // We don't want two git pulls running concurrently
         let _guard = CATALOGUE_UPDATE_LOCK.lock();
 
         let url = Url::parse(SPIN_ENV_REPO)?;
         let git_source = GitSource::new(&url, None, &self.git_root);
         if self.git_root.exists() {
-            git_source.pull().await
+            git_source.pull().await.unwrap();
```
kate-goldenring: Why are we unwrapping here? This could panic on a network error
```suggestion
            git_source.pull().await?;
```

crates/environments/src/environment/catalogue.rs
```diff
@@ -17,29 +21,96 @@ impl Catalogue {
...
+
     fn new(git_root: PathBuf) -> Self {
         Self {
             git_root: git_root.clone(),
             envs_root: git_root.join(ENVS_DIR_IN_REPO),
         }
     }
 
+    /// Updates if we have not updated recently, ignoring
+    /// failures. The scenario here is unverioned environments,
```
kate-goldenring: nit
```suggestion
    /// failures. The scenario here is unversioned environments,
```

kate-goldenring (commented):
just the `unwrap` handling and then LGTM

crates/environments/src/environment/catalogue.rs
```diff
@@ -17,29 +21,96 @@ impl Catalogue {
...
+                };
+                let Ok(diff) = now.duration_since(last) else {
+                    return false;
+                };
+                diff < RECENCY_WINDOW
+            }
+        }
+    }
+
+    fn last_update_file(&self) -> Result<PathBuf, ()> {
```
fibonacci1729: Would `-> Option<PathBuf>` be more idiomatic here?

fibonacci1729 (approved):
LGTM pending Kate's `unwrap()` and typos.

kate-goldenring (approved):

Issue #3520, opened 2026-05-20
Deprecate support for WAGI


calebschoepp:
WAGI is rarely used and imposes a large maintenance burden on Spin. I propose that we make the breaking change of deprecating support for WAGI. Some discussion around this has already happened in this [Slack thread](https://cloud-native.slack.com/archives/C089NJ9G1V0/p1779221988733209) in the CNCF Slack.

Ivan raised a good point in the thread that we should be cautious about WAGI usage in downstream Spin hosted services. It would make sense to encourage those to deprecate WAGI first before we eventually deprecate WAGI in Spin.

lann:
A phased approach:

- Soon: start emitting deprecation warnings, e.g.: `You're using wagi but we're getting rid of it; complain here: https://github.com/spinframework/spin/issues/new`
- Spin N+1: turn the warning into a fatal error with the option to keep it working with e.g. `SPIN_ENABLE_DEPRECATED_WAGI=1`
- Spin N+2: remove wagi

PR #3526, opened 2026-05-21
Remove wagi examples

lann:
A journey of a thousand miles begins with a single step.

Ref #3520


fibonacci1729:
Should we wait until we go through the deprecation process before removing these?

lann:
I don't think it's necessary. To me examples should reflect a project's recommended approach to solving a problem and I don't think we'd recommend wagi for rust or cpp...

lann:
(templates on the other hand probably warrant at least more discussion)

fibonacci1729 (approved):

PR #3534, opened 2026-05-26
Asyncify MySQL interface

dicej:
This adds a new `spin:mysql/mysql@3.0.0` interface with async functions for opening, executing, and querying.

The implementation uses `Arc<Mutex<_>>` to control access to the host state and individual connections, meaning at most one `open` operation may proceed at a time, and each connection will accept at most one `execute` or `query` operation at a time, with any other operations queuing up behind that one.  In no case will the guest be blocked (assuming `mysql-async` is well-behaved), so it can do other things concurrently while it's waiting for DB I/O.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8877,6 +8877,7 @@ dependencies = [
  "spin-factors-test",
  "spin-resource-table",
  "spin-telemetry",
+ "spin-wasi-async",
  "spin-world",
  "tokio",
  "tracing",
--- a/crates/factor-outbound-mysql/Cargo.toml
+++ b/crates/factor-outbound-mysql/Cargo.toml
@@ -22,6 +22,7 @@ spin-factors = { path = "../factors" }
 spin-resource-table = { path = "../table" }
 spin-telemetry = { path = "../telemetry" }
 spin-world = { path = "../world" }
+spin-wasi-async = { path = "../wasi-async" }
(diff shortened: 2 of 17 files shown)


itowlson (approved):
I am a bit confused about the `v2::Connection` and `v3::Connection` stuff going on here.  The places where I see it opening a connection and getting a client all deal in terms of `v2::Connection`  so I am not clear where the `v3::Connection` is coming from.  But probably this is something I already do in the other factors and I have just forgotten about it...!

Anyway LGTM except for that small befuddlement on my part!

dicej:
> I am a bit confused about the `v2::Connection` and `v3::Connection` stuff going on here. The places where I see it opening a connection and getting a client all deal in terms of `v2::Connection` so I am not clear where the `v3::Connection` is coming from. But probably this is something I already do in the other factors and I have just forgotten about it...!
> 
> Anyway LGTM except for that small befuddlement on my part!

It's [this line](https://github.com/dicej/spin/blob/9bb305e17e2845aa2d3910d7a4e580f113301579/crates/world/src/lib.rs#L51).  I figured since `v2::Connection` is just an empty type anyway I could reuse it, but on reflection I'm realizing that could cause confusing for both human readers (hence your comment) and the compiler, so I'm going to change that tomorrow.

In short, bad decision on my part and I'll fix it.

PR #3535, opened 2026-05-27
Target envs: pull WITs direct from OCI

itowlson:
Allows target envs to use packages pushed with `wkg oci push` (so that no special wkg registry endpoint is needed on the server).

~This is untested so far - I can't create a test OCI because of some weirdness in `wkg` (issue raised).  @vdice is going to try it, leaving as draft until then.~ Tested it. Worked on my machine which is about as much as we can say of any of this so far!


vdice (approved):
Works great!  Tested with a local environment config referencing a WIT file pushed via `wkg oci push`.  Having this OCI option is very convenient, thank you.

PR #3537, opened 2026-05-27
Cap Number of Open TCP and UDP Sockets

rylev:
This adds the ability to limit the number of open TCP and UDP sockets to a set number so that embedders can more easily control the number of open file descriptors they have.


crates/factor-wasi/src/lib.rs
```diff
@@ -4,27 +4,539 @@ mod wasi_2023_10_18;
...
+use wasmtime_wasi::p2::bindings::sockets::udp_create_socket as p2_udp_create;
+use wasmtime_wasi::p2::{DynInputStream, DynOutputStream, DynPollable};
 use wasmtime_wasi::random::{WasiRandom, WasiRandomCtx};
-use wasmtime_wasi::sockets::{WasiSockets, WasiSocketsCtxView};
+use wasmtime_wasi::sockets::{TcpSocket, UdpSocket, WasiSockets, WasiSocketsCtxView};
 use wasmtime_wasi::{DirPerms, FilePerms, ResourceTable, WasiCtx, WasiCtxBuilder, WasiCtxView};
 
 pub use wasmtime_wasi::sockets::SocketAddrUse;
 
+/// Shared state for tracking per-socket semaphore permits. Permits are
```
lann: Given that it is >doubling the size of this file it might be nice to move all of this into e.g. `sockets.rs`.

crates/factor-wasi/src/lib.rs
```diff
@@ -4,27 +4,539 @@ mod wasi_2023_10_18;
...
+        network: Resource<Network>,
+        remote_address: IpSocketAddress,
+    ) -> wasmtime_wasi::p2::SocketResult<()> {
+        if let Some(state) = &self.permit_state {
+            // If we have a permit state, we need to acquire a permit before allowing the connection to proceed.
+            let socket_rep = this.rep();
+            let Ok(permit) = Arc::clone(&state.semaphore).try_acquire_owned() else {
+                // wasi has no "quota exceeded" error code. ConnectionRefused is the closest available.
+                return Err(SocketErrorCode::ConnectionRefused.into());
+            };
```
lann: The outbound http version of this _waits_ for a permit. I don't have a strong opinion at the moment about whether that is better or worse than failing but it seems like we should be consistent.

lann: Separately, it we do end up returning an error here we should probably at least emit a tracing event.

rylev: I'm not sure that it does make sense to have the same semantics for outbound HTTP and raw sockets. Outbound HTTP slots are freed by the runtime when the response comes back. There's no explicit guest action required to free the resource and open the slot. This means in the fullness of time slots are _guaranteed_ to eventually become available. Socket slots, by contrast, are only freed when the guest explicitly drops the resource. This means a guest holding N sockets and waiting for an N+1th permit will deadlock. It can never release the permits it holds while suspended waiting for a new one.

rylev:

PR #3540, opened 2026-05-27
Drop `BetweenBytesTimeoutBody` inner body eagerly when timeout happens

rylev:
This fixes a bug where when a between-bytes timeout fired, the inner HTTP body was kept alive, leaking the underlying TCP connection until the Wasm guest eventually dropped its resource handle. The fix wraps body in `Option<B>` so it can be set to `None` (dropping the connection eagerly) right before returning the `ConnectionReadTimeout` error.

### DIFF
--- a/crates/factor-outbound-http/src/wasi.rs
+++ b/crates/factor-outbound-http/src/wasi.rs
@@ -275,18 +275,17 @@ impl<B: Body<Error = p2_types::ErrorCode> + Unpin> Body for BetweenBytesTimeoutB
             Pin::new(body).poll_frame(cx)
         };
 
+        let mut record_body_size_once = |body_size: u64| {
+            if let Some(span) = me.span.take() {
+                // `http.response.body.size` is incubating (behind semconv_experimental)
+                // in opentelemetry-semantic-conventions 0.28. Leave as literal to avoid
+                // enabling the experimental feature.
+                span.record("http.response.body.size", body_size);
+            }
+        };
         match poll_result {
             Poll::Ready(value) => {
                 me.sleep.as_mut().set(None);
-
-                let mut record_body_size_once = |body_size: u64| {
-                    if let Some(span) = me.span.take() {
-                        // `http.response.body.size` is incubating (behind semconv_experimental)
-                        // in opentelemetry-semantic-conventions 0.28. Leave as literal to avoid
-                        // enabling the experimental feature.
-                        span.record("http.response.body.size", body_size);
-                    }
-                };
 
                 match &value {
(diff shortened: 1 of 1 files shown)


crates/factor-outbound-http/src/wasi.rs
```diff
@@ -302,17 +312,25 @@ impl<B: Body<Error = p2_types::ErrorCode>> Body for BetweenBytesTimeoutBody<B> {
...
+                *me.body = None;
+
                 Poll::Ready(Some(Err(p3_types::ErrorCode::ConnectionReadTimeout)))
             }
         }
     }
 
     fn is_end_stream(&self) -> bool {
-        self.body.is_end_stream()
+        self.body.as_ref().is_some_and(|b| b.is_end_stream())
```
lann: `body.is_none()` means we've already returned an error, right?

```suggestion
        self.body.as_ref().is_none_or(|b| b.is_end_stream())
```

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -302,17 +312,25 @@ impl<B: Body<Error = p2_types::ErrorCode>> Body for BetweenBytesTimeoutBody<B> {
                     me.sleep.as_mut().set(Some(tokio::time::sleep(*me.timeout)));
                 }
                 task::ready!(me.sleep.as_pin_mut().unwrap().poll(cx));
+
+                // Drop the inner body immediately to close the underlying TCP
+                // connection rather than waiting for the guest to release the resource.
+                *me.body = None;
+
```
lann, line 319: Tangential but seems like a good change while we're here...

```suggestion
                            record_body_size_once(*me.byte_count);
```

crates/factor-outbound-http/src/wasi.rs
```diff
@@ -302,17 +312,25 @@ impl<B: Body<Error = p2_types::ErrorCode>> Body for BetweenBytesTimeoutBody<B> {
                     me.sleep.as_mut().set(Some(tokio::time::sleep(*me.timeout)));
                 }
                 task::ready!(me.sleep.as_pin_mut().unwrap().poll(cx));
+
+                // Drop the inner body immediately to close the underlying TCP
+                // connection rather than waiting for the guest to release the resource.
```
lann: nit: This is probably true for incoming HTTP/1.1 but isn't true for outgoing bodies and probably not for HTTP/2.

```suggestion
                // Drop the inner body immediately to free resources (like sockets)
                // rather than waiting for the guest to release the resource.
```

dicej (approved):

PR #3541, opened 2026-05-27
Abort tokio task when client disconnects so outbound HTTP connections close rather than leaking until timeout

rylev:
Wraps the WASI HTTP response body in an `AbortOnDropBody` that holds a `tokio::task::AbortHandle` pointing at the handle task, so that when Hyper drops the body on client disconnect, the component task is immediately aborted. This causes the Wasm Store to be dropped, which cascades to drop any pending outbound `AbortOnDropJoinHandle` resources and close their underlying TCP connections.

### DIFF
--- a/crates/trigger-http/src/wasi.rs
+++ b/crates/trigger-http/src/wasi.rs
@@ -1,15 +1,19 @@
 use std::io::IsTerminal;
 use std::net::SocketAddr;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
-use anyhow::{Context, Result, anyhow};
-use futures::TryFutureExt;
+use anyhow::{Context as _, Result, anyhow};
 use http::{HeaderName, HeaderValue};
+use http_body::{Body as HttpBody, Frame, SizeHint};
+use http_body_util::BodyExt;
 use hyper::{Request, Response};
 use spin_factor_outbound_http::wasi_2023_10_18::Proxy as Proxy2023_10_18;
 use spin_factor_outbound_http::wasi_2023_11_10::Proxy as Proxy2023_11_10;
 use spin_factors::RuntimeFactors;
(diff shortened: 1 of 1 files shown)


crates/trigger-http/src/wasi.rs
```diff
@@ -130,29 +134,38 @@ impl<S: HandlerState> HttpExecutor for WasiHttpExecutor<'_, S> {
...
+                        // Task was aborted because the incoming request was cancelled
+                        // (e.g. client disconnected). This is expected; don't log it as an error.
+                        Err(e) if e.is_cancelled() => {}
+                        result => {
+                            let result =
+                                result.context("guest invocation panicked").and_then(|r| {
+                                    r.map_err(anyhow::Error::from)
+                                        .context("guest invocation failed")
+                                });
+                            if let Err(e) = result {
```
lann, line 152: minor: This error handling logic was already hard to follow... I suspect it could be simplified by extracting it into a function which would let `?` work. (my kingdom for stable `try` blocks 😭)

lann (commented):
I think this is changing intentional behavior. WASI handlers are allowed keep running after a response is complete, successfully or otherwise.

lann (changes requested):

rylev:
@lann this is a good point. I was mostly thinking of this from the "client disconnects" perspective which I think one could argue might be reason to kill the guest since the entity that initially kicked off the request is no longer there. In any case, in retrospective, if we wanted to make this change, it's best to do that in an issue or a SIP first as it does change behavior in a non-trivial manner.

Additionally, as you point out this has the added effect of killing a running guest even when the request/response has been fully handled. While I've never personally used Spin for the use case of "continuing side effects after the request/response handling is done", this does seem to be a legitimate usage, and so this change probably should not make it in.

PR #3542, opened 2026-05-27
[Backport v4.0] fix: support falling back to webpki roots in default TLS config

fibonacci1729:
Backporting #3536 to v4.0


kate-goldenring (approved):
We need to first backport https://github.com/spinframework/spin/pull/3491

PR #3543, opened 2026-05-27
outbound-mqtt: add runtime config for mqtt max payload size

kate-goldenring:
Sets a default max MQTT payload side for publish calls to 1MiB with the option to configure a max via runtime configuration.

### DIFF
--- a/crates/factor-outbound-mqtt/tests/factor_test.rs
+++ b/crates/factor-outbound-mqtt/tests/factor_test.rs
@@ -9,6 +9,7 @@ use spin_factor_variables::VariablesFactor;
 use spin_factors::{RuntimeFactors, anyhow};
 use spin_factors_test::{TestEnvironment, toml};
 use spin_world::spin::mqtt::mqtt::{Error, Qos};
+use spin_world::v2::mqtt as v2;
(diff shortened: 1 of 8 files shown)


crates/factor-outbound-mqtt/tests/factor_test.rs
```diff
@@ -139,3 +137,110 @@ async fn exercise_publish() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn oversized_payload_rejected() -> anyhow::Result<()> {
```
rylev, line 142: The tests all seem to be `v2` (the `fermyon:spin` world) based. Would it make sense for us to also test the `spin:up` world based implementation? I suppose it would be fine to leave this `v2` based and open an issue for switching mqtt tests to test the latest implementation.

rylev (approved):
🎉

lann (dismissed):
The default makes this a breaking change. Does this actually need a default?

kate-goldenring:
> The default makes this a breaking change. Does this actually need a default?

@lann good call out. I removed the default

PR #3545, opened 2026-05-28
Limit the number of redis connections an app can have across instances

rylev:
Similar to https://github.com/spinframework/spin/pull/3285 and https://github.com/spinframework/spin/pull/3537 this limits the number of open connections an app can have with Redis across all instances of that app. 

Writing tests for this is a bit awkward unfortunately.


rylev:

PR #3546, opened 2026-05-28
Limit the number of pg/mysql connections an app can have across instances

rylev:
Same thing as #3545 but for MySQL and Postgres.


lann:
Should this be a single connection limit across all interfaces? 😬

rylev:
@lann Yes I think so but not exclusively. I think it is useful to have individual limits for factors _and_ a global limit. This would give users the ability to do things like limit total connections to say 100 but limit Redis specifically to 5. I went ahead and implemented this in https://github.com/spinframework/spin/commit/626b698e2b12a1c43272fd66608a9812e879f6d8.

If the idea of a combined global and factor specific connections limiting system sounds good to you, the question becomes: how do you best want to review this. Would you prefer if I collapsed #3537, #3545, #3546, and https://github.com/spinframework/spin/commit/626b698e2b12a1c43272fd66608a9812e879f6d8 all into one branch and then opened up a new PR? Such a PR would be pretty big but then you'd have all the context in one place. Otherwise, we could merge the factor specific PRs first and then open up a PR for https://github.com/spinframework/spin/commit/626b698e2b12a1c43272fd66608a9812e879f6d8 separately.

lann:
Assuming 626b698 is refactoring significant chunks of the other PRs I think a single PR would mean less redundant review.

rylev:

PR #3547, opened 2026-05-29
Deprecate max_concurrent_requests

rylev:
*Setting to draft so we can let #3545, #3546, and #3537 settle first since this brings outbound http into line with the changes to the other factors made in those PRs.*

This deprecates the `max_concurrent_requests` runtime config setting to bring it more into line with the naming conventions for other factors introduced in #3545, #3546, and #3537. It also gets rid of the weird "0 means 1" semantics in the newly introduced option that I am responsible for despite @lann telling me it was a bad idea (I SHOULD HAVE LISTENED). 

This is a backwards compatible change, and we can consider removing the option completely in some future release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8991,6 +8991,7 @@ dependencies = [
  "spin-factors-test",
  "spin-telemetry",
  "spin-world",
+ "terminal",
  "tokio",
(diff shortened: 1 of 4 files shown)


rylev:

PR #3551, opened 2026-06-02
Skip allowed outbound host entries that are empty due to empty variable resolution

### LINKED ISSUE #3549: No support for optional allowed_outbound_hosts driven by an unset Spin variable
### Summary

When a Spin variable is used to supply an optional outbound host URL (e.g., a telemetry endpoint that may or may not be configured), there is no way to conditionally include that host in allowed_outbound_hosts. If the variable is left unset or defaults to an empty string "", Spin fails to start because an empty string is not a valid host value.

### Steps to reproduce

Given a component in spin.toml with an optional telemetry endpoint:

```
[variables]
telemetry_host = { default = "" }
telemetry_path = { default="/v1/traces" }

[component.my-app]
source = "my_app.wasm"
allowed_outbound_hosts = ["{{ telemetry_host }}"]

[component.my-app.variables]
telemetry_endpoint = "{{ telemetry_host }}{{ telemetry_path }}"
```

### Actual behavior

When telemetry_endpoint is not set (or defaults to ""), Spin fails at startup with an error similar to:

```

Caused by:
   2: Failed to load component `my-app`
   3: `allowed_outbound_hosts` is malformed
   4: using default variable value(s) with template "{{ telemetry_host }}" results in invalid config ""
   5: "" is not an allowed outbound host format.
Hosts must be in the form <scheme>://<host>[:<port>], with '*' wildcards allowed for each.
If you intended to allow all outbound networking, you can use '*://*:*' - this will obviate all network sandboxing.
Learn more: https://spinframework.dev/v3/http-outbound#granting-http-permissions-to-components
```

### Proposed Solution

A possible implementation would be to treat an empty or unset variable in allowed_outbound_hosts as "no entry". I.e., the host is simply omitted from the list rather than causing a startup failure. This would allow developers to implement optional outbound integrations without compromising on least-privilege networking.

### Workaround

Use the .invalid TLD as the default. Per [RFC 6761 §6.4](https://www.rfc-editor.org/info/rfc6761/#section-6.4), .invalid is guaranteed never to resolve:

```
telemetry_host = { default = "https://disabled.invalid" }
```

The workaround requires the developer to add defensive logic in application code to avoid actually attempting the request based on the `.invalid` TLD. It works, but is not a clean solution.

itowlson:
Fixes #3549.

I had to introduce a resolver trait instead of using PreparedResolver for the annoying reason that I couldn't initialise a PreparedResolver with empty strings for testing.  I tried a `#[cfg(test)]` method on PreparedResolver to initialise it with fixed values, but it didn't work across the crate boundary.

### DIFF
--- a/crates/expressions/src/lib.rs
+++ b/crates/expressions/src/lib.rs
@@ -132,6 +132,22 @@ pub struct Resolver {
     component_configs: HashMap<String, HashMap<String, Template>>,
 }
 
+impl SyncResolver for Resolver {
+    fn resolve_variable(&self, key: &str) -> Result<String> {
+        let var = self
+            .variables
+            .get(key)
+            // This should have been caught by validate_template
+            .ok_or_else(|| Error::InvalidName(key.to_string()))?;
+
+        var.default.clone().ok_or_else(|| {
+            Error::Provider(anyhow::anyhow!(
+                "no provider resolved required variable {key:?}"
+            ))
+        })
+    }
+}
+
 impl Resolver {
     /// Creates a Resolver for the given Tree.
     pub fn new(variables: impl IntoIterator<Item = (String, Variable)>) -> Result<Self> {
@@ -196,20 +212,6 @@ impl Resolver {
         Ok(template)
     }
 
-    fn resolve_variable(&self, key: &str) -> Result<String> {
-        let var = self
-            .variables
-            .get(key)
-            // This should have been caught by validate_template
-            .ok_or_else(|| Error::InvalidName(key.to_string()))?;
-
-        var.default.clone().ok_or_else(|| {
-            Error::Provider(anyhow::anyhow!(
-                "no provider resolved required variable {key:?}"
-            ))
-        })
-    }
-
     fn validate_template(&self, template: String) -> Result<Template> {
         let template = Template::new(template)?;
         // Validate template variables are valid
@@ -229,15 +231,16 @@ impl Resolver {
     }
 }
 
-/// A resolver who has resolved all variables.
-#[derive(Default)]
-pub struct PreparedResolver {
-    variables: HashMap<String, String>,
-}
-
-impl PreparedResolver {
-    /// Resolves a the given template.
-    pub fn resolve_template(&self, template: &Template) -> Result<String> {
+/// Resolves a template to a string.
+///
+/// Because this trait is synchronous, implementations should not
+/// perform I/O. Mostly this is an abstraction over PreparedResolver
+/// to allow tests to inject fakes.
+pub trait SyncResolver {
+    /// Resolves the given template.
+    ///
+    /// Do not override this implementation.
+    fn resolve_template(&self, template: &Template) -> Result<String> {
         let mut resolved_parts: Vec<Cow<str>> = Vec::with_capacity(template.parts().len());
         for part in template.parts() {
             resolved_parts.push(match part {
@@ -248,6 +251,20 @@ impl PreparedResolver {
         Ok(resolved_parts.concat())
     }
 
+    /// Resolves a variable to a string.
+    ///
+    /// This must be implemented, but consumers should not usually
+    /// call it directly: call `resolve_template` instead.
+    fn resolve_variable(&self, key: &str) -> Result<String>;
+}
+
+/// A resolver who has resolved all variables.
+#[derive(Default)]
+pub struct PreparedResolver {
+    variables: HashMap<String, String>,
+}
+
+impl SyncResolver for PreparedResolver {
     fn resolve_variable(&self, key: &str) -> Result<String> {
         self.variables
             .get(key)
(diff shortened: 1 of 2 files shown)


crates/expressions/src/lib.rs
```diff
@@ -229,15 +229,9 @@ impl Resolver {
...
 }
 
-/// A resolver who has resolved all variables.
-#[derive(Default)]
-pub struct PreparedResolver {
-    variables: HashMap<String, String>,
-}
-
-impl PreparedResolver {
+pub trait SyncResolver {
```
rylev, line 239: Would be good to have doc comments for the trait itself.

crates/outbound-networking-config/src/allowed_hosts.rs
```diff
@@ -431,11 +431,21 @@ impl PartialAllowedHostConfig {
...
         match self {
-            Self::Exact(h) => Ok(h),
-            Self::Unresolved(t) => AllowedHostConfig::parse(resolver.resolve_template(&t)?),
+            Self::Exact(h) => Ok(Some(h)),
+            Self::Unresolved(t) => {
+                let resolved = resolver.resolve_template(&t)?;
+                // An empty variable could result in an empty host. We are
+                // should ignore these, so as to support an "optional endpoint"
+                // scenario.
+                if resolved.is_empty() || resolved == "http://" || resolved == "https://" {
```
rylev: Should we trim resolved so that `"http:// " and other whitespace wrapped `resolved`s work?

crates/outbound-networking-config/src/allowed_hosts.rs
```diff
@@ -431,11 +431,21 @@ impl PartialAllowedHostConfig {
...
+        resolver: &impl spin_expressions::SyncResolver,
+    ) -> anyhow::Result<Option<AllowedHostConfig>> {
         match self {
-            Self::Exact(h) => Ok(h),
-            Self::Unresolved(t) => AllowedHostConfig::parse(resolver.resolve_template(&t)?),
+            Self::Exact(h) => Ok(Some(h)),
+            Self::Unresolved(t) => {
+                let resolved = resolver.resolve_template(&t)?;
+                // An empty variable could result in an empty host. We are
+                // should ignore these, so as to support an "optional endpoint"
```
lann:
```suggestion
                // An empty variable could result in an empty host. We
                // ignore these, so as to support an "optional endpoint"
```

crates/outbound-networking-config/src/allowed_hosts.rs
lann: Given that whitespace is never allowed in a URL I probably wouldn't. 🤷

rylev: For example: `"https://{{ my_var }} "` would lead to `"https://foo.com "` when `my_var` == `"foo.com"` which I think we should permit since the space is trailing, and `"https:// "` when `my_var` == `""`. 

Erroring in that case seems like an unnecessary paper cut.

lann: True, but there is another scenario: `"https://{{ my_var }}"` and accidentally setting `my_var` to `" "`. If your guest logic checks `my_var.is_empty()` (which is what https://github.com/spinframework/spin/issues/3549 is about) then you'll have a runtime error (hopefully) which could have been a manifest validation failure. This is definitely subjective but I'll take a little bit of extra safety at the expense of a similarly-sized paper cut. 🤷

lann: (I would probably also disallow whitespace in `allowed_outbound_hosts` but I guess that ship has sailed)

fibonacci1729: This only handles `http://` and `https://`; should we consider the other outbound schemes as well? (i.e. `redis://`, `mysql://`, `postgres://`, `mqtt://`).

If deferring we should at least leave a TODO.

crates/outbound-networking-config/src/allowed_hosts.rs
fibonacci1729, line 453: I think the empty-skip-test needs to be applied here too. The actual error text in the issue seems to indicate this is the source of the error: `using default variable value(s) with template "{{ telemetry_host }}" results in invalid config ""`

The test coverage doesn't reflect this because it only exercises `parse`. There's no test for `validate` with an empty-default variable which is the scenario described in the issue.

itowlson: Good catch - thanks.  I wonder if this should be calling `self.parse` rather than replicating logic... anyway I will add tests for it and figure it out.

fibonacci1729: That or a standalone helper makes sense to me!

fibonacci1729 (approved):
Looks good!

PR #3552, opened 2026-06-02
Connection limits

rylev:
*This PR is probably best reviewed commit by commit. Each factor's changes are largely independent and follow the same pattern.*

This adds connection limits (both per factor and global) for the following factors (which originally had separate PRs which have been closed in favor of a single PR):
* Wasi (sockets): https://github.com/spinframework/spin/pull/3537
* Redis: https://github.com/spinframework/spin/pull/3545
* MySQL: https://github.com/spinframework/spin/pull/3546
* Postgres: https://github.com/spinframework/spin/pull/3546
* MQTT

Additionally the global connection limit is connected to the existing outbound http connection limiting introduced in https://github.com/spinframework/spin/pull/3285. The runtime config toml has also been updated to be consistent across all factors while retaining the now deprecated existing toml naming for backwards compatibility.

### Open Issues/Questions
* I was not able to wrap the wasi P3 Socket implementation in a wrapper type that could hold on to a `ConnectionSemaphore`. I may need to talk with @dicej about introducing changes in `wasmtime-wasi` to allow this.
* The wasi socket connections error immediately when no quota is available where as all other implementations `.await` for quote blocking the guest from making progress. Should we always immediately error so that guests can decide what to do when no capacity is available? We treated sockets differently because waiting would risk deadlock if a component holds sockets open across async yield points.
* KV is not limited. Should we limit KV with the same mechanism, a different mechanism, or not at all?

### Breaking changes

This does contain one breaking change (beyond factor APIs which I believe we do not have any stability guarantees at this point): the `outbound_http.concurrent_connection_permits_acquired` metric has been renamed to `outbound_connection_permits_acquired` with the label `kind` set to `outbound_http`. Maintaining backward compatibility here is ugly but doable if we believe it is important.


crates/connection-semaphore/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            }
+        }
+        let mut waited = false;
+        let start = std::time::Instant::now();
+
+        let (global, factor_specific) = match (&self.global, &self.factor_specific) {
+            (None, None) => (None, None),
+            (Some(g), None) => (Some(acquire_one(g, &mut waited, "global").await?), None),
+            (None, Some(f)) => (None, Some(acquire_one(f, &mut waited, "factor").await?)),
+            // Loop until we acquire both. We have to be careful to avoid holding one permit while waiting for the other.
```
lann: I think this could cause problems. Lets say we have tasks A, B, C, D all in the same factor code, all on different threads.

- A: already has permits
- B: waiting on G.acquire
- C: waiting on F.acquire
- A: finishes, drop(G+F)
- B: acquires G
- C: acquires F
- B: F.try_acquire fails
- C: G.try_acquire fails
- B: drop(G)
- C: drop(F)
- D: swoops in and acquires G and F
- Now B & C have to start over at the backs of the semaphore queues

lann: I _think_ this should work?

(replacing the whole outer `match`)
```rust
let factor_specific = match &self.factor_specific {
    Some(f) => Some(acquire_one(f, &mut waited, "factor").await?),
    None => None,
};

let global = match &self.global {
    Some(g) => Some(acquire_one(g, &mut waited, "global").await?),
    None => None,
}; 
```

The problem with acquiring global first is that the task might then wait on the factor-specific semaphore and block a task in a different factor that could otherwise make progress, but I think reversing the order fixes that; if a task acquires factor-specific and waits on global it will only block other tasks on the same factor _which would be blocked on global anyway_.

lann: FWIW Gemini agrees with me (after pointing out that the `if let` in my first version was bogus 😅).

lann (commented):
~My chain of 3 comments would have made more sense if I had remembered I was in a review. Imagine gaps and edits between them...~ Never mind I guess github remembers for you...

dicej:
> I was not able to wrap the wasi P3 Socket implementation in a wrapper type that could hold on to a `ConnectionSemaphore`. I may need to talk with @dicej about introducing changes in `wasmtime-wasi` to allow this.

What issues did you run into there?  I'm happy to start working on any changes needed.

rylev:
@dicej you can see code plus a print out of the compilation errors in [this commit](https://github.com/spinframework/spin/commit/70d0436dc07073c25ecdb0e18d5f7c8b2a8b6e2a). Essentially there's no way to make the `store.with_getter` call to work since the generic `T` is just a plain generic that the implementor cannot set.

crates/connection-semaphore/src/lib.rs
rylev: Done: 3e9fb87f5a1b4772de49746c0d02d91bda82cabf

dicej:
> @dicej you can see code plus a print out of the compilation errors in [this commit](https://github.com/spinframework/spin/commit/70d0436dc07073c25ecdb0e18d5f7c8b2a8b6e2a). Essentially there's no way to make the `store.with_getter` call to work since the generic `T` is just a plain generic that the implementor cannot set.

Thanks, I'll check out that commit and play around with it.

rylev:
> I was not able to wrap the wasi P3 Socket implementation in a wrapper type that could hold on to a ConnectionSemaphore. I may need to talk with @dicej about introducing changes in wasmtime-wasi to allow this.

Update: thanks to @dicej and @alexcrichton for helping figure this out. With the addition of https://github.com/bytecodealliance/wasmtime/pull/13549, we will be able to support this. The code for that can be see in https://github.com/dicej/spin/tree/connection-limits. We'll want to wait until the linked changes land in a release of wasmtime to adopt this, but the path forward is now clear.

lann:
Do we have an issue tracking p3 socket support?

crates/connection-semaphore/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+        let handle = tokio::spawn(async move {
+            let _p = sem2.acquire().await.expect("should eventually acquire");
+        });
+
+        drop(permit); // release so the spawned task can proceed
+        handle.await.expect("task should complete");
+    }
+
+    /// Verifies that when factor-specific is exhausted, acquire() releases
+    /// the global permit while waiting — so other connection types aren't blocked.
```
lann:
```suggestion
    /// Verifies that when factor-specific is exhausted, acquire() doesn't hold
    /// a global permit while waiting — so other connection types aren't blocked.
```

rylev:
> Do we have an issue tracking p3 socket support?

I do not see [any](https://github.com/spinframework/spin/issues?q=is%3Aissue%20p3).

crates/connection-semaphore/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// Acquire both configured semaphore slots, returning a permit that holds
+    /// them until dropped.
+    ///
+    /// When both a global and a factor-specific semaphore are configured, this
+    /// method acquires factor-specific first, then global, ensuring the global
+    /// permit is never held while blocking on a factor-specific backlog.
+    ///
+    /// If `wait_timeout` is configured and the permits cannot be acquired within
+    /// that duration, an error is returned.
+    pub async fn acquire(&self) -> anyhow::Result<ConnectionPermit> {
```
lann, line 62: I think a fast path check here would make it more obviously correct when `wait_timeout == Some(Duration::ZERO)` (ensuring it will try at least once) and might even be (slightly) faster on average since it won't need the syscall in `Instant::now`.

```suggestion
    pub async fn acquire(&self) -> anyhow::Result<ConnectionPermit> {
        // Fast path: all required permits are already available
        if let Ok(permit) = self.try_acquire_permits() {
            spin_telemetry::monotonic_counter!(
                outbound_connection_permits_acquired = 1,
                kind = self.factor,
                waited = false
            );
            return Ok(permit);
        }
```

lann (approved):

PR #3558, opened 2026-06-04
telemetry: Reduce cardinality of some otel span names

lann:
This PR removes a lot of detail from some span names.

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.56.0/specification/trace/api.md#span

> The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable. [...] Generality SHOULD be prioritized over human-readability.

Given that a large number of `#[instrument(...)]` attrs were being touched anyway, I took this opportunity to split most of them across two (or more) lines, including some lines (but not files) that weren't otherwise updated.


crates/telemetry/src/db.rs
lann, line 1: I spent some time trying to simplify this but everything I did broke one or more test cases below and the test cases all look pretty reasonable.

crates/factor-outbound-mqtt/src/host.rs
```diff
@@ -145,32 +142,26 @@ impl v3::HostConnectionWithStore for crate::MqttFactorData {
...
 
     #[instrument(name = "spin_outbound_mqtt.publish", skip(accessor, connection, payload), err(level = Level::INFO),
-        fields(otel.kind = "producer", otel.name = format!("{} publish", topic), messaging.operation = "publish",
-        messaging.system = "mqtt"))]
+        fields(
+            otel.kind = "producer",
+            otel.name = "publish",
+            messaging.operation = "publish",
+            messaging.system = "mqtt",
+            messaging.destination.name = topic,
```
lann, line 170: Moved topic here (per otel semantic conventions)

crates/factor-outbound-mqtt/src/host.rs
```diff
@@ -145,8 +145,13 @@ impl v3::HostConnectionWithStore for crate::MqttFactorData {
     }
 
     #[instrument(name = "spin_outbound_mqtt.publish", skip(accessor, connection, payload), err(level = Level::INFO),
-        fields(otel.kind = "producer", otel.name = format!("{} publish", topic), messaging.operation = "publish",
-        messaging.system = "mqtt"))]
+        fields(
+            otel.kind = "producer",
+            otel.name = "publish",
```
rylev: `publish` seems too general. I don't understand why using the `"spin_outbound_mqtt.publish"` name field would be inappropriate here?

crates/factor-outbound-mqtt/src/host.rs
```diff
@@ -218,8 +223,13 @@ impl v2::HostConnection for InstanceState {
     /// current trace context into the payload yourself.
     /// https://w3c.github.io/trace-context-mqtt/#mqtt-v3-recommendation.
     #[instrument(name = "spin_outbound_mqtt.publish", skip(self, connection, payload), err(level = Level::INFO),
-        fields(otel.kind = "producer", otel.name = format!("{} publish", topic), messaging.operation = "publish",
-        messaging.system = "mqtt"))]
+        fields(
+            otel.kind = "producer",
+            otel.name = "publish",
```
rylev: Same as above.

crates/factor-outbound-mysql/src/host.rs
```diff
@@ -159,14 +159,16 @@ impl<C: Client> v3::HostConnectionWithStore for MysqlFactorData<C> {
 impl<C: Client> v2::Host for InstanceState<C> {}
 
 impl<C: Client> v2::HostConnection for InstanceState<C> {
-    #[instrument(name = "spin_outbound_mysql.open", skip(self, address), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", db.address = Empty, server.port = Empty, db.namespace = Empty))]
+    #[instrument(name = "spin_outbound_mysql.open", skip(self, address), err(level = Level::INFO),
```
rylev, line 181: Nit: shouldn't we use `skip_all` instead of `skip(self, address)`?

crates/factor-outbound-mysql/src/host.rs
```diff
@@ -159,14 +159,16 @@ impl<C: Client> v3::HostConnectionWithStore for MysqlFactorData<C> {
...
+        fields(otel.kind = "client", db.system = "mysql", db.address = Empty, server.port = Empty, db.namespace = Empty))]
     async fn open(&mut self, address: String) -> Result<Resource<v2::Connection>, v2::Error> {
         let mut state = self.0.lock().await;
         state.otel.reparent_tracing_span();
         state.open_connection(&address).await.map(Resource::new_own)
     }
 
-    #[instrument(name = "spin_outbound_mysql.execute", skip(self, connection, params), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", otel.name = statement))]
+    #[instrument(name = "spin_outbound_mysql.execute", skip(self, connection, params), err(level = Level::INFO),
+        fields(otel.kind = "client", db.system = "mysql", otel.name = spin_telemetry::db::sql_span_name(&statement)))]
```
rylev: In general, I'm not convinced that using some sort of simplified query representation is better than name being  `"spin_outbound_mysql.execute"` and the statement being captured...

crates/factor-outbound-mysql/src/host.rs
```diff
@@ -184,7 +186,8 @@ impl<C: Client> v2::HostConnection for InstanceState<C> {
             .map_err(track_db_error_on_span)
     }
 
-    #[instrument(name = "spin_outbound_mysql.query", skip(self, connection, params), err(level = Level::INFO), fields(otel.kind = "client", db.system = "mysql", otel.name = statement))]
+    #[instrument(name = "spin_outbound_mysql.query", skip(self, connection, params), err(level = Level::INFO),
+        fields(otel.kind = "client", db.system = "mysql", otel.name = spin_telemetry::db::sql_span_name(&statement)))]
     async fn query(
```
rylev, line 218: I don't love that otel consumers lose the knowledge that this comes from the `query` method.

rylev (commented):
I'm really not sure the query parsing is worth the hassle. I agree with the goal of this PR, but I'm not convinced that we need to be this clever.

crates/telemetry/src/db.rs
```diff
@@ -0,0 +1,207 @@
...
+        .unwrap_or(rest.len());
+
+        let (token, rest) = rest.split_at(token_end);
+        self.rest = rest;
+        Some(token)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
calebschoepp: Is this implementation robust to CTEs, subqueries, etc.

I wonder if there is an easy way to fuzz/[property-based test](https://hegel.dev/) this

crates/factor-outbound-mysql/src/host.rs
calebschoepp, line 202: Do we need to do anything to obfuscate the statement in the attribute?

calebschoepp (commented):
Overall I think this PR is plugging a big gap in our tracing cardinality risk — I could be convinced to approve this ~as is.

However, I'm feeling vaguely uncomfortable about two things.

1. The implementation of `sql_span_name` seems like an endlessly deep tunnel of edge cases. It seems like the kind of code we're going to have to keep coming back to time after time to fix little things. I like that it is dependency free, but did we fully rule out that there isn't an existing library that could solve this problem for us?
2. It feels like we have a mix of OTel convention span names and "spin style" span names. Maybe this is just fine and we can't do much better, but I feel weird about them being mixed. (I'm also aware this is a bit outside the scope of your PR so feel free to ignore this concern)

crates/factor-outbound-mqtt/src/host.rs
lann: Oh that would probably be fine. I was just going for minimal changes: "{topic} publish" -> "publish". 🤷

crates/factor-outbound-mysql/src/host.rs
lann: It depends a lot on your tooling. It doesn't really matter for something like Honeycomb which gives you a lot of flexibility in how you visualize traces, but for e.g. Jaeger you only see a few specific attributes at the top level and span name is a big one.

crates/factor-outbound-mysql/src/host.rs
lann: That one is tricky. There really _shouldn't_ be anything sensitive in the statement, but of course that isn't going to stop people from doing it. My current opinion is that we should leave it up to the user's telemetry pipeline to deal with that.

crates/factor-outbound-mysql/src/host.rs
lann: That sounds fine but I don't really want to audit all arguments in this PR.

crates/factor-outbound-mysql/src/host.rs
lann: That's fair, though there is quiet a bit of other info available:

```
code.filepath: [...SNIP...]/crates/factor-outbound-mysql/src/host.rs
code.lineno: 57
code.namespace: spin_factor_outbound_mysql::host
```

crates/telemetry/src/db.rs
lann: CTEs: no; I think it would probably just report `WITH`.
Subqueries: a little; see `select_anonymous_table_has_no_target` below.

This is very much focused on lowering cardinality first and retaining a little bit of readability second.

lann:
I'm fine dropping the SQL stuff but if we aren't going to try to converge on otel conventions then we should come up with our own rationale for what we _do_ want to converge on.

lann:
The last commit here resolves most of the comments.

calebschoepp (approved):

itowlson:
@calebschoepp I rebased onto main and fixed conflicts with #3567.  Please re-review, and merge if you are still happy with it - thanks!

calebschoepp (approved):

PR #3564, opened 2026-06-10
Make the right features line up

rylev:
Currently `spin-oci` does not compile on its own but does compile when compiling the entire workspace. This is due to the cargo feature resolver being funky. We essentially need to force the version of `reqwest` that `docker-registry` uses to use the right feature.


vdice (approved):
🎉

PR #3567, opened 2026-06-12
Update OpenTelemetry dependencies and use experimental semantic convention constants

ChihweiLHBird:
### Summary

- Upgrade the OpenTelemetry dependency stack to the 0.29-compatible releases.
  - We can't upgrade to 0.30 cleanly due to a upstream breaking change in `opentelemetry_sdk::metrics::data::ResourceMetrics`.
- Replace hard-coded semantic convention attribute strings with constants from `opentelemetry-semantic-conventions`.
- Migrate outbound database and networking spans away from deprecated attribute names where the semantic conventions crate provides newer names. (may be **breaking**)
- Adjust OpenTelemetry SDK usage for the updated log processor resource API.

### Details

Continue previous work of compile-time references for the emitted attribute keys and makes future semantic convention migrations easier to identify during dependency updates.

The workspace enables `semconv_experimental` because Spin was already emitting experimental semantic convention attributes as string literals before this change, including attributes such as `http.response.body.size` and `db.namespace`. Enabling the feature does not introduce new telemetry concepts by itself; it lets the existing experimental attribute names come from the same upstream conventions crate as the stable attributes.

### Breaking Changes / Compatibility

This changes some emitted OpenTelemetry span attribute names as part of replacing deprecated semantic convention attributes with their newer equivalents. Consumers that collects these attributes should update their telemetry rules accordingly.

Notable replacements include:

- `db.system` -> `db.system.name`
- `db.address` (not a semantic convention) -> `server.address`

### DIFF
--- a/crates/trigger-http/Cargo.toml
+++ b/crates/trigger-http/Cargo.toml
@@ -15,7 +15,7 @@ http = { workspace = true }
 http-body = { workspace = true }
 http-body-util = { workspace = true }
 hyper = { workspace = true }
-hyper-util = { workspace = true }
+hyper-util = { workspace = true, features = ["server-auto"] }
 pin-project-lite = { workspace = true }
 rand.workspace = true
 rustls = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -290,21 +290,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-global-executor"
-version = "2.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "05b1b633a2115cd122d73b955eadd9916c18c8f510ec9cd1686404c60ad1c29c"
-dependencies = [
- "async-channel 2.3.1",
- "async-executor",
- "async-io",
- "async-lock",
- "blocking",
- "futures-lite 2.5.0",
- "once_cell",
-]
-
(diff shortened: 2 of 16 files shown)


crates/trigger-http/Cargo.toml
```diff
@@ -15,7 +15,7 @@ http = { workspace = true }
 http-body = { workspace = true }
 http-body-util = { workspace = true }
 hyper = { workspace = true }
-hyper-util = { workspace = true }
+hyper-util = { workspace = true, features = ["server-auto"] }
```
calebschoepp, line 18: What's up with this change?

ChihweiLHBird: `spin-trigger-http` uses `hyper_util::server::conn::auto::Builder`, which is behind `hyper-util`'s `server-auto` feature.

Before this PR, `opentelemetry-proto 0.28` happened to enable that feature indirectly via its sub-dependencies path. After the OTel upgrade, that feature is no longer enabled, so `spin-trigger-http` needs to declare the feature it directly uses.

This showed up as a CI failure on an earlier version of this PR, before I force-pushed it away:
https://github.com/spinframework/spin/actions/runs/27437706149/job/81103748687

calebschoepp (approved):
LGTM!

PR #3570, opened 2026-06-13
Fix HTTP test port race and self-request origin

ChihweiLHBird:
## Summary

Fixes flaky HTTP integration test startup caused by a test harness port race.

The harness previously selected a port by binding `localhost:0`, reading the assigned port, dropping that listener, and later launching `spin up --listen 127.0.0.1:{port}`. In parallel test runs, another process could claim that port before this Spin instance bound it. The old readiness check only proved that something accepted TCP connections on the port, so a test could accidentally send requests to the wrong Spin app and see a spurious `404`.

### Production code

- Track the actual bound HTTP listener address and use it for self-request origin setup.

### Test framework

- Start Spin with `--listen 127.0.0.1:0`, then parse the actual bound port from this Spin process's startup output.
- Report early Spin startup failures with stdout/stderr.

## Why the server change is needed

After the test harness switched to `127.0.0.1:0`, `HttpServer::listen_addr` could remain `127.0.0.1:0` even though the actual listener was bound to an assigned port. Self requests must use the real bound address, not the configured `:0` address. The server now stores `listener.local_addr()` after binding and uses that address when configuring `SelfRequestOrigin`.

For in-process tests, `InProcessSpin` builds a server and drives `handle()` directly without calling `serve()`, so no socket address is recorded. That path falls back to the configured listen address.

### DIFF
--- a/crates/trigger-http/src/server.rs
+++ b/crates/trigger-http/src/server.rs
@@ -3,7 +3,7 @@ use std::{
     future::Future,
     io::{ErrorKind, IsTerminal},
     net::SocketAddr,
-    sync::Arc,
+    sync::{Arc, OnceLock},
     time::Duration,
 };
 
@@ -60,8 +60,14 @@ pub const MAX_RETRIES: u16 = 10;
 
 /// An HTTP server which runs Spin apps.
 pub struct HttpServer<F: RuntimeFactors> {
-    /// The address the server is listening on.
+    /// The address the server was configured to listen on (the `--listen` value).
     listen_addr: SocketAddr,
+    /// The address the server is actually bound to, captured once after binding.
+    ///
+    /// This can differ from `listen_addr` when the OS assigns the port — e.g.
+    /// `--listen 127.0.0.1:0` or `--find-free-port`. Self-request origins must use
+    /// this real address rather than the configured one.
+    local_addr: OnceLock<SocketAddr>,
     /// The TLS configuration for the server.
     tls_config: Option<TlsConfig>,
     /// The maximum buffer size for an HTTP1 connection.
@@ -151,6 +157,7 @@ impl<F: RuntimeFactors> HttpServer<F> {
             .collect::<anyhow::Result<_>>()?;
         Ok(Self {
             listen_addr,
+            local_addr: OnceLock::new(),
             tls_config,
             find_free_port,
             router,
@@ -206,6 +213,8 @@ impl<F: RuntimeFactors> HttpServer<F> {
                 }
             })?
         };
+
+        let _ = self.local_addr.set(listener.local_addr()?);
 
         if let Some(tls_config) = self.tls_config.clone() {
             self.serve_https(listener, tls_config).await?;
@@ -366,6 +375,10 @@ impl<F: RuntimeFactors> HttpServer<F> {
         }
     }
 
+    fn get_local_addr(&self) -> SocketAddr {
+        self.local_addr.get().copied().unwrap_or(self.listen_addr)
+    }
+
     async fn respond_wasm_component(
         self: &Arc<Self>,
         req: Request<Body>,
@@ -386,7 +399,9 @@ impl<F: RuntimeFactors> HttpServer<F> {
             .context(
             "The wasi HTTP trigger was configured without the required wasi outbound http support",
         )?;
-        let origin = SelfRequestOrigin::create(server_scheme, &self.listen_addr.to_string())?;
+
+        let self_addr = self.get_local_addr();
+        let origin = SelfRequestOrigin::create(server_scheme, &self_addr.to_string())?;
         outbound_http.set_self_request_origin(origin);
         outbound_http.set_request_interceptor(OutboundHttpInterceptor::new(self.clone()))?;
 
--- a/tests/testing-framework/src/runtimes/spin_cli.rs
+++ b/tests/testing-framework/src/runtimes/spin_cli.rs
@@ -56,13 +56,15 @@ impl SpinCli {
         spin_config: SpinConfig,
         env: &mut TestEnvironment<R>,
     ) -> anyhow::Result<Self> {
-        let port = get_random_port()?;
         let mut spin_cmd = Command::new(spin_config.binary_path);
         let child = spin_cmd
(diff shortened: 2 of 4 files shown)


crates/trigger-http/src/server.rs
```diff
@@ -386,7 +395,11 @@ impl<F: RuntimeFactors> HttpServer<F> {
             .context(
             "The wasi HTTP trigger was configured without the required wasi outbound http support",
         )?;
-        let origin = SelfRequestOrigin::create(server_scheme, &self.listen_addr.to_string())?;
+        // Fall back to the configured `listen_addr` when the bound address hasn't been
+        // recorded: the in-process runtime (`InProcessSpin`) drives `handle` directly via
```
itowlson: InProcessSpin is a test concern; I'm not convinced we should be modifying the production trigger for it.

crates/trigger-http/src/server.rs
```diff
@@ -582,8 +595,11 @@ impl<F: RuntimeFactors> HttpServer<F> {
         }
     }
 
-    fn print_startup_msgs(&self, scheme: &str, listener: &TcpListener) -> anyhow::Result<()> {
-        let local_addr = listener.local_addr()?;
+    fn print_startup_msgs(&self, scheme: &str) -> anyhow::Result<()> {
+        let local_addr = self
+            .local_addr
+            .get()
+            .context("listen address not recorded before printing startup messages")?;
```
itowlson: This seems like it's introducing temporal coupling with some other function recording an address into the `.local_addr` field.  Why can't we continue printing `listener.local_addr()` here?

tests/testing-framework/src/runtimes/spin_cli.rs
```diff
@@ -278,9 +288,25 @@ enum IoMode {
     None,
 }
 
-/// Uses a track to ge a random unused port
-fn get_random_port() -> anyhow::Result<u16> {
-    Ok(std::net::TcpListener::bind("localhost:0")?
-        .local_addr()?
-        .port())
+fn parse_serving_port(output: &str) -> Option<u16> {
+    parse_plain_serving_port(output).or_else(|| parse_json_serving_port(output))
```
itowlson, line 292: Wait, don't we know whether it's printing JSON or plain text?

itowlson (commented):
I'll defer to @rylev on the test stuff, but I'd like to understand how often we are seeing port race failures - I don't think I've ever seen one myself.  Unless you're being actively affected by them, I would tend to be conservative with changes.

I can see the issue with self-requests though.  So we probably need to fix that and if the rest of it comes along for the ride well fair enough.

ChihweiLHBird:
@itowlson Thanks for the feedback! Actually, I saw more than once impact of this issue, sometimes it's very implicit, shown as 404 not found error, when a test code hits another test spin instance which doesn't have the expected path.

crates/trigger-http/src/server.rs
ChihweiLHBird: In theory, if removing the fallback, the `InProcessSpin` code path may panic... but if this is really not a concern, I can remove the fallback.

tests/testing-framework/src/runtimes/spin_cli.rs
ChihweiLHBird: There was a test I added in the previous PR for testing JSON output, and all other tests are plain text. I can update the test code to force every test to output JSON, which is easier to handle. How does that sound?

crates/trigger-http/src/server.rs
ChihweiLHBird: Just wanted unified the way to get `local_addr` from self (single source of truth) and avoid re-computing it. But yes, I can revert it back to `listener.local_addr()` if that style is preferred. What do you think?

tests/testing-framework/src/runtimes/spin_cli.rs
itowlson: I was thinking more that we would be able to know when calling `parse_serving_port` whether the output was set to JSON.  But maybe that info is buried in some Command arg that appears only in a callback somewhere and is inaccessible.  It's fine to leave this as is.

itowlson:
> I saw more than once impact of this issue, sometimes it's very implicit, shown as 404 not found error, when a test code hits another test spin instance which doesn't have the expected path.

Interesting.  If it's happening that often, I'm surprised we don't get a lot more flakes in CI - I mean, yes, we do get flakes, but if they show up in manual testing then I'd expect them to absolutely wreck CI, which runs about a gazillion times on every PR.  But of course we'd want it to be correct anyway!

crates/trigger-http/src/server.rs
itowlson: Yeah, given that we _have_ the `Listener`  at the moment we call `print_startup_msgs` I'd be inclined to use that as the source of truth rather than relying on another variable being hopefully set up in `self`.

crates/trigger-http/src/server.rs
itowlson: I would say it's fine to just remove the mention of InProcessSpin.  I'd probably pull `self.local_addr.get().copied().unwrap_or(self.listen_addr)` into a `fn addr(&self) -> SocketAddr` or something like that and call that without worrying about InProcessSpin - being infallible is reason enough.

ChihweiLHBird: Ah got it, I can remove the mention of InProcessSpin. And let me try the function style

ChihweiLHBird: Just removed and wrapped in a function.

ChihweiLHBird:
> > I saw more than once impact of this issue, sometimes it's very implicit, shown as 404 not found error, when a test code hits another test spin instance which doesn't have the expected path.
> 
> Interesting. If it's happening that often, I'm surprised we don't get a lot more flakes in CI - I mean, yes, we do get flakes, but if they show up in manual testing then I'd expect them to absolutely wreck CI, which runs about a gazillion times on every PR. But of course we'd want it to be correct anyway!

Actually, I saw that from one CI run in my another PR and started investigating into it. I didn't catch it locally

itowlson (approved):

PR #3571, opened 2026-06-14
Miscellaneous integration test fixes

ChihweiLHBird:
Three small test-reliability fixes:

- **WASI HTTP p3 fixture:** `test_wasi_http_p3_double_echo` ran the p2 fixture; point it at `wasi-http-p3-streaming`.
- **Plugin uninstall:** the cleanup `Drop` now sets `SPIN_DATA_DIR=./plugins` (matching install) so uninstall doesn't fail/panic.
- **Process teardown:** `kill_process` now SIGTERMs with a ~5s grace, escalates to SIGKILL, and reaps with `wait()` — no more stray/zombie Spin processes (Windows hard-kills + reaps).

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1223,8 +1223,12 @@ route = "/..."
         impl Drop for Uninstaller<'_> {
             fn drop(&mut self) {
                 let mut uninstall = std::process::Command::new(spin_binary());
-                uninstall.args(["plugins", "uninstall", "example"]);
-                self.0.run_in(&mut uninstall).unwrap();
+                uninstall
+                    .args(["plugins", "uninstall", "example"])
+                    .env("SPIN_DATA_DIR", "./plugins");
+                self.0
+                    .run_in(&mut uninstall)
+                    .expect("failed to uninstall test plugin");
             }
         }
         let _u = Uninstaller(&env);
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1362,7 +1362,7 @@ route = "/..."
     #[test]
     fn test_wasi_http_p3_double_echo() -> anyhow::Result<()> {
         wasi_http_echo(
-            "wasi-http-p2-streaming",
+            "wasi-http-p3-streaming",
             "double-echo".into(),
             Some("echo".into()),
         )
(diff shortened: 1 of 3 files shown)


tests/testing-framework/src/runtimes/spin_cli.rs
```diff
@@ -257,15 +257,36 @@ pub struct SpinConfig {
...
+                }
+            }
+        }
+        log::warn!("Spin did not exit within ~5s of SIGTERM; escalating to SIGKILL");
     }
+    // Hard-kill the process. On Windows this is the only termination
+    // (TerminateProcess); on Unix it's the SIGKILL
+    let _ = process.kill();
+
+    let _ = process.wait();
```
itowlson: I am not sure about this.  It seems like if a process failed to be killed for some reason then the test suite would not complete and you'd _still_ be left with a zombie Spin process.  Could you expand on what problem you're seeing that this solves?

tests/integration.rs
```diff
@@ -1223,8 +1223,12 @@ route = "/..."
         impl Drop for Uninstaller<'_> {
             fn drop(&mut self) {
                 let mut uninstall = std::process::Command::new(spin_binary());
-                uninstall.args(["plugins", "uninstall", "example"]);
-                self.0.run_in(&mut uninstall).unwrap();
+                uninstall
+                    .args(["plugins", "uninstall", "example"])
+                    .env("SPIN_DATA_DIR", "./plugins");
```
itowlson, line 1228: Good catch.  But now I wonder why I've never seen a leftover `example` plugin in my current directory.  cc @rylev in case I am missing something

tests/integration.rs
```diff
@@ -1362,7 +1366,7 @@ route = "/..."
     #[test]
     fn test_wasi_http_p3_double_echo() -> anyhow::Result<()> {
         wasi_http_echo(
-            "wasi-http-p2-streaming",
+            "wasi-http-p3-streaming",
```
itowlson, line 1369: Ouch.  Good find.

itowlson (commented):
Most of this looks good but I'm not sure the `kill_process` changes are worth the complexity (and am I deeply unsure about the "wait for kill" thing).  Could you say more about what problem this is solving please?  Thanks!

tests/testing-framework/src/runtimes/spin_cli.rs
ChihweiLHBird: A hard kill usually force the program to exit, but in some rare states, like uninterruptible sleep, it may stop there, but yes, I can remove this to address the concern.

ChihweiLHBird: Removed in https://github.com/spinframework/spin/pull/3571/commits/a3a959a8de25704a7b34de88baca6f42fd1abd9a

itowlson (approved):
LGTM - thanks!

PR #3573, opened 2026-06-14
Bring back Python smoke test

### LINKED ISSUE #2457: Python smoke test is failing with build errors for unknown reasons.
The `integration_tests::http_python_template_smoke_test` test [is failing](https://github.com/fermyon/spin/actions/runs/8702958149/job/23877188850?pr=2456#step:10:2246) repeatedly with an error about mismatching hashes when resolving the `componentize-py` dependency. Strangely it looks like the download of the dependency is ending before the package is fully downloading leading to the hash not matching what is expected. We have ignored the test in https://github.com/fermyon/spin/pull/2456. In order to unignore the test, we'll need to figure out why it is failing.

itowlson:
Fixes #2457.

In case something has miraculously happened to make it fix itself in the last two years.

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -724,29 +724,29 @@ mod integration_tests {
 
     #[test]
     #[cfg(feature = "extern-dependencies-tests")]
-    #[ignore = "https://github.com/spinframework/spin/issues/2457"]
-    // TODO: Check why python is not picking up the spin_sdk from site_packages
-    // Currently installing to the local directory to get around it.
     fn http_python_template_smoke_test() -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


itowlson:
Survived its first CI pass. Re-running to see if that's repeatable or if it was just a beautiful dream.

itowlson:
This has now passed _twice,_ which is at least once more than most of my tests, so... I am not sure what the net change has been but it seems to work now.  So I suggest we merge it, wait for just long enough for everyone to forget about this PR, and then act surprised when the test starts flaking again.

dicej (approved):

vdice (approved):
🎉

PR #3574, opened 2026-06-15
chore(deps): bump/use upstream rust-oci-client

vdice:
- Switches/bumps the `oci-client` (formerly `oci-distribution`) dependency to the latest commit from the upstream repo (https://github.com/oras-project/rust-oci-client)
- Replaces `oci_distribution` with `oci_client` references per the package naming update

~~As noted in the comments, there is a PR to land the needed functionality upstream.  If that lands or if we relocate the logic into dependent systems (the aka and cloud plugins), then we can eliminate the fork.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,7 +40,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.15",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


itowlson (approved):

itowlson:
@vdice Is this draft because of waiting on that oci-client PR?

vdice:
@itowlson yes, was optimistic it might merge soon.  Would we rather merge this one and just follow-up with using upstream once ready?

itowlson:
I've pinged the upstream reviewer to ask if they can merge it.  Let's give it a day or two, and if still no joy then, land this as is.

PR #3576, opened 2026-06-17
Every month, update Wasmtime to latest pre-release

itowlson:
The early warning system!

This can't currently trigger workflows on the PR it creates - a human needs to close-reopen to do that.  A custom token would solve that.  My guess is there's stuff like DCO that will cause ructions too...


itowlson:
I messed up, sorry!

PR #3577, opened 2026-06-17
Every month, update Wasmtime to latest pre-release

itowlson:
The early warning system!

This can't currently trigger workflows on the PR it creates - a human needs to close-reopen to do that. A custom token would solve that. My guess is there's stuff like DCO that will cause ructions too...

_I think I messed up less this time.  My local test of the bump PR is failing with a 1.94 Rust requirement which suggests that the script is doing something plausible..._

### DIFF
--- a/.github/workflows/bump-wasmtime.yml
+++ b/.github/workflows/bump-wasmtime.yml
@@ -1,0 +1,40 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Bump wasmtime (prerelease)
+on:
+  schedule:
+    - cron: 0 0 9 * * # 9th of each month, at 00:00 UTC
+  workflow_dispatch:
+
+jobs:
+  bump-wasmtime:
+    name: Bump Wasmtime to prerelease
+    runs-on: "ubuntu-22.04"
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Get latest Wasmtime release
+        id: wasmtime-release
+        run: |
+          LAST_WASMTIME=$(git ls-remote https://github.com/bytecodealliance/wasmtime 'refs/heads/release-*' | awk '{print $2}' | sed -e "s/^refs\/heads\/release-//" | sort -V | tail -n1)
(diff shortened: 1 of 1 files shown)


itowlson:
Marked this ready for review despite lack of vast confidence.  I guess the best way to try it is to get it into main and start creating PRs, and see where the pain points are.

Note this does not do the update to RTM later in the month.  If we trial-and-error this one into shape then doing one for RTM should follow.

.github/workflows/bump-wasmtime.yml
```diff
@@ -0,0 +1,40 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Bump wasmtime (prerelease)
+on:
+  schedule:
+    - cron: 0 0 9 * *
```
karthik2804: Can we add a comment for the non-cron speakers like me?
```suggestion
    - cron: 0 0 9 * * # Once a month, on the 9th at midnight UTC
```

.github/workflows/bump-wasmtime.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  bump-wasmtime:
+    name: Bump Wasmtime to prerelease
+    runs-on: "ubuntu-22.04"
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v3
```
vdice:
```suggestion
      - uses: actions/checkout@v6
```

Might as well start with the latest

vdice (commented):
Not sure if you've already tested this, but you can do so by pushing to your fork and triggering, etc.  Otherwise, as usual, happy to fix up as needed after testing here on origin.

itowlson:
@vdice yeah, I tested it on my fork (a sea of red attests to my triumph) - here's a PR it generated: https://github.com/itowlson/spin/pull/5

itowlson:
@vdice but my fork doesn't have all the DCO stuff and all that so it's not a super reliable indicator of the woes we're going to hit in the real repo

fibonacci1729 (approved):
Release the Kraken.

PR #3579, opened 2026-06-17
Fix IPv6 allowed outbound host parsing

ChihweiLHBird:
Fix parsing of `allowed_outbound_hosts` entries containing IPv6 literals or IPv6 CIDR ranges without an explicit port.

Previously, `rsplit_once(':')` could mistake an IPv6 address segment for a port. These entries now use the scheme’s default port:

- `http://[::1]`
- `https://[::1]`
- `http://ff00::/8`

The updated parsing also:
- Validates and normalizes wildcard domain suffixes.
- Rejects malformed paths and multiple trailing slashes while allowing one empty trailing slash.
- Keeps RFC 3986-compatible empty ports, such as `http://example.com:/`, accepted.

Added regression coverage for IPv6 parsing and matching, malformed authorities, wildcard domains, paths, and error messages.

### DIFF
--- a/crates/outbound-networking-config/src/allowed_hosts.rs
+++ b/crates/outbound-networking-config/src/allowed_hosts.rs
@@ -126,15 +126,25 @@ impl AllowedHostConfig {
                 ),
             }
         };
-        let (host, rest) = rest.rsplit_once(':').unwrap_or((rest, ""));
-        let port = match rest.split_once('/') {
-            Some((port, path)) => {
-                if !path.is_empty() {
-                    bail!("{url:?} has a path but is not allowed to");
+        let (host, port) = match rest.rsplit_once(':') {
+            None => (rest, ""),
+            Some((h, "")) => (h, ""),
+            Some((h, tail)) => {
+                if HostConfig::parse(rest).is_ok() {
+                    (rest, "")
+                } else {
+                    let port = match tail.split_once('/') {
+                        Some((port, path)) => {
+                            if !path.is_empty() {
+                                bail!("{url:?} has a path but is not allowed to");
+                            }
+                            port
+                        }
+                        None => tail,
+                    };
+                    (h, port)
                 }
-                port
             }
-            None => rest,
         };
 
         let port = PortConfig::parse(port, scheme)
@@ -267,7 +277,7 @@ impl HostConfig {
             return Ok(Self::Cidr(net));
         }
 
-        host = host.trim_end_matches('/');
+        host = host.strip_suffix('/').unwrap_or(host);
         if host.contains('/') {
             bail!("must not include a path");
         }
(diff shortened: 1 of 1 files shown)


crates/outbound-networking-config/src/allowed_hosts.rs
```diff
@@ -126,15 +126,19 @@ impl AllowedHostConfig {
...
-            Some((port, path)) => {
-                if !path.is_empty() {
-                    bail!("{url:?} has a path but is not allowed to");
+        let (host, port) = match rest.rsplit_once(':') {
+            None => (rest, ""),
+            Some((h, "")) => (h, ""),
+            Some((h, tail)) => {
+                let port_candidate = tail.strip_suffix('/').unwrap_or(tail);
+                if !port_candidate.contains('/')
+                    && PortConfig::parse(port_candidate, scheme).is_ok()
```
rylev: This is correct but does make error messages a bit worse. For example, for the string "example.com:99999", instead of the errror being `Invalid allowed host port "99999"`, it ends up as `Invalid allowed host "example.com:99999"`.

crates/outbound-networking-config/src/allowed_hosts.rs
```diff
@@ -1210,4 +1228,315 @@ mod test {
...
+            "http://*.[2001:db8::1]",
+        ] {
+            assert!(
+                AllowedHostConfig::parse(bad).is_err(),
+                "expected {bad:?} to be rejected because wildcard suffixes must be domain names"
+            );
+        }
+    }
+
+    #[ignore = "FIXME"]
```
rylev: I think we should remove this and open an issue instead since the issue was pre-existing. The issue should include this test though.

ChihweiLHBird: I just checked and empty ports in URL like this is permitted by RFC 3986, I will clean this up.
```
authority = [ userinfo "@" ] host [ ":" port ]
port      = *DIGIT
```

crates/outbound-networking-config/src/allowed_hosts.rs
ChihweiLHBird: Let me look into this

ChihweiLHBird: It should has been fixed now, thanks!

crates/outbound-networking-config/src/allowed_hosts.rs
```diff
@@ -1210,4 +1234,334 @@ mod test {
...
+            "http://*.example.com:8O8O",     // letter 'O' instead of zero
+        ] {
+            assert!(
+                AllowedHostConfig::parse(bad).is_err(),
+                "expected {bad:?} to be rejected, not silently accepted"
+            );
+        }
+    }
+
+    /// Bonus (finding #4): validating the wildcard domain also normalises its
```
rylev: Please remove all references to `finding #`. These don't mean anything to the reader.

ChihweiLHBird: Oh I'm sorry, will do

ChihweiLHBird: Just pushed an update to consolidate tests and remove all unnecessary sloppy comments.

rylev (approved):

Issue #3566, opened 2026-06-11
Add WASI 0.3.0 support


dicej:
WASI 0.3.0 was finalized today, and Wasmtime 46 will support it by default.  So should Spin.

Note that, just as with 0.2.0, we'll continue supporting the 0.3.0-rc-2026-03-15 snapshot indefinitely as well.

dicej:
> This is also a good opportunity to consider whether Spin wants to pin to the 0.3.0 final snapshot or continue being flexible with snapshot versions — the rc snapshot was the only 0.3.x version with significant production deployment, and breaking it quietly would strand WASI 0.3.0-rc modules that users have already built against Spin's WIT worlds.

Hence this part of [the issue description](https://github.com/spinframework/spin/issues/3566#issue-4642836201):

> Note that, just as with 0.2.0, we'll continue supporting the 0.3.0-rc-2026-03-15 snapshot indefinitely as well.

itowlson:
@dicej Do you reckon there's anything more to do after #3595 and #3603?  If so, I'm happy to (attempt to) help; otherwise, shall we close?

dicej:
Thanks for the reminder.  I think we're ready to close this.

PR #3583, opened 2026-06-17
Bump Wasmtime to v46.0.0 prerelease

spinframeworkbot:
Bumps Wasmtime to v46.0.0 prerelease

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,27 +4,18 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.26.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "59317f77929f0e679d39364702289274de2f0f0b22cbf50b2b8cff2169a0b27a"
 dependencies = [
- "gimli 0.31.1",
-]
-
-[[package]]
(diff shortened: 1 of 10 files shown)


itowlson:
@alexcrichton @dicej I updated the async host traits to the new way of doing generics, but there are a bunch of other errors I don't understand.  At first it seemed like only a couple, but I tried naively fixing this line:

```
ProxyHandler::new(handler_state, ProxyPre::P3(pre)),
```

and that unlocked a whole world of errors I had no idea what they were so I am guessing that was not the right thing!

I have kept my changes separate from the bot ones so if anyone has bandwidth to take a look at what I did and polish off what I didn't do, that would be awesome - thanks!

dicej:
@itowlson sorry for not publicizing it, but @rylev and I already did the work to update to Wasmtime 46 here: https://github.com/dicej/spin/tree/connection-limits.  We had to do it as part of the connection limit PR he was working on, although we ended up deferring that part until 46 is actually released.

dicej:
Also, this will inevitably depend on #3566 since 46 will support only WASI 0.3.0 final and _not_ 0.3.0-rc-2026-03-15, meaning we'll need to crate a 0.3.0-rc-2026-03-15 shim on top of the 0.3.0 implementation similar to what we did for the 0.2.0 RCs.  All of which I was planning to do on top of the branch I linked to above, but I'll be out on vacation for the next week and a half, so feel free to tackle that yourself if you're motivated.

PR #3590, opened 2026-06-25
Set rustls crypto provider explicitly

### LINKED ISSUE #3104: Avoid automatic TLS crypto provider detection
As can be seen in https://github.com/spinframework/spin/pull/2830 and https://github.com/spinframework/spin/pull/3103 - rustls has [automatic detection of the crypto provider](https://github.com/rustls/rustls/blob/88dccec476595321af9a32708ffbf63200701a98/rustls/src/crypto/mod.rs#L261-L282) used for TLS crypto operations. This can very easily go wrong in a complex project like Spin where many different dependencies come together to determine which features are enabled. 

Instead, we should explicitly set the crypto provider once so that automatic detection does not happen. The only question is where this should happen. The crypto provider is a global variable which should only be set once. Therefore, it probably makes most sense to set it in main and allow each subcrate of Spin to be ignorant of this entire situation.

ChihweiLHBird:
Fixes #3104.

## Summary

- Install Spin's process-wide rustls crypto provider at startup.
- Add a shared `spin-trigger` helper for the CLI and trigger harness paths.
- Switch `reqwest` to the no-provider rustls feature so it uses the process default.

A follow up PR can change the provider from `ring` to `aws-lc-rs` and optionally upgrade `reqwest` to 0.13


Cargo.toml
```diff
@@ -170,7 +170,7 @@ quote = "1"
 rand = "0.9"
 redis = "0.32.5"
 regex = "1"
-reqwest = { version = "0.12", features = ["stream", "blocking", "rustls-tls-native-roots"] }
+reqwest = { version = "0.12", features = ["stream", "blocking", "rustls-tls-native-roots-no-provider"] }
```
lann: I think this could be a breaking change for consumers of spin crates that use this workspace dependency.

Is this actually necessary? It looks like reqwest will use an already-installed provider: https://github.com/seanmonstar/reqwest/blob/797df2b96a88ee49a636337f87beebe87f6212fe/src/async_impl/client.rs#L583-L593

ChihweiLHBird: This is not strictly needed, let me try changing it back.

lann (approved):

src/bin/spin.rs
```diff
@@ -2,6 +2,7 @@ use spin_cli::subprocess::ExitStatusError;
 
 #[tokio::main]
 async fn main() {
+    spin_trigger::crypto::install_default_crypto_provider();
```
itowlson: It's not clear to me why this is in `spin_trigger`.

crates/trigger/src/crypto.rs
```diff
@@ -0,0 +1,15 @@
...
+
+/// Install Spin's process-wide rustls crypto provider.
+///
+/// This is idempotent for Spin's own duplicate calls from `main` and the trigger,
+/// but fails loudly if something else installed a rustls provider first.
+pub fn install_default_crypto_provider() {
+    INSTALL_DEFAULT_CRYPTO_PROVIDER.call_once(|| {
+        rustls::crypto::ring::default_provider()
+            .install_default()
+            .unwrap_or_else(|_| panic!("failed to install rustls ring crypto provider"));
```
itowlson: Is this equivalent to `.expect("...")` or am I showing my ignorance? _grin_

crates/trigger/src/cli.rs
```diff
@@ -180,6 +180,7 @@ pub struct NoCliArgs;
 impl<T: Trigger<B::Factors>, B: RuntimeFactorsBuilder> FactorsTriggerCommand<T, B> {
     /// Create a new TriggerExecutorBuilder from this TriggerExecutorCommand.
     pub async fn run(self) -> Result<()> {
+        crate::crypto::install_default_crypto_provider();
```
itowlson: This would presumably already have been called at the top of `main`.  Is the idea of calling it again here to support custom triggers/the SpinKube shim/etc.

src/bin/spin.rs
ChihweiLHBird: Or it can be in a dedicated crate?

crates/trigger/src/cli.rs
ChihweiLHBird: Yes, I think custom trigger bypasses the spin's `main`, like https://github.com/spinframework/spin/blob/main/examples/spin-timer/src/main.rs

crates/trigger/src/crypto.rs
ChihweiLHBird: oops

src/bin/spin.rs
itowlson: Yeah, maybe.  I don't think there's a good place for it at the moment - _maybe_ `spin-common`, but that's meant to be reserved for helpers and conventions, and I am cautious about putting load-bearing state in there.  I am not sure.  It's just that if I thought "where is that default crypto thingy" then `spin-trigger` would not be on my list of places to look.  @lann any thoughts?

ChihweiLHBird: I put it into dedicated crate called `spin-tls`. Let me know if this works/looks better.

lann (approved):

PR #3595, opened 2026-07-02
Update to Wasmtime 46 and WASI 0.3.0, retaining support for 0.3.0-rc-2026-03-15

dicej:
The first commit just updates Wasmtime; the second commit adds shims to support 0.3.0-rc-2026-03-15.

~This is a draft for now, while I investigate an odd regression in `test_max_memory_size_violated`.~ EDIT: this seems to be due to how my local version of Rust compiles the guest code rather than anything that changed in Spin or Wasmtime.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,9 @@ jobs:
   dependency-review:
     name: Review new dependencies for known vulnerabilities
     runs-on: "ubuntu-22.04"
-    if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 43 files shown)


alexcrichton (approved):
Thanks for doing this! Clearly a lot of work went into this and I'm glad at least that it all worked out without having any hiccups to address upstream in addition to this. I looked over everything and it all looks good to me 👍 

The one bit I'm not the most comfortable reviewing is the CI adjustments, but if you've already discussed that with others that sounds reasonable too

itowlson (approved):
Maintainer rubber-stamp for Alex's review!

PR #3597, opened 2026-07-05
Miscellaneous Cleanups

ChihweiLHBird:
Some small cleanups

- Replace manually implemented `partition_results` with [`partition_result` from `itertools`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.partition_result).
  - Upgrade itertools to 0.15
- Replace manually implemented bytes parsing with [bytes::Buf](https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html).
- Using the file URL parsing from reqwest in tests (I can remove this if it seems to be too annoying for tests)

### DIFF
--- a/crates/factor-outbound-pg/src/types/interval.rs
+++ b/crates/factor-outbound-pg/src/types/interval.rs
@@ -36,7 +36,7 @@ impl ToSql for Interval {
 impl FromSql<'_> for Interval {
     fn from_sql(
         _ty: &Type,
-        raw: &'_ [u8],
+        mut raw: &'_ [u8],
     ) -> std::result::Result<Self, Box<dyn std::error::Error + Sync + Send>> {
         const EXPECTED_LEN: usize = size_of::<i64>() + size_of::<i32>() + size_of::<i32>();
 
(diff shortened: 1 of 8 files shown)


crates/factor-outbound-pg/src/types/interval.rs
```diff
@@ -36,20 +36,19 @@ impl ToSql for Interval {
...
 
-        let (micro_bytes, rest) = raw.split_at(size_of::<i64>());
-        let (day_bytes, rest) = rest.split_at(size_of::<i32>());
-        let month_bytes = rest;
-        let months = i32::from_be_bytes(month_bytes.try_into().unwrap());
-        let days = i32::from_be_bytes(day_bytes.try_into().unwrap());
-        let micros = i64::from_be_bytes(micro_bytes.try_into().unwrap());
+        use bytes::Buf;
+
+        let micros = raw.get_i64();
```
itowlson, line 49: Nice!  TIL

itowlson (approved):
Always here for some itertools.  Thanks!

PR #3599, opened 2026-07-07
Parse WAGI host authorities with http::uri::Authority

ChihweiLHBird:
## Summary

- Replace WAGI host parsing's ad hoc `splitn(':')` logic with `http::uri::Authority`.
- Add focused tests for host parsing precedence and edge cases.

## Behavior Change

- Bracketed IPv6 authorities such as `[::1]:3000` are now parsed correctly.
- Invalid authorities and empty-host authorities such as `:8443` are now ignored instead of partially updating the parsed host/port pair.

### DIFF
--- a/crates/http/src/wagi/mod.rs
+++ b/crates/http/src/wagi/mod.rs
@@ -8,6 +8,7 @@ use http::{
     HeaderMap, HeaderValue, Response, StatusCode,
     header::{HOST, HeaderName},
     request::Parts,
+    uri::Authority,
 };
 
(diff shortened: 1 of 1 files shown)


itowlson:
@ChihweiLHBird Is this causing you problems, or is it something you noticed and thought should be tidied/improved?  I ask because we provisionally have WAGI on a deprecation path (https://github.com/spinframework/spin/issues/3520), so we'd not normally be investing in quality-of-life changes to it.  However, if you're using WAGI, that's valuable data for _not_ deprecating it (and of course in that case happy to review the PR).

PR #3601, opened 2026-07-08
Add async MySQL to deny adapter

itowlson:
COMPUTER SAYS NO


fibonacci1729 (approved):

itowlson:
~Ugh this was meant to be a draft because it wasn't building right.  Sorry Brian.  (I guess that's the trouble with building the binary locally - you can't see if it does what it says on the tin.)~ mixup

itowlson:
~Although apparently now it is building right with no changes.  So... maybe some sort of weird Cargo cache thing?  I will test main, and send a new PR if something has gone awry.~ mixup

itowlson:
I'm an idiot.  Got this mixed up with 3603.  Sorry.

PR #3602, opened 2026-07-08
Bump middleware interface to 0.3.0

fibonacci1729:


itowlson:
I think you'll need to update it in the tests, which means bringing in the WITs, etc. etc. (which is what I was doing when I hit my deny adapter woe, but happy to leave this with you).

fibonacci1729:
Eh seems like youre further down the path on this, so ill close this. I thought this was a quick win I could do quickly on the side in between compiles.

PR #3603, opened 2026-07-08
WASI P3 - non-core migration bits

itowlson:
~WIP - currently the deny adapter does not build with a bunch of weird errors - pushing for folks who know bindgen better to take a look (thanks!).~ Fixed and ready for review.


itowlson:
Bumpetty bump.

crates/capabilities/src/lib.rs
```diff
@@ -43,9 +43,11 @@ const ALLOWED_OUTBOUND_HOSTS: &[&str] = &[
     "spin:postgres/postgres@3.0.0",
     "spin:postgres/postgres@4.2.0",
     "spin:redis/redis@3.0.0",
+    "wasi:http/client@0.3.0",
     "wasi:http/client@0.3.0-rc-2026-03-15",
```
fibonacci1729, line 48: I just realized we are missing `spin:mysql/mysql@3.0.0` from this list. Would you mind including it here?

fibonacci1729 (approved):
One comment to resolve. Otherwise, looks good.

crates/capabilities/src/lib.rs
itowlson: Guess who undoubtedly only looked at the directories!  Thanks for the catch

fibonacci1729 (approved):

PR #3604, opened 2026-07-08
Re-enable the review-dependencies action with the problem GHSA excluded

itowlson:
...wait, why am I doing this?  That action enrages me all the time


itowlson:
Oh, this isn't a fair test because I think I introduce any new dependencies here.  Let me fake something up.

itowlson:
Okay I think that has been a few tries.  I also regenerated the timer sample lockfile which I think is (at least part of) what caused the initial trouble, so hopefully this validates that the exclusion is working.

dicej (approved):

PR #3606, opened 2026-07-09
Fix wasmtime bump regexes

itowlson:
Fixes the bug laid horribly bare by https://github.com/spinframework/spin/pull/3605.

I couldn't test properly in my fork because of tokens and keys and whatnot but I made it cat the munged Cargo.toml and it looked uh not obviously wrong.  Well not obviously wrong _to me_ anyway.

https://github.com/itowlson/spin/actions/runs/28987200598/job/86018887021#step:4:849


fibonacci1729 (approved):
not a bitcoin miner!

itowlson:
> not a bitcoin miner!

I would use awk rather than sed for that.

PR #3608, opened 2026-07-09
Run the bump-wasmtime workflow everyday

fibonacci1729:
Runs everyday and PRs changes only when the release branch needs to bump.

### DIFF
--- a/.github/workflows/bump-wasmtime.yml
+++ b/.github/workflows/bump-wasmtime.yml
@@ -3,7 +3,7 @@
 name: Bump wasmtime (prerelease)
 on:
   schedule:
-    - cron: 0 0 9 * * # 9th of each month, at 00:00 UTC
+    - cron: 0 0 * * * # every day at 00:00 UTC
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


itowlson:
Could you say more about why you're proposing this?  Our actions dashboard is already quite noisy, and a daily action that prints "nothing to do" seems to have limited value.  Is it a hedge against Wasmtime changing its release schedule?

fibonacci1729:
That was my understanding after discussing with @tschneidereit last week. Could you weigh in or correct me if I misunderstood.

tschneidereit:
> Is it a hedge against Wasmtime changing its release schedule?

It is exactly that, yes: we want to maximize the amount of time we have between finding an issue that needs a Wasmtime fix and the Wasmtime release happening, so we want to check as close to the branch date as possible. OTOH, there have been instances of the branch being delayed, so we need to be robust in the face of that.

> Our actions dashboard is already quite noisy, and a daily action that prints "nothing to do" seems to have limited value.

I had honestly not considered that there's value in having a clean actions dashboard, because I didn't realize that anyone actively checks that dashboard. If you think that is an important enough thing to optimize for, I can think of some other ways to achieve the goals here, but they involve external mechanisms and a bunch more work, so I'm a bit wary of them.

itowlson:
Given that the Wasmtime 47 PR has been up for most of a week and nobody has looked at it, we have a ways to go as a project to benefit from "maximis[ing] the amount of time" from finding an issue.  _grin_

For what it's worth, we already delay by a couple of days to allow for small release delays and for "oops we meant to include X" churn in the immediate aftermath of starting a release train.  But yes, if the release were delayed longer than that, the cron would miss - I agree it would be nice to be more robust against that.

tschneidereit:
Discussing this with @fibonacci1729 just now I realized that my reasoning here was flawed to begin with: for this workflow, what we care about is the date the release branch is created, not the date the actual release itself happens. And luckily the branch creation [is fully automatic](https://github.com/bytecodealliance/wasmtime/blob/ff639ef68baed46aaeda8fefcf3f1bdba9e59bd2/.github/workflows/release-process.yml#L23) and (barring GitHub Actions infra failures, which of course never ever happen) doesn't get delayed.

So we could run this job with a cron schedule of `'0 12 5 * *'`, 12 hours after the Wasmtime branch schedule.

itowlson:
@fibonacci1729 Given Till's new information and proposal, what do you prefer: close, merge as is, rework, something else?

fibonacci1729:
Closing this. I'll PR the updated cron schedule.

Issue #3609, opened 2026-07-09
Expose a hard per-request deadline hook for spin-trigger-http embedders


carsonfarmer:
### What is the version of your Spin CLI?

Spin main / upcoming 4.1.0 pre-release.

This is an embedder-facing API suggestion for `spin-trigger-http`, not a CLI behavior bug.

### What is your suggestion?

Expose a small public API that lets embedders of `spin-trigger-http` apply a hard execution deadline to each HTTP request's underlying Spin `Store`.

One possible shape:

```rust
let reuse_config = InstanceReuseConfig::single_use_with_request_deadline(timeout);

let server = HttpServer::new(
    listen_addr,
    tls_config,
    find_free_port,
    trigger_app,
    http1_max_buf_size,
    reuse_config,
    output_format,
)?;
```

Internally, this mode would:

- set a hard request deadline on each request's Spin `Store`
- keep the existing response/request timeout behavior as the response-level guard
- disable instance reuse in this first API shape, so every request gets a fresh store with a request-specific deadline
- cover the legacy Spin HTTP executor, WASI HTTP executor, Wagi executor while supported, and the WASIp3 handler-state path

The default config and CLI-created config do not need to change. This can be opt-in for embedders that need hard request limits.

### Why would this improve Spin?

Embedders can currently wrap `HttpServer::handle(...)` in an async timeout, but that only limits how long the host waits for a response. It does not reliably interrupt CPU-bound guest Wasm. If a component enters a tight loop, the guest can continue executing unless the Wasmtime store itself is interrupted.

Wasmtime epoch deadlines are the right primitive for this, and Spin already exposes `Store::set_deadline`. The missing piece is a small hook in the HTTP trigger path, where the request store is actually created and guest HTTP code is invoked.

Without this hook, embedders that need hard request limits are pushed toward reimplementing or copying private Spin HTTP internals: routing, executor selection, Wagi/WASI/Spin ABI handling, outbound HTTP integration, request body plumbing, and WASIp3 worker behavior. A small public deadline config keeps that logic inside Spin and gives embedders a safer, lower-maintenance path.

### Are you willing to submit PRs to contribute to this feature or improvement?

- [x] Yes, I am willing to implement it.

itowlson:
@carsonfarmer Is this fixed by your #3610?  If so is it okay to close - I see you raised a new issue for remaining work right?  Thanks!

carsonfarmer:
Yes, apologies, I should have referenced this issue directly in that PR!

PR #3610, opened 2026-07-09
Add a hard request deadline config for embedded HTTP servers

carsonfarmer:
This adds an opt-in request deadline mode for embedders of `spin-trigger-http`.

The new API is:

```rust
let reuse_config =
    InstanceReuseConfig::single_use_with_request_deadline(timeout);
```

That config keeps the existing HTTP request timeout behavior, and also applies a hard deadline to the underlying Spin `Store` before guest HTTP code runs. This matters for CPU-bound guest code: an outer async timeout can stop waiting for a response, but it does not necessarily interrupt a tight Wasm loop. A store deadline does.

For this first version, the deadline config intentionally uses single-use instances. That keeps the API small and avoids subtle deadline-reset behavior for reused workers. Embedders that need hard per-request limits get correctness first, while the default config and CLI-created config remain unchanged.

The deadline is wired through the legacy Spin HTTP executor, WASI HTTP executor, Wagi executor, and the WASIp3 handler-state path.

Related issue: #3609

Validation:

```sh
cargo check -p spin-trigger-http
cargo test -p spin-trigger-http request_deadline_config_is_single_use
```

### DIFF
--- a/crates/trigger-http/src/lib.rs
+++ b/crates/trigger-http/src/lib.rs
@@ -239,6 +239,7 @@ pub struct InstanceReuseConfig {
     max_instance_reuse_count: Range<usize>,
     max_instance_concurrent_reuse_count: Range<usize>,
     request_timeout: Option<Range<Duration>>,
+    request_deadline: Option<Duration>,
     idle_instance_timeout: Range<Duration>,
 }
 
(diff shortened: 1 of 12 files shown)


crates/trigger-http/src/lib.rs
```diff
@@ -250,6 +251,24 @@ impl Default for InstanceReuseConfig {
...
             ),
             request_timeout: DEFAULT_REQUEST_TIMEOUT,
+            request_deadline: None,
+            idle_instance_timeout: DEFAULT_IDLE_INSTANCE_TIMEOUT,
+        }
+    }
+}
+
+impl InstanceReuseConfig {
+    /// Creates a single-use instance reuse configuration with a hard request deadline.
```
dicej: Calling this a "hard request deadline" is a bit misleading given that the precision depends on the value of `EngineBuilder::epoch_tick_interval` (default: 10ms), the fuzziness of `std::thread::sleep`, and how often the epoch comparison happens in the compiled guest code.  Perhaps we should note that the guest may run a bit longer than the specified deadline, but not arbitrarily longer?

dicej (commented):
Thanks, @carsonfarmer!

This looks reasonable to me, although I think we should eventually combine `request_timeout` and `request_deadline` into a single field, making use of [Store::epoch_deadline_async_yield_and_update](https://docs.rs/wasmtime/latest/wasmtime/struct.Store.html#method.epoch_deadline_async_yield_and_update) to allow it to work with or without instance reuse and across the various flavors of handler (Spin, WASIp2, WASIp3, and WAGI).  It's confusing to have two separate fields with overlapping responsibilities.

I'm fine with using the approach for now and opening an issue for future improvements, though.

dicej:
Also, would you be up for adding a simple integration test which verifies that the deadline is enforced for e.g. a WASIp2 guest that infinitely loops?

crates/trigger-http/src/lib.rs
carsonfarmer: Totally, fair point! I'll add that as part of the comment here.

carsonfarmer:
> Also, would you be up for adding a simple integration test which verifies that the deadline is enforced for e.g. a WASIp2 guest that infinitely loops?

Good call, yes, absolutely. Integration test added.

crates/trigger-http/src/lib.rs
carsonfarmer: I updated the language to be more tempered.

carsonfarmer:
> I'm fine with using the approach for now and opening an issue for future improvements, though.

I created a follow up issue for future improvements. My confidence in implementing that is low, but I am willing to work on it.

dicej (approved):
Thanks for doing this and adding the integration test!

PR #3612, opened 2026-07-10
Miscellaneous improvements and cleanup (batch 2)

ChihweiLHBird:
- Use `split_once` for first-delimiter parsing in environment variables and OCI basic authentication
- Use `Itertools::join` in the dotted PascalCase template filter. These changes remove temporary collections without changing behavior.
- ~~Search for `spin.toml` with `Path::ancestors` to remove the depth limit~~
- Use `exactly_one` to detect a single GitHub tarball root directory
- Move owned WASI OpenTelemetry data into OTel types instead of cloning it.

### DIFF
--- a/crates/common/src/paths.rs
+++ b/crates/common/src/paths.rs
@@ -1,6 +1,7 @@
 //! Resolves a file path to a manifest file
 
 use anyhow::{Context, Result, anyhow};
+use std::env::current_dir;
 use std::path::{Path, PathBuf};
 
 use crate::ui::quoted_path;
@@ -71,8 +72,8 @@ pub fn search_upwards_for_manifest() -> Option<(PathBuf, usize)> {
         return Some((candidate, 0));
     }
 
-    for distance in 1..20 {
-        let inferred_dir = PathBuf::from("../".repeat(distance));
+    let current_dir = current_dir().ok()?;
+    for (distance, inferred_dir) in current_dir.ancestors().enumerate() {
         if !inferred_dir.is_dir() {
             return None;
         }
@@ -82,7 +83,7 @@ pub fn search_upwards_for_manifest() -> Option<(PathBuf, usize)> {
             return Some((candidate, distance));
         }
 
-        if is_git_root(&inferred_dir) {
+        if is_git_root(inferred_dir) {
             return None;
         }
     }
--- a/crates/common/src/paths.rs
+++ b/crates/common/src/paths.rs
@@ -1,7 +1,6 @@
 //! Resolves a file path to a manifest file
 
 use anyhow::{Context, Result, anyhow};
-use std::env::current_dir;
 use std::path::{Path, PathBuf};
 
 use crate::ui::quoted_path;
@@ -72,8 +71,8 @@ pub fn search_upwards_for_manifest() -> Option<(PathBuf, usize)> {
         return Some((candidate, 0));
     }
 
-    let current_dir = current_dir().ok()?;
-    for (distance, inferred_dir) in current_dir.ancestors().enumerate() {
+    for distance in 1..20 {
+        let inferred_dir = PathBuf::from("../".repeat(distance));
         if !inferred_dir.is_dir() {
             return None;
         }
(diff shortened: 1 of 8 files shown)


crates/common/src/paths.rs
```diff
@@ -71,8 +72,8 @@ pub fn search_upwards_for_manifest() -> Option<(PathBuf, usize)> {
         return Some((candidate, 0));
     }
 
-    for distance in 1..20 {
-        let inferred_dir = PathBuf::from("../".repeat(distance));
+    let current_dir = current_dir().ok()?;
+    for (distance, inferred_dir) in current_dir.ancestors().enumerate() {
```
itowlson: The depth limit is intentional (if anything I would reduce it).

Also, I have a feeling we intentionally went with the `..` prefix so that the printout would show something like `../spin.toml` rather than switching to an absolute path.

crates/oci/src/auth.rs
```diff
@@ -67,18 +67,14 @@ impl AuthConfig {
 
         let bytes = base64::Engine::decode(&base64::engine::general_purpose::STANDARD, encoded)?;
         let decoded = std::str::from_utf8(&bytes)?;
-        let parts: Vec<&str> = decoded.splitn(2, ':').collect();
+        let (username, secret) = decoded
+            .split_once(':')
+            .context("expected secret as second element of the decoded auth")?;
```
itowlson, line 72: I see we are no longer checking for the decoded auth having a first element, but I _think_ the existing check would never fire, so I don't think there's any loss here... right?

crates/templates/src/source.rs
```diff
@@ -253,13 +254,12 @@ fn bypass_gh_added_root(unpack_dir: PathBuf) -> PathBuf {
...
     // return the root of the unpacking. (The take(2) here is because we don't need to traverse
     // the full list - we only care whether there is more than one.)
-    let dirs = dirs.filter_map(|de| de.ok()).take(2).collect::<Vec<_>>();
-    if dirs.len() != 1 {
+    let Ok(dir) = dirs.filter_map(Result::ok).exactly_one() else {
         return unpack_dir;
-    }
+    };
 
     // If we get here, there is a single directory (dirs has a single element). Look in it to see if it's a plausible repo root.
```
itowlson: You can probably remove this comment, as the array indexing has nicely gone away and we know that `dir` is a scalar (yay type systems).

itowlson (changes requested):
Nice clean ups, but I'd like to avoid the change to the path search behaviour - everything else looks fine though!

itowlson:
cc @calebschoepp for the OTel changes which I don't feel qualified to review

crates/templates/src/source.rs
ChihweiLHBird: Not sure what it does... but yes, I can [remove it](https://github.com/ChihweiLHBird/spin/commit/59cafdf91c7eac0e9e67b3fc28190b356adc8e1f)

crates/oci/src/auth.rs
ChihweiLHBird: Do you mean `context("expected secret as second element of the decoded auth")` would be unnecessary or?

crates/common/src/paths.rs
ChihweiLHBird: Got it, thanks! Will undo it.

crates/oci/src/auth.rs
itowlson: No, I mean there used to be a context about "expected username as first element of the decoded auth," and now there isn't.  And I'm thinking this is okay because I reckon "expected username" could never have been raised.  But I want someone to check my reasoning!

crates/templates/src/source.rs
itowlson: It was previously useful because it explained why `dirs[0]` is safe (and why we don't need to check `dirs[1]` etc.).  But it is now redundant (and indeed incorrect) because we now have scalar `dir` instead of vector `dirs`.

itowlson (approved):

crates/templates/src/source.rs
ChihweiLHBird: Ah got it. Sorry, I misunderstood it previously

calebschoepp (approved):
OTel parts LGTM

PR #3617, opened 2026-07-14
add http middleware template

michelleN:
Opened up this draft PR as an additional resource to visualize https://github.com/spinframework/spin/issues/3618. 
This is an attempt to add an http middleware template for both the `spin new` and `spin add` experience. I ran into a few issues.

### DIFF
--- a/templates/http-middleware/content/spin.toml
+++ b/templates/http-middleware/content/spin.toml
@@ -1,0 +1,14 @@
+spin_manifest_version = 2
+
+[application]
+name = "{{project-name | kebab_case}}"
+version = "0.1.0"
+authors = ["{{authors}}"]
+description = "{{project-description}}"
+
+[component.{{project-name | kebab_case}}]
+source = "target/wasm32-wasip2/release/{{project-name | snake_case}}.wasm"
+allowed_outbound_hosts = []
+[component.{{project-name | kebab_case}}.build]
+command = "cargo build --target wasm32-wasip2 --release"
+watch = ["src/**/*.rs", "Cargo.toml"]
--- a/templates/http-middleware/metadata/snippets/component.txt
+++ b/templates/http-middleware/metadata/snippets/component.txt
@@ -1,0 +1,7 @@
+[component.{{project-name | kebab_case}}]
+source = "{{ output-path }}/target/wasm32-wasip2/release/{{project-name | snake_case}}.wasm"
+allowed_outbound_hosts = []
+[component.{{project-name | kebab_case}}.build]
+command = "cargo build --target wasm32-wasip2 --release"
+workdir = "{{ output-path }}"
+watch = ["src/**/*.rs", "Cargo.toml"]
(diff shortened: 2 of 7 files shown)


templates/http-middleware/metadata/snippets/component.txt
```diff
@@ -0,0 +1,10 @@
+[[trigger.http]]
```
michelleN: The current templating system requires the add snippet to contain a trigger section which doesn't make sense for a middleware component (or any library component).

templates/http-middleware/metadata/spin-template.toml
```diff
@@ -0,0 +1,14 @@
+manifest_version = "1"
+id = "http-middleware"
+description = "http middleware rust template"
+tags = []
+
+[add_component]
+skip_files = ["spin.toml"]
+[add_component.snippets]
+component = "component.txt"
```
michelleN, line 12: Adding middleware to an existing app would require post add instructions including adding the middleware to dependencies under the correct trigger section.

templates/http-middleware/content/spin.toml
```diff
@@ -0,0 +1,14 @@
+spin_manifest_version = 2
```
michelleN, line 1: There's not really a clear way to test middleware as a standalone spin app unless we revive (_I think_)`spin test`

templates/http-middleware/metadata/snippets/component.txt
itowlson: Huh, weird.  I think we used to check the trigger to make sure it was the same as existing triggers.  I don't think that would be needed any more - let's see how the trigger info is being depended on and see if we can bin it off.

templates/http-middleware/content/spin.toml
itowlson: When I briefly looked at a middleware template, I assumed it would have to be add-only (not available in `spin new`).  Even with `spin test` it's not clear how a standalone piece of middleware could work.

templates/http-middleware/metadata/spin-template.toml
itowlson: This may be unavoidable - presumably the reason you're building your thing as middleware rather than including it in component source code is that you want to apply it to multiple triggers.  (Although I guess another case is "my app is an off-the-shelf component like the fileserver but I want to wrap it."  So maybe sometimes we do want to be able to auto-add it to triggers.  But yeah, we don't have syntax for that yet.)

templates/http-middleware/metadata/snippets/component.txt
michelleN: PR for removing trigger validation is merged so going to mark this as resolved

templates/http-middleware/content/spin.toml
michelleN: +1 updated template to make it add only

PR #3619, opened 2026-07-16
Add ability to control metric output level through spin_telemetry cargo feature

rylev:
This gives embedders a way to control at which level telemetry should be emitted at. I think this is the best way to allow metrics which are emitted deep from within Spin's codebase to still be controlled by embedders which may choose to statically strip out traces and would thus never see these metrics.

### DIFF
--- a/crates/telemetry/src/metrics.rs
+++ b/crates/telemetry/src/metrics.rs
@@ -88,6 +88,50 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
     Ok(MetricsLayer::new(meter_provider))
 }
 
+// The two mutually exclusive features cannot both select a default level.
+#[cfg(all(
+    feature = "metrics-default-level-debug",
+    feature = "metrics-default-level-info"
+))]
+compile_error!(
+    "features `metrics-default-level-debug` and `metrics-default-level-info` are mutually exclusive"
+);
+
+/// The [`tracing::Level`] at which the metrics macros (`counter!`, `histogram!`,
+/// `monotonic_counter!`, `gauge!`) emit their events when no explicit `level:` is given.
+///
+/// Defaults to [`tracing::Level::TRACE`]. Enable the `metrics-default-level-debug` or
+/// `metrics-default-level-info` feature on `spin-telemetry` to raise it to `DEBUG` or `INFO`.
+#[cfg(all(
+    not(feature = "metrics-default-level-debug"),
+    not(feature = "metrics-default-level-info")
+))]
+pub const DEFAULT_METRICS_LEVEL: tracing::Level = tracing::Level::TRACE;
+
+/// The [`tracing::Level`] at which the metrics macros (`counter!`, `histogram!`,
+/// `monotonic_counter!`, `gauge!`) emit their events when no explicit `level:` is given.
+///
+/// Defaults to [`tracing::Level::DEBUG`]. Enable the `metrics-default-level-info` or
+/// remove the `metrics-default-level-debug` feature on `spin-telemetry` to raise it to `INFO`
+/// or lower it to `TRACE` respectively.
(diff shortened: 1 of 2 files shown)


lann:
It would be a larger change but I'd be inclined to switch metrics to use opentelemetry directly rather than through the wonky `tracing` interface.

crates/telemetry/src/metrics.rs
```diff
@@ -88,20 +88,66 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
     Ok(MetricsLayer::new(meter_provider))
 }
 
+// The two mutually exclusive features cannot both select a default level.
```
lann, line 91: FWIW we do already have `cfg-if` in the dep tree...

Looking forward to [bumping the MSRV](https://blog.rust-lang.org/2026/04/16/Rust-1.95.0/#cfg-select) too.

crates/telemetry/src/metrics.rs
```diff
@@ -88,20 +88,66 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
...
+#[cfg(all(
+    feature = "metrics-default-level-debug",
+    feature = "metrics-default-level-info"
+))]
+compile_error!(
+    "features `metrics-default-level-debug` and `metrics-default-level-info` are mutually exclusive"
+);
```
lann, line 98: Given that these are sort of "logically additive" it would be nice to treat this the same as just enabling `metrics-default-level-info` so that `--all-features` doesn't break.

lann: ...though annoyingly that already appears to be broken by `llm-metal` / `llm-cublas`.

lann (approved):

calebschoepp (approved):
> It would be a larger change but I'd be inclined to switch metrics to use opentelemetry directly rather than through the wonky tracing interface.

I share the same desire, but not blocking of course.

rylev:
> It would be a larger change but I'd be inclined to switch metrics to use opentelemetry directly rather than through the wonky tracing interface.

@lann @calebschoepp I'd be fine with this, but that would be a breaking change. Not sure how we want to handle that.

calebschoepp:
> @lann @calebschoepp I'd be fine with this, but that would be a breaking change. Not sure how we want to handle that.

Breaking change for embedders, or for something else I'm misunderstanding?

lann:
Not if they're using the macros! 😅

rylev:
For all users of spin. It's not a breaking _code_ change, but if we use opentelemetry directly, users of spin will suddenly start seeing metrics even if they had previously filtered them out with RUST_LOG=info before.

lann:
I think you already have to opt-in with env vars don't you? https://github.com/spinframework/spin/blob/ae349c2b9cf908be831b1e2831ddbfb4a8d75c85/crates/telemetry/src/env.rs#L29-L41

calebschoepp:
> For all users of spin. It's not a breaking code change, but if we use opentelemetry directly, users of spin will suddenly start seeing metrics even if they had previously filtered them out with RUST_LOG=info before.

I personally don't see this as a problem. I'd be really surprised if people were explicitly using `RUST_LOG` to filter out metrics while still opting into forwarding stuff to the OTel collector. The fact that metrics can be filtered out by `RUST_LOG` is just a weird side effect of our implementation that I think we should move away from.

itowlson:
I see two approvals but a lot of ongoing discussion.  Should I merge this or are folks thinking there is more work/thought needed?

rylev:
@itowlson I've provided an alternative implementation in https://github.com/spinframework/spin/pull/3623 - let's see what @lann and @calebschoepp think

rylev:
Closing in favor of https://github.com/spinframework/spin/pull/3623

PR #3620, opened 2026-07-16
OCD tidying

fibonacci1729:


itowlson (approved):
I asked Joel about this.  Apparently we now need folders only for multi-file packages.  So we should make a call on what we want to do going forward.

But if we prefer to keep using folders structure despite them being no longer needed, I'm fine with that too!

fibonacci1729:
Yeah I understand that but maintaining a mix is harder and more annoying to automate against. Case in point, I forgot to port this over when updating a downstreams WITs because the mysql file was down below all the other directories and i missed it.

fibonacci1729:
So my vote would be to conventionalize subdirectories even if a single wit file.

itowlson:
@fibonacci1729 ~So, once again my masterful strategy of blindly copying the entire deps directory without thinking or nuance pays off~ Yeah, good point - there's muscle memory that the only bare file is the world, and if we're going to change that, we need to change it visibly enough that it's _highly_ visible to readers.

PR #3623, opened 2026-07-17
Emit otel metrics directly instead of through tracing

rylev:
Alternative to https://github.com/spinframework/spin/pull/3619

Instead of doing some compile time magic, we instead get rid of our dependency on `tracing` for emitting metrics and instead emit the otel metrics directly.

This does require duplicating macros to specify types as as far as I can tell we can't infer the type param we need to pass to the `opentelemetry::metrics` types.

I also deleted some tests in crates/connection-semaphore/src/lib.rs where were testing that we were emitting specific metrics. This is harder to test with the new implementation, and I'm not convinced we actually got that much out of those tests.

### DIFF
--- a/crates/telemetry/src/lib.rs
+++ b/crates/telemetry/src/lib.rs
@@ -51,7 +51,7 @@ pub use propagation::inject_trace_context;
 /// Examples of emitting metrics from Spin:
 ///
 /// ```no_run
-/// spin_telemetry::metrics::monotonic_counter!(spin.metric_name = 1, metric_attribute = "value");
+/// spin_telemetry::metrics::monotonic_counter_u64!(spin.metric_name = 1, metric_attribute = "value");
 /// ```
 ///
 /// `histogram_buckets` lets callers override the OTel default histogram boundaries for specific
@@ -83,21 +83,11 @@ pub fn init(spin_version: String, histogram_buckets: Vec<HistogramBuckets>) -> a
         None
     };
 
-    let otel_metrics_layer = if otel_metrics_enabled() {
-        Some(
-            metrics::otel_metrics_layer(spin_version.clone(), histogram_buckets)
-                .context("failed to initialize otel metrics")?,
-        )
-    } else {
-        None
-    };
-
     let alert_in_dev_layer = alert_in_dev::alert_in_dev_layer();
 
     // Build a registry subscriber with the layers we want to use.
     registry()
         .with(otel_tracing_layer)
-        .with(otel_metrics_layer)
         .with(fmt_layer)
         .with(alert_in_dev_layer)
         .init();
@@ -105,6 +95,12 @@ pub fn init(spin_version: String, histogram_buckets: Vec<HistogramBuckets>) -> a
     // Used to propagate trace information in the standard W3C TraceContext format. Even if the otel
     // layer is disabled we still want to propagate trace context.
     opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
+
+    if otel_metrics_enabled() {
+        let meter_provider = metrics::metrics_provider(spin_version.clone(), histogram_buckets)
+            .context("failed to initialize otel metrics")?;
+        opentelemetry::global::set_meter_provider(meter_provider);
+    }
 
     if otel_logs_enabled() {
         logs::init_otel_logging_backend(spin_version)
--- a/crates/telemetry/src/lib.rs
+++ b/crates/telemetry/src/lib.rs
@@ -25,10 +25,13 @@ pub use propagation::inject_trace_context;
 /// Initializes telemetry for Spin using the [tracing] library.
 ///
 /// Under the hood this involves initializing a [tracing::Subscriber] with multiple [Layer]s. One
-/// [Layer] emits [tracing] events to stderr, another sends spans to an OTel collector, and another
-/// sends metrics to an OTel collector.
+/// [Layer] emits [tracing] events to stderr, and another sends spans to an OTel collector. Metrics
+/// are handled separately from the tracing [Layer]s: a global OTel meter provider is registered
+/// directly, and the metric macros in [`metrics`] record to it without going through `tracing`.
 ///
-/// Configuration for the OTel layers is pulled from the environment.
+/// Configuration for the OTel layers and the meter provider is pulled from the environment. This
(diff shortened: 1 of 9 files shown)


crates/telemetry/src/metrics.rs
```diff
@@ -81,82 +86,176 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
...
-/// counter!(spin.metric_name = 1, metric_attribute = "value");
 /// ```
-macro_rules! counter {
-    ($metric:ident $(. $suffixes:ident)*  = $metric_value:expr $(, $attrs:ident=$values:expr)*) => {
-        tracing::trace!(counter.$metric $(. $suffixes)* = $metric_value $(, $attrs=$values)*);
-    }
+/// spin_telemetry::metrics::monotonic_counter_f64!(spin.metric_name = 1.5, metric_attribute = "value");
+/// ```
+#[macro_export]
+macro_rules! monotonic_counter_f64 {
```
lann: I think we should omit this (and strip _u64 from the other one). I can't think of any use for a float counter in this kind of project. _What does counting even mean?!_

lann:
> This does require duplicating macros to specify types as as far as I can tell we can't infer the type param we need to pass to the `opentelemetry::metrics` types.

This seems more robust than what tracing-opentelemetry does anyway: https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/struct.MetricsLayer.html#floating-point-numbers

crates/telemetry/src/lib.rs
```diff
@@ -106,6 +96,12 @@ pub fn init(spin_version: String, histogram_buckets: Vec<HistogramBuckets>) -> a
     // layer is disabled we still want to propagate trace context.
     opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
 
+    if otel_metrics_enabled() {
+        let meter_provider = metrics::metrics_provider(spin_version.clone(), histogram_buckets)
+            .context("failed to initialize otel metrics")?;
+        opentelemetry::global::set_meter_provider(meter_provider);
+    }
```
lann, line 106: I'm not 100% confident but I _think_ we could replace this (and most of `metrics_provider(...)` with the default behavior that [pulls config from env vars](https://docs.rs/opentelemetry-otlp/latest/opentelemetry_otlp/#environment-variables). I think we had this explicit check previously because the extra tracing layer had a significant effect on tracing performance under high load.

lann: @calebschoepp do you remember what the setup in `metrics_provider` is all about?

crates/telemetry/src/metrics.rs
```diff
@@ -81,82 +86,176 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
...
 ///
-/// The increment may only be a positive i64 or f64. You must not mix types for the same metric.
+/// Unlike `monotonic_counter_*`, the delta may be negative. This maps to OTel's `UpDownCounter`.
 ///
-/// Takes advantage of monotonic counter support in [tracing-opentelemetry](https://docs.rs/tracing-opentelemetry/0.32.0/tracing_opentelemetry/struct.MetricsLayer.html).
+/// ```
+/// spin_telemetry::metrics::counter_f64!(spin.metric_name = -1.5, metric_attribute = "value");
+/// ```
+#[macro_export]
+macro_rules! counter_f64 {
```
lann: Ditto.

lann (commented):
Minor quibbles. I think this is better than tracing.

lann (approved):

crates/telemetry/src/lib.rs
lann, line 25: I suppose this should be updated to mention that it sets the global metrics provider. An LLM might read it...

crates/telemetry/src/metrics.rs
```diff
@@ -81,82 +86,176 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
...
+            $crate::metrics::opentelemetry::metrics::Gauge<f64>, f64_gauge, record, $($tt)*
+        )
+    };
 }
 
-pub use counter;
-pub use gauge;
-pub use histogram;
-pub use monotonic_counter;
+pub use counter_f64;
```
calebschoepp: Nit: I think it's more clear to expose these with names matching the underlying OTel types i.e. `Counter` and `UpDownCounter` vs. `MonotonicCounter` and `Counter`. Not blocking and feel free to stick with your opinion - just thought I'd voice that it's a bit confusing to me.

crates/telemetry/src/lib.rs
calebschoepp: This metrics layer is gone now

crates/telemetry/src/metrics.rs
```diff
@@ -81,82 +86,176 @@ pub(crate) fn otel_metrics_layer<S: Subscriber + for<'span> LookupSpan<'span>>(
...
+}
+
+/// Records an increment to the named monotonic counter (as a `u64`) with the given attributes.
 ///
-/// The increment may only be an i64 or f64. You must not mix types for the same metric.
+/// ```
+/// spin_telemetry::metrics::monotonic_counter_u64!(spin.metric_name = 1, metric_attribute = "value");
+/// ```
+#[macro_export]
+macro_rules! monotonic_counter_u64 {
```
calebschoepp: I think it's important to document clearly somewhere where the consumers of these macros will see that to emit a metric from multiple locations they need to wrap the macro in a function and call that function instead (because of the static nature). You kind of allude to this but I think it should be more clear.

crates/telemetry/src/lib.rs
calebschoepp: I'd say lets just not mess with this. You might be right Lann, but this would be a bigger refactor across tracing and logs too and also risk more potentially breaking DX changes.

calebschoepp (approved):
I like this approach a lot more

crates/telemetry/src/metrics.rs
rylev: I'm not sure I understand what you mean. The `INSTRUMENT` is static but each macro invocation leads to a different `$record_method` call...

PR #3624, opened 2026-07-18
Miscellaneous performance optimizations

ChihweiLHBird:
### Summary

Five small, low-risk hot-path cleanups (one commit each), no behavior change for the CLI:

- **telemetry:** cache the OTel logs enablement check (`OnceLock`) instead of re-reading env vars on every guest log write
- **trigger-http:** resolve the app name once at server construction instead of deserializing metadata on every request
- **key-value-spin:** `exists` uses `SELECT 1 … LIMIT 1` instead of fetching the whole value BLOB via `get`
- ~~**llm-remote-http:** move embedding vectors out of the response instead of cloning them~~
- ~~**llm-local:** clone only `eos_token_id` instead of the whole llama `Config`~~

### Notes

- Malformed `name` metadata now fails at server construction instead of per request (already caught earlier by the default factor stack, so unobservable in practice).
- SQLite `exists` no longer decodes the value, matching the other backends' native existence probes.

### DIFF
--- a/crates/llm-local/src/llama.rs
+++ b/crates/llm-local/src/llama.rs
@@ -89,7 +89,7 @@ impl InferencingModel for LlamaModels {
         let mut cache = self.cache.clone();
         // Try to retrieve the End of Sentence (EOS) token ID from config or
         // default to a single EOS token. EOS token is used to determine when to stop.
-        let eos_token_id = config.clone().eos_token_id.or_else(|| {
+        let eos_token_id = config.eos_token_id.clone().or_else(|| {
             tokenizer
                 .token_to_id(EOS_TOKEN)
                 .map(llama::LlamaEosToks::Single)
--- a/crates/llm-remote-http/src/open_ai/schemas.rs
+++ b/crates/llm-remote-http/src/open_ai/schemas.rs
@@ -63,15 +63,6 @@ pub struct CreateEmbeddingResponse {
     usage: EmbeddingUsage,
 }
 
-impl CreateEmbeddingResponse {
-    fn embeddings(&self) -> Vec<Vec<f32>> {
-        self.data
-            .iter()
-            .map(|embedding| embedding.embedding.clone())
-            .collect()
-    }
-}
-
 #[derive(Deserialize)]
 struct EmbeddingUsage {
     prompt_tokens: u32,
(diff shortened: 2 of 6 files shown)


crates/telemetry/src/logs.rs
```diff
@@ -26,7 +26,8 @@ pub fn handle_app_log(buf: &[u8], component_id: &str) {
 
 /// Forward the app log to OTel.
 fn app_log_to_otel(buf: &[u8], component_id: &str) {
-    if !otel_logs_enabled() {
+    static CELL: OnceLock<bool> = OnceLock::new();
+    if !*CELL.get_or_init(otel_logs_enabled) {
```
itowlson: Can we move the CELL into `otel_logs_enabled`?  It feels like the caching shouldn't be a concern of the consumer.

crates/llm-local/src/llama.rs
```diff
@@ -89,7 +89,7 @@ impl InferencingModel for LlamaModels {
         let mut cache = self.cache.clone();
         // Try to retrieve the End of Sentence (EOS) token ID from config or
         // default to a single EOS token. EOS token is used to determine when to stop.
-        let eos_token_id = config.clone().eos_token_id.or_else(|| {
+        let eos_token_id = config.eos_token_id.clone().or_else(|| {
```
itowlson: I wouldn't bother with LLM changes.

crates/llm-remote-http/src/open_ai/schemas.rs
```diff
@@ -95,7 +86,7 @@ impl From<CreateChatCompletionResponse> for wasi_llm::InferencingResult {
 impl From<CreateEmbeddingResponse> for wasi_llm::EmbeddingsResult {
     fn from(value: CreateEmbeddingResponse) -> Self {
         Self {
-            embeddings: value.embeddings(),
+            embeddings: value.data.into_iter().map(|e| e.embedding).collect(),
```
itowlson: again would skip LLM changes

itowlson (approved):
LGTM - thanks!

@calebschoepp any concerns on the OTel tweak?

calebschoepp (approved):
Nice catch!

PR #3627, opened 2026-07-21
Workflow to finalize Wasmtime upgrade

fibonacci1729:
First attempt at adding a workflow to complement the `bump-wasmtime` automation.

I broke this up into a manual workflow_dispatch workflow. There are two phases `update` and `backport`. The `update` phase updates the PR created from the bump-wasmtime workflow to point at the released crates.io version. The `backport` job can then be run to backport the merged PR from the `update` phase to the latest release branch. This backport PR will contain the patch version bump as well. All that is left is to push the tag to cut the release.


itowlson:
> Is this a crypto miner? No.

That's assurance enough for me!

.github/workflows/finalize-wasmtime.yml
```diff
@@ -0,0 +1,269 @@
...
+# release *branch* (a git dependency) so integration can begin before wasmtime
+# ships. This workflow watches crates.io and, once the release is published:
+#   1. Rewrites the bump PR's git/branch dependency back to the crates.io
+#      version and pushes it to the PR branch.
+#   2. Backports the entire bump PR onto the last Spin release branch and opens
+#      a PR that also bumps the patch version, ready to be tagged as a release.
+name: Finalize wasmtime release
+on:
+  schedule:
+    - cron: 0 13 * * * # Daily; polls crates.io for the published wasmtime release
```
itowlson: Suggestion: for the first run or two, leave this as manual dispatch only.  That might reduce the fear factor because we can run it at a time when we can examine and sanity-check the output.

.github/workflows/finalize-wasmtime.yml
```diff
@@ -0,0 +1,269 @@
...
+
+      - name: Discover bump PR, published version, and release branch
+        id: discover
+        shell: bash
+        env:
+          INPUT_PR_NUMBER: ${{ github.event.inputs.pr_number }}
+          INPUT_CRATES_VERSION: ${{ github.event.inputs.crates_version }}
+          INPUT_RELEASE_BRANCH: ${{ github.event.inputs.release_branch }}
+        run: |
+          set -eu
```
itowlson: Is it viable to put this in a separate `.sh` file?  That would likely make it easier to maintain, because if there's one thing I bar more than editing bash, it's editing bash _embedded in YAML._

.github/workflows/finalize-wasmtime.yml
```diff
@@ -0,0 +1,269 @@
...
+
+          if grep -q 'bytecodealliance/wasmtime' Cargo.toml examples/spin-timer/Cargo.toml; then
+            echo "::error::Some wasmtime git references remain after the rewrite; the Cargo.toml format may have changed. Update the dependency manually."
+            exit 1
+          fi
+
+          cargo update
+          cargo update --manifest-path examples/spin-timer/Cargo.toml
+
+          git commit -a -m "Point wasmtime at crates.io v${CRATES_VERSION}"
```
itowlson:
```suggestion
          git commit -a -m "Update wasmtime to v${CRATES_VERSION}"
```

.github/workflows/finalize-wasmtime.yml
```diff
@@ -0,0 +1,269 @@
...
+          git push origin "HEAD:${PR_BRANCH}"
+
+      - name: Backport and prepare the patch release
```
itowlson: Should we wait until the PR has been checked and possibly amended before doing the backport?

.github/workflows/finalize-wasmtime.yml
```diff
@@ -0,0 +1,269 @@
...
+        if: ${{ steps.discover.outputs.proceed == 'true' && steps.discover.outputs.backport_exists == 'false' }}
+        shell: bash
+        env:
+          PR_NUMBER: ${{ steps.discover.outputs.pr_number }}
+          CRATES_VERSION: ${{ steps.discover.outputs.crates_version }}
+          RELEASE_BRANCH: ${{ steps.discover.outputs.release_branch }}
+          NEXT_VERSION: ${{ steps.discover.outputs.next_version }}
+          BACKPORT_BRANCH: ${{ steps.discover.outputs.backport_branch }}
+        run: |
+          set -eu
```
itowlson: Consider a `.sh` file

itowlson (commented):
Per your comment about improving confidence, my suggestions would be:

* Split the bump and backport-patch.  That gives us an opportunity to review the bump PR, and address any concerns, before kicking off the backport.  After several months of everything going perfectly smoothly, as it surely will, we can review whether to re-combine them.
* Begin with the manual (`workflow_dispatch`) trigger only.  That means we can run it at a time of our choosing when we are ready to keep an eye out for any shenanigans.  Move to automatic later, when our confidence is high.

.github/workflows/finalize-wasmtime.yml
fibonacci1729: Great idea.

.github/workflows/finalize-wasmtime.yml
fibonacci1729: YES. I was just being lazy.

.github/workflows/finalize-wasmtime.yml
fibonacci1729: Everything is manual dispatch for now. PTAL.

.github/workflows/finalize-wasmtime.yml
fibonacci1729: The backport job now looks for the updated and merged PR so marking this as resolved.

.github/workflows/finalize-wasmtime.yml
```diff
@@ -0,0 +1,101 @@
...
+name: Finalize wasmtime release
+on:
+  workflow_dispatch:
+    inputs:
+      action:
+        description: Which phase to run
+        type: choice
+        required: true
+        default: update
+        options:
```
itowlson: To me it feels conceptually simpler to have two workflows here.  For example `release_branch` only makes sense in the backport flow, and `crates_version` presumably makes sense only in the update flow (because backport is just blindly backporting an existing PR, like the existing backport script, it doesn't care about Wasmtime versions).  And there don't seem to be any common steps across the two flows...?

scripts/wasmtime-release.sh
```diff
@@ -0,0 +1,234 @@
...
+
+  cargo update
+  cargo update --manifest-path examples/spin-timer/Cargo.toml
+
+  git commit -a -m "Update wasmtime to v${crates_version}"
+  git push origin "HEAD:${pr_branch}"
+  log "Pushed crates.io update to ${pr_branch}"
+}
+
+cmd_backport() {
```
itowlson, line 142: Can we leverage the existing backport script for this?

itowlson (approved):
You might want to get someone else to look at the bash, and I have a couple of suggestions, but generally LGTM - thanks!

scripts/wasmtime-release.sh
fibonacci1729: Not very cleanly. The backport job here also bumps the version for the patch release.

.github/workflows/finalize-wasmtime.yml
fibonacci1729: Updated to 2 workflows (update & backport).

PR #3629, opened 2026-07-21
remove single trigger compatibility validation

### LINKED ISSUE #3626: `spin add`: remove trigger compatibility check
https://github.com/spinframework/spin/pull/2027 reinstated a check in `spin add` that the component being added had the same trigger type as the current ones.  This was correct at the time because we then supported only a single trigger type in an app.  However, time moves on (spoilers), and an app can now cheerfully (well, moderately cheerfully) have more than one trigger type.  So we should remove the check.

Thanks @michelleN for spotting that the check was still there!  (I thought I had removed it when we did multi-trigger-type, but I guess that was just a beautiful dream.)

michelleN:
+ resolves #3626

### DIFF
--- a/crates/templates/src/app_info.rs
+++ b/crates/templates/src/app_info.rs
@@ -8,7 +8,6 @@ use std::{
     sync::LazyLock,
 };
 
-use anyhow::ensure;
 use serde::Deserialize;
 use spin_manifest::schema::v1;
 
@@ -16,7 +15,11 @@ use crate::store::TemplateLayout;
 
 pub(crate) struct AppInfo {
     manifest_format: u32,
-    trigger_type: Option<String>, // None = v2 template does not contain any triggers yet
+    // Used for v1 manifests, which have exactly one trigger type and are the
+    // only manifests subject to a compatibility check. v2 manifests may host
+    // multiple trigger types. Triggerless templates also allowed for all
+    // manifest versions.
+    trigger_types: Option<Vec<String>>,
 }
 
 impl AppInfo {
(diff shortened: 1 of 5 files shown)


crates/templates/src/app_info.rs
```diff
@@ -8,15 +8,18 @@ use std::{
...
 use crate::store::TemplateLayout;
 
 pub(crate) struct AppInfo {
     manifest_format: u32,
-    trigger_type: Option<String>, // None = v2 template does not contain any triggers yet
+    // Used for v1 manifests, which have exactly one trigger type and are the
+    // only manifests subject to a compatibility check. v2 manifests may host
+    // multiple trigger types. Triggerless templates also allowed for all
+    // manifest versions.
+    trigger_types: Option<Vec<String>>,
```
itowlson: I _think_ we can leave this as `Option<String>`, because if it's v1, it can only be a single type, and if it's v2, we don't care about the type anyway.

But however we do it we need some way to distinguish between "this is a v1 with this trigger type" and "this is a v2 which currently has only one trigger type."

Although... really, at this point, our templates can't add to a v1 manifest anyway, because the trigger and component syntax is all wrong.  So perhaps what we should do is:

* detect if we are adding to a v1 manifest
  * if so, fail gracefully
* remove everything to do with trigger types and trigger type checking

It wouldn't be a breaking change because it's _ahem_ already broken...

crates/templates/src/template.rs
```diff
@@ -345,7 +345,7 @@ impl Template {
 
     fn infer_trigger_type(layout: &TemplateLayout) -> TemplateTriggerCompatibility {
         match crate::app_info::AppInfo::from_layout(layout) {
-            Some(Ok(app_info)) => match app_info.trigger_type() {
+            Some(Ok(app_info)) => match app_info.trigger_types().and_then(|types| types.first()) {
```
itowlson: With the current implementation this feels like it would constrain a v2 application to the first trigger type in its manifest, but maybe I'm misreading it?

(But per comment above I think we should ditch all checking anyway, and simply fail on v1 manifests.)

crates/templates/src/app_info.rs
```diff
@@ -219,7 +225,33 @@ mod test {
 
         let info = AppInfo::from_template_text(tpl).unwrap();
         assert_eq!(2, info.manifest_format);
-        assert_eq!("triggy", info.trigger_type.unwrap());
+        assert_eq!(vec!["triggy".to_owned()], info.trigger_types.unwrap());
+    }
+
+    #[test]
+    fn can_read_app_info_from_multi_trigger_template_v2() {
```
itowlson: We probably want to test somewhere that we can add a new trigger type to an existing (single-trigger-type) manifest.

crates/templates/src/template.rs
michelleN: yeah so if you trace the function call, `infer_trigger_type` is only being used for v1 manifests but your suggestion on ditching the check altogether makes sense to me.

crates/templates/src/app_info.rs
michelleN: That makes sense to me. I reworked this PR to reflect that flow. Lmk what you think.

itowlson (approved):
Wow, so much code that does nothing except check trigger types.  Good riddance to it!  Thanks for this, it looks good!

crates/templates/src/app_info.rs
itowlson: Did we end up with a test for this?  It's not a blocker, just a passing thought.

michelleN: No we didn't since all the trigger validation logic is gone now.

PR #3630, opened 2026-07-22
`spin targets` command to list/update envs

### LINKED ISSUE #3621: Environment management commands?
I had hoped to make the whole environments stuff mostly invisible but I am wondering if we need to surface commands after all to:

* force-update the local environments catalogue (currently it updates only when the user references an env, and the env is 1. missing, or 2. unversioned and it's been more than N hours)
* list the envs (in case someone needs to remember the exact spelling of their favourite env)

I don't feel great about this because envs should be something that people are directed to (just use the one your platform docs tell you to!), not something that they browse or have to manage, and a new command is more annoying cognitive load that's always there even though you only use it every blue moon, but... do we think we are okay without these commands?

(I guess an alternative is to have an update command under the hidden `spin maintenance` tree, and not have a list.  But I am not in love with telling users to run hidden maintenance commands.)

itowlson:
Fixes #3621.

```
$ spin targets list
akamai-functions
spin-up@3.2
spin-up@3.4
spin-up@3.5
spin-up@3.6

$ spin targets update
Target environments updated
```

Reviewers are advised to strap themselves in for a rollercoaster thrill ride of high-stakes sorting code.

### DIFF
--- a/src/commands/target_environments.rs
+++ b/src/commands/target_environments.rs
@@ -1,0 +1,164 @@
+use clap::Subcommand;
+use semver::Version;
+use std::cmp::Ordering;
+
+/// Commands for the target environments catalogue.
+#[derive(Subcommand, Debug)]
+pub enum TargetEnvironmentCommands {
+    /// List known target environments.
+    List,
+
+    /// Update the target environments from the remote repository.
+    Update,
+}
+
+impl TargetEnvironmentCommands {
+    pub async fn run(self) -> anyhow::Result<()> {
+        match self {
+            Self::List => list().await,
+            Self::Update => update().await,
+        }
+    }
+}
+
+async fn list() -> anyhow::Result<()> {
+    let catalogue = spin_environments::Catalogue::try_default()?;
+
+    let mut envs = catalogue.list().await;
+    envs.sort_by(|p, q| order_versioned(p, q));
+
+    if envs.is_empty() {
+        eprintln!(
+            "No target environments found. Run `spin targets update` to fetch target environments."
+        );
+    }
+
+    for env in &envs {
+        println!("{env}");
+    }
+
+    Ok(())
+}
+
+async fn update() -> anyhow::Result<()> {
+    let catalogue = spin_environments::Catalogue::try_default()?;
+
+    catalogue.update().await?;
+
+    eprintln!("Target environments updated");
+    Ok(())
+}
+
+fn order_versioned(first: &str, second: &str) -> Ordering {
+    let (name1, ver1) = name_and_version(first);
+    let (name2, ver2) = name_and_version(second);
+
+    match name1.cmp(name2) {
+        Ordering::Equal => match (ver1, ver2) {
+            (None, None) => Ordering::Equal,
+            (None, Some(_)) => Ordering::Less,
+            (Some(_), None) => Ordering::Greater,
+            (Some(ver1), Some(ver2)) => order_env_ver(ver1, ver2),
+        },
+        other => other,
+    }
+}
+
+fn name_and_version(env_name: &str) -> (&str, Option<&str>) {
+    match env_name.rsplit_once('@') {
+        Some((n, v)) => (n, Some(v)),
+        None => (env_name, None),
+    }
+}
+
+fn order_env_ver(first: &str, second: &str) -> Ordering {
+    // Environment versions are expected to be major.minor, but this allows for
+    // some crazy future where a bold young experimentalist creates a major-only
+    // or patch-specific env.
+
+    // Leverage semver for parsing and ordering
+    fn padded_version(v: &str) -> Option<Version> {
+        let padded = match v.split('.').count() {
+            1 => format!("{v}.0.0"),
+            2 => format!("{v}.0"),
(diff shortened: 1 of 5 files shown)


fibonacci1729:
I think a neat followup maybe would be a `spin targets check` for devs that didnt include a `-E` when scaffolding but want to answer the question : "can I deploy this to X?"

In any case, ill review this in the AM.

THANK YOU

src/commands/target_environments.rs
```diff
@@ -0,0 +1,160 @@
...
+            order_versioned("capybara@3.2.1", "capybara")
+        );
+        assert_eq!(
+            Ordering::Greater,
+            order_versioned("capybara@3.2.1", "amphipod")
+        );
+
+        assert_eq!(
+            Ordering::Less,
+            order_versioned("capybara.2.1", "cassowary@1.2.3")
```
fibonacci1729: Should this be `capybara@2.1`?

src/commands/target_environments.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+fn name_and_version(env_name: &str) -> (&str, Option<&str>) {
+    match env_name.rsplit_once('@') {
+        Some((n, v)) => (n, Some(v)),
+        None => (env_name, None),
+    }
+}
+
+fn order_semver(first: &str, second: &str) -> std::cmp::Ordering {
```
fibonacci1729:
```suggestion
fn order_semver(first: &str, second: &str) -> Ordering {
```

`Ordering` is imported in prelude.

src/commands/target_environments.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    for env in &envs {
+        println!("{env}");
+    }
+
+    Ok(())
+}
+
+async fn update() -> anyhow::Result<()> {
```
fibonacci1729, line 43:
```suggestion
async fn update() -> Result<()> {
```

`list` used the unprefixed `Result`. Don't care either way but prefer consistency.

src/commands/target_environments.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+async fn list() -> Result<()> {
+    let catalogue = spin_environments::Catalogue::try_default()?;
+
+    let mut envs = catalogue.list().await?;
+    envs.sort_by(|p, q| order_versioned(p, q));
+
+    if envs.is_empty() {
```
fibonacci1729, line 30: I believe this is only going to fire if the directory exists but is empty. I.e. on a new machine (or after just deleting the cache) a user will see: `No such file or directory ...` instead of the hint (since in `Catalogue::list` the read_dir will error on missing and propagate through here). Ideally `Catalogue::list` would do this check and return the empty vec.

src/commands/target_environments.rs
fibonacci1729: I don't think this is quite correct. `Version::parse` requires `major.minor.patch` so `3.2` fails to parse and falls back to the `else` branch which lexicographically  string compares which will yield incorrect sorts once a 2-digit minor version is present. E.g. this test fails:

```assert_eq!(Ordering::Greater, order_versioned("capybara@3.10", "capybara@3.2"));```

**Review by fibonacci1729 — CHANGES_REQUESTED** [review]

File: src/commands/target_environments.rs (hunk 2 above)
Comment (fibonacci1729) [reply src/commands/target_environments.rs]: Verified that this works:

```rust

fn order_semver(first: &str, second: &str) -> Ordering {
    match (padded_version(first), padded_version(second)) {
        (Some(v1), Some(v2)) => v1.cmp(&v2),
        _ => first.cmp(second),
    }
}

fn padded_version(v: &str) -> Option<Version> {
    let padded = match v.split('.').count() {
        1 => format!("{v}.0.0"),
        2 => format!("{v}.0"),
        _ => v.to_string(),
    };
    Version::parse(&padded).ok()
}
```

src/commands/target_environments.rs
itowlson: Oh man that really puts a hole in the supposed thoroughness of this test.  ~Yes it should~ and thanks for catch! (correction: should have been `@3.2.1`)

src/commands/target_environments.rs
itowlson: Good catch.  The only other place I use `Catalogue::list()` I `unwrap_or_default()` it; it's better that `list` itself should handle that.

src/commands/target_environments.rs
itowlson: I think WIT versions are always 3-digit though aren't they?  (and yeah sorry I had misremembered that test case was meant to be 3 digits too)  If versions don't parse as semver `Version` values then a lot of other stuff might break too...

fibonacci1729: Possibly? But these versions come from the environment name.

itowlson: I'm an idiot.  You're right.  Thanks!

itowlson: Fixed now - thank you for saving me from myself.

fibonacci1729 (approved):

PR #3635, opened 2026-07-23
Add spin targets check command

### LINKED ISSUE #3634: Add a `spin targets check`
Today the _only_ way to answer "can I deploy this to X?" is to declare `targets` in the manifest and run `spin build`. A developer who scaffolded their application without `-E` has no ad-hoc way to ask the question without editing their manifest.

Alternatively we could add a `--targets` flag to `spin build` but I think the naming proposed in this issue is the more discoverable option.

fibonacci1729:


src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+        long = "from",
+        alias = "file",
+    )]
+    pub app_source: Option<PathBuf>,
+
+    /// The target environment(s) to check against. This may be a catalogue id (for
+    /// example `spin-up@3.6`), an `http:` URL, or a `file:` path, and may be specified
+    /// multiple times. If omitted, the targets declared in the application manifest
+    /// are checked.
+    #[clap(short = 'E', long = "target")]
```
itowlson, line 79: Elsewhere the long form is defaulted to `--target-environment`.

src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+    pub target_environment: Vec<String>,
+
+    /// The build profile to use when building the application before checking.
+    #[clap(long)]
+    #[arg(add = clap_complete::ArgValueCandidates::new(crate::completions::profiles))]
+    pub profile: Option<String>,
+
+    /// Check the application as already built, rather than building it first.
+    #[clap(long)]
+    pub no_build: bool,
```
itowlson, line 90: In most other commands, the default is to _not_ build; `--build` is the option that can be added.  The counterexample is `spin watch`, but that's essentially build and run commands in a trenchcoat, and I'm not sure this is.  But agree this one can be argued either way.

itowlson, line 90: I believe our usual convention for things like this is `skip-` (cf. `spin watch --skip-build`, `spin build --skip-target-checks`).

src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+}
+
+async fn check(cmd: CheckCommand) -> anyhow::Result<()> {
```
itowlson, line 93: Everywhere else the convention is `impl FooCmd { async fn run(&self) }` - is there a reason for doing this differently?

(Okay, there are exceptions like `spin plugins update`, but I think those are only when the command has no arguments.)

src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+    // Target checking inspects each component's compiled Wasm, so the application
+    // must be built first. We skip Spin's own manifest-driven target checks here,
+    // because we run our own check against the requested environments below.
+    if !cmd.no_build {
+        spin_build::build(
+            &manifest_file,
+            profile,
+            &[],
+            spin_build::TargetChecking::Skip,
+            spin_build::GenerateDependencyWits::Generate,
```
itowlson, line 110: It feels a bit weird to me that running a check command would modify the dependency WITs - a check command feels like it should be non-side-effectful.  I guess it's possble that build would fail if we didn't, is that the idea?

src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+        .await?;
+    }
+
+    let manifest = spin_manifest::manifest_from_file(&manifest_file)
+        .context("Failed to read application manifest for target environment checking")?;
+
+    // Environments named on the command line take precedence; otherwise fall back
+    // to the targets declared in the manifest.
+    let target_refs: Vec<TargetEnvironmentRef> = if cmd.target_environment.is_empty() {
+        manifest.application.targets.clone()
```
itowlson, line 122: This seems equivalent to a `spin build`.  Is the idea to enable `spin targets check --skip-build` to validate without rebuilding?

src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+    } else {
+        cmd.target_environment
+            .iter()
+            .map(|env| crate::parse_env::parse_env(env))
+            .collect()
+    };
+
+    if target_refs.is_empty() {
```
itowlson, line 130: Maybe we should check this _before_ slogging through a potentially lengthy build!

src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+    }
+
+    let application =
+        spin_environments::ApplicationToValidate::new(manifest, &[], profile, &app_dir)
+            .await
+            .context("unable to load application for checking against target environments")?;
+
+    let targets = spin_environments::Targets {
+        default: &target_refs,
+        overrides: Default::default(),
```
itowlson, line 143: Hmm, this is interesting.  If the app says "but hey I intend to deploy the `draw-an-owl` component in the `mighty-gpu` environment," is it correct to fail a `spin targets check` with "no, `draw-an-owl` doesn't work in the app environment"?  (Almost certainly not.)  Is it correct to fail a `spin targets check -E no-gpus-here` on `draw-an-owl`?  (I am not sure.)

src/commands/target_environments.rs
```diff
@@ -49,6 +59,113 @@ async fn update() -> anyhow::Result<()> {
...
+        targets,
+        None,
+        &app_dir,
+    )
+    .await
+    .context("unable to check if the application is compatible with the target environments")?;
+
+    if validation.is_ok() {
+        terminal::step!(
+            "Compatible",
```
itowlson, line 157: If I read the `step` macro right, this will print "Compatible the application is compatible..."  Is that what you see or am I misreading?  If that's correct, we should add some punctuation.

itowlson (changes requested):
A few consistency quirks, but the main issue is how to handle per-component targets already declared in the app.

I am pretty sure the proposed behaviour is incorrect if the user runs `spin target check` to check against the env in the manifest.  I feel like this should succeed iff `spin build` succeeds - it would be confusing for `spin build` to say "you're good to go" and "spin targets check" to say "not so fast, buddy."  I am not sure what the correct behaviour is if the user runs `spin target check -E` - on the one hand, the manifest says "hey these components are gonna run in a different environment", on the other hand, you did ask if the _app_ would run in _this specific environment._

Understanding the use cases might be helpful for figuring out the right answer.  I find the ad hoc validation use case a bit hard to get my head around so I am probably not seeing the context that would help clarify this!

itowlson:
Also realised, if we are doing this then I think we should add `target` as an alias for `targets` because, as you can see from my review, I for one will always, _always_ forget it in this command!

itowlson:
It occurred to me this should force a catalogue update if the env is unversioned - we don't do this during build because we don't want to delay every. single. build. but when the user specifically asks for a check I think it's worth the network hop to make sure they're checking the latest env.

src/commands/target_environments.rs
fibonacci1729: Changing this to `build`.

itowlson:
@fibonacci1729 I saw a bunch of thumbs but I am not sure if there have been new pushes - happy to re-review when you're ready.

fibonacci1729:
> Re: how should `spin targets check` treat components that declare their own targets in the manifest?

I think i'd propose:
1. No `-E`: mirror spin build exactly
2. With `-E`: Replaces the app-level default, per-component targets still win (print a note listing which components used their own targets).

Quite frankly I forgot about per-component targets and I find it deeply confusing when trying to rationalise it with the app level targets.

itowlson:
@fibonacci1729 That makes sense.  Without `-E` it's asking "is this app compliant with where it expects to go"; with `-E` it's asking "is this app compliant with _this specific env_".  We can always revisit if the whole "I want to deploy different components to different envs" stuff gets traction (and there are enough envs for people to want to check against ones other than the app one).

fibonacci1729:
Closing this until there is a concrete usecase.

PR #3644, opened 2026-07-28
feat(kv): support configuring concurrent operation limits

kate-goldenring:
- uses a semaphore to control concurrent KV operations
- follows a similar pattern that we took with other factors here https://github.com/spinframework/spin/pull/3552
- limit is still unset in cosmos and redis KV factors

### DIFF
--- a/crates/factor-key-value/src/host.rs
+++ b/crates/factor-key-value/src/host.rs
@@ -332,10 +332,8 @@ impl<T> v3::HostStoreWithStore<T> for crate::KeyValueFactorData {
             (host.get_store(store).cloned(), host.semaphore.clone())
         });
         let store = store_result.map_err(|_| v3::Error::NoSuchStore)?;
-        let _permit = permit_fut
-            .acquire()
+        let _permit = acquire_permit_v3(&permit_fut)
             .await
-            .map_err(log_error_v3)
             .map_err(track_error_on_span_v3)?;
         store
             .get(&key, MAX_HOST_BUFFERED_BYTES)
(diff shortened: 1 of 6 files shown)


crates/factor-key-value/src/lib.rs
```diff
@@ -87,9 +90,26 @@ impl Factor for KeyValueFactor {
...
         }
 
+        let app_id: Arc<str> = ctx
+            .app()
+            .get_metadata(APP_NAME_KEY)?
+            .unwrap_or_else(|| "<unnamed>".into())
+            .into();
+
+        let semaphore = ConnectionSemaphore::new(
+            None,
```
rylev, line 103: It would be good to leave a comment calling out that we are not including the global connection limit here because we want to treat kv limits independently.

crates/factor-key-value/src/host.rs
```diff
@@ -280,13 +326,17 @@ impl<T> v3::HostStoreWithStore<T> for crate::KeyValueFactorData {
...
+            let host = access.get();
+            host.otel.reparent_tracing_span();
+            (host.get_store(store).cloned(), host.semaphore.clone())
+        });
+        let store = store_result.map_err(|_| v3::Error::NoSuchStore)?;
+        let _permit = permit_fut
+            .acquire()
+            .await
+            .map_err(log_error_v3)
+            .map_err(track_error_on_span_v3)?;
```
rylev, line 337: Can we update `track_error_on_span` functions to not treat `Error::Other` as a host error when the error is about too many requests?

rylev (approved):
Some nits, but looks good. Thanks!

crates/factor-key-value/src/host.rs
kate-goldenring: good call. i also made sure we always map the error to a user friendly "too many requests" error

PR #3649, opened 2026-08-04
Middleware host requirement

### LINKED ISSUE #3637: Middleware host requirement
The CLI and manifest format now allow for middleware.  If a downstream host doesn't implement middleware, we want that middleware to set off alarm bells rather than being quietly ignored.   The mechanism for this is to list a host requirement in the lock file (similar to what we do for service chaining).  This would also allow it to participate in target env checks.

itowlson:
Fixes #3637.

### DIFF
--- a/crates/loader/src/local/trigger_components.rs
+++ b/crates/loader/src/local/trigger_components.rs
@@ -1,6 +1,8 @@
 use std::collections::{HashMap, HashSet};
 
-use spin_locked_app::locked::{LockedApp, LockedComponent, LockedTrigger};
+use spin_locked_app::locked::{
+    HOST_REQ_REQUIRED, LockedApp, LockedComponent, LockedTrigger, TRIGGER_DEPENDENCIES_KEY,
+};
 
 /// We want all component/composition graph information to be in the component,
 /// because the component ID is how Spin looks this stuff up. So if a trigger
@@ -76,6 +78,9 @@ fn move_deps_from_triggers_to_components(locked: &mut LockedApp) {
                 "resolve-trigger-dependencies-using".into(),
                 trigger.trigger_type.clone().into(),
             );
+            component
+                .host_requirements
+                .insert(TRIGGER_DEPENDENCIES_KEY.into(), HOST_REQ_REQUIRED.into());
             trigger.trigger_dependencies.clear();
         }
     }
--- a/crates/loader/src/local/trigger_components.rs
+++ b/crates/loader/src/local/trigger_components.rs
@@ -1,7 +1,7 @@
 use std::collections::{HashMap, HashSet};
 
 use spin_locked_app::locked::{
-    HOST_REQ_REQUIRED, LockedApp, LockedComponent, LockedTrigger, TRIGGER_DEPENDENCIES_KEY,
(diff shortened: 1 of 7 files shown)


lann (approved):
This is using trigger dependencies as a _proxy_ for middleware, right? I think that's pragmatically fine given that we have no other use for trigger dependencies on the horizon, but just confirming that its a compromise we're making on purpose.

itowlson:
@lann Yeah, I raised the "middleware vs trigger-deps vs HTTP-trigger-deps vs ..." in discussion elsewhere, and you said you didn't have an opinion.  So I went with trigger-deps because it's easiest, but you are correct, that means that hosts are promising "I understand trigger deps" rather than "I understand _this particular_ trigger dep."  I can change it back to be middleware-specific: that requires the lockfile builder to have knowledge of individual triggers, but that's not a blocker, merely a compromise of Spin's hitherto-flawless ladder of architectural purity.

fibonacci1729:
Oof sorry I should have taken my OOB position more strongly. This point is exactly why I think host requirements should not be generic.

itowlson:
Okay, so is the consensus to make this specifically detect the middleware situation?  Not a problem, just want to be sure before I make the change - thanks!

lann:
I don't really understand the work involved but it does seem like it would be a better solution.

fibonacci1729:
> requires the lockfile builder to have knowledge of individual triggers

Is this actually true? Maybe I am looking at the wrong place but why not just plop the metadata in when reassigning trigger dependencies.

itowlson:
Yes, that's part of the lockfile builder.  It is currently agnostic and just transfers trigger deps.  Now it will need to have specific understanding of the trigger dep key "middleware."

itowlson:
Which is fine and I have it ready to go if that is what we agree.

itowlson:
Is that a thumb of "yes I agree this is how it should work" or a thumb of "I acknowledge the status"? _grin_

fibonacci1729:
Agreeing to making the change and how it should work.

crates/loader/src/local/trigger_components.rs
```diff
@@ -76,6 +78,11 @@ fn move_deps_from_triggers_to_components(locked: &mut LockedApp) {
                 "resolve-trigger-dependencies-using".into(),
                 trigger.trigger_type.clone().into(),
             );
+            if component.trigger_dependencies.contains_key("middleware") {
```
fibonacci1729, line 81:
```suggestion
            if component.trigger_dependencies.contains_key(MIDDLEWARE_KEY) {
```

fibonacci1729, line 81: Oh easy peasy

fibonacci1729 (approved):

crates/loader/src/local/trigger_components.rs
itowlson: No, `MIDDLEWARE_KEY` is the hostreq key.  There's no guarantee that this is the same as the manifest key.

_ETA:_ It _does_ need to line up with what the `spin-trigger-http` composer looks at, but I don't want to create a cross-dependency from the loader to the trigger.

PR #3651, opened 2026-08-05
Do not pass Connection header to guest

### LINKED ISSUE #3650: Drop incoming `Connection` header before passing request to guest
Because it causes composition/middleware to error.  See discussion at https://github.com/spinframework/spin-rust-sdk/issues/97#issuecomment-5179927388

itowlson:
Fixes #3650.

I am not sure if this constitutes a breaking change, or if it cause any other problems, because if there is one thing I have learned, it is that IT IS NEVER THAT EASY.

### DIFF
--- a/crates/trigger-http/src/headers.rs
+++ b/crates/trigger-http/src/headers.rs
@@ -88,6 +88,9 @@ pub fn strip_forbidden_headers(req: &mut Request<Body>) {
     {
         headers.remove("Host");
     }
+    // This header is transport information - only meaningful to the server
+    // itself - and causes problems if guests need to forward requests.
+    headers.remove("Connection");
 }
 
 pub fn prepare_request_headers(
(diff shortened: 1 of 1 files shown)


crates/trigger-http/src/headers.rs
```diff
@@ -88,6 +88,9 @@ pub fn strip_forbidden_headers(req: &mut Request<Body>) {
     {
         headers.remove("Host");
     }
+    if headers.contains_key("Connection") {
+        headers.remove("Connection");
+    }
```
lann: Remove already basically does "if contains_key" internally. More importantly: this could also use a comment, maybe:
```suggestion
    // This header is transport information - only meaningful
    // to the server itself - and has caused problems in guests.
    headers.remove("Connection");
```

lann:
I don't think this is "logically" breaking. Someone could be using it in some awful hacky way to sniff some transport detail but that would almost certainly be unportable anyway.

lann (approved):
Thanks!

PR #3653, opened 2026-08-05
Bump Wasmtime to v48.0.0

spinframeworkbot:
Bumps Wasmtime to v48.0.0 prerelease

### DIFF
--- a/crates/core/tests/integration_test.rs
+++ b/crates/core/tests/integration_test.rs
@@ -10,7 +10,6 @@ use spin_factor_wasi::{DummyFilesMounter, WasiFactor};
 use spin_factors::{App, AsInstanceState, RuntimeFactors};
 use spin_locked_app::locked::LockedApp;
 use tokio::fs;
-use wasmtime_wasi::I32Exit;
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_max_memory_size_obeyed() {
@@ -40,11 +39,7 @@ async fn test_max_memory_size_violated() {
     )
     .await
     .unwrap_err();
-    let trap = err
-        .root_cause() // The error returned is a backtrace. We need the root cause.
-        .downcast_ref::<I32Exit>()
-        .expect("trap error was not an I32Exit");
-    assert_eq!(trap.0, 1);
+    assert!(format!("{err:?}").contains("command failed"))
 }
 
 // FIXME: racy timing test
--- a/crates/factor-wasi/src/lib.rs
+++ b/crates/factor-wasi/src/lib.rs
@@ -393,6 +393,7 @@ impl Factor for WasiFactor {
         ctx: PrepareContext<T, Self>,
     ) -> anyhow::Result<InstanceBuilder> {
         let mut wasi_ctx = WasiCtxBuilder::new();
+        wasi_ctx.allow_tcp(true).allow_udp(true);
 
         // Mount files
         let mount_ctx = MountFilesContext { ctx: &mut wasi_ctx };
@@ -451,8 +452,7 @@ impl MountFilesContext<'_> {
         } else {
             FsPerms::ReadOnly
         };
-        self.ctx
-            .preopened_dir(host_path, guest_path, perms)?;
+        self.ctx.preopened_dir(host_path, guest_path, perms)?;
         Ok(())
     }
 }
@@ -520,8 +520,7 @@ impl InstanceBuilder {
         } else {
             FsPerms::ReadOnly
         };
-        self.ctx
-            .preopened_dir(host_path, guest_path, perms)?;
+        self.ctx.preopened_dir(host_path, guest_path, perms)?;
         Ok(())
     }
 }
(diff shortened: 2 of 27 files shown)


itowlson:
I bumped it to 1.95 and fixed up the low-hanging fruit around filesystem permissions, but there's a big pile of errors around changes to the sockets API that I don't know how to start on.

tests/manual/pg-ssl-root-certs/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "pg-app"
 authors = ["Kate Goldenring <kate.goldenring@fermyon.com>"]
 description = ""
 version = "0.1.0"
-rust-version = "1.94"
+rust-version = "54"
```
fibonacci1729: Err

examples/open-ai-rust/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "open-ai-rust"
 authors = ["Spin Framework Contributors"]
 description = "Example showing using openAI with Spin"
 version = "0.1.0"
-rust-version = "1.94"
+rust-version = "54"
```
fibonacci1729: Ope

tests/manual/pg-ssl-root-certs/Cargo.toml
itowlson: bloody hell

itowlson:
I updated the SocketAddrUse stuff based on https://github.com/bytecodealliance/wasmtime/pull/13934, but it needs someone to check it because I am somewhat guessing!

A lot of the upstream sockets stuff has changed from sync to async or from async to sync.  For some of it I was able to propagate the change without obvious problems, but there are some cases (I think all implementing RCs) where we are constrained by a sync trait but the underlying API is now async, and my usual kludges for bridging over that feel ill-advised.  So @alexcrichton or @dicej I think I am going to need to punt those to you - sorry.

dicej:
Thanks for getting this going, @itowlson!  I'm addressing the remaining issues and will push an update.

crates/factor-wasi/src/lib.rs
```diff
@@ -393,6 +393,7 @@ impl Factor for WasiFactor {
         ctx: PrepareContext<T, Self>,
     ) -> anyhow::Result<InstanceBuilder> {
         let mut wasi_ctx = WasiCtxBuilder::new();
+        wasi_ctx.allow_tcp(true).allow_udp(true);
```
alexcrichton, line 396: One thing on this -- this probably only wants to happen if sockets/addresses are otherwise allowed for this component. Otherwise this is technically a vector by which guests can create fds on the host and may cause minor issues. Would it be possible to make these conditional based on other settings? Or are guests expected to always be able to create a socket and outgoing connections?

crates/core/tests/integration_test.rs
```diff
@@ -40,11 +39,7 @@ async fn test_max_memory_size_violated() {
     )
     .await
     .unwrap_err();
-    let trap = err
-        .root_cause() // The error returned is a backtrace. We need the root cause.
-        .downcast_ref::<I32Exit>()
-        .expect("trap error was not an I32Exit");
-    assert_eq!(trap.0, 1);
+    assert!(format!("{err:?}").contains("command failed"))
```
alexcrichton, line 42: This looks a bit suspect to me, is `I32Exit` no longer showing up? If so that might be a bug

crates/factor-wasi/src/lib.rs
lann: Should be possible to extend `OutboundAllowedHosts` with a "is this protocol _ever_ allowed" method.

crates/core/tests/integration_test.rs
dicej: I've gone back as far as Spin v3.6.0 and Rust 1.90 (and various versions in between), and this test is failing in that version and every version since (and possibly earlier).  I have no idea how CI has been green all that time.

This came up during the Wasmtime 46 upgrade as well, but I wrote it off because it only seemed to be happening on my machine and not on CI.  Now it's happening in both places, and I can't tell what changed.

alexcrichton: Ah ok if this is just a historical thing then that's fine yeah, I mostly just wasn't aware of anything in 48.0.0 that would have affected this

dicej: I _think_ this could be related to the version of Rust used to build the Wasm code (and consquently the version of WASI-SDK used?) which could influence whether an error exit means returning an `err` from `wasi:cli/run` vs. calling `wasi:cli/exit{_with_code}`, but I haven't had a chance to confirm.

dicej: If the above is correct, then probably the best thing to do here is look for an `I32Exit` _or_ a "command failed", since either could be valid depending on how the component was built.

alexcrichton: Ah yeah I wouldn't worry much about this, it's definitely unrelated to the 47->48 update which is all I wanted to bottom out

itowlson:
I have run the update workflow so this is now on Wasmtime 48 final.  Looks like the script ran flawlessly @fibonacci1729!

alexcrichton (approved):

crates/factor-wasi/src/lib.rs
dicej: I opened an issue for this: https://github.com/spinframework/spin/issues/3680

dicej (approved):

PR #3654, opened 2026-08-08
Switch rustls crypto provider from ring to aws_lc_rs

### LINKED ISSUE #3502: FIPS Support
## Feature

I'd like to be able to run a build of Spin (via SpinKube) that has wasi-http configured to run in FIPS mode. This would mean that all https requests and responses (as well as new outbound requests) would use a NIST-certified crypto module.
This could just be a separate build of Spin or containerd-shim-spin, and wouldn't need to be dynamic per-request or per-VM.

## Benefit

If I want to run SpinKube in a US Gov "FedRAMP Moderate" or higher environment, I need to be able to show auditors that any encryption-in-transit is done with a crypto module that's been NIST-certified as FIPS-140-2 or FIPS-140-3 compliant. As an example, the Golang 1.24+ crypto module 1.0.0 was [recently validated](https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/5247), so any golang programs [built with the appropriate flags](https://go.dev/doc/security/fips140#the-gofips140-environment-variable) are guaranteed to use the appropriate crypto implementation at runtime. Ideally users of Spin and SpinKube could rely on something similar for their WASI implementations, and the wasi-http components just get it for free, regardless of how they were compiled or what language was used.

This is a follow-on issue to https://github.com/bytecodealliance/wasmtime/issues/13293, created at the request of @fibonacci1729.  We chatted in the CNCF slack and he had a really promising proposal for how it could work.

## Concrete Items

* A FIPS build of spin available as part of the release from main.
* A FIPS build of [containerd-shim-spin](https://github.com/spinframework/containerd-shim-spin) available as part of the release from main.
* A docs page for [spinkube.dev](https://www.spinkube.dev/docs/install/) on how to run it with FIPS (including calling out the need for a FIPS OpenSSL hosting environment for the non-rust, dynamically-linked deps).

ChihweiLHBird:
Install `aws_lc_rs` as the process-wide provider at startup. ring is still compiled via transitive `rustls` features; removing it is a follow-up.

Replace the install helper with `spin_tls::get_or_install_default_crypto_provider()`, which returns the installed default, respecting a provider installed by embedders (the old helper would panic), and installs aws-lc-rs if none is set. The outbound client and `trigger-http` server TLS configs now build with `aws_lc_rs` provider explicitly.

Partially fix https://github.com/spinframework/spin/issues/3502

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -492,6 +492,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.17.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4342d8937fc7e5dd9b1c60292261c0670c882a2cd1719cfc11b1af41731e32ad"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.42.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d9ceb1da931507a12f4fccea479dccd00da1943e1b4ae72d8e502d707361444"
+dependencies = [
+ "cc",
+ "cmake",
+ "dunce",
+ "fs_extra",
+ "pkg-config",
+]
+
+[[package]]
 name = "aws-runtime"
 version = "1.5.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -8111,6 +8134,7 @@ version = "0.23.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6b92b125634d9b795e7beca796cc790df15a7fb38323bf3196fda83292d06b1f"
 dependencies = [
+ "aws-lc-rs",
  "log",
  "once_cell",
  "ring",
@@ -8239,6 +8263,7 @@ version = "0.103.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "61c429a8649f110dddef65e2a5ad240f747e85f7758a6bccc7e5777bd33f756e"
 dependencies = [
+ "aws-lc-rs",
  "ring",
  "rustls-pki-types",
  "untrusted",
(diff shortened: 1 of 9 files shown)


lann:
> rustls auto-detection would panic on paths that bypass Spin's entrypoints

Could you say more about this? Where would one need to bypass Spin's entrypoints? 🤔

ChihweiLHBird:
> > rustls auto-detection would panic on paths that bypass Spin's entrypoints
> 
> Could you say more about this? Where would one need to bypass Spin's entrypoints? 🤔

Sorry... this is still a draft. The scenario in my mind was if another thing uses Spin as a crate and install ring as default. I know it's not supposed to be used this way, but in theory it might happen. So, I left that in draft and will refine it later.

lann:
If something is embedding spin it _should_ be able to pick the provider. If we're changing the default for `spin up` then that should just be scoped to the `spin` binary crate I think.

ChihweiLHBird:
@lann the original way expects the default is not installed:
```rust
rustls::crypto::ring::default_provider()
    .install_default()
    .expect("failed to install rustls ring crypto provider");
```

It's from my previous PR https://github.com/spinframework/spin/pull/3590

The implementation in this PR is theoradically safer even though the panic scenario is very unlikely to be triggerred.

lann:
Right but that is only called from spin trigger implementations (which are effectively the root binary crate in this case).

ChihweiLHBird:
@lann Sorry for the confusion. I looked into it again, and found my earlier explanation was wrong.

Flipping the workspace `rustls` feature to `aws_lc_rs` does not remove `ring`, so without these guards of installing default and build `ClientConfig` with a specific provider via `builder_with_provider`, it may panic because 2 providers are compiled in. In many code paths, e.g. some spin tests, who doesn't call the install default helper, will panic, if without `builder_with_provider`.

Four tests in `crates/factor-outbound-networking/src/tls.rs` and `tests/factor_test.rs` all build a `ClientConfig` without going through a Spin entrypoint. That's what the `builder_with_provider` calls guard.

I can also split the helper into `install_default_crypto_provider()` (entrypoints only) and `default_crypto_provider()` for the config builders. Does that sound like a good idea?

itowlson:
@lann bump

PR #3655, opened 2026-08-10
Use async (tokio) tar crate for OCI

### LINKED ISSUE #2846: Replace async-tar dependency

itowlson:
Fixes #2846.

This mostly reverts (the relevant bits of) #3340 but now with `astral-tokio-tar` (which appears to be the current inheritor of the lineage) instead of the old async-std-based `async-tar`.

It is now PERFECT and FLAWLESS and I will NEVER HAVE TO FAFF AROUND WITH `tar` CRATES EVER AGAIN.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,6 +206,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
+name = "astral-tokio-tar"
+version = "0.6.4"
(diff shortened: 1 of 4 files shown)


itowlson:
Bump

lann (approved):

PR #3656, opened 2026-08-10
BREAKING: change default for plugins list to only show summary and not the full list

### LINKED ISSUE #3648: Change default behavior of `spin plugins list`
The list of plugin version being shown when running `spin plugins list` is really long. I'd like to propose changing it to only show latest version (and installed version of a given plugin, if it's not the latest).

Current
```shell
trigger-command 0.1.0
trigger-command 0.2.0
trigger-command 0.2.1
trigger-command 0.2.2
trigger-command 0.3.0
trigger-command 0.3.1
trigger-command 0.3.2
trigger-command 0.4.0
trigger-command 0.4.1
trigger-command 0.5.0
trigger-command 0.5.1
trigger-command 0.5.2
trigger-command 0.5.3
trigger-command 0.6.0
trigger-command 0.7.0
```

Proposed
```shell
trigger-command 0.6.0 [installed]
trigger-command 0.7.0 [latest]
```

Having had conversations with LLMs there's an option to do a strict migration or soft deprecation of default behavior with the existing `--summary` flag.

**Strict cutover with --all kept**

New default: spin plugins list shows latest only.
--all means show all versions (same name as today).
Remove --summary entirely.
No compatibility aliasing.

_Impact:_
Users relying on default output (all versions) will see changed behavior.
Users already using --all in scripts should continue to work with little or no change.

**Soft migration with --all kept**

Same new default: latest only.
Keep --all for all versions.
Keep --summary temporarily as deprecated (hidden or warned) for one release cycle.
Optionally print deprecation warning when --summary is used.

_Impact:_
Minimal disruption for users who learned/used --summary.
Easier comms and rollout, then remove --summary later.
What this means compared to renaming to --all-versions
Keeping --all removes one migration burden: no need to teach a new flag name.
Most compatibility work shifts to handling --summary only.

In practice, strict vs soft mostly becomes:
Strict: remove --summary now.
Soft: deprecate --summary, remove later.

Recommendation
Keep --all and do soft migration if you want lowest user friction.
Keep --all and do strict cutover if you want the cleanest CLI immediately and can tolerate minor breakage for --summary users.

I'd like some feedback on which path to go down - soft or strict. Thanks.

mikkelhegn:
Implements #3648

### DIFF
--- a/src/commands/plugins.rs
+++ b/src/commands/plugins.rs
@@ -755,11 +755,11 @@ pub struct List {
     #[clap(long, group = "which")]
     pub installed: bool,
 
-    /// List all versions of plugins. This is the default behaviour.
+    /// List all versions of plugins.
     #[clap(long, group = "which")]
     pub all: bool,
 
-    /// List latest and installed versions of plugins.
+    /// List latest and installed versions of plugins. This is the default behaviour.
     #[clap(long, group = "which")]
     pub summary: bool,
 
@@ -798,7 +798,7 @@ impl List {
             }
         }?;
 
-        if self.summary {
+        if self.should_summarise() {
             plugins = summarise(plugins);
         }
 
@@ -850,6 +850,10 @@ impl List {
 
     fn target_environment(&self) -> OptionalValueFlag {
         (&self.target_environment).into()
+    }
+
+    fn should_summarise(&self) -> bool {
+        self.summary || (!self.installed && !self.all)
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/commands/plugins.rs
```diff
@@ -851,6 +851,10 @@ impl List {
     fn target_environment(&self) -> OptionalValueFlag {
         (&self.target_environment).into()
     }
+
+    fn should_summarise(&self) -> bool {
+        self.summary || !self.all
```
itowlson: You probably don't want to summarise in the presence of `--installed` either.

mikkelhegn: Got it, added a bunch of tests as well :-)

src/commands/plugins.rs
```diff
@@ -1192,4 +1207,46 @@ mod test {
...
+
+    #[test]
+    fn list_summary_enables_summary_mode() {
+        let list = List::parse_from(["list", "--summary"]);
+        assert!(list.should_summarise());
+    }
+
+    #[test]
+    fn list_summary_overrides_all_when_constructed() {
+        let list = test_list(false, true, true);
```
itowlson: clap should not allow this state (or the states in the next two tests) - at most one of `--installed`, `--all` and `--summary` can be set.

src/commands/plugins.rs
```diff
@@ -851,6 +851,10 @@ impl List {
     fn target_environment(&self) -> OptionalValueFlag {
         (&self.target_environment).into()
     }
+
+    fn should_summarise(&self) -> bool {
+        !self.installed && (self.summary || !self.all)
```
itowlson: This is correct, but I had to work through the logic to be sure of it!  The way I think of it is: summarise if `--summary` is set or _none_ of the flags are set (i.e. default). I.e. `summary || (!inst && !all)`

But yours is equivalent, just less self-explanatory to me!  And the tests document and verify the behaviour.

itowlson (approved):
Thanks for this.  A couple of comments but no blockers (except, of course, the whole breaking change thing that means we can't merge this until Spin 5).

src/commands/plugins.rs
mikkelhegn: Good suggestion :-)

mikkelhegn:
@itowlson I trust you've got this in terms of when it's appropriate to merge.

PR #3660, opened 2026-08-13
Spin deps CLI

fibonacci1729:
Implementation of [SIP 024](https://github.com/spinframework/spin/blob/main/docs/content/sips/024-spin-deps-cli-dx.md).


crates/capabilities/src/collect.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use crate::{CAPABILITY_SETS, InheritConfiguration};
+
+impl InheritConfiguration {
+    /// Collect iterates the imports of a Wasm component source and determines where each import
+    /// is part of one of the capabilities sets (i.e. AI_MODELS, ALLOWED_OUTBOUND_HOSTS, etc.).
+    /// For whichever set it is matched to, the lowercase name is appended to `capabilities` which
+    /// is returned as `InheritConfiguration::Some(...)`. If nothing matches `InheritConfiguration::None`
+    /// is returned.
+    pub fn collect(source: &[u8]) -> anyhow::Result<Option<Self>> {
```
itowlson: The naming here was not obvious to me (what am I collecting).  And I am confused about the return type.  `Result::Err` presumably means the bytes are garbage.  `Ok(Some(IC::Some))` means the component has imports which match Spin capabilities.  `Ok(Some(IC::None))` means the component has _no_ imports which match Spin capabilities.  But what does `Ok(None)` mean?  Looking at the code it seems to be what the comment says is represented by `Ok(Some(IC::None))`.  Might we have more cases than are necessary?

I am in two minds about representing the inferred capability sets as an `InheritConfiguration`.  On the one hand, the list of sets is structurally identical to an `InheritConfiguration`.  On the other hand, no configuration is being inherited at this point: this is collecting a list of capabilities to _ask_ about inheriting.

If it helps, the current iteration of the dependency isolation work/experiment represents custom capability configuration as a new variant of the `InheritConfiguration` enum.  I don't yet know if that's correct long term; but if it is, then the enum becomes less suitable as a return type for the inferred capabilities function.

crates/capabilities/src/collect.rs
```diff
@@ -0,0 +1,183 @@
...
+            "wasi:sockets/tcp@0.2.6",
+            "fermyon:spin/variables@2.0.0",
+        ]);
+        let result = InheritConfiguration::collect(&bytes).unwrap();
+        let InheritConfiguration::Some(caps) = result.unwrap() else {
+            panic!("expected Some capabilities");
+        };
+        assert_eq!(
+            caps,
+            vec!["ai_models", "allowed_outbound_hosts", "variables"]
```
itowlson: nit: order should not matter here: ideally we would compare as sets rather than vecs

crates/capabilities/src/collect.rs
```diff
@@ -0,0 +1,183 @@
...
+        let bytes = component.finish();
+        let result = InheritConfiguration::collect(&bytes).unwrap();
+        assert!(result.is_none());
+    }
+
+    #[test]
+    fn single_ai_models_import() {
+        let bytes = build_component(&["fermyon:spin/llm@2.0.0"]);
+        let result = InheritConfiguration::collect(&bytes).unwrap();
+        let InheritConfiguration::Some(caps) = result.unwrap() else {
```
itowlson: I feel like some of this would flow more naturally if we didn't distinguish the empty case at the type level.  If we just returned the set of capabilities then the assert would error on the empty case without needing the unwrap and the IC variant check.

crates/capabilities/src/collect.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    #[test]
+    fn duplicate_set_entries_are_deduped() {
+        let bytes =
+            build_component(&["wasi:http/outgoing-handler@0.2.6", "wasi:sockets/tcp@0.2.6"]);
+        let result = InheritConfiguration::collect(&bytes).unwrap();
+        let InheritConfiguration::Some(caps) = result.unwrap() else {
+            panic!("expected Some capabilities");
+        };
+        // Both map to allowed_outbound_hosts — should appear once.
```
itowlson, line 144: I apologise that I keep wanging on about the return type but if `collect` returned a set then this would be guaranteed at the type level.

crates/dependency-wit/src/lib.rs
```diff
@@ -9,6 +9,97 @@ use wit_parser::Span;
 
 const GENERATED_COMMENT: &str = "// This file is automatically generated by Spin\n// It is not intended for manual editing.\n\n";
 
+/// List the exports of a Wasm component as formatted dependency name strings.
+///
+/// Each export is formatted as `namespace:package/interface@version` for qualified
+/// interface exports, or as a plain kebab-case name for simple named exports.
```
itowlson: What is a simple named export?  Is that like where the world exports a function?  If so, could we say "for exported functions" or is it a gazillion times more complicated and nuanced than that?

(Of course it's a gazillion times more complicated and nuanced, Ivan. It always is.)

src/lib.rs
```diff
@@ -111,6 +112,8 @@ enum SpinApp {
     Registry(RegistryCommands),
     #[clap(alias = "b")]
     Build(BuildCommand),
+    #[clap(subcommand, alias = "dep")]
+    Deps(DepsCommands),
```
itowlson, line 116: I would vote for `Dependencies(DependenciesCommands)` as the variant and type names, with `dep` and `deps` aliases.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+    #[clap(
+        name = APP_MANIFEST_FILE_OPT,
+        short = 'f',
+        long = "from",
+        alias = "file",
+    )]
+    app_source: Option<PathBuf>,
+}
+
+/// Parsed representation of the user-supplied source string.
```
itowlson, line 58: Consider pulling this and `ResolvedSource` into a sub-module?  Just to slightly help tame this large file (as they are well-bounded chunks of functionality).

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        package: Option<String>,
+    },
+    Component {
+        id: String,
+    },
+}
+
+impl AddCommand {
+    pub async fn run(self) -> Result<()> {
+        // Locate and parse the manifest.
```
itowlson, line 115: Possibly the first thing to check here is "do we have an interactive terminal because this command is not going to work without one".

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        let app_root = parent_dir(&manifest_file)?;
+        let manifest = spin_manifest::manifest_from_file(&manifest_file)?;
+
+        // Resolve the source to Wasm bytes plus the metadata needed to record it.
+        let (wasm_bytes, dep_source) = self.resolve_source(&app_root, &manifest).await?;
+
+        // A component that both imports and exports wasi:http/handler is HTTP
+        // middleware, which is attached to a trigger rather than a component.
+        if spin_dependency_wit::is_http_middleware(&wasm_bytes)
+            .context("Failed to inspect the component's interfaces")?
```
itowlson, line 133: This feels like a core check of "is this a dependable thing" is being tucked away as `is_http_middleware`, which works of course but reading this I would be thinking "okay we are checking for middleware" not "okay we are checking for validity."

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        {
+            self.add_middleware(&manifest_file, &manifest, &wasm_bytes, dep_source)
+                .await
+        } else {
+            self.add_component_dependency(
+                &manifest_file,
+                &app_root,
+                &manifest,
+                &wasm_bytes,
```
itowlson, line 142: And having checked for validity we are passing on the raw bytes rather than a parsed component or an object that models the imports and exports?  (But perhaps I'm misunderstanding what's been validated?)

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+    async fn add_component_dependency(
+        &self,
+        manifest_file: &std::path::Path,
+        app_root: &std::path::Path,
+        manifest: &AppManifest,
+        wasm_bytes: &[u8],
+        dep_source: ResolvedSource,
+    ) -> Result<()> {
+        // Select the target component.
+        let component_id = self.resolve_component_id(manifest)?;
```
itowlson, line 159: I would maybe not use `resolve` for these functions, because that makes me think of "okay we are resolving a Wasm blob to an interface" rather than "we are going to badger the user to pick one."  The comments are, to me, much better names:


```suggestion
        let component_id = self.select_target_component(manifest)?;
```

`prompt` is also a good word although I guess the concern is that with a single component we would not actually prompt.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+        // Select the interface to import.
+        let selected = resolve_interface(wasm_bytes)?;
```
itowlson, line 162:
```suggestion
        let selected = select_interface(wasm_bytes)?;
```

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+        // Determine capability inheritance.
+        let (inherit_config, required_caps) =
+            resolve_inherit_configuration(manifest, &component_id, wasm_bytes)?;
```
itowlson, line 166: This is about the third time we have processed the raw bytes.  Could we introduce some intermediate representation to save doing this over and over?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+        // Build and write the dependency into the manifest.
+        let dep_name: DependencyName = selected.parse().with_context(|| {
+            format!("Failed to parse selected interface '{selected}' as a dependency name")
+        })?;
+        let dep_value = build_component_dependency(&dep_source, inherit_config.clone())?;
+        write_dependency_to_manifest(manifest_file, &component_id, &dep_name, &dep_value)?;
+
+        // Regenerate spin-dependencies.wit for the component.
+        let manifest = spin_manifest::manifest_from_file(manifest_file)?;
```
itowlson, line 176: Wait, do we not already have the manifest?  ... oh, is this because we edited the manifest and now need to reload it from disk?

itowlson, line 176: Consider moving this stanza out to a function (e.g. `regenerate_dependencies_wit`).  The mixing of nicely chunked operations with sudden "low level file management and string parsing" is a bit head spinning - it would be nice to maintain the linear flow of high level steps.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+        // Select the interface to import.
+        let selected = resolve_interface(wasm_bytes)?;
+
+        // Determine capability inheritance.
+        let (inherit_config, required_caps) =
+            resolve_inherit_configuration(manifest, &component_id, wasm_bytes)?;
+
+        // Build and write the dependency into the manifest.
+        let dep_name: DependencyName = selected.parse().with_context(|| {
```
itowlson, line 169: Consider breaking these few lines out to a well named helper function

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+    /// Attach the resolved source as HTTP middleware to a selected trigger.
+    async fn add_middleware(
+        &self,
+        manifest_file: &std::path::Path,
+        manifest: &AppManifest,
+        wasm_bytes: &[u8],
+        dep_source: ResolvedSource,
+    ) -> Result<()> {
+        println!("Detected HTTP middleware (imports and exports wasi:http/handler).");
```
itowlson, line 216: Is the WASI-level explanation of "detected middleware" helpful to the user?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        println!();
+
+        // Enumerate HTTP triggers that have a route; private (route-less)
+        // endpoints are not eligible for middleware.
+        let http_triggers = manifest.triggers.get("http").cloned().unwrap_or_default();
```
itowlson, line 221: This function would benefit from the same higher-level chunking as the non-middleware one.  Like the next bunch of lines could be a `select_trigger` function.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+                .items(&routes)
+                .interact()
+                .context("Failed to select route")?;
+            routed[sel]
+        };
+        let route = trigger_route(trigger)
+            .context("selected trigger has no route")?
+            .to_string();
+
+        // Existing middleware entries on the trigger (used for positioning).
```
itowlson, line 247: Again a `select_pipeline_position` function would help to make the flow obvious.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+            .map(|d| d.0.iter().map(trigger_dependency_label).collect())
+            .unwrap_or_default();
+
+        // Choose the pipeline position (append when the pipeline is empty).
+        let index = if existing_labels.is_empty() {
+            0
+        } else {
+            let mut items = vec!["At the end (closest to the application component)".to_string()];
+            for label in &existing_labels {
+                items.push(format!("Before {label}"));
```
itowlson, line 260: I guess the library doesn't have a "mobile selector that goes between list items" thing.  Given that, I wonder if we should be more explicit a la "Between X and Y" rather than just "Before Y".  But I guess that's too much clutter, too much repetitive information for the reader to parse.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+            .dependencies
+            .get("middleware")
+            .map(|d| d.0.iter().map(trigger_dependency_label).collect())
+            .unwrap_or_default();
+
+        // Choose the pipeline position (append when the pipeline is empty).
+        let index = if existing_labels.is_empty() {
+            0
+        } else {
+            let mut items = vec!["At the end (closest to the application component)".to_string()];
```
itowlson, line 258: Am I reading the ordering right?  Reading the code this looks like it will go e.g.

```
> At the end
  Before the first item
  Before the second item
  Before the third item
```

which visually implies an order which is not correct - "at the end" is actually after the third item - am I misunderstanding/misreading?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+                .interact()
+                .context("Failed to select pipeline position")?;
+            if sel == 0 {
+                existing_labels.len()
+            } else {
+                sel - 1
+            }
+        };
+
+        // Capability inheritance (from the component the trigger routes to).
```
itowlson, line 275: `prompt_inherit_config()` or something

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        let required_caps = collect_required_capabilities(wasm_bytes)?;
+        let inherit_config = if required_caps.is_empty() {
+            None
+        } else {
+            println!(
+                "This middleware requires the following capabilities: {}",
+                required_caps.join(", ")
+            );
+            let selections = dialoguer::MultiSelect::new()
+                .with_prompt("Select capabilities to inherit from the trigger's component")
```
itowlson, line 285: Man this really feels so weird.  It's like "This middleware needs these capabilities! Choose whether to grant them all or have it fail at runtime!"

I wonder if we should do something like: "This middleware needs these capabilities. If you proceed, it will be able to access the following: (list of caps, maybe even showing what it will inherit, not sure).  Continue / Cancel / I know what I'm doing, let me grant individual caps."  I am not sure.  Seeing this written out feels strange.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+                .items(&required_caps)
+                .interact()
```
itowlson, line 287: I haven't gone back and checked others but these should usually be `interact_opt` so that the user can go "what the heck have I got myself into" and Esc out

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+            component_id.as_deref(),
+            &required_caps,
+            inherit_config.as_ref(),
+        );
+
+        Ok(())
+    }
+
+    /// Determine which component to add the dependency to (interactive).
+    fn resolve_component_id(&self, manifest: &AppManifest) -> Result<String> {
```
itowlson, line 323: This needs a way to represent cancellation.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+                } else {
+                    std::env::current_dir()
+                        .context("Failed to get current directory")?
+                        .join(path)
+                };
+                if !resolved.exists() {
+                    bail!("Dependency not found: {}", resolved.display());
+                }
+                // Store the path relative to app_root for the manifest.
+                let rel_path = resolved
```
itowlson, line 368: Consider https://docs.rs/pathdiff/latest/pathdiff/fn.diff_paths.html instead of `strip_prefix`

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+                    actual_digest == digest,
+                    "invalid content digest; expected {digest}, downloaded {actual_digest}"
+                );
+
+                let dest = cache.wasm_path(&digest);
+                tokio::fs::write(dest, &bytes).await?;
+
+                Ok((
+                    bytes.to_vec(),
+                    ResolvedSource::Http {
```
itowlson, line 429: Oh this is the information we need to construct the manifest entry!

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+                let temp_dep = ComponentDependency::Package {
+                    version: version_str.clone(),
+                    registry: self.registry.clone(),
+                    package: Some(package.package.to_string()),
+                    export: None,
+                    inherit_configuration: None,
+                };
+
+                let (wasm_path, _export) = loader
+                    .load_dependency_content(&dep_name, &temp_dep)
```
itowlson, line 454: Hmm, I wonder if it's worth cracking WasmLoader open to accept the fields directly instead of forcing you to construct a fake dependency.  It looks like `load_registry_source` exists, just not public.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+    if exports.is_empty() {
+        bail!("The Wasm component has no exports to use as a dependency");
+    }
+
+    if exports.len() == 1 {
+        return Ok(exports[0].clone());
+    }
+
+    // Group exports by package, preserving first-seen order. Plain-named exports
```
itowlson, line 510: Why is first-seen order significant here?  Do we feel it's preferable to alphabetical order?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+    let mut groups: Vec<(Option<String>, Vec<String>)> = Vec::new();
+    for export in &exports {
+        let pkg_key = export.parse::<DependencyPackageName>().ok().map(|p| {
+            let mut key = p.package.to_string();
+            if let Some(v) = &p.version {
+                key.push_str(&format!("@{v}"));
+            }
+            key
+        });
+        match groups.iter_mut().find(|(k, _)| *k == pkg_key) {
```
itowlson, line 521: Is this preferred to using a map because of the first-seen order thing?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+}
+
+/// Collect the capability sets the dependency requires, inferred from its imports.
+fn collect_required_capabilities(wasm_bytes: &[u8]) -> Result<Vec<String>> {
+    Ok(
+        match spin_capabilities::InheritConfiguration::collect(wasm_bytes)
+            .context("Failed to collect capability requirements from the dependency")?
+        {
+            Some(spin_capabilities::InheritConfiguration::Some(caps)) => caps,
+            _ => vec![],
```
itowlson, line 566: I'm again questioning the value of the Option in this type, sorry

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+/// Determine the `inherit_configuration` value for the dependency (interactive).
+///
+/// Returns the value to record (never `All(true)` — always an explicit list) and
+/// the full set of capabilities the dependency requires (for post-add guidance).
+fn resolve_inherit_configuration(
+    manifest: &AppManifest,
+    component_id: &str,
+    wasm_bytes: &[u8],
+) -> Result<(Option<InheritConfiguration>, Vec<String>)> {
```
itowlson, line 579: It's not clear to me what this return type means.  What does it mean for the option to be None?  If the result is `Some(IC::Some)`, why do we need a caps list in the tuple, doesn't `IC::Some` already have a caps list?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+    }
+
+    // Edge case: if the component already sets the blanket
+    // `dependencies_inherit_configuration`, that covers all dependencies, so we
+    // neither prompt nor write a per-dependency `inherit_configuration`.
+    if let Ok(kebab) = TryInto::<KebabId>::try_into(component_id.to_string())
+        && let Some(component) = manifest.components.get(&kebab)
+        && component.dependencies_inherit_configuration.is_some()
+    {
+        return Ok((None, required));
```
itowlson, line 592: Oh, is this what's going on, there's a scenario where we don't want to write an inherit but we need to know the list of caps anyway?  Maybe that case should be explained in the function comments.  (To this naive reader it is not clear why we need to caps list if we are not going to write an inherit.)

(But nice catch on the blanket inheritance case - I completely missed that one.)

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+    // Always record the explicit list of selected capabilities. We deliberately
+    // never emit `inherit_configuration = true`, even when every listed
+    // capability is selected, so that a future version of the dependency that
+    // imports a new capability does not silently inherit it.
+    let selected: Vec<String> = selections
+        .into_iter()
+        .map(|i| required[i].clone())
+        .collect();
+    Ok((Some(InheritConfiguration::Some(selected)), required))
```
itowlson, line 618: Oh, right, this is this "post-add guidance" thing, we are separately recording "what the user has chosen to inherit" and "what the user has chosen not to inherit" so we can chastise them.  Okay, I get it now.  Maybe a struct with field names would help, or perhaps expand the comment a bit?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        .iter()
+        .filter(|c| !inherited.contains(c))
+        .cloned()
+        .collect();
+
+    println!();
+    if !inherited.is_empty() {
+        println!("NOTE: This dependency inherits: {}.", inherited.join(", "));
+        println!(
+            "Ensure component '{component_id}' declares these capabilities so the dependency can use them."
```
itowlson, line 646: We can check this can't we?  Is there a point in badgering the user if the caps already exist?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        );
+    }
+    if !declined.is_empty() {
+        println!(
+            "NOTE: The dependency also uses {} which was not inherited; it will be denied these at runtime.",
```
itowlson, line 651: "was" is going to read oddly if multiple capabilities are declined.

I am not sure if "denied" is too internal lingo: perhaps "it will fail if it tries to use these"?  Not sure

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+            declined.join(", ")
+        );
+    }
+}
+
+/// Build the `ComponentDependency` value from the resolved source.
+fn build_component_dependency(
```
itowlson, line 658: This sounds like you are building the dependency.  `dependency_manifest_entry()` or something like that perhaps?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+            registry: registry.clone(),
+            package: package.clone(),
+            export: None,
+            inherit_configuration: inherit_config,
+        }),
+        ResolvedSource::Component { id } => {
+            let component: KebabId = id
+                .clone()
+                .try_into()
+                .map_err(|e| anyhow!("'{id}' is not a valid component id: {e}"))?;
```
itowlson, line 689: Haven't we already checked this?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        .get_mut(component_id)
+        .and_then(|c| c.as_table_like_mut())
+        .with_context(|| format!("Component '{component_id}' not found in manifest"))?;
+
+    // Ensure [component.<id>.dependencies] exists.
+    if component.get("dependencies").is_none() {
+        component.insert("dependencies", Item::Table(Table::new()));
+    }
+    let deps_table = component
+        .get_mut("dependencies")
```
itowlson, line 730: They say that if you say the `"dependencies"` literal three times, the bogeyman jumps out of a ~review~ mirror and tells you to create a `const` for it.

itowlson, line 730: We either just got or just inserted this element.  Can we not use the result of `get/insert` instead of re-looking it up?

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        .and_then(|d| d.as_table_like_mut())
+        .context("Failed to access dependencies table")?;
+
+    let dep_key = dep_name.to_string();
+    if deps_table.contains_key(&dep_key) {
```
itowlson, line 735: This feels like a thing we could check earlier, before tormenting them with decisions about capabilities that turn out to be completely nugatory.  (We could even _not offer_ keys that already exist.)

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+fn serialize_component_dependency(dep: &ComponentDependency) -> Result<toml_edit::Item> {
+    match dep {
+        ComponentDependency::Version(version) => Ok(toml_edit::value(version.as_str())),
+        ComponentDependency::Local {
+            path,
+            export,
+            inherit_configuration,
+        } => {
+            let mut table = toml_edit::InlineTable::new();
+            table.insert(
```
itowlson, line 761: Disappointing that we can't use `toml` serialisation and slurp that into a `toml_edit::Item`.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+/// A short label for the source being added (for the confirmation message).
+fn source_label(source: &ResolvedSource) -> String {
+    match source {
+        ResolvedSource::Local { path } => path.display().to_string(),
+        ResolvedSource::Http { url, .. } => url.clone(),
+        ResolvedSource::Registry {
+            package, version, ..
+        } => match package {
+            Some(p) => format!("{p}@{version}"),
+            None => version.clone(),
```
itowlson, line 868: Not sure where this is used but the comment implies it's human-facing - if so, I'm somewhat worried that a bare version may not be very informative unless it has the package name nearby.  But like I say I am missing context on this

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+        },
+        ResolvedSource::Component { id } => id.clone(),
+    }
+}
+
+/// Serialize the resolved source as a middleware entry inline table.
+fn serialize_trigger_dependency(
```
itowlson, line 875: This seems very reminiscent of `serialise_component_dependency`.  Even if we can't consolidate them, I'd consider locating them next to each other in the file (or even in a sub-module) so that someone maintaining them is more likely to see them both at once.

src/commands/deps.rs
```diff
@@ -0,0 +1,1077 @@
...
+    let idx = index.min(middleware.len());
+    middleware.insert(idx, entry);
+
+    std::fs::write(manifest_file, doc.to_string()).context("Failed to write manifest file")?;
+
+    Ok(())
+}
+
+/// Print guidance about the capabilities the middleware needs.
+fn print_middleware_capability_guidance(
```
itowlson, line 967: This seems very similar to the other "print capability guidance" function except for the word "middleware" in the messages.  Could we have a single function that takes a parameter?  Do we even need to say "middleware" instead of "dependency" given that middleware is enough of a dependency to live under the dependencies command?

itowlson (dismissed):
This looks good!  I struggled with a few things I'm afraid, and I have some naming nits.  My main feeling is that I would really like to see the main flows as something like:

```
fn add_component_dependency() {
  let component_id = select_component();         // }
  let selected_interface = select_interface();   // } as you already have it
  let inheritance = select_inheritance();        // }
  write_dependency_to_manifest();
  regenerate_dependencies_wit();
  print_report();
}
```

instead if inlining the last half, which for me meant I lost track of what is really a very simple linear flow.

But all that aside the bones seem really solid and it is mostly stylistic confusion on my part.  Great to see this under sail!

crates/capabilities/src/collect.rs
fibonacci1729: Made this a free standing function that returns a set.

crates/dependency-wit/src/lib.rs
fibonacci1729: Clarified to mention exported function and inline interfaces.

src/commands/deps.rs
fibonacci1729: First-seen matches WIT declaration order, but happy to sort alphabetically if preferred.

itowlson: I'm not sure why WIT declaration order is helpful to the person using the component.  If it's correlated to importance/relevance then that would make sense.  Otherwise I worry it will feel random.  (Although there will typically be so few interfaces that maybe it's moot.)

crates/capabilities/src/collect.rs
```diff
@@ -0,0 +1,163 @@
...
+use crate::CAPABILITY_SETS;
+
+/// Infer the Spin capability sets a component requires, by inspecting its
+/// top-level (component) imports.
+///
+/// Each import is matched against the known capability sets (`ai_models`,
+/// `allowed_outbound_hosts`, etc.); every set that matches contributes its name
+/// to the result. The returned set is deduplicated and sorted. An empty set
+/// means the component imports nothing that maps to a Spin capability.
+pub fn required_capabilities(source: &[u8]) -> anyhow::Result<BTreeSet<String>> {
```
itowlson: Could this be `&'static str` instead of `String`?

crates/capabilities/src/collect.rs
```diff
@@ -0,0 +1,163 @@
...
+        let mut imports = ComponentImportSection::new();
+        for name in import_names {
+            imports.import(name, ComponentTypeRef::Instance(0));
+        }
+        component.section(&imports);
+
+        component.finish()
+    }
+
+    fn caps(names: &[&str]) -> BTreeSet<String> {
```
itowlson: The function name implies either capabilities or capitalisation but the body says different?

crates/capabilities/src/collect.rs
```diff
@@ -0,0 +1,163 @@
...
+        names.iter().map(|s| s.to_string()).collect()
```
itowlson: Again, not quite sure why ownership is needed?

...oh, is all the owning because of Rust refusing to compare `&str` with `&&str` or some such silliness?

src/commands/deps/source.rs
```diff
@@ -0,0 +1,338 @@
...
+            ParsedSource::Http(url) => resolve_http(url, digest).await,
+            ParsedSource::Registry { package } => {
+                resolve_registry(package, registry, app_root).await
+            }
+            ParsedSource::Component(id) => resolve_component(id, app_root, manifest).await,
+        }
+    }
+}
+
+async fn resolve_local(path: &Path, app_root: &Path) -> Result<(Vec<u8>, ResolvedSource)> {
```
itowlson, line 114: I feel like there's a lot of overlap between this set of functions and the WasmLoader methods.  E.g. it somewhat bothers me to see us rewriting the HTTP downloader.  What prevents us unifying these flows with existing ones?

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+    /// package reference.
+    ///
+    /// Examples:
+    ///   ./path/to/component.wasm
+    ///   https://example.com/component.wasm
+    ///   my:package@1.0.0
+    source: ParsedSource,
+
+    /// SHA-256 digest used to verify HTTP downloads. Required for HTTP sources,
+    /// ignored otherwise.
```
itowlson, line 48: I kinda feel like this and registry should be error on wrong source type, rather than ignored.  Open to being convinced otherwise!

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+        alias = "file",
+    )]
+    app_source: Option<PathBuf>,
+}
+
+impl AddCommand {
+    pub async fn run(self) -> Result<()> {
+        // These flows are interactive, so fail fast if there's no terminal to prompt on.
+        ensure_interactive()?;
+
```
itowlson, line 74: We should probably verify that HTTP sources have digests, and registry sources have registries, before trying to load the manifest.  I think "your command is bad and you should feel bad" slightly trumps "and your manifest is bad too"...?

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+    let dep_value = dep_source.to_component_dependency(inheritance.to_write.clone());
+    write_dependency_to_manifest(manifest_file, component_id, &dep_name, &dep_value)?;
+    regenerate_dependencies_wit(manifest_file, app_root, component_id).await?;
+
+    println!("Added {selected} to {target}");
+    println!("Run `spin build` to generate language bindings for the new dependency.");
+    print_capability_guidance(&target, Some(component), &inheritance);
+
+    Ok(())
```
itowlson, line 157: I really like the flow of this function.  The clear top-to-bottom-with-cancellation format makes it super obvious to the reader what is happening, and there is basically zero clutter to distract from that clarity.

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+    );
+    print_capability_guidance(&target, component.map(|(_, c)| c), &inheritance);
+
+    Ok(())
+}
+
+/// The add flows are interactive; make sure there's a terminal to prompt on.
+fn ensure_interactive() -> Result<()> {
+    use std::io::IsTerminal;
+    if std::io::stdin().is_terminal() && std::io::stdout().is_terminal() {
```
itowlson: I believe we usually check `stderr` rather than `stdout`, because that is what `dialoguer` interacts on.  I don't think we care about `stdin`.

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+        Ok(())
+    } else {
+        bail!("`spin dependencies add` is interactive and requires a terminal.");
+    }
+}
+
+/// Report that the user cancelled, leaving the manifest untouched.
+fn cancelled() -> Result<()> {
+    println!("No changes were made.");
```
itowlson: Nice little detail!

Nit: I'm wondering if this (and maybe some others?) should be `eprintln` - not sure.

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+        .context("Failed to select capabilities")?
+    else {
+        return Ok(None);
+    };
+
+    let inherited: Vec<String> = match choice {
+        0 => required.clone(),
+        1 => vec![],
+        _ => {
+            let Some(selections) = dialoguer::MultiSelect::new()
```
itowlson: Consider breaking this block out into a named function, so that it's clear to the reader that this is the "choose individually" case.

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+        .required
+        .iter()
+        .map(String::as_str)
+        .filter(|c| !inheritance.inherited.iter().any(|i| i == c))
+        .collect();
+
+    if !undeclared.is_empty() {
+        println!();
+        println!(
+            "NOTE: {target} does not yet declare: {}. Add these so the dependency can use them.",
```
itowlson: I am not sure what "declared" means in this context.

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+    let declined: Vec<&str> = inheritance
+        .required
+        .iter()
+        .map(String::as_str)
+        .filter(|c| !inheritance.inherited.iter().any(|i| i == c))
+        .collect();
+
+    if !undeclared.is_empty() {
+        println!();
+        println!(
```
itowlson: Should we use something like `terminal::warn!` or `terminal::einfo!` here?

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+            "NOTE: {target} does not yet declare: {}. Add these so the dependency can use them.",
+            undeclared.join(", ")
+        );
+    }
+    if !declined.is_empty() {
+        println!();
+        println!(
+            "NOTE: Not inherited: {}. The dependency's calls to these capabilities will fail at runtime.",
```
itowlson: "Not inherited" is quite terse - can we make it more explanatory without it turning into _War and Peace_?

Also, should we reword "the dependency's calls" - I am thinking of ye olde Python case where the dep may never call the import but may have it anyway from the SDK - "if the dependency calls these capabilities" perhaps?  abuse of language a bit though, I know

src/commands/deps.rs
```diff
@@ -0,0 +1,834 @@
...
+        .as_array_mut()
+        .context("Failed to access the middleware array")?;
+
+    let idx = index.min(middleware.len());
+    middleware.insert(idx, entry);
+
+    std::fs::write(manifest_file, doc.to_string()).context("Failed to write manifest file")?;
+
+    Ok(())
+}
```
itowlson, line 849: I really enjoyed reading this whole file.  Thank you.

itowlson (approved):
Couple of wording and erroring suggestions which we can iterate on over time, no blockers.

I admit I found all the Wasm parser stuff a bit hairy but the main UI stuff was a model of clarity, and I really appreciate the work that went into structuring it so cleanly - thank you!

crates/capabilities/src/collect.rs
fibonacci1729: This went away with the move to `&'static str`

src/commands/deps/source.rs
fibonacci1729: Definitely interested in unifying these further in a follow up -- the tricky bit is that these resolve relative to CWD not an app root.

PR #3663, opened 2026-08-14
Allow embedders flexibility in how they register their metrics

rylev:
Exposes `metrics_provider` so that the embedders can modify and and ultimately register through `set_meter_provider` in the way they see fit.

### DIFF
--- a/crates/telemetry/src/metrics.rs
+++ b/crates/telemetry/src/metrics.rs
@@ -1,5 +1,5 @@
 use anyhow::{Result, bail};
-use opentelemetry_otlp::WithHttpConfig;
+use opentelemetry_otlp::{WithExportConfig, WithHttpConfig};
 use opentelemetry_sdk::{
     Resource,
     metrics::{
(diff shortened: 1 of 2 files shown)


crates/telemetry/src/metrics.rs
```diff
@@ -32,42 +32,68 @@ pub struct HistogramBuckets {
...
+///
 /// The caller is responsible for registering the returned provider as the global one (e.g. via
 /// [`opentelemetry::global::set_meter_provider`]). Instruments created by the macros in this
 /// module (e.g. [`counter`](crate::counter)) bind to whatever meter
 /// provider is global *at the time they're first used*, and never rebind afterwards.
-pub(crate) fn metrics_provider(
-    spin_version: String,
+///
+/// Exposed publicly so embedders that manage their own [tracing::Subscriber] (and so can't call
+/// [`crate::init`], which also installs one) can still set up Spin's metrics: build a provider
```
itowlson, line 54: This made me wonder if `crate::init` should be offering more flexibility to consumers (with the CLI defaults hived off to an `init_default` kind of function).  But I'm not familiar with the telemetry subsystem so I'm fine with this less invasive change!  But I'd like @calebschoepp to look it over too.

itowlson (commented):
I had one question but basically this looks fine as far as I know, but I would really like it if Caleb could take a look.  If he doesn't have bandwidth then I'm happy to merge it.

itowlson:
@calebschoepp bump

calebschoepp (approved):
This all LGTM and I don't think we need to pursue a deeper change.

PR #3668, opened 2026-08-18
Accept boolish values for env-settable bool flags

ChihweiLHBird:
Supporting broader boolish values, like `1`, `0`, `yes` by adding [BoolishValueParser](https://gitlab-com.gitlab.io/support/toolbox/gitlabsar/clap/builder/struct.BoolishValueParser.html).

Note that some input arguments are impacted by a conflicting issue caused by a clap issue https://github.com/clap-rs/clap/issues/5591 which is a pre-existing issue before this PR. An example is even if `DISABLE_WASMTIME_CACHE` is set to false, it still conflicts to `WASMTIME_CACHE_FILE`. Clap check whether it's set rather than whether it's set to true.

### DIFF
--- a/crates/trigger/src/cli.rs
+++ b/crates/trigger/src/cli.rs
@@ -73,6 +73,7 @@ pub struct FactorsTriggerCommand<T: Trigger<B::Factors>, B: RuntimeFactorsBuilde
         long = "disable-cache",
         env = DISABLE_WASMTIME_CACHE,
         conflicts_with = WASMTIME_CACHE_FILE,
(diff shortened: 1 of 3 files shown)


itowlson:
To clarify, does this PR cause us to run into the Clap `false` conflicts issue, or do we already have that issue?

ChihweiLHBird:
> To clarify, does this PR cause us to run into the Clap `false` conflicts issue, or do we already have that issue?
> 

We already have it before this PR. I discovered the issue during working on this PR.

itowlson (approved):

PR #3676, opened 2026-08-20
Avoid reinstalling env templates if Git source has not changed

### LINKED ISSUE #3672: Delay when creating an app using environment templates
When you run `spin new -E`, it always fetches latest templates, which incurs a delay.

We would like it to not incur a delay.

The challenge with this is: the current behaviour ensures that the app you create is lined up with the environment definition.  If we use old (cached) templates, it may not line up with the environment definition.  So how do we ensure that templates are current, and yet not incur a delay when they are?

itowlson:
Fixes #3672.

The adopted fix is to record against each template the repo and tag that it was installed from, assume that tags and branches are immutable, and update only if the actual tag or branch listed in the env file has changed from the one used for installation.

But: two considerations:

1. This will result in missed updates if the env owner moves a tag or branch in their templates repo. For this to work, env owners have to consider refs immutable, and if they want to change their templates, they have to update their env with a new tag.  I am not sure if this is a safe assumption.  We could have a Plan B to re-fetch periodically, but either this kick in too often (and Thorsten would remain frustrated), or it would be too infrequent to be useful.  Feedback welcome.

2. This does not address the delay to re-snapshot the environment, which happens if the environment is unversioned and our snapshot is more than 1 hour old (basically if no unversioned env has been used in the last hour).  This is trickier because we can have absolutely _no_ way to tell if the upstream has changed, without a network round-trip.  We could increase the recency window to reduce the incidence of these delays, but again that just makes it more and more likely that a user will hit the window.  Although we do have the `spin targets update` command to mitigate that, so maybe it would be okay.

So: feedback and discussion first, so we don't do the wrong thing; but at the same time, it would be nice to get any fix we are happy with into 4.1.  So no mixed messages then.

(cc @ThorstenHans who flagged the original issue)

### DIFF
--- a/crates/templates/src/manager.rs
+++ b/crates/templates/src/manager.rs
@@ -25,20 +25,32 @@ pub trait ProgressReporter {
 #[derive(Debug)]
 pub struct InstallOptions {
     exists_behaviour: ExistsBehaviour,
+    skip_if_all_match: bool,
 }
 
 impl InstallOptions {
     /// Sets the option to update existing templates. If `update` is true,
     /// existing templates are updated. If false, existing templates are
     /// skipped.
-    pub fn update(self, update: bool) -> Self {
+    pub fn update(mut self, update: bool) -> Self {
         let exists_behaviour = if update {
             ExistsBehaviour::Update
         } else {
             ExistsBehaviour::Skip
         };
 
-        Self { exists_behaviour }
+        self.exists_behaviour = exists_behaviour;
+        self
+    }
+
+    /// If set, the installer will skip if the manager has templates,
+    /// and all templates were installed from the install source.
+    /// This allows a consumer to avoid downloading from a
+    /// remote source, at the expense of missing an update if the
+    /// source is mutable (such as a Git tag or branch).
+    pub fn skip_if_all_match(mut self, skip_if_all_match: bool) -> Self {
+        self.skip_if_all_match = skip_if_all_match;
+        self
     }
 }
 
@@ -46,6 +58,7 @@ impl Default for InstallOptions {
     fn default() -> Self {
         Self {
             exists_behaviour: ExistsBehaviour::Skip,
+            skip_if_all_match: false,
         }
     }
 }
@@ -133,6 +146,18 @@ impl TemplateManager {
         options: &InstallOptions,
         reporter: &impl ProgressReporter,
     ) -> anyhow::Result<InstallationResults> {
+        if options.skip_if_all_match && self.all_match(source).await {
+            let existing = self.list().await.map(|lr| lr.templates).unwrap_or_default(); // don't fail if we can't list
+            return Ok(InstallationResults {
+                installed: Default::default(),
+                skipped: existing
+                    .into_iter()
+                    .map(|t| (t.id().to_string(), SkippedReason::AlreadyExists))
+                    .collect(),
+                removed: Default::default(),
+            });
+        }
+
         if source.requires_copy() {
             reporter.report("Copying remote template source");
         }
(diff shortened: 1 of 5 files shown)


crates/templates/src/template.rs
```diff
@@ -254,11 +254,23 @@ impl Template {
             .is_some_and(|r| r == source_repo.as_str())
     }
 
+    /// Determines if the template was installed from the expected repo and
+    /// branch. This does not prove that the template is up to date, as the
+    /// tag or branch may have moved in the interim.
+    pub fn is_installed_from(&self, expected_repo: &str, expected_tag: Option<&str>) -> bool {
```
fibonacci1729:
```suggestion
    pub fn is_installed_from_git_ref(&self, expected_repo: &str, expected_tag: Option<&str>) -> bool {
```

Would help disambiguate from `is_from_source_repo` above.

crates/templates/src/manager.rs
```diff
@@ -317,6 +342,23 @@ impl TemplateManager {
...
+
+    async fn all_match(&self, source: &TemplateSource) -> bool {
+        let existing = self.list().await.map(|lr| lr.templates).unwrap_or_default();
+        if existing.is_empty() {
+            // We don't expect a template source to be empty. If the manager is empty, count it as no match.
+            return false;
+        }
+
+        let (expected_repo, expected_tag) = match source {
+            TemplateSource::Git(g) => (g.repo(), g.branch()),
```
fibonacci1729, line 354: I think maybe we should bail out here when the source branch is `None`. Untagged envs will keep today's behavior. This effectively makes the "refs are immutable" an opt-in contract the env author signs by pinning rather than an assumption imposed on people who deliberately didn't pin. The only downside here is that untagged envs get no speedup at all.

fibonacci1729:
AFAICT `spin targets update` only refreshes the environment catalogue. To add an escape hatch when a ref moves (for whatever reason) I think we should either:
1. Refresh the installed env templates stores during `target update`
2. Add a `--refresh-templates/--force` flag to `spin new -E`
3. Or a TTL on the skip to bound staleness

itowlson:
Ooh I like `spin targets update` refreshing the templates.

itowlson:
Okay `spin targets update` updating the templates is going to need a wee bit more thought.  `update` updates the entire catalogue which contains _all_ envs, so we need to figure out which envs the user is actually using and update only those templates, which is new - we don't currently have a concept of "the user is using an environment," it is all on-demand.  I think we can get away with "do we have cached templates from this env" but I would like to defer it to another PR so I can think about it separately.

itowlson:
Updated with your suggestions @fibonacci1729 - please take a look - thanks!

ThorstenHans:
What about retrofitting this into `spin templates upgrade`? From a user perspective, a template contributed by a target-environment is no different from a "regular" Spin template.

- If we know `http-yolo` is a template that's provided by the `yolo` environment, I would expect `spin templates list` to visualize this as well
- When running `spin templates upgrade` users should be able to select an environment and trigger upgrade of all templates contributed by the environment

Handing off the template management to those commands grouped under `spin templates` would allow `spin new` to remain the way it was prior to Spin 4.x. and have the lifecycle for templates decoupled from the actual scaffolding.

Installing templates defined by an environment should IMO be possible through two routes:

- `spin templates install -E yolo`: To install all templates defined by an environment called yolo
- `spin targets use yolo`: This would set the `yolo` environment as my personal default (e.g, `SPIN_DEFAULT_TARGET_ENVIRONMENT=yolo`) and install templates for that environment after user confirmation.

ℹ️: 2nd approach is just an idea that came into my mind

fibonacci1729 (approved):

PR #3679, opened 2026-08-21
Allow an environment to constrain component permissions

### LINKED ISSUE #3677: Express configuration constraints in environment definition
In addition to limits on which APIs are available, some environments constrain how APIs can be configured. For example, an environment may support only a single `"default"` key-value store.

While environments can catch this stuff with maximum fidelity at deployment time, it would be nice to find out about it at dev time, _before_ you built too much code that relied on separating your KV stores.

We'd need feedback from env owners on what to constrain but here's my sighting shot on what items could usefully support configuration constraints:

| Configuration item | Constrainable |
|------------------|-----------------|
| Allowed outbound hosts | No (they can exclude APIs like HTTP, Redis, Postgres, etc. but that's at the interface level; I can't see constraining AOH patterns) |
| Files  | No (they can exclude the filesystem API but not individual patterns, that just seems implausible) |
| Environment variables | No (they can exclude the env var API but individual variables why) |
| KV stores | Yes |
| SQLite databases | Yes |
| AI models | Yes |
| Variables | No (can exclude the API but again I can't see why an env could ban individual vars) |

itowlson:
Fixes #3677.

### DIFF
--- a/crates/environments/src/environment.rs
+++ b/crates/environments/src/environment.rs
@@ -10,8 +10,8 @@ mod env_loader;
 mod lockfile;
 
 pub use catalogue::Catalogue;
-pub use definition::EnvironmentDefinition;
 use definition::WorldName;
+pub use definition::{ConfigurationConstraints, EnvironmentDefinition};
 pub use env_loader::load_environment_def;
 
 use crate::Targets;
@@ -27,6 +27,7 @@ pub struct TargetEnvironment {
     trigger_capabilities: HashMap<TriggerType, Vec<String>>,
     unknown_trigger: UnknownTrigger,
     unknown_capabilities: Vec<String>,
+    configuration_constraints: ConfigurationConstraints,
 }
 
 pub(crate) struct RealisedTargets {
@@ -95,6 +96,10 @@ impl TargetEnvironment {
         self.trigger_capabilities
             .get(trigger_type)
             .unwrap_or(&self.unknown_capabilities)
+    }
+
+    pub fn configuration_constraints(&self) -> &ConfigurationConstraints {
+        &self.configuration_constraints
     }
 }
 
@@ -278,6 +283,7 @@ mod test {
             trigger_capabilities: Default::default(),
             unknown_trigger: UnknownTrigger::Deny,
             unknown_capabilities: Default::default(),
+            configuration_constraints: Default::default(),
         }
     }
 
@@ -315,7 +321,12 @@ mod test {
             trigger_capabilities: Default::default(),
             unknown_trigger: UnknownTrigger::Allow(candidate_worlds),
             unknown_capabilities: Default::default(),
+            configuration_constraints: Default::default(),
         })
+    }
+
+    fn fake_component() -> spin_manifest::schema::v2::Component {
+        toml::from_str(r#"source = "dummy.wasm""#).unwrap()
     }
 
     #[tokio::test]
@@ -332,7 +343,8 @@ mod test {
         assert!(env.supports_trigger_type(&"s".to_owned()));
         assert!(!env.supports_trigger_type(&"t".to_owned()));
 
-        let component = crate::ComponentToValidate::new("scomp", "scomp.wasm", wasm, vec![]);
+        let component =
+            crate::ComponentToValidate::new("scomp", "scomp.wasm", wasm, vec![], fake_component());
         let errs =
             crate::validate_component_against_environments(&[env], &"s".to_owned(), &component)
                 .await;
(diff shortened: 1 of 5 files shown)


crates/environments/src/environment/definition.rs
```diff
@@ -164,6 +170,13 @@ impl std::fmt::Display for WorldName {
     }
 }
 
+#[derive(Clone, Debug, Default, serde::Deserialize)]
```
fibonacci1729, line 174: I think we should add a `#[serde(deny_unknown_fields)]` here.

crates/environments/src/lib.rs
```diff
@@ -271,6 +274,46 @@ fn validate_host_reqs(
...
+    configuration_constraints: &environment::ConfigurationConstraints,
+    component: &ComponentToValidate,
+    errors: &mut Vec<anyhow::Error>,
+) {
+    let mut validate_only_permitted = |descriptor,
+                                       permitted_values: &Option<Vec<String>>,
+                                       required_values| {
+        if let Some(permitted_values) = permitted_values.as_ref() {
+            for required_value in required_values {
+                if !permitted_values.contains(required_value) {
```
fibonacci1729, line 287: Should we normalize the configurations? IIRC apps can specify `my-store` or `my_store` to mean the same thing since labels are deserizlied via `kebab_or_snake_case`.

crates/environments/src/environment.rs
```diff
@@ -687,6 +714,156 @@ mod test {
...
+            &[Arc::new(env)],
+            &"s".to_owned(),
+            &component,
+        )
+        .await;
+        assert!(errs.is_empty());
+    }
+
+    #[tokio::test]
+    async fn empty_configuration_allows_all() {
```
fibonacci1729:
```suggestion
    async fn empty_configuration_allows_none() {
```

crates/environments/src/environment.rs
```diff
@@ -687,6 +714,156 @@ mod test {
...
+        assert!(
+            err.contains("Component cscomp can't run in environment test"),
+            "unexpected error {err}"
+        );
+        assert!(err.contains("key-value store"), "unexpected error {err}");
+        assert!(err.contains("the-kv-i-need"), "unexpected error {err}");
+    }
+
+    #[tokio::test]
+    async fn supported_kv_store_does_not_invalidates_component() {
```
fibonacci1729:
```suggestion
    async fn supported_kv_store_does_not_invalidate_component() {
```

crates/environments/src/lib.rs
fibonacci1729: This comment is now out of date with these changes, specifically "there may be up to one error ..."

crates/environments/src/lib.rs
itowlson: Oh man

crates/environments/src/environment/definition.rs
```diff
@@ -164,6 +170,13 @@ impl std::fmt::Display for WorldName {
...
+pub struct ConfigurationConstraints {
+    pub key_value_stores: Option<Vec<String>>,
```
fibonacci1729, line 178: Document these pub items.

crates/environments/src/loader.rs
```diff
@@ -148,27 +178,32 @@ impl ApplicationToValidate {
...
         };
 
         Ok(Some(ComponentToValidate {
-            id: component.id,
-            source_description: source_description(component.source),
+            id: component_source.id,
+            source_description: source_description(component_source.source),
             wasm,
             host_requirements,
+            manifest: component.clone(),
```
fibonacci1729, line 206: I don't think you need a clone here since everything is already lifetime bound by `'a`, no?

itowlson: You are technically correct but it makes testing a monstrous chore fighting the borrow checker.

fibonacci1729 (approved):

crates/environments/src/lib.rs
itowlson: Are you sure about this?  Looking at `kebab_or_snake_case` it appears to be _validating_ rather than normalising.

fibonacci1729: You're right it validates. But nevertheless, what should the behavior be if an env declares `my-store` but an app specifies `my_store` (both valid spellings of the same intent).

itowlson: Are you sure those are interchangeable?

fibonacci1729: I am operating off a vague memory so no. 🤣

PR #3695, opened 2026-08-27
Update manifest JSON schema with HTTP static_response

### LINKED ISSUE #3693: Recommendation for HTTP trigger `static_response` or maybe a manifest schema drift?
According to the documentation, users can specify `static_response` as part of the HTTP trigger definition:

```toml
[[trigger.http]]
route = "/foo/bar"
static_response = { status_code = 200, body = "Ok" }
```

See docs at https://spinframework.dev/v4/manifest-reference#additional-fields-for-triggerhttp-tables

However, the latest schema does not contain this property as optional property on the `HttpTriggerSchema` (see https://github.com/spinframework/schemas.spinframework.dev/blob/19acd0235cff11cd4eda579d301fca6d345b9a3f/spin/manifest-v2/latest.json#L572)

Are there plans to deprecate that capability or is the schema just missing this? 

I've tested with latest Spin and it works as documented. However, my editor treats the `static_response` property as an error.

itowlson:
Fixes #3693 

(After merge we'll need to run the GH action to push an updated JSON schema.)

### DIFF
--- a/crates/manifest/src/schema/json_schema.rs
+++ b/crates/manifest/src/schema/json_schema.rs
@@ -30,6 +30,8 @@ pub struct HttpTriggerSchema {
     /// `components = { ... }`
     #[serde(default, skip_serializing_if = "Map::is_empty")]
     pub components: Map<String, OneOrManyComponentSpecs>,
+    /// `static_response = { ... }`
+    pub static_response: Option<StaticResponseSchema>,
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

thorstenhans (approved):
LGTM :shipit:

PR #3704, opened 2026-08-31
Break WASIp3 HTTP server ownership cycle

### LINKED ISSUE #3703: WASIp3 handler keeps the HTTP server alive
## What happens

After Spin handles a WASIp3 HTTP request, dropping the last external owner of the HTTP server does not release the server or its application state.

The HTTP server owns the WASIp3 handler state. That handler state stores a strong reference back to the HTTP server so that it can build new instances. These references keep each other alive.

This can leave old application state in memory when an embedder replaces an HTTP server.

I reproduced this on `main` at `bf863202c93422b174481633bd3e22e62f616b2f`. A regression test sends two WASIp3 requests, drops the in-process test environment, and then finds that a weak reference to the server can still be upgraded.

## Expected result

The HTTP server and its application state should be released after their external owners and active workers are gone.

## Suggested fix

Store a weak reference to the HTTP server in the WASIp3 handler state. Upgrade it when Spin creates a new WASIp3 worker. Return an error if the server has already gone away.

An active worker can still keep the server alive until that worker ends. This issue is about the permanent reference cycle that remains after the worker ends.

carsonfarmer:
Fixes #3703.

The WASIp3 handler state stored a strong reference to `HttpServer`. The server also owns that handler state. After the first WASIp3 request, these references kept each other alive and prevented the server and its application state from being released.

This change stores a weak reference in the handler state. Spin upgrades it while it creates a new WASIp3 worker. If the server is already gone, worker creation returns an error.

The regression test sends two single-use WASIp3 requests. It then drops the in-process server and verifies that the server was released. The test failed on upstream `main` before the production change and passes with this change.

Active workers can still keep the server alive until they end. This change only removes the permanent server-to-handler-to-server reference cycle.

No dependencies or services were added.

## Tests

- `make lint`
- `make build`
- `cargo test -p spin-trigger-http --no-fail-fast -- --nocapture`
- `cargo test --test integration wasi_http_p3 -- --nocapture`
- `cargo test --test integration wasip3_handler_does_not_retain_http_server -- --nocapture`

### DIFF
--- a/crates/trigger-http/src/server.rs
+++ b/crates/trigger-http/src/server.rs
@@ -4,7 +4,7 @@ use std::{
     marker::PhantomData,
     net::SocketAddr,
     pin::Pin,
-    sync::{Arc, OnceLock},
+    sync::{Arc, OnceLock, Weak},
     task::{Context, Poll},
     time::{Duration, Instant},
 };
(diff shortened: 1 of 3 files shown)


itowlson:
@carsonfarmer Thanks for the report and the PR.  It looks good to my inexpert eye, but I've flagged a couple of knowledgeable folks for a review.

For our guidance, how urgent is this for you?  And would you require it in a release, or do you pin to a SHA?

carsonfarmer:
I'm just pinning to sha, so no rush on this honestly.

dicej (approved):
Thanks so much for catching, fixing, and adding a test for this, @carsonfarmer!

PR #3707, opened 2026-09-01
Count TCP sockets from creation

### LINKED ISSUE #3706: TCP socket limits do not count unconnected sockets
## Problem

`max_socket_connections` is applied when a TCP socket starts connecting, not when the socket is created.

Creating a TCP socket already opens an operating-system socket. A guest can therefore create many unconnected sockets without using the configured limit.

I reproduced this on current `main` with a limit of 2: creating a third unconnected socket still succeeds. UDP sockets already count against the limit when they are created.

Could TCP sockets use the same lifecycle: acquire quota at creation and release it when the socket resource is dropped?

I have a small fix with P2 and P3 tests and am happy to submit it.

carsonfarmer:
Fixes #3706.

TCP socket creation already opens an operating-system socket, but Spin did not acquire quota until connect. Unconnected sockets could therefore bypass `max_socket_connections` and `max_total_connections`.

This change acquires quota when P2 or P3 creates the socket. The existing resource tracking releases it when the socket is dropped.

Tests cover both APIs, socket and instance cleanup, and shared limits.

## Tests

- `cargo test -p spin-factor-wasi`
- `cargo test -p spin-factor-outbound-networking`
- `make test-unit`
- `make lint`
- `make build`

### DIFF
--- a/crates/factor-outbound-networking/tests/factor_test.rs
+++ b/crates/factor-outbound-networking/tests/factor_test.rs
@@ -17,6 +17,7 @@ use wasmtime_wasi::p2::bindings::sockets::network::{ErrorCode, IpAddressFamily};
 use wasmtime_wasi::p2::bindings::sockets::tcp as p2_tcp;
 use wasmtime_wasi::p2::bindings::sockets::tcp_create_socket as p2_tcp_create;
 use wasmtime_wasi::p2::bindings::sockets::udp_create_socket as p2_udp_create;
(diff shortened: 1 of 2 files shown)


carsonfarmer:
~~Apologies, I think I can see a simpler way to handle this. Moving to draft until I can properly confirm!~~

Much better/simpler.

dicej (approved):
LGTM, thanks!  @rylev implemented the original connection limiting code, so it will be good to have his eyes on this.

PR #3708, opened 2026-09-02
SIP for dependency isolation

itowlson:
Technology POC with terrible names, poor formatting, and questionable consistency at https://github.com/itowlson/spin/tree/dep-isolation with a partial test case at https://github.com/itowlson/spin-dep-isolation-test.

(The POC doesn't yet have KV because Wasmtime 47 had a bug with named-importing interfaces with hyphens in them: Wasmtime 48 fixed that and I'll get the POC KVed up shortly.)

### DIFF
--- a/docs/content/sips/026-dependency-isolation.md
+++ b/docs/content/sips/026-dependency-isolation.md
@@ -1,0 +1,203 @@
+title = "SIP 026 - Dependency isolation"
+template = "main"
+date = "2026-09-02T00:00:00Z"
+---
+
+Summary: Enable dependency capabilities that are independent of the parent component.
+
+## Context
+
+Spin component dependencies want to be able to access resources such as configuration variables, databases, and remote APIs.
+However, until recently, the component model has enforced that if a composition imports an interface, then those
+imports are indistiguishable: all imports of, say, the `spin:sqlite/sqlite` interface get the exact same
+host implementation.
+
+The component model has now added an "implements" feature, which allows a component (for our purposes, a composition)
+to import an interface under a different name:
+
+```
+world w {
+    import my-db: spin:sqlite/sqlite;
+}
+```
+
+The host must mark the interface as `named_imports` in `wasmtime::component::bindgen!`: this generates additional,
+parallel `Host` traits which are the same as the normal host traits with an extra ID parameter (roughly, the name
+under which the guest imported the interface).
+
+## How does this help?
+
+Consider a situation where a primary component wants to access the `finance` database, and a dependency
+wants to access the `customer` database:
+
+```toml
+# hypothetical
+[component.shop]
+sqlite_databases = ["finance"]
+[component.shop.dependencies]
+"customer:customer/account" = { path = "...", capabilities = { sqlite_databases = ["customer"] } }
+```
+
+The naive composition graph looks like this. Note how the host sees only one import of the interface.
+
+```
+{----------------------- Wasm binary -------------------}
+
+[ main comp  ] <---
+                   \
+                    --- [ import spin:sqlite/sqlite ]
+                   /
+[ dependency ] <---
+```
+
+The idea is the during composition, we generate a nemed import of `spin:sqlite/sqlite`, and plug
+that onto the dependency:
+
+```
+{--------------------- Wasm binary -----------------}           {-------- synthesised by Spin --------}
+
+[ main comp  ] <--- [ import spin:sqlite/sqlite ]
+
+[ dependency ] <--- [ import spin:sqlite/sqlite ] <-- PLUG IN -- [ import custo: spin:sqlite/sqlite ]
+```
+
+Now the host sees separate named and unnamed imports of the interface.
+
+> Note that the _dependency_ is not aware of the named import. As far as the dependency is concerned,
+> it's still importing `spin:sqlite/sqlite`. The named import is a composition-time artefact which
+> is outside what the dependency can know.
+
+The host can now distinguish the dependency's import from the primary component's import, and can
+apply different validation rules to each.
+
+> In principle, the host could supply completely different implementations to each. That's not useful
+> here: Spin only cares about restricting access.
+
+We can do this by extending the behaviour of the existing capabilities code. Currently, the capabilities
+code is limited to composing a deny adapter onto a dependency import. The new behaviour would be, for each
+interface:
+
+* Is there an isolated capability for the interface? If so, compose a named import onto the interface. Otherwise:
+* Is there an inherit specified for the interface? If so, do nothing (as today). Otherwise:
+* Compose the deny adapter onto the interface (as today).
+
+## Named imports and capability sets
+
+The actual name of the named import doesn't matter: it is a private contract between the capabilities
+composer and the host trait. What matters is:
+
+1. Dependencies should not be able to forge import names. If a malicious developer crafts a component
+   which named-imports `super-secret-db-password`, and persuades me to use it in a composition,
+   Spin should not mistake the crafted import for an application-granted permission to the
+   super secret DB password.
+2. Spin must be able to map the import name - which is, roughly, what we get at the host trait - to
+   a capability set defined in the manifest. 
+
+So as an initial proposal, I suggest that we name the imports using everybody's favourite highly
+readable keys, GUIDs. That said, we may want something more debuggable in the vanishingly unlikely
+event that this doesn't work infallibly first time - perhaps a combination of a GUID for unpredictability
+with an identifying string for tracing back to what it actually represents. We can iterate on this.
+
+My initial sense is that a key will correspond to a capability and be dependency scoped - for example,
+a key for the `sqlite_databases` capability on the `foo` dependency, a key for the the `key_value_stores`
+capability on the `foo` dependency, a key for the `sqlite_databases` capability on the `bar` dependency.
+This makes point 2, mapping keys to capability sets, easier. But because a capability
+(`allowed_outbound_hosts`) can enable multiple interfaces, we will need
+to further distinguish named imports by interface.  We therefore end up with:
+
+* capability set key (CSK): identifies a specific capability set (AOH on `foo`, sqlite on `bar`)
+* named import key: identifies a specific named import (`foo`'s import of `spin:postgres`)
+
+The host trait will receive the _named import key_.  Therefore, it must be possible to derive
+the capability set key from the named import key (as the capability set key is what we use to
+look up capability sets).
+
+Because the CSKs are not predictable ones like database names, we will need to maintain a separate
+map of CSKs to capability sets. The flow here is something like:
+
+* When loading a component, generate a CSK for each combination of dep and capability,
+  and record this in the lockfile.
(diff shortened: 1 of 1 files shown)


docs/content/sips/026-dependency-isolation.md
```diff
@@ -0,0 +1,178 @@
...
+
+```
+{--------------------- Wasm binary -----------------}           {-------- synthesised by Spin --------}
+
+[ main comp  ] <--- [ import spin:sqlite/sqlite ]
+
+[ dependency ] <--- [ import spin:sqlite/sqlite ] <-- WAC PLUG -- [ import custo: spin:sqlite/sqlite ]
+```
+
+Now the host seems separate named and unnamed imports of the interface.
```
dicej:
```suggestion
Now the host sees separate named and unnamed imports of the interface.
```

docs/content/sips/026-dependency-isolation.md
```diff
@@ -0,0 +1,178 @@
...
+
+With this in place, when the factor handles an import via its named import host trait:
+
+* It derives the CSK from the named import key (received via the trait method)
+* It looks up the CSK in the instance state, and gets the capability set
+* It verifies access against that capability set (instead of against the primary component capabilities)
+  * This is the exact same behaviour as normal access validation, just against a different list.
+    It returns the same errors etc.
+
+> An alternative approach is to have a per-dependency instance state, which is potentially
```
dicej, line 136: Leaving it up to the factor to decide what parts of the instance state should be shared across dependencies and what parts should be dependency-specific feels like the most flexible option, so I agree with your conclusion here.

docs/content/sips/026-dependency-isolation.md
```diff
@@ -0,0 +1,178 @@
...
+* Is there an isolated capability for the interface? If so, compose a named import onto the interface. Otherwise:
+* Is there an inherit specified for the interface? If so, do nothing (as today). Otherwise:
+* Compose the deny adapter onto the interface (as today).
+
+## Named imports and capability sets
+
+The actual name of the named import doesn't matter: it is a private contract between the capabilities
+composer and the host trait. What matters is:
+
+1. Dependencies should not be able to forge import names. If a malicious developer crafts a component
```
dicej, line 89: I'm not sure I understand the value the GUIDs would add here.  If we were to deny-adapt dependency imports by default (regardless of whether they're named imports or not) and require that the app developer specify explicitly which imports they want to make available to the dependency in a restricted or unrestricted form, there would be no way for the dependency to forge anything.

In your example, a dependency that named-imports `super-secret-db-password` would have that import denied by default, and the app developer would have to explicitly indicate if they wanted it hooked up to anything.

I guess I don't understand what the "and persuades me to use it in a composition" scenario looks like and how GUIDs would help avoid it.

docs/content/sips/026-dependency-isolation.md
```diff
@@ -0,0 +1,178 @@
...
+Spin doesn't implement the WASI interfaces: instead it hands off to the `wasmtime_wasi` and
+`wasmtime_wasi_http` crates. And these do not implement named imports.
+
+If we add, say, `wasi:http/handler` to the `spin_world` `named_imports`, we do get something,
+but the something is unwieldy: we have to implement every HTTP host trait, using a set of
+types which is separate from the `wasmtime_wasi_http` ones. And we have to link those traits,
+which is a bit of a dark art in itself given `WasiCtx` and `WasiCtxView` and Uncle Tom Cobbley
+and all.
+
+My proposed solution to this is to go cap in hand to Wasmtime and ask them to do it. We
```
dicej, line 175: Agreed!

dicej (commented):
Thanks for writing this up!

alexcrichton (approved):
Agreed with all the conclusions here and it all seems pretty reasonable to me 👍

docs/content/sips/026-dependency-isolation.md
```diff
@@ -0,0 +1,178 @@
...
+{----------------------- Wasm binary -------------------}
+
+[ main comp  ] <---
+                   \
+                    --- [ import spin:sqlite/sqlite ]
+                   /
+[ dependency ] <---
+```
+
+The idea is the during composition, we generate a nemed import of `spin:sqlite/sqlite`, and `wac plug`
```
fibonacci1729: nit: we probably shouldn't use `wac plug` specifically because it runs the risk of over-plugging. I think it suffices to say "wire up" or "plug/compose" here.

lann:
This looks viable for the current Spin interfaces, but since @tschneidereit isn't here to do it I'll point out that in the long run it would be great to use named imports for this directly, replacing the pattern of `open("my-store")` with an import named `my-store`.

In theory we could even make this additive with existing interfaces, e.g. 

```wit
package spin:key-value@3.1.0;

interface opener {
  open: async func() -> result<store, error>;
}

interface key-value {
  resource store {
    // This stays, but can't be isolated!
    open: static async func(label: string) -> result<store, error>;
```

Dependency isolation would then be pushed back to composition time rather than run time.

docs/content/sips/026-dependency-isolation.md
```diff
@@ -0,0 +1,178 @@
...
+* named import key: identifies a specific named import (`foo`'s import of `spin:postgres`)
+
+The host trait will receive the _named import key_.  Therefore, it must be possible to derive
+the capability set key from the named import key (as the capability set key is what we use to
+look up capability sets).
+
+Because the CSKs are not predictable ones like database names, we will need to maintain a separate
+map of CSKs to capability sets. The flow here is something like:
+
+* When loading a component, generate a CSK for each combination of dep and capability,
```
fibonacci1729, line 118: IIUC random GUIDs will make composition non-deterministic. This means the composed Wasm differs on every `spin build`. Given that, I'd suggest a deterministic structured key derived from (component id, dependency name, capability, interface) -- e.g. `spin-cap:<component>/<dep>/<capabilitiy>/<interface>`.

docs/content/sips/026-dependency-isolation.md
itowlson: Hmm, maybe you're right.  I am not sure how the deny adapter would apply itself to named imports.  But I worry there is still a case where a dependency has a legitimate import (dependency imports SQLite and this is known and allowed), which will be mapped to a named import, but the dependency also _forges_ a pre-named import of the same interface.  In the prototype, I link _all_ the dep's named imports.  But this could be somewhat mitigated by passing the list of "names we made up" and only linking those.  But I still worry that there is a more-complicated-still scenario where Evil Dep forges a name that it guesses will be used by Good Dep and we can't tell them apart.  Maybe we can do additional decoration to avoid this.

But basically my worry is kind of if we trust names as indicators of permissions, we are trusting something potentially controlled by an attacker.  Perhaps I am seeing phantoms!

itowlson:
@lann Yes, I thought about that, but that involves more rework of the SDKs than I suspect we want to take on as part of this.

The flow there is certainly simpler: we validate each dependency's set of named imports agains `spin.toml`, then compose naively.  I think we still have considerations around name collisions, although I guess we have those in the current proposal too.  If folks think this is the way to go then I can have a tinker.

docs/content/sips/026-dependency-isolation.md
itowlson: If we're confident that the forgery concern is a non-issue then I'm happy to go back to this (this was what the prototype did until I got in a funk over predictable names).

docs/content/sips/026-dependency-isolation.md
dicej: To me, the key is that we would deny-adapt _all_ dependency imports by default, regardless of whether they are named or not, and require the app manifest to specify explicitly which imports they want to connect to the host, if any.  A dep that named-imports `super-secret-db-password` would only get access to a host resource of the same name if the app explicitly says it should by saying (in TOML-ese) "connect dep $x's `super-secret-db-password` to the host's `super-secret-db-password`".  Alternatively, the app manifest could say "connect dep $x's `super-secret-db-password` to the host's `fake-db-password-ha-ha`".  But unless the app manifest explicitly says what to connect that import to, it just gets denied.

The metaphor I'm thinking of is mobile (e.g. Android) app permissions.  An app can say "I need access to your camera to do my job, may I?", but you as the user get to decide (and are indeed _required_ to decide) what the answer is.  Similarly, a dep component can say "I need access to such-and-such resource to do my job, may I?", and the app author can say "yes", "no" (the default), or "here's something else with the right shape (but maybe fake), make do".  Indeed some aftermarket mobile OSes let you do the equivalent of the last option by giving an app access to fake location data when it asks for your location.

itowlson: Okay, I see where you're coming from.  We proactively deny all named imports on each dep, and then plug unnamed imports to named imports.  This will work today: I guess in the future Till envisages and Lann notes, we will want/need to rethink all this stuff anyway.

So this allows us to trust names, as any attacker controlled names will be squished (even in precomposition scenarios), and we can ditch the GUIDs in favour of predictable (or even reversible) names.

itowlson: @fibonacci1729 I'm trying to POC the "proactively deny all named imports on each dep" and am not sure how to identify them.  If a component imports `ns:nm/itf@1.2.3` both normally _and_ as a named import called `spork`, then the WAC imports list (`&graph.types()[graph[pkgid].ty()].imports`, as we use in the existing capabilities code) shows:~

| Name   | Kind  |
|-------|-------|
| ns:nm/itf@1.2.3 | Instance (InterfaceId 0) |
| spork | Instance (InterfaceId 1) |

~There doesn't seem to be anything in this that lets me see "okay `spork` is a named import of `ns:nm/itf`, I need to squish it."  I'm on WAC main.  Am I missing something?  Do I need to find the named imports some other way?

itowlson: Oh, never mind: I think it doesn't matter, because the idea is to scan _all_ the dep's imports, and squish _everything_ that doesn't match an interface name that is permitted for inheritance or capabilitification.  We don't need to care if the import is named or vanilla.

itowlson: No, I might have been right, there might be an issue: we don't know what type of deny plug to squish into the named import socket.  Is that what the subtype checking in `apply_deny_adapter` (https://github.com/spinframework/spin/blob/3b479bcad9b012f2c09e959b5add9615ab54eb02/crates/capabilities/src/deny.rs#L69) is handling?  Non-inherited implementations get THE SQUISH even if they try to hide under a different name?

Basically is there work I need to do to squish named imports of Spin/WASI interfaces or does Spin already do that, either by masterful planning or by good fortune?

fibonacci1729: Yeah you're right. We'll need to invert the loop to iterate the dependency's imports, resolve each instance import to its implemented interface, and decide allow/deny on that name while plugging by label.

EDIT: https://github.com/spinframework/spin/pull/3711

itowlson:
Looking at Alex's Wasmtime PR it occurs to me I could stop fartarsing around with digests and just give the named imports sequential numbers like `spin-capability-set-n43-i12`, fie.

PR #3710, opened 2026-09-05
Bump Wasmtime to v49.0.0 prerelease

spinframeworkbot:
Bumps Wasmtime to v49.0.0 prerelease

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -204,9 +204,9 @@ wasm-metadata = "0.247.0"
 wasm-pkg-client = "0.11"
 wasm-pkg-common = "0.11"
 wasmparser = "0.247.0"
-wasmtime = "48.0.0"
-wasmtime-wasi = { version = "48.0.0", features = ["p3"] }
-wasmtime-wasi-http = { version = "48.0.0", features = ["p3", "component-model-async"] }
+wasmtime = { git = "https://github.com/bytecodealliance/wasmtime", branch = "release-49.0.0" }
+wasmtime-wasi = { git = "https://github.com/bytecodealliance/wasmtime", branch = "release-49.0.0", features = ["p3"] }
+wasmtime-wasi-http = { git = "https://github.com/bytecodealliance/wasmtime", branch = "release-49.0.0", features = ["p3", "component-model-async"] }
 wit-component = "0.247.0"
 wit-parser = "0.247.0"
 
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -204,9 +204,9 @@ wasm-metadata = "0.247.0"
 wasm-pkg-client = "0.11"
 wasm-pkg-common = "0.11"
 wasmparser = "0.247.0"
-wasmtime = "48.0.0"
-wasmtime-wasi = { version = "48.0.0", features = ["p3"] }
-wasmtime-wasi-http = { version = "48.0.0", features = ["p3", "component-model-async"] }
+wasmtime = { git = "https://github.com/bytecodealliance/wasmtime", branch = "release-49.0.0" }
+wasmtime-wasi = { git = "https://github.com/bytecodealliance/wasmtime", branch = "release-49.0.0", features = ["p3"] }
+wasmtime-wasi-http = { git = "https://github.com/bytecodealliance/wasmtime", branch = "release-49.0.0", features = ["p3", "component-model-async"] }
 wit-component = "0.247.0"
```
alexcrichton, line 210: I realize this is automation-generated but one possible tweak, now that this is the first release this is in effect, would be to [use 49.0.0-rc.1 on crates.io](https://crates.io/crates/wasmtime/49.0.0-rc.1) which would avoid the need for git deps

alexcrichton (approved):
lgtm, although if more is needed other than a Rust update in CI I can look more closely at the other changes too

itowlson:
I've re-run the tests in case of flakiness, but `internal_http_p3_p2` (service chaining across P3 / P2 boundary, https://github.com/spinframework/spin/tree/main/tests/runtime-tests/tests/internal-http-p3-p2) seems to be failing consistently.  (Both failed jobs are the same test, just different OSes.)  My only changes were sync/async chores in the sockets stack, so if this is real then I think it is likely down to something in Wasmtime.  Giving it one more for luck but I think this might need some @dicej / @alexcrichton love - thanks!

alexcrichton:
Definitely looks like a Wasmtime-side issue, I'll poke at it today

alexcrichton:
Ok I've attempted a draft fix for that issue at https://github.com/bytecodealliance/wasmtime/pull/14302, will report back here when that's ready.

alexcrichton:
Ok that PR is now merged so if you update the git rev of the release-49.0.0 branch, I think CI might then be passing

itowlson:
Thanks Alex.  Updated!

itowlson:
further bumpings for a Cargo.lock conflict

people I must insist no merging to main from the 5th of the month to the 20th

itowlson (approved):

PR #3711, opened 2026-09-08
Deny named imports

fibonacci1729:
A dependency importing one interface under several labels now has all of them denied (previously at most one could be plugged).


itowlson (approved):
I admit to not really understanding the main flow, but the tests make it very clear what's going on - thanks for writing them and for finding the right helpers to make them concise and clear!  So LGTM when you feel ready.

PR #3716, opened 2026-09-14
Fix WIT extractor emitting wrong interface when duplicated name

### LINKED ISSUE #3715: spin build: dependency WIT picks the wrong interface when a dependency's own imports share the interface name
Filing this per @itowlson's request on spinframework/spin-docs#259.

`spin build` writes the wrong interface into the `root` world of the generated `spin-dependencies.wit` when the dependency component's *own imports* include an interface with the same bare name as the one named in `[component.*.dependencies]`. `spin_sdk::dependencies!()` then generates bindings for that wrong interface and the app fails to compile, with an error that points nowhere near the cause.

Composition itself is unaffected — `spin up` works, since Spin composes against the real component imports rather than the generated file. It is only the generated WIT, and therefore `dependencies!()`, that is wrong.

* Spin version: `spin 4.1.0`
* Plugins: none relevant (stock install)
* Also present on `main` as of `7ff011c`

### Repro

Minimal repro, two throwaway components: **https://gist.github.com/mileszim/179ad82be60345c206b6757e567bb0fb** — `bash repro.sh`, needs `spin`, `cargo`, `wasm32-wasip2`.

The dependency exports `example:dep/store@0.1.0` and imports `wasi:config/store@0.2.0-draft-2024-09-27`. Both interfaces are named `store`; that is the whole trigger. The app declares it the ordinary way:

```toml
[component.app.dependencies]
"example:dep/store" = { path = "../dep/target/wasm32-wasip2/release/example_dep.wasm" }
```

Expected `spin-dependencies.wit`:

```wit
world root {
  import example:dep/store@0.1.0;
}
```

Actual:

```wit
package root:component;

world root {
  import wasi:config/store@0.2.0-draft-2024-09-27;
}
world impo-world0 {
  import example:dep/store@0.1.0;
}
```

`root` — the world `spin_sdk::dependencies!()` generates from — got the dependency's own `wasi:config` import, so the app fails with:

```
error[E0433]: cannot find module or crate `example` in this scope
```

### Why

[`one_import`](https://github.com/spinframework/spin/blob/7ff011c40ac6bd6e242b063c212967f24beaa601/crates/dependency-wit/src/lib.rs#L389-L398) searches every interface in the resolve decoded from the dependency binary, and matches on the bare interface name:

```rust
.find(|i| i.1.name == Some(name.to_string()))
```

That resolve holds the interfaces of the dependency's imports as well as its exports, so the match lands on whichever interface called `store` sits earlier in the arena — here `wasi:config/store` at index 0, against `example:dep/store` at 15. A component that happens to get the right one is getting it by arena order rather than by matching. One way to fix it would be to match on the fully qualified name; the manifest key already spells it out, and `ImportKind::Interface` currently carries only the bare `KebabId` while the package sits unused in the `DependencyName::Package` beside it.

This is not an exotic shape — `store`, `types`, `handler` and `config` are common interface names, and any dependency that itself uses `wasi:config`, `wasi:keyvalue` or `wasi:blobstore` brings a `store` into the resolve.

### Where we hit it

A secrets component that exports `seekrit:secrets/store@0.1.0` and imports `wasi:config/store@0.2.0-draft-2024-09-27` — same collision, same result, observed on Spin 4.1.0.

* Guide and manifest: https://seekrit.dev/docs/guides/spin
* Component: `ghcr.io/seekritdev/secrets-spin:0.3.0`
* A full sample app using it: spinframework/spin-docs#259

### Workaround

Declare the import in your own `wit/` and call `spin_sdk::wit_bindgen::generate!` instead of `spin_sdk::dependencies!()`. It costs one WIT file, and is what the docs PR currently tells readers to do — the two entries there describe the bug so a reader hitting `E0433` knows why.

itowlson:
Fixes #3715.

When a dep imports and exports distinct interfaces with the same _unqualified_ name (but from different packages), the extractor could pick an imported one.  In the original bug report, for example, a component exported its own secrets `store` interface, but imported a WASI config `store`.

This PR limits the extractor to considering only exported interfaces (yes, it calls them imports, but that means after the dep has been importised).  I am still not sure if there is a different edge case where a dep exports two interfaces with the same unqualified name - @fibonacci1729 I'd love your advice here as my brain is all muddled about it, although we can probably punt that to a separate PR.

@mileszim I have a unit test which I believe replicates your scenario, but if you have bandwidth to test this against your real code, that would help with validation.  Thanks for the detailed bug report (and for taking the time to port it from the docs repo)!


mileszim:
@itowlson looks like its working! amazing!

crates/dependency-wit/src/lib.rs
```diff
@@ -840,4 +863,94 @@ world colors {
...
+            DependencyName::Package("my:impexp/zounds@1.0.0".to_string().try_into().unwrap());
+        let dep_src = ComponentDependency::Local {
+            path: dep_file,
+            export: None,
+            inherit_configuration: None,
+        };
+        let deps = std::iter::once((&dep_name, &dep_src));
+
+        let wit = extract_wits(deps, ".").await?;
+        eprintln!("{wit}");
```
dicej: Stray debug println?

crates/dependency-wit/src/lib.rs
```diff
@@ -387,11 +387,21 @@ fn as_func(wi: &wit_parser::WorldItem) -> Option<&wit_parser::Function> {
...
     let id = wasm
         .resolve()
-        .interfaces
+        .worlds
         .iter()
-        .find(|i| i.1.name == Some(name.to_string()))
+        .flat_map(|(_wid, w)| w.imports.values())
+        .flat_map(as_interface)
+        .flat_map(|iid| find_interface(wasm.resolve(), iid))
+        .find(|itf| itf.1.name == Some(name.to_string()))
```
dicej: Nit: could avoid an allocation by turning this into e.g. `itf.1.name.as_deref() == Some(name)`

dicej (approved):
I haven't looked at this code before, so I don't have full context for it, but the change looks reasonable to me.

PR #3717, opened 2026-09-15
Add schema directive to empty template

### LINKED ISSUE #3698: `spin add` to an empty manifest prints target env warning
```sh
$ spin new http-empty my-test -a

$ cd my-test

$ spin add http-rust my-comp -a
Warning: Couldn't determine target environment; using your installed templates.

$ spin add http-rust my-other-comp -a
# no warning
```

Looking at the code, the only way this seems like it could happen is if the empty template fails to TOML-deserialise as an app manifest.  My guess is that the `triggers` element is required at the serialiser level - looking at the code it doesn't have a `#[serde(default)]` on it.  If so, maybe we should move that to a validation step, rather than making it "a manifest with no triggers is syntactically malformed, and we won't even load it, no, not even for non-execution purposes, how dare you."

### LINKED ISSUE #3699: `http-empty` template doesn't have manifest schema
Most `spin.toml` templates now have a Taplo directive for the manifest schema.  `http-empty` doesn't.

(That said, it's possible that `http-empty` isn't compliant, because of the mandatory trigger thing.  So maybe we shouldn't reference a schema that will make it go red.  On the other hand, as soon as you've done your first `spin add`, it's vexing that it's not there.  And maybe it's okay for the schema to say "you're missing this required thing.")

itowlson:
Fixes #3699.

The current `AppManifest` serialisation type _requires_ that there be at least one trigger.  This makes the empty manifest invalid under the current schema.

I've proposed to fix this by allowing an empty set of triggers at the serialisation level, but validating during load.  This changes the error output when you `spin up` an app with no triggers:

```
# Before

$ spin up
Error: TOML parse error at line 1, column 1
  |
1 | spin_manifest_version = 2
  | ^^^^^^^^^^^^^^^^^^^^^^^^^
missing field `trigger`

# After

$ spin up

Caused by:
   2: The application must have at least one trigger
```

The alternative is to keep the existing requirement and accept that there will be red squigglies in a new empty app.  However, like the "before" error, the message will be correct but it will point to the wrong place, so I have pretty mixed feelings about this.

Note that we will want to publish a new schema directive before tagging this template - at the moment we merge it, the directive will continue to point to the "trigger required" schema, which fails.  But as noted, the failure at least comes with a good error message, so eh.


itowlson:
Pleasingly, this also fixes #3698 _does the bug smiting dance_

fibonacci1729 (approved):

PR #3719, opened 2026-09-16
Add OpenBao Variable Provider

thorstenhans:
The underlying [`vaultrs`](https://docs.rs/vaultrs/latest/vaultrs/) allows us connecting to Vault and OpenBao. I've added dedicated provider and shared the impl between Vault and OpenBao. 

Although this being mainly sugar coating on top of the already existing implementation, it's worth it IMO. OpenBao is a truly community driven alternative under the CNCF umbrella umbrella besides Vault that should be supported.


lann:
Is there any difference between the two implementations? If not then perhaps this could just be a `#[serde(alias = "openbao")]` on the `VariableProviderConfiguration::Vault` variant?

lann:
or rename everything `Vault` -> `OpenBao` and make it `#[alias = "vault"]` - I don't really have a strong opinion either way 🙂

thorstenhans:
@lann 🤦🏻 yeah.. obviously the way smarter approach for the implementation. Haven't considered that as an feasible approach... Let' me test locally before refactoring it in that way.

PR #3720, opened 2026-09-17
Parameterize Azure Cosmos key-value queries

ChihweiLHBird:
Keys and store ids were interpolated into the Cosmos SQL text, so a key containing `'` could break out of the literal and read or match other rows. This binds them as query parameters via `Query::with_params` instead.

Verified against a live Cosmos DB container through `get`, `exists`, `get_keys`, `get_many`, `increment` and compare-and-swap with quoted and injection-shaped keys.


itowlson (approved):

itowlson:
Nice!  Thanks for finding this!

